PR #19, opened 2019-09-06
Fixed build error on mising file

lzap:
I have noticed an error when building the project. The attributes file was including file named "build" which no longer exists since build option is now passed via command line. This is a regression introduced in 8773d0d4d57cf64680d88121d0bb424b669cdb78.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+---
+language: minimal
+before_install:
+  - sudo apt-get install -y ruby make
+install: gem install asciidoctor
+script:
+  - cd guides
+  - make clean html BUILD=satellite
+  - make clean html BUILD=foreman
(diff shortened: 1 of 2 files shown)


lzap:
@melcorr when you checkout this branch, delete leftover file `guides/common/build*` and do `make clean` and then `make html`. Does it work for you?

ehelms:
This is now turned on -- https://travis-ci.org/theforeman/foreman-documentation

ehelms:
You may need to force push to re-trigger

lzap:
Great, there is now CI pipeline which builds the documentation both upstream and downstream. It only builds it and does not upload the artifacts anywhere but it is a good start. At least I will not be pushing breaking changes from now on. Merging!

melcorr:
Wow, thank you!

PR #22, opened 2019-09-13
Added support for MacOS

lzap:
Two changes needed in the Makefile in order to build on MacOS. I use my old Macbook for "evening browsing", so will likely be using that to do some work here and there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.DS_Store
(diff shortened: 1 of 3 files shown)


rahulbajaj0509:
Looks straight forward!  Merging this :)

Thanks @lzap!

PR #25, opened 2019-09-14
Upsteamize app-host-params chapter

rahulbajaj0509:


rahulbajaj0509:
One more thing to note in this chapter is, on line number 103, we have `Red Hat Openstack Platform Parameters`. Do we need only OpenStack for the upstream or this is fine?

lzap:
Yes, please introduce variable named OpenStack and for upstream it should be just OpenStack and downstream the full product name Red Hat OpenStack Platform. There do not appear other forms of this product name in this guide so we should be good. Other than that looks good.

rahulbajaj0509:
@lzap I have also arranged the attributes in alphabetical order, this creates no confusion on where to add a new attribute :)

lzap:
No English involved, merging. :-)

PR #26, opened 2019-09-17
Infoblox upstreamization

lzap:
Easy peasy, one thing I hate hostname `Infoblox_hostname` I vote for changing this to simply `infoblox-host`. Issues:

* hostnames should always be lowercase
* dashes preferred over underscore
* it is a host not hostname


melcorr:
Hi Lzap, if you raise those as an Issue, I'll do it upstream and downstream.

melcorr:
Yeah, you're right!

PR #27, opened 2019-09-19
Correct the Google resolver IP

ekohl:
Per https://developers.google.com/speed/public-dns/

I do wonder if we should point to possible other services. There's [Quad9's 9.9.9.9](https://www.quad9.net/) and [Cloudflare's 1.1.1.1](https://www.cloudflare.com/learning/dns/what-is-1.1.1.1/) as well.

Technically forwarders aren't even needed.


melcorr:
If it needs to be fixed, it should be fixed :

melcorr:
Thanks

melcorr:
Created downstream ticket to resolve this there too: 

https://bugzilla.redhat.com/show_bug.cgi?id=1754004

PR #28, opened 2019-09-19
[WIP] upstreamize fips topic

melcorr:
Upstreamized content for FIPS templates

### DIFF
--- a/guides/doc-Provisioning_Guide/topics/app-Deploy_FIPS_hosts.adoc
+++ b/guides/doc-Provisioning_Guide/topics/app-Deploy_FIPS_hosts.adoc
@@ -38,7 +38,7 @@ Before creating the FIPS-compliant templates in {Project}, you must identify tho
 ----
 $ hammer location list
 ---|-----------------
-ID | NAME            
+ID | NAME
 ---|-----------------
 2  | Default Location
 ---|-----------------
@@ -50,11 +50,12 @@ Note the value in the `NAME` column of those locations to which you want to depl
 +
 .Example
 ----
+$ hammer organization list
 ---|----------------------|----------------------|------------
 ID | NAME                 | LABEL                | DESCRIPTION
 ---|----------------------|----------------------|------------
-1  | Default Organization | Default_Organization |            
-2  | Sales                | Sales_Department     |            
+1  | Default Organization | Default_Organization |
+2  | Sales                | Sales_Department     |
 ---|----------------------|----------------------|------------
 ----
 +
@@ -65,14 +66,6 @@ Note the value in the `NAME` column of those organizations to which you want to 
 .Example
 ----
 $ hammer os list
----|-----------------|--------------|-------
-ID | TITLE           | RELEASE NAME | FAMILY
----|-----------------|--------------|-------
-2  | RedHat 6.6      |              | Redhat
-3  | RedHat 7.1      |              | Redhat
-1  | RedHat 7.2      |              | Redhat
-4  | RedHat 6.7      |              | Redhat
----|-----------------|--------------|-------
 ----
 +
 Note the value in the `TITLE` column of those operating systems to which you want to deploy FIPS-compliant hosts.
@@ -94,7 +87,7 @@ This repository contains the following Embedded RuBy (ERB) templates. These are 
   ** Updated PXELinux template
 * `fips_packages.erb`
   ** Packages required by FIPS mode (for example, `dracut-fips`)
-* `{Project}_Kickstart_Default_FIPS.erb`
+* `Kickstart_Default_FIPS.erb`
   ** Kickstart template with modifications to call the `fips_packages` snippet
 * `puppet.conf.erb`
   ** Updated `puppet.conf` configuration file with updated (SHA256) message digest algorithm
@@ -112,7 +105,7 @@ $ hammer template create  --name "Kickstart Default PXELinux FIPS" \
   --type PXELinux
 ----
 +
-Replace the placeholder values `LOCATIONS`, `ORGANIZATION`, and `OS` with the values you noted in xref:identifying_relevant_operating_systems_locations_organizations[]. If any value contains non-aphabetical characters, enclose the value in quotation marks (").
+Replace the placeholder values `LOCATIONS`, `ORGANIZATION`, and `OS` with the values you noted in xref:identifying_relevant_operating_systems_locations_organizations[]. If any value contains non-alphabetical characters, enclose the value in quotation marks (").
 +
 The message `Config template created` indicates success.
 +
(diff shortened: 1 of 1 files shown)


melcorr:
Hi @lzap I went ahead with the FIPS provisioning templates section, there is one section that points to templates in the Satellite github account. I have wrapped this in an ifeval but am wondering is there a repository of Foreman FIPS provisioning templates? If not, maybe these docs won't work here.

guides/doc-Provisioning_Guide/topics/app-Deploy_FIPS_hosts.adoc
```diff
@@ -83,10 +76,12 @@ The FIPS provisioning templates are provided in a git repository. In this proced
 
 . On the {ProjectServer}, clone the git repository containing the FIPS enabled templates, then change into the repository's directory.
 +
+ifeval::["{Build}" == "satellite"]
```
ekohl: Why are these satellite-only? I'd expect them to work on Foreman as well. My only question is why these are not part of community-templates so we ship them to every user.

The whole section refers to this repository and doesn't really make sense without it.

@lzap can we just integrate these templates upstream which means we'll eventually ship them out of the box?

melcorr: Because it is the Satellite github @ekohl - I need the link for the Foreman community templates, hence the WIP :D

melcorr: PS: they did not make it into the regular shipped templates in Satellite, so we were asked to link to the Github page. I am relying on you a bit to let me know whether they are in Foreman, if so, I can write a Foreman-specific section.

ekohl: While it may be the Satellite org, this is publicly available so I'd release this without hiding it. The proper upstreaming would probably mean they become part of all the existing files and there would be no need for anything other than setting the `fips_enabled` parameter to `true` on the host to provision. That should then flow back into downstream and the Satellite customer would only need to set that one parameter. At least, that's how I'd see this feature if it was properly finished.

melcorr: By that logic, I think most of the changes in this PR can be reverted, because all the templates will have Satellite rather than Foreman if we are just pulling straight from the Satellite github and if none of the templates are available in the upstream files. 

By the sounds of it, what is required is an ifeval with a note for Foreman saying that they can use the Satellite resources for this. 

What do you think @ekohl @lzap ?

ekohl: So it looks like what I suggested has already happened and this entire chapter is downstream-only. When those changes make it into downstream, the entire chapter can also be dropped.

ekohl:
Looking at the Puppet part that might still use some improvement to enforce sha256 on all systems if that is indeed required.

guides/doc-Provisioning_Guide/topics/app-Deploy_FIPS_hosts.adoc
lzap: Well I think this is just a proposal and we can come to agreement. It does not make any sense to hide it.

lzap:
If you spot any technical change or mistake, please file an issue in this repo. We want to have those PRs only focused on upstreamizing the process. Thank you.

lzap:
I just noticed that FIPS templates have been merged into community repo. I will go ahead and update README and I am creating a ticket for us to update this chapter from scratch. All templates are now completely different, also there is no need to checkout any git: https://github.com/theforeman/foreman-documentation/issues/33

ekohl:
The line:
> Edit the `/etc/puppet/puppet.conf` file, adding the line `digest_algorithm = sha256` in the `[main]` stanza.

In upstream there will already be a block with `digest_algorithm = sha256` (https://github.com/theforeman/community-templates/blob/1b2d5ebf427422b0572f58f94f2fd960816788ed/provisioning_templates/snippet/puppet.conf.erb#L43-L45). At least the instruction should be changed but if we add a `puppet_use_sha256_digest_algorithm` parameter (or similar) then we can have easier user instructions.

> I just noticed that FIPS templates have been merged into community repo. I will go ahead and update README and I am creating a ticket for us to update this chapter from scratch. All templates are now completely different, also there is no need to checkout any git: #33

:+1: That was my point as well about hiding the chapter. It's good to rewrite it from scratch because 80% of the content will be different.

lzap:
Absolutely, not in this PR tho. It's still valid even for Foreman - you can grab those templates from the git repo and use it that way. At least until we refactor this content.

ekohl:
In this PR you could say something like:
> Ensure `digest_algorithm = sha256` is in the `[main]` stanza

melcorr:
@ekohl and @lzap sorry for the break. I am a bit confused: do we need this upstream section or not? If the templates exist in the upstream, we should just add information saying to 

> set the `fips_enabled` parameter to `true` on the host to provision

lzap:
In this PR please let's focus on upstreaming this. Let's not hide any sections or the whole chapter and for now keep it pointing to the Satellite6 github repo. Let's create an issue to rework the chapter from ground up for Satellite 6.6 because templates are not part of upstream.

guides/doc-Provisioning_Guide/topics/app-Deploy_FIPS_hosts.adoc
```diff
@@ -199,16 +194,20 @@ Now that the new FIPS templates have been added to {Project}, they must be set a
...
-54 | Satellite Kickstart Default FIPS      | provision
-47 | Satellite Kickstart Default User Data | user_data
-50 | subscription_manager_registration     | snippet  
+46 | {Project} Kickstart Default           | provision
+48 | {Project} Kickstart Default Finish    | finish
+54 | {Project} Kickstart Default FIPS      | provision
+47 | {Project} Kickstart Default User Data | user_data
+ifeval::["{Build}" == "satellite"]
+50 | subscription_manager_registration     | snippet
+endif::[]
```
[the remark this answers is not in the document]
lzap: The example output can be done in the chapter rewrite later on, just make sure you don't hide anything from it.

guides/doc-Provisioning_Guide/topics/app-Deploy_FIPS_hosts.adoc
```diff
@@ -65,14 +66,6 @@ Note the value in the `NAME` column of those organizations to which you want to
...
 ----
 $ hammer os list
----|-----------------|--------------|-------
-ID | TITLE           | RELEASE NAME | FAMILY
----|-----------------|--------------|-------
-2  | RedHat 6.6      |              | Redhat
-3  | RedHat 7.1      |              | Redhat
-1  | RedHat 7.2      |              | Redhat
-4  | RedHat 6.7      |              | Redhat
----|-----------------|--------------|-------
```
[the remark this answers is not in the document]
melcorr: Yeah I just wanted to get rid of the output. In docs, we only display output if it is complex and needs explaining. This is simple.

PR #40, opened 2019-10-02
Add note about RHEL 8 repositories

melcorr:
As part of

Bug 1757152 - Need to add a note in Satellite Provisioning Guide, as to
which repositories should be included in the synced Kickstart Repository
for a rhel 8 Operating System

https://bugzilla.redhat.com/show_bug.cgi?id=1757152


lzap:
Looks good.

Issue #32, opened 2019-09-20
Replace Infoblox_hostname with infoblox_hostname


lzap:
I hate hostname Infoblox_hostname I vote for changing this to simply infoblox-host. Issues:

*    hostnames should always be lowercase
*    dashes preferred over underscore
*    it is a host not hostname

PR #59, opened 2019-10-10
s/Infoblox_hostname/infoblox_hostname

melcorr:
As part of Issue #32

### DIFF
--- a/guides/doc-Provisioning_Guide/topics/Infoblox_Integration.adoc
+++ b/guides/doc-Provisioning_Guide/topics/Infoblox_Integration.adoc
@@ -34,18 +34,18 @@ You can download the certificate from the Infoblox web UI, or you can use the fo
 [options="nowrap" subs="+quotes"]
 ----
 # update-ca-trust enable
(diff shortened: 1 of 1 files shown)


melcorr:
@lzap From what I can tell, your link checker can't handle the example links we add to the docs?

lzap:
Yes, this will not work. I suggest to drop this practice completely, in tech we like to use fully qualified domains anyways and the best practice IMHO is to use "infoblox.example.com".

http://www.example.com : Example Domain

_This domain is established to be used for illustrative examples in documents. You may use this domain in examples without prior coordination or asking for permission._

However, this was not the reason why Travis failed. If you open it up, you will see this: Error: ReadTimeout: HTTPConnectionPool(host='ipxe.org', port=80): Read timed out. (read timeout=60) 

This means that the http://ipxe.org links could not be verified because the remote host did not respond. This will happen from time to time, just rebase and force push to trigger tests again to retest. I just tested this in a train from Wien to Munich and I saw zero errors :-)

melcorr:
Done!

lzap:
Nice it looks like it ignored the underscores. Great.

PR #72, opened 2019-11-27
Fix path for puppet.conf file

melcorr:
As part of

Bug 1763347 - Wrong puppet.conf path on Appendix C.4 of provisioning
guide

https://bugzilla.redhat.com/show_bug.cgi?id=1763347


melcorr:
Thanks

PR #76, opened 2020-01-10
A tool to help to navigate modularized docs

lzap:
I wrote a simple tool which creates symlinks in folder called `.alink`
which follows the order in which adoc includes appear in documents. This
makes navigation through modularized docs with a breeze.

Better to see it in a video:

https://www.youtube.com/watch?v=w03XYn-p1K0

Test it, share it, enjoy.

### DIFF
--- a/guides/.gitignore
+++ b/guides/.gitignore
@@ -1,3 +1,4 @@
 /build/
 /common/build.adoc
 /upstreamize_report.md
+.alink
--- a/guides/asciidoc-include-linker.rb
+++ b/guides/asciidoc-include-linker.rb
@@ -1,0 +1,33 @@
+#!/usr/bin/env ruby
+require "find"
+require "fileutils"
+
+DIR=ARGV[1] || '.'
+ALDIR="#{DIR}/.alink"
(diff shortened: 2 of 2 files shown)


melcorr:
Hi @lzap 
I watched the video and think this is great. 
When we modularize docs, the include files won't all be in the master.adoc file, but in separate files for the chapters, so for example "chap_Networking.adoc" would have include files for all the networking content for that chapter and the master.adoc will just have "include::chap_Networking.adoc".

Regardless, I think this will be easy for contributors who are unfamiliar with the file structures so your tool will be helpful for contributors to navigate the content.

We must type up and add the video to the readme me also. 

Thank you for this.

lzap:
> When we modularize docs, the include files won't all be in the master.adoc file, but in separate files for the chapters, so for example "chap_Networking.adoc" 

This should actually work too, the script searches all adoc files for includes not just master.adoc. For example in this repo it's this file:

topics/000-Bare_Metal.adoc-Common_Compute_Resource_Prereqs.adoc

PR #94, opened 2020-02-04
Bz#1797957

melcorr:
adding conditionals for operating systems
add snippet for firewall-cmd usage

### DIFF
--- a/guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
+++ b/guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
@@ -5,9 +5,9 @@
 To configure an external DHCP server to use with {ProductName}, on a Red{nbsp}Hat Enterprise Linux server, you must install the ISC DHCP Service and Berkeley Internet Name Domain (BIND) packages.
 You must also share the DHCP configuration and lease files with {ProductName}. The example in this procedure uses the distributed Network File System (NFS) protocol to share the DHCP configuration and lease files.
 
-include::snip_firewalld.adoc[]
 
 .Procedure
+include::snip_firewalld.adoc[]
 To configure an external DHCP server to use with {ProductName}, complete the following steps:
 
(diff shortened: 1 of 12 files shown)


guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
```diff
@@ -5,6 +5,8 @@
 To configure an external DHCP server to use with {ProductName}, on a Red{nbsp}Hat Enterprise Linux server, you must install the ISC DHCP Service and Berkeley Internet Name Domain (BIND) packages.
 You must also share the DHCP configuration and lease files with {ProductName}. The example in this procedure uses the distributed Network File System (NFS) protocol to share the DHCP configuration and lease files.
 
+include::snip_firewalld.adoc[]
```
tahliar: This include is outside the .Procedure markup, while the others are inside it. I don't mind which one you pick, but they should be consistent.

guides/doc-Provisioning_Guide/topics/Resources.adoc
```diff
@@ -483,5 +483,10 @@ endif::[]
 # firewall-cmd --add-port=5900-5930/tcp
 # firewall-cmd --add-port=5900-5930/tcp --permanent
 ----
+ifeval::["{build}" == "foreman"]
```
tahliar: Does this ifeval need to be here, since the file itself also contains an ifeval?

melcorr: I need this ifeval to nest it with the ++ markup. Elseses it is a sentence that breaks the procedure. It's a bit like what Leo DiCaprio dealt with in that film Inception, I know, but with asciidoc, you take what is given to you. 

TL;DR: The + signs are the main object I need in that ifeval to frame the include.

tahliar: Asciidoc strikes again!

PR #103, opened 2020-02-17
Remove all support info and just keep link

melcorr:
As part of

Bug 1802078 - [DDF] I guess intention of this statement was for intended
for PXE Less provisioning.

https://bugzilla.redhat.com/show_bug.cgi?id=1802078


melcorr:
@lzap go ahead!

PR #100, opened 2020-02-13
Debian version of the Installing Foreman guide

melcorr:
This is the early draft of the Foreman installation guide on Debian. 
If @lzap thinks that this could work, I will implement the changes for external Smart Proxy on Debian also. 
As part of 

Bug 1803033 - [upstream] add Installing Foreman on Debian system guide.
https://bugzilla.redhat.com/show_bug.cgi?id=1803033

PR #105, opened 2020-02-17
Overall review of Installing Satellite guide

melcorr:
As part of

Bug 1797957 - Write upstream/downstream conditions for Installing
Satellite Server (Connected) guide

https://bugzilla.redhat.com/show_bug.cgi?id=1797957


guides/common/modules/installing_sos.adoc
```diff
@@ -2,6 +2,10 @@
 
 = Installing the SOS Package on the Host Operating System
 
+ifeval::["{Build}" == "foreman"]
+For Red Hat systems only.
+endif::[]
```
lzap: Actually Debian contains sosreport as well. I'd probably remove this note but remove references to Red Hat Technical Support and replace this with our community site.

melcorr: I will deal with this when talking with Baboon. Ignore this for now.

melcorr: I will handle this in the debian PRs, if that is OK.

guides/common/modules/configuring_satellite_external_idm_dns.adoc
```diff
@@ -2,6 +2,10 @@
 
 = Configuring {Project} or {SmartProxy} with External IdM DNS
 
+ifeval::["{Build}" == "foreman"]
+This applies only to Red Hat-based systems.
+endif::[]
```
lzap: I have noticed we use `Build` and `build` in our guides, it's likely that this is not case sensitive. I haven't checked, but I actually like `build` more as it comply with other variables. Maybe we could unify this sometimes (not is this PR please).

melcorr: @lzap I am addressing this in the Debian PR #100 !! Stay tuned!

guides/common/modules/installing_sos.adoc
lzap: Sure, to give you some context: we ship tool called `foreman-debug`, a dirty shell script which I wrote. Last year Pavel Moravec rewrote it into an official sosreport plugin and it's now part of sos package, it will be available in Fedoras and RHEL and we would like to sunset our old tool sometime. However Debian is much more slower paced even then RHEL, so we should keep suggesting the old tool for some time specifically for debian users.

guides/common/modules/proc_configuring-repositories.adoc
```diff
@@ -40,9 +44,10 @@ ifeval::["{context}" == "capsule"]
 ----
 +
 endif::[]
+ifeval::["{build}" == "satellite"]
 NOTE: If you are installing {ProductName} as a virtual machine hosted on Red{nbsp}Hat Virtualization (RHV), you must also enable the *Red{nbsp}Hat Common* repository, and install RHV guest agents and drivers. For more information, see https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.3/html/virtual_machine_management_guide/installing_guest_agents_and_drivers_linux#Installing_the_Guest_Agents_and_Drivers_on_Red_Hat_Enterprise_Linux[Installing the Guest Agents and Drivers on Red Hat Enterprise Linux] in the _Virtual Machine Management Guide_ for more information.
-
 +
+endif::[]
```
lzap: This is a valid paragraph, just replace RH Virtualization with our variable which renders as oVirt upstream.

PR #106, opened 2020-02-18
Final changes for Capsule upstreaming

melcorr:
As part of

Bug 1797970 - Write upstream/downstream conditions for Installing
Capsule guide

https://bugzilla.redhat.com/show_bug.cgi?id=1797970

### DIFF
--- a/guides/common/assembly_configuring-capsule-custom-server-certificate.adoc
+++ b/guides/common/assembly_configuring-capsule-custom-server-certificate.adoc
@@ -9,9 +9,9 @@ If you configure {ProjectServer} to use a custom SSL certificate, you must also 
 
 To configure your {SmartProxyServer} with a custom certificate, complete the following procedures on each {SmartProxyServer}:
 
-. xref:creating-a-custom-certificate_capsule[]
-. xref:deploying-a-custom-ssl-certificate-to-capsule-server_capsule[]
-. xref:deploying-a-custom-ssl-certificate-to-hosts_capsule[]
+. xref:creating-a-custom-certificate_{smart-proxy-context}[]
+. xref:deploying-a-custom-ssl-certificate-to-capsule-server_{smart-proxy-context}[]
+. xref:deploying-a-custom-ssl-certificate-to-hosts_{smart-proxy-context}[]
 
 //Creating a Custom SSL Certificate for {SmartProxyServer}
 include::modules/proc_creating-a-custom-certificate.adoc[leveloffset=+1]
@@ -24,4 +24,3 @@ include::modules/proc_deploying-a-custom-ssl-certificate-to-hosts.adoc[leveloffs
 
 ifdef::parent-context[:context: {parent-context}]
 ifndef::parent-context[:!context:]
-
--- a/guides/doc-Installing_Capsule_Server/master.adoc
+++ b/guides/doc-Installing_Capsule_Server/master.adoc
@@ -8,7 +8,7 @@ endif::[]
 
 include::common/attributes.adoc[]
 
-:context: {SmartProxy}
+:context: {smart-proxy-context}
 :ProductName: {SmartProxyServer}
 
 = Installing {SmartProxyServer}
--- a/guides/doc-Installing_Capsule_Server/master.adoc
+++ b/guides/doc-Installing_Capsule_Server/master.adoc
@@ -8,7 +8,7 @@ endif::[]
 
 include::common/attributes.adoc[]
 
-:context: capsule
+:context: {SmartProxy}
 :ProductName: {SmartProxyServer}
 
 = Installing {SmartProxyServer}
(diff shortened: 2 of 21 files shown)


melcorr:
Hi @tahliar 
I don't know how to handle this `context` attribute related problem: 

I changed the master.adoc context to `{SmartProxy}`. I have done this for the Satellite guide also, changing context to {Project} and this is handled very nicely upstream and downstream using the attributes set in the ifevals of the attributes.adoc file. 

However I am caught in `proc_creating-a-custom-certificate.adoc` file

The mod docs effort used this: 

DNS.1 = _{context}.example.com_ <3>

# openssl genrsa -out `/root/{context}_cert/{context}_cert_key.pem` 4096


. Create the `/root/{context}_cert/openssl.cnf`

Would it be possible for me to add ifevals with context = satellite context = capsule and defining the actual attribute to use from the build perspective in there? 

Would that mess anything up?

tahliar:
> Hi @tahliar
> I don't know how to handle this `context` attribute related problem:
> 
> I changed the master.adoc context to `{SmartProxy}`. I have done this for the Satellite guide also, changing context to {Project} and this is handled very nicely upstream and downstream using the attributes set in the ifevals of the attributes.adoc file.
> 
> However I am caught in `proc_creating-a-custom-certificate.adoc` file
> 
> The mod docs effort used this:
> 
> DNS.1 = _{context}.example.com_ <3>
> # openssl genrsa -out `/root/{context}_cert/{context}_cert_key.pem` 4096
> 
> . Create the `/root/{context}_cert/openssl.cnf`
> 
> Would it be possible for me to add ifevals with context = satellite context = capsule and defining the actual attribute to use from the build perspective in there?
> 
> Would that mess anything up?

This is a good find, and good to bring up while we're still early in the process. 

Making the context {SmartProxy} is a great idea in theory, and works well in some cases, but there's this issue you mention here, and also, since upstream Smart Proxy is two words, it will probably cause problems with links, because we use {context} in section IDs (so you might have a link ending in "section-id_Smart Proxy" or some jumble of characters in place of the white space). 

Therefore, it might be a good idea to have a new attribute, something like {smart-proxy-id}, which evaluates to smart-proxy or capsule in lowercase. That would then work in IDs and in examples like this one. 

What do you think?

lzap:
Unsure about the context thing, I was only looking into technical aspects of the patch.

melcorr:
> Unsure about the context thing, I was only looking into technical aspects of the patch.

No problem, Lzap, @tahliar 's proposal solves everything and I will implement after this call.

guides/doc-Installing_Capsule_Server/master.adoc
```diff
@@ -8,7 +8,7 @@ endif::[]
 
 include::common/attributes.adoc[]
 
-:context: capsule
+:context: {SmartProxy}
```
lzap: From our meeting: My concern is that if we going to use any other text than "capsule" we also need to change all the "ifeval" statement as they compare only against this word currently. The same for "satellite".

melcorr: Will start testing.

melcorr: Hi @lzap I looked into this and (with @tahliar 's help ) found that we can create an attribute to replace satellite and capsule. I tested and found it to work everywhere:
`ifeval::["{context}" == "{project-context}"]`
`ifeval::["{context}" == "{smart-proxy-context}"]`

`context` is defined as `{project-context}` or `{smart-proxy-context}` in the master.adoc files. 

When we build, the context is set through the attribute. 

I went through both Foreman/Satellite and Capsule/Smart Proxy scenarios and think that this works. 

If you have any thoughts, let me know.

guides/common/assembly_configuring-capsule-custom-server-certificate.adoc
```diff
@@ -9,9 +9,9 @@ If you configure {ProjectServer} to use a custom SSL certificate, you must also
 
 To configure your {SmartProxyServer} with a custom certificate, complete the following procedures on each {SmartProxyServer}:
 
-. xref:creating-a-custom-certificate_capsule[]
-. xref:deploying-a-custom-ssl-certificate-to-capsule-server_capsule[]
-. xref:deploying-a-custom-ssl-certificate-to-hosts_capsule[]
+. xref:creating-a-custom-certificate_{smart-proxy-context}[]
+. xref:deploying-a-custom-ssl-certificate-to-capsule-server_{smart-proxy-context}[]
+. xref:deploying-a-custom-ssl-certificate-to-hosts_{smart-proxy-context}[]
```
lzap: Just a heads up as I haven't tried to build this - I saw that variables cannot be used in some instances like in anchors. Doublecheck this renders correctly.

melcorr: Before I committed these changes I tested this throughout all versions of the guide and this works! Thanks.

melcorr:
> There is still one left behind, is this intentional?
> 
> ```
> [lzap@box guides]$ ag context | grep ifeval
> common/modules/con_ports-and-firewalls-requirements.adoc:7:ifeval::["{context}" == "capsule"]
> ```

Not intentional. I missed this. Fixed.  I want this ag tool. I just use grep.

lzap:
Heh "ag" is just a nicer grep that respects git ignores etc:

```
$ dnf whatprovides /usr/bin/ag
the_silver_searcher-2.1.0-5.fc31.x86_64 : Super-fast text searching tool (ag)
```

PR #110, opened 2020-02-21
Add CLI commands to Provisioning Cloud Instances in GCE

spetrosi:
Bug 1777312 - [RFE] Add Hammer commands for provisioning with GCE

https://bugzilla.redhat.com/show_bug.cgi?id=1777312

### DIFF
--- a/guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
+++ b/guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
@@ -21,12 +21,22 @@ Use this procedure to add a GCE connection to {Project} to be able to add images
 .Procedure
 
 . In GCE, generate a service account key in JSON format and upload this file to the `/usr/share/foreman/` directory on {ProjectServer}.
-. On {ProjectServer}, configure permissions for the service account key to ensure that the file is readable by the `foreman` user:
+. On {ProjectServer}, change owner for the service account key to the `foreman` user:
 +
 [options="nowrap" subs="+quotes"]
 ----
 # chown foreman /usr/share/foreman/_gce_key_.json
+----
+. Configure permissions for the service account key to ensure that the file is readable:
++
+[options="nowrap" subs="+quotes"]
+----
 # chmod 0600 /usr/share/foreman/_gce_key_.json
+----
+. Restore SELinux context for the service account key:
++
+[options="nowrap" subs="+quotes"]
+----
 # restorecon -vv /usr/share/foreman/_gce_key_.json
 ----
 . In the {Project} web UI, navigate to *Infrastructure* > *Compute Resources* and click *Create Compute Resource*.
@@ -44,12 +54,24 @@ Use this procedure to add a GCE connection to {Project} to be able to add images
 
 . In GCE, generate a service account key in JSON format and upload this file to the `/usr/share/foreman/` directory on {ProjectServer}.
 
-. On {ProjectServer}, configure permissions for the service account key to ensure that the file is readable by the `foreman` user:
+. On {ProjectServer}, change owner for the service account key to the `foreman` user:
 +
 [options="nowrap" subs="+quotes"]
 ----
 # chown foreman /usr/share/foreman/_gce_key_.json
+----
+
+. Configure permissions for the service account key to ensure that the file is readable:
++
+[options="nowrap" subs="+quotes"]
+----
 # chmod 0600 /usr/share/foreman/_gce_key_.json
+----
+
+. Restore SELinux context for the service account key:
++
+[options="nowrap" subs="+quotes"]
+----
 # restorecon -vv /usr/share/foreman/_gce_key_.json
 ----
 
@@ -85,7 +107,7 @@ GCE uses image-based provisioning to create hosts. You must add image details to
 
 .For CLI Users
 
-To add an image, enter the following command. With the `--username` option, specify a user other than `root`, because the `root` user cannot connect to a GCE instance using SSH keys. The username must begin with a letter and consist of lowercase letters and numbers.
+Use the `hammer compute-resource image create` command to add an image to the GCE compute resource. With the `--username` option, specify a user other than `root`, because the `root` user cannot connect to a GCE instance using SSH keys. The username must begin with a letter and consist of lowercase letters and numbers.
 
 [options="nowrap" subs="+quotes"]
 ----
@@ -180,7 +202,7 @@ To create a host, enter the following command:
 --compute-profile "_gce_profile_name_" \
 --image _gce_image_name_
 --provision-method 'image' \
---root-password _changeme_ \
+--root-password "_your_root_password_" \
 --interface "type=interface,domain_id=1,managed=true,primary=true,provision=true" \
 --puppet-environment-id _1_ \
 --puppet-ca-proxy-id _1_ \
--- a/guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
+++ b/guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
@@ -20,8 +20,6 @@ Use this procedure to add a GCE connection to {Project} to be able to add images
 
 .Procedure
 
-To add a connection, complete the following steps:
-
 . In GCE, generate a service account key in JSON format and upload this file to the `/usr/share/foreman/` directory on {ProjectServer}.
 . On {ProjectServer}, configure permissions for the service account key to ensure that the file is readable by the `foreman` user:
 +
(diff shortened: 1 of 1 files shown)


kgaikwad:
Looks good to me :+1: 
@jyejare, could you please go through once?

guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
```diff
@@ -42,15 +40,38 @@ To add a connection, complete the following steps:
 . From the *Zone* list, select the GCE zone to use.
 . Click *Submit*.
 
+.For CLI Users
+
+. In GCE, generate a service account key in JSON format and upload this file to the `/usr/share/foreman/` directory on {ProjectServer}.
```
jyejare: Not required to repeat as this is the same as UI.

What do you say ?

guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
```diff
@@ -42,15 +40,38 @@ To add a connection, complete the following steps:
...
+
+. On {ProjectServer}, configure permissions for the service account key to ensure that the file is readable by the `foreman` user:
```
jyejare: same here.

guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
```diff
@@ -80,6 +115,25 @@ To add details to a compute profile, complete the following steps:
...
+# hammer compute-profile create --name _gce_profile_
+----
+
+. Add GCE details to the compute profile. 
++
+[options="nowrap" subs="+quotes"]
+----
+# hammer compute-profile values create --compute-profile _gce_profile_ \
+--compute-resource '_gce_cr_' \
+--volume "_size_gb=20_" \
```
jyejare: Does _size_gb with prior underscore actually appear in Doc preview. If yes, That should be fixed as size_gb. Its a size parameter and is a constant/fixed name.

guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
```diff
@@ -80,6 +115,25 @@ To add details to a compute profile, complete the following steps:
...
+[options="nowrap" subs="+quotes"]
+----
+# hammer compute-profile create --name _gce_profile_
+----
+
+. Add GCE details to the compute profile. 
++
+[options="nowrap" subs="+quotes"]
+----
+# hammer compute-profile values create --compute-profile _gce_profile_ \
```
jyejare: Looks like the '--image_id' option is missing ? @kavita, I see that is also missing from hammer values compute-profile create --help.

guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
spetrosi: @jyejare, the underscores are used to highlight text as italics. In the built doc this line will appear in the following format:
 --volume "_size_gb=20_" \ 
^Does that look correct?

The same works for all other lines - surrounded by underscores=renders as italics.

guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
kgaikwad: To make consistent with other compute resources, user won't be able to configure "image" value at compute profile level. I have created ticket to remove this field from compute-resource form  -  https://projects.theforeman.org/issues/29183

guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
spetrosi: I perceive those two procedures - the UI and CLI - as two different procedures. So I expect users who only use CLI to jump right to the CLI procedure and not look at the UI steps, that is why I repeat the same required steps in both procedures.

jyejare: Make sense ! ACK!

The change i m looking for might needs to be done globaly.

guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
jyejare: ACK

guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
jyejare: Make sense! Thanks for the clarification. Got to know something else.

guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
jyejare: @kgaikwad, Is that an UI change ? And when are we planning to release that ?

kgaikwad: @jyejare, 
Yes, it is UI change/fix and it is merged now in upstream which will remove image field from compute-profile page so no need to pass --image_id while creating compute-profile.

spetrosi: The `--image` option is not available with the `hammer compute-profile values create` command, but it is there in the `hammer host create` command. I am marking this conversation as resolved.

guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
```diff
@@ -42,15 +40,38 @@ To add a connection, complete the following steps:
...
+
+. In GCE, generate a service account key in JSON format and upload this file to the `/usr/share/foreman/` directory on {ProjectServer}.
+
+. On {ProjectServer}, configure permissions for the service account key to ensure that the file is readable by the `foreman` user:
++
+[options="nowrap" subs="+quotes"]
+----
+# chown foreman /usr/share/foreman/_gce_key_.json
+# chmod 0600 /usr/share/foreman/_gce_key_.json
+# restorecon -vv /usr/share/foreman/_gce_key_.json
```
melcorr: Sergei, several times in different reviews, you have made me separate several commands into individual steps. Can you do the same here?

guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
```diff
@@ -62,6 +83,20 @@ To add an image, complete the following steps:
 . Optional: If the selected image supports `cloud-init`, select the *User data* check box to enable user data input.
 . Click *Submit* to save the image details.
 
+.For CLI Users
+
+To add an image, enter the following command. With the `--username` option, specify a user other than `root`, because the `root` user cannot connect to a GCE instance using SSH keys. The username must begin with a letter and consist of lowercase letters and numbers.
```
melcorr: Please be consistent with the way you refer to the command. In line 56, you use one sentence structure, and use another now. It is better if we are consistent.

guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
```diff
@@ -62,6 +83,20 @@ To add an image, complete the following steps:
...
+
+[options="nowrap" subs="+quotes"]
+----
+# hammer compute-resource image create --name '_gce_image_name_' \
+--compute-resource '_gce_cr_' \
+--operatingsystem-id 1 \
+--architecture-id 1 \
+--uuid '_378010XXXXXXXXX_' \
```
melcorr: Is this example, with all these XXXs really necessary? Is the UUID in this format? It is usually with dashes. If this value has no relation to an actual UUID, perhaps just replace with _UUID_

guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
```diff
@@ -112,3 +166,25 @@ ifeval::["{build}" == "foreman"]
...
+[options="nowrap" subs="+quotes"]
+----
+# hammer host create --name "_GCE_VM_" \
+--organization "_Your_Organization_" \
+--location "_Your_Location_" \
+--compute-resource _gce_cr_name_
+--compute-profile "_gce_profile_name_" \
+--image _gce_image_name_
+--provision-method 'image' \
+--root-password _changeme_ \
```
melcorr: lol, let's change this to something more generic and in line with other examples of creating a host.

guides/doc-Provisioning_Guide/topics/Cloud-GCE.adoc
spetrosi: @ kgaikwad, can you clarify if the uuid in this example is a real value or a dummy value?

kgaikwad: I have updated the real UUID with XXXs. You can replace XXXs with some numbers.

spetrosi: @kgaikwad, are you able to provide a real UUID from GCE? Using a real value is always better in docs than a dummy value of XXXs. Thank you.

kgaikwad: @spetrosi, real UUID from GCE  - `3780108136525169178`

melcorr:
Great work and thanks everyone!

Issue #111, opened 2020-02-21
Better book titles upstream


lzap:
On our front page we currently have:

* Installing Foreman on RHEL/CentOS
* Installing Smart Proxy on RHEL/CentOS
* Installing Foreman on Debian/Ubuntu
* Installing Smart Proxy on Debian/Ubuntu 

https://docs.theforeman.org/web/

But when you open these it reads e.g. "Installing Foreman server from a Connected Network". We should probably change these titles to match the front page or come up with something that works fine.

melcorr:
Directory names have been updated as part of #114

PR #114, opened 2020-02-24
Rename directories and conditionalize titles for Installation guides

melcorr:
As part of

 #111
Bug 1806543 - [upstream] improve directory names for Installing guides
upstream

https://bugzilla.redhat.com/show_bug.cgi?id=1806543


lzap:
I'd probably go just "doc-Installing_on_Red_Hat" and "doc-Installing_on_Debian":

1) I'd avoid project/product name from directory names, paths and URLs just in case downstream would adopt this. Things can change.

2) Red Hat OS family includes CentOS, Oracle Enterprise Linux, Scientific Linux and other derivates too. The same for Debian (includes Ubuntu, KUbuntu etc).

3) Shorter directory names are better in general.

Alternatively, "Installation_Guide_for_Red_Hat" as we use "Guide" for most (all?) other books. I am not sure if you are leaving this naming convention or actually implementing it downstream.

PR #118, opened 2020-02-26
Change unclear $(hostname) to the example URL of smart proxy

spetrosi:
Bug 1792263 - [DDF] In example you add $(hostname) which represents FQDN of capsule server, but in Satellite provisioning guide in

https://bugzilla.redhat.com/show_bug.cgi?id=1792263

### DIFF
--- a/guides/common/modules/proc_configuring-dns-dhcp-and-tftp.adoc
+++ b/guides/common/modules/proc_configuring-dns-dhcp-and-tftp.adoc
@@ -53,7 +53,7 @@ ifeval::["{context}" == "{project-context}"]
 --foreman-proxy-dhcp-nameservers __172.17.13.2__ \
 --foreman-proxy-tftp true \
 --foreman-proxy-tftp-managed true \
---foreman-proxy-tftp-servername $(hostname)
+--foreman-proxy-tftp-servername _{smartproxy-example-com}_
 ----
 
 You can monitor the progress of the `{foreman-installer}` command displayed in your prompt. You can view the logs in `/var/log/foreman-installer/satellite.log`. You can view the settings used, including the `admin_password` parameter, in the `/etc/foreman-installer/scenarios.d/satellite-answers.yaml` file.
@@ -78,7 +78,7 @@ ifeval::["{context}" == "{smart-proxy-context}"]
 --foreman-proxy-dhcp-nameservers _172.17.13.2_ \
 --foreman-proxy-tftp true \
 --foreman-proxy-tftp-managed true \
---foreman-proxy-tftp-servername $(hostname)
+--foreman-proxy-tftp-servername _{smartproxy-example-com}_
 ----
 endif::[]
 
--- a/guides/common/modules/proc_configuring-dns-dhcp-and-tftp.adoc
+++ b/guides/common/modules/proc_configuring-dns-dhcp-and-tftp.adoc
@@ -43,7 +43,6 @@ ifeval::["{context}" == "{project-context}"]
 --foreman-proxy-dns-managed true \
 --foreman-proxy-dns-interface __eth0__ \
 --foreman-proxy-dns-zone __example.com__ \
---foreman-proxy-dns-forwarders __192.0.2.1__ \
 --foreman-proxy-dns-reverse __2.0.192.in-addr.arpa__ \
 --foreman-proxy-dhcp true \
 --foreman-proxy-dhcp-managed true \
@@ -53,10 +52,10 @@ ifeval::["{context}" == "{project-context}"]
 --foreman-proxy-dhcp-nameservers __192.0.2.2__ \
 --foreman-proxy-tftp true \
 --foreman-proxy-tftp-managed true \
---foreman-proxy-tftp-servername _192.0.2.0_
+--foreman-proxy-tftp-servername _192.0.2.3_
 ----
 
-You can monitor the progress of the `{foreman-installer}` command displayed in your prompt. You can view the logs in `/var/log/foreman-installer/satellite.log`. You can view the settings used, including the `admin_password` parameter, in the `/etc/foreman-installer/scenarios.d/satellite-answers.yaml` file.
+You can monitor the progress of the `{foreman-installer}` command displayed in your prompt. You can view the logs in `/var/log/foreman-installer/{project-context}.log`. You can view the settings used, including the `initial_admin_password` parameter, in the `/etc/foreman-installer/scenarios.d/{project-context}-answers.yaml` file.
 endif::[]
 
 ifeval::["{context}" == "{smart-proxy-context}"]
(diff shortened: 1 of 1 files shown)


guides/common/modules/proc_configuring-dns-dhcp-and-tftp.adoc
```diff
@@ -53,7 +53,7 @@ ifeval::["{context}" == "{project-context}"]
 --foreman-proxy-dhcp-nameservers __172.17.13.2__ \
 --foreman-proxy-tftp true \
 --foreman-proxy-tftp-managed true \
---foreman-proxy-tftp-servername $(hostname)
+--foreman-proxy-tftp-servername _{smartproxy-example-com}_
```
ekohl: I'd suggest to use IP addresses. The code to do the lookups is complex and can lead to failures. The satellite/capsule servers should have static IPs and is thus more reliable without additional maintenance overhead.

guides/common/modules/proc_configuring-dns-dhcp-and-tftp.adoc
```diff
@@ -53,7 +53,7 @@ ifeval::["{context}" == "{project-context}"]
 --foreman-proxy-dhcp-nameservers __172.17.13.2__ \
```
ekohl: Unrelated note, but https://community.theforeman.org/t/using-ips-and-domains-in-tests-documentation/16600 documents which IP ranges should be used for documentation.

guides/common/modules/proc_configuring-dns-dhcp-and-tftp.adoc
spetrosi: Thank you @ekohl, I will change the option to use an IP address.

guides/common/modules/proc_configuring-dns-dhcp-and-tftp.adoc
spetrosi: Interesting, I will change IP addresses in the guides to correspond with the rule.

melcorr:
Hi @spetrosi  can you please rebase this as the latest changes from #114 fix the broken links.

guides/common/modules/proc_configuring-dns-dhcp-and-tftp.adoc
```diff
@@ -43,17 +43,17 @@ ifeval::["{context}" == "{project-context}"]
...
---foreman-proxy-dhcp-range "__172.17.13.100__ __172.17.13.150__" \
---foreman-proxy-dhcp-gateway __172.17.13.1__ \
---foreman-proxy-dhcp-nameservers __172.17.13.2__ \
+--foreman-proxy-dhcp-range "__192.0.2.100__ __192.0.2.150__" \
+--foreman-proxy-dhcp-gateway __192.0.2.1__ \
+--foreman-proxy-dhcp-nameservers __192.0.2.2__ \
 --foreman-proxy-tftp true \
 --foreman-proxy-tftp-managed true \
---foreman-proxy-tftp-servername $(hostname)
+--foreman-proxy-tftp-servername _192.0.2.0_
```
ekohl: `.0` IPs are usually not valid IPs

guides/common/modules/proc_configuring-dns-dhcp-and-tftp.adoc
```diff
@@ -68,17 +68,17 @@ ifeval::["{context}" == "{smart-proxy-context}"]
...
---foreman-proxy-dhcp-range "_172.17.13.100_ _172.17.13.150_" \
---foreman-proxy-dhcp-gateway _172.17.13.1_ \
---foreman-proxy-dhcp-nameservers _172.17.13.2_ \
+--foreman-proxy-dhcp-range "_192.0.2.100_ _192.0.2.150_" \
+--foreman-proxy-dhcp-gateway _192.0.2.1_ \
+--foreman-proxy-dhcp-nameservers _192.0.2.2_ \
 --foreman-proxy-tftp true \
 --foreman-proxy-tftp-managed true \
---foreman-proxy-tftp-servername $(hostname)
+--foreman-proxy-tftp-servername _192.0.2.0_
```
ekohl: Same

guides/common/modules/proc_configuring-dns-dhcp-and-tftp.adoc
```diff
@@ -43,17 +43,17 @@ ifeval::["{context}" == "{project-context}"]
...
+--foreman-proxy-dhcp-range "__192.0.2.100__ __192.0.2.150__" \
+--foreman-proxy-dhcp-gateway __192.0.2.1__ \
+--foreman-proxy-dhcp-nameservers __192.0.2.2__ \
 --foreman-proxy-tftp true \
 --foreman-proxy-tftp-managed true \
---foreman-proxy-tftp-servername $(hostname)
+--foreman-proxy-tftp-servername _192.0.2.0_
 ----
 
 You can monitor the progress of the `{foreman-installer}` command displayed in your prompt. You can view the logs in `/var/log/foreman-installer/satellite.log`. You can view the settings used, including the `admin_password` parameter, in the `/etc/foreman-installer/scenarios.d/satellite-answers.yaml` file.
```
ekohl: These paths should use the scenario name. The `admin_password` parameter is called `initial_admin_password` for a few (upstream) releases but has also made its way into downstream.

guides/common/modules/proc_configuring-dns-dhcp-and-tftp.adoc
```diff
@@ -43,17 +43,17 @@ ifeval::["{context}" == "{project-context}"]
 --foreman-proxy-dns-managed true \
 --foreman-proxy-dns-interface __eth0__ \
 --foreman-proxy-dns-zone __example.com__ \
---foreman-proxy-dns-forwarders __172.17.13.1__ \
---foreman-proxy-dns-reverse __13.17.172.in-addr.arpa__ \
+--foreman-proxy-dns-forwarders __192.0.2.1__ \
```
ekohl: Forwarders are optional and not strictly needed. You are pointing here to the gateway and that may be confusing. We're getting very much off topic here which is why I usually separate changes like this into their own PRs.

IMHO we should not recommend forwarders by default. Then explain why they're needed. *if* you run a managed nameserver (bind), then it can be used to point to an upstream recursor which does all the heavy lifting. If you don't specify it, bind will reach out to the root servers and rely on its own cache. If you don't run a managed nameserver (for example because you use Infoblox or point to an existing external nameserver), it's ignored.

spetrosi: This command is an example of a basic configuration that is shown in the Installing guide, which means that this can be the command that is run for the first time, so I will keep the `--scenario` option.
As for forwarders, as far as I understand, they are used in a configuration that uses an external server as a recursor. This sounds like something further than basic, so I will remove this option from the command.

spetrosi:
Hi @ekohl, is that fine that I removed the `...-forwarders` options as per my comment [1]?
[1] https://github.com/theforeman/foreman-documentation/pull/118#discussion_r386442482

spetrosi:
Thank you for pointing this out @ekohl, I will remember this the next time :)

PR #123, opened 2020-03-04
BZ#1698333 - FDI instructions updated

lzap:
Couple of changes, the main change is that it's not possible to Publish via HTTP anymore in Katello. Therefore local repository URLs must be used. This means the procedure MUST be either done on Satellite (which we actually disencourage to do) or files should be copied over to a different server. Maybe we should change our mind and simply tell the user to do this on the Satellite server. That would do it, it's harmless.


melcorr:
Hi @spetrosi 

The work in this PR is related to the following Bugzilla ticket: 

Bug 1698333 - Building a Satellite Discovery Image inconsistencies

https://bugzilla.redhat.com/show_bug.cgi?id=1698333

spetrosi:
Looks perfect, thank you.

PR #125, opened 2020-03-05
Add example outputs of katello-certs-check and capsule-certs-generate

spetrosi:
Bug 1810218 - Command for installing customer certs to satellite missing

https://bugzilla.redhat.com/show_bug.cgi?id=1810218

### DIFF
--- a/guides/common/modules/proc_configuring-capsule-default-certificate.adoc
+++ b/guides/common/modules/proc_configuring-capsule-default-certificate.adoc
@@ -32,6 +32,80 @@ To configure {SmartProxyServer} with a default server certificate, complete the 
 ----
 +
 Retain a copy of the `{foreman-installer}` command that the `{certs-generate}` command returns for deploying the certificate to your {SmartProxyServer}.
++
+ifeval::["{build}" == "satellite"]
+.Example output of capsule-certs-generate
+[options="nowrap", subs="+quotes"]
+----
+Installing             Done                                               [100%]
+  Success!
+
+  To finish the installation, follow these steps:
+
+  If you do not have the Capsule registered to the Satellite instance, then please do the following:
+
+  1. yum -y localinstall http://_satellite.example.com_.com/pub/katello-ca-consumer-latest.noarch.rpm
+  2. subscription-manager register --org "_Default_Organization_"
+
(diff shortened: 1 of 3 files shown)


melcorr:
I am really against this.

lzap:
> I am really against this.

Can you elaborate? Is the reason having hard time maintaining those in the future? That one, I can comment :-)

What if we come up with an automated way of creating output of these commands as .txt files. On every release, I can run such script on freshly installed machine and get us output which we can include into our documentation. The only problem is that we usually want to replace some texts e.g. something.example.com. We could easily search and replace these.

Honestly, these outputs are sometimes better than three paragraphs of texts. People rush reading docs, they quickly scan for commands and output to find what they want. Keeping these up to date would be great thing to have.

Having said that, we all have plenty tasks to do, we could only put this to our TODO backlog. Also, this approach will defeinitely not work for all commands.

melcorr:
Hi @lzap 

Myself and @spetrosi have been discussing this. In general, it is wasteful to reproduce in the docs the information that is presented to the user, especially if it is clear. I think we should only put output of commands into the docs when it is complex and needs explanation.

However, we have had requests from customers to add this back to the documentation, and we have to listen to that. 

In talking to @spetrosi , we have agreed to compromise and to keep the commands but to remove all of the text: 

----
 Success!
  To finish the installation, follow these steps:
  If you do not have the Capsule registered to the Satellite instance, then please do the following:
  1. yum -y localinstall http://_satellite.example.com_.com/pub/katello-ca-consumer-latest.noarch.rpm
  2. subscription-manager register --org "_Default_Organization_"
  Once this is completed run the steps below to start the Capsule installation:
  1. Ensure that the satellite-capsule package is installed on the system.
  2. Copy the following file _/root/capsule_cert/capsule_certs.tar_ to the system _capsule.example.com_ at the following location _/root/capsule_certs.tar_
  scp _/root/capsule_cert/capsule_certs.tar_ _root@capsule.example.com:/root/capsule_certs.tar_
  3. Run the following commands on the Capsule (possibly with the customized
     parameters, see satellite-installer --scenario capsule --help and
     documentation for more info on setting up additional 
----

I think it is too much and as you said, can cause the docs to age with the slightest change. 
I like your idea but don't think it is urgent.

PR #127, opened 2020-03-09
Bz#1811623

melcorr:
While experimenting with syncing the Installation guides upstream to downstream, the build breaks downstream because of the `ifeval` statements on the doc titles. This relates to added complication of Docbook xml that we don't have upstream. 

To work around this @tahliar suggested that we use an attribute for the doc titles, and on experimenting, this worked well. 

This PR adds attributes for the different titles that we want to use upstream and downstream for the Installation guides. The distinction between connected and disconnected relates to scenarios that don't exist upstream, and also upstream, the installation of an external Smart Proxy needs more clarification than it does downstream.


melcorr:
As part of this PR, I also updated the master.adoc reference to an assembly that was updated as part of the modularization effort for the downstream version of the Smart Proxy guide.

PR #128, opened 2020-03-12
Disable HTML link validator server cert check

lzap:
There is still an error ATM, it's because the target does not work.


melcorr:
Thank you !

PR #129, opened 2020-03-16
Add the `foreman-proxy-plugin-dns-infoblox-dns-view` option with desc…

spetrosi:
…ription

Bug 1812900 - [DDF] Please add the option "--foreman-proxy-plugin-dns-infoblox-dns-view" to set the DNS view on the configuration file

https://bugzilla.redhat.com/show_bug.cgi?id=1812900


melcorr:
We can ignore the error. It's a known issue.

PR #130, opened 2020-03-17
Ignore cdn.redhat.com during link checking

lzap:
It needs X509 client cert which we don't have and don't want to use actually.


melcorr:
Thank you!

PR #134, opened 2020-03-20
[WIP] Bz#1796973 rhev cloud-init

melcorr:
Hi @shiramax

Could you take a look and perhaps help me figure out the prerequisites. 

I copied over as much of the KBase as made sense to me. I am a bit lost with this part: 

> We install cloud-init from the rhel-7-server-rpms repository:
> yum install -y cloud-init

Where does this happen? On Satellite or on the VM that we then use to make an image? 

Can you take a look at this PR so far and let me know where it needs to be improved? 

Thank you! 

Melanie


shiramax:
@melcorr 
installing cloud-init should be on a VM inside ovirt, and then you should create a template out of this VM. 

later in foreman, you should create an image from this template. 

I hope it's clearer now if you have any more questions please let me know.

melcorr:
Changed to @spetrosi for review.

PR #141, opened 2020-04-03
Update documentation on configuring project with HTTP proxy

spetrosi:
* Replace the procedure to create a proxy entry in Satellite with a new procedure that uses the new method
* Create a new title Verifying the HTTP Proxy connection to a
* Add CLI commands to set hammer settings

Bug 1806587 - Update documentation to not refer to --katello-proxy-* options

https://bugzilla.redhat.com/show_bug.cgi?id=1806587


spetrosi:
> I am good. It is worth mentioning that this is poorly implemented - HTTP Proxies are not used by all communication. There are plenty of places where Foreman would ignore this setting and fallback to Administer - Settings - proxy. Like most Compute Resources do this.

Hi @lzap, thank you for checking on this. Sad to hear that this does not work consistently now. We definitely shall state this fact in the documentation. It would be fair to add more details about what are the drawbacks here so users know what to expect. Do you know exactly what

PR #149, opened 2020-04-27
Rename Bare metal hosts chapter

lzap:
chapter entitled "Provisioning Bare Metal Hosts" is just wrong for multiple reasons:

* It describes PXE workflow.
* PXE can be used on bare metal, virtualization and even on some clouds.
* It's confusing.
* We want to add new chapters on HTTP Boot provisioning as well.

Therefore I suggest to rename this.


melcorr:
Perhaps "Using PXE to Provision Hosts" would be better or "Provisioning Hosts Using PXE" since we don't need to call out bare metal or virtualized specifically.

lzap:
Okay, I am renaming also the anchor above if that's ok.

PR #161, opened 2020-05-12
Add DHCP override note - BZ#1777244

lzap:
A customer reported that --dhcp-no-override setting in dnsmasq breaks the Foreman TFTP directory setup. We need to drop a word about this.

### DIFF
--- a/guides/common/modules/proc_configuring-repositories.adoc
+++ b/guides/common/modules/proc_configuring-repositories.adoc
@@ -2,13 +2,12 @@
 
 = Configuring Repositories
 
-ifeval::["{build}" == "foreman"]
-This procedure is only for Katello plug-in and {RHEL}-based operating system users.
-endif::[]
-
 Use this procedure to enable the repositories that are required to install {ProductName}.
 
 .Procedure
+ifeval::["{build}" == "foreman"]
+This procedure is only for Katello plug-in users. Skip steps involving `subscription-manager` when not installing on the {RHEL} operating system.
+endif::[]
 To configure the required repositories, complete the following steps:
 
 . Disable all repositories:


spetrosi:
Hi @lzap, thank you for submitting this PR. The Install guide was modularized, so I will need to add this note to a new file by rebasing this PR.
For now, I have rewritten your writing to improve the style and add structure. How does this look?
~~~
Disable the override setting, for example, the `no-override` setting on dnsmasq, on the external DHCP server. The override setting disables re-use of the DHCP servername and filename fields as extra option space. This is required because {Project} creates configuration files on the TFTP server under the `grub2/` subdirectory. While with the override setting enabled,  clients fetch the bootloader and its configuration from the `root` directory, which might cause an error.
~~~

lzap:
I tried to rebase but this version looks completely different and I don't know more details about how to modularize this, probably this needs new file or something with notes for various DHCP servers. This note is only for dnsmasq server which we distribute in RHEL.

> Disable the override setting, for example, the `no-override` setting on dnsmasq, on the external DHCP server. The override setting disables re-use of the DHCP servername and filename fields as extra option space. This is required because {Project} creates configuration files on the TFTP server under the `grub2/` subdirectory. While with the override setting enabled,  clients fetch the bootloader and its configuration from the `root` directory, which might cause an error.

Generally I am okay but this configuration value is actually only available in dnsmasq server: http://www.thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html (`--dhcp-no-override`). Just reword it in a way that his is not a generic advice but dnsmasq only advice. E.g. "When using dnsmasq, make sure to...".

spetrosi:
I understand now. I will open a new PR to add this to the guide.
Rebase did not help because @tahliar modularized this, changed files contents, files names...

spetrosi:
I cloned the same changes in #180

PR #166, opened 2020-05-15
Bz#1825227

tahliar:
Bug 1825227 - Modularise "Configuring External Services"

https://bugzilla.redhat.com/show_bug.cgi?id=1825227


lzap:
A friendly advice: can you guys make sure the title is utilized on github? When I do my inbox, it really helps me if I need to step in with my technical notes or not, that was this case. Thanks!

PR #171, opened 2020-05-19
xref that evaded me because it was older than the old one

tahliar:


tahliar:
> The change looks good. Why the Redmine check fails though?

I'm not sure. Looks like it's been going for 18 hours though, which doesn't seem normal. What does this check actually do, and do we need it for docs PRs?

spetrosi:
@lzap, can you please clarify to us what this Redmine issue job does and take a look why it takes so long this time?

PR #172, opened 2020-05-20
Implement the new {foreman-client} attribute

spetrosi:
Get rid of the workaround that listed the foreman-client repository for el7 only, which might be confusing

Bug 1835718 - Make the wording of Satellite Tools repository consistent throughout the guides

https://bugzilla.redhat.com/show_bug.cgi?id=1835718

### DIFF
--- a/guides/common/attributes.adoc
+++ b/guides/common/attributes.adoc
@@ -66,8 +66,8 @@ ifeval::["{build}" == "foreman"]
 :smartproxy-example-com: smartproxy.example.com
 :smartproxy_port: 8443
 :Team: Foreman developers
-:RepoRHEL7ServerSatelliteToolsProductVersion: https://yum.theforeman.org/client/{ProjectVersion}/el7/x86_64/foreman-client-release.rpm
-:foreman-client: https://yum.theforeman.org/client/{ProjectVersion}/
+:project-client-RHEL7-url: https://yum.theforeman.org/client/{ProjectVersion}/el7/x86_64/foreman-client-release.rpm
+:project-client-name: https://yum.theforeman.org/client/{ProjectVersion}/
 endif::[]
 
 ifeval::["{build}" == "satellite"]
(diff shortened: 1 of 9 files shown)


guides/common/attributes.adoc
```diff
@@ -67,6 +67,7 @@ ifeval::["{build}" == "foreman"]
 :smartproxy_port: 8443
 :Team: Foreman developers
 :RepoRHEL7ServerSatelliteToolsProductVersion: https://yum.theforeman.org/client/{ProjectVersion}/el7/x86_64/foreman-client-release.rpm
+:foreman-client: https://yum.theforeman.org/client/{ProjectVersion}/
```
tahliar, line 70: Is there a name we can call this so it fits sentences the same way Satellite Tools does? Like "Foreman Client" or something? If not then I'm fine with the solution you've used here.

spetrosi: We can call it Foreman Client, the problem is, we do not define Foreman Client anywhere. There are procedures on enabling and syncing Satellite Tools repo, where we include some info about what this repo does, but there are no such procedures for Foreman Client. So  I assumed that we can just point to the URL of Foreman Client so that users can go and find the repo for their OSes. 
While Satellite Tools can be found in Red Hat repoes, Foreman Client cannot be found anywhere except the URL. Foreman Client also can be added to clients by yum, no need to enable and sync in on Foreman Server actually.

spetrosi:
> I'd probably go with `{foreman-client-repo-url}` and `{foreman-client-repo-name}` or something like that. This variable is really like some client command name rather than URL or repo name.

Right, I took an existing attribute, I should rename it for consistency. I also like the suggested attribute names, will do.

lzap:
Actually this should probably be project-client-repo-url or somethin like that. We try to avoid foreman and satellite in variables :-)

guides/common/attributes.adoc
ekohl: That's because historically Foreman itself has always avoided writing client tooling and relied on others to provide information/execution (like Puppet). You can see there's basically 3 packages there:
* Ansible collection (very recent)
* Katello agent/host tools
* SCAP client

It's only when the unification with Katello took more shape that a client tooling repo was created to replace katello-client. Even then, there's only one for RPMs and none on Debian.

In a vanilla Foreman setup I think very few users actually use the client repo.

Note that in Katello 4.0 (next version) we'll drop katello-agent as well which further limits the usefulness.

spetrosi: Thank you for the info @ekohl.Now we include a prerequisite in a number of procedures that use packages from this client repo. We can reconsider this in future than things change.

lzap:
LGTM

PR #179, opened 2020-05-27
Bz#1825223 Convert Installing Satellite Server chapter to mod docs

tahliar:
Bug 1825223 - Modularise "Installing Satellite Server"

https://bugzilla.redhat.com/show_bug.cgi?id=1825223


tahliar:
Travis's broken links are expected, as they were changed in this PR so are not live yet.

guides/common/assembly_installing-satellite-server-connected.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+Use the following procedures to install {ProjectServer}, perform the initial configuration, and import subscription manifests.
+
+ifeval::["{build}" != "foreman-deb"]
+Before you continue, consider which manifests are relevant for your environment. For more information on manifests, see link:{BaseURL}content_management_guide/managing_subscriptions/[Managing Subscriptions] in the _Content Management Guide_.
+endif::[]
+
+
+NOTE: You cannot register {ProjectServer} to itself.
+endif::[]
```
spetrosi, line 15: There is `ifeval::["{build}" != "foreman-deb"]` statement inside of ` ifeval::["{build}" == "satellite"]`.  This does not make sense because the content inside of the first ` ifeval::["{build}" == "satellite"]` will be shown in satellite build only.

guides/common/assembly_installing-satellite-server-connected.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+ifeval::["{build}" != "satellite"]
+Use the following procedures to install {ProjectServer} and perform the initial configuration.
```
spetrosi, line 18: This sentence is included twice, once in `ifeval::["{build}" == "satellite"]` on line 7, then here in `ifeval::["{build}" != "satellite"]`. Can you put it out from ifevals for simplicity?

guides/common/assembly_installing-satellite-server-connected.adoc
```diff
@@ -0,0 +1,86 @@
...
+endif::[]
+
+include::modules/proc_installing-the-satellite-server-packages.adoc[leveloffset=+1]
+
+include::modules/proc_synchronizing-the-system-clock-with-chronyd.adoc[leveloffset=+1]
+
+include::modules/proc_installing-the-sos-package.adoc[leveloffset=+1]
+
+[id="configuring-server_{context}"]
+== Configuring {ProjectServer}
```
spetrosi, line 69: Why is this section not a separate assembly?

guides/common/assembly_installing-satellite-server-disconnected.adoc
```diff
@@ -0,0 +1,45 @@
...
+include::modules/proc_installing-from-the-offline-repositories.adoc[leveloffset=+1]
+
+include::modules/proc_resolving-package-dependancy-errors.adoc[leveloffset=+1]
+
+include::modules/proc_synchronizing-the-system-clock-with-chronyd.adoc[leveloffset=+1]
+
+include::modules/proc_installing-the-sos-package.adoc[leveloffset=+1]
+
+[id="configuring-server_{context}"]
+== Configuring {ProjectServer}
```
spetrosi, line 28: Why is this not in a separate assembly?

spetrosi: Also, the content of Configuring {ProjectServer} for connected and disconnected is vry similar, seems like there is a chance to reuse.

guides/common/modules/proc_configuring-satellite-manually.adoc
```diff
@@ -0,0 +1,56 @@
...
+--foreman-initial-location "_initial_location_name_" \
+--foreman-initial-admin-username _admin_user_name_ \
+--foreman-initial-admin-password _admin_password_
+----
++
+ifeval::["{build}" == "satellite"]
+The script displays its progress and writes logs to `/var/log/foreman-installer/satellite.log`.
+endif::[]
+ifeval::["{build}" == "foreman"]
+The script displays its progress and writes logs to `/var/log/foreman-installer/foreman-installer.log`.
```
spetrosi, line 45: @lzap on my Foreman, I cannot locate `/var/log/foreman-installer/foreman-installer.log`, only `/var/log/foreman-installer/foreman.log`. Can you please verify this? If this is the case, we can use attribute in the form `/var/log/foreman-installer/{project-context}.log`

guides/common/assembly_installing-satellite-server-connected.adoc
tahliar: Oops! That was an error :smile:

guides/common/assembly_installing-satellite-server-connected.adoc
tahliar: They are slightly different sentences (the first one includes "and import subscription manifests"), but actually the second one should only be != foreman-deb so I do need to adjust it.

guides/common/assembly_installing-satellite-server-connected.adoc
tahliar: Because it isn't one. An assembly is meant to be a start-to-finish user story, and although we do stretch that definition a bit in some of our earlier assemblies, in this case it's well outside that definition as it's two different options for doing the same thing.

guides/common/assembly_installing-satellite-server-disconnected.adoc
tahliar: I considered putting it in a concept file, but it would only ever be used in this context, so I think putting it directly in the installing-satellite-server assembly is the convention in that case (or it was last time I did the training). I'll have a think about it though.

guides/common/assembly_installing-satellite-server-connected.adoc
spetrosi: Got it, thank you

guides/common/assembly_installing-satellite-server-disconnected.adoc
spetrosi: Thank you for the explanation

guides/common/assembly_installing-satellite-server-connected.adoc
tahliar: I ended up moving the ifevals around into a more manageable order. There were so many "ifs" for manifests in Foreman (katello + red hat OS) that I decided to just not mention it in the intro for the foreman build as it seemed like overkill.

PR #184, opened 2020-05-29
Improve upstreamize script - options statement

lzap:
SSIA


spetrosi:
Works perfectly, I will follow up with a PR that actually runs this script. Thanks a lot @lzap

PR #198, opened 2020-06-18
Add proxy configuration for PXE smart proxy

lzap:
This was brought by a user on IRC. It is not relevant downstream as the only content we support is RHEL via CDN and that goes into Katello which do not need to be proxied at all.


melcorr:
For upstream only content, would I have permission to merge @spetrosi @tahliar @suyogsainkar ?

tahliar:
> For upstream only content, would I have permission to merge @spetrosi @tahliar @suyogsainkar ?

Sounds reasonable to me, if the others agree too.

spetrosi:
@melcorr Do it!

Issue #201, opened 2020-06-20
Tech Previews still listed upstream in Admin guide


melcorr:
In the Administrating guide, tech previews are still listed: 

http://docs.theforeman.org/guides/build/doc-Administering_Red_Hat_Satellite/index-foreman.html#chap-Red_Hat_Satellite-Administering_Red_Hat_Satellite-Configuring_External_Authentication

PR #202, opened 2020-06-22
Hide tech preview upstream

melcorr:
As part of #201


tahliar:
Now Travis is upset about http://yum.postgresql.org/9.5/redhat/ but I can access it just fine.

PR #205, opened 2020-06-23
Use attribute for yum when manipulating packages from Capsules

spetrosi:
Bug 1825798 - [RFE] Document that satellite-maintain package feature on Capsule`

https://bugzilla.redhat.com/show_bug.cgi?id=1825798

### DIFF
--- a/guides/common/attributes.adoc
+++ b/guides/common/attributes.adoc
@@ -30,7 +30,6 @@ ifeval::["{build}" == "foreman"]
 :project-context: foreman
 :foreman-installer: foreman-installer
 :foreman-maintain: foreman-maintain
-:packages-install: yum install
 :foreman-example-com: foreman.example.com
 :installer-scenario: foreman-installer --scenario katello
 :installer-scenario-smartproxy: foreman-installer --no-enable-foreman
@@ -38,6 +37,8 @@ ifeval::["{build}" == "foreman"]
 :package-update: yum update
 :package-clean: yum clean
 :package-remove: yum remove
+:package-install-project: yum install
+:package-remove-project: yum remove
 :certs-generate: foreman-proxy-certs-generate
 :certs-proxy-context: foreman-proxy
 :project-change-hostname: katello-change-hostname
@@ -84,11 +85,12 @@ ifeval::["{build}" == "satellite"]
 :package-install: yum install
 :package-update: yum update
 :package-clean: yum clean
-:package-remove: satellite-maintain packages remove
+:package-remove: yum remove
+:package-install-project: satellite-maintain packages install
+:package-remove-project: satellite-maintain packages remove
 :certs-generate: capsule-certs-generate
 :certs-proxy-context: capsule
 :foreman-maintain: satellite-maintain
-:packages-install: satellite-maintain packages install
 :foreman-example-com: satellite.example.com
 :project-change-hostname: satellite-change-hostname
 :lorax-composer: Red{nbsp}Hat Image Builder
@@ -135,8 +137,10 @@ ifeval::["{build}" == "foreman-deb"]
 :package-install: apt-get install
 :package-update: apt-get update
 :package-clean: apt-get clean
-:certs-generate: foreman-proxy-certs-generate
 :package-remove: apt-get remove
+:package-install-project: apt-get install
+:package-remove-project: apt-get remove
+:certs-generate: foreman-proxy-certs-generate
 :installer-scenario-smartproxy: foreman-installer --no-enable-foreman
 :project-change-hostname: katello-change-hostname
 :oVirt: oVirt
--- a/guides/common/attributes.adoc
+++ b/guides/common/attributes.adoc
@@ -84,7 +84,7 @@ ifeval::["{build}" == "satellite"]
 :package-install: yum install
 :package-update: yum update
 :package-clean: yum clean
-:package-remove: yum remove
+:package-remove: satellite-maintain packages remove
 :certs-generate: capsule-certs-generate
 :certs-proxy-context: capsule
 :foreman-maintain: satellite-maintain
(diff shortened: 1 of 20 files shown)


guides/common/attributes.adoc
```diff
@@ -84,7 +84,7 @@ ifeval::["{build}" == "satellite"]
 :package-install: yum install
```
suyogsainkar, line 84: @spetrosi Hi Sergei, similar to package-remove, does package-install and package-update also need updating with satellite-maintain packages install/update? Everything else looks good to me.

guides/common/attributes.adoc
```diff
@@ -135,8 +137,10 @@ ifeval::["{build}" == "foreman-deb"]
 :package-install: apt-get install
 :package-update: apt-get update
 :package-clean: apt-get clean
-:certs-generate: foreman-proxy-certs-generate
 :package-remove: apt-get remove
+:package-install-project: apt-get install
+:package-remove-project: apt-get remove
+:certs-generate: foreman-proxy-certs-generate
```
ekohl, line 143: This is a bit misleading. This simply doesn't exist on Debian and there is no equivalent.

guides/doc-Managing_Hosts/topics/Synchronizing_Templates.adoc
```diff
@@ -334,6 +334,7 @@ To avoid errors after removing the foreman_templates plugin:
 
 . Uninstall the plug-in:
 +
+[options="nowrap" subs="+quotes,attributes"]
 ----
-# yum remove tfm-rubygem-foreman_templates
+# {package-remove-project} tfm-rubygem-foreman_templates
```
ekohl, line 339: This only works on EL7. The package name is different on EL8 and Debian. On RPMs you can use the virtual:
```suggestion
# {package-remove-project} foreman-plugin-templates
```

(different plugin, but same story)
```
[root@foreman ~]# yum remove foreman-plugin-monitoring
Loaded plugins: fastestmirror
Resolving Dependencies
--> Running transaction check
---> Package tfm-rubygem-foreman_monitoring.noarch 0:1.0.1-4.fm2_1.el7 will be erased
--> Finished Dependency Resolution

Dependencies Resolved

======================================================================================================================================================================================
 Package                                                 Arch                            Version                                      Repository                                 Size
======================================================================================================================================================================================
Removing:
 tfm-rubygem-foreman_monitoring                          noarch                          1.0.1-4.fm2_1.el7                            @foreman-plugins                          459 k

Transaction Summary
======================================================================================================================================================================================
Remove  1 Package

Installed size: 459 k
```

evgeni: I think that's a good idea.

guides/common/attributes.adoc
spetrosi: Elaborate on this, please. This is used to make the guides that build for Debian (those are the install guides only)  to render both the `package-install-project` and `package-install` attributes as `apt-get install`. And the same for remove.

guides/doc-Administering_Red_Hat_Satellite/topics/Maintenance.adoc
```diff
@@ -102,24 +102,24 @@ If you use an untypical file system (other than for example ext3, ext4, or xfs),
...
+=== Managing Packages on the Base Operating System of {Project} or {SmartProxy}
 
-To install and update packages on the {Project} base operating system, you must enter the `{foreman-maintain} packages` command.
+To install and update packages on the {Project} or {SmartProxy} base operating system, you must enter the `{foreman-maintain} packages` command.
 
-{Project} prevents users from installing and updating packages with `yum` because `yum` might also update the packages related to {Project} and result in system inconsistency.
+{Project} prevents users from installing and updating packages with `yum` because `yum` might also update the packages related to {Project} or {SmartProxy} and result in system inconsistency.
 
-IMPORTANT: The `{foreman-maintain} packages` command restarts some services on {ProjectServer} because it runs the `{foreman-installer} --upgrade` command after installing packages.
+IMPORTANT: The `{foreman-maintain} packages` command restarts some services on the operating system where you run it because it runs the `{foreman-installer} --upgrade` command after installing packages.
```
ekohl, line 111: Using `--upgrade` with `foreman-installer` is deprecated. Since Foreman 2.1 the flag is just a dummy and will be dropped entirely in the future (https://projects.theforeman.org/issues/29296).

spetrosi: Nice catch, Is saying that the `satellite-maintain packages` command runs `satellite-installer` is accurate then?

spetrosi: This is what the script itself tells:
WARNING: This script runs satellite-installer after the yum execution 
to ensure the Satellite is in a consistent state.
I have added a commit to fix this.

guides/common/attributes.adoc
spetrosi: Good question. Technically it makes sense because you can run `satellite-maintain packages update`, but we do not use package updates in our docs. This is because the `satellite-maintain packages install` command does the same thing - checks if the package is installed and updates it to a latest version if it is not the latest version already. So we do not need attributes for an update.

guides/common/attributes.adoc
ekohl: Specifically `foreman-proxy-certs-generate` does not exist. There is no CA tooling. The closest we get is using the Puppet CA but I don't think you want to write up instructions for that.

guides/doc-Administering_Red_Hat_Satellite/topics/Maintenance.adoc
ekohl: That's indeed the best way.

guides/common/attributes.adoc
spetrosi: I've opened the issue #213 to address that and other your feedback.

guides/common/attributes.adoc
suyogsainkar: @spetrosi OK, understood - thanks.

spetrosi:
The broken link is expected, I work on it currently and will fix it in another PR when ready.

PR #207, opened 2020-06-26
Clarify DNS providers - ISC DNS vs ISC BIND

illumino:
The Internet Systems Consortium produce a DNS server (and related tools) distributed as ISC BIND (it used to be an acronym, see BIND on Wikipedia, but it is no longer developed by UCB, rather ISC). A google search for "ISC DNS" will take you to the ISC BIND9 page, so I don't feel we should confuse readers with separate references to ISC DNS (a name possibly unique to this document) and the actual ISC product.

### DIFF
--- a/guides/doc-Planning_Guide/topics/Capsule_Server_Overview.adoc
+++ b/guides/doc-Planning_Guide/topics/Capsule_Server_Overview.adoc
@@ -20,7 +20,7 @@ Infrastructure and host management services:
 
 * *DHCP* – {SmartProxy} can act as a DHCP server or it can integrate with an existing solution, including ISC DHCP servers, Active Directory, and Libvirt instances.
 
-* *DNS* – {SmartProxy} can act as a DNS server or it can integrate with an existing solution, including ISC BIND or Active Directory.
+* *DNS* – {SmartProxy} can act as a DNS server or it can integrate with an existing solution, including ISC BIND and Active Directory.
 
 * *TFTP* – {SmartProxy} can act as a TFTP server or integrate with any UNIX-based TFTP server.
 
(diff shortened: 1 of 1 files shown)


illumino:
1. I wasn't sure if I should use "including ISC BIND _or_ Active Directory" which matches the original wording, or "including ISC BIND _and_ Active Directory" - which sounds better to me (for what that is worth)
2. CI Breaks seem unrelated to my simple change, so I haven't looked into them.

guides/doc-Planning_Guide/topics/Capsule_Server_Overview.adoc
```diff
@@ -20,7 +20,7 @@ Infrastructure and host management services:
 
 * *DHCP* – {SmartProxy} can act as a DHCP server or it can integrate with an existing solution, including ISC DHCP servers, Active Directory, and Libvirt instances.
 
-* *DNS* – {SmartProxy} can act as a DNS server or it can integrate with an existing solution, including ISC DNS, Active Directory, or BIND.
+* *DNS* – {SmartProxy} can act as a DNS server or it can integrate with an existing solution, including ISC BIND or Active Directory.
```
spetrosi, line 23:
```suggestion
* *DNS* – {SmartProxy} can act as a DNS server or it can integrate with an existing solution, including ISC BIND and Active Directory.
```
You are totally correct about using `and` here, it works better.

ekohl: Technically the Smart Proxy provider we use is using nsupdate so any [RFC2136](https://tools.ietf.org/html/rfc2136) compatible nameserver can be used. ISC bind is the one where we can also manage the server.

IMHO `{SmartProxy} can act as a DNS server is incorrect. The Smart Proxy can only integrate with DNS servers. The installer can deploy ISC bind on the same server. The same goes for DHCP and TFTP.

melcorr:
Thanks @illumino for providing this update!

PR #212, opened 2020-07-03
Ignore access.redhat.com/solutions

lzap:
These links are often 401 (paywalled), the tool is unable to check them as there is no way to configure it that 401 should be considered as success. This removes these links.

Example failure:

```
URL        `https://access.redhat.com/solutions/3397771'
Name       `Impact of Disk Speed on Satellite Operations'
Parent URL file:///home/travis/build/theforeman/foreman-documentation/guides/build/doc-Installing_Proxy_on_Debian/index-satellite.html, line 219, col 101
Real URL   https://access.redhat.com/solutions/3397771
Check time 2.051 seconds
Result     Error: 403 Forbidden
```

### DIFF
--- a/guides/common/Makefile
+++ b/guides/common/Makefile
@@ -36,7 +36,7 @@ browser: html
 	${BROWSER_OPEN} "file://$(realpath $(ROOTDIR)/$(DEST_HTML))"
 
 linkchecker: html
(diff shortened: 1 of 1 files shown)


melcorr:
Thanks for this!

PR #214, opened 2020-07-10
Fix yum install in Managing Hosts guide

spetrosi:
Bug 1825798 - [RFE] Document that satellite-maintain package feature on Capsule

https://bugzilla.redhat.com/show_bug.cgi?id=1825798


spetrosi:
Hi @suyogsainkar, I found one more instance of `yum` that must be changed upstream.

PR #215, opened 2020-07-22
Update docs to correspond to a new UI label name

spetrosi:
Bug 1852359 - [Docs] [Administering Red Hat Satellite] Update references to the impersonation UI label

https://bugzilla.redhat.com/show_bug.cgi?id=1852359


spetrosi:
> @spetrosi looks good to me. I noticed that after you click the impersonation icon in the UI, a popup window is displayed to confirm the action/stop session - but I think the UI is self explanatory so there's no need to document that step. Merging now.

Good catch! And a good notice that this is self-explanatory too. Thank you for the review.

PR #217, opened 2020-07-23
Fixes 404 link for AWS support site

lzap:
The article no longer exists. This fixes the link in the text.

Also this is a refactoring of Makefile, linkchecker is now called via `xargs` which gives much better results for viewing (single call). Also this supports now linkchecker-tryer which is a tool that tries to retry on links when error is returned.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,15 @@
 ---
-language: minimal
+language: python
+python:
+  - "2.7"
 before_install:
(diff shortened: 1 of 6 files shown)


lzap:
Do not merge yet.

lzap:
I merged this, this improves link checking in a way that it tries several time before it marks a link as invalid. This solves varous timeout errors.

PR #221, opened 2020-07-31
Use attribute for installing package on Satellite

spetrosi:
Bug 1859359 - [DDF] This command fails, I'm presuming because of package lock. The command in step 2 installs it if missing though

https://bugzilla.redhat.com/show_bug.cgi?id=1859359

### DIFF
--- a/guides/doc-Managing_Hosts/topics/Running_Jobs_on_Hosts.adoc
+++ b/guides/doc-Managing_Hosts/topics/Running_Jobs_on_Hosts.adoc
@@ -273,8 +273,9 @@ To set up {Project} to use Kerberos authentication for remote execution on hosts
 
 . To install the `tfm-rubygem-net-ssh-krb` package, enter the following command:
 +
+[options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-# yum install tfm-rubygem-net-ssh-krb
+# {packages-install} tfm-rubygem-net-ssh-krb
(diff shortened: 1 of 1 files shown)


guides/doc-Managing_Hosts/topics/Running_Jobs_on_Hosts.adoc
```diff
@@ -273,8 +273,9 @@ To set up {Project} to use Kerberos authentication for remote execution on hosts
 
 . To install the `tfm-rubygem-net-ssh-krb` package, enter the following command:
 +
+[options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-# yum install tfm-rubygem-net-ssh-krb
+# {packages-install} tfm-rubygem-net-ssh-krb
 ----
 +
```
ekohl, line 280: I don't understand the point of this. Why do you need to do this if running the installer with `--foreman-proxy-plugin-remote-execution-ssh-ssh-kerberos-auth true` already does this? An additional benefit is that it uses the correct package name (the current example is incorrect on RHEL8 and Debian). Just drop this?

spetrosi: Thank you @ekohl! Good to know this, we run the installer in the next step, so this first step is completely redundant. I will remove it on master too in a separate PR.

PR #227, opened 2020-08-05
properly add the cname at the right spot

evgeni:
@lzap as one couldn't test it -- it didn't work ;-)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -131,9 +131,6 @@ jobs:
       - name: Delete all files
         run: rm -rf *
 
-      - name: Add CNAME for docs.theforeman.org
-        run: echo docs.theforeman.org > CNAME
-
       - name: Create directory for branch
         run: mkdir -p public/master
 
--- a/web/CNAME
+++ b/web/CNAME
@@ -1,0 +1,1 @@
+docs.theforeman.org


.github/workflows/deploy.yml
```diff
@@ -131,9 +131,6 @@ jobs:
       - name: Delete all files
         run: rm -rf *
 
-      - name: Add CNAME for docs.theforeman.org
-        run: echo docs.theforeman.org > CNAME
```
ekohl, line 135: The alternative to this is to create `public/master/CNAME`. Of course you need to move this to below the `mkdir -p` statement.

evgeni: `public/CNAME`, without the master, but I think just having the file is cleaner?

evgeni:
btw, @lzap there is a `.nojekyll` file in `gh-pages` which probably should get the same treatment? (it's not removed as you don't remove dot files, but still weird).

.github/workflows/deploy.yml
ekohl: Yes, probably is.

evgeni:
Seems it did?

spetrosi:
Yup all works! Thank you for fixing this @evgeni

ekohl:
It was added here:
https://github.com/theforeman/foreman-documentation/commit/60c56d312fc0ce6cc85108344934e33847ef1763#diff-adc4bfdb0829dae99e3699393e3fbaa4

PR #231, opened 2020-08-10
[WIP] Add a task to upload single html files from each build dir

spetrosi:


lzap:
What your intentions are? Maybe I can help?

spetrosi:
> What your intentions are? Maybe I can help?

I was trying to upload single HTML files as artefacts. This way, one can simply open an artefact and see the built document right away. One thing I do not know is if the artefact will be uploaded as a single file or as an archive. I did this at the end of my working hours so did not have time to test properly, and then closed because did not want someone to merge it by accident.

ekohl:
Nowadays you can create a PR as a draft: https://github.blog/2019-02-14-introducing-draft-pull-requests/

IMHO this is nicer than using `[WIP]` because it actually prevents accidental merges and clearly communicates the intent.

As for the actual intention: based on my experiments it's always a zip, even if it's a single file.

spetrosi:
This was my intention. I even saw the *Create as draft* button when creating the PR, but was afraid that the checks would not work this way. And when the PR was created, I did not find any way to set it to be a draft PR.

spetrosi:
It's a shame that Git archives a single file.

PR #232, opened 2020-08-11
Add command that patches Satellite base OS

spetrosi:
Mark the whole section Managing Packages on the Base Operating System of Satellite Server to be built for Satellite only as package lock is not enabled on Foreman by default

Bug 1859747 - [RFE] Add a section on patching the Operating System in the "Administering Red Hat Satellite documentation

https://bugzilla.redhat.com/show_bug.cgi?id=1859747

### DIFF
--- a/guides/common/attributes.adoc
+++ b/guides/common/attributes.adoc
@@ -38,6 +38,7 @@ ifeval::["{build}" == "foreman"]
 :package-clean: yum clean
 :package-remove: yum remove
 :package-install-project: yum install
+:package-update-project: yum update
 :package-remove-project: yum remove
 :certs-generate: foreman-proxy-certs-generate
 :certs-proxy-context: foreman-proxy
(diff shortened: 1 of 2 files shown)


spetrosi:
The PR has been approved by @lzap and @suyogsainkar, so I am merging it now.

guides/common/attributes.adoc
```diff
@@ -139,6 +141,7 @@ ifeval::["{build}" == "foreman-deb"]
 :package-clean: apt-get clean
 :package-remove: apt-get remove
 :package-install-project: apt-get install
+:package-update-project: apt-get update
```
ekohl, line 144: `apt-get update` only updates the metadata. You're looking for `apt-get upgrade`.

Technically speaking `apt-get upgrade` is equivalent to `yum update --cacheonly`

lzap: This is a good catch, can we fix this @spetrosi please?

suyogsainkar:
@spetrosi the updates look good to me from a peer review point of view. once @ekohl / @lzap's suggestion is applied, the bug can be moved to verified. thanks.

guides/common/attributes.adoc
spetrosi: Please merge #255 to fix it @lzap

PR #238, opened 2020-08-12
Add help to satellite-sync and ensure that it copies itself to…

spetrosi:
… the downstream repo


guides/satellite-sync.sh
```diff
@@ -3,14 +3,49 @@
...
+# Hacked repeately by Melanie Corr to accommodate transitions towards Upstream-Downstream Satellite
+# To use the script to synchronize changes from the upstream foreman-documentation GitHub repository to the downstream docs-Red_Hat_Satellite_6 repository, complete the following steps:"
+# 1. Create a directory to store repositories for synchronization. You can name the directory whatever you want, for example SyncCenter:"
+# $ mkdir ~/Documents/SyncCenter"
+#
+# 2. Copy this script to the SyncCenter directory."
+#
+# 3. Change to the SyncCenter directory."
+#
+# 4. In the upstream repository, ensure that the commits that you want to synchronized are cherry-picked to the related branch, for example, to the SATELLITE-6.8-beta branch. In the upstream repository, all related to downstream branches are in the format 'SATELLITE-x.y'"
```
suyogsainkar, line 15: In the upstream repository, ensure that the commits you want to synchronize are first cherry-picked to the branch that corresponds to the downstream branch, for example, to the SATELLITE-6.8-beta branch. In the upstream repository, branches that correspond to the downstream branches are in the format 'SATELLITE-x.y'"

guides/satellite-sync.sh
```diff
@@ -3,14 +3,49 @@
...
+  echo "1. Create a directory to store repositories for synchronization. You can name the directory whatever you want, for example SyncCenter:"
+  echo "$ mkdir ~/Documents/SyncCenter"
+  echo ""
+  echo "2. Copy this script to the SyncCenter directory."
+  echo ""
+  echo "3. Change to the SyncCenter directory."
+  echo ""
+  echo "4. In the upstream repository, ensure that the commits that you want to synchronized are cherry-picked to the related branch, for example, to the SATELLITE-6.8-beta branch. In the upstream repository, all related to downstream branches are in the format 'SATELLITE-x.y'"
+  echo ""
+  echo "5. Run the script and specify the upstream branch to copy changes from and the downstream branch to copy changes to. For example, to synchronize between the upstream SATELLITE-6.8-beta branch and the downstream master branch, enter the following command:"
```
suyogsainkar, line 41: I understand this is just an example. But in reality, do we commonly sync between the upstream SATELLITE-6.8-beta branch and the downstream master branch, or the downstream 6.8-beta branch? I was unsure myself so I didn't try out the syncing step. If it's the latter, I would suggest updating the text and example to reflect equivalent beta branches to avoid confusion.

spetrosi: The master branch is the branch that is supposed to have all the latest changes. Every commit that belong to the very recent Satellite version, in our case it is unreleased 6.8-beta, should be cherry-picked to master first. So you always need to synchronize changes to master anyway.
The overall structure of repositories look like this:
Upstream `master` is in sync with downstream `master`
Upstream `SATELLITE-6.7` is in sync with downstream `6.7`
Upstream `SATELLITE-6.8-beta` is in sync with downstream `6.8-beta`
In general, upstream `SATELLITE-x.y` is in sync with downstream `x.y`.

That is the state that we need to maintain. Here is one method how you can maintain it:

1. Upstream, create branch with your work from `master`, push changes, create a PR and have it merged. Note that if you need to change something for an older version ONLY, for example if the change is related to `6.7` but should not be reflected on `master`, you push this commit to `SATELLITE-6.7` directly, though this is not a common scenario.

2. After the PR is merged, cherry-pick the changes to the relevant upstream branches.

- The change can be only relevant to the `6.8-beta` version of Satellite, for examle the Foreman Ansible Modulse PR. In this case, only cherry-pick it to `SATELLITE-6.8-beta`. 
- The change can also be relevant to more versions, for example to `6.7` and `6.8-beta`. In this case, cherry-pick the change to all relevant SATELLITE branches.

3. When the changes are in the relevant branches, you can use the `satellite-sync.sh` script to copy them to downstream branches. You can do this by running a command for each branch, for example, for Foreman Ansible modules PR, you will need to run those two commands:
`sh satellite-sync.sh master master`
`sh satellite-sync.sh SATELLITE-6.8-beta 6.8-beta`

That's it, I will try to implement this somehow to the satellite-sync readme.

guides/scripts/satellite-sync.sh
```diff
@@ -90,12 +112,12 @@ if ls $US_REPO/guides/doc-* 1> /dev/null 2>&1; then
...
+  cp $US_REPO/guides/scripts/satellite-sync.sh $DS_REPO/scripts/satellite-sync.sh
 fi
 # Add downstream build attributes and find and replace attributes with fixed terms
 
 cd $DS_REPO
-sed 's/:SatelliteAnsibleVersion: 2.8/:SatelliteAnsibleVersion: 2.8\n:build: satellite/g' common/attributes.adoc > common/attributes.tmp
+sed '1s/^/\:build\: satellite\n/' common/attributes.adoc > common/attributes.tmp
 mv common/attributes.tmp common/attributes.adoc
 sed 's/:TargetVersion: 6.7-beta/:TargetVersion: 6.7/g' common/attributes.adoc > common/attributes.tmp
 mv common/attributes.tmp common/attributes.adoc
```
spetrosi, line 123: @melcorr What does lines 122-127 in this file do? Please elaborate why they replace version numbers. I would probably add comments about it to make this clear for future generations.

Note the change that I've done on line 120. The attribute `SatelliteAnsibleVersion` will change with time, so instead of using it as a reference point, I made `sed` to add the line `:build\: satellite` in the beginning of attributes.adoc and put a newline after it.

guides/satellite-sync.sh
spetrosi: Done.

guides/scripts/satellite-sync.sh
melcorr: Sergei, the link checker fails and all the docs links are broken if we use the latest release number in the upstream. 
This was the workaround we came up with. 
I have absolutely no idea what you mean by build there. Sorry. Makes no sense to me.

spetrosi: > Sergei, the link checker fails and all the docs links are broken if we use the latest release number in the upstream.
> This was the workaround we came up with.
Got it. For this release, I solved it by keeping the working `6.7` version on the upstream master branch, but changing the version to `6.8-beta` on upstream SATELLITE-6.8-beta and downstream 6.8-beta branches. I will probably remove this workaround to avoid confusion.
> I have absolutely no idea what you mean by build there. Sorry. Makes no sense to me.
Anyway, this works :) So the `sed` command on line 129 inserts the line `:build: satellite` after the line `:SatelliteAnsibleVersion: 2.8` in the downstream attributes.adoc. I changed it to insert `:build: satellite` at the beginning of the attributes.adoc file.

melcorr: All of this was experimental at the time. If it works for you, it works :)

PR #246, opened 2020-08-24
Add section that configures alternate CNAME

spetrosi:
Bug 1828219 - [RFE] Document using multi-homed

https://bugzilla.redhat.com/show_bug.cgi?id=1828219

### DIFF
--- a/guides/common/assembly_configuring-an-alternate-cname.adoc
+++ b/guides/common/assembly_configuring-an-alternate-cname.adoc
@@ -1,6 +1,6 @@
 [id='configuring-an-alternate-cname_{context}']
 = Configure an Alternate CNAME for {Project}
-You can configure an alternate CNAME for {Project}. This might be useful if you want to deploy the {Project} web interface on a different domain name than the one that is used by client systems to connect to {Project}.
+You can configure an alternate CNAME for {Project}. This might be useful if you want to deploy the {Project} web interface on a different domain name than the one that is used by client systems to connect to {Project}. You must plan the alternate CNAME configuration in advance prior to installing {SmartProxies} and registering hosts to {Project} to avoid redeploying new certificates to hosts.
 
 ifeval::["{build}" != "satellite"]
 This procedure is only for Katello plug-in users.
--- a/guides/common/assembly_configuring-an-alternate-cname.adoc
+++ b/guides/common/assembly_configuring-an-alternate-cname.adoc
@@ -6,5 +6,5 @@ ifeval::["{build}" != "satellite"]
 This procedure is only for Katello plug-in users.
 endif::[]
 
-include::modules/proc_configuring-satellite-with-an-alternate-cname.adoc[leveloffset=+1]
+include::modules/proc_configuring-project-with-an-alternate-cname.adoc[leveloffset=+1]
 include::modules/proc_configuring-hosts-to-use-an-alternate-cname-for-content-management.adoc[leveloffset=+1]
--- a/guides/common/assembly_configuring-an-alternate-cname.adoc
+++ b/guides/common/assembly_configuring-an-alternate-cname.adoc
@@ -1,0 +1,6 @@
+[id='configuring-an-alternate-cname_{context}']
+= Configure an Alternate CNAME for {Project}
+You can configure an alternate CNAME for {Project}. This can be useful if <placeholder>.
+
+include::modules/proc_configuring-satellite-with-an-alternate-cname.adoc[leveloffset=+1]
+include::modules/proc_configuring-hosts-to-use-an-alternate-cname-for-content-management.adoc[leveloffset=+1]
--- a/guides/common/assembly_configuring-an-alternate-cname.adoc
+++ b/guides/common/assembly_configuring-an-alternate-cname.adoc
@@ -1,6 +1,10 @@
 [id='configuring-an-alternate-cname_{context}']
 = Configure an Alternate CNAME for {Project}
-You can configure an alternate CNAME for {Project}. This can be useful if <placeholder>.
+You can configure an alternate CNAME for {Project}. This might be useful if you want to deploy the {Project} web interface on a different domain name than the one that is used by client systems to connect to {Project}.
+
+ifeval::["{build}" != "satellite"]
+This procedure is only for Katello plug-in users.
+endif::[]
 
 include::modules/proc_configuring-satellite-with-an-alternate-cname.adoc[leveloffset=+1]
 include::modules/proc_configuring-hosts-to-use-an-alternate-cname-for-content-management.adoc[leveloffset=+1]
--- a/guides/common/modules/proc_configuring-hosts-to-use-an-alternate-cname-for-content-management.adoc
+++ b/guides/common/modules/proc_configuring-hosts-to-use-an-alternate-cname-for-content-management.adoc
@@ -1,7 +1,7 @@
 [id='configuring-hosts-to-use-an-alternate-cname-for-content-management_{context}']
 = Configuring Hosts to Use an Alternate {Project} CNAME for Content Management
 
-If {Project} is configured with an alternate CNAME, you can configure hosts to use the alternate {Project} CNAME for content management. To do this, you must point hosts tot eh alternate CNAME prior to registering the hosts to {Project}. You can do this manually or using the bootstrap script.
+If {Project} is configured with an alternate CNAME, you can configure hosts to use the alternate {Project} CNAME for content management. To do this, you must point hosts to the alternate {Project} CNAME prior to registering the hosts to {Project}. You can do this using the bootstrap script or manually.
 
 .Configuring Hosts with the bootstrap Script
 
--- a/guides/common/modules/proc_configuring-hosts-to-use-an-alternate-cname-for-content-management.adoc
+++ b/guides/common/modules/proc_configuring-hosts-to-use-an-alternate-cname-for-content-management.adoc
@@ -1,11 +1,20 @@
 [id='configuring-hosts-to-use-an-alternate-cname-for-content-management_{context}']
-= Configuring Hosts to Use an Alternate CNAME for Content Management
+= Configuring Hosts to Use an Alternate {Project} CNAME for Content Management
 
-After you configure {ProjectServer} with an alternate CNAME, you must point Red{nbsp}Hat Subscription Manager on {Project} hosts to use the alternate name for content management. You can do this manually or using the bootstrap script.
+If {Project} is configured with an alternate CNAME, you can configure hosts to use the alternate {Project} CNAME for content management. To do this, you must point hosts tot eh alternate CNAME prior to registering the hosts to {Project}. You can do this manually or using the bootstrap script.
+
+.Configuring Hosts with the bootstrap Script
+
+On the host, run the bootstrap script with the `--server _alternate_fqdn.example.com_` option to register the host to the alternate {Project} CNAME:
+
+[options="nowrap" subs="+quotes,attributes"]
+----
+# ./bootstrap.py --server _alternate_fqdn.example.com_
+----
 
 .Configuring Hosts Manually
 
-On each host, edit the `/etc/rhsm/rhsm.conf` file to update `hostname` and `baseurl` settings to point to the alternate host name, for example:
+On the host, edit the `/etc/rhsm/rhsm.conf` file to update `hostname` and `baseurl` settings to point to the alternate host name, for example:
 
 [options="nowrap" subs="+quotes,attributes"]
 ----
@@ -20,11 +29,4 @@ _content omitted_
 baseurl=https://_alternate_fqdn.example.com_/pulp/repos
 ----
 
-.Configuring Hosts with the bootstrap Script
-
-On each host, run the bootstrap script with the `--server _alternate_fqdn.example.com_` option:
-
-[options="nowrap" subs="+quotes,attributes"]
-----
-# ./bootstrap.py --server _alternate_fqdn.example.com_
-----
+Now you can register the host with the `subscription-manager`.
--- a/guides/common/modules/proc_configuring-hosts-to-use-an-alternate-cname-for-content-management.adoc
+++ b/guides/common/modules/proc_configuring-hosts-to-use-an-alternate-cname-for-content-management.adoc
@@ -1,0 +1,24 @@
+[id='configuring-hosts-to-use-an-alternate-cname-for-content-management_{context}']
+= Configuring Hosts to Use an Alternate CNAME for Content Management
+
+After you configure {ProjectServer} with and alternate CNAME, you must point Red{nbsp}Hat Subscription Manager on {Project} hosts to use the alternate name for content management. You can do this manually or using a bootstrap script.
+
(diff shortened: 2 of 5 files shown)


guides/common/assembly_configuring-an-alternate-cname.adoc
```diff
@@ -0,0 +1,6 @@
+[id='configuring-an-alternate-cname_{context}']
+= Configure an Alternate CNAME for {Project}
+You can configure an alternate CNAME for {Project}. This can be useful if <placeholder>.
```
spetrosi, line 3: @evgeni can you suggest what this feature can be useful for in real life?

evgeni: This is useful if you want to run the {Project} web interface on a different name than the one used by client systems to connect to {Project}.

guides/common/modules/proc_configuring-hosts-to-use-an-alternate-cname-for-content-management.adoc
```diff
@@ -0,0 +1,24 @@
...
+
+After you configure {ProjectServer} with and alternate CNAME, you must point Red{nbsp}Hat Subscription Manager on {Project} hosts to use the alternate name for content management. You can do this manually or using a bootstrap script.
+
+.Configuring Hosts Manually
+
+On each host, in the `/etc/rhsm/rhsm.conf` file, edit the `baseurl` setting to point to the alternate host name, for example:
+
+[options="nowrap" subs="+quotes,attributes"]
+----
+[rhsm]
```
evgeni, line 12: You also need to change the following entry in the `server` section:

```ini
[server]
# Server hostname:
hostname = _alternate_fqdn.example.com
```

guides/common/modules/proc_configuring-hosts-to-use-an-alternate-cname-for-content-management.adoc
```diff
@@ -0,0 +1,24 @@
...
+# Content base URL:
+baseurl=https://https://_alternate_fqdn.example.com_/pulp/repos
```
evgeni, line 14:
```suggestion
baseurl=https://_alternate_fqdn.example.com_/pulp/repos
```

guides/common/modules/proc_configuring-hosts-to-use-an-alternate-cname-for-content-management.adoc
```diff
@@ -0,0 +1,24 @@
...
+----
+
+.Configuring Hosts with a bootstrap Script
+
+On each host, run the bootstrap script with the `--server https://_alternate_fqdn.example.com_` option:
```
evgeni, line 19:
```suggestion
On each host, run the bootstrap script with the `--server _alternate_fqdn.example.com_` option:
```

no https needed, bootstrap will figure that out itself.

guides/common/modules/proc_configuring-hosts-to-use-an-alternate-cname-for-content-management.adoc
```diff
@@ -0,0 +1,24 @@
...
+On each host, in the `/etc/rhsm/rhsm.conf` file, edit the `baseurl` setting to point to the alternate host name, for example:
+
+[options="nowrap" subs="+quotes,attributes"]
+----
+[rhsm]
+# Content base URL:
+baseurl=https://https://_alternate_fqdn.example.com_/pulp/repos
+----
+
+.Configuring Hosts with a bootstrap Script
```
evgeni, line 17: I wonder if it should be "with THE bootstrap" script?

guides/common/modules/proc_configuring-satellite-with-an-alternate-cname.adoc
```diff
@@ -0,0 +1,16 @@
+[id='configuring-satellite-with-an-alternate-cname_{context}']
```
evgeni, line 1: is it intentional that the ID and the filename have "satellite" in them?

guides/common/modules/proc_configuring-hosts-to-use-an-alternate-cname-for-content-management.adoc
```diff
@@ -0,0 +1,24 @@
+[id='configuring-hosts-to-use-an-alternate-cname-for-content-management_{context}']
+= Configuring Hosts to Use an Alternate CNAME for Content Management
+
+After you configure {ProjectServer} with and alternate CNAME, you must point Red{nbsp}Hat Subscription Manager on {Project} hosts to use the alternate name for content management. You can do this manually or using a bootstrap script.
```
ekohl, line 4:
```suggestion
After you configure {ProjectServer} with an alternate CNAME, you must point Red{nbsp}Hat Subscription Manager on {Project} hosts to use the alternate name for content management. You can do this manually or using a bootstrap script.
```

guides/common/modules/proc_configuring-satellite-with-an-alternate-cname.adoc
```diff
@@ -0,0 +1,16 @@
...
+If you use a default {Project} certificate and want to configure {Project} with an alternate CNAME, enter the following command on {Project} to include a new CNAME into the {Project} certificate:
+
+[options="nowrap" subs="+quotes,attributes"]
+----
+{foreman-installer} --certs-cname _alternate_fqdn_
+----
+
+.For Custom Certificate Users
+
+If you use {Project} with a custom certificate, when creating a custom certificate, include the alternate CNAME records to the custom certificate. For more information, see {BaseURL}installing_satellite_server_from_a_connected_network/performing-additional-configuration#creating-a-custom-certificate_satellite[Creating a Custom SSL Certificate for Satellite Server].
```
ekohl, line 16: This still mentions Satellite

guides/common/modules/proc_configuring-satellite-with-an-alternate-cname.adoc
```diff
@@ -0,0 +1,16 @@
+[id='configuring-satellite-with-an-alternate-cname_{context}']
+= Configuring {Project} with an Alternate CNAME
+Use this procedure to configure {Project} with an alternate CNAME. Note that the procedures for users of a default {Project} certificate and custom certificate differ.
```
ekohl, line 3: Also note that this only works on Katello, not vanilla Foreman.

spetrosi: Noted this at the beginning of the procedure

guides/common/modules/proc_configuring-satellite-with-an-alternate-cname.adoc
spetrosi: Well, this is consistent, as this is done throughout the guides. For now, let\s stick to consistency, and if we want to change it in future we can do it everywhere in one run.

evgeni: Works for me, was mostly curious :)

lzap: So far we have removed the word from directory names. I suggest to make the switch rather earlier. This will ultimately break links in downstream tho.

By the way can't we simply create TWO ids (archors)? One with satellite for backward compatibility and one with foreman (or with something else like "project" or whatever.

guides/common/modules/proc_configuring-satellite-with-an-alternate-cname.adoc
```diff
@@ -0,0 +1,16 @@
...
+
+.For Default {Project} Certificate Users
+
+If you use a default {Project} certificate and want to configure {Project} with an alternate CNAME, enter the following command on {Project} to include a new CNAME into the {Project} certificate:
+
+[options="nowrap" subs="+quotes,attributes"]
+----
+{foreman-installer} --certs-cname _alternate_fqdn_
```
mirekdlugosz, line 11: Looking at comments in https://bugzilla.redhat.com/show_bug.cgi?id=1160344 , this only works during initial installation of Satellite - you can't change, or specify alternate, cname on already installed server.

But it seems we never mention that requirement explicitly. Moreover, placement of this section suggests this is something that can be freely done as post-install configuration. That tripped me up when I tried to follow these instructions.

evgeni: I think you *can* do this after the fact, but then need to pass one of the --regenerate-certs (can't remember the exact name) parameters, as the installer otherwise won't touch existing certs.

This comes at a cost of probably invalidating client configurations, tho.

mirekdlugosz: True, the option is `--certs-reset`. With that, I could get new certs and complete all instructions.

Do we expect people to just know that?

evgeni: Given the implications, I would neither expect people to know, nor document it to be used here.

The normal user workflow should be "do this during initial installation"

guides/common/modules/proc_configuring-satellite-with-an-alternate-cname.adoc
spetrosi: Two IDs will be complicated to maintain and create garbage in files. I will substitute `satellite` with `project` in the filename and in ID here.

guides/common/modules/proc_configuring-satellite-with-an-alternate-cname.adoc
spetrosi: We try to separate different scenarios into different procedures to not end up having a large procedure just on installing with many conditional notes in it. I have written a procedure for those who have already installed Satellite, and informed them that they need to configure it with an alternate CNAME before they register any host or Capsule to avoid complications with certs. In this procedure, I used the `--certs-reset` option. I also created another conditional bullet for those who have not installed Satellite yet to say that they can install it with an alternate cname right away,

guides/common/modules/proc_configuring-project-with-an-alternate-cname.adoc
```diff
@@ -0,0 +1,18 @@
+[id='configuring-project-with-an-alternate-cname_{context}']
+= Configuring {Project} with an Alternate CNAME
+Use this procedure to configure {Project} with an alternate CNAME. Note that the procedures for users of a default {Project} certificate and custom certificate differ. Note that if you have any hosts registered to {Projec}, you must update the `katello-ca-consumer` package on the hosts after regenerating the {Project} SSL certificate.
```
mirekdlugosz, line 3: Typo:
```suggestion
Use this procedure to configure {Project} with an alternate CNAME. Note that the procedures for users of a default {Project} certificate and custom certificate differ. Note that if you have any hosts registered to {Project}, you must update the `katello-ca-consumer` package on the hosts after regenerating the {Project} SSL certificate.
```

guides/common/modules/proc_configuring-hosts-to-use-an-alternate-cname-for-content-management.adoc
```diff
@@ -0,0 +1,32 @@
+[id='configuring-hosts-to-use-an-alternate-cname-for-content-management_{context}']
+= Configuring Hosts to Use an Alternate {Project} CNAME for Content Management
+
+If {Project} is configured with an alternate CNAME, you can configure hosts to use the alternate {Project} CNAME for content management. To do this, you must point hosts tot eh alternate CNAME prior to registering the hosts to {Project}. You can do this manually or using the bootstrap script.
```
mirekdlugosz, line 4: Typo:
```suggestion
If {Project} is configured with an alternate CNAME, you can configure hosts to use the alternate {Project} CNAME for content management. To do this, you must point hosts to the alternate CNAME prior to registering the hosts to {Project}. You can do this manually or using the bootstrap script.
```

guides/common/modules/proc_configuring-project-with-an-alternate-cname.adoc
```diff
@@ -0,0 +1,18 @@
...
+.For Default {Project} Certificate Users
+
+* If you have installed {Project} with a default {Project} certificate and want to configure {Project} with an alternate CNAME, enter the following command on {Project} to generate a new default {Project} SSL certificate with an additional CNAME.
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foreman-installer} --certs-cname _alternate_fqdn_ --certs-reset
+----
+
+* If you have not install {Project}, you can run the `satellite-maintain  --certs-cname _alternate_fqdn_` command  to install {Project} with an alternate CNAME.
```
mirekdlugosz, line 14: English grammar was never my string suite, but shouldn't that be "If you have not install**ed** ..."?

guides/common/modules/proc_configuring-satellite-with-an-alternate-cname.adoc
mirekdlugosz: I understand this might be seen as more complex, and there are many conditionals, but I do like this new version much better. We are clear that alternative CNAME is something you should think about before installing Satellite. If you already installed Satellite, we give working command - but warn that there will be additional work on clients and capsules. And we also give exact command to use when Satellite is installed for the first time.

Overall, great job!

guides/common/modules/proc_configuring-hosts-to-use-an-alternate-cname-for-content-management.adoc
mirekdlugosz, line 4: > You can do this manually or using the bootstrap script.

Since below we cover bootstrap first, would it be better to switch order in that sentence?
I mean:

> You can do this with bootstrap script or manually.

spetrosi: Good catch, I changed the order of methods because the bootstrap method is automated and seems like the one that users should go for ideally, however they can do things manually too if needed. Will change the order in this sentence too.

guides/common/modules/proc_configuring-project-with-an-alternate-cname.adoc
```diff
@@ -0,0 +1,18 @@
...
+.For Default {Project} Certificate Users
+
+* If you have installed {Project} with a default {Project} certificate and want to configure {Project} with an alternate CNAME, enter the following command on {Project} to generate a new default {Project} SSL certificate with an additional CNAME.
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foreman-installer} --certs-cname _alternate_fqdn_ --certs-reset
+----
+
+* If you have not installed {Project}, you can add the `--certs-cname _alternate_fqdn_` option to the `satellite-maintain` command to install {Project} with an alternate CNAME.
```
mirekdlugosz, line 14: Wait, `satellite-maintain`? Shouldn't that be `satellite-installer`?

Also, I guess it should be written as variable: `{foreman-installer}`.

Sorry, haven't noticed that yesterday

spetrosi: Fixed! Thank you for pointing this out, I do not know what I was thinking when writing it.

suyogsainkar:
@spetrosi looks good to me; thanks. merging; great job.

PR #250, opened 2020-08-26
Remove outdated statements that Capsule and Sat versions must match

spetrosi:
Bug 1871844 - Need to update a note from Sat6.8 installation guide since
we support N-1 capsules
https://bugzilla.redhat.com/show_bug.cgi?id=1871844


suyogsainkar:
@spetrosi lgtm; merging.

PR #251, opened 2020-08-27
Add info about multiple DHCP subnets

lzap:
This is constantly being asked, finally it is making its way into docs.

### DIFF
--- a/guides/doc-Provisioning_Guide/topics/Networking.adoc
+++ b/guides/doc-Provisioning_Guide/topics/Networking.adoc
@@ -211,6 +211,46 @@ To configure network services on {Project}'s integrated {SmartProxy}, complete t
 # hammer proxy info --name "_{foreman-example-com}_"
 ----
 
+[[Configuring_Networking-Multiple_Subnets_or_Domains_via_Installer]]
+==== Multiple Subnets or Domains via Installer
+
+The {foreman-installer} options allow only for a single DHCP subnet or DNS domain. One way to define more than one subnet is by using custom configuration file.
+
+For every additional subnet or domain, create an entry in `/etc/foreman-installer/custom-hiera.yaml` file:
+
+----
+dhcp::pools:
+ isolated.lan:
+   network: 192.168.99.0
+   mask: 255.255.255.0
+   gateway: 192.168.99.1
+   range: 192.168.99.5 192.168.99.49
+
+dns::zones:
+  # creates @ SOA $::fqdn root.example.com.
+  # creates $::fqdn A $::ipaddress
+  example.com: {}
+
+  # creates @ SOA test.example.net. hostmaster.example.com.
+  # creates test.example.net A 192.0.2.100
+  example.net:
+    soa: test.example.net
+    soaip: 192.0.2.100
+    contact: hostmaster.example.com.
+
+  # creates @ SOA $::fqdn root.example.org.
+  # does NOT create an A record
+  example.org:
+    reverse: true
+
+  # creates @ SOA $::fqdn hostmaster.example.com.
+  2.0.192.in-addr.arpa:
+    reverse: true
+    contact: hostmaster.example.com.
+----
+
+Execute {foreman-installer} to perform the changes and verify that the `/etc/dhcp/dhcpd.conf` contains appropriate entries. Subnets must be then defined in {Project} database.
+
 ==== DHCP, DNS, and TFTP Options for Network Configuration
 
 .DHCP Options


melcorr:
Is there anything here that we need to flag for downstream?

guides/doc-Provisioning_Guide/topics/Networking.adoc
```diff
@@ -211,6 +211,36 @@ To configure network services on {Project}'s integrated {SmartProxy}, complete t
 # hammer proxy info --name "_{foreman-example-com}_"
 ----
 
+[[Configuring_Networking-Multiple_DHCP_Subnets_via_Puppet]]
+==== Multiple DHCP Subnets via Puppet
+
+The installer have options available only for a single DHCP subnet. One way to define more than one subnet is via Puppet, the tool that the installer utilizes to perform the configuration.
```
melcorr, line 217:
```suggestion
The {foreman-installer} options allow only for a single DHCP subnet. One way to define more than one subnet is by using Puppet.
```

melcorr: I'm not sure we need to mention that Puppet is in the installer, since this will change and also since it has no real relevance to using Puppet to add more subnets.

guides/doc-Provisioning_Guide/topics/Networking.adoc
```diff
@@ -211,6 +211,36 @@ To configure network services on {Project}'s integrated {SmartProxy}, complete t
...
+==== Multiple DHCP Subnets Manually
+
+DHCP server configuration can be also managed manually, or with an external configuration tool. In that case, re-run the installer with a switch that disables management of the DHCP configuration completely while keeping the DHCP feature turned on.
+
+[options="nowrap" subs="+quotes,attributes"]
+----
+# foreman-installer --foreman-proxy-dhcp true --foreman-proxy-dhcp-managed false
+----
+
+From now on, DHCP configuration (e.g. `/etc/dhcp/dhcpd.conf`) is completely out of {foreman-installer} and Puppet control. {Project} will only connect to the DHCP server via OMAPI to perform DHCP record management.
```
ekohl, line 242: IMHO we shouldn't mention Puppet explicitly. That's just an implementation detail of the installer.

guides/doc-Provisioning_Guide/topics/Networking.adoc
```diff
@@ -211,6 +211,36 @@ To configure network services on {Project}'s integrated {SmartProxy}, complete t
...
+   network: 192.168.99.0
+   mask: 255.255.255.0
+   gateway: 192.168.99.1
+   range: 192.168.99.5 192.168.99.49
+----
+
+Execute {foreman-installer} to perform the changes and verify that the `/etc/dhcp/dhcpd.conf` contains appropriate entries. Subnets must be then defined in {Project} database.
+
+[[Configuring_Networking-Multiple_DHCP_Subnets_Manually]]
+==== Multiple DHCP Subnets Manually
```
ekohl, line 233: We should discourage this and mention that you are on your own to ensure things work. For example, users who upgrade won't get HTTPBoot support working out of the box. There are no instructions for users to maintain this.

guides/doc-Provisioning_Guide/topics/Networking.adoc
ekohl: Indeed. Puppet is an implementation detail of the installer.

ekohl:
Btw, you did add these instructions to the Katello manual: https://theforeman.org/plugins/katello/3.16/installation/index.html#multiple-subnets-and-domains

guides/doc-Provisioning_Guide/topics/Networking.adoc
lzap: I disagree. Then remove the option from the installer.

guides/doc-Provisioning_Guide/topics/Networking.adoc
lzap: I can remove the reference here.

guides/doc-Provisioning_Guide/topics/Networking.adoc
lzap: I removed this paragraph completely.

lzap:
Rebased.

guides/doc-Provisioning_Guide/topics/Networking.adoc
ekohl: What I'm saying is that the user must be able to understand the implication. It's a trade off: they get flexibility, but also no support in upgrades. If a fix or feature in the DHCP config is needed, they are responsible. HTTPBoot is one example.

lzap: My point is I am getting asked this quite often - I am not using Puppet, I have/want to deploy my own DNS server. And I would like to have a place where I can just share link to.

I can just hide this downstream and only show this upstream if that makes it any better.

ekohl: My point is that this option is valid, but users should understand the implication. Right now this looks like a thing without downsides.

Imagine a user did this with 6.7 and is happy. Then 6.8 comes around and HTTPBoot doesn't work. So support is contacted, it takes a while to figure out and we will end up with various suggestions on how to configure HTTPBoot. There will be frustration.

That's why the user must be informed about the trade off. If they manually manage their server, it's their responsibility and certain features may not work unless they configure it the right way.

lzap: I have deleted the whole block.

melcorr:
@lzap 

Do we need to wrap this for upstream or will Satellite users benefit from this also? Other than that this looks good to merge.

lzap:
This is valid for both. Failed test was caused by Microsoft site down, we can merge.

Issue #192, opened 2020-06-09
baseURL parameter not fulfilling upstream requirements - how to handle URLS


melcorr:
Hi, 

The http://docs.theforeman.org/guides/build/doc-Deploying_on_AWS/index-foreman.html#_use_cases_known_to_work section has links only to Satellite docs, when all these doc links are now upstream. 

Please add an equivalent.

lzap:
Yeah we need to get rid of absolute URLs, there is a macro for base url when needed.

melcorr:
Hi @spetrosi @lzap 

I see that the baseURL is a common attribute that is not defined dependent on upstream/downstream

Are we in a position now to set the upstream baseurls? 

@spetrosi there are only small guides that are left to migrate upstream, aren't there?

spetrosi:
I will test how guides work with BaseURL pointing to docs.theforeman

spetrosi:
My tests resulted in sad results:
Let's compare links for the same section up- and downstream:
- https://docs.theforeman.org/master/Administering_Red_Hat_Satellite/index-foreman.html#sect-Red_Hat_Satellite-Administering_Red_Hat_Satellite-Installing_the_Katello_Root_CA_Certificate
- https://access.redhat.com/documentation/en-us/red_hat_satellite/6.7/html/administering_red_hat_satellite/chap-red_hat_satellite-administering_red_hat_satellite-accessing_red_hat_satellite#sect-Red_Hat_Satellite-Administering_Red_Hat_Satellite-Installing_the_Katello_Root_CA_Certificate

The way the sections are "reffered" in the URLs are very different. If we were to use the BaseURL attribute, the links would look like that:
- {BaseURL}Administering_Red_Hat_Satellite-Installing_the_Katello_Root_CA_Certificate
- {BaseURL}/administering_red_hat_satellite/chap-red_hat_satellite-administering_red_hat_satellite-accessing_red_hat_satellite#sect-Red_Hat_Satellite-Administering_Red_Hat_Satellite-Installing_the_Katello_Root_CA_Certificate
I will try to investigate if we have something to do to make the paths in links identical.

melcorr:
Thanks @spetrosi 
I had similar findings yesterday...but i was hoping it was because I was doing something stupid because I forget things.

melcorr:
@lzap 
Any idea how we can get around this without writing ifevals for links?

spetrosi:
[...]

PR #116, opened 2020-02-25
Initial version of unmanaged IPv6

lzap:
This is very rough version of the "unmanaged IPv6 provisioning" setup I have tested. This is a feature for Foreman 2.1. The initial version is all written in a single procedure adoc file, it should probably be broken into more files. Let me know how do you prefer to approach this, I wanted to write it little bit in advance because my colelagues will be testing this stuff - it's not a priority at the moment.

There is one big open question in regard to grub2 - there are currently several bugs in RHEL version of grub2 and we are currently in the process of talking to maintainers how to proceed.

PR #254, opened 2020-09-02
Add content about installing project on IPv6

spetrosi:
Bug 1826824 - [RFE] IPv6 support

https://bugzilla.redhat.com/show_bug.cgi?id=1826824O

### DIFF
--- a/guides/common/assembly_preparing-environment-for-installation-in-ipv6-network.adoc
+++ b/guides/common/assembly_preparing-environment-for-installation-in-ipv6-network.adoc
@@ -1,9 +1,9 @@
 [id="preparing-environment-for-installation-in-ipv6-network"]
 = Preparing your Environment for {Project} Installation in an IPv6 Network
 
-To provision hosts in an IPv6 network, you must install {Project} in an IPv6 network and configure {Project} for the UEFI HTTP boot provisioning. Before installing {Project} in an IPv6 network, view the limitations and ensure that you meet the requirements.
+You can install and use {Project} in an IPv6 network. Before installing {Project} in an IPv6 network, view the limitations and ensure that you meet the requirements.
 
-After installing {Project}, you will need to configure {Project} for provisioning. For more information, see xref:configuring-for-uefi-http-boot-provisioning-in-an-ipv6-network_{context}[].
+To provision hosts in an IPv6 network, after installing {Project}, you must also configure {Project} for the UEFI HTTP boot provisioning. For more information, see xref:configuring-for-uefi-http-boot-provisioning-in-an-ipv6-network_{context}[].
 
 include::modules/con_limitations-of-installation-in-an-ipv6-network.adoc[leveloffset=+1]
 
--- a/guides/common/assembly_preparing-environment-for-installation-in-ipv6-network.adoc
+++ b/guides/common/assembly_preparing-environment-for-installation-in-ipv6-network.adoc
@@ -1,4 +1,4 @@
-[id="preparing-environment-for-{project}-installation-ipv6"]
+[id="preparing-environment-for-installation-in-ipv6-network"]
 = Preparing your Environment for {Project} Installation in an IPv6 Network
 
 To provision hosts in an IPv6 network, you must install {Project} in an IPv6 network and configure {Project} for the UEFI HTTP boot provisioning. Before installing {Project} in an IPv6 network, view the limitations and ensure that you meet the requirements.
--- a/guides/common/assembly_preparing-environment-for-installation-in-ipv6-network.adoc
+++ b/guides/common/assembly_preparing-environment-for-installation-in-ipv6-network.adoc
@@ -3,6 +3,8 @@
 
 To provision hosts in an IPv6 network, you must install {Project} in an IPv6 network and configure {Project} for the UEFI HTTP boot provisioning. Before installing {Project} in an IPv6 network, view the limitations and ensure that you meet the requirements.
 
+After installing {Project}, you will need to configure {Project} for provisioning. For more information, see xref:configuring-for-uefi-http-boot-provisioning-in-an-ipv6-network_{context}[].
+
 include::modules/con_limitations-of-installation-in-an-ipv6-network.adoc[leveloffset=+1]
 
 include::modules/con_requirements-for-installation-in-an-ipv6-network.adoc[leveloffset=+1]
--- a/guides/common/assembly_preparing-environment-for-satellite-installation-ipv6.adoc
+++ b/guides/common/assembly_preparing-environment-for-satellite-installation-ipv6.adoc
@@ -1,7 +1,7 @@
 [id="preparing-environment-for-{project}-installation-ipv6"]
 = Preparing your Environment for {Project} Installation in an IPv6 Network
 
-If you must provision hosts in IPv6 network, you can install {Project} in a IPv6 network and provision using UEFI HTTP Boot. Before installing {Project} in an IPv6 network, view the limitations that such scenario has and ensure that you meet the requirements.
+To provision hosts in an IPv6 network, you must install {Project} in an IPv6 network and configure {Project} for the UEFI HTTP boot provisioning. Before installing {Project} in an IPv6 network, view the limitations and ensure that you meet the requirements.
 
 include::modules/con_limitations-of-installation-in-an-ipv6-network.adoc[leveloffset=+1]
 
--- a/guides/common/assembly_preparing-environment-for-satellite-installation-ipv6.adoc
+++ b/guides/common/assembly_preparing-environment-for-satellite-installation-ipv6.adoc
@@ -1,4 +1,4 @@
-[id="preparing-environment-for-satellite-installation-ipv6"]
+[id="preparing-environment-for-{project}-installation-ipv6"]
 = Preparing your Environment for {Project} Installation in an IPv6 Network
 
 If you must provision hosts in IPv6 network, you can install {Project} in a IPv6 network and provision using UEFI HTTP Boot. Before installing {Project} in an IPv6 network, view the limitations that such scenario has and ensure that you meet the requirements.
--- a/guides/common/assembly_preparing-environment-for-satellite-installation-ipv6.adoc
+++ b/guides/common/assembly_preparing-environment-for-satellite-installation-ipv6.adoc
@@ -1,0 +1,9 @@
+[id="preparing-environment-for-satellite-installation-ipv6"]
+= Preparing your Environment for {Project} Installation in an IPv6 Network
+
+If you must provision hosts in IPv6 network, you can install {Project} in a IPv6 network and provision using UEFI HTTP Boot. Before installing {Project} in an IPv6 network, view the limitations that such scenario has and ensure that you meet the requirements.
+
+include::modules/con_limitations-of-installation-in-an-ipv6-network.adoc[leveloffset=+1]
+
+include::modules/con_requirements-for-installation-in-an-ipv6-network.adoc[leveloffset=+1]
+
--- a/guides/common/modules/con_limitations-of-installation-in-an-ipv6-network.adoc
+++ b/guides/common/modules/con_limitations-of-installation-in-an-ipv6-network.adoc
@@ -3,5 +3,5 @@
 
 {Project} installation in an IPv6 network has the following limitations:
 
-* DHCP server in {RHEL} (ISC DHCP) does not provide an integration API for managing IPv6 records, therefore {SmartProxy} DHCP plug-in that provides DHCP management of IPv6 subnets does not exist. In this scenario, DHCP IPv6 must be deployed as a separate unmanaged service to boostrap client nodes into grub2, which then configures IPv6 networking either using DHCPv6 or via static IPv6 address.
+* DHCP server in {RHEL} (ISC DHCP) does not provide an integration API for managing IPv6 records, therefore {SmartProxy} DHCP plug-in that provides DHCP management is limited to IPv4 subnets. In this scenario, DHCP IPv6 must be deployed as a separate unmanaged service to bootstrap client nodes into grub2, which then configures IPv6 networking either using DHCPv6 or via static IPv6 address.
 * Although provisioning templates which ship with {Project} include IPv6 support for PXE and HTTP (iPXE) scenarios, the only tested and certified workflow is UEFI HTTP Boot.
--- a/guides/common/modules/con_limitations-of-installation-in-an-ipv6-network.adoc
+++ b/guides/common/modules/con_limitations-of-installation-in-an-ipv6-network.adoc
@@ -4,4 +4,4 @@
 {Project} installation in an IPv6 network has the following limitations:
 
 * DHCP server in {RHEL} (ISC DHCP) does not provide an integration API for managing IPv6 records, therefore {SmartProxy} DHCP plug-in that provides DHCP management is limited to IPv4 subnets. In this scenario, DHCP IPv6 must be deployed as a separate unmanaged service to bootstrap client nodes into grub2, which then configures IPv6 networking either using DHCPv6 or via static IPv6 address.
-* Although provisioning templates which ship with {Project} include IPv6 support for PXE and HTTP (iPXE) scenarios, the only tested and certified workflow is UEFI HTTP Boot.
+* Although {Project} provisioning templates include IPv6 support for PXE and HTTP (iPXE) provisioning, the only tested and certified provisioning workflow is the UEFI HTTP Boot provisioning. This limitation only relates to users who plan to use {Project} to provision hosts.
--- a/guides/common/modules/con_limitations-of-installation-in-an-ipv6-network.adoc
+++ b/guides/common/modules/con_limitations-of-installation-in-an-ipv6-network.adoc
@@ -3,6 +3,6 @@
 
 {Project} installation in an IPv6 network has the following limitations:
 
-* You can install {Project} and {SmartProxies} in IPv6-only systems, dual-stack installation not supported.
+* You can install {Project} and {SmartProxies} in IPv6-only systems, dual-stack installation is not supported.
 
 * Although {Project} provisioning templates include IPv6 support for PXE and HTTP (iPXE) provisioning, the only tested and certified provisioning workflow is the UEFI HTTP Boot provisioning. This limitation only relates to users who plan to use {Project} to provision hosts.
--- a/guides/common/modules/con_limitations-of-installation-in-an-ipv6-network.adoc
+++ b/guides/common/modules/con_limitations-of-installation-in-an-ipv6-network.adoc
@@ -1,0 +1,7 @@
+[id="limitations-of-installation-in-an-ipv6-network_{context}"]
+= Limitations of {Project} Installation in an IPv6 Network 
+
+{Project} installation in an IPv6 network has the following limitations:
+
+* DHCP server in {RHEL} (ISC DHCP) does not provide an integration API for managing IPv6 records, therefore {SmartProxy} DHCP plug-in that provides DHCP management of IPv6 subnets does not exist. In this scenario, DHCP IPv6 must be deployed as a separate unmanaged service to boostrap client nodes into grub2, which then configures IPv6 networking either using DHCPv6 or via static IPv6 address.
+* Although provisioning templates which ship with {Project} include IPv6 support for PXE and HTTP (iPXE) scenarios, the only tested and certified workflow is UEFI HTTP Boot.
--- a/guides/common/modules/con_limitations-of-installation-in-an-ipv6-network.adoc
+++ b/guides/common/modules/con_limitations-of-installation-in-an-ipv6-network.adoc
@@ -3,5 +3,6 @@
 
 {Project} installation in an IPv6 network has the following limitations:
 
-* DHCP server in {RHEL} (ISC DHCP) does not provide an integration API for managing IPv6 records, therefore {SmartProxy} DHCP plug-in that provides DHCP management is limited to IPv4 subnets. In this scenario, DHCP IPv6 must be deployed as a separate unmanaged service to bootstrap client nodes into grub2, which then configures IPv6 networking either using DHCPv6 or via static IPv6 address.
+* You can install {Project} and {SmartProxies} in IPv6-only systems, dual-stack installation not supported.
+
 * Although {Project} provisioning templates include IPv6 support for PXE and HTTP (iPXE) provisioning, the only tested and certified provisioning workflow is the UEFI HTTP Boot provisioning. This limitation only relates to users who plan to use {Project} to provision hosts.
--- a/guides/common/modules/con_requirements-for-installation-in-an-ipv6-network.adoc
+++ b/guides/common/modules/con_requirements-for-installation-in-an-ipv6-network.adoc
@@ -8,12 +8,12 @@ ifeval::["{build}" == "satellite"]
 endif::[]
 
 ifeval::["{build}" != "satellite"]
-* {SmartProxy} must be installed on a system with grub2 version 2.05 or higher or system with fixes for HTTP Boot, for example CentOS version 7.9 or 8.3 or higher. For other operating systems, copy newest grub build to `/var/lib/tftpboot/grub2/grubx64.efi`.
+* {SmartProxy} must be installed on a system with grub2 version 2.05 or higher or system with fixes for HTTP Boot, for example CentOS version 7.9 or 8.3 or higher. For other operating systems, copy the newest grub build to `/var/lib/tftpboot/grub2/grubx64.efi`.
 endif::[]
 
-* {Project} or {SmartProxy} can be only installed on IPv6-only systems, dual-stack is known to work but not supported.
+* You can install {Project} and {SmartProxy} in IPv6-only systems, dual-stack installation not supported.
 
-* It is assumed, native IPv6 internet connectivity is availble. If not, it's recommended to configure a HTTP proxy in order to download RPM packages:
+* It is assumed, native IPv6 internet connectivity is available. If not, it's recommended to configure a HTTP proxy in order to download RPM packages:
 +
 ifeval::["{build}" == "satellite"]
 [options="nowrap" subs="+quotes,attributes"]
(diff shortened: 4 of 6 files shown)


guides/common/assembly_preparing-environment-for-satellite-installation-ipv6.adoc
```diff
@@ -0,0 +1,9 @@
+[id="preparing-environment-for-satellite-installation-ipv6"]
+= Preparing your Environment for {Project} Installation in an IPv6 Network
+
+If you must provision hosts in IPv6 network, you can install {Project} in a IPv6 network and provision using UEFI HTTP Boot. Before installing {Project} in an IPv6 network, view the limitations that such scenario has and ensure that you meet the requirements.
```
ekohl, line 4: I'd use an impersonal form.
```suggestion
To provision hosts in IPv6 networks, {Project} should be installed in an IPv6 network and UEFI HTTP Boot configured. Before installing {Project} in an IPv6 network, view the limitations on the scenario and ensure that the requirements are met.
```

guides/common/modules/con_limitations-of-installation-in-an-ipv6-network.adoc
```diff
@@ -0,0 +1,7 @@
+[id="limitations-of-installation-in-an-ipv6-network_{context}"]
+= Limitations of {Project} Installation in an IPv6 Network 
+
+{Project} installation in an IPv6 network has the following limitations:
+
+* DHCP server in {RHEL} (ISC DHCP) does not provide an integration API for managing IPv6 records, therefore {SmartProxy} DHCP plug-in that provides DHCP management of IPv6 subnets does not exist. In this scenario, DHCP IPv6 must be deployed as a separate unmanaged service to boostrap client nodes into grub2, which then configures IPv6 networking either using DHCPv6 or via static IPv6 address.
```
ekohl, line 6:
```suggestion
* DHCP server in {RHEL} (ISC DHCP) does not provide an integration API for managing IPv6 records, therefore {SmartProxy}'s DHCP plug-in that provides DHCP management is limited to IPv4. In this scenario, DHCP IPv6 must be deployed as a separate unmanaged service to bootstrap client nodes into grub2, which then configures IPv6 networking either using DHCPv6 or via static IPv6 address.
```

guides/common/modules/con_requirements-for-installation-in-an-ipv6-network.adoc
```diff
@@ -0,0 +1,33 @@
...
+= Requirements for {Project} Installation in an IPv6 Network 
+
+Before installing {Project} in an IPv6 network, ensure that you meet the following requirements:
+
+ifeval::["{build}" == "satellite"]
+* {SmartProxy} must be installed on {RHEL} version 7.9 or higher.
+endif::[]
+
+ifeval::["{build}" != "satellite"]
+* {SmartProxy} must be installed on a system with grub2 version 2.05 or higher or system with fixes for HTTP Boot, for example CentOS version 7.9 or 8.3 or higher. For other operating systems, copy newest grub build to `/var/lib/tftpboot/grub2/grubx64.efi`.
```
ekohl, line 11:
```suggestion
* {SmartProxy} must be installed on a system with grub2 version 2.05 or higher or system with fixes for HTTP Boot, for example CentOS version 7.9 or 8.3 or higher. For other operating systems, copy the newest grub build to `/var/lib/tftpboot/grub2/grubx64.efi`.
```

guides/common/modules/con_requirements-for-installation-in-an-ipv6-network.adoc
```diff
@@ -0,0 +1,33 @@
...
+endif::[]
+
+* {Project} or {SmartProxy} can be only installed on IPv6-only systems, dual-stack is known to work but not supported.
+
+* It is assumed, native IPv6 internet connectivity is availble. If not, it's recommended to configure a HTTP proxy in order to download RPM packages:
```
ekohl, line 16:
```suggestion
* It is assumed, native IPv6 internet connectivity is available. If not, it's recommended to configure a HTTP proxy in order to download RPM packages:
```

guides/common/modules/con_requirements-for-installation-in-an-ipv6-network.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+ifeval::["{build}" == "satellite"]
+* {SmartProxy} must be installed on {RHEL} version 7.9 or higher.
+endif::[]
+
+ifeval::["{build}" != "satellite"]
+* {SmartProxy} must be installed on a system with grub2 version 2.05 or higher or system with fixes for HTTP Boot, for example CentOS version 7.9 or 8.3 or higher. For other operating systems, copy newest grub build to `/var/lib/tftpboot/grub2/grubx64.efi`.
+endif::[]
+
+* {Project} or {SmartProxy} can be only installed on IPv6-only systems, dual-stack is known to work but not supported.
```
ekohl, line 14: I'd keep this downstream only. Upstream we don't really have this distinction. If it's know to work, users can use it.

guides/common/assembly_preparing-environment-for-satellite-installation-ipv6.adoc
spetrosi: I applied your technical corrections here and fixed the style a bit. The problem with an impersonal form, is that it is harder to translate and to understand. For example, with passive voice you do not see who performs the action. So as per the IBM style guide, we use active voice as often as possible.

guides/common/modules/con_requirements-for-installation-in-an-ipv6-network.adoc
spetrosi: Done! Thank you.

guides/common/modules/proc_configuring-for-uefi-http-boot-ipv6-provisioning.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+* Ensure that the UDP ports 67 and 68 are accessible by the client so that the client can send and receive a DHCP request and offer.
+
+* Ensure that the TCP port 8000 is open for the client to download files and Kickstart templates from {SmartProxy}.
+
+* Ensure that the host provisioning interface subnet has a TFTP {SmartProxy}, and Templates {SmartProxy} set. For more information, see https://access.redhat.com/documentation/en-us/red_hat_satellite/6.7/html/provisioning_guide/configuring_networking#Configuring_Networking-Adding_a_Subnet_to_the_Satellite_Server[Adding a Subnet to {ProjectServer}] in the _Provisioning Guide_.
+
+.Procedure
+
+. Since clients that are booting from network cannot be identified by remote IPv6 address due to unmanaged DHCPv6 service, you must enable provisioning tokens. To enable provisioning tokens, navigate to *Administer* > *Settings* > *Provisioning* and set the *Token duration* setting to 0.
```
ekohl, line 18: I thought that setting the token duration to 0 disables them rather than enable. Note that this makes it insecure.

guides/common/modules/proc_configuring-for-uefi-http-boot-ipv6-provisioning.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+Use this procedure to configure {Project} to provision hosts in an IPv6 network with UEFI HTTP Boot provisioning.
+
+.Prerequisites
+
+* Ensure that your clients can access DHCP and TFTP servers.
+
+* Ensure that the UDP ports 67 and 68 are accessible by the client so that the client can send and receive a DHCP request and offer.
+
+* Ensure that the TCP port 8000 is open for the client to download files and Kickstart templates from {SmartProxy}.
```
ekohl, line 12:
```suggestion
* Ensure that the TCP port 8000 is open for clients to download files and Kickstart templates from {SmartProxy}.
```

guides/common/modules/proc_configuring-for-uefi-http-boot-ipv6-provisioning.adoc
```diff
@@ -0,0 +1,39 @@
+[id="configuring-for-uefi-http-boot-ipv6-provisioning_{context}"]
+= Configuring {Project} for UEFI HTTP Boot IPv6 Provisioning
+
+Use this procedure to configure {Project} to provision hosts in an IPv6 network with UEFI HTTP Boot provisioning.
+
+.Prerequisites
+
+* Ensure that your clients can access DHCP and TFTP servers.
+
+* Ensure that the UDP ports 67 and 68 are accessible by the client so that the client can send and receive a DHCP request and offer.
```
ekohl, line 10: To quote [RFC2131 section 4.1](https://tools.ietf.org/html/rfc2131#section-4.1):

>   DHCP uses UDP as its transport protocol.  DHCP messages from a client
>   to a server are sent to the 'DHCP server' port (67), and DHCP
>   messages from a server to a client are sent to the 'DHCP client' port
>   (68).

That means UDP port 67 must be open on the DHCP server in the `INPUT` chain and UDP port 68 in the `OUTPUT` chain. The current phrasing suggests you should open port 68 on the server to incoming traffic.

```suggestion
* Ensure that the UDP ports 67 and 68 are accessible by clients so clients can send DHCP requests and receive DHCP offers.
```

guides/common/modules/proc_configuring-for-uefi-http-boot-ipv6-provisioning.adoc
```diff
@@ -0,0 +1,39 @@
...
+. Optional: If the host and the DHCP server are separated by a router, configure the DHCP relay agent and point to the DHCP server.
+
+. Update the `grub2-efi` package to the latest version:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {package-install-project} grub2-efi
+----
+
+. Optional: If native IPv6 connectivity is not available, configure an HTTP proxy in {Project} and set it as the default proxy in settings. For more information, see https://access.redhat.com/documentation/en-us/red_hat_satellite/6.8-beta/html/installing_satellite_server_from_a_connected_network/performing-additional-configuration#adding-a-default-http-proxy_satellite[Adding a Default HTTP Proxy to Satellite].
```
ekohl, line 33: Is there a way to link to this content internally, without linking to the RH docs?

guides/common/assembly_preparing-environment-for-satellite-installation-ipv6.adoc
```diff
@@ -0,0 +1,9 @@
+[id="preparing-environment-for-{project}-installation-ipv6"]
```
ekohl, line 1: Do you really need to mention the project in the ID? Isn't it better to avoid it when possible?

guides/common/modules/proc_configuring-for-uefi-http-boot-ipv6-provisioning.adoc
spetrosi: This is being discussed in the issue #192. In this case, this module is likely to be reused, and if it is reused in another guide, xref won't work. Hence the link.

guides/common/assembly_preparing-environment-for-satellite-installation-ipv6.adoc
spetrosi: That's a good point, it is probably better to avoid using project in the ID. However, in some cases there will be similar modules for Foreman and Smart Proxy. In this cases we need to specify this in the ID anyway. For example, in `proc_installing-the-satellite-server-packages.adoc` and `proc_installing-capsule-server-packages.adoc`.

ekohl: Agreed, but I'd suggest to leave it out when you can. It can even avoid confusion. Here the user may be looking for a document to prepare the capsule but as I read this, it's more of an integral env preparation.

guides/common/modules/proc_configuring-for-uefi-http-boot-ipv6-provisioning.adoc
spetrosi: Is there a way to provision securely? If not, let' be fair and provide more info about what exactly is insecure.

spetrosi:
The PR is based on info from #116

guides/common/modules/proc_configuring-for-uefi-http-boot-ipv6-provisioning.adoc
ekohl: Without provisioning tokens there is no authentication on provisioning templates, other than checking the IP address. These templates can include private data like passwords or other secrets. The attack would need to be able to spoof an IP on the target network so it's not trivial but it certainly lowers security.

guides/common/assembly_preparing-environment-for-installation-in-ipv6-network.adoc
```diff
@@ -0,0 +1,11 @@
+[id="preparing-environment-for-installation-in-ipv6-network"]
+= Preparing your Environment for {Project} Installation in an IPv6 Network
+
+To provision hosts in an IPv6 network, you must install {Project} in an IPv6 network and configure {Project} for the UEFI HTTP boot provisioning. Before installing {Project} in an IPv6 network, view the limitations and ensure that you meet the requirements.
```
lzap, line 4: I'd probably break up these two requirements because the first one is required always, however UEFI HTTP boot provisioning is only required when you plan to do network boot provisioning.

guides/common/modules/con_limitations-of-installation-in-an-ipv6-network.adoc
```diff
@@ -0,0 +1,7 @@
+[id="limitations-of-installation-in-an-ipv6-network_{context}"]
+= Limitations of {Project} Installation in an IPv6 Network 
+
+{Project} installation in an IPv6 network has the following limitations:
+
+* DHCP server in {RHEL} (ISC DHCP) does not provide an integration API for managing IPv6 records, therefore {SmartProxy} DHCP plug-in that provides DHCP management is limited to IPv4 subnets. In this scenario, DHCP IPv6 must be deployed as a separate unmanaged service to bootstrap client nodes into grub2, which then configures IPv6 networking either using DHCPv6 or via static IPv6 address.
+* Although provisioning templates which ship with {Project} include IPv6 support for PXE and HTTP (iPXE) scenarios, the only tested and certified workflow is UEFI HTTP Boot.
```
lzap, line 7: This is limitation for provisioning, not for generic use on IPv6. You can still install and use project on IPv6 without provisioning just fine - e.g. syncing content, calling remote execution jobs etc.

guides/common/modules/con_requirements-for-installation-in-an-ipv6-network.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+Before installing {Project} in an IPv6 network, ensure that you meet the following requirements:
+
+ifeval::["{build}" == "satellite"]
+* {SmartProxy} must be installed on {RHEL} version 7.9 or higher.
+endif::[]
+
+ifeval::["{build}" != "satellite"]
+* {SmartProxy} must be installed on a system with grub2 version 2.05 or higher or system with fixes for HTTP Boot, for example CentOS version 7.9 or 8.3 or higher. For other operating systems, copy the newest grub build to `/var/lib/tftpboot/grub2/grubx64.efi`.
+endif::[]
```
lzap, line 12: These two notes are only valid if you plan provisioning.

guides/common/modules/proc_configuring-for-uefi-http-boot-ipv6-provisioning.adoc
```diff
@@ -0,0 +1,39 @@
+[id="configuring-for-uefi-http-boot-provisioning-in-an-ipv6-network_{context}"]
+= Configuring {Project} for UEFI HTTP Boot Provisioning in an IPv6 Network
+
+Use this procedure to configure {Project} to provision hosts in an IPv6 network with UEFI HTTP Boot provisioning.
+
+.Prerequisites
+
+* Ensure that your clients can access DHCP and TFTP servers.
```
lzap, line 8: Replace TFTP with HTTP. TFTP is not used at all in this scenario.

guides/common/modules/proc_configuring-for-uefi-http-boot-ipv6-provisioning.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+* Ensure that the UDP ports 67 and 68 are accessible by clients so clients can send DHCP requests and receive DHCP offers.
+
+* Ensure that the TCP port 8000 is open for clients to download files and Kickstart templates from {SmartProxy}.
+
+* Ensure that the host provisioning interface subnet has a TFTP {SmartProxy}, and Templates {SmartProxy} set. For more information, see https://access.redhat.com/documentation/en-us/red_hat_satellite/6.7/html/provisioning_guide/configuring_networking#Configuring_Networking-Adding_a_Subnet_to_the_Satellite_Server[Adding a Subnet to {ProjectServer}] in the _Provisioning Guide_.
```
lzap, line 14: Replace TFTP with HTTP Boot.

guides/common/modules/proc_configuring-for-uefi-http-boot-ipv6-provisioning.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+.Procedure
+
+. Since clients that are booting from network cannot be identified by a remote IPv6 address due to unmanaged DHCPv6 service, you must disable provisioning tokens. To disable provisioning tokens, navigate to *Administer* > *Settings* > *Provisioning* and set the *Token duration* setting to 0.
```
lzap, line 18: This is a typo made by me - it is actually the very opposite. So ask readers to "ensure the token duration is set to some reasonable time duration and not zero".

guides/common/modules/proc_configuring-for-uefi-http-boot-ipv6-provisioning.adoc
lzap: Let's continue the discussion there, we have many of these and this needs to be solved somehow.

lzap (changes requested):
Some comments.

guides/common/modules/proc_configuring-for-uefi-http-boot-ipv6-provisioning.adoc
spetrosi: Oh that is good, @ekohl pointed out that disabling tokens is insecure.
Please elaborate why prov tokens must be enabled here. I do not see the direct correlation between the sentences "{SmartProxy} cannot identify clients that are booting from the network by a remote IPv6 address because of unmanaged DHCPv6 service" and "you must enable provisioning tokens". How are tokens related to Smart Proxy identifying clients?

ekohl: > Oh that is good, @ekohl pointed out that disabling tokens is insecure.

Discussion @ https://github.com/theforeman/foreman-documentation/pull/254#discussion_r482046568.

> This is a typo made by me - it is actually the very opposite. So ask readers to "ensure the token duration is set to some reasonable time duration and not zero".

Glad to hear that tokens should be used.

The default for `token_duration` is is 6 * 60 minutes, so 6 hours. It's probably good to recommend users who are unsure about the exact value to choose the default, right @lzap?

guides/common/modules/con_requirements-for-installation-in-an-ipv6-network.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+* You can install {Project} and {SmartProxy} in IPv6-only systems, dual-stack installation not supported.
+
+* It is assumed, native IPv6 internet connectivity is available. If not, it's recommended to configure a HTTP proxy in order to download RPM packages:
```
spetrosi, line 16: Is not this the same thing as in this step in guides/common/modules/proc_configuring-for-uefi-http-boot-ipv6-provisioning.adoc within this PR?

> Optional: If native IPv6 connectivity is not available, configure an HTTP proxy in {Project} and set it as the default proxy in settings. For more information, see https://access.redhat.com/documentation/en-us/red_hat_satellite/6.8-beta/html/installing_satellite_server_from_a_connected_network/performing-additional-configuration#adding-a-default-http-proxy_satellite[Adding a Default HTTP Proxy to Satellite].

guides/common/modules/proc_configuring-for-uefi-http-boot-ipv6-provisioning.adoc
lzap: Your concern about security is valid but what I am saying is that we actually want users to ensure tokens are in use. Therefore it's more secure. In the original text I require the opposite (unsecure). So there's is nothing to worry about.

guides/common/assembly_preparing-environment-for-satellite-installation-ipv6.adoc
lzap: Another discussion we need to have. The obvious solution is to remove these from all IDs however this will likely break links in existing Satellite docs. Maybe we can add two IDs to those places with a comment that the old one is legacy for keeping links alive.

guides/common/modules/proc_configuring-for-uefi-http-boot-ipv6-provisioning.adoc
lzap: Yeah we actually do not require any action here to be honest. I wonder if we should delete this line from the procedure. We can perhaps move this to Requirements only in the form of "Ensure tokens are not disabled...".

I am just doublechecking, I wonder why I added it initially. For IPv6 DHCP is unmanaged, therefore system boots up, picks (random) IPv6, gets TFTP config which contains the token, yeah. This will work, no reason to disable them.

guides/common/modules/con_requirements-for-installation-in-an-ipv6-network.adoc
lzap: Yes it is just a generic info about how we recommend to setup proxy when native IPv6 is not there. Feel free to make this a module.

spetrosi:
@lzap I have applied all feedback that you left, can you please reveiw again?

guides/common/modules/con_limitations-of-installation-in-an-ipv6-network.adoc
```diff
@@ -0,0 +1,7 @@
+[id="limitations-of-installation-in-an-ipv6-network_{context}"]
+= Limitations of {Project} Installation in an IPv6 Network 
+
+{Project} installation in an IPv6 network has the following limitations:
+
+* DHCP server in {RHEL} (ISC DHCP) does not provide an integration API for managing IPv6 records, therefore {SmartProxy} DHCP plug-in that provides DHCP management is limited to IPv4 subnets. In this scenario, DHCP IPv6 must be deployed as a separate unmanaged service to bootstrap client nodes into grub2, which then configures IPv6 networking either using DHCPv6 or via static IPv6 address.
```
lzap, line 6: that provides management of ISC DHCP

is probably better because we also ship Infoblox DHCP and that one could work (we haven't tested actually).

lzap (approved):
One nitpick but RTM

guides/common/modules/con_limitations-of-installation-in-an-ipv6-network.adoc
spetrosi: Where shall I put your suggestion @lzap ?

guides/common/modules/con_limitations-of-installation-in-an-ipv6-network.adoc
```diff
@@ -3,5 +3,6 @@
 
 {Project} installation in an IPv6 network has the following limitations:
 
-* DHCP server in {RHEL} (ISC DHCP) does not provide an integration API for managing IPv6 records, therefore {SmartProxy} DHCP plug-in that provides DHCP management is limited to IPv4 subnets. In this scenario, DHCP IPv6 must be deployed as a separate unmanaged service to bootstrap client nodes into grub2, which then configures IPv6 networking either using DHCPv6 or via static IPv6 address.
+* You can install {Project} and {SmartProxies} in IPv6-only systems, dual-stack installation not supported.
```
lzap, line 6: I'd prefer "untested, known to work" or similar. But I am fine.

guides/common/modules/con_requirements-for-installation-in-an-ipv6-network.adoc
```diff
@@ -11,4 +11,8 @@ ifeval::["{build}" != "satellite"]
...
 endif::[]
 
-* You can install {Project} and {SmartProxies} in IPv6-only systems, dual-stack installation not supported.
+* You must deploy an external DHCP IPv6 server as a separate unmanaged service to bootstrap clients into GRUB2, which then configures IPv6 networking either using DHCPv6 or or assigning static IPv6 address. This is required because the DHCP server in {RHEL} (ISC DHCP) does not provide an integration API for managing IPv6 records, therefore the {SmartProxy} DHCP plug-in that provides DHCP management is limited to IPv4 subnets.
+
+* You must deploy an external IPv4 HTTP proxy server. This is required because {Project} distributes content only over IPv4 networks, therefore you must use an IPv4 proxy to redirect that content to hosts in your IPv6 network.
+
+* You must configure {Project} to use this IPv4 HTTP proxy server as the default proxy. For more information, see https://access.redhat.com/documentation/en-us/red_hat_satellite/6.8-beta/html/installing_satellite_server_from_a_connected_network/performing-additional-configuration#adding-a-default-http-proxy_satellite[Adding a Default HTTP Proxy to Satellite].
```
lzap, line 18: Add a sentence that shows up when "foreman" build is done: This is only relevant for Katello content.

spetrosi: Done

lzap (approved):

guides/common/modules/con_limitations-of-installation-in-an-ipv6-network.adoc
suyogsainkar: "dual-stack installation _is_ not supported."

suyogsainkar:
@spetrosi I've added a minor comment, everything else looks good to me. By the way, this PR is still marked as draft so it's currently unmergable.

guides/common/modules/con_limitations-of-installation-in-an-ipv6-network.adoc
spetrosi: Thanks Suyog, fixed this, please merge

PR #259, opened 2020-09-10
Add script that runs satellite-sync for each downstream repository

spetrosi:
Bug 1877670 - Add script upstream that runs satellite-sync for each downstream repository

https://bugzilla.redhat.com/show_bug.cgi?id=1877670


lzap:
I noticed that the script which is being called contains hostname(s) of internal system(s). I suggest to extract this to an argument or ENV variable so it's not exposed to the public.

spetrosi:
> I noticed that the script which is being called contains hostname(s) of internal system(s). I suggest to extract this to an argument or ENV variable so it's not exposed to the public.

I do not know how to do this. Can you please show me an example with one instance of sensitive info so that I can learn and apply it everywhere?

suyogsainkar:
@spetrosi the PR looks good to me if it's been tested to run as expected. I'll approve it from my end instead of merging just yet, since you're awaiting lzap's input before addressing his latest comment. thanks.

suyogsainkar (approved):

guides/scripts/satellite-sync.sh
```diff
@@ -86,7 +86,8 @@ then
   cd ..
 else
   echo "Local copy of downstream repository not found on server - cloning..."
-  git clone git@gitlab.cee.redhat.com:satellite-6-documentation/docs-Red_Hat_Satellite_6.git
+  # GitLab repository is hidden for security reasons. You must define the $SATELLITE_DOCS_REPOSITORY before running the script by entering the `export GITLABHOSTNAME=<link-to-repository>` command. To find the link to repository, navigate to the GitLab repository, click *Clone*, then click *Copy URL* to the right of the *Clone with SSH* URL.
+  git clone $SATELLITE_DOCS_REPOSITORY
```
spetrosi, line 90: @lzap Here I've hidden the hostname. Does this seem correct?

lzap: Excellent.

lzap (approved):
Haven't tested tho! LGTM

PR #260, opened 2020-09-11
Add a step that installs rh-postgresql12-postgresql-evr on PostgreSQL…

spetrosi:
… external DB server

Bug 1860473 - external postgres 12 not working on satellite 6.8

https://bugzilla.redhat.com/show_bug.cgi?id=1860473

### DIFF
--- a/guides/common/modules/proc_installing-postgresql.adoc
+++ b/guides/common/modules/proc_installing-postgresql.adoc
@@ -10,15 +10,11 @@ include::snip_firewalld.adoc[]
 . To install PostgreSQL, enter the following command:
 +
 ----
-# yum install rh-postgresql12-postgresql-server rh-postgresql12-syspaths
+# yum install rh-postgresql12-postgresql-server \
+rh-postgresql12-syspaths \
+rh-postgresql12-postgresql-evr
 ----
 
-. Install the `rh-postgresql12-postgresql-evr` package:
-+
-----
-# yum install rh-postgresql12-postgresql-evr
-----
-+
 . To initialize, start, and enable PostgreSQL service, enter the following commands:
 +
 [options="nowrap" subs="verbatim,quotes"]
--- a/guides/common/modules/proc_installing-postgresql.adoc
+++ b/guides/common/modules/proc_installing-postgresql.adoc
@@ -12,6 +12,12 @@ include::snip_firewalld.adoc[]
 ----
 # yum install postgresql-server
 ----
+
+. Install the `rh-postgresql12-postgresql-evr` package:
++
+----
+# yum install rh-postgresql12-postgresql-evr
+----
 +
 . To initialize, start, and enable PostgreSQL service, enter the following commands:
 +
--- a/guides/common/modules/proc_installing-postgresql.adoc
+++ b/guides/common/modules/proc_installing-postgresql.adoc
@@ -10,7 +10,7 @@ include::snip_firewalld.adoc[]
 . To install PostgreSQL, enter the following command:
 +
 ----
-# yum install postgresql-server
+# yum install rh-postgresql12-postgresql-server rh-postgresql12-syspaths
 ----
 
 . Install the `rh-postgresql12-postgresql-evr` package:
--- a/guides/common/modules/proc_installing-postgresql.adoc
+++ b/guides/common/modules/proc_installing-postgresql.adoc
@@ -22,11 +22,11 @@ rh-postgresql12-postgresql-evr
 # postgresql-setup initdb
 ----
 +
-. Edit the `/var/lib/pgsql/data/postgresql.conf` file:
+. Edit the `/var/opt/rh/rh-postgresql12/lib/pgsql/data/postgresql.conf` file:
 +
 [options="nowrap" subs="verbatim,quotes"]
 ----
-# vi /var/lib/pgsql/data/postgresql.conf
+# vi /var/opt/rh/rh-postgresql12/lib/pgsql/data/postgresql.conf
 ----
 +
 . Remove the `#` and edit to listen to inbound connections:
@@ -36,11 +36,11 @@ rh-postgresql12-postgresql-evr
 listen_addresses = '*'
 ----
 +
-. Edit the `/var/lib/pgsql/data/pg_hba.conf` file:
+. Edit the `/var/opt/rh/rh-postgresql12/lib/pgsql/data/pg_hba.conf` file:
 +
 [options="nowrap" subs="verbatim,quotes"]
 -----
-# vi /var/lib/pgsql/data/pg_hba.conf
+# vi /var/opt/rh/rh-postgresql12/lib/pgsql/data/pg_hba.conf
 -----
 +
 . Add the following line to the file:
--- a/guides/common/modules/proc_installing-postgresql.adoc
+++ b/guides/common/modules/proc_installing-postgresql.adoc
@@ -1,7 +1,7 @@
 [id="installing-postgresql_{context}"]
 = Installing PostgreSQL
 
-You can install only the same version of PostgreSQL that is installed with the `{foreman-installer}` tool during an internal database installation. You can install PostgreSQL using Red Hat Enterprise Linux Server 7 repositories or from an external source, as long as the version is supported. {Project} supports PostgreSQL version 9.2.
+You can install only the same version of PostgreSQL that is installed with the `{foreman-installer}` tool during an internal database installation. You can install PostgreSQL using Red Hat Enterprise Linux Server 7 repositories or from an external source, as long as the version is supported. {Project} supports PostgreSQL version 12.1.
 
 include::snip_firewalld.adoc[]
 
--- a/guides/common/modules/proc_installing-postgresql.adoc
+++ b/guides/common/modules/proc_installing-postgresql.adoc
@@ -15,13 +15,11 @@ rh-postgresql12-syspaths \
 rh-postgresql12-postgresql-evr
 ----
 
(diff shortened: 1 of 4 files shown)


guides/common/modules/proc_installing-postgresql.adoc
```diff
@@ -12,6 +12,12 @@ include::snip_firewalld.adoc[]
 ----
 # yum install postgresql-server
```
ekohl, line 13: This should match the EVR package, which means `rh-postgresql12-postgresql-server`. You may want to suggest `rh-postgresql12-syspaths`, but that's not strictly needed.

guides/common/modules/proc_installing-postgresql.adoc
```diff
@@ -12,6 +12,12 @@ include::snip_firewalld.adoc[]
...
 ----
+
+. Install the `rh-postgresql12-postgresql-evr` package:
++
+----
+# yum install rh-postgresql12-postgresql-evr
```
ekohl, line 19: Note that we only build this package on EL7 now so the host must be EL7. Upstream we do build it for EL8 where it's named `postgresql-evr`. I don't know about downstream.

ekohl (changes requested):
I can't comment on line 4, but it states:
> as long as the version is supported. {Project} supports PostgreSQL version 9.2.

That's no longer true and current versions require at least version 10.

The whole guide is also very RHEL specific since subscription-manager isn't really used on CentOS.

spetrosi:
> I can't comment on line 4, but it states:
> 
> > as long as the version is supported. {Project} supports PostgreSQL version 9.2.
> 
> That's no longer true and current versions require at least version 10.
This has been reported and is being processed in [BZ#1871192](https://bugzilla.redhat.com/show_bug.cgi?id=1871192)
> 
> The whole guide is also very RHEL specific since subscription-manager isn't really used on CentOS.
It is, I do not have bandwidth to address it now, hopefully some contributors will contribute some time

guides/common/modules/proc_installing-postgresql.adoc
spetrosi: Do you say that the command `yum install postgresql-server` doesn't work and should be substituted by `yum install rh-postgresql12-syspaths`?

guides/common/modules/proc_installing-postgresql.adoc
spetrosi: Downstream we say that the external database must be installed on RHEL7, so keeping the EL7 name works fine.

guides/common/modules/proc_installing-postgresql.adoc
ekohl: Yes. Effectively you want to install `yum install ${scl_prefix}postgresql-server ${scl_prefix}postgresql-evr`. On EL7 `$scl_prefix` is `rh-postgresql12-` (right now) but on EL8 it would be empty.

If you don't install `rh-postgresql12-syspaths` the service also becomes `rh-postgresql12-postgresql` instead of `postgresql` so `systemctl` commands would also need to be modified. If the syspaths package is installed, you can keep using all the regular paths, but it will conflict with the regular postgresql packages.

Note that if users have an external postgresql they also need to upgrade their installation. https://github.com/theforeman/foreman-installer/blob/71f60592f33a3056f52791a86876b9ae3ce9235f/hooks/pre/30-el7_upgrade_postgresql.rb is what our installer does for systems where it's bundled. Also note that the storage path changed to `/var/opt/rh/rh-postgresql12/lib/pgsql` instead of `/var/lib/pgsql`. This may have impact on storage requirements and backup procedures.

spetrosi: I used the packages for EL7 because the guides say to install PostgreSQL on RHEL8.

guides/common/modules/proc_installing-postgresql.adoc
```diff
@@ -10,7 +10,13 @@ include::snip_firewalld.adoc[]
...
 +
 ----
-# yum install postgresql-server
+# yum install rh-postgresql12-postgresql-server rh-postgresql12-syspaths
+----
+
+. Install the `rh-postgresql12-postgresql-evr` package:
++
+----
+# yum install rh-postgresql12-postgresql-evr
```
ekohl, line 19: I'd suggest to do this in a single `yum install` line.

guides/common/modules/proc_installing-postgresql.adoc
spetrosi: Updated paths and PostgreSQL supported version in #271

ekohl (changes requested):
The installing proc still has `as long as the version is supported. {Project} supports PostgreSQL version 9.2.` but we require at least version 10 for a while.

It also has `/var/lib/pgsql/data` which also changed to `/var/opt/rh/rh-postgresql12/lib/pgsql/data`.

Also I'd also change the structure a bit: first run `postgresql-setup initdb`, then modify files in the data directory, then run `systemctl enable --now postgresql` (which is the same as enable + start). This saves a restart.

Instead of the psql shell, you can also use command line commands such as `createuser` and `createdb`.

Another thing to note is that the external database must be created with the correct locale and language. If it's done wrong, the application fails.
```
createuser --pwprompt foreman
createdb --owner=foreman --encoding=UTF-8 --locale=en_US.utf8 foreman
```

I don't know if Candlepin has the same requirements. Note that this changes the command to prompt for a password. The benefit is that it's not saved to your shell history. That last part is also a concern with the ping tests below. By default psql should prompt for a password but you can use `psql --password`.

spetrosi:
> The installing proc still has `as long as the version is supported. {Project} supports PostgreSQL version 9.2.` but we require at least version 10 for a while.
> 
> It also has `/var/lib/pgsql/data` which also changed to `/var/opt/rh/rh-postgresql12/lib/pgsql/data`.
I am updating paths and PostgreSQL supported version in #271. This PR focuses on the required packages.
> 
> Also I'd also change the structure a bit: first run `postgresql-setup initdb`, then modify files in the data directory, then run `systemctl enable --now postgresql` (which is the same as enable + start). This saves a restart.
Done that.
> 
> Instead of the psql shell, you can also use command line commands such as `createuser` and `createdb`.
I will stick to what we have got now to not spend QE time testing other methods.

guides/common/modules/proc_installing-postgresql.adoc
```diff
@@ -10,16 +10,16 @@ include::snip_firewalld.adoc[]
...
+. To initialize PostgreSQL, enter the following command:
 +
 [options="nowrap" subs="verbatim,quotes"]
 ----
 # postgresql-setup initdb
-# systemctl start postgresql
-# systemctl enable postgresql
 ----
 +
 . Edit the `/var/lib/pgsql/data/postgresql.conf` file:
```
ekohl, line 25: This location is from the non-SCL verison and must be updated.

ekohl:
>  I am updating paths and PostgreSQL supported version in #271. This PR focuses on the required packages.

The config paths are part of the correct package. rh-postgresql12-postgresql doesn't read the locations so the instructions are broken. IMHO you must update all package names and fix all paths to match.

ekohl:
Let's first finish #271 and then rebase this.

guides/common/modules/proc_installing-postgresql.adoc
```diff
@@ -1,7 +1,7 @@
 [id="installing-postgresql_{context}"]
 = Installing PostgreSQL
 
-You can install only the same version of PostgreSQL that is installed with the `{foreman-installer}` tool during an internal database installation. You can install PostgreSQL using Red Hat Enterprise Linux Server 7 repositories or from an external source, as long as the version is supported. {Project} supports PostgreSQL version 9.2.
+You can install only the same version of PostgreSQL that is installed with the `{foreman-installer}` tool during an internal database installation. You can install PostgreSQL using Red Hat Enterprise Linux Server 7 repositories or from an external source, as long as the version is supported. {Project} supports PostgreSQL version 12.1.
```
ekohl, line 4: Upstream still states PG 10+. On Ubuntu 18.04 this is the standard.

Also wondering if Red Hat Enterprise Linux Server 7 shouldn't use some macro so upstream can mention CentOS. 

However, the whole document is written very much with Satellite downstream in mind. Perhaps there should be a separate effort to make this guide really usable in upstream.

guides/common/modules/proc_installing-postgresql.adoc
```diff
@@ -10,23 +10,23 @@ include::snip_firewalld.adoc[]
 . To install PostgreSQL, enter the following command:
 +
 ----
-# yum install postgresql-server
+# yum install rh-postgresql12-postgresql-server \
+rh-postgresql12-syspaths \
+rh-postgresql12-postgresql-evr
```
ekohl, line 15: It's important to note that the evr package is only needed for Katello. However, the rest of the document is also Katello specific by mentioning Candlepin.

suyogsainkar:
@spetrosi looks good to me from a peer review point of view. please feel free to merge if the above two items/comments from @ekohl have been addressed. thanks.

suyogsainkar (approved):

guides/common/modules/proc_installing-postgresql.adoc
spetrosi: I would like to keep it as a separate effort. This will require a review and tests of community members.

ekohl:
Thanks. I don't know if you already have a guide documenting upgrades, but users upgrading from 6.7 to 6.8 with an external DB need to be aware of this as well.

spetrosi:
Thank you, @ekohl. This has been taken care of.

PR #265, opened 2020-09-16
Apply changes from PR258 about Azure provisioning using custom images.

suyogsainkar:
PR for 6.8-Beta


suyogsainkar:
Related PR#258.

PR #271, opened 2020-09-22
Update supported version of PostgreSQL to 12.1

spetrosi:
Bug 1871192 - postgres version path of postgres not updated in satellite 6.8 beta Document
    
https://bugzilla.redhat.com/show_bug.cgi?id=1871192


ekohl (changes requested):
If you're changing the file paths, you must also update the `yum install postgresql-server` package line to match. Otherwise they're out of sync with each other.

spetrosi:
This changes will be applied as part of #260

PR #305, opened 2020-11-04
initial draft of new Ansible guides

melcorr:
Hi @spetrosi 

Perhaps you can help me with my thought process. 

I have stripped out the Ansible parts from the Admin guide and the Managing Hosts guide, created modules from them, and made the chapters into assemblies. 

I was thinking of stripping out the Running Jobs on Hosts section into modules, making an assembly and adding the assembly both to the new Ansible guide and the existing Managing Hosts guide becaue it is relevant both to Ansible users and those with their own script. 

I was thinking of adding the Tower section and upstreaming the Ansible Tower term so it is of use upstream too. 

I think the Satellite Ansible Content Connections is actually the Foreman Ansible modules. You can use this to manage Foreman with Ansible. I think a better title to that section would be Managing Foreman with Ansible. 

I will also write an introduction to this guide that gives a breakdown of everything that can be found in the guide.  

With the bit I have done so far, everything is building. I will sort the TOC levels tomorrow. 

If you've any thoughts, let me know. I will continue tomorrow.

PR #307, opened 2020-11-05
fix command syntax

evgeni:


melcorr:
If you don't mind, I'll fix this in PR #305

evgeni:
sure, go ahead

PR #311, opened 2020-11-12
Disable weak encryption using installer commands

spetrosi:
The doc say to make manual edits to files while the same can be done automatically by running the installer

Bug 1895336 - [Doc Bug] Disable weak ciphers out of date and are already updated in the documentation

https://bugzilla.redhat.com/show_bug.cgi?id=1895336

### DIFF
--- a/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
+++ b/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
@@ -14,7 +14,7 @@ Apache and Qpid services in {Project} have TLS 1.0 and 1.1 encryption enabled by
 # vi /etc/foreman-installer/custom-hiera.yaml
 ----
 
-. Add the following entry:
+. Add the following entries:
 +
 [options="nowrap"]
 ----
--- a/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
+++ b/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
@@ -27,12 +27,12 @@ To disable weak encryption for {Project}, complete the following steps:
 # Apache
 apache::mod::ssl::ssl_protocol: [ 'ALL' , '-SSLv3' , '-TLSv1' , '-TLSv1.1' , '+TLSv1.2' ]
 ----
-+
+
 . Enter the `{foreman-installer}` command with the following arguments:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {foreman-installer} --foreman-proxy-tls-disabled-versions=1.1 \
+# {foreman-installer} \
 --foreman-proxy-content-qpid-router-ssl-protocols=TLSv1.2
 ----
 +
@@ -67,7 +67,8 @@ apache::mod::ssl::ssl_protocol: [ 'ALL' , '-SSLv3' , '-TLSv1' , '-TLSv1.1' , '+T
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {foreman-installer} --foreman-proxy-content-qpid-router-ssl-protocols=TLSv1.2
+# {foreman-installer} \
+--foreman-proxy-content-qpid-router-ssl-protocols=TLSv1.2
 ----
 +
 . Restart the `{foreman-maintain}` services:
--- a/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
+++ b/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
@@ -20,6 +20,10 @@ Apache and Qpid services in {Project} have TLS 1.0 and 1.1 encryption enabled by
 ----
 # Apache
 apache::mod::ssl::ssl_protocol: [ 'ALL' , '-SSLv3' , '-TLSv1' , '-TLSv1.1' , '+TLSv1.2' ]
+
+# QPID Dispatch
+foreman_proxy_content::qpid_router_ssl_protocols: [ 'TLSv1.2' ]
+foreman_proxy_content::qpid_router_ssl_ciphers: 'ALL:!aNULL:+HIGH:-SSLv3:!IDEA-CBC-SHA'
 ----
 
 . Enter the `{foreman-installer}` command to configure Qpid to use TLS 1.2 only:
--- a/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
+++ b/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
@@ -1,11 +1,7 @@
-[[Disabling_Weak_SSL_Encryption]]
-
-== Disabling Weak Encryption
-
-You might want to change the encryption settings for {Project} depending on the security requirements of your infrastructure or to fix vulnerabilities quickly. Use the following sections to disable weak SSL encryption and 64-bit cipher suites.
-
 [[Disabling_Weak_SSL_2-0_and_3-0_Encryption]]
-=== Disabling TLS 1.0 and 1.1 Encryption
+== Disabling TLS 1.0 and TLS 1.1 Encryption
+
+You might want to change the encryption settings for {Project} depending on the security requirements of your infrastructure or to fix vulnerabilities quickly.
 
 Apache and Qpid services in {Project} have TLS 1.0 and 1.1 encryption enabled by default. Use this procedure on {Project} and {SmartProxy} to configure {Project} and {SmartProxy} to only allow TLS 1.2, as required, for example, by PCI-DSS regulations.
 
@@ -39,35 +35,4 @@ apache::mod::ssl::ssl_protocol: [ 'ALL' , '-SSLv3' , '-TLSv1' , '-TLSv1.1' , '+T
 [options="nowrap" subs="+quotes,attributes"]
 ----
 # {foreman-maintain} service restart
-----
-
-[[Disabling_64-bit_Block_Size_Cipher_Suites]]
-=== Disabling 64-bit Block Size Cipher Suites (SWEET32)
-
-If you want to update your cipher suites for {Project}, you can edit the ciphers and then add your changes to the `/etc/foreman-installer/custom-hiera.yaml` file to make these changes persistent.
-
-You can use the following procedure to update your cipher suite.
-
-The minimum browser requirements for the following Ciphers is Firefox 27.
-
-.Procedure
-
-. Open the `/etc/foreman-installer/custom-hiera.yaml` file for editing:
-+
-[options="nowrap" subs="+quotes,attributes"]
-----
-# vi /etc/foreman-installer/custom-hiera.yaml
-----
-
-. Add the following entry for `apache`:
-+
-----
-apache::mod::ssl::ssl_cipher: ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256
-----
-
-. Run the `{foreman-installer}` tool to add the changes to the Apache configuration:
-+
-[options="nowrap" subs="+quotes,attributes"]
-----
-# {installer-scenario}
 ----
--- a/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
+++ b/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
@@ -59,28 +59,15 @@ To disable weak encryption for {SmartProxy}, complete the following steps:
 +
 [options="nowrap"]
 ----
-# Foreman Proxy
-foreman_proxy::tls_disabled_versions: [ '1.1' ]
-
-# Dynflow
-foreman_proxy::plugin::dynflow::tls_disabled_versions: [ '1.1' ]
-
 # Apache
 apache::mod::ssl::ssl_protocol: [ 'ALL' , '-SSLv3' , '-TLSv1' , '-TLSv1.1' , '+TLSv1.2' ]
-
-# QPID Dispatch
-foreman_proxy_content::qpid_router_ssl_protocols: [ 'TLSv1.2' ]
-foreman_proxy_content::qpid_router_ssl_ciphers: 'ALL:!aNULL:+HIGH:-SSLv3:!IDEA-CBC-SHA'
-
-# PULP
-pulp::ssl_protocol: "ALL -SSLv3 -TLSv1 -TLSv1.1 +TLSv1.2"
 ----
 +
 . Rerun the `{foreman-installer}` tool with no arguments:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {foreman-installer} 
+# {foreman-installer} --foreman-proxy-content-qpid-router-ssl-protocols=TLSv1.2
 ----
 +
 . Restart the `{foreman-maintain}` services:
--- a/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
+++ b/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
@@ -63,7 +63,7 @@ To disable weak encryption for {SmartProxy}, complete the following steps:
 apache::mod::ssl::ssl_protocol: [ 'ALL' , '-SSLv3' , '-TLSv1' , '-TLSv1.1' , '+TLSv1.2' ]
 ----
 +
-. Rerun the `{foreman-installer}` tool with no arguments:
+. Enter the `{foreman-installer}` command with the following arguments:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
--- a/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
+++ b/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
@@ -32,10 +32,3 @@ foreman_proxy_content::qpid_router_ssl_ciphers: 'ALL:!aNULL:+HIGH:-SSLv3:!IDEA-C
 # {foreman-installer} \
 --foreman-proxy-content-qpid-router-ssl-protocols=TLSv1.2
 ----
-
-. Restart the `{foreman-maintain}` services:
-+
-[options="nowrap" subs="+quotes,attributes"]
-----
-# {foreman-maintain} service restart
-----
--- a/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
+++ b/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
@@ -7,11 +7,9 @@ You might want to change the encryption settings for {Project} depending on the 
 [[Disabling_Weak_SSL_2-0_and_3-0_Encryption]]
 === Disabling Weak SSL 2.0 and SSL 3.0 Encryption
 
-Apache and Qpid services in {Project} use SSL 2.0 and SSL 3.0 encryption by default. If your {Project} or {SmartProxy} fails Nessus scans because of SSL vulnerabilities, or your security infrastructure requires that you disable SSL 2.0 and SSL 3.0, complete this procedure on {Project} or {SmartProxy} to remove weak encryption.
+Apache and Qpid services in {Project} have SSL encryption disabled by default, however, TLS 1.0 and 1.1 encryption is enabled. If your {Project} and {SmartProxy} fail Nessus scans because of encryption vulnerabilities, or your security infrastructure requires that you disable TSL 1.0 and TSL 1.1, complete this procedure on {Project} and {SmartProxy} to remove weak encryption.
 
 .Procedure
-
-To disable weak encryption, complete the following steps on {Project} or {SmartProxy}:
 
 . Open the `/etc/foreman-installer/custom-hiera.yaml` file for editing:
 +
@@ -28,7 +26,7 @@ To disable weak encryption, complete the following steps on {Project} or {SmartP
 apache::mod::ssl::ssl_protocol: [ 'ALL' , '-SSLv3' , '-TLSv1' , '-TLSv1.1' , '+TLSv1.2' ]
 ----
 
-. Enter the `{foreman-installer}` command with the following arguments:
+. Enter the `{foreman-installer}` command to configure Qpid to use TLS 1.2 only:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
--- a/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
+++ b/guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
@@ -5,9 +5,9 @@
 You might want to change the encryption settings for {Project} depending on the security requirements of your infrastructure or to fix vulnerabilities quickly. Use the following sections to disable weak SSL encryption and 64-bit cipher suites.
 
 [[Disabling_Weak_SSL_2-0_and_3-0_Encryption]]
-=== Disabling Weak SSL 2.0 and SSL 3.0 Encryption
+=== Disabling TLS 1.0 and 1.1 Encryption
 
-Apache and Qpid services in {Project} have SSL encryption disabled by default, however, TLS 1.0 and 1.1 encryption is enabled. If your {Project} and {SmartProxy} fail Nessus scans because of encryption vulnerabilities, or your security infrastructure requires that you disable TSL 1.0 and TSL 1.1, complete this procedure on {Project} and {SmartProxy} to remove weak encryption.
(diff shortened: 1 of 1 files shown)


guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
```diff
@@ -24,28 +24,16 @@ To disable weak encryption for {Project}, complete the following steps:
...
-foreman_proxy_content::qpid_router_ssl_ciphers: 'ALL:!aNULL:+HIGH:-SSLv3:!IDEA-CBC-SHA'
 ----
 +
-. Rerun the `{foreman-installer}` tool with no arguments:
+. Enter the `{foreman-installer}` command with the following arguments:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {foreman-installer}
+# {foreman-installer} --foreman-proxy-tls-disabled-versions=1.1 \
```
ekohl, line 35: TLS 1.1 is gone since Foreman Proxy 2.1 (https://github.com/theforeman/smart-proxy/commit/ebdde78c329d6bf67b198446dae64d8952d0ac4a). Please drop this.

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
```diff
@@ -24,28 +24,16 @@ To disable weak encryption for {Project}, complete the following steps:
...
-foreman_proxy::tls_disabled_versions: [ '1.1' ]
-
-# Dynflow
-foreman_proxy::plugin::dynflow::tls_disabled_versions: [ '1.1' ]
-
 # Apache
 apache::mod::ssl::ssl_protocol: [ 'ALL' , '-SSLv3' , '-TLSv1' , '-TLSv1.1' , '+TLSv1.2' ]
-
-# Tomcat / Candlepin
-candlepin::tls_versions: [ '1.2' ]
```
ekohl, line 37: FYI: since 2.1 this is redundant: https://projects.theforeman.org/issues/29625

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
```diff
@@ -24,28 +24,16 @@ To disable weak encryption for {Project}, complete the following steps:
 +
 [options="nowrap"]
 ----
-# Foreman Proxy
-foreman_proxy::tls_disabled_versions: [ '1.1' ]
-
-# Dynflow
-foreman_proxy::plugin::dynflow::tls_disabled_versions: [ '1.1' ]
```
ekohl, line 31: Redundant since https://projects.theforeman.org/issues/29624 which dropped TLS 1.1

ekohl (changes requested):
I think (even with Satellite 6.8) only Apache and qpid may accept TLS 1.1. Apache because I wasn't sure about older yum clients and qpid because I didn't think about it.

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
spetrosi: The option `foreman-proxy-tls-disabled-versions=1.1` disables TLS version 1.1. From the help output:
`--foreman-proxy-tls-disabled-versions  List of TLS versions that will be disabled from the default`

ekohl: But it's redundant. If you don't specify it, the Smart Proxy won't use it either. The only use it has now is to disable TLS 1.2 and only use TLS 1.3, but that's not really relevant for most deployment.

spetrosi: Got it, thank you for the explanation. What with Satellites that exist longer, say for those who upgrade from 6.4. Will TLS 1.1 be enabled on them? Does installer disable TLS 1.1 when upgrading?

ekohl: It will still configure it but the smart proxy application **always** disables some protocols:
https://github.com/theforeman/smart-proxy/blob/7c818d0cf1606a7d16c8334ab4fd256370eac365/lib/launcher.rb#L73-L79
Currently that's SSLv2, SSLv3, TLS 1.0 and (since Foreman 2.1 / Satellite 6.8) TLS 1.1. It then parses the disabled option:
https://github.com/theforeman/smart-proxy/blob/7c818d0cf1606a7d16c8334ab4fd256370eac365/lib/launcher.rb#L81-L90
So effectively what happens is:
```ruby
options |= OpenSSL::SSL::OP_NO_TLSv1_1 # hardcoded
options |= OpenSSL::SSL::OP_NO_TLSv1_1 # via options
```

Since it doesn't do anything, I didn't bother cleaning up options in a migration. The user will still only use strong encryption.

spetrosi: Thank you for the clarification, I have removed this option.

ekohl (changes requested):
Can we also change the headers?

`Disabling Weak SSL 2.0 and SSL 3.0 Encryption` -> I'm uncertain about qpid, but all other services disable SSL 2 and 3 out of the box. Only Apache and qpid allow TLS 1.0 and TLS 1.1 by default. I mostly wasn't sure about older clients.

Also note that the instructions for `{Project}` and `{SmartProxy}` are now the same so they can be merged.

As for `Disabling 64-bit Block Size Cipher Suites` I have 2 notes.
* If this is really a problem, then it should be done by default [here](https://github.com/theforeman/foreman-installer/blob/4be4372dbc70aa3cd39718f13a72324ffae6a322/config/foreman.hiera/security.yaml#L2). However, I think that https://github.com/theforeman/foreman-installer/commit/e89c65e602243134aa533e82bc9c16134d5e44db already took care of this. If so, then it's been solved since 1.19 (Satellite 6.5?).

However, if it's still relevant then the user should only edit the hiera setting and then run the installer. Manual modification of files + restarting services should not be done.

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
```diff
@@ -24,28 +24,16 @@ To disable weak encryption for {Project}, complete the following steps:
...
-
 # Apache
 apache::mod::ssl::ssl_protocol: [ 'ALL' , '-SSLv3' , '-TLSv1' , '-TLSv1.1' , '+TLSv1.2' ]
-
-# Tomcat / Candlepin
-candlepin::tls_versions: [ '1.2' ]
-
-# QPID Dispatch
-foreman_proxy_content::qpid_router_ssl_protocols: [ 'TLSv1.2' ]
-foreman_proxy_content::qpid_router_ssl_ciphers: 'ALL:!aNULL:+HIGH:-SSLv3:!IDEA-CBC-SHA'
```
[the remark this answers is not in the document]
spetrosi: Does the `--foreman-proxy-content-qpid-router-ssl-protocols=TLSv1.2` option replaces this config line?

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
```diff
@@ -24,28 +24,16 @@ To disable weak encryption for {Project}, complete the following steps:
...
 [options="nowrap"]
 ----
-# Foreman Proxy
-foreman_proxy::tls_disabled_versions: [ '1.1' ]
-
-# Dynflow
-foreman_proxy::plugin::dynflow::tls_disabled_versions: [ '1.1' ]
-
 # Apache
 apache::mod::ssl::ssl_protocol: [ 'ALL' , '-SSLv3' , '-TLSv1' , '-TLSv1.1' , '+TLSv1.2' ]
```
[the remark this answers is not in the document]
spetrosi: Following the Chesterton’s Fence principle, I would leave this here for now.

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
```diff
@@ -7,94 +7,35 @@ You might want to change the encryption settings for {Project} depending on the
 [[Disabling_Weak_SSL_2-0_and_3-0_Encryption]]
 === Disabling Weak SSL 2.0 and SSL 3.0 Encryption
 
-If your {Project} fails Nessus scans because of SSL vulnerabilities, or your security infrastructure requires that you disable SSL 2.0 and SSL 3.0, you can edit the `/etc/foreman-installer/custom-hiera.yaml` file to remove weak encryption.
+Apache and Qpid services in {Project} use SSL 2.0 and SSL 3.0 encryption by default. If your {Project} or {SmartProxy} fails Nessus scans because of SSL vulnerabilities, or your security infrastructure requires that you disable SSL 2.0 and SSL 3.0, complete this procedure on {Project} or {SmartProxy} to remove weak encryption.
```
spetrosi, line 10: @ekohl can you help to verify the first sentence here. Is that correct that Apache and Qpid use SSL 2.0 and 3.0?
You suggested changing the title, can you elaborate on how to change it exactly?

spetrosi: I have a default installation of Satellite 6.7. I see the following on it: 
```
PORT    STATE SERVICE
443/tcp open  https
| ssl-enum-ciphers: 
|   SSLv3: No supported ciphers found
|   TLSv1.0: 
|     ciphers: 
|       TLS_DHE_RSA_WITH_AES_128_CBC_SHA - strong
|       TLS_DHE_RSA_WITH_AES_256_CBC_SHA - strong
|       TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA - strong
|       TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA - strong
|       TLS_RSA_WITH_AES_128_CBC_SHA - strong
|       TLS_RSA_WITH_AES_256_CBC_SHA - strong
|     compressors: 
|       NULL
|   TLSv1.1: 
|     ciphers: 
|       TLS_DHE_RSA_WITH_AES_128_CBC_SHA - strong
|       TLS_DHE_RSA_WITH_AES_256_CBC_SHA - strong
|       TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA - strong
|       TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA - strong
|       TLS_RSA_WITH_AES_128_CBC_SHA - strong
|       TLS_RSA_WITH_AES_256_CBC_SHA - strong
|     compressors: 
|       NULL
|   TLSv1.2: 
|     ciphers: 
|       TLS_DHE_RSA_WITH_AES_128_CBC_SHA - strong
|       TLS_DHE_RSA_WITH_AES_128_CBC_SHA256 - strong
|       TLS_DHE_RSA_WITH_AES_128_GCM_SHA256 - strong
|       TLS_DHE_RSA_WITH_AES_256_CBC_SHA - strong
|       TLS_DHE_RSA_WITH_AES_256_CBC_SHA256 - strong
|       TLS_DHE_RSA_WITH_AES_256_GCM_SHA384 - strong
|       TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA - strong
|       TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256 - strong
|       TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 - strong
|       TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA - strong
|       TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384 - strong
|       TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 - strong
|       TLS_RSA_WITH_AES_128_CBC_SHA - strong
|       TLS_RSA_WITH_AES_128_CBC_SHA256 - strong
|       TLS_RSA_WITH_AES_128_GCM_SHA256 - strong
|       TLS_RSA_WITH_AES_256_CBC_SHA - strong
|       TLS_RSA_WITH_AES_256_CBC_SHA256 - strong
|       TLS_RSA_WITH_AES_256_GCM_SHA384 - strong
|     compressors: 
|       NULL
|_  least strength: strong
```
I see that qpid is running when entering `systemctl status qpidd`. Sounds like it does not use SSL by default as well.

ekohl: It would have surprised me if it did, especially SSL v2 which is really really ancient. Even SSLv3 was considered insecure back in 2014 which is already 6 years ago. If RHEL shipped something insecure by default, that would have surprised me.

spetrosi: Shall we then dump this whole chapter 7. Disabling Weak Encryption because it is irrelevant is the current RHEL?

spetrosi: Wait, 7.2. Disabling 64-bit Block Size Cipher Suites (SWEET32) is still relevant.

ekohl: > Shall we then dump this whole chapter 7. Disabling Weak Encryption because it is irrelevant is the current RHEL?

I'd say that currently the instructions are valid for some. Various institutions only allow TLS 1.2 (I think PCI requires it?). Since the default install allows TLS 1.0, that is still useful. However, it should be clearly stated that its disabling TLS 1.0 and 1.1, not SSL v2/v3.

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
```diff
@@ -7,94 +7,33 @@ You might want to change the encryption settings for {Project} depending on the
 [[Disabling_Weak_SSL_2-0_and_3-0_Encryption]]
 === Disabling Weak SSL 2.0 and SSL 3.0 Encryption
 
-If your {Project} fails Nessus scans because of SSL vulnerabilities, or your security infrastructure requires that you disable SSL 2.0 and SSL 3.0, you can edit the `/etc/foreman-installer/custom-hiera.yaml` file to remove weak encryption.
+Apache and Qpid services in {Project} have SSL encryption disabled by default, however, TLS 1.0 and 1.1 encryption is enabled. If your {Project} and {SmartProxy} fail Nessus scans because of encryption vulnerabilities, or your security infrastructure requires that you disable TSL 1.0 and TSL 1.1, complete this procedure on {Project} and {SmartProxy} to remove weak encryption.
```
ekohl, line 10: `SSL encryption` is ambiguous. In many places SSL and TLS are used interchangably. I would not mention SSLv2 and SSLv3 at all. In 2020 no product should even think about shipping them IMHO.
```suggestion
Apache and Qpid services in {Project} have TLS 1.0 and 1.1 encryption enabled by default. This procedure tightens the security to only allow TLS 1.2, as required in various places (such as PCI-DSS regulations).
```

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
```diff
@@ -7,94 +7,33 @@ You might want to change the encryption settings for {Project} depending on the
 [[Disabling_Weak_SSL_2-0_and_3-0_Encryption]]
 === Disabling Weak SSL 2.0 and SSL 3.0 Encryption
```
ekohl, line 8:
```suggestion
=== Disabling TLS 1.0 and 1.1 Encryption
```

ekohl (changes requested):

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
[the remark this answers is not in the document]
ekohl: No, only the line above.

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
ekohl: > Wait, 7.2. Disabling 64-bit Block Size Cipher Suites (SWEET32) is still relevant.

I think your example proves it doesn't. I don't see DES ciphers in there. Looks like it can be dropped.

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
spetrosi: Done.

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
spetrosi: Done.

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
spetrosi: Are you sure that it can be dropped? Ganesh has a customer who used this in Satellite 6.5, see [BZ#1683832](https://bugzilla.redhat.com/show_bug.cgi?id=1683832).

ekohl: I can't speak for 6.5 since that's rather ancient. It should have the changes. Note that the BZ is about making the changes persistent. However, I think that the entire instructions are no longer needed.

ekohl: Used an internal reference Satellite 6.5 and that is not vulnerable to it. It was relevant on older releases but those are now EOL.

spetrosi: I am more than happy to dump it then

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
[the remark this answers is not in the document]
ekohl: Turns out that the default installation needs this. I think that this line should even be the default. Without it, qpid is vulnerable to SWEET32. Now I am wondering if this is sufficient. We have a qpid router, but there's also another qpid service listening. That may need a similar line.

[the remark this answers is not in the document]
spetrosi: I will not remove these lines and return them then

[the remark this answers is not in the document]
ekohl: I think it's better to use the command line option, similar to the qpid router ssl protocols.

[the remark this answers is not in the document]
spetrosi: Now this PR adds the following command:
```
# {foreman-installer} \
--foreman-proxy-content-qpid-router-ssl-protocols=TLSv1.2
```
The Bug this PR is opened for also suggests to add the option `--foreman-proxy-tls-disabled-versions=1.1` but you said it is not required, so I removed it. The qpid-router option is still there.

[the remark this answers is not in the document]
ekohl: The bug is correct that the command line option is `--foreman-proxy-tls-disabled-versions`. However, the need to pass that option was removed.

> I think it's better to use the command line option, similar to the qpid router ssl protocols.

What I meant was that you can pass `--foreman-proxy-content-qpid-router-ssl-ciphers '...'` which is better than setting it in `custom-hiera.yaml`.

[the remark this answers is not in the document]
spetrosi: I do not have such an option on Satellite 6.7, I will find some Satellite 6.8 instance and check there next week.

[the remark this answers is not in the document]
ekohl: It should be there. did you check `--full-help`?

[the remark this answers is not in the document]
spetrosi: My bad, the option is there under full-help. I cannot figure out how to use it, I tried the original syntax from the custom-hiera.yml and separating by commas, neither worked:
```
# satellite-installer --foreman-proxy-content-qpid-router-ssl-ciphers 'ALL, !aNULL, +HIGH, -SSLv3, !IDEA-CBC-SHA'
ERROR: Scenario (/etc/foreman-installer/scenarios.d/SLv3,.yaml) was not found, can not continue
# satellite-installer --foreman-proxy-content-qpid-router-ssl-ciphers 'ALL:!aNULL:+HIGH:-SSLv3:!IDEA-CBC-SHA'
ERROR: Scenario (/etc/foreman-installer/scenarios.d/SLv3:!IDEA-CBC-SHA.yaml) was not found, can not continue
```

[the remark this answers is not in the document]
ekohl: I think you may be hitting [Bug #30218](https://projects.theforeman.org/issues/30218) where `-S` (in `-SSLv3`) is interpreted as a scenario switch. The linked BZ suggests it hasn't landed in downstream yet.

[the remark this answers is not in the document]
ekohl: Created https://github.com/theforeman/puppet-foreman_proxy_content/pull/301 to set it out of the box. Probably doesn't help you for now though. Perhaps we should keep this in `custom-hiera.yaml` for now until we have it solved.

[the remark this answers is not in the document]
spetrosi: Thank you, @ekohl. I have removed the config line that configures qpid tls version because this is done via the command line option, I have kept the config line that configures SSL cyphers.

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
```diff
@@ -1,151 +1,41 @@
...
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# vi /etc/foreman-installer/custom-hiera.yaml
-----
-+
-. Add the following entries:
-+
-[options="nowrap"]
+# {foreman-installer} \
+--foreman-proxy-content-qpid-router-ssl-protocols=TLSv1.2
```
ekohl, line 33: So I noticed that when the ciphers were tightened it also stopped listening on TLS 1.0 and 1.1 which would make that redundant. Should we drop this line?

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
```diff
@@ -1,151 +1,41 @@
...
-----
-# {foreman-installer} 
-----
-+
 . Restart the `{foreman-maintain}` services:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
 # {foreman-maintain} service restart
 ----
```
ekohl, line 41: The installer should restart any services. Not sure why you would restart them again after.
```suggestion
```

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
```diff
@@ -1,151 +1,41 @@
...
 . Open the `/etc/foreman-installer/custom-hiera.yaml` file for editing:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
 # vi /etc/foreman-installer/custom-hiera.yaml
 ----
-+
-. Add the following entries:
+
+. Add the following entry:
```
ekohl, line 17: It's 2 entries again
```suggestion
. Add the following entries:
```

ekohl (commented):
Slowly getting there.

guides/doc-Administering_Red_Hat_Satellite/topics/Disabling_Weak_Encryption.adoc
spetrosi: Do you say that we do not need to run the installer at all and that the config lines in custom-hiera.yaml would do the trick? If so, I can remove this step completely and return the step that restarts the services.

ekohl: No, still run the installer but without the argument. Running the installer is crucial to actually making the config file changes.

spetrosi: Did you notice that I removed the `foreman_proxy_content::qpid_router_ssl_protocols: [ 'TLSv1.2' ]` config line? Now we say to only set the `foreman_proxy_content::qpid_router_ssl_ciphers: 'ALL:!aNULL:+HIGH:-SSLv3:!IDEA-CBC-SHA'` configuration. Will that do the trick? Want to check before committing it.

ekohl: It appeared to be when I tried it out. Now I'm not sure if it was just nmap's reporting. Will try to figure it out.

If you're unsure, you can do both for now and we can simplify it in a future iteration.

ekohl (approved):
Overall I think it looks good. Some thoughts:

* By default qpid is still vulnerably to SWEET32. Previously that was not listed as something that should be fixed in the (now removed) chapter. However, I don't know the exact implications of the stricter ciphers regarding client support. I think the current state of this PR probably good enough while we figure it out at a project level
* `--foreman-proxy-content-qpid-router-ssl-protocols=TLSv1.2` appeared to be redundant from my testing. If it is, I'd drop it from the instructions to keep them minimal.
* `+TLSv1.2` is still redundant in the Apache instructions

I have no objection to merging it in the current form and correcting it later.

spetrosi:
Thanks, Ewoud. I will merge it now, we can always correct if any future objections will raise.

PR #312, opened 2020-11-16
Add Adam Ruzicka's feedback

melcorr:
Add ifevals for Ansible context

### DIFF
--- a/guides/common/modules/proc_configuring-an-alternative-directory-to-execute-remote-jobs-on-clients.adoc
+++ b/guides/common/modules/proc_configuring-an-alternative-directory-to-execute-remote-jobs-on-clients.adoc
@@ -2,7 +2,17 @@
 
 = Configuring {Project} to Use an Alternative Directory to Execute Remote Jobs on Hosts
 
+ifeval::["{context}" == "managing-hosts"]
+
 By default, {Project} uses the `/var/tmp` directory on the client system to execute the remote execution jobs. If the client system has `noexec` set for the `/var/` volume or file system, you must configure {Project} to use an alternative directory because otherwise the remote execution job fails since the script cannot be run.
+
+endif::[]
+
+ifeval::["{context}" == "ansible"]
+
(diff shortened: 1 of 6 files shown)


guides/doc-Configuring_Ansible/master.adoc
```diff
@@ -7,9 +7,10 @@ endif::[]
 
 include::common/attributes.adoc[]
 
-:configuring-ansible:
 :context: ansible
 
+:configuring-ansible:
```
spetrosi, line 12: All looks good, In only have a question to this change. Why you do this? Won't it break something?

melcorr: I had a problem, and was troubleshooting, comparing the layout of the Installation docs master pages to what i had done with this new guide. I moved things around trying to figure out what was wrong. This was completely unrelated in the end. I will move this back. Thanks for catching.

guides/common/modules/proc_configuring-an-alternative-directory-to-execute-remote-jobs-on-clients.adoc
```diff
@@ -2,8 +2,18 @@
...
 
+ifeval::["{context}" == "managing-hosts"]
+
 By default, {Project} uses the `/var/tmp` directory on the client system to execute the remote execution jobs. If the client system has `noexec` set for the `/var/` volume or file system, you must configure {Project} to use an alternative directory because otherwise the remote execution job fails since the script cannot be run.
 
+endif::[]
+
+ifeval::["{context}" == "ansible"]
+
+By default, {Project} uses the `/var/tmp` directory on the client system to execute the remote execution jobs. Ansible puts its own files it requires into the `$HOME/.ansible/tmp` directory, where `$HOME` is the home directory of the remote user.
```
adamruzicka, line 13: For ansible, nothing should be going into `/var/tmp` so we can drop the entire sentence

PR #313, opened 2020-11-18
Change the PostgreSQL 12 mount point

spetrosi:
Bug 1896703 - Change directory to "/var/opt/rh/rh-postgresql12/" in Storage Requirements section to avoid warning during upgrade

https://bugzilla.redhat.com/show_bug.cgi?id=1896703

### DIFF
--- a/guides/common/modules/ref_satellite-storage-requirements.adoc
+++ b/guides/common/modules/ref_satellite-storage-requirements.adoc
@@ -27,7 +27,6 @@ ifdef::installing-satellite-server-disconnected[30 GB]
 
 |/var/log/ |10 MB |10 GB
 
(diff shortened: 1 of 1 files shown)


guides/common/modules/ref_satellite-storage-requirements.adoc
```diff
@@ -28,6 +28,7 @@ ifdef::installing-satellite-server-disconnected[30 GB]
 |/var/log/ |10 MB |10 GB
 
 |/var/opt/rh/rh-postgresql12/lib/pgsql/ |100 MB |10 GB
```
ekohl, line 30: Perhaps this should be dropped in favor of the parent path?

spetrosi: Sure, thank you for catching this!!!

ekohl (approved):

PR #175, opened 2020-05-21
Refs #26925 - document host register

ezr-ondrej:
Documentation for theforeman/foreman#6813 by no means complete, more as a reminder.
I do not feel like the right person to do this as: I do not understand provisioning, I do not understand how this documentation is being written properly.

Glad to learn though!

Appreciate any help, or even takeovers :)

PR #316, opened 2020-11-24
Fixes #30475 - Global Registration Documentation

stejskalleos:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1903456

### DIFF
--- a/guides/doc-Managing_Hosts/topics/Global_Registration.adoc
+++ b/guides/doc-Managing_Hosts/topics/Global_Registration.adoc
@@ -1,0 +1,185 @@
+[id="Global_Registration"]
+= Registering Host with Global Registration
+
+With Global Registration feature hosts can be registered to the Foreman with
+just one command.
+Main power of this feature came from utilization of provisioning templates,
+which gives user complete control over the process of the host registration.
+
+
+.Prerequisites
+- Foreman user must have `create_hosts` permission
+- Host OS must have assigned Registration template.
+  Go to the template detail > Associations & select the OS
+  Then on OS detail (Hosts > Operating Systems) set default registration template
+- Root privileges on the host that's going to be registered.
+
+
+.Procedure
+. Login to {ProjectName}
+. Go to Hosts > All Hosts > Register Host
+. Set required parameters and generate the command
+. Log in to the Host
+. Run command generated in {ProjectServer}
++
+[options="nowrap", subs="+quotes,attributes"]
+----
+# Example:
+curl http://_{foreman-example-com}_/register | bash
+----
++
+
+
+.Templates & snippets
+
+[cols=2*]
+|===
+|Global Registration Template
+|Contains steps for registration of host to the Foreman. Rendered when accessing `/register` endpoint.
+
+|Host Registration Template
+|In Foreman named `Linux registration default`, contains commands for Host setup.
+
+|`remote_execution_ssh_keys` snippet
+| Used in Host registration template, deploy SSH keys to the host. Only when `host_registration_remote_execution` parameter is `true`. Part of Host registration template
+
+|`insights` snippet
+| Download & install insights client. Only for Red Hat OS & when global parameter `host_registration_insights` is set to true. Part of Host registration template
+|===
+
+
+.Registration parameters
+* Organization
+* Location
+* Host Group
+* Operating System - No need to specify for registration with subscription-manager
+* Proxy
+* Setup Insights - Install insights client & register host. Override value of `host_registration_insights` global parameter on the Host level. Plus only for Red Hat OS.
+* Remote Execution - Override value of `host_registration_remote_execution` global parameter on the Host level.
+* Token lifetime
+
+Parameters can be extended from the plugins, see ee https://github.com/theforeman/foreman/blob/develop/developer_docs/how_to_create_a_plugin.asciidoc
+
+
+.Authentication
+For authentication we use JSON Web Tokens (JWT):
+[options="nowrap", subs="+quotes,attributes"]
+----
+curl -X GET "https://foreman.example.com/register" -H 'Authorization: Bearer ...
+----
+
+The user who generated the command with JWT is also the one who's permissions
+will be used for authorization.
+This means that if user will lost or gain some additional permissions,
+the JWT will "have" these permissions to.
+Deleting or blocking the user will also invalidate the token.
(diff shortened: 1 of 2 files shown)


spetrosi:
Seems like this PR makes the same thing that #175 that @ezr-ondrej opened

melcorr:
> Seems like this PR makes the same thing that #175 that @ezr-ondrej opened

@stejskalleos  was aware of that PR and has created a more updated version, if I understand correctly?

stejskalleos:
@spetrosi @melcorr 

This PR seems to be same as https://github.com/theforeman/foreman-documentation/pull/175, but in fact it's about different feature (the naming is confusing).

**PR#175**
* Is about registering (or let's say _setting up_) of *already created host* in the Foreman.

**This PR (Global Registration Feature)** 
* It's about creating completely new host to the Foreman
* Delivered in Foreman 2.3, available in Satellite 6.9

We probably should discuss the naming conventions, I understand that having two different features with same name can be confusing.

guides/doc-Managing_Hosts/topics/Global_Registration.adoc
```diff
@@ -0,0 +1,138 @@
...
+.Registration parameters
+* Organization
+* Location
+* Host Group
+* Operating System - No need to specify for registration with subscription-manager
+* Proxy
+* Setup Insights - Install insights client & register host. Override value of `host_registration_insights` global parameter on the Host level. Plus only for Red Hat OS.
+* Remote Execution - Override value of `host_registration_remote_execution` global parameter on the Host level.
+* Token lifetime
+
```
ares, line 60: perhaps we could mention here that plugins may extend this list?

guides/doc-Managing_Hosts/topics/Global_Registration.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+.Settings
+* For changing default Global Registration template go to Administer > Settings > Provisioning
```
ares, line 63: It would be good to mention why would one want to change that. E.g. if you want to customize the default registration behavior, instead of unlocking and modifying the default template, you can clone it and change it via settings. The default template gets overriden on every Foreman upgrade so custom changes would be lost. (please change the text as needed)

guides/doc-Managing_Hosts/topics/Global_Registration.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+
+.Global Parameters
+* `host_registration_remote_execution` - used in `remote_execution_ssh_keys` snippet, default value is `true`.
+* `host_registration_insights` - used in `insights` snippet, default value is `false` for the Foreman and `true` for the Satellite
```
ares, line 68: this should be somehow conditional, for Foreman it's false, for Satellite it's true, @lzap can this be done as one sentence changing the value based on `{ProjectName}`?

guides/doc-Managing_Hosts/topics/Global_Registration.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+
+.With Smart proxy
+For registration through proxy you need to enable Registration module
```
ares, line 72: it would be good to say how, e.g. foreman-installer line example

guides/doc-Managing_Hosts/topics/Global_Registration.adoc
```diff
@@ -0,0 +1,138 @@
...
+[options="nowrap", subs="+quotes,attributes"]
+----
+foreman-installer --foreman-proxy-registration
+----
+
+Then you can call `/register` endpoint on the proxy:
+[options="nowrap", subs="+quotes,attributes"]
+----
+curl http://proxy.example.com/register | bash
+----
```
ares, line 82: it would be also good to explain, that templates will then render curl commands with proxy url so the whole communication flows through the proxy

also we should mention the necessity of templates module and the impact if it's not enabled (host will remain in build mode)

guides/doc-Managing_Hosts/topics/Global_Registration.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+.SSL
+All calls in the templates to the Foreman are secure & use https.
+However, the first call to the Foreman (`curl /register`) requires to have certificate authority
+installed on the host or you can call the register command with `--insecure` parameter.
```
ares, line 87: it would be great to mention how to fetch that SSL cert from the client machine or how to get it on Foreman host (what path to take it from). It's good you mention the `--insecure`, it would also benefit from saying that GRT deploys the certificate for the registration so the rest of the communication is then properly secured.

ares (commented):
This is excellent write up. I left few comments but would add one more thing - the JWT description. We should clarify that it gives user the same permissions the user who generated it has. Also we should mention somewhere the same registration command can be used for multiple hosts, in fact until the JWT expires. Perhaps we could create some 'advanced' chapter where all this would be explained and the flow chart would be added.

guides/doc-Managing_Hosts/topics/Global_Registration.adoc
```diff
@@ -0,0 +1,185 @@
...
+- Host OS must have assigned Registration template.
+  Go to the template detail > Associations & select the OS
+  Then on OS detail (Hosts > Operating Systems) set default registration template
+- Root privileges on the host that's going to be registered.
+
+
+.Procedure
+. Login to {ProjectName}
+. Go to Hosts > All Hosts > Register Host
+. Set required parameters and generate the command
```
pondrejk, line 21: Are there parameters that are always required?

stejskalleos: The only required parameter is `activation_key`, but only for Red Hat family OS.
I see that I forgot to mention it in the list of parameters.

guides/doc-Managing_Hosts/topics/Global_Registration.adoc
```diff
@@ -0,0 +1,185 @@
...
+. Log in to the Host
+. Run command generated in {ProjectServer}
++
+[options="nowrap", subs="+quotes,attributes"]
+----
+# Example:
+curl http://_{foreman-example-com}_/register | bash
+----
```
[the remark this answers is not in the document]
pondrejk: oh I see it below, scratch that

PR #317, opened 2020-11-25
Add orchestrate rake tasks to Troubleshooting

lzap:
This happens quite often, I think it is a good idea to have it here.


spetrosi (commented):
Hi @lzap, thank you for adding this troubleshooting info. I have pushed a commit where I restructured the procedure you added to have all steps within one troubleshooting procedure, and have all preview and perform steps written out so that users know what to do in the first place. How do you like it? Is it correct?

lzap (commented):
I am good, thank you.

PR #318, opened 2020-12-02
Replace outdated xref with a direct URL

spetrosi:


lzap:
We should not be using HTTP URLs like that but we should prefer `{BaseURL}` attribute.

PR #321, opened 2020-12-04
Issue 192: Handle URLs to point to upstream docs upstream

spetrosi:


spetrosi:
Hi @lzap, for tests purposes I made this one link to point to upstream docs upstream and downstream docs downstream. I had to change some internals to make it work.

I had to change Makefile because downstream we have titles in lowercase, and upstream they were in title case (Managing_Hosts), so links did not work both upstream and downstream. I changed Makefile to publish with lowercase upstream too.

I also made use of the {index} attribute. Consider these upstream and downstream links pointing to the same section:
https://docs.theforeman.org/master/Managing_Hosts/index-foreman.html#configuring-and-setting-up-remote-jobs_managing-hosts
https://access.redhat.com/documentation/en-us/red_hat_satellite/6.8/html-single/managing_hosts/index#configuring-and-setting-up-remote-jobs_managing-hosts
While downstream uses just `index#`, upstream has `index-foreman.html`. The solution I came up with seems not elegant but it works. The other solution would be to publish upstream the same way - with `index#`. Do you know if it is possible?

The big issue I faced is links to the Installation guides - the URLs to them are different upstream and downstream - `/Installing_Server_on_Red_Hat/` vs `installing_satellite_server_from_a_connected_network`. Sounds like we will need to use ifevals for these links.

Thank you

spetrosi:
I will need to change links to be in lower case in `web/index.html` too

ekohl:
It looks like it uses the git branch in the URL. May I suggest to rename the default branch to `nightly` so it maps to names Foreman uses for releases?

guides/common/attributes.adoc
```diff
@@ -169,9 +171,11 @@ ifeval::["{build}" == "satellite"]
 :ansible-namespace: `redhat.satellite`
 :ansible-namespace-example: `redhat.satellite._module_name_`
 :awx: Ansible Tower
+:index:
```
lzap, line 174: This is confusing, I suggest that the `index` variable holds either `index.html` or `index-foreman.html`. URLs will also look a bit cleaner.

lzap:
> It looks like it uses the git branch in the URL. May I suggest to rename the default branch to `nightly` so it maps to names Foreman uses for releases?

Where do you see nightly? We use the term `master`, I picked it because I want a consistent master branch name: https://docs.theforeman.org/master/Planning_Guide/index-foreman.html

We already have quite few links shared, I'd rather keep `master` it's not big deal. It's expected for contributors and a random reader won't care much.

lzap (changes requested):
Nice work indeed, I'd like to propose slight variation.

We do define few attributes in the individual documents in `master.adoc` files. I suggest that we add one more variable:

    :SelfURL:{BaseURL}managing_hosts/{index_html}

So in the final text we can just use `see {SelfURL}/running_jobs#sect-blah-blah[Blah blah]`. For links to other docs we need to use what you propose, I suggest to improve `{index_html}` variable a bit (see my other comment). For the special case we can do if statement but it will be present only once, in `master.adoc`. The rest will be pretty consistent.

Alternatively, we can define all those URLs in the `attributes.adoc` for each individual book: `ManagingHostDocURL`, `InstallRedHatDocURL` etc.

guides/common/attributes.adoc
spetrosi: Good notice, will do

spetrosi:
> Nice work indeed, I'd like to propose slight variation.
Thank you!
> 
> We do define few attributes in the individual documents in `master.adoc` files. I suggest that we add one more variable:
> 
> ```
> :SelfURL:{BaseURL}managing_hosts/{index_html}
> ```
> 
> So in the final text we can just use `see {SelfURL}/running_jobs#sect-blah-blah[Blah blah]`. For links to other docs we need to use what you propose, I suggest to improve `{index_html}` variable a bit (see my other comment). For the special case we can do if statement but it will be present only once, in `master.adoc`. The rest will be pretty consistent.
> 
> Alternatively, we can define all those URLs in the `attributes.adoc` for each individual book: `ManagingHostDocURL`, `InstallRedHatDocURL` etc.
This is a perfect solution!
Maybe we can avoid creating more attributes and keep it simple by defining the BaseURL attribute in the master.adoc file for each guide. This will also solve the issue of the Install guides' different titles. I will now try to implement this for the test URL I have worked on in this bug, then make global changes.

ekohl:
> > It looks like it uses the git branch in the URL. May I suggest to rename the default branch to `nightly` so it maps to names Foreman uses for releases?
> 
> Where do you see nightly? We use the term `master`, I picked it because I want a consistent master branch name: https://docs.theforeman.org/master/Planning_Guide/index-foreman.html
> 
> We already have quite few links shared, I'd rather keep `master` it's not big deal. It's expected for contributors and a random reader won't care much.

I opened https://github.com/theforeman/foreman-documentation/issues/322 to not distract too much. IMHO it is a big deal.

spetrosi:
@lzap, I have begun by adding two attributes for the Managing Hosts and Ansible guide, so far everything works as expected. Please review the work I have done, maybe you might find something that can be improved. You can merge this PR if you like the changes.
I am planning to continue doing the same thing for all other URLs and all other guides by opening a PR for URLs pointing to a single guide each time, it would take around 10 PRs. This way it would be easier for me to do the work and review it. How does this sound?

guides/common/attributes.adoc
```diff
@@ -97,6 +96,11 @@ ifeval::["{build}" == "foreman"]
 :ansible-namespace: `theforeman.foreman`
 :ansible-namespace-example: `theforeman.foreman._module_name_`
 :awx: AWX
+:BaseURL: https://docs.theforeman.org/master/
+:ManagingHostsDocURL: {BaseURL}Managing_Hosts/index-foreman.html#
+:ConfiguringAnsibleDocURL: {BaseURL}Configuring_Ansible/index-foreman.html#
+:BaseURL: https://access.redhat.com/documentation/en-us/red_hat_satellite/{ProductVersion}/html/
+// Remove the above BaseURL attribute when work with URLs is ready. This attribute redefines the BaseURL attribute to ensure that links that only use {BaseURL} work, we won't need that when all links use a specific-to-guide attribute instead
```
lzap, line 103: So this hack is there to be able to pass link validator, or what is the reason?

lzap:
I wonder what if we choose to use relative URLs. If every content is in the same level, then simply `../../whatever` should always work no matter if it's upstream or downstream.

spetrosi:
> I wonder what if we choose to use relative URLs. If every content is in the same level, then simply `../../whatever` should always work no matter if it's upstream or downstream.

Some guides have different titles up- and downstream, and hence have different relative URLs. For example, the Install guides have the following relative URLs:
https://docs.theforeman.org/master/Installing_Server_on_Red_Hat/index-foreman.html
https://access.redhat.com/documentation/en-us/red_hat_satellite/6.8/html/installing_satellite_server_from_a_connected_network/index
Therefore, it is not possible to use relative URLs in all cases.

guides/common/attributes.adoc
spetrosi: You are correct. But not only to pass the validator, more like to keep the links working. Many links use the BaseURL attribute now, so I cannot simply remove it. I will work on all links one by one and in the end all links would use the relative attribute, e.g. `ConfiguringAnsibleDocURL` in them, so it then would be possible to remove the BaseURL attribute without any issue.

guides/common/attributes.adoc
```diff
@@ -97,6 +96,11 @@ ifeval::["{build}" == "foreman"]
 :ansible-namespace: `theforeman.foreman`
 :ansible-namespace-example: `theforeman.foreman._module_name_`
 :awx: AWX
+:BaseURL: https://docs.theforeman.org/master/
```
melcorr, line 99: Instead of master could we use {ProductVersion}, then define that as master upstream for now?
I think that if we are going to get serious about replacing Foreman manuals, we need to start branching.
I think that we should think seriously about this in the next month because we need the process worked out before @spetrosi leaves the docs team.

lzap: Let's start branching after we migrate installation guide. I was hoping to get to it soon.

lzap: We can modify `master` later tho, I won't block the PR.

lzap (approved):
Since the attribute overriding is temporary I am fine with the changes as long as HTML validator passes.

lzap:
@melcorr any other comments? Let's move forward?

guides/common/attributes.adoc
melcorr: Yeah for sure, but as part of this PR I'd like {ProductVersion} used to define the upstream version, so that it is then set and we don't have to modify that again.

spetrosi: Let's do this as part of the issue #322 after the issue #192 is resolved. I plan to finish with #192 by the end of January.

melcorr: OK.

spetrosi: lol I didn't reload this page for hours and did not see your comments prior to adding mine. The issue is that ProductVersion is now defined globally for all builds, so we will need to move this to each builds' attributes. Then, there is a bunch of similar attributes like ProductVersionPrevious, that are very related and that I want to keep all in one place, so we will need to move the altogether and find new names for them in each build. This is a completely different work from what this PR relates to. Let's shoot at all those moving targets one by one rather than fixing them all here.

melcorr: I'm happy with this as long as they are defined somewhere, like in Issues, so we can look at the upstream implications also.

melcorr:
> @melcorr any other comments? Let's move forward?

Nope, the issue around using an attribute for `master` was all that struck me for this PR.

guides/common/attributes.adoc
spetrosi: Thanks, @melcorr, can you merge this PR now that it is approved?

PR #323, opened 2020-12-09
Add a note about DNS for provisioning

lzap:
It is important that DNS is working both on Server and Smart Proxy.


melcorr:
Thank you for your contribution!

PR #325, opened 2020-12-18
Correct HTTP Boot proxy requirement

lzap:
I don't know why we have TFTP there, probably a copy paste problem.


melcorr:
Thanks! Probably my bad!

PR #332, opened 2021-01-05
Fixes 1911570 - SELinux in disabled is not supported

lzap:
SSIA we don't test this

### DIFF
--- a/guides/common/modules/ref_system-requirements.adoc
+++ b/guides/common/modules/ref_system-requirements.adoc
@@ -62,7 +62,7 @@ endif::[]
 
 ifeval::["{build}" != "foreman-deb"]
 .SELinux Mode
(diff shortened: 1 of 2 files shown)


guides/common/modules/ref_system-requirements.adoc
```diff
@@ -61,6 +61,8 @@ ifeval::["{build}" == "satellite"]
 endif::[]
 
 ifeval::["{build}" != "foreman-deb"]
+.SELinux Mode
+SELinux must be in either enabled or permissive mode, installation with disabled SELinux is not supported.
```
ekohl, line 65: I think both enforcing and permissive are considered enabled. How about this?
```suggestion
SELinux must be enabled, either in enforcing or permissive mode. Installation with disabled SELinux is not supported.
```

lzap:
Sure, please squash on commit.

PR #333, opened 2021-01-05
Fixes #1911299 - only postfix is tested email server

lzap:
We don't test others, gray-listing is usually causing issues these days.

### DIFF
--- a/guides/common/modules/proc_configuring-satellite-for-outgoing-emails.adoc
+++ b/guides/common/modules/proc_configuring-satellite-for-outgoing-emails.adoc
@@ -4,6 +4,10 @@
 
 To send email messages from {ProjectServer}, you can use either an SMTP server, or the `sendmail` command.
 
+.Prerequisites
+
+* Some SMTP servers with anti-spam protection or grey-listing features are known to cause problems. To setup outgoing email with such a service either install and configure a vanilla SMTP service on {ProjectServer} for relay or use the `sendmail` command instead.
+
 .Procedure
 
 . In the {Project} web UI, navigate to *Administer* -> *Settings*.


guides/common/modules/proc_configuring-satellite-for-outgoing-emails.adoc
```diff
@@ -5,6 +5,7 @@ To send email messages from {ProjectServer}, you can use either an SMTP server,
 
 .Prerequisites
 
+* Postfix from {RHEL} is the recommended email server to use. Some SMTP servers with anti-spam protection or grey-listing features are known to cause problems, to setup outgoing email with such service either install and configure postfix on {ProjectServer} for relay or use `sendmail` command instead.
```
ekohl, line 8: Can you cite a source on this? Note that Debian ships exim by default.

guides/common/modules/proc_configuring-satellite-for-outgoing-emails.adoc
```diff
@@ -5,6 +5,7 @@ To send email messages from {ProjectServer}, you can use either an SMTP server,
...
 * If you have upgraded from a previous release, rename or remove the configuration file `/usr/share/foreman/config/email.yaml` and restart the `httpd` service. For example:
```
ekohl, line 9: Unrelated, this "previous release" was Foreman 1.15 and 1.16 removed it. Since https://github.com/theforeman/puppet-foreman/commit/e6ead4595d3008040e56a55cc40a424ee5cb87d7 the installer removes it. No user will ever see this. I opened https://github.com/theforeman/foreman-documentation/pull/334 to get rid of it.

guides/common/modules/proc_configuring-satellite-for-outgoing-emails.adoc
lzap: The context is: https://bugzilla.redhat.com/show_bug.cgi?id=1911299

Red Hat QA do actually perform regression testing when Satellite is configured via local Postfix. Exim will work fine as long as some anti-spam protection isn't configured. I cannot be specific what exactly can cause issues tho, the purpose of this change is to inform user that it's safest to get an email from Foreman into a proper SMTP agent as soon as possible so it can be delivered properly.

lzap:
Ok I rebased this to informative: _Some SMTP servers with anti-spam protection or grey-listing features are known to cause problems, to setup outgoing email with such service either install and configure a vanilla SMTP service on {ProjectServer} for relay or use `sendmail` command instead._

guides/common/modules/proc_configuring-satellite-for-outgoing-emails.adoc
```diff
@@ -3,6 +3,10 @@
 
 To send email messages from {ProjectServer}, you can use either an SMTP server, or the `sendmail` command.
 
+.Prerequisites
+
+* Some SMTP servers with anti-spam protection or grey-listing features are known to cause problems, to setup outgoing email with such service either install and configure a vanilla SMTP service on {ProjectServer} for relay or use `sendmail` command instead.
```
ekohl, line 8: Minor grammar suggestions.

```suggestion
* Some SMTP servers with anti-spam protection or grey-listing features are known to cause problems. To setup outgoing email with such a service either install and configure a vanilla SMTP service on {ProjectServer} for relay or use the `sendmail` command instead.
```

Also, doesn't the sendmail command rely on a local mailserver already?

ekohl (commented):
I'd like it if someone could make sure my grammar suggestions are actually correct. Otherwise :+1:

guides/common/modules/proc_configuring-satellite-for-outgoing-emails.adoc
melcorr: Just need a comma before either :)

lzap: > doesn't the sendmail command rely on a local mailserver already?

Well it does as far as I know, but the gist of my message is: don't configure against a remote enterprise-configured (tightened) SMTP. Use a local one and let it do the job of the delivery, not Ruby.

Amended, commas - I don't use so no idea where exactly to put them ;-)

ekohl (approved):

lzap:
Thanks, merged.

PR #334, opened 2021-01-05
Remove outdated email.yaml references

ekohl:
Foreman 1.16 dropped support email.yaml support. It no longer reads this file so the instructions are completely useless. There never was a reason to restart httpd since the process never read it anyway.

Since Foreman 1.20 the installer also removes the outdated email.yaml. No user will on a supported version will ever see this file.


melcorr:
Thank you @ekohl

PR #354, opened 2021-01-15
Edit ilnks to the Configuring with Load Balancer guide

spetrosi:
Issue #192


melcorr (approved):
Short and sweet. This works

Issue #326, opened 2020-12-29
Using one sentence per line


maximiliankolb:
Hi!
I want to make a proposal to simplify PRs and code reviews: sticking to **one sentence per line** which makes handling `git diff` much easier. Doing so does not change the output, as asciidoc does not handle

```
sentence 1. sentence 2. the whole paragraph on one line :(
```

any different to

```
sentence 1.
sentence 2 on a new line.
:)
```

as long as there are no empty lines in between.

If you allow me, I will create PRs doing exactly this, chapter by chapter. We at ATIX have been doing this with great success.

Stay safe everyone,
Maximilian.

melcorr:
@spetrosi I just had @maximiliankolb explain this to me, and reading the proposal, it has no impact to the output, but makes reviewing easier. I think that this would be a positive improvement. What do you think? Ideally we find some way to script this so that we don't have to manually add linebreaks to every existing doc. 
What do you think @lzap

maximiliankolb:
I've converted the two guides as suggested. The second one was a bit of a pain, as there are many files included ...
I did this by hand, as I see no way to (quickly) automate this. There are many edge cases; I feel like going chapter by chapter is just fine.
Disclaimer: I did not touch any tables yet and might have missed some sentences starting after a colon or semicolon.

lzap:
Well, this will break the readability of the source a bit, but there is a clear value in better diffs. I don't mind this, I'd probably prefer something completely different myself: no newlines for sentences and hard wrapping at 79th column. But that's me, not all contributors do have powerful text editor with hard wrapping options.

If we are landing on this as the standard, I suggest to formalize this in the README:

* No wrapping.
* Every sentence on a new line.
* UTF-8 blah blah

maximiliankolb:
My plan was to convert all guides and add a second commit to the last PR to amend the `README` accordingly.

lzap:
Great, thanks for your contributions!

PR #361, opened 2021-01-25
Convert to one sentence per line

maximiliankolb:
This PR converts the "Administering Red Hat Satellite" guide to follow the one sentence per line idea.
See #326

One table is excluded by choice & will be fixed at a later stage.


maximiliankolb:
Thanks @spetrosi !

PR #365, opened 2021-01-27
Change link to point to the new KCS solution for port matrix

spetrosi:
Bug 1896012 - [RFE] Include a comprehensive overview of connections

https://bugzilla.redhat.com/show_bug.cgi?id=1896012


lzap (approved):
Both are redirects, looks good to me.

PR #367, opened 2021-01-28
Proposal: Add plugin documentation skeleton

maximiliankolb:
Hey all.

We're currently evaluating adding the plugin documentation of [Foreman ACD](https://github.com/ATIX-AG/foreman_acd) to the upstream documentation. What do you think about this skeleton as a base for plugin documentation?

We basically propose to have a new directory called `plugins`, which in turn contains the plugin documentation like `doc-foreman_acd`, similarly to all the other usage guides.

This is still early work; we haven't fully made up our mind yet. But seeing the [thread](https://community.theforeman.org/t/volunteer-needed-documentation-conversion/22121) by @lzap made us go forward with this.

Looking forward to hearing your feedback.
Thanks and stay safe
Maximilian.

### DIFF
--- a/plugins/Makefile
+++ b/plugins/Makefile
@@ -1,0 +1,1 @@
+guides/Makefile
--- a/plugins/README.adoc
+++ b/plugins/README.adoc
@@ -1,0 +1,15 @@
+Foreman Plugins
+
+.Foreman Plugins
+|===
+|Plugin |Compatible to | Maintained by
+
+|https://github.com/username/project[Foreman plugin 1]
+|>= Foreman 2.1
+|https://example.com/[Example Company]
+
+|https://github.com/username/project[Foreman plugin 2]
+|>= Foreman 2.3
+|https://example.com/[Example Individual]
+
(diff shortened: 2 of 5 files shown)


melcorr:
This is contra to everything we are trying to do: create guides not plugin manuals - the next generation of Foreman docs.

lzap:
I am reviewing some unread notifications, just to prevent confusion: I am sure @melcorr is not against having the documentation here, it just needs to be integrated into current documentation. I think this could be a good fit for our Ansible guide: https://docs.theforeman.org/nightly/Configuring_Ansible/index-foreman-el.html

If you prepare the doc as a new chapter that will have information that the whole chapter applies to foreman_acd plugin, then I think it would be great to have it here. Just a technical note it must not be rendered with "satellite" context because Red Hat is not shipping this plugin at the moment.

Installation is easy (just enable the plugin, so it can live in the Ansible guide, or we can probably create individual chapters in the installation guide(s). What you think @melcorr ?

PR #368, opened 2021-01-28
Add the /var/spool/squid/ directory to Proxy storage reqs table

spetrosi:
Bug 1822391 - [DDF] Squid Storage Requirements are missing on Capsule
installation guide

https://bugzilla.redhat.com/show_bug.cgi?id=1822391

### DIFF
--- a/guides/common/modules/ref_capsule-storage-requirements.adoc
+++ b/guides/common/modules/ref_capsule-storage-requirements.adoc
@@ -14,5 +14,6 @@ The runtime size was measured with Red{nbsp}Hat Enterprise Linux 6, 7, and 8 rep
 |/var/cache/pulp/ |1 MB | 20 GB (Minimum)
 |/var/lib/pulp/ |1 MB |300 GB
 |/var/lib/mongodb/ |3.5 GB |50 GB
+|/var/spool/squid/ |0 GB |10 GB
 |/opt | 500 MB | Not Applicable
 |====


melcorr:
> Here it gets complicated: in nightly this directory is no longer really used. We just dropped Pulp 2 as active content. It's now just there on the main Katello machine for katello-agent. That means it really only should be applied to stable branches. upstream documentation doesn't have that.

I think we're at the point where we really need to figure this out. 
> 
> On the other hand, `/var/lib/mongodb` and `/var/cache/pulp` are also irrelevant in nightly. Also, the PostgreSQL database with Pulp 3 will grow so that needs to be planned for as well. That means you may want to merge this now, but update it right after for the new Pulp 3 world.

This is fun. 
IMHO, the upstreamized versions of the Sat installation guides are largely worthless to the community. What do you think?

ekohl:
I should note that it was just merged hours ago, right in time for branching. That's why I kept insisting on a branching structure in this repository. If you have that, you can take version x.y, apply the change there and cherry pick it into downstream. At the same time, you can continuously update the documentation for the next release.

melcorr:
> I should note that it was just merged hours ago, right in time for branching. That's why I kept insisting on a branching structure in this repository. If you have that, you can take version x.y, apply the change there and cherry pick it into downstream. At the same time, you can continuously update the documentation for the next release.

We need to add a readme explaining branching and start branching both with Foreman and Katello releases in mind. 
We also should have some doc that we maintain with info on plugin versions, as per your comment on Discourse, which I will reply to later. 
I'm +1 branching, but we just need to write it up formally so that the new TWs and other contributors can understand what they need to do.

suyogsainkar:
thanks @spetrosi looks good to me; merging.

suyogsainkar (approved):

PR #370, opened 2021-02-01
Add version to upstream titles

lzap:
We have a branch support online now:

https://docs.theforeman.org/2.3/Planning_Guide/index-foreman.html

However it's not clear what guide version are you looking at. This introduces version number upstream to all titles, downstream there is a subtitle inserted but we don't do this workflow upstream. This will do it.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -57,12 +57,6 @@ jobs:
           make html BUILD=foreman-deb
           make html BUILD=satellite
 
-      - name: List files
-        run: find build
-
-      - name: Remove Last Modified
-        run: sed -i -E 's/^Last updated ([[:digit:]]+-[[:digit:]]+)-[[:digit:]]+ [[:digit:]]+:[[:digit:]]+:[[:digit:]]+ UTC$/Last updated \1/' build/*/*.html
-
       - name: Upload HTML
         uses: actions/upload-artifact@v2
         with:
@@ -152,7 +146,7 @@ jobs:
           path: web/public/
 
   publish:
-    if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/2.3'
+    if: github.repository == 'theforeman/foreman-documentation' && startsWith(github.ref, 'refs/heads/(master|[0-9]+\.[0-9]+)')
     needs:
       - build-html
       - build-web
--- a/guides/common/attributes.adoc
+++ b/guides/common/attributes.adoc
@@ -1,5 +1,10 @@
-:ProjectVersion: 2.0
-:KatelloVersion: 3.15
+// NOTE: Avoid empty lines so it can be included in the header section
+// Document state: "nightly" for master, "stable" for last two releases,
+// "unsupported" for the rest and "satellite" for satellite build
+:DocState: nightly
+// Versions used in text and code
+:ProjectVersion: 2.3
+:KatelloVersion: 3.18
 :TargetVersion: 6.8
 :TargetVersionMaintainUpgrade: 6.8
 // The above attribute should point to the GA version number (x.y) for all releases including Beta
@@ -17,7 +22,6 @@
 //Do not update the puppet4 repo versions. They must stay at 6.3.
 :SatelliteSub: Red Hat Satellite Infrastructure Subscription
 // Change to "Red Hat Satellite Infrastructure Subscription (Beta)" for beta releases
-
 :RepoRHEL7Server: rhel-7-server-rpms
 :RepoRHEL7ServerSoftwareCollections: rhel-server-rhscl-7-rpms
 :RepoRHEL7ServerOptional: rhel-7-server-optional-rpms
@@ -25,10 +29,9 @@
 :SatelliteAnsibleVersion: 2.9
 :SpecialCaseProductVersion: 6.8
 //the above attribute is for Package Manifests etc that will fail the upstream link-checker during the beta see Issue #115 on GitHub
-
 ifeval::["{build}" == "foreman"]
-:project-installation-guide-title: Installing Foreman server
-:smart-proxy-installation-guide-title: Installing an External Smart Proxy Server
+:project-installation-guide-title: Installing Foreman {ProjectVersion} server
+:smart-proxy-installation-guide-title: Installing an External Smart Proxy Server {ProjectVersion}
 :smart-proxy-context: smart-proxy
 :project-context: foreman
 :foreman-installer: foreman-installer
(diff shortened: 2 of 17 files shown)


lzap:
The PDF version looks a bit ugly, I think maybe I can put the version below as a new title or subtitle?

image

lzap:
There appears to be some meta information which renders to the header and footer, but I can't get `{ProjectVersion}` variable working there. Weird. https://docs.asciidoctor.org/asciidoc/latest/document/header/

guides/common/attributes.adoc
```diff
@@ -1,5 +1,5 @@
-:ProjectVersion: 2.0
-:KatelloVersion: 3.15
+:ProjectVersion: 2.3
+:KatelloVersion: 3.18
```
ekohl, line 2: This is the nightly version so shouldn't this be 2.4 and Katello 4.0?

lzap: Yes I will add a new commit once we agree, I want to merge this into 2.3 branch.

lzap:
Ok I figured out how to put metadata into the AsciiDoc header, now it renders nice both for HTML and PDF. It even puts version into footer in the end of the HTML version:

image

lzap:
I rebased, let me know what you think about "Version X.Y, published YEAR". I wanted to signal clearly to the reader that it might be quite old version if they happen to randomly land on an older version from Google.

guides/common/attributes.adoc
ekohl: FYI: nightly is about to bump to 2.5. Would you also branch 2.4?

melcorr: From what I understand, Katello releases come slightly after a Foreman release. Is it possible to have a branch, for example, for Foreman 2.4 & Katello 3.18 AND Foreman 2.4 and Katello 4?

ekohl: No, we only had 1 example where there was Katello 3.9 and Katello 3.10 on a single Foreman release but I don't really see that happening again. The biggest difference between 3.9 and 3.10 was EL8 support because it missed the 3.9 deadline. Overall it was a painful experience. I certainly hope we can avoid it.

melcorr: Ok, even better. Less complications. Thank you @ekohl

lzap:
So? What do you guys say? Shall I add a new commit on top of this one with the latest versions and shall we merge this? I'd like to backport this into 2.3 so it's online.

melcorr:
If @ekohl is Ok with this, it LGTM.

guides/doc-Administering_Red_Hat_Satellite/master.adoc
```diff
@@ -1,11 +1,13 @@
+include::common/attributes.adoc[]
 :imagesdir: images
 :administering-red-hat-satellite:
 :numbered:
 ifeval::["{build}" != "satellite"]
 :toc:
 :toc-placement: left
+:revnumber: {ProjectVersion}
+:revdate: published {docyear}
```
ekohl, line 9: Is a year sufficient? Perhaps it needs to be a date?

ekohl (approved):
There does appear to be a lot of duplication but overall I think this makes sense.

guides/doc-Administering_Red_Hat_Satellite/master.adoc
lzap: Date is at the end of each page (PDF version does not have it). I can add month but I would like to avoid full date.

In the github action, I actually `sed` date in the format of YYYY-MM-DD HH:MM to simply YYYY-MM because I don't think readers need to know exact date. It creates a change every time we commit into `gh-pages` and I would like to have sane history.

Alternatively, I can introduce some "nightly" variable which will make this to only appear on released branches. As long as we won't forget to flip it that could be the best solution.

lzap: That variable might be actually a good idea - we could leverage it with values "nightly, stable, old" and utilize it to render some warnings for old and nightly content.

ekohl: I think it makes sense to have smaller changes, but as a reader it implies it's only published yearly. At least a PDF gives that impression.

In our current manual we also have something like "this isn't the latest version, see $stable" and "this is nightly, see $stable for a supported version". Having a similar thing here does make sense to me. Perhaps even something more complex: note when it's supported. Currently Foreman 2.2 and 2.3 are supported, but 2.1 and earlier can get a big warning that the release is end of life. Especially for an update guide that's great to have. Nobody should install an EOL release unless they know what they're doing.

lzap:
Ok here is another version. Much more clean now.

When built as "foreman" or "foreman-deb" with "nightly" attribute the header reads `Version Nightly, published Feb 2021` and the footer reads `Version Nightly` and on the second line `Last updated Feb 2021`. With "stable" attribute it is the same except there is a version instead "Nightly" and the date is `Jan 15 2021`. And unsupported is the same except version reads `Version 2.1 (unsupported)`. When built as "satellite" nothing is added, versioning is done differently downstream.

.github/workflows/deploy.yml
```diff
@@ -57,12 +57,6 @@ jobs:
           make html BUILD=foreman-deb
           make html BUILD=satellite
 
-      - name: List files
-        run: find build
-
-      - name: Remove Last Modified
-        run: sed -i -E 's/^Last updated ([[:digit:]]+-[[:digit:]]+)-[[:digit:]]+ [[:digit:]]+:[[:digit:]]+:[[:digit:]]+ UTC$/Last updated \1/' build/*/*.html
-
```
lzap, line 65: This hack is no longer needed since `Last updated` is now rendered correctly due to `-a docdatetime="$(DATE_MY)"` `Makefile` change.

guides/doc-Administering_with_Ansible/master.adoc
```diff
@@ -1,11 +1,22 @@
...
+:revdate: published {date_my}
+endif::[]
+ifeval::["{DocState}" == "stable"]
+:revnumber: {ProjectVersion}
+:revdate: published {date_mdy}
+endif::[]
+ifeval::["{DocState}" == "unsupported"]
+:revnumber: {ProjectVersion} (unsupported)
+:revdate: published {date_mdy}
+endif::[]
```
ekohl, line 19: Is there a way to avoid this duplication everywhere? Perhaps via an include?

lzap:
Note to myself: Edit Github Action and modify this line if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/2.3' to include the new version. Only last two stable versions are needed.

This should be done in the branch. Maybe we can find a generic way, regexp like X.Y @ekohl if you have an idea.

guides/doc-Administering_with_Ansible/master.adoc
lzap: How I could miss this! Of course.

lzap:
Ok @ekohl helped me to figure out better solution, no need to update GH Action if it works. Let's see.

lzap:
All green, I am unable to test if deployment now works so let's just merge this if there are no objections and I will fix it in the branch if needed.

PR #376, opened 2021-02-04
Update versions to 2.4/4.0

lzap:
The next major version pair 2.4/4.0.

I am still having issues with that if statement, looks like the startsWith does not recognize that glob. It can only be used in the "on" statement. I wonder if I remove the test and only test owner if this is what I want:

* it triggers only for master and X.Y branches
* for pull requests artifacts are built
* for theforeman owner site is also deployed

Do not merge yet.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -146,7 +146,7 @@ jobs:
           path: web/public/
 
   publish:
-    if: github.repository_owner == 'theforeman' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/[0-9]+.[0-9]+'))
+    if: github.repository_owner == 'theforeman' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/2.') || startsWith(github.ref, 'refs/heads/3.') || startsWith(github.ref, 'refs/heads/4.'))
     needs:
       - build-html
       - build-web
--- a/guides/common/attributes.adoc
+++ b/guides/common/attributes.adoc
@@ -3,8 +3,8 @@
 // "unsupported" for the rest and "satellite" for satellite build
 :DocState: nightly
 // Versions used in text and code
-:ProjectVersion: 2.3
-:KatelloVersion: 3.18
+:ProjectVersion: 2.4
+:KatelloVersion: 4.0
 :TargetVersion: 6.8
 :TargetVersionMaintainUpgrade: 6.8
 // The above attribute should point to the GA version number (x.y) for all releases including Beta


guides/common/attributes.adoc
```diff
@@ -4,7 +4,7 @@
 :DocState: nightly
 // Versions used in text and code
 :ProjectVersion: 2.3
```
ekohl, line 6:
```suggestion
:ProjectVersion: 2.4
```

ekohl (changes requested):
Btw, I do think we should branch 2.4 and keep master as 2.5 to match Foreman nightly.

guides/common/attributes.adoc
lzap: Yeah this was an overlook. I wanted to do this change.

.github/workflows/deploy.yml
```diff
@@ -146,7 +146,7 @@ jobs:
           path: web/public/
 
   publish:
-    if: github.repository_owner == 'theforeman' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/[0-9]+.[0-9]+'))
+    if: github.repository_owner == 'theforeman'
```
lzap, line 149: Deploy was executed for this PR and it failed. I struggle to find the correct condition. I want the deploy job to be executed if:

* this was not a PR
* owner of the github repo is "theforeman"
* branch is master or X.Y (this can't be done at the moment I will do a workaround for this)

lzap:
I read the github actions documentation over and over again yet I don't know how to setup what I want.

For all PRs in this repo, I want all artifacts to be built (so we test they do build).

For all pushes into "master" and "X.Y" I want all artifacts to be built + deployed.

I am not sure if the correct solution would be to split the YAML into two, but the problem is that deploy depends on all the three build jobs.

.github/workflows/deploy.yml
ekohl: > this was not a PR

Looking at https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#contexts and in particular https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context I'd consider looking at `event_name` and check whether it's push.

lzap:
Now it builds fine, no deployment. It's terrible but I can't figure out better solution.

ekohl (approved):
I'm going to merge this so we can branch 2.4. Then master should be updated to track nightly which means 2.5/4.1.

.github/workflows/deploy.yml
```diff
@@ -146,7 +146,7 @@ jobs:
           path: web/public/
 
   publish:
-    if: github.repository_owner == 'theforeman' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/[0-9]+.[0-9]+'))
+    if: github.repository_owner == 'theforeman' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/2.') || startsWith(github.ref, 'refs/heads/3.') || startsWith(github.ref, 'refs/heads/4.'))
```
ekohl, line 149: I've said it before, but I think we should follow Foreman's convention. That means `develop` for nightly and `x.y-stable` for stable branches. Then you can check it with an `endsWith` instead of a number.

lzap: Well it looks like the glob syntax does not work with startsWith despite some advice on stackoverflow.com.

PR #380, opened 2021-02-08
Add vendor neutral provisioning user data sequence

maximiliankolb:
This PR replaces the "provisioning user data sequence" graphic with a more vendor neutral one.


melcorr:
@maximiliankolb  would you mind adding an != satellite & == satellite ifeval here because Red Hat will want the less vendor neutral one ;-)

maximiliankolb:
Fixed and rebased.

lzap (approved):
Nice, SVG renders fine both in HTML and PDF. Maybe we can use SVGs directly everywhere from now on, is there a requirement that all images have to be PNG, @melcorr ?

guides/doc-Provisioning_Guide/topics/Virt-VMWare.adoc
```diff
@@ -288,7 +288,12 @@ You can use the same procedures to set up a VMware compute resource within {Proj
 
 .VMware cloud-init Provisioning Overview
 
+ifeval::["{build}" == "satellite"]
 image:ProvisioningUserData-Sequence.png[]
+endif::[]
+ifeval::["{build}" != "satellite"]
+image:provisioning_user_data_sequence.svg[]
+endif::[]
```
lzap, line 296: One thing - we are setting a standard here - we should probably figure out some naming convention to distinquish upstream and satellite. How about `user-data-sequence-foreman.svg` and `user-data-sequence-satellite.{svg,png}`?

Then we should probably update the README with some instructions about how to properly name diagramy. Would you mind doing that?

melcorr:
> Nice, SVG renders fine both in HTML and PDF. Maybe we can use SVGs directly everywhere from now on, is there a requirement that all images have to be PNG, @melcorr ?

Those illustrations are done by RH Design team who seem to have a particular aesthetic. I'm not sure if we can request svgs if I'm honest. There is a uniformity throughout all of the RH images, so they have some style guide. They also do not share the source so we cannot edit those files ;-)

guides/doc-Provisioning_Guide/topics/Virt-VMWare.adoc
melcorr: This would be the behaviour of a group following a standard! +1 from me.

melcorr (approved):
I like this. It looks lovely. 
If you want to change to the suggested convention, that is up to you. 
We could also discuss the convention and rename at a later stage. 
I'm just glad to have these drawings.

guides/doc-Provisioning_Guide/topics/Virt-VMWare.adoc
maximiliankolb: I like the idea of simply suffixing `-foreman` and `-satellite`. I prefer underscores but can live with dashes, too. 

I've reworked this PR and added a new convention.

guides/README.md
```diff
@@ -128,6 +128,7 @@ This makes `git diff` much easier to read and helps reviewing changes.
 * No trailing whitespace on lines and in files.
 Whitespace after partial files has to be handled in the file using the `include::` directive.
 * Source files use UTF-8 character encoding.
+* Image file names use dashes (`-`) and suffix a build target, e.g. `foreman`.
```
lzap, line 132: Nice, we could drop a note that we prefer SVG file format for upstream and PNG for downstream? I am actually not sure, SVG works upstream and it's smaller and don't need to have a PNG copy. On the other hand, you probably need more modern browser to open them... like from 2015+ or something. Not big deal.

guides/doc-Provisioning_Guide/topics/Virt-VMWare.adoc
lzap: Dashes was just an idea, I don't mind I just saw the "Virt-VMWare.adoc" filename here and I thought that's the preferred way in this repo.

lzap:
Thanks.

guides/README.md
maximiliankolb: Good point. I can expand the bullet point with my next PR replacing graphics.

PR #382, opened 2021-02-09
Fix broken multiline ifdef directives

maximiliankolb:
When converting all source files to "one sentence per line", I probably broke two `ifdef` directives.

This should be fixed now; see `git grep -i -n "ifdef:" | grep -v "\]$"`: there are no more "ifdef" occurrences without a closed square bracket at the end of the line.


lzap (approved):
Thanks.

PR #381, opened 2021-02-08
Update version number in docinfo.xml to 6.9 Beta for the upstream guides

suyogsainkar:
Update attributes.adoc and docinfo files for 6.9
https://bugzilla.redhat.com/show_bug.cgi?id=1918485

PR #383, opened 2021-02-09
Bz#1918485 docinfo 6.9beta

suyogsainkar:
Cherrypicked from https://github.com/theforeman/foreman-documentation/pull/381

### DIFF
--- a/guides/doc-Administering_Red_Hat_Satellite/docinfo.xml
+++ b/guides/doc-Administering_Red_Hat_Satellite/docinfo.xml
@@ -1,6 +1,6 @@
 <title>Administering Red Hat Satellite</title>
 <productname>Red Hat Satellite</productname>
-<productnumber>6.8</productnumber>
+<productnumber>6.9 Beta</productnumber>
 <subtitle>A guide to administering Red Hat Satellite.</subtitle>
 <abstract>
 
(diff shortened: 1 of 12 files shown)


melcorr:
Why is there a PR for a cherry pick?

lzap:
We actually tend to review cherry picks too in the code, it gives a better visibility.

But I guess it's not needed given the fact that those SATELLITE branches are not interesting for the Foreman community. Therefore I suggest CP PRs only for Foreman stable branches, e.g. `2.3`.

lzap:
By the way, I suggest you doing fast-forward merging as long as possible. Most of the upstream changes are still valid and you could actually be fast-forwarding until you need a downstream only merge. Once you start cherry picking, the history diverges and you need to keep cherry picking forever.

If you would have fast-fotwarded, you would still have this (still useful) fix: https://github.com/theforeman/foreman-documentation/commit/80b65d711813247a1d0d53291523eba7d48251c8

suyogsainkar:
@melcorr I create my own topic branch from the version branch (SATELLITE-6.9-beta in this case) and perform _git cherry-pick <commit-id>_ This workflow created this new PR when I pushed the cherry-picked commits. @spetrosi clarified this is redundant and I can directly work in the version branch for cherry-picking.

PR #385, opened 2021-02-10
new module for enabling insights

annibond:
new module for enabling insights on a capsule server

### DIFF
--- a/guides/common/modules/proc_installing-postgresql.adoc
+++ b/guides/common/modules/proc_installing-postgresql.adoc
@@ -79,7 +79,7 @@ $ su - postgres -c psql
 +
 [options="nowrap" subs="verbatim,quotes"]
 ----
-CREATE USER "foreman-el" WITH PASSWORD '_Foreman_Password_';
+CREATE USER "foreman" WITH PASSWORD '_Foreman_Password_';
 CREATE USER "candlepin" WITH PASSWORD '_Candlepin_Password_';
 CREATE DATABASE foreman OWNER foreman;
 CREATE DATABASE candlepin OWNER candlepin;


guides/common/modules/proc_enable-insights-with-capsule.adoc
```diff
@@ -0,0 +1,21 @@
+[id='register-capsule-with-insights_{context}']
+
+= Register a Capsule server with Red Hat Insights
```
melcorr, line 3: Please use the appropriate variable for Capsule server for the heading too. 
In Satellite docs and mod docs, and according to IBM style, we always use a gerund in our headings.

guides/common/modules/proc_enable-insights-with-capsule.adoc
```diff
@@ -0,0 +1,21 @@
+[id='register-capsule-with-insights_{context}']
+
+= Register a Capsule server with Red Hat Insights
+
+Red Hat Insights enables you to proactively diagnose systems and downtime related to security exploits, performance degradation and stability failures.
+You can use the dashboard to quickly identify key risks to stability, security, or performance.
```
melcorr, line 6: In Satellite docs, we followed IBM style guide, so we did not use phrases like "enables you" "allows you" etc, which would be classed as Anthropomorphism according to pg 8 of IBM Style.

melcorr (changes requested):
Hi @annibond welcome to docs. 

I see that you are pushing from the master of your fork. This is fine but it means that you'll have to reset the head of the your local master branch after this is merged to be able to pull in the latest changes to your fork from the upstream without errors. It's best to create a branch so you you can pull directly from your upstream into your master and keep it always up to date.. 

If you've any specific git questions, you can ask them here and someone will help.

PR #386, opened 2021-02-11
GlobalReg procedure: Selecting SmartProxy is not optional

swadeley:
Hello

1] As per subject, we need to mark that step as optional and mention having an AK in a HG also works.
2] If a user does not select the smart proxy to use, it defaults to the internal one and that might not be what the user expected and if the user connected without using FQDN there will be problems..So lets not say that step is optional.

Thank you

### DIFF
--- a/guides/doc-Managing_Hosts/topics/proc_registering-a-host-to-project-using-the-global-registration-template.adoc
+++ b/guides/doc-Managing_Hosts/topics/proc_registering-a-host-to-project-using-the-global-registration-template.adoc
@@ -56,7 +56,7 @@ Optional: If the *Registration* feature is not enabled on your {SmartProxy}, ent
 
 . From the *Operating System* list, select the operating system of hosts that you want to register.
 
-. Optional: If you want to register hosts through {SmartProxy}, from the *{SmartProxy}* list, select the {SmartProxy} to register hosts to.
+. From the *{SmartProxy}* list, select the {SmartProxy} to register hosts through {SmartProxy}.
 
 ifeval::["{build}" == "satellite"]
 . From the *Insights* list, select whether you want to register the hosts to Insights or not.
@@ -77,7 +77,7 @@ Therefore, do not delete, block, or change permissions of the user during the to
 If you keep this field blank, {Project} uses the default network interface.
 
 ifeval::["{build}" == "satellite"]
-. In the *Activation Key(s)* field, enter one or more activation keys to assign to hosts.
+. Optional: In the *Activation Key(s)* field, enter one or more activation keys to assign to hosts if you are not using a Host Group with an activation key.
 endif::[]
 
 ifeval::["{build}" != "satellite"]
--- a/guides/doc-Managing_Hosts/topics/proc_registering-a-host-to-project-using-the-global-registration-template.adoc
+++ b/guides/doc-Managing_Hosts/topics/proc_registering-a-host-to-project-using-the-global-registration-template.adoc
@@ -77,7 +77,7 @@ Therefore, do not delete, block, or change permissions of the user during the to
 If you keep this field blank, {Project} uses the default network interface.
 
 ifeval::["{build}" == "satellite"]
-. Optional: In the *Activation Key(s)* field, enter one or more activation keys to assign to hosts if you are not using a Host Group with an activation key.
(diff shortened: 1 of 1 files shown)


swadeley:
@lstejska@redhat.com please review

guides/doc-Managing_Hosts/topics/proc_registering-a-host-to-project-using-the-global-registration-template.adoc
```diff
@@ -56,7 +56,7 @@ Optional: If the *Registration* feature is not enabled on your {SmartProxy}, ent
 
 . From the *Operating System* list, select the operating system of hosts that you want to register.
 
-. Optional: If you want to register hosts through {SmartProxy}, from the *{SmartProxy}* list, select the {SmartProxy} to register hosts to.
+. From the *{SmartProxy}* list, select the {SmartProxy} to register hosts through {SmartProxy}.
```
spetrosi, line 59: Isn't it an optional step? Users need to do this only in the case when they want to register through Capsule, isn't it? I would keep the `Optional:` part in this step to avoid confusion.

swadeley: IIUC, the internal Capsule will be selected if they do not select any Capsule, and that can cause problems in some cases e.g. where they connected using an IP not a FQDN

spetrosi: Oh, so one can select the internal Capsule. Yup, then this field is always required.

spetrosi:
```suggestion
. From the *{SmartProxy}* list, select the {SmartProxy} to register hosts through. You must select the internal {SmartProxy} if you do not want to use an external {SmartProxy}.
```
How about making it explicit like that?

swadeley: I think lstejska wants to make it default to that setting in the web UI, not just in the code, so that it is clear what will happen if you do nothing. In which case "You must" would be a little strong. Lets wait now for lstejska to give feedback.

Thank you

stejskalleos: > that can cause problems in some cases e.g. where they connected using an IP not a FQDN.

To be honest we didn't count with the users that connect to Satellite via IP and not a FQDN,
there is an open issue for it: https://projects.theforeman.org/issues/31821

So my advice would be to select Capsule every time, just in case.

guides/doc-Managing_Hosts/topics/proc_registering-a-host-to-project-using-the-global-registration-template.adoc
```diff
@@ -77,7 +77,7 @@ Therefore, do not delete, block, or change permissions of the user during the to
 If you keep this field blank, {Project} uses the default network interface.
 
 ifeval::["{build}" == "satellite"]
-. In the *Activation Key(s)* field, enter one or more activation keys to assign to hosts.
+. Optional: In the *Activation Key(s)* field, enter one or more activation keys to assign to hosts if you are not using a Host Group with an activation key.
```
stejskalleos, line 80: Activation key is always required, without it registration will fail:
```
Complete!
Error: Must specify an activation key
```

Getting AKs from host group will be supported in 6.10

guides/doc-Managing_Hosts/topics/proc_registering-a-host-to-project-using-the-global-registration-template.adoc
swadeley: Thank you

guides/doc-Managing_Hosts/topics/proc_registering-a-host-to-project-using-the-global-registration-template.adoc
swadeley: OK, I have removed that change, thank you

swadeley:
Seems I cannot force push to remove the early git commit, so I will close and make new PR

swadeley:
see https://github.com/theforeman/foreman-documentation/pull/388

PR #390, opened 2021-02-12
Drop all mongo references

ehelms:


jlsherrill (approved):

spetrosi:
Hi @ehelms, is this related to Satellite? In what version of Satellite docs we want this change to appear?

ekohl:
@spetrosi this follows current nightly which is the Satellite version that's in development. It's also a major change. This is why I strongly emphasized that this repo needed a branching strategy. Luckily that was resolved recently.

spetrosi:
@ekohl will this go to Satellite 6.10?
The branching strategy is indeed cool.

ehelms:
Aye, this is for 6.10

lzap:
> The branching strategy is indeed cool.

The docs team rushed creating SAT branches and I think we should start over. This is a mess.

Ideal solution:

* github
  * master
  * X.Y (stable branches)
* gitlab (fork)
  * SATELLITE-X.Y (branches of X.Y from upstream)

For some reason, docs team insisted they don't want to introduce another git repository, I think we can live with SATELLITE-X.Y branches on github, probably not big deal, but they needs to be branched off from the correct upstream branches for sure. We can't cherry pick everything.

ekohl:
>     * X.Y (stable branches)

We can be a bit more explicit and follow Foreman which has `X.Y-stable` (see https://github.com/theforeman/foreman/branches). However, that's minor.

> For some reason, docs team insisted they don't want to introduce another git repository, I think we can live with SATELLITE-X.Y branches on github, probably not big deal, but they needs to be branched off from the correct upstream branches for sure. We can't cherry pick everything.

IMHO we should not have downstream names in upstream. If the docs team doesn't want another repository, they should remember to use the right X.Y-stable.

We can't rely on SATELLITE-X.Y because downstream always skips an upstream version. If foreman-documentation is supposed to become the source for documentation, we need more versions. Or explicitly, there is no downstream representation of Foreman 2.4 but upstream we will need that. Having both 2.3-stable and SATELLITE-6.9 is also confusing. Mixing SATELLITE-6.9 and 2.4-stable (but not 2.3-stable) is also a recipe for disaster.

> I suggest to start over, delete all of them and use fast-forward as much as possible. Most of the changes are fast-forwardable until we introduce documentation for a new feature and I don't see any.

:+1: to this. I noticed the same thing.

PR #393, opened 2021-02-12
Change foreman build to foreman-el to better identify enterprise linu…

ehelms:
…x installs


ehelms:
I'd like to have this change prior to introducing a katello build as I do here -- https://github.com/theforeman/foreman-documentation/pull/394

That way we can more easily pull apart Foreman on EL vs Deb vs Katello specifics

melcorr:
Cool @ehelms - will get to this today.

melcorr:
Hi @ehelms, 

I am testing this and am not finding problems except that there are missed "foreman" that haven't been converted to "foreman-el"
They are scattered throughout. 
Would it be useful for me to list each one?

melcorr:
I wonder if the problem is that you ran the sed over only the common/modules folders, while not all of the guides have been modularised yet, so content exists under guides/<guide-name>/topics/*.adoc for some of the guides. For example, the provisioning guide.

ehelms:
I must have been in a sub-directory when I ran my grep. Updated now and double checked myself.

melcorr (approved):
I can't find any issue with this now, thanks for implementing the change, @ehelms 
I would just like to clarify one point because I'm unsure if it was discussed anywhere:
Are we going to start generating Debian/other OS versions of each guide?

ehelms:
I think we are already generating different versions at least for install based on https://docs.theforeman.org/ -- the bigger change, that starts to bring this all into a more easy to manage view point of foreman EL vs Debian vs applies to both vs Katello is the change here: https://github.com/theforeman/foreman-documentation/pull/394

lzap:
This change actually broke our site in a way that we have a welcome page that links `index-BUILD_TAG.html` pages:

    https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman.html

After the change it should be:

    https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-el.html

We haven't noticed because currently the deploy script only copies new files to gh-pages branch, it does not delete old content. We should probably update this, but we need to make sure we do not delete all files in `gh-pages`, e.g. when deploying into `/nightly` we must only remove `/nightly/*` and not stable directories like `/2.3`.

Can you update our web in `web/` directory? It's hugo generated, you can easily figure out which file to change.

lzap:
For the record, we keep files because the action we use only allows to keep all or delete all:

```yaml
        - name: Deploy
          uses: peaceiris/actions-gh-pages@v3
          with:
            github_token: ${{ secrets.GITHUB_TOKEN }}
            publish_dir: ./public
            keep_files: true
```

ehelms:
We thought that is what this file handles -- https://github.com/theforeman/foreman-documentation/blob/master/web/layouts/shortcodes/guide-list.html  

Was in incorrect in updating that for these new identifiers? Or are you saying that it did not replace that file with this updated one?

lzap:
The deploy action is probably broken, cos I touched it yeah, and I need to fix this: 1896b2cf4850df2203f4e1b5bc65ae09bf19a517

lzap:
Ok I found out it's not working, need to take a look tomorrow: https://github.com/theforeman/foreman-documentation/issues/406

PR #394, opened 2021-02-12
Add katello install

ehelms:
This switches to a model where build sets a "variable". By setting a variable, the docs can use `ifdef` statements which allow complex `and` and `or` conditionals that `ifeval` does not allow. So, for example that allows closer Katello and Satellite sharing but Satellite specific items where necessary or for shared statements that apply to Foreman on EL or Debian, e.g.:

```
ifeval::["{context}" == "{project-context}"]
* A minimum of 4 GB RAM is required for {ProjectServer} to function.
{Project} running with less RAM than the minimum value might not operate correctly.
endif::[]
endif::[]

ifdef::katello,satellite[]
ifeval::["{context}" == "{project-context}"]
* A minimum of 20 GB RAM is required for {ProjectServer} to function.
In addition, a minimum of 4 GB RAM of swap space is also recommended.
{Project} running with less RAM than the minimum value might not operate correctly.
endif::[]
ifeval::["{context}" == "{smart-proxy-context}"]
* A minimum of 12 GB RAM is required for {SmartProxyServer} to function.
In addition, a minimum of 4 GB RAM of swap space is also recommended.
{SmartProxy} running with less RAM than the minimum value might not operate correctly.
endif::[]
endif::[]
```


ehelms:
Requires https://github.com/theforeman/foreman-documentation/pull/393

jlsherrill:
Looking at the generated PDFs, this generates a whole set of guides for each build  deb/katello/el/sat.  Would it make sense to 'prevent' some guides from being created for some builds? (i.e. dont' build the content guide for foreman-el?)  We don't have to do this now, just curious what the thought is going forward

ehelms:
> Looking at the generated PDFs, this generates a whole set of guides for each build deb/katello/el/sat. Would it make sense to 'prevent' some guides from being created for some builds? (i.e. dont' build the content guide for foreman-el?) We don't have to do this now, just curious what the thought is going forward

Definitely. Let's track that here https://github.com/theforeman/foreman-documentation/issues/396

ehelms:
I rebased this PR. This is by no means a complete overhaul of the install to separate out the Katello parts. But as its a start and introduces the Katello build I would like to get this part in and then keep working on updates in follow up PRs.

guides/common/modules/ref_storage-guidelines.adoc
```diff
@@ -73,19 +79,23 @@ After a full synchronization task is completed, the content is moved to the `/va
 For content in RPM format, each RPM file is moved to the `/var/lib/pulp` directory after it is synchronized.
 A maximum of 5 RPM files are stored in the `/var/cache/pulp/` directory at any time.
 Up to 8 RPM content synchronization tasks can run simultaneously by default, with each using up to 1 GB of metadata.
+endif::[]
 
-ifeval::["{build}" != "foreman-deb"]
+ifndef::foreman-deb[]
```
melcorr, line 84: With the ifdefs, we could define what it is rather than what it isn't. 
Here, for example, we were trying to express that this was for both foreman, and satellite, but not Foreman with Debian.

guides/common/modules/ref_storage-guidelines.adoc
```diff
@@ -95,8 +105,9 @@ If you plan to use ISO images for installing or updating, you must provide exter
 For example, if you are synchronizing four ISO files, each 4 GB in size, this requires a total of 16 GB in the `/var/cache/pulp/` directory.
 Consider the number of ISO files you intend synchronizing because the temporary disk space required for them typically exceeds that of RPM content.
 endif::[]
+endif::[]
 
-ifeval::["{build}" != "foreman-deb"]
+ifndef::foreman-deb[]
```
melcorr, line 110: same as above.

guides/common/modules/ref_storage-requirements.adoc
```diff
@@ -1,16 +1,12 @@
-[id="satellite-storage-requirements_{context}"]
+[id="storage-requirements_{context}"]
 = Storage Requirements
 
-ifeval::["{build}" == "foreman-el"]
-By default, the storage requirements include usage of the Katello plug-in and the requirements for storing and managing content.
-endif::[]
+ifdef::katello,satellite[]
```
melcorr, line 4: this should not be for Satellite

guides/common/modules/ref_system-requirements.adoc
```diff
@@ -4,11 +4,19 @@
...
+ifdef::satellite[]
 * The latest version of Red Hat Enterprise Linux 7 Server
 endif::[]
 * 4-core 2.0 GHz CPU at a minimum
 
+ifdef::foreman-el,foreman-deb[]
+ifeval::["{context}" == "{project-context}"]
+* A minimum of 4 GB RAM is required for {ProjectServer} to function.
+{Project} running with less RAM than the minimum value might not operate correctly.
+endif::[]
```
melcorr, line 16: I can't test the switching of these statements until the error is fixed.

melcorr (changes requested):
Thanks @ehelms just one small error to fix in this PR. I will re-review once fixed. You just made the changes only for the installation guides? I'm not sure if you intended to grep everything and forget the /topics folders here also, but this is just an observation, not a blocker.

@maximiliankolb - would this change complicate anything for you guys? 

Note to self: Content Management Guide has no numbering while the other guides do... fix that.

guides/common/modules/ref_storage-requirements.adoc
ehelms: Why is this is part not for Satellite?

guides/common/modules/ref_storage-guidelines.adoc
ehelms: Fair, I switched to only declaring what build types should be rather than including all but.

guides/common/assembly_preparing-environment-for-installation.adoc
```diff
@@ -1,13 +1,13 @@
-[id="preparing-environment-for-satellite-installation"]
+[id="preparing-environment-for-installation"]
 = Preparing your Environment for Installation
```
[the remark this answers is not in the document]
ehelms: Fixed, I thought I had got them all.

ehelms:
> Thanks @ehelms just one small error to fix in this PR. I will re-review once fixed. You just made the changes only for the installation guides? I'm not sure if you intended to grep everything and forget the /topics folders here also, but this is just an observation, not a blocker.

I was working my way through the guide building up changes, using the presence of ifeval as sort of way to mark things that had not been edited yet. If you prefer that I make the change across the board from ifeval to ifdef I can.

melcorr:
> > Thanks @ehelms just one small error to fix in this PR. I will re-review once fixed. You just made the changes only for the installation guides? I'm not sure if you intended to grep everything and forget the /topics folders here also, but this is just an observation, not a blocker.
> 
> I was working my way through the guide building up changes, using the presence of ifeval as sort of way to mark things that had not been edited yet. If you prefer that I make the change across the board from ifeval to ifdef I can.

No no, that is fine by me! I was just making sure that this was conscious rather than accidental. Thanks.

guides/common/modules/ref_storage-requirements.adoc
melcorr: Oh I see what you did now. Ignore me.

guides/common/assembly_preparing-environment-for-installation.adoc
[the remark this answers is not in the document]
melcorr: Thank you!

melcorr (approved):
This is nice. 
The more I look at this, the more I think it will benefit ATIX using these docs also.

maximiliankolb:
Oh, already merged. I did test it -and lucky us- it works with Antora/our use case as well.

I am curious for future usage of `ifeval` and `ifdef`. 👍

melcorr:
> Oh, already merged. I did test it -and lucky us- it works with Antora/our use case as well.
> 
> I am curious for future usage of `ifeval` and `ifdef`. +1

Sorry @maximiliankolb  I should have waited. The more I played with this, the more confident I got that you would be OK ! Aaand there's always revert ;)

guides/common/assembly_preparing-environment-for-installation.adoc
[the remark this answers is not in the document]
lzap: Just a sidenote - we kept "satellite" in ids because this will ultimately break links downstream. We can probably create two `id=` statements one with satellite (legacy) and one newone if we want to be precise, but I am okay with this - people still land on the correct page, they just need to scroll. So I am good.

guides/common/attributes.adoc
```diff
@@ -194,7 +210,8 @@ ifeval::["{build}" == "satellite"]
 :PlanningDocURL: {BaseURL}planning_for_red_hat_satellite/index#
 :ProvisioningDocURL: {BaseURL}provisioning_guide/index#
 endif::[]
-ifeval::["{build}" == "foreman-deb"]
+
+ifdef::foreman-deb[]
```
lzap, line 214: Ah, we need to avoid spaces in this file:

```
// NOTE: Avoid empty lines so it can be included in the header section
// Document state: "nightly" for master, "stable" for last two releases,
// "unsupported" for the rest and "satellite" for satellite build
```

I haven't found that it breaks anything, but according to docs everything until an empty line is considered to be a "header". Shall we fix this or just ignore this? I am a bit confused by asciidoc here.

guides/common/attributes.adoc
```diff
@@ -29,7 +29,22 @@
...
+endif::[]
+ifeval::["{build}" == "foreman-deb"]
+:foreman-deb:
+endif::[]
+ifeval::["{build}" == "katello"]
+:katello:
+endif::[]
+ifeval::["{build}" == "satellite"]
+:satellite:
+endif::[]
```
lzap, line 45: Nice find! This thing is quite clunky, but when it works results are great.

guides/common/attributes.adoc
melcorr: I checked that build, and it was working. Weird. Let's fix.

lzap:
I like it, so what is the plan. Shall we search-replace other guides as well?

What I am missing is edit of the README to advertise new instructions.

guides/common/modules/ref_system-requirements.adoc
lzap, line 16: I don't understand why you used `{context}` and not simply `context`:
```
doc-Installing_Server_on_Debian/master.adoc
9::context: {project-context}

doc-Installing_Server_on_Red_Hat/master.adoc
11::context: {project-context}
```

ehelms: I don't either. I just copied this from somewhere :)

PR #395, opened 2021-02-12
Refactor Importing Content Guides

jlsherrill:
This tries to unify the Red Hat and non-Red Hat import content sections
* Changes download speed section to use pulp3 concurrency control
* removes puppet content references
* Tried to consolidate syncing instructions


guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
```diff
@@ -1,28 +1,272 @@
-[[Importing_Red_Hat_Content]]
-== Importing Red{nbsp}Hat Content
+[[Importing_Content]]
+== Importing Content
 
-This section describes how to use products and repositories in {Project}, and to create synchronization plans to ensure your {Project} content remains up to date with the content on the Red{nbsp}Hat Content delivery network (CDN).
+This chapter outlines how you can import {customcontent} of different types to {Project}.
```
melcorr, line 4: In Satellite docs, this will render as custom content of different types...

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
```diff
@@ -1,28 +1,272 @@
...
+In the Content Credentials window, click *Create Content Credential*.
+. In the *Name* field, enter a name for your SSL certificate.
+. From the *Type* list, select *SSL Certificate*.
+. In the *Content Credentials Content* field, paste your SSL certificate, or click *Browse* to upload your SSL certificate.
+. Click *Save*.
+
+[[Importing_Content-Importing_a_Custom_GPG_Key]]
+=== Importing a {customgpgtitle} Key
+
+Before you create {customcontent}, ensure that you import any GPG keys that this content is signed with.  This will allow clients to validate the signature of the rpms before installation.  For Red Hat content, this is handled automatically.
```
melcorr, line 50:
```suggestion
Before you create {customcontent}, ensure that you import any GPG keys that this content is signed with. Clients use these keys to validate the signature of the RPMs before installation. For Red{nbsp}Hat content, this is handled automatically.
```

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
```diff
@@ -1,28 +1,272 @@
...
+
+.Prerequisites
+
+You will want a copy of the GPG key used to sign the rpm content that will be served.  Most RPM distribution providers will provide their GPG Key on their website.  You can also extract is manually from an RPM:
+
```
melcorr, line 55:
```suggestion
Ensure that you have a copy of the GPG key used to sign the RPM content that you want to use and manage in {Project}. Most RPM distribution providers provide their GPG Key on their website. You can also extract this manually from an RPM:

```

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
```diff
@@ -1,28 +1,272 @@
-[[Importing_Red_Hat_Content]]
-== Importing Red{nbsp}Hat Content
+[[Importing_Content]]
+== Importing Content
 
-This section describes how to use products and repositories in {Project}, and to create synchronization plans to ensure your {Project} content remains up to date with the content on the Red{nbsp}Hat Content delivery network (CDN).
+This chapter outlines how you can import {customcontent} of different types to {Project}.
+If you want to import Rpms, Files, or different content types to {Project}, it is done with largely the same procedures in this chapter.
```
melcorr, line 5:
```suggestion
If you want to import RPMs, files, or different content types to {Project}, the underlying procedures are the same:
```

@jlsherrill when we met about Katello docs, we talked about how what was missing was context/a scenario/ a defined use case. You mentioned how users need to realise that they don't need to always do CVs, that syncing content might be enough for people. Do you think that we could create a list here of all the scenarios that importing content will serve? I can use that content as part of an introduction with a link to this section also ? Or if you prefer we can do this as part of another PR, that is fine, but since we've tackled an entire chapter in this section, it would be nice to contextualise why and who this is important to.

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
```diff
@@ -1,28 +1,272 @@
...
+Both Red Hat content and non-RH {customcontent} in {Project} have similarities:
+endif::[]
+
+* The relationship between a product and its repositories is the same and the repositories still require synchronization.
+* {customproduct}s require a subscription for clients to access, similar to subscriptions to Red Hat Products.
+{Project} creates a subscription for each {customproduct} you create.
+
+For more information about creating and packaging RPMs, see the https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html-single/rpm_packaging_guide/[RPM Packaging Guide] in the Red{nbsp}Hat Enterprise Linux documentation.
+
+Red Hat Content is already organized into Products, while other {customcontent} can be organized into Products however you wish.  For example, Red{nbsp}Hat Enterprise Linux Server is a _Product_ in {Project}, the repositories for that product consist of different versions, different architectures, and different add-ons.
```
melcorr, line 30:
```suggestion
Red Hat Content is already organized into Products, while other {customcontent} can be organized into Products however you want. For example, {RHEL} Server is a _Product_ in {Project}. The repositories for that Product consist of different versions, architectures, and add-ons.
```

melcorr (commented):
Hi @jlsherrill 

Forgive me, but I am trying to understand what is the benefit of the consolidation of both chapters? 
As I mentioned in an earlier comment, when we talked about where Katello docs could get better, we talked about how there is a lack of defined context, so users end up thinking they need to do things, for example create CVs, when they might not need that at all for their deployment. I started reviewing this purely from a tech docs standpoint, but when I review the new chapter outline, I wonder are we just creating the exact problem we want to avoid. 
At the moment we have two chapters, and by merging it into one chapter, we are essentially having sections that apply in one case but not in another all grouped together. Does this add clarity? Perhaps you can explain to me where the problem was with the way the information is currently laid out?

jlsherrill:
Hey @melcorr !     
No worries, this is all good discussion.  

> we are essentially having sections that apply in one case but not in another all grouped together.

The previous docs were actually the opposite.  There were lots of cases that applied to both cases, but were only in one section.  
These are the only items that are not part by both 'custom content' and 'redhat content':

* uploading rpms
* gpg keys
* ssl certificates

And then the two ways to 'create' content are different:

* creating a repo (custom content)
* enabling a red hat repo

While the old docs seemed to imply that various ways of syncing, sync plans, http proxies, download policies, repo recovery, etc... were all Red hat specific, when they weren't.  I'm open to other suggestions (and its possible ti should be structured differently from upstream to downstream).

A logical breakout would be:
* Creating custom content
   * uploading rpms
   * gpg keys
   * ssl certificates
* Enabling RH Content
* All topics that apply to both (Everything else)

Although i'm not sure that makes sense from a user perspective

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
jlsherrill: Maybe 'how you can import different types of {customcontent}' sounds better?

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
jlsherrill: Thats a great idea. i'll add some context!

melcorr (commented):
Hi @jlsherrill 
Thank you for explaining this to me. It is a lot clearer now. 
> 
> While the old docs seemed to imply that various ways of syncing, sync plans, http proxies, download policies, repo recovery, etc... were all Red hat specific, when they weren't. I'm open to other suggestions (and its possible ti should be structured differently from upstream to downstream).

This is great. Makes no sense to leave it as is. 

I think this bit could be expanded in the introduction with a Red Hat and a non-Red Hat example: 
`
Red Hat Content is already organized into Products, while other content can be organized into Products however you wish. For example, Red Hat Enterprise Linux Server is a Product in Foreman, the repositories for that product consist of different versions, different architectures, and different add-ons. For Red Hat repositories, products are created automatically after enabling the repository. See Enabling Red Hat Repositories`

Then we could explicitly state or list that once the user has the product set up depending on the content type, they can proceed to the following actions depending on their requirements.
> 
> A logical breakout would be:
 I would tweak it slightly to this:
     * Creating a custom Product
     ** uploading rpms
     ** gpg keys
     ** ssl certificates
     ** Creating a Product
(in whichever order makes sense)
Then the easy RH enable automagic stuff:
> 
>     * Enabling RH Content
> 
>     * All topics that apply to both (Everything else)

What do you think?

PS: I just noted another few nitpicks here and there. Nothing serious.

jlsherrill:
I reworked it a bit @melcorr, curious your thoughts.  I went with the approach of showing only the items needed before creating the custom repo (Ssl certs), then going through creating a custom repo, and then enabling a redhat repo and then everything else.  For the GPG section i go into more depth about how its only for Custom repos.  This isn't really needed before creating and is more of an advanced feature

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
```diff
@@ -294,7 +265,176 @@ do
...
+You can set the default download policy that {Project} applies to repositories that you create in all organizations.
+
+ifeval::["{build}" == "satellite"]
+
+Depending on whether it is a Red Hat or non-Red Hat {customrepo}, {Project} uses separate settings.
+Changing the default value does not change existing settings.
+
+endif::[]
+
+ifeval::["{build}" != "satellite"]
```
melcorr, line 313: If the messages are the same here, we can probably drop the ifevals?

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
```diff
@@ -571,3 +665,93 @@ done
...
+
+[[Importing_Content-Limiting_Synchronization_Speed]]
+
+=== Limiting Synchronization Concurrency
+
+By default each Repository Synchronization job will fetch up to 10 files at a time.  This can be adjusted on a per repository basis.
+
+Increasing the limit may improve performance, but can cause the upstream server to be overloaded or start rejecting requests.  If you are seeing Repository syncs fail due to the upstream servers rejecting requests, you may want to try lowering the limit.
+
+To do so from the cli:
```
melcorr, line 677:
```suggestion
To do so from the CLI:
```

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
```diff
@@ -571,3 +665,93 @@ done
 ----
 # hammer product list --organization $ORG --sync-plan $SYNC_PLAN
 ----
+
+[[Importing_Content-Limiting_Synchronization_Speed]]
+
+=== Limiting Synchronization Concurrency
+
+By default each Repository Synchronization job will fetch up to 10 files at a time.  This can be adjusted on a per repository basis.
```
melcorr, line 673:
```suggestion
By default each Repository Synchronization job can fetch up to 10 files at a time.  This can be adjusted on a per repository basis.
```

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
```diff
@@ -571,3 +665,93 @@ done
...
+
+[options="nowrap" subs="verbatim,quotes"]
+----
+# hammer repository update --organization $ORG --download-concurrency 5 --id $REPO_ID
+----
+
+[[Importing_Content-Importing_a_Custom_GPG_Key]]
+=== Importing a {customgpgtitle} Key
+
+When clients are consuming signed {customcontent}, you will want ensure the clients are configured to validate the installation of rpms with the appropriate GPG Key.  This helps to ensure that only packages from authorized sources can be installed.
```
melcorr, line 687:
```suggestion
When clients are consuming signed {customcontent}, ensure that the clients are configured to validate the installation of RPMs with the appropriate GPG Key.  This helps to ensure that only packages from authorized sources can be installed.
```

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
```diff
@@ -571,3 +665,93 @@ done
...
+
+Red Hat content will already be configured with the appropriate GPG key and thus GPG Key management of Red Hat Repositories is not supported.
```
melcorr, line 689:
```suggestion
Red Hat content is already configured with the appropriate GPG key and thus GPG Key management of Red Hat Repositories is not supported.
```

melcorr (changes requested):
Hi @jlsherrill 
I like this a lot. I added a few nits - just tech docs style stuff that you don't need to worry about. 
I think we are good now once these small fixes are made, thank you so much!

jlsherrill:
Thanks @melcorr I think i've addressed all comments!

melcorr (approved):
Thanks @jlsherrill

PR #411, opened 2021-02-18
Update installation repositories configuration

ehelms:
Working on updates to the installation guide to support Katello 4.0 and Foreman on EL and Debian for 2.4/nightly of Foreman. I am trying to scrub satellite from filenames and ensure the right information gets generated per combination. At the same time I am trying to match up information from the current Foreman install guide from theforeman.org.

My tactic is to follow the current breakdown to end up with the following guides:

 * Foreman on Enterprise Linux
   * includes: CentOS 7, CentOS 8, RHEL 7, RHEL 8
 * Foreman on Debian
   * includes Debian 10, Ubuntu 18.04
 * Katello on Enterprise Linux
    * includes: CentOS 7, CentOS 8, RHEL 7, RHEL 8

One tricky part is that there needs to be different instructions in places for EL 7 vs EL 8. Paths may be different on disk, yum vs dnf to name a few.


guides/common/assembly_preparing-environment-for-installation.adoc
```diff
@@ -17,7 +17,7 @@ include::modules/ref_supported-operating-systems.adoc[leveloffset=+1]
 
 include::modules/ref_supported-browsers.adoc[leveloffset=+1]
 
-include::modules/ref_satellite-ports-and-firewalls-requirements.adoc[leveloffset=+1]
+include::modules/ref_ports-and-firewall-requirements.adoc[leveloffset=+1]
```
lzap, line 20: These are fine, just keep in mind that changing reference "anchors" will ultimately break links, so we need to to this via two stages: https://github.com/theforeman/foreman-documentation/pull/416

guides/common/modules/ref_ports-and-firewall-requirements.adoc
```diff
@@ -1,5 +1,5 @@
-[id="satellite-ports-and-firewalls-requirements_{context}"]
-= Ports and Firewalls Requirements
+[id="ports-and-firewall-requirements_{context}"]
+= Ports and Firewall Requirements
```
lzap, line 2: Please keep the old id along with the new one, we need to do this at least for one release not to break our CI.

guides/common/modules/ref_ports-and-firewall-requirements.adoc
```diff
@@ -15,35 +15,41 @@ This includes the base operating system on which {SmartProxyServer} is running.
...
 |====
 | Port | Protocol | Service | Required For
 | 443 | TCP | HTTPS | Subscription Management Services (access.redhat.com) and connecting to the Red{nbsp}Hat CDN (cdn.redhat.com).
 |====
+endif::[]
+endif::[]
 
-ifeval::["{build}" == "satellite"]
+ifdef::satellite[]
+ifeval::["{mode}" == "connected"]
```
lzap, line 42: Oh is the plan to merge connected and disconnected guides into single content? That would be nice.

lzap:
I haven't looked into technical correctness but I think this is going the good direction.

guides/common/modules/proc_configuring-foreman-repositories-el-8.adoc
```diff
@@ -0,0 +1,34 @@
...
+. Install the `katello-repos-latest.rpm` package
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# yum localinstall https://fedorapeople.org/groups/katello/releases/yum/{KatelloVersion}/katello/el8/x86_64/katello-repos-latest.rpm
+----
+endif::[]
+ifdef::foreman-el,katello[]
++
+. Install the `puppet6-release-el-7.noarch.rpm` package:
```
ekohl, line 29:
```suggestion
. Install the `puppet6-release-el-8.noarch.rpm` package:
```

guides/common/modules/proc_configuring-foreman-repositories-el-8.adoc
```diff
@@ -0,0 +1,34 @@
+[id="configuring-foreman-repositories-el-8"]
```
ekohl, line 1: This looks like an exact copy of el7 with `s/el7/el8/g`. Can that somehow be combined with something like a major version variable?

guides/common/modules/ref_ports-and-firewall-requirements.adoc
melcorr: I think both guides are currently made from the same source!

melcorr:
I agree with the guide breakdown as described above. 
In Satellite docs, RHEL 6, 7, 8 were described in subsections of the one procedure if necessary. If it's clear, the user can jump to that section easily. I think the alternative would be quite convoluted.

ehelms:
@melcorr There are two sections in this PR where I have attempted to apply a per OS sub-section approach: configuring repositories and storage requirements. Could you take a look and let me know if that is how you were thinking of approaching splitting them out?

guides/doc-Installing_Server_on_Red_Hat/master.adoc
```diff
@@ -1,6 +1,6 @@
 :imagesdir: common/images
 :numbered:
-ifeval::["{build}" != "satellite"]
+ifdef::foreman-el,foreman-deb,katello[]
```
melcorr, line 3: This is not working in my tests.

melcorr (commented):
Yep, the OSes look fine, but I wonder why the menu nav sidebar will not work with the ifdef. 
I wonder was this what I hit upon yesterday trying to give the Katello build of this guide a better name than "Installing Foreman 2.5 server".

guides/doc-Installing_Server_on_Red_Hat/master.adoc
ehelms: Ahh, becuase the attributes file is loaded after this block, so it previously always worked by happy accident.

guides/common/modules/proc_configuring-foreman-repositories-el-8.adoc
ehelms: Consolidated into a single guide using a variable.

ehelms:
One decent size change made in here is I have dropped the notion of "manual" and "answer file" based installations as two separate methods. Users should *not* edit the answers file directly and thus I have removed all notions of this. This instead focuses on a single method to install: running the installer itself.

guides/common/modules/ref_supported-operating-systems.adoc
```diff
@@ -1,53 +1,42 @@
 [id="supported-operating-systems_{context}"]
 = Supported Operating Systems
 
-ifeval::["{build}" == "satellite"]
+ifdef::satellite[]
 You can install the operating system from a disc, local ISO image, kickstart, or any other method that Red{nbsp}Hat supports.
 Red{nbsp}Hat {ProductName} is supported only on the latest versions of {RHEL} 7 Server that is available at the time when {ProductName} {ProductVersion} is installed.
 Previous versions of Red{nbsp}Hat Enterprise Linux including EUS or z-stream are not supported.
 endif::[]
```
melcorr, line 9: The satellite guide does not build after this point, so something has gone awry :)

melcorr (changes requested):
Hi @ehelms 
The Katello guide looks great. 
I like the titles.
I cannot review the Satellite guide as the build for it is breaking somewhere in a sneaky way that throws no errors. I've marked the place where it breaks down and if I figure it out I'll comment again. I think that there's a missing endif. 
There is a warning when building for the installation katello, foreman-el, and foreman-deb guides: 
asciidoctor: WARNING: common/modules/proc_installing-server-packages-el.adoc: line 3: id assigned to block already in use: configuring-foreman-repositories-el-yum
Though it appears to build for me. I will try and figure this out. 
I really like the way that you've gotten around and through the package management issues. I wouldn't have thought to do this so I really appreciate it.

guides/common/modules/ref_supported-operating-systems.adoc
```diff
@@ -1,53 +1,42 @@
...
+| CentOS 8 | x86_64 only | EPEL is not supported.
+| {RHEL} 7 | x86_64 only | EPEL is required.
 endif::[]
-| Ubuntu 16.04 (Xenial) | amd64 |
+ifdef::foreman-deb[]
 | Ubuntu 18.04 (Bionic) | amd64 |
-| Debian 9 (Stretch) | amd64 |
+| Debian 10 (Buster) | amd64 |
+endif::[]
 |====
```
melcorr, line 28: Perhaps an endif is needed here?

melcorr:
Looks good, @ehelms 
One question: no RHEL 8 for Satellite yet?

ehelms:
> One question: no RHEL 8 for Satellite yet?

Not till Satellite 7+

melcorr:
> > One question: no RHEL 8 for Satellite yet?
> 
> Not till Satellite 7+

Cool. Thanks for all your work on this.

PR #412, opened 2021-02-18
Cherry picks from 6.9-beta into 2.3

lzap:
To remove all Satellite related branches, I would like to reconfigure RH Docs gitlab to fetch directly from upstream branch, that would be `2.3` in this case.

Since the `SATELLITE-6.9-beta` branch already quite diverged, I think the best idea not to disrupt downstream is to cherry pick all changes that went into the branch into `2.3`.

Then I can proceed with changing the synchronization script in gitlab and finally I can remove all Satellite branches.

More info: https://community.theforeman.org/t/cleanup-of-foreman-documentation-branches/22445/2

@mcorr @spetrosi


lzap:
For the record, before deleting those branches here are shas just in case we'd need to recover them for any reason:

```
  remotes/origin/6.7                            61a0bfa Clarify what client to click in the Keycloak web UI
  remotes/origin/6.8-beta                       2649ced properly add the cname at the right spot
  remotes/origin/SATELLITE-6.7                  6800acf Add the /var/spool/squid/ directory to Proxy storage reqs table (#368)
  remotes/origin/SATELLITE-6.8                  2afb5ee Fix broken multiline ifdef directives
  remotes/origin/SATELLITE-6.8-beta             44f125d Add docinfo.xml to the Provisioning guide directory for downstream (#288)
  remotes/origin/SATELLITE-6.9-beta             fc20460 Remove redundant line in note -Config Repositories section
```

There are two leftovers which were probably accidentaly pushed, going to remove those as well:

```
  remotes/origin/add-script                     3af44f9 Adding upstream/downstream sync script
  remotes/origin/provisioning_guides            8de6b30 [WIP] Provisioning guide
```

lzap:
One note, I will actually not delete SATELLITE-6.7  and SATELLITE-6.8 but rename them to 1.24 and 2.1 respectively so fixes can be done for downstream only. We would not publish those versions upstream.

lzap:
I created `1.24` and `2.1` branches. This is now ready to be merged.

PR #414, opened 2021-02-19
Improve 404 redirect and foreman attribute

### LINKED ISSUE #406: Fix deploy action and JS moved content
* Links are changed now: `foreman.html` is now `foreman-el.html` - we need to update the JS that suggest where the new content is at least.
* Deploy action does not work - it did not update the site - need to fix.
* Deploy action should delete all files before copying new content but only in nightly or X.Y directories (not all that would delete all content)

lzap:
This improves redirect from `index.html` to `index-el.html` in 404 error. Files haven't been deployed yet, I am going to look into that in a moment. This fixes #406.

Second change is a proposal to introduce `foreman` attribute when `foreman-el` or `foreman-deb` is set so we can write easier if statements.

### DIFF
--- a/guides/README.md
+++ b/guides/README.md
@@ -102,8 +102,28 @@ To use them, use "attributes" keyword:
 Hide or show specific blocks, paragraphs, warnings or chapters via special variable called "build".
 Its value can be set either to "foreman-el" or "satellite":
 
-	ifeval::["{build}" == "foreman-el"]
+	ifeval::["{build}" == "katello"]
 	NOTE: This part is only relevant for deployments with Katello plugin.
+	endif::[]
+
+This syntax does not allow multiple conditionals, in that case use, now preferred, `ifdef` syntax:
+
+	ifdef::katello[]
+	NOTE: This part is only relevant for deployments with Katello plugin.
+	endif::[]
+
+Use comma for logic "or":
+
+	ifdef::katello,satellite[]
+	NOTE: This part is only relevant for deployments with Katello plugin or in Satellite environment.
+	endif::[]
+
+Some files are included in different contexts, there are attributes to find the correct context. In these cases use both `ifdef` and `ifeval`:
+
+	ifdef::foreman-el,foreman-deb[]
+	ifeval::["{context}" == "{project-context}"]
+	* A minimum of 4 GB RAM is required for {ProjectServer} to function.
+	endif::[]
 	endif::[]
 
 When doing review, consider checking out the topic branch and putting necessary changes on top of author's work to making many comments on github.
--- a/guides/common/attributes.adoc
+++ b/guides/common/attributes.adoc
@@ -29,13 +29,14 @@
 :SatelliteAnsibleVersion: 2.9
 :SpecialCaseProductVersion: 6.8
 //the above attribute is for Package Manifests etc that will fail the upstream link-checker during the beta see Issue #115 on GitHub
-
 // Define properties to represent each build. Allows doing 'or' and 'and' operations for conditions.
 ifeval::["{build}" == "foreman-el"]
 :foreman-el:
+:foreman:
 endif::[]
 ifeval::["{build}" == "foreman-deb"]
 :foreman-deb:
+:foreman:
 endif::[]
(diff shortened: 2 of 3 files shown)


lzap:
I am also improving our docs about if statements.

lzap:
Btw you can't test redirection because hugo local server does not have a 404 custom page. Just verify the JS syntax and merge it :-) @ehelms or @ekohl would you mind reviewing the JS code?

lzap:
A link you can test with (after this is merged and deployed): https://docs.theforeman.org/nightly/Planning_Guide/index-foreman.html

guides/common/attributes.adoc
```diff
@@ -126,7 +126,6 @@ ifdef::katello,foreman-el[]
 :PlanningDocURL: {BaseURL}Planning_Guide/index-foreman.html#
 :ProvisioningDocURL: {BaseURL}Provisioning_Guide/index-foreman.html#
 endif::[]
-
```
lzap, line 129: Empty lines will make all those attributes not to be part of a header. I am not sure what it makes, looks like it works fine, but I would rather keep it the old way.

lzap:
We need to merge https://github.com/theforeman/foreman-documentation/pull/417 first and then rebase on top of it due to the broken links.

web/layouts/404.html
```diff
@@ -4,10 +4,17 @@ <h2>Page not found!</h2>
 
 <p>We are sorry, but this page does not exist.
 <script>
-    if (document.location.toString().includes("/master/")) {
-        document.write("This page has been <a href=\"")
-        document.write(document.location.toString().replace("/master/", "/nightly/"))
-        document.write("\">moved</a>.")
+    var loc = document.location.toString();
```
rabajaj0509, line 7: can we use `const` instead of var here?

web/layouts/404.html
```diff
@@ -4,10 +4,17 @@ <h2>Page not found!</h2>
...
+    if (loc.includes("/master/") || loc.includes("/index.html")) {
+        newLocation = document.location.toString()
```
rabajaj0509, line 9: I think  instead of `document.location.toString()` you can use `window.location.href` at all places.

web/layouts/404.html
```diff
@@ -4,10 +4,17 @@ <h2>Page not found!</h2>
...
+                .replace("/master/", "/nightly/")
+                .replace("/index.html", "/index-el.html");
+        document.write("This page has been <a href=\"");
+        document.write(newLocation);
+        document.write("\">moved</a>. Redirecting in 3 seconds.");
```
rabajaj0509, line 14: you can do string interpolation like: 
```js
`This page has been moved to ${newlocation}. Redirecting in 3 seconds`
```

rabajaj0509 (commented):
I think this should make the code look a little better :)

guides/common/attributes.adoc
ehelms: If it must be that way to work, OK. But empty lines make this file more readable, the "blocks" more identifiable.

guides/README.md
```diff
@@ -102,10 +102,30 @@ To use them, use "attributes" keyword:
...
+	endif::[]
+
+Some files are included in different contexts, there are attributes to find the correct context. In these cases use both `ifdef` and `ifeval`:
+
+	ifdef::foreman-el,foreman-deb[]
+	ifeval::["{context}" == "{project-context}"]
+	* A minimum of 4 GB RAM is required for {ProjectServer} to function.
+	endif::[]
+	endif::[]
+
```
ehelms, line 128: Thanks for adding this section on how you can use the logic.

melcorr:
The text in the readme LGTM.

web/layouts/404.html
lzap: Thanks, these are all valid points but what I have is kind-of-a tested and since everyone is okay I am gonna merge this to shorten the time we don't have this in our 404 page. I learned a lot, cheers! :-)

lzap:
Aaaand I had a typo there: https://github.com/theforeman/foreman-documentation/pull/419 :)

PR #416, opened 2021-02-19
Prepare anchors for upstreamization

lzap:
This prepares anchors so we can start removing product references from them.

Please cherry pick it into 2.3.


ehelms (approved):
@melcorr could you take a look here? This likely impacts other PRs so would be a good base to have in to rebase on to.

melcorr (approved):
I built and checked different scenarios. 
Debian is not building but unrelated to this and I will raise a PR for that. 
Good to go I think.

Issue #310, opened 2020-11-09
Change all the xrefs to hard links 


melcorr:
As part of #305 a lot of content was modularized so it could suit both upstream/downstream as well as Ansible REX and plain REX guides. 

The xrefs can be problematic with FCC, so it's best to add hard links. 

Once the PR is merged, I will go back and swap out the xrefs for hard links.

tahliar:
@melcorr I can do this one, too!

melcorr:
Thank you @tahliar

tahliar:
Resolved by PR #427

PR #427, opened 2021-03-02
Change xrefs to hard links

tahliar:
For Issue #310 

I've done this for the content in the common directory only, as it is likely to be shared. Let me know if you'd also like me to do the same for content that's not in common, for future-proofing. 

The remaining xrefs are only the ones in assemblies that include the referenced module (therefore will always work), or ones that are carefully marked up to be shared between similar guides (e.g. content shared between the Foreman and Smart Proxy installation guides). In one case, I maintained this standard by moving content from a concept module back into the assembly (if intro content applies only to a single assembly, it should ideally stay in that assembly rather than be a concept module).

I also moved an included module that was randomly splitting up a section.

Cherry-pick into:


melcorr (changes requested):
Hey @tahliar when I go to test this, the Installation guides aren't building for me.
I think that if you rebase the branch, everything should be fixed! 
Thank you for this!

melcorr (changes requested):
@tahliar sorry to be a pain but would you mind reverting that "merge master into" commit and rebase your branch from master, just to keep the commits history cleaner!

guides/common/modules/proc_adding-life-cycle-environments.adoc
```diff
@@ -23,12 +23,7 @@ To add a life cycle environment to {SmartProxyServer}, complete the following st
 . To synchronize the content on the {SmartProxy}, click the *Overview* tab and click *Synchronize*.
 . Select either *Optimized Sync* or *Complete Sync*.
 +
-ifeval::["{context}" == "{smart-proxy-context}"]
-For definitions of each synchronization type, see {ContentManagementDocURL}Importing_Red_Hat_Content-Recovering_a_Repository[Recovering a Repository] in the _Content Management Guide_.
-endif::[]
-ifeval::["{context}" == "content-management"]
-For definitions of each synchronization type, see xref:Importing_Red_Hat_Content-Recovering_a_Repository[].
-endif::[]
+For definitions of each synchronization type, see {ContentManagementDocURL}Importing_Red_Hat_Content-Recovering_a_Repository[Recovering a Repository].
```
melcorr, line 26: This link does not work for me on a Foreman build but works on a Satellite build...

melcorr (changes requested):
Hi Tahlia, 
Everything looks OK now except one link that I've highlighted doesn't point to the section, just resets to the guide for Foreman but works for Satellite. i'm wondering if you'd know a way around this?

guides/common/modules/proc_adding-life-cycle-environments.adoc
tahliar: My best guess would be that it's a link to a section that we hide in the Foreman build, but that doesn't make sense if the xref worked fine before I changed it to a link. I'll investigate.

tahliar: Ah, the answer is that the xref was broken before! The ID changed from Importing_Red_Hat_Content to Importing_Content and there are a few broken xrefs as a result. I'll fix them here.

melcorr (approved):
I am approving but I am not cherry-picking for safety's sake. 
Thank you very much for your contribution. 
As far as doing the non-modularized guides, I would welcome PRs doing this, but I would not demand it :-)

PR #432, opened 2021-03-03
Correct links for 2.3 branch

lzap:
Due to our foreman > foreman-el refactoring all links for 2.3 branch were invalidated. Instead using a shared include file, we need to make a copy for this particular case.


web/content/version-2.3/index.md
```diff
@@ -1,5 +1,6 @@
 ---
```
ekohl, line 1: This feels like a weird design. I would expect something like:

* https://docs.theforeman.org/index.html a generated list of all available version. No link directly to any docs. The source for this is in the `gh-pages` branch.
* https://docs.theforeman.org/nightly/index.html -> comes from `master` branch and lists the documents in nightly
* https://docs.theforeman.org/2.3/index.html -> comes from `2.3` branch and lists the documents in 2.3

Another possible scheme:

* master branch is just the site for the index. This also determines which versions are linked
* nightly branch is nightly docs
* x.y branch is for x.y docs
* gh-pages is generated from a combination of the above branches

lzap: Problem is, directories with content generated via our AsciiDoc can't be shared with the static website. But I could create `index.html` welcome page and then `nightly.html` and `2.3.html` pages with the branch landing pages. That would work just fine.

The scheme you suggest is not bad, I'd like to keep this PR simple but it's something I will keep in mind. It's not a big change and it should bring more clarity to the deployment proces while it should not burden docs writers too much - they will probably only work in the `nightly` branch. Migration path should be pretty easy.

lzap: Uh now I understand why I have chosen the clunky `version-2.3/` naming scheme - I could not get hugo to build a content file in the root folder: https://gohugo.io/content-management/organization/

lzap: Damn I really don't know how to do this with hugo, I was hoping to use some simple and fast tool but hugo seems to be quite complex. I need to ask: https://discourse.gohugo.io/t/how-to-get-content-my-file-md-rendered/31582

lzap: So found the solution, I can enable ugly links and have `/2.3.html`. Let's do this?

melcorr: I don't mind!

lzap:
How about this? A bit cleaner.

melcorr:
> How about this? A bit cleaner.

@ekohl 's opinion counts here more than mine.

web/layouts/partials/header.html
```diff
@@ -18,6 +18,7 @@
           <li class="dropdown">
             <a role="button" aria-haspopup="true" aria-expanded="false">Contribute</a>
             <div class="dropdown-content">
+              <a href="https://www.theforeman.org">Project site</a>
               <a href="https://community.theforeman.org">Discourse forum</a>
```
lzap, line 22: I stuffed these two changes as well - clicking on logo gets you back to the welcome page and there's new Project site link in the Contribute menu now. This feels better to me.

lzap:
Do you mind @ekohl ?

lzap:
If you are okay let's merge? @melcorr

PR #441, opened 2021-03-05
Better separate the smart proxy installation instructions between Katello and Foreman

ianballou:
This PR adds some better separation between vanilla (no Katello), content, and Satellite smart proxies.

I've removed the parts that say "Katello only" and hid those sections instead.

Let me know if I missed anything.

### DIFF
--- a/guides/common/modules/proc_installing-capsule-server-packages.adoc
+++ b/guides/common/modules/proc_installing-capsule-server-packages.adoc
@@ -1,8 +1,17 @@
 [id="installing-capsule-server-packages_{context}"]
+[id="installing-the-satellite-capsule-server-packages_{context}"]
 
 = Installing {SmartProxyServer} Packages
 
 Before installing {SmartProxyServer} packages, you must update all packages that are installed on the base operating system.
+
+ifdef::foreman-el,katello[]
+* xref:#centos-7-package[Red Hat Enterprise Linux 7 / CentOS Linux 7]
+* xref:#centos-8-package[Red Hat Enterprise Linux 8 / CentOS Linux 8]
+endif::[]
+ifdef::satellite[]
+* xref:#rhel-7-package[Red Hat Enterprise Linux 7]
+endif::[]
 
 .Procedure
 To install {SmartProxyServer}, complete the following steps:
@@ -28,4 +37,24 @@ ifndef::satellite[]
 ----
 # {package-install} foreman-proxy-content
 ----
+
+ifdef::foreman-el,katello[]
+== [[centos-7-package]]Red Hat Enterprise Linux 7 / CentOS Linux 7
+endif::[]
+
+ifdef::satellite[]
+== [[rhel-7-package]]Red Hat Enterprise Linux 7
+endif::[]
+
+ifdef::foreman-el,katello,satellite[]
+:package-manager: yum
+include::proc_installing-proxy-packages-el.adoc[]
+endif::[]
+
+ifdef::foreman-el,katello[]
+== [[centos-8-package]]Red Hat Enterprise Linux 8 / CentOS Linux 8
+
+:package-manager: dnf
+include::proc_installing-proxy-packages-el.adoc[]
+
 endif::[]
--- a/guides/common/modules/proc_registering-capsule-to-satellite-server.adoc
+++ b/guides/common/modules/proc_registering-capsule-to-satellite-server.adoc
@@ -1,9 +1,13 @@
-[id="registering-to-server_{context}"]
-[id="registering-to-satellite-server_{context}"]
+[id="registering-capsule-to-server_{context}"]
+[id="registering-capsule-to-satellite-server_{context}"]
 
 = Registering to {ProjectServer}
 
 Use this procedure to register the base operating system on which you want to install {SmartProxyServer} to {ProjectServer}.
+
+ifdef::katello[]
+Registering your {SmartProxyServer} as a content host is optional unless you wish to download the installation packages from your synced repositories.
+endif::[]
 
 .Subscription Manifest Prerequisites
 * On {ProjectServer}, a manifest must be installed and it must contain the appropriate repositories for the organization you want {SmartProxy} to belong to.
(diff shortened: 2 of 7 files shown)


ehelms:
I'd suggest we drop `include::modules/proc_configuring-repositories-proxy.adoc[leveloffset=+1]` in favor of 
https://github.com/theforeman/foreman-documentation/blob/master/guides/common/modules/proc_configuring-repositories.adoc

guides/common/modules/proc_installing-capsule-server-packages.adoc
```diff
@@ -13,19 +13,27 @@ To install {SmartProxyServer}, complete the following steps:
 ----
 # {package-update}
 ----
-ifeval::["{build}" == "satellite"]
```
ehelms, line 16: I'd suggest we either drop this file all together and use https://github.com/theforeman/foreman-documentation/blob/master/guides/common/modules/proc_installing-the-satellite-server-packages.adoc or create a similarly structure file for smart-proxy/capsule. As there is a lot of nuance missing here per OS.

ehelms:
The storage requirements will need an update to account for OS versions similar to https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-katello.html#storage-requirements_foreman

ianballou:
@ehelms I updated the smart proxy storage docs.  Let me know how the numbers and columns look. I removed `/opt` from CentOS 8 to match them main install storage requirements, but there might be something I missed to add.

ehelms:
@ianballou Looks like you need a rebase.

lzap:
LGTM, this was rebased. Any objections to merge?

ehelms:
When rendered, this has a section titled `2.1. Registering to Foreman server` which talks about sub-man registering the host to Foreman before installing. That is not actually required and I wonder if it should be an optional step later in the process.

guides/common/modules/proc_installing-capsule-server-packages.adoc
```diff
@@ -28,4 +34,28 @@ ifndef::satellite[]
 ----
 # {package-install} foreman-proxy-content
 ----
+=======
```
ehelms, line 37: I think this line is breaking rendering.

ianballou:
> When rendered, this has a section titled `2.1. Registering to Foreman server` which talks about sub-man registering the host to Foreman before installing. That is not actually required and I wonder if it should be an optional step later in the process.

@ehelms I agree that it's optional.  I feel like it's there in case the smart proxy-related packages are expected to be downloaded from synced repositories on the Katello server.  I feel like it makes more sense to have this first then to mention it later. Otherwise I'd argue to remove it completely.  What do you think about just adding a note mentioning that this step is optional?

guides/common/modules/proc_installing-capsule-server-packages.adoc
ianballou: Looks like it, I've pushed the fix.

ehelms:
> @ehelms I agree that it's optional. I feel like it's there in case the smart proxy-related packages are expected to be downloaded from synced repositories on the Katello server. I feel like it makes more sense to have this first then to mention it later. Otherwise I'd argue to remove it completely. What do you think about just adding a note mentioning that this step is optional?

I like either optional, with a note about if you plan to sync the Foreman and Katello repositories locally then you need to do this step or make this an advanced method that comes after the "default" way of using the hosted repositories directly.

ianballou:
Alright, so I fixed it by creating a separate `common/modules/proc_registering-capsule-to-satellite-server.adoc` file since I needed to get the "optional" warning underneath the bold title.

lzap:
@ehelms or @ekohl can we get the final blessing?

lzap:
I am about to branch off 2.4 series, what is the status please? I can hold off the branching process if this is something that should be in 2.4/4.0. Let me know, thanks.

ianballou:
@lzap I don't think this PR is worth holding back branching.

guides/common/modules/proc_installing-capsule-server-packages.adoc
```diff
@@ -28,4 +37,27 @@ ifndef::satellite[]
...
+ifdef::foreman-el,katello[]
+== [[centos-8-package]]CentOS 8
+
+:package-manager: dnf
+include::proc_installing-proxy-packages-el.adoc[]
+
+endif::[]
+
+ifdef::foreman-el,katello,satellite[]
+== [[rhel-7-package]]Red Hat Enterprise Linux 7
```
ekohl, line 58: For non-Satellite builds, can we combine it into RHEL 7 / CentOS 7 or are they really different?

guides/common/modules/proc_installing-capsule-server-packages.adoc
```diff
@@ -28,4 +37,27 @@ ifndef::satellite[]
...
+ifdef::foreman-el,katello[]
+== [[centos-7-package]]CentOS 7
+
+:package-manager: yum
+include::proc_installing-proxy-packages-el.adoc[]
+
+endif::[]
+
+ifdef::foreman-el,katello[]
+== [[centos-8-package]]CentOS 8
```
ekohl, line 50: Don't we support RHEL 8 upstream?

ekohl (commented):
I think overall this is ok and probably already an improvement. My biggest note (also inline) is that upstream I'd like to see `RHEL x / CentOS x` where the instructions are identical.

Also a note that CentOS 8 is discouraged by the CentOS project. They now prefer explicit CentOS Linux 8 or CentOS Stream 8. There may be many more of these, but I want to raise awareness to it.

guides/common/modules/proc_registering-capsule-to-satellite-server.adoc
```diff
@@ -0,0 +1,53 @@
...
+.Subscription Manifest Prerequisites
+* On {ProjectServer}, a manifest must be installed and it must contain the appropriate repositories for the organization you want {SmartProxy} to belong to.
+* The manifest must contain repositories for the base operating system on which you want to install {SmartProxy}, as well as any clients that you want to connect to {SmartProxy}.
+* The repositories must be synchronized.
+
+For more information on manifests and repositories, see {ContentManagementDocURL}Managing_Subscriptions/[Managing Subscriptions] in the _{ProjectName} Content Management Guide_.
+
+.Proxy and Network Prerequisites
+* The {ProjectServer} base operating system must be able to resolve the host name of the {SmartProxy} base operating system and vice versa.
+ifndef::foreman-deb[]
```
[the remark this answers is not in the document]
ianballou: I'll remove it.

guides/common/modules/proc_installing-capsule-server-packages.adoc
ianballou: I'm going to assume yes unless anyone else says otherwise.

guides/common/modules/ref_capsule-storage-requirements.adoc
```diff
@@ -7,6 +7,34 @@ These values are based on expected use case scenarios and can vary according to
...
 
+ifdef::foreman-el,katello[]
+== [[storage-centos-7]]CentOS 7
+
+.Storage Requirements for {SmartProxyServer} Installation
+[cols="1,1,1",options="header"]
+|====
+|Directory |Installation Size |Runtime Size
+|/var/lib/pulp/ |1 MB |300 GB
+|/var/opt/rh/rh-postgresql12/lib/pgsql |100 MB |10 GB
```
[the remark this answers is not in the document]
ianballou: I'm not seeing what would need the variable. Let me know if my recent push helped at all with this.

jturel (approved):
LGTM

@ehelms @ekohl it appears that all feedback has been addressed for some time. is it OK to be merged?

melcorr:
Would you mind rebasing, and I will merge! TY

ianballou:
@melcorr I've rebased just now.

PR #452, opened 2021-03-12
Update common content to adhere to templates

tahliar:
I've made some edits to get (almost) everything in `common` in compliance with the module templates. Some notes:

Assemblies:

* Added opening/closing context ifdefs to assemblies where it was missing. Technically those lines are only required when an assembly has a context of its own (to preserve the original guide-wide context value and reinstate it at the end of the assembly), such as in assembly_configuring-keycloak-external-authentication.adoc, but it doesn't hurt to just always include them.
  * Addendum to this, I actually had to not include those lines in more deeply nested assemblies, because it turns out if you nest master.adoc (with a context) > assembly (with a context) > assembly (doesn't matter if it has a context), it only reverts back to the second-level context, not the master file's context. I wonder if the template authors have a solution for that, aside from "don't nest assemblies that far" :laughing:  which I am in favour of, too, to be fair!
I can probably get rid of the deep nesting in another PR though, as the third-level assembly in question, assembly_configuring-project-settings-for-keycloak-authentication.adoc, has a module for using the web UI and a module for using the CLI, so could be converted into a single procedure file using the .Procedure/.For CLI Users structure we use in other places (that or just include those modules in the second-level assembly instead, with leveloffset=+2).

* Assemblies included in guides/other assemblies only really need their own context if you want to include the same module in multiple assemblies in the same guide, so I didn't give every assembly its own context (that would have ballooned out to changing a bunch of xrefs and links), but that's an option you could consider if you want strict template adherence.

* In assemblies that do have their own context, I moved it under the assembly title as per the assembly template.

Modules:

* I only corrected markup/placement that went against the templates. I did not correct tech writing elements, such as "Start each step with an active verb", even though that is written in the procedure template. I might consider doing that in later PRs.

* Some of the external services procedures were a bit convoluted with sub-headings and such, so I didn't change those, but may do so in the future.

* I'd like to address the structure/markup of proc_configuring-repositories.adoc and proc_installing-the-satellite-server-packages.adoc (there's no .Procedure markup, and proc_configuring-repositories-el.adoc and similar files should be snippet files), but I haven't yet as I see there's an open PR adding a similar structure for installing capsule packages, so maybe I'll do all these sections at a later date.

Cherry-pick into:

### DIFF
--- a/guides/common/assembly_configuring-keycloak-external-authentication-with-totp.adoc
+++ b/guides/common/assembly_configuring-keycloak-external-authentication-with-totp.adoc
@@ -1,8 +1,9 @@
 ifdef::context[:parent-context: {context}]
-:context: keycloak-totp
 
 [id="configuring-keycloak-authentication-with-totp_{context}"]
 = Configuring {Keycloak} Authentication with TOTP
+
+:context: keycloak-totp
 
 Use this section to configure {Project} to use {Keycloak} as an OpenID provider for external authentication with TOTP cards.
 
--- a/guides/common/assembly_configuring-an-alternate-cname.adoc
+++ b/guides/common/assembly_configuring-an-alternate-cname.adoc
@@ -1,5 +1,8 @@
-[id='configuring-an-alternate-cname_{context}']
-= Configure an Alternate CNAME for {Project}
+ifdef::context[:parent-context: {context}]
+
+[id="configuring-an-alternate-cname_{context}"]
+= Configuring an Alternate CNAME for {Project}
+
 You can configure an alternate CNAME for {Project}.
 This might be useful if you want to deploy the {Project} web interface on a different domain name than the one that is used by client systems to connect to {Project}.
 You must plan the alternate CNAME configuration in advance prior to installing {SmartProxies} and registering hosts to {Project} to avoid redeploying new certificates to hosts.
@@ -10,3 +13,6 @@ endif::[]
 
 include::modules/proc_configuring-project-with-an-alternate-cname.adoc[leveloffset=+1]
 include::modules/proc_configuring-hosts-to-use-an-alternate-cname-for-content-management.adoc[leveloffset=+1]
+
+ifdef::parent-context[:context: {parent-context}]
+ifndef::parent-context[:!context:]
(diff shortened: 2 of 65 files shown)


guides/common/assembly_configuring-keycloak-external-authentication-with-totp.adoc
```diff
@@ -1,9 +1,10 @@
 ifdef::context[:parent-context: {context}]
-:context: keycloak-totp
 
 [id="configuring-keycloak-authentication-with-totp_{context}"]
 = Configuring {Keycloak} Authentication with TOTP
 
+:context: keycloak-totp
```
melcorr, line 6: Should we not define the context above where the context will be invoked? To me this makes no sense to define it after it is called in the title. I think this is problematic?

melcorr (commented):
Just one comment. I might be wrong.

guides/common/assembly_configuring-keycloak-external-authentication-with-totp.adoc
tahliar: Yeah this is one of the quirks of nesting assemblies (for simplicity I'm referring to the master.adoc file as the top assembly).

The key here is that the context defined in the assembly is only used for the modules included in that assembly, just as the context in the master.adoc file is used for the assemblies included in that master.adoc file. 

So, this assembly should build its ID from the context of the assembly it's included in (master.adoc), and then define its own context for its included modules to build their IDs with. So, using the :parent-context: ifdefs to preserve and then reinstate the top assembly's context, it looks something like this (at least for visualisation purposes): 

```
Master Assembly (master.adoc)
:context: master

  include::assembly1.adoc
    ID=assembly1_master
    Assembly 1
    :context: assembly1
  
      include::module1.adoc
        ID=module1_assembly1
        Module 1
    
      include::module2.adoc
        ID=module2_assembly1
        Module 2
    
  include::assembly2.adoc
    ID=assembly2_master
    Assembly 2
    :context: assembly2
    
      include::module1.adoc
        ID=module1_assembly2
        Module 1
        
      include::module3.adoc
        ID=module3_assembly2
        Module 3
```
    
I hope that makes sense and isn't just gibberish! It's described (probably better) here: https://redhat-documentation.github.io/modular-docs/#nesting-assemblies

melcorr: Thank you for explaining this to me!

melcorr (approved):
Thank you so much for this. 
I would really appreciate if you removed some of the nesting in some of our guides if you ever feel like it. 
What I would really love is to rip out chapter 13 of Admin guide into an Auth guide. I was looking at it today and it is so difficult to browse because of the levels of nesting. Each subsection should almost be a chapter on its own. 
But anyways, I so appreciate your contributions to the project.

PR #456, opened 2021-03-12
Add line to enable PostgreSQL 12 module on EL8

ehelms:
Cherry-pick into:

* [x] Foreman 2.4 / Katello 4.0

### DIFF
--- a/guides/common/modules/proc_configuring-repositories.adoc
+++ b/guides/common/modules/proc_configuring-repositories.adoc
@@ -58,6 +58,23 @@ ifdef::katello[]
 # {package-manager} config-manager --set-enabled powertools
 ----
 endif::[]
+ifdef::foreman-el,katello[]
++
+. Enable the PostgreSQL 12 module:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {package-manager} module enable postgresql:12
+----
++
+. If the PostgreSQL 10 module has already been enabled, a module reset will need to be performed.
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {package-manager} module reset postgresql
+# {package-manager} module enable postgresql:12
+----
+endif::[]
 
 ifdef::foreman-el,katello,satellite[]
 == [[rhel-7]]Red Hat Enterprise Linux 7


ehelms:
CC @ekohl

ehelms:
CC @ekohl

guides/common/modules/proc_configuring-repositories.adoc
```diff
@@ -58,6 +58,15 @@ ifdef::katello[]
 # {package-manager} config-manager --set-enabled powertools
 ----
 endif::[]
+ifdef::foreman-el,katello[]
```
ekohl, line 61: I'm not sure if this needs a guard on katello. IMHO this is valid for vanilla Foreman on EL8 as well.

guides/common/modules/proc_configuring-repositories.adoc
```diff
@@ -58,6 +58,15 @@ ifdef::katello[]
...
++
+. Enable the PostgreSQL 12 module:
```
ekohl, line 63: I wonder if we need some reference documentation on what to do if a different module is already enabled etc.

guides/common/modules/proc_configuring-repositories.adoc
ehelms: That's not a guard, but rather a "which guides to include this in" line. So in this case, include it in the Foreman Enterprise Linux guide and the Katello guide.

guides/common/modules/proc_configuring-repositories.adoc
ehelms: Added a note

lzap:
I am just about to branch off 2.4 series, this is something that should go into 2.4/4.0. Can we merge now?

ehelms:
> I am just about to branch off 2.4 series, this is something that should go into 2.4/4.0. Can we merge now?

This applies only to nightly and Foreman 2.5+.

lzap:
Thank you!

guides/common/modules/proc_configuring-repositories.adoc
ekohl: So I read `ifdef` as an AND, not an OR. IMHO this should be an OR. If it is, then consider this addressed.

ekohl (approved):
If `ifdef` works as an OR then :+1: on this.

Since this is the setup guide, I think it wouldn't be bad to do this even on a released version (2.4). Upgrading users shouldn't read it. Plus, the whole docs.theforeman.org is still considered in development. Regular Foreman users are probably not reading it as a primary source yet.

lzap:
> If `ifdef` works as an OR then 👍 on this.

That is my understanding as well but frankly I haven't tried to verify this myself.

ehelms:
> If `ifdef` works as an OR then  on this.

`ifdef` can work as AND or OR:

OR:
```
ifdef::foreman-el,katello[]
```

AND:
```
ifdef::foreman-el+katello[]
```

Ref: https://docs.asciidoctor.org/asciidoc/latest/directives/ifdef-ifndef/#checking-multiple-attributes

melcorr:
Can I merge this?

ekohl:
I think it can also be cherry picked into the 2.4 branch. @ehelms?

ehelms:
> I think it can also be cherry picked into the 2.4 branch. @ehelms?

Yes, it can apply there as well

ekohl:
How do we deal with cherry picks in this repo? Is that a separate PR? If so, who's supposed to create it?

melcorr:
Me - I'll do it !

PR #453, opened 2021-03-12
Add Application centric deployment guide

maximiliankolb:
Hi everyone!

This PR provides a guide for **Application Centric Deployment** via the [Foreman ACD](https://github.com/ATIX-AG/foreman_acd) & [Smart Proxy ACD](https://github.com/ATIX-AG/smart_proxy_acd) plugins.

Along with the content, there are also commits to allow building a new "orcharhino" target, including a `.css` file, attributes, and a small extension to the README. I would love some feedback if anything else can be/needs to be adjusted.

Feedback is very welcome. 😄 

Thanks.
Maximilian.

PR #460, opened 2021-03-18
Big cleanup of attributes

lzap:
I know this is a change that will affect all of us, will break existing PRs and cause some cherry picking change but overall I think it's an important change to make rather sooner than later as another build flag is pending merge (https://github.com/theforeman/foreman-documentation/pull/453) and the attributes file is getting out of control.

**The problem**

As we migrated form Satellite docs, we started with Satellite attributes and enclosed them in if-statement and made a copy for "foreman" build. Then we added "katello" build, then we broke "foreman" into two and now there's another one incoming soon. Some if-blocks are copies (e.g. satellite vs foreman), other if-blocks are just overriding other values.

We need to have a consistent way of defining attributes and overriding them.

**The proposal**

Let's break huge attributes.adoc file into several files:

* attributes-base.adoc: The base file, all attributes which are supposed to be overridden should be here, with exception of attributes which are not supposed to be overridden per BUILD (e.g. versions).
* attributes-BUILD.adoc: Override files, for each BUILD a one. Only attributes which are defined in attributes.adoc can be here, so we make sure that someone guide is not defining own sandbox of attributes and everything is shared across the whole project. Builds can also define their own attributes visible only for that particular build.
* attributes.adoc: The main file which is included from guides, contains only the basic attributes (versions) and includes.

Initially, I thought it would be good idea to keep all attribute files sorted, but it turned out to be problematic as some attributes depend on others. Also it is useful to create groups of attributes, for example:

* BaseURL + URLs we use for xrefs.
* Own attributes visible only for the particular build.
* Overriden base attributes.

What is important is that instead of overriding satellite attributes, this patch turns it the other way around to the more logical: **base attributes are upstream attributes**, satellite overrides just like other builds. This makes much more sense to me.

**How to test this patch** (BEFORE YOU MERGE)

We must absolutely make sure this does not break our docs, so before merging MAKE SURE to build documentation **without** this patch, copy HTML pages to a safe place, then checkout this branch, rebase, make clean and generate all HTML pages. Then **compare** to the original and there must be exactly ZERO changes as there are no text changes in this patch on purpose.

**The oritinal proposal**

_Since I changed my mind after one night, I am keeping the original proposal here just for the record, skip to the next section. In short: I proposed some very strict rules about attributes and alphabetical order which turned not very practical._

Let's break huge attributes.adoc file into several files:

* attributes-base.adoc: The base file, ALL attributes ever defined must be here. It's the file which is ALWAYS included no matter which build is done. With exception of attributes which are not supposed to be overriden per BUILD (e.g. versions).
* attributes-BUILD.adoc: Override files, for each BUILD a one. Only attributes which are defined in attributes.adoc can be here, so we make sure that someone guide is not defining own sandbox of attributes and everything is shared across the whole project.
* attributes.adoc: The main file which is included from guides, contains only the basic attributes (versions) and includes.

What is important is that instead of overriding satellite attributes, this patch turns it the other way around to the more logical: base attributes are upstream attributes, satellite overrides many of them. This makes much more sense to me.

I do not want to stop there, if we are okay with this approach, I would like to write a script that will tighten up even more how we treat attributes and build would fail unless the following conditions are met:

* Attributes in all attribute-*.adoc files must be in the aplhanumeric order.
* Attributes overriden must be always present in the attributes-base.adoc file.
* If overriden attribute has the same value as the base one, it must be removed (no useless copies).
* Attributes referenced in attributes-*.adoc files (e.g. BaseURL) must be defined in the main attributes.adoc file to prevent ordering problems.
* In the future more rules (e.g. all downcase with dash characters).

I did split and review all attributes, then I sorted all files alphabetically, then I used a diff tool to remove those which are just copies (the value is the same). I also found one that is only defined in one BUILD but it should be present in the attributes-base.adoc (there is FIXME comment there - I do not want to change any of the content in this patch so will fix later).

### DIFF
--- a/guides/common/attributes-foreman-deb.adoc
+++ b/guides/common/attributes-foreman-deb.adoc
@@ -1,0 +1,30 @@
+// URLs
+:BaseFilenameURL: index-{build}.html
+:AdministeringDocURL: {BaseURL}Administering_Red_Hat_Satellite/{BaseFilenameURL}#
+:ConfiguringAnsibleDocURL: {BaseURL}Configuring_Ansible/{BaseFilenameURL}#
+:ConfiguringLoadBalancerDocURL: {BaseURL}Configuring_Load_Balancer/{BaseFilenameURL}#
+:ContentManagementDocURL: {BaseURL}Content_Management_Guide/{BaseFilenameURL}#
+:InstallingProjectDocURL: {BaseURL}Installing_Server_on_Red_Hat/{BaseFilenameURL}#
+:InstallingSmartProxyDocURL: {BaseURL}Installing_Proxy_on_Red_Hat/{BaseFilenameURL}#
+:ManagingHostsDocURL: {BaseURL}Managing_Hosts/{BaseFilenameURL}#
+:PlanningDocURL: {BaseURL}Planning_Guide/{BaseFilenameURL}#
+:ProvisioningDocURL: {BaseURL}Provisioning_Guide/{BaseFilenameURL}#
+
+// Attributes only for foreman-deb build
+:smart-proxy-installation-guide-title: Installing an External Smart Proxy Server on Debian
+
+// Overrides for foreman-deb build
+:foreman-installer-package: foreman-installer
+:installer-scenario-smartproxy: foreman-installer --no-enable-foreman
+:installer-scenario-smartproxy: foreman-installer --no-enable-foreman
+:installer-scenario: foreman-installer
+:installer-scenario: foreman-installer --scenario foreman
+:package-clean: apt-get clean
+:package-install-project: apt-get install
+:package-install: apt-get install
+:package-remove-project: apt-get remove
+:package-remove: apt-get remove
+:package-update-project: apt-get upgrade
+:package-update: apt-get upgrade
+:project-installation-guide-title: Installing Foreman {ProjectVersion} server on Enterprise Linux
+:project-installation-guide-title: Installing Foreman server on Debian
--- a/guides/common/attributes.adoc
+++ b/guides/common/attributes.adoc
@@ -1,8 +1,8 @@
-// NOTE: Avoid empty lines so it can be included in the header section
 // Document state: "nightly" for master, "stable" for last two releases,
 // "unsupported" for the rest and "satellite" for satellite build
 :DocState: nightly
-// Versions used in text and code
+
+// Version numbers
 :ProjectVersion: 2.4
 :KatelloVersion: 4.0
 :TargetVersion: 6.8
@@ -10,25 +10,20 @@
 // The above attribute should point to the GA version number (x.y) for all releases including Beta
 :ProductVersion: 6.9-beta
 :ProductVersionPrevious: 6.8
-:ProductVersionRepoTitle: Beta
-// For Beta, change to "Beta". For GA releases, change to, for example, "6.8".
-:RepoRHEL7ServerSatelliteServerProductVersion: rhel-server-7-satellite-6-beta-rpms
-:RepoRHEL7ServerSatelliteServerProductVersionPrevious: rhel-7-server-satellite-6.8-rpms
-:RepoRHEL7ServerSatelliteCapsuleProductVersion: rhel-server-7-satellite-capsule-6-beta-rpms
-:RepoRHEL7ServerSatelliteToolsProductVersion: rhel-7-server-satellite-tools-6-beta-rpms
-:RepoRHEL7ServerSatelliteMaintenanceProductVersion: rhel-7-server-satellite-maintenance-6-beta-rpms
-:RepoRHEL7ServerSatelliteServerPuppetVersion: rhel-7-server-satellite-6.3-puppet4-rpms
-:RepoRHEL7ServerSatelliteCapsulePuppetVersion: rhel-7-server-satellite-capsule-6.3-puppet4-rpms
-//Do not update the puppet4 repo versions. They must stay at 6.3.
-:SatelliteSub: Red Hat Satellite Infrastructure Subscription
-// Change to "Red Hat Satellite Infrastructure Subscription (Beta)" for beta releases
-:RepoRHEL7Server: rhel-7-server-rpms
-:RepoRHEL7ServerSoftwareCollections: rhel-server-rhscl-7-rpms
-:RepoRHEL7ServerOptional: rhel-7-server-optional-rpms
-:RepoRHEL7ServerAnsible: rhel-7-server-ansible-2.9-rpms
 :SatelliteAnsibleVersion: 2.9
+// For Package Manifests etc that will fail the upstream link-checker during the beta see Issue #115 on GitHub
 :SpecialCaseProductVersion: 6.8
-//the above attribute is for Package Manifests etc that will fail the upstream link-checker during the beta see Issue #115 on GitHub
+
+//
+// WHERE ARE MY ATTRIBUTES?
+//
+// All attributes were moved into separate files:
+// * attributes-base.adoc (upstream attributes common for all builds)
+// * attributes-foreman-el.adoc (attributes overridden or unique to Red Hat Enteprise Linux and clones)
+// * attributes-foreman-deb.adoc (attributes overridden or unique to Debian/Ubuntu)
+// * attributes-katello.adoc (attributes overridden or unique to Katello)
+// * attributes-satellite.adoc (attributes overridden or unique to Satellite)
+
 // Define properties to represent each build. Allows doing 'or' and 'and' operations for conditions.
 ifeval::["{build}" == "foreman-el"]
 :foreman-el:
@@ -45,242 +40,24 @@ ifeval::["{build}" == "satellite"]
 :satellite:
 endif::[]
 
-ifdef::katello,foreman-el[]
-:smart-proxy-installation-guide-title: Installing an External Smart Proxy Server {ProjectVersion}
-:smart-proxy-context: smart-proxy
-:project-context: foreman
-:foreman-installer: foreman-installer
-:foreman-maintain: foreman-maintain
-:foreman-example-com: foreman.example.com
-:package-install: yum install
-:package-update: yum update
-:package-clean: yum clean
-:package-remove: yum remove
-:package-install-project: yum install
-:package-update-project: yum update
-:package-remove-project: yum remove
-:certs-generate: foreman-proxy-certs-generate
-:certs-proxy-context: foreman-proxy
-:project-change-hostname: katello-change-hostname
-:oVirt: oVirt
-:oVirtShort: oVirt
-:oVirtEngine: oVirt Engine
-:LoraxCompose: Lorax Composer
-:Cockpit: Cockpit
-:ovirt-example-com: ovirt.example.com
-:KubeVirt: KubeVirt
-:OpenStack: OpenStack
-:FreeIPA: FreeIPA
-:Keycloak: Keycloak
-:Keycloak-short: Keycloak
-:PIV: PIV
-:ProjectNameXY: Foreman{nbsp}1.22
-:ProjectNameX: Foreman
-:ProjectName: Foreman
-:ProjectXY: Foreman{nbsp}1.22
-:ProjectX: Foreman
-:ProjectServer: Foreman{nbsp}server
-:Project: Foreman
-:Project_Link: Red_Hat_Satellite
-:provision-script: OS installer recipe
-:RHEL: Red{nbsp}Hat Enterprise Linux
-:RHELServer: Red{nbsp}Hat Enterprise Linux Server
-:SmartProxyServer: Smart{nbsp}Proxy{nbsp}server
-:SmartProxies: Smart{nbsp}Proxies
-:SmartProxy: Smart{nbsp}Proxy
-:smartproxy-example-com: smartproxy.example.com
-:smartproxy_port: 8443
-:Team: Foreman developers
-:project-client-RHEL7-url: https://yum.theforeman.org/client/{ProjectVersion}/el7/x86_64/foreman-client-release.rpm
-:project-client-name: https://yum.theforeman.org/client/{ProjectVersion}/
-:customcontent: content
-:customproduct: product
-:customproducttitle: Product
-:customgpgtitle: GPG
-:customssltitle: SSL
-:customssl: SSL
-:customrpmtitle: RPM
-:customrpm: RPM
-:customrepo: repository
-:customfiletypetitle: File Type
-:customfiletype: file type
-:customostreecontenttitle: OSTree Content
-:customostreecontent: OSTree content
-:ansiblefilepath: /usr/share/ansible/collections/ansible_collections/theforeman/foreman/plugins/modules/
-:ansible-doc-activation_key: ansible-doc theforeman.foreman.activation_key
-:ansible-galaxy: https://galaxy.ansible.com/theforeman/foreman
-:ansible-namespace: `theforeman.foreman`
-:ansible-namespace-example: `theforeman.foreman._module_name_`
-:awx: AWX
+// Load base attributes
 :BaseURL: https://docs.theforeman.org/nightly/
-:ManagingHostsDocURL: {BaseURL}Managing_Hosts/index-foreman-el.html#
-:ConfiguringAnsibleDocURL: {BaseURL}Configuring_Ansible/index-foreman-el.html#
-:AdministeringDocURL: {BaseURL}Administering_Red_Hat_Satellite/index-foreman-el.html#
-:InstallingSmartProxyDocURL: {BaseURL}Installing_Proxy_on_Red_Hat/index-foreman-el.html#
-:InstallingProjectDocURL: {BaseURL}Installing_Server_on_Red_Hat/index-foreman-el.html#
-:ContentManagementDocURL: {BaseURL}Content_Management_Guide/index-foreman-el.html#
-:ConfiguringLoadBalancerDocURL: {BaseURL}Configuring_Load_Balancer/index-foreman-el.html#
-:PlanningDocURL: {BaseURL}Planning_Guide/index-foreman-el.html#
-:ProvisioningDocURL: {BaseURL}Provisioning_Guide/index-foreman-el.html#
-:UpgradingDocURL: {BaseURL}Upgrading_and_Updating/index-foreman-el.html#
-endif::[]
-ifdef::satellite[]
-:DocState: satellite
-:project-installation-guide-title: Installing Satellite Server from a Connected Network
-:smart-proxy-installation-guide-title: Installing Capsule Server
-:smart-proxy-context: capsule
-:project-context: satellite
-:foreman-installer: satellite-installer
-:foreman-installer-package: satellite-installer
-:installer-scenario: satellite-installer --scenario satellite
-:installer-scenario-smartproxy: satellite-installer --scenario capsule
-:package-install: yum install
-:package-update: yum update
-:package-clean: yum clean
-:package-remove: yum remove
-:package-install-project: satellite-maintain packages install
-:package-update-project: satellite-maintain packages update
-:package-remove-project: satellite-maintain packages remove
-:certs-generate: capsule-certs-generate
-:certs-proxy-context: capsule
-:foreman-maintain: satellite-maintain
-:foreman-example-com: satellite.example.com
-:project-change-hostname: satellite-change-hostname
-:LoraxCompose: Red{nbsp}Hat Image Builder
-:Cockpit: Red{nbsp}Hat web console
-:oVirt: Red{nbsp}Hat{nbsp}Virtualization
-:oVirtShort: RHV
-:oVirtEngine: Red{nbsp}Hat Virtualization Manager
-:ovirt-example-com: rhv.example.com
-:KubeVirt: Container-native Virtualization
-:OpenStack: Red{nbsp}Hat OpenStack Platform
-:FreeIPA: Red{nbsp}Hat Identity Management
-:Keycloak: Red{nbsp}Hat Single Sign-On
-:Keycloak-short: RHSSO
-:PIV: CAC
-:ProjectNameXY: Red{nbsp}Hat Satellite{nbsp}{ProductVersionRepoTitle}
-:ProjectNameX: Red{nbsp}Hat Satellite{nbsp}6
-:ProjectName: Red{nbsp}Hat Satellite
-:ProjectXY: Satellite{nbsp}{ProductVersionRepoTitle}
-:ProjectX: Satellite{nbsp}6
-:ProjectServer: Satellite{nbsp}Server
-:Project: Satellite
-:Project_Link: Red_Hat_Satellite
-:provision-script: kickstart
-:RHEL: Red{nbsp}Hat Enterprise Linux
-:RHELServer: Red{nbsp}Hat Enterprise Linux Server
-:SmartProxyServer: Capsule{nbsp}Server
-:SmartProxies: Capsules
-:SmartProxy: Capsule
-:smartproxy-example-com: capsule.example.com
-:smartproxy_port: 9090
-:Team: Red{nbsp}Hat
-:project-client-RHEL7-url: {RepoRHEL7ServerSatelliteToolsProductVersion}
-:project-client-name: Satellite Tools {ProductVersionRepoTitle}
-:customcontent: custom content
-:customproduct: custom product
-:customproducttitle: Custom Product
-:customgpgtitle: Custom GPG
-:customssltitle: Custom SSL
-:customssl: custom SSL
-:customrpmtitle: Custom RPM
-:customrpm: custom RPM
-:customrepo: custom repository
-:customfiletypetitle: Custom File Type
-:customfiletype: custom file type
-:customostreecontenttitle: Custom OSTree Content
-:customostreecontent: custom OSTree content
-:ansiblefilepath: /usr/share/ansible/collections/ansible_collections/redhat/satellite/plugins/modules/
-:ansible-doc-activation_key: ansible-doc redhat.satellite.activation_key
-:ansible-galaxy: https://cloud.redhat.com/ansible/automation-hub/redhat/satellite/docs
-:ansible-namespace: `redhat.satellite`
-:ansible-namespace-example: `redhat.satellite._module_name_`
-:awx: Ansible Tower
-:BaseURL: https://access.redhat.com/documentation/en-us/red_hat_satellite/{ProductVersion}/html-single/
-:ManagingHostsDocURL: {BaseURL}managing_hosts/index#
-:ConfiguringAnsibleDocURL: {BaseURL}configuring_satellite_to_use_ansible/index#
-:AdministeringDocURL: {BaseURL}administering_red_hat_satellite/index#
-:InstallingSmartProxyDocURL: {BaseURL}installing_capsule_server/index#
-:InstallingProjectDocURL: {BaseURL}installing_satellite_server_from_a_connected_network/index#
-:ContentManagementDocURL: {BaseURL}content_management_guide/index#
-:ConfiguringLoadBalancerDocURL: {BaseURL}configuring_capsules_with_a_load_balancer/index#
-:PlanningDocURL: {BaseURL}planning_for_red_hat_satellite/index#
-:ProvisioningDocURL: {BaseURL}provisioning_guide/index#
-:UpgradingDocURL: {BaseURL}upgrading_and_updating_red_hat_satellite/index#
+include::attributes-base.adoc[]
+
+// Load overrides for specific scenarios
+ifdef::foreman-el[]
+include::attributes-foreman-el.adoc[]
 endif::[]
 
 ifdef::foreman-deb[]
-:smart-proxy-context: smart-proxy
-:project-context: foreman
-:project-installation-guide-title: Installing Foreman server on Debian
-:smart-proxy-installation-guide-title: Installing an External Smart Proxy Server on Debian
-:foreman-installer: foreman-installer
-:foreman-maintain: foreman-maintain
-:foreman-example-com: foreman.example.com
-:installer-scenario: foreman-installer
-:package-install: apt-get install
-:package-update: apt-get upgrade
-:package-clean: apt-get clean
-:package-remove: apt-get remove
-:package-install-project: apt-get install
-:package-update-project: apt-get upgrade
-:package-remove-project: apt-get remove
-:certs-generate: foreman-proxy-certs-generate
-:installer-scenario-smartproxy: foreman-installer --no-enable-foreman
-:project-change-hostname: katello-change-hostname
-:oVirt: oVirt
-:oVirtShort: oVirt
-:oVirtEngine: oVirt Engine
-:ovirt-example-com: ovirt.example.com
-:LoraxCompose: Lorax Composer
-:Cockpit: Cockpit
-:KubeVirt: KubeVirt
-:OpenStack: OpenStack
-:FreeIPA: FreeIPA
-:Keycloak: Keycloak
-:Keycloak-short: Keycloak
-:PIV: PIV
-:ProjectNameXY: Foreman{nbsp}1.22
-:ProjectNameX: Foreman
-:ProjectName: Foreman
-:ProjectXY: Foreman{nbsp}1.22
-:ProjectX: Foreman
-:ProjectServer: Foreman{nbsp}server
-:Project: Foreman
-:Project_Link: Red_Hat_Satellite
-:provision-script: OS installer recipe
-:RHEL: Red{nbsp}Hat Enterprise Linux
-:RHELServer: Red{nbsp}Hat Enterprise Linux Server
-:SmartProxyServer: Smart{nbsp}Proxy{nbsp}server
-:SmartProxies: Smart{nbsp}Proxies
-:SmartProxy: Smart{nbsp}Proxy
-:smartproxy-example-com: smartproxy.example.com
-:smartproxy_port: 8443
-:Team: Foreman developers
-:project-client-RHEL7-url: https://yum.theforeman.org/client/{ProjectVersion}/el7/x86_64/foreman-client-release.rpm
-:project-client-name: https://yum.theforeman.org/client/{ProjectVersion}/
-:BaseURL: https://docs.theforeman.org/nightly/
-:ManagingHostsDocURL: {BaseURL}Managing_Hosts/index-foreman-deb.html#
-:ConfiguringAnsibleDocURL: {BaseURL}Configuring_Ansible/index-foreman-deb.html#
-:AdministeringDocURL: {BaseURL}Administering_Red_Hat_Satellite/index-foreman-deb.html#
-:InstallingSmartProxyDocURL: {BaseURL}Installing_Proxy_on_Red_Hat/index-foreman-deb.html#
-:InstallingProjectDocURL: {BaseURL}Installing_Server_on_Red_Hat/index-foreman-deb.html#
-:ContentManagementDocURL: {BaseURL}Content_Management_Guide/index-foreman-deb.html#
-:ConfiguringLoadBalancerDocURL: {BaseURL}Configuring_Load_Balancer/index-foreman-deb.html#
-:PlanningDocURL: {BaseURL}Planning_Guide/index-foreman-deb.html#
-:ProvisioningDocURL: {BaseURL}Provisioning_Guide/index-foreman-deb.html#
+include::attributes-foreman-deb.adoc[]
 endif::[]
-// Overrides for specific scenarios from the defaults
-ifdef::foreman-deb,foreman-el[]
-:foreman-installer-package: foreman-installer
-:project-installation-guide-title: Installing Foreman {ProjectVersion} server on Enterprise Linux
-:installer-scenario: foreman-installer --scenario foreman
-:installer-scenario-smartproxy: foreman-installer --no-enable-foreman
-endif::[]
+
 ifdef::katello[]
-:foreman-installer-package: foreman-installer-katello
-:project-installation-guide-title: Installing Foreman {ProjectVersion} server with Katello {KatelloVersion} plugin on Enterprise Linux
-:installer-scenario: foreman-installer --scenario katello
-:installer-scenario-smartproxy: foreman-installer --scenario foreman-proxy-content
-:InstallingProjectDocURL: {BaseURL}Installing_Server_on_Red_Hat/index-katello.html#
+include::attributes-katello.adoc[]
+endif::[]
+
+ifdef::satellite[]
+:BaseURL: https://access.redhat.com/documentation/en-us/red_hat_satellite/{ProductVersion}/html-single/
+include::attributes-satellite.adoc[]
 endif::[]
--- a/guides/.gitignore
+++ b/guides/.gitignore
@@ -1,6 +1,8 @@
 /build/
+/build-*/
 /common/build.adoc
 /upstreamize_report.md
 .alink
 /scripts/docs-Red_Hat_Satellite_6/
 /scripts/foreman-documentation/
+*.diff
--- a/guides/README.md
+++ b/guides/README.md
@@ -19,11 +19,11 @@ In MacOS required tools can be installed via brew but instead "make" call "gmake
 
 Alternatively, XCode development environment can be installed to have make utility available on PATH, however this takes about an hour to download and install and requires several gigabytes of HDD space:
 
-	xcode-select --install
+		xcode-select --install
 
 If AsciiDoctor is not available in repositories or under RVM/rbenv, simply install it from rubygems:
 
-  	gem install asciidoctor asciidoctor-pdf --pre
+		gem install asciidoctor asciidoctor-pdf --pre
 
 Then simply run `make` or `make html` which builds HTML artifacts.
 Generating PDF output is slow, therefore command `make pdf` must be used separately.
@@ -49,7 +49,6 @@ Currently there are three different versions:
 `make BUILD=katello` - this generates a katello build of the guide
 
 `make BUILD=foreman-deb` - This generates  a version for Foreman installed on Debian.
-
 
 The final artifacts can be found in the ./build subdirectory.
 Note that GNU Makefile tracks changes and only builds relevant artifacts, to trigger full rebuild use `make clean` to delete build directory and start over.
@@ -94,7 +93,14 @@ Therefore, never write "Foreman" or "Satellite" words directly, but use the foll
 | {SmartProxy} | Smart Proxy | Capsule |
 
 The table only covers the most frequent terms.
-The rest are defined in the [attributes file](common/attributes.adoc).
+The rest are defined in the attribute files:
+
+* [attributes.adoc](common/attributes.adoc): version definitions and includes for other attribute files.
+* [attributes-base.adoc](common/attributes-base.adoc): base attributes common for all builds.
+* [attributes-foreman-el.adoc](common/attributes-foreman-el.adoc): base overrides for foreman-el build.
+* [attributes-foreman-deb.adoc](common/attributes-foreman-deb.adoc): base overrides for foreman-deb build.
+* [attributes-katello.adoc](common/attributes-katello.adoc): base overrides for katello build.
+* [attributes-satellite.adoc](common/attributes-satellite.adoc): base overrides for satellite build.
 
 Variables cannot be used in shell or code examples.
 To use them, use "attributes" keyword:
--- a/guides/common/attributes-base.adoc
+++ b/guides/common/attributes-base.adoc
@@ -1,0 +1,96 @@
+// URLs
+:AdministeringDocURL: {BaseURL}Administering_Red_Hat_Satellite/index-foreman-el.html#
+:ConfiguringAnsibleDocURL: {BaseURL}Configuring_Ansible/index-foreman-el.html#
+:ConfiguringLoadBalancerDocURL: {BaseURL}Configuring_Load_Balancer/index-foreman-el.html#
+:ContentManagementDocURL: {BaseURL}Content_Management_Guide/index-foreman-el.html#
+:InstallingProjectDocURL: {BaseURL}Installing_Server_on_Red_Hat/index-foreman-el.html#
+:InstallingSmartProxyDocURL: {BaseURL}Installing_Proxy_on_Red_Hat/index-foreman-el.html#
+:ManagingHostsDocURL: {BaseURL}Managing_Hosts/index-foreman-el.html#
+:PlanningDocURL: {BaseURL}Planning_Guide/index-foreman-el.html#
+:ProvisioningDocURL: {BaseURL}Provisioning_Guide/index-foreman-el.html#
+:UpgradingDocURL: {BaseURL}Upgrading_and_Updating/index-foreman-el.html#
+
+// Repositories and subscriptions (used both in Katello and Satellite guides)
+:SatelliteSub: Red Hat Satellite Infrastructure Subscription
+:RepoRHEL7Server: rhel-7-server-rpms
+:RepoRHEL7ServerSoftwareCollections: rhel-server-rhscl-7-rpms
+:RepoRHEL7ServerOptional: rhel-7-server-optional-rpms
+:RepoRHEL7ServerAnsible: rhel-7-server-ansible-2.9-rpms
+// For Beta, change to "Beta". For GA releases, change to, for example, "6.8".
+:RepoRHEL7ServerSatelliteServerProductVersion: rhel-server-7-satellite-6-beta-rpms
+:RepoRHEL7ServerSatelliteServerProductVersionPrevious: rhel-7-server-satellite-6.8-rpms
+:RepoRHEL7ServerSatelliteCapsuleProductVersion: rhel-server-7-satellite-capsule-6-beta-rpms
+:RepoRHEL7ServerSatelliteToolsProductVersion: rhel-7-server-satellite-tools-6-beta-rpms
+:RepoRHEL7ServerSatelliteMaintenanceProductVersion: rhel-7-server-satellite-maintenance-6-beta-rpms
+// Do not update the puppet4 repo versions. They must stay at 6.3.
+:RepoRHEL7ServerSatelliteServerPuppetVersion: rhel-7-server-satellite-6.3-puppet4-rpms
+:RepoRHEL7ServerSatelliteCapsulePuppetVersion: rhel-7-server-satellite-capsule-6.3-puppet4-rpms
+
+// Base attributes
+:ansible-doc-activation_key: ansible-doc theforeman.foreman.activation_key
+:ansible-galaxy: https://galaxy.ansible.com/theforeman/foreman
+:ansible-namespace-example: `theforeman.foreman._module_name_`
+:ansible-namespace: `theforeman.foreman`
+:ansiblefilepath: /usr/share/ansible/collections/ansible_collections/theforeman/foreman/plugins/modules/
+:awx: AWX
+:certs-generate: foreman-proxy-certs-generate
+:certs-proxy-context: foreman-proxy
+:Cockpit: Cockpit
+:customcontent: content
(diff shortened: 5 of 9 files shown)


lzap:
Ah interesting problem, when attribute files are sorted, this broke dependencies :-)

Some attributes use others, therefore this introduced regressions. E.g. BaseURL:

```
common/attributes-satellite.adoc
1::AdministeringDocURL: {BaseURL}administering_red_hat_satellite/index#
8::BaseURL: https://access.redhat.com/documentation/en-us/red_hat_satellite/{ProductVersion}/html-single/
12::ConfiguringAnsibleDocURL: {BaseURL}configuring_satellite_to_use_ansible/index#
...
```

I will move BaseURL to the top attributes.adoc file then.

I will add this as a TODO item to implement in the check script to prevent this in the future.

guides/common/attributes.adoc
```diff
@@ -1,16 +1,18 @@
...
 :DocState: nightly
+
 // Versions used in text and code
 :ProjectVersion: 2.4
 :KatelloVersion: 4.0
 :TargetVersion: 6.8
 :TargetVersionMaintainUpgrade: 6.8
+
 // The above attribute should point to the GA version number (x.y) for all releases including Beta
 :ProductVersion: 6.9-beta
```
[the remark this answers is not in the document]
lzap: I think we should probably follow this pattern in this case:

```
:ProjectNameXY: Foreman{nbsp}1.22
:ProjectNameX: Foreman
:ProjectName: Foreman
:ProjectXY: Foreman{nbsp}1.22
:ProjectX: Foreman
:ProjectServer: Foreman{nbsp}server
:Project: Foreman
```

Therefore I would expect something like:

    :ProjectServerXY: Foreman{nbsp}1.22

Or better:

    :ProjectServerXY: Foreman{nbsp}{ProjectVersion}

I can fix that.

maximiliankolb:
> base attributes are upstream attributes, satellite overrides many of them. This makes much more sense to me.

Yes, 100% agree. This makes it easier for me: I can overwrite some attributes via an `attributes-orcharhino.adoc` file, but I don't get into trouble if I don't.

Thanks for handling this @lzap!

lzap:
I gave this a second thought and I slightly updated the proposal - I think fixed order is not very practical and a tool that would enforce some rules would be also probably for the worse. So I am just breaking attributes up into more logical BASE vs OVERRIDE files and cleaning them a bit. But other than that, this should still be a huge cleanup.

guides/common/attributes.adoc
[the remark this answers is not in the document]
lzap: Actually, I agree but I want to keep this PR without any changes so we can easily test it. Let's do this later.

ehelms:
This looks great, thanks for tackling @lzap . I have not had a chance to check through some generated guides yet but will do so later today.

guides/common/attributes.adoc
```diff
@@ -19,16 +23,16 @@
 :RepoRHEL7ServerSatelliteMaintenanceProductVersion: rhel-7-server-satellite-maintenance-6-beta-rpms
 :RepoRHEL7ServerSatelliteServerPuppetVersion: rhel-7-server-satellite-6.3-puppet4-rpms
 :RepoRHEL7ServerSatelliteCapsulePuppetVersion: rhel-7-server-satellite-capsule-6.3-puppet4-rpms
```
ekohl, line 25: It feels weird to have this. It's not used anywhere and 6.3 is EOL. Should this at least be moved to a Satellite attributes file that's only included in downstream builds? The moving part refers to all these things, all the way up to TargetVersion.

ekohl (commented):
In general this makes sense to me.

guides/common/attributes.adoc
lzap: Agreed, let me move all those RepoRHEL7* attributes into attributes-satellite.adoc alongside with SatelliteSub. These should not be used upstream at all.

lzap:
Rebased, this actually clears attributes.adoc completely, we only have versions and include statements there from now on :-)

Maybe I will drop a warning comment where to find those.

lzap:
I dropped a comment in the attributes.adoc explaining where attributes are now.

lzap:
This is now ready for your final blessings, this will hurt let's do it quick.

After this is merged I will drop a comment in all opened PRs about possible conflicts and what just happened.

guides/common/attributes.adoc
```diff
@@ -11,24 +11,20 @@
 :ProductVersion: 6.9-beta
 :ProductVersionPrevious: 6.8
 :ProductVersionRepoTitle: Beta
```
ekohl, line 13: Isn't this also a downstream only attribute that should live in Satellite's attributes?
```console
$ rg ProductVersionRepoTitle
guides/common/attributes.adoc
13::ProductVersionRepoTitle: Beta

74::project-client-name: Satellite Tools {ProductVersionRepoTitle}
82::ProjectNameXY: Red{nbsp}Hat Satellite{nbsp}{ProductVersionRepoTitle}
85::ProjectXY: Satellite{nbsp}{ProductVersionRepoTitle}
```

guides/common/attributes.adoc
```diff
@@ -11,24 +11,20 @@
...
+// For Package Manifests etc that will fail the upstream link-checker during the beta see Issue #115 on GitHub
 :SpecialCaseProductVersion: 6.8
-//the above attribute is for Package Manifests etc that will fail the upstream link-checker during the beta see Issue #115 on GitHub
+
+//
+// WHERE ARE MY ATTRIBUTES?
+//
+// All attributes were moved into separate files:
+// * attributes-base.adoc (upstream attributes common for all builds)
+// * attributes-foreman-el.adoc (attributes overridden or unique to the specific build)
```
ekohl, line 23: Does it make sense to describe the builds here or is that duplication?
```suggestion
// * attributes-foreman-el.adoc (attributes overridden or unique to the build for Foreman on Enterprise Linux)
```

guides/common/attributes.adoc
lzap: Yes, changing.

guides/common/attributes.adoc
lzap: Sure.

ehelms (approved):

melcorr:
Through my inbox after PTO, and testing this now!

guides/common/attributes-foreman-deb.adoc
```diff
@@ -0,0 +1,31 @@
...
+
+// Overrides for foreman-deb build
+:foreman-installer-package: foreman-installer
+:installer-scenario-smartproxy: foreman-installer --no-enable-foreman
+:installer-scenario-smartproxy: foreman-installer --no-enable-foreman
+:installer-scenario: foreman-installer
+:installer-scenario: foreman-installer --scenario foreman
+:package-clean: apt-get clean
+:package-install-project: apt-get install
+:package-install: apt-get install
```
melcorr, line 24: This appears in the planning guide unexpectedly? 

`# apt-get install grub2-efi`

melcorr (changes requested):
Hi @lzap 
I can follow the logic of this proposal, and do think we need to do *something* to keep this under control, so a +1 for this. 
On testing, most things work just fine. I found some broken things. I will try and dig into why they are broken a little later. Here I am noting what I have found for foreman-el build. 

Can you please add an entry to the readme about this. @maximiliankolb has been reminding me of all the tribal knowledge that we have not put anywhere. I think none of the RH writers have been around for a while either, so please update that so when they return, they will know what has been happening with this big change. 

These irregularities appear in the following guides:

"For more information about using an external database, see Using External Databases with Foreman in Installing Foreman server on Debian."

 For information about how to configure a Smart Proxy with external services, see Configuring Smart Proxy server with External Services in Installing an External Smart Proxy Server on Debian.

These are not working in the QuickStart Guide: 

Enable the following repositories:

# subscription-manager repos --enable={RepoRHEL7Server} \
--enable {RepoRHEL7ServerOptional} \
--enable {RepoRHEL7ServerSoftwareCollections}

The Installation guide generated by foreman-el is now called

Installing Foreman server on Debian

And it is interspersed with yum and apt commands :o) 

Same attributes are broken in this guide also: 

   Enable the following repositories:

    # subscription-manager repos --enable={RepoRHEL7Server} \
    --enable {RepoRHEL7ServerOptional} \
    --enable {RepoRHEL7ServerSoftwareCollections}

    Disable all repositories and enable only the following repositories:

    # subscription-manager repos --disable '*'
    # subscription-manager repos --enable={RepoRHEL7ServerSoftwareCollections} \
    --enable={RepoRHEL7Server}

Same here. 

Installing Smart Proxy on EL 

Installing Smart Proxy on EL guide is now titled Installing Smart Proxy on Debian also :+1: 


Foreman Infrastructure is included with all subscriptions that include Smart Management. For more information, see the Red Hat Knowledgebase solution Foreman Infrastructure Subscriptions MCT3718 MCT3719.

Subscriptions are classified as available if they are not already attached to a system. If you are unable to find an available Foreman subscription, see the Red Hat Knowledgebase solution How do I figure out which subscriptions have been consumed by clients registered under Red Hat Subscription Manager? to run a script to see if your subscription is being consumed by another system.
Procedure

These attributes are not rendering: 
    Identify the Pool ID of the Foreman Infrastructure subscription:

    # subscription-manager list --all --available --matches '{SatelliteSub}'

    The command displays output similar to the following:

    Subscription Name:   {SatelliteSub}
    Provides:            Red Hat Satellite
                          Red Hat Software Collections (for RHEL Server)
          
`apt-get install` is throughout this guide also. 

==== Upgrade Guide ====== 

apt-get is throughout

These attributes also don't render:

Updating Disconnected Foreman server to the Next Minor Version

    Create a new configuration file as follows:

    # vi /etc/yum.repos.d/redhat-local.repo

    [{RepoRHEL7ServerAnsible}]
    name=Ansible {ForemanAnsibleVersion} RPMs for Red Hat Enterprise Linux 7 Server x86_64
    baseurl=file:///var/lib/pulp/published/yum/https/repos/Default_Organization/Library/content/dist/rhel/server/7/7Server/x86_64/ansible/2.9/os/
    enabled=1

    [{RepoRHEL7Server}]
    name=Red Hat Enterprise Linux 7 Server RPMs x86_64
    baseurl=file:///var/lib/pulp/published/yum/https/repos/Default_Organization/Library/content/dist/rhel/server/7/7Server/x86_64/os/
    enabled=1

    [{RepoRHEL7ServerForemanServerProductVersion}]
    name=Foreman for RHEL 7 Server RPMs x86_64
    baseurl=file:///var/lib/pulp/published/yum/https/repos/Default_Organization/Library/content/dist/rhel/server/7/7Server/x86_64/satellite/6.9-beta/os/
    enabled=1

    [{RepoRHEL7ServerForemanMaintenanceProductVersion}]
    name=Foreman Maintenance 6 for RHEL 7 Server RPMs x86_64
    baseurl=file:///var/lib/pulp/published/yum/https/repos/Default_Organization/Library/content/dist/rhel/server/7/7Server/x86_64/sat-maintenance/6/os/
    enabled=1

    [{RepoRHEL7ServerSoftwareCollections}]
    name=Red Hat Software Collections RPMs for Red Hat Enterprise Linux 7 Server x86_64
    baseurl=file:///var/lib/pulp/published/yum/https/repos/Default_Organization/Library/content/dist/rhel/server/7/7Server/x86_64/rhscl/1/os/
    enabled=1

Deploying on AWS 

This guide now has apt-get references and references to deploying on Debian. 

Configuring Load Balancing guide also has apt-get replacing yum in places through it

Provisioning guide has apt-get install running through it also. 
I can stop here I think. I'm finding the same issue over and over but the rest looks OK. 
I'll retest in full if we clear up these issues. 

Thank you for doing this!!

lzap:
Updating README, the issue was caused by an overlook, I included `foreman-deb` when `foreman-el` was built.

I compared all builds old and new with GNU diff and all is good except `foreman-deb`, there is still some problem. Looking into it.

lzap:
So I believe I have fixed all issues now. I have created a script which builds all HTML guides with all BUILD flags and stores them in `build-BUILD-BRANCH` directories. Then it does the same for `master` branch and finally it uses GNU diff to compare the branch with master.

This resulted in zero differences for my last commit, so I am pushing:

```
$ LC_ALL=C ls -lah *diff
-rw-r--r--. 1 lzap lzap 0 Apr  1 11:51 foreman-deb-master-vs-master.diff
-rw-r--r--. 1 lzap lzap 0 Apr  1 11:51 foreman-el-master-vs-master.diff
-rw-r--r--. 1 lzap lzap 0 Apr  1 11:51 katello-master-vs-master.diff
-rw-r--r--. 1 lzap lzap 0 Apr  1 11:51 satellite-master-vs-master.diff
```

When trying this make sure the branch is rebased on top of master so there are no differences other than my commit. I am pretty confident saying the patch works as expected.

lzap:
I am pushing shortly, updating the script a bit so it is more useful for other cases.

lzap:
Damn I accidentally compared master vs master, there were no differences, now I see many issues. Looking into this. :-D

lzap:
So I was able to fix hopefully everything, except in Debian version there will be some differences introduced by this patch because previously due to the mess (this PR aims to clean) some variables were missing so Debian got rendered incorrectly. Granted we do not publish those documents at the moment but we can at some point.

Example: https://docs.theforeman.org/nightly/Administering_Red_Hat_Satellite/index-foreman-deb.html#listing-using-satellite-ansible-modules_admin

Variable `{ansiblefilepath}` was previously not defined (in the master branch) but now it is correctly defined.

The same for (already published) https://docs.theforeman.org/nightly/Quickstart_Guide/index-foreman-deb.html - incorrect variables (Enterprise Linux) in the header etc.

The full diff is here:

```diff
diff -r build-foreman-deb-cleanup-attributes/Administering_Red_Hat_Satellite/index-foreman-deb.html build-foreman-deb-master/Administering_Red_Hat_Satellite/index-foreman-deb.html
823c823
< <pre class="nowrap"># ansible-doc -l `theforeman.foreman`</pre>
---
> <pre class="nowrap"># ansible-doc -l {ansible-namespace}</pre>
831c831
< <pre class="nowrap"># ls /usr/share/ansible/collections/ansible_collections/theforeman/foreman/plugins/modules/</pre>
---
> <pre class="nowrap"># ls {ansiblefilepath}</pre>
835c835
< <p>Alternatively, you can also see the complete list of Foreman Ansible modules and other related information at <a href="https://galaxy.ansible.com/theforeman/foreman" class="bare">https://galaxy.ansible.com/theforeman/foreman</a>.</p>
---
> <p>Alternatively, you can also see the complete list of Foreman Ansible modules and other related information at {ansible-galaxy}.</p>
838c838
< <p>All modules are in the `theforeman.foreman` namespace and can be referred to in the format `theforeman.foreman._module_name_`.
---
> <p>All modules are in the {ansible-namespace} namespace and can be referred to in the format {ansible-namespace-example}.
843c843
< <pre class="nowrap">$ ansible-doc theforeman.foreman.activation_key</pre>
---
> <pre class="nowrap">$ {ansible-doc-activation_key}</pre>
diff -r build-foreman-deb-cleanup-attributes/Administering_with_Ansible/index-foreman-deb.html build-foreman-deb-master/Administering_with_Ansible/index-foreman-deb.html
64c64
< <pre class="nowrap"># ansible-doc -l `theforeman.foreman`</pre>
---
> <pre class="nowrap"># ansible-doc -l {ansible-namespace}</pre>
72c72
< <pre class="nowrap"># ls /usr/share/ansible/collections/ansible_collections/theforeman/foreman/plugins/modules/</pre>
---
> <pre class="nowrap"># ls {ansiblefilepath}</pre>
76c76
< <p>Alternatively, you can also see the complete list of Foreman Ansible modules and other related information at <a href="https://galaxy.ansible.com/theforeman/foreman" class="bare">https://galaxy.ansible.com/theforeman/foreman</a>.</p>
---
> <p>Alternatively, you can also see the complete list of Foreman Ansible modules and other related information at {ansible-galaxy}.</p>
79c79
< <p>All modules are in the `theforeman.foreman` namespace and can be referred to in the format `theforeman.foreman._module_name_`.
---
> <p>All modules are in the {ansible-namespace} namespace and can be referred to in the format {ansible-namespace-example}.
84c84
< <pre class="nowrap">$ ansible-doc theforeman.foreman.activation_key</pre>
---
> <pre class="nowrap">$ {ansible-doc-activation_key}</pre>
diff -r build-foreman-deb-cleanup-attributes/Configuring_Ansible/index-foreman-deb.html build-foreman-deb-master/Configuring_Ansible/index-foreman-deb.html
59c59
< <li><a href="#integrating-ansible-tower_ansible">4. Integrating Foreman and AWX</a>
---
> <li><a href="#integrating-ansible-tower_ansible">4. Integrating Foreman and {awx}</a>
61c61
< <li><a href="#adding-server-as-a-dynamic-inventory-item_ansible">4.1. Adding Foreman&#160;server to AWX as a Dynamic Inventory Item</a></li>
---
> <li><a href="#adding-server-as-a-dynamic-inventory-item_ansible">4.1. Adding Foreman&#160;server to {awx} as a Dynamic Inventory Item</a></li>
1426c1426
< <h2 id="integrating-ansible-tower_ansible">4. Integrating Foreman and AWX</h2>
---
> <h2 id="integrating-ansible-tower_ansible">4. Integrating Foreman and {awx}</h2>
1429c1429
< <p>You can integrate Foreman and AWX to use Foreman&#160;server as a dynamic inventory source for AWX.</p>
---
> <p>You can integrate Foreman and {awx} to use Foreman&#160;server as a dynamic inventory source for {awx}.</p>
1432,1433c1432,1433
< <p>You can also use the provisioning callback function to run playbooks on hosts managed by Foreman, from either the host or AWX.
< When provisioning new hosts from Foreman&#160;server, you can use the provisioning callback function to trigger playbook runs from AWX.
---
> <p>You can also use the provisioning callback function to run playbooks on hosts managed by Foreman, from either the host or {awx}.
> When provisioning new hosts from Foreman&#160;server, you can use the provisioning callback function to trigger playbook runs from {awx}.
1437c1437
< <h3 id="adding-server-as-a-dynamic-inventory-item_ansible">4.1. Adding Foreman&#160;server to AWX as a Dynamic Inventory Item</h3>
---
> <h3 id="adding-server-as-a-dynamic-inventory-item_ansible">4.1. Adding Foreman&#160;server to {awx} as a Dynamic Inventory Item</h3>
1439c1439
< <p>To add Foreman&#160;server to AWX as a dynamic inventory item, you must create a credential for a Foreman&#160;server user on AWX, add an AWX user to the credential, and then configure an inventory source.</p>
---
> <p>To add Foreman&#160;server to {awx} as a dynamic inventory item, you must create a credential for a Foreman&#160;server user on {awx}, add an {awx} user to the credential, and then configure an inventory source.</p>
1449c1449
< <p>For non-admin users, you must assign the <code>AWX Inventory Reader</code> role to your Foreman&#160;server user.
---
> <p>For non-admin users, you must assign the <code>{awx} Inventory Reader</code> role to your Foreman&#160;server user.
1453c1453
< <p>You must host your Foreman&#160;server and AWX on the same network or subnet.</p>
---
> <p>You must host your Foreman&#160;server and {awx} on the same network or subnet.</p>
1461,1462c1461,1462
< <p>In the AWX web UI, create a credential for your Foreman.
< For more information about creating credentials, see <a href="http://docs.ansible.com/ansible-tower/latest/html/userguide/credentials.html#add-a-new-credential">Add a New Credential</a> and <a href="http://docs.ansible.com/ansible-tower/latest/html/userguide/credentials.html#red-hat-satellite-6">Foreman Credentials</a> in the <em>AWX User Guide</em>.</p>
---
> <p>In the {awx} web UI, create a credential for your Foreman.
> For more information about creating credentials, see <a href="http://docs.ansible.com/ansible-tower/latest/html/userguide/credentials.html#add-a-new-credential">Add a New Credential</a> and <a href="http://docs.ansible.com/ansible-tower/latest/html/userguide/credentials.html#red-hat-satellite-6">Foreman Credentials</a> in the <em>{awx} User Guide</em>.</p>
1504,1505c1504,1505
< <p>Add an AWX user to the new credential.
< For more information about adding a user to a credential, see <a href="http://docs.ansible.com/ansible-tower/latest/html/userguide/credentials.html#getting-started-with-credentials">Getting Started with Credentials</a> in the <em>AWX User Guide</em>.</p>
---
> <p>Add an {awx} user to the new credential.
> For more information about adding a user to a credential, see <a href="http://docs.ansible.com/ansible-tower/latest/html/userguide/credentials.html#getting-started-with-credentials">Getting Started with Credentials</a> in the <em>{awx} User Guide</em>.</p>
1509c1509
< For more information, see <a href="http://docs.ansible.com/ansible-tower/latest/html/userguide/inventories.html#add-a-new-inventory">Add a new inventory</a> in the <em>AWX User Guide</em>.</p>
---
> For more information, see <a href="http://docs.ansible.com/ansible-tower/latest/html/userguide/inventories.html#add-a-new-inventory">Add a new inventory</a> in the <em>{awx} User Guide</em>.</p>
1513c1513
< For more information, see <a href="https://docs.ansible.com/ansible-tower/latest/html/userguide/inventories.html#add-source">Add Source</a> in the <em>AWX User Guide</em>.</p>
---
> For more information, see <a href="https://docs.ansible.com/ansible-tower/latest/html/userguide/inventories.html#add-source">Add Source</a> in the <em>{awx} User Guide</em>.</p>
1579,1580c1579,1580
< <p>When you create hosts in Foreman, you can use AWX to run playbooks to configure your newly created hosts.
< This is called <em>provisioning callback</em> in AWX.</p>
---
> <p>When you create hosts in Foreman, you can use {awx} to run playbooks to configure your newly created hosts.
> This is called <em>provisioning callback</em> in {awx}.</p>
1583c1583
< <p>The provisioning callback function triggers a playbook run from AWX as part of the provisioning process.
---
> <p>The provisioning callback function triggers a playbook run from {awx} as part of the provisioning process.
1587c1587
< <p>For more information about provisioning callbacks, see <a href="http://docs.ansible.com/ansible-tower/latest/html/userguide/job_templates.html#provisioning-callbacks">Provisioning Callbacks</a> in the <em>AWX User Guide</em>.</p>
---
> <p>For more information about provisioning callbacks, see <a href="http://docs.ansible.com/ansible-tower/latest/html/userguide/job_templates.html#provisioning-callbacks">Provisioning Callbacks</a> in the <em>{awx} User Guide</em>.</p>
1610,1611c1610,1611
< <p>Before you can configure provisioning callbacks, you must add Foreman as a dynamic inventory in AWX.
< For more information, see <a href="https://docs.theforeman.org/nightly/Configuring_Ansible/index-foreman-deb.html#integrating-ansible-tower_ansible">Integrating Foreman and AWX</a>.</p>
---
> <p>Before you can configure provisioning callbacks, you must add Foreman as a dynamic inventory in {awx}.
> For more information, see <a href="https://docs.theforeman.org/nightly/Configuring_Ansible/index-foreman-deb.html#integrating-ansible-tower_ansible">Integrating Foreman and {awx}</a>.</p>
1614c1614
< <p>In the AWX web UI, you must complete the following tasks:</p>
---
> <p>In the {awx} web UI, you must complete the following tasks:</p>
1621c1621
< For more information, see <a href="https://docs.ansible.com/ansible-tower/latest/html/userguide/credentials.html#add-a-new-credential">Add a New Credential</a> in the <em>AWX User Guide</em>.</p>
---
> For more information, see <a href="https://docs.ansible.com/ansible-tower/latest/html/userguide/credentials.html#add-a-new-credential">Add a New Credential</a> in the <em>{awx} User Guide</em>.</p>
1625c1625
< For more information, see <a href="https://docs.ansible.com/ansible-tower/latest/html/userguide/projects.html">Projects</a> in the <em>AWX User Guide</em>.</p>
---
> For more information, see <a href="https://docs.ansible.com/ansible-tower/latest/html/userguide/projects.html">Projects</a> in the <em>{awx} User Guide</em>.</p>
1629c1629
< For more information, see <a href="https://docs.ansible.com/ansible-tower/latest/html/userguide/job_templates.html#create-a-job-template">Job Templates</a> in the <em>AWX User Guide</em>.</p>
---
> For more information, see <a href="https://docs.ansible.com/ansible-tower/latest/html/userguide/job_templates.html#create-a-job-template">Job Templates</a> in the <em>{awx} User Guide</em>.</p>
1633c1633
< For more information about job templates, see <a href="http://docs.ansible.com/ansible-tower/latest/html/userguide/job_templates.html#">Job Templates</a> in the <em>AWX User Guide</em>.</p>
---
> For more information about job templates, see <a href="http://docs.ansible.com/ansible-tower/latest/html/userguide/job_templates.html#">Job Templates</a> in the <em>{awx} User Guide</em>.</p>
1688,1689c1688,1689
< <p>The fully qualified domain name (FQDN) of your AWX.
< Do not add <code>https</code> because this is appended by AWX.</p>
---
> <p>The fully qualified domain name (FQDN) of your {awx}.
> Do not add <code>https</code> because this is appended by {awx}.</p>
1711c1711
< <p>The host configuration key that your job template generates in AWX.</p>
---
> <p>The host configuration key that your job template generates in {awx}.</p>
1743c1743
< <pre class="nowrap">SAT_host systemd[1]: Started Provisioning callback to AWX...</pre>
---
> <pre class="nowrap">SAT_host systemd[1]: Started Provisioning callback to {awx}...</pre>
1751c1751
< <p>You can use the provisioning callback URL and the host configuration key from a host to call AWX.
---
> <p>You can use the provisioning callback URL and the host configuration key from a host to call {awx}.
diff -r build-foreman-deb-cleanup-attributes/Configuring_Load_Balancer/index-foreman-deb.html build-foreman-deb-master/Configuring_Load_Balancer/index-foreman-deb.html
634c634
< If you set a wildcard value, you must add the <code>-t foreman-proxy</code> option when you use the <code>katello-certs-check</code> command.</p>
---
> If you set a wildcard value, you must add the <code>-t {certs-proxy-context}</code> option when you use the <code>katello-certs-check</code> command.</p>
701c701
< <p>If you set the <code>commonName=</code> to a wildcard value <code>*</code>, you must add the <code>-t foreman-proxy</code> option to the <code>katello-certs-check</code> command.</p>
---
> <p>If you set the <code>commonName=</code> to a wildcard value <code>*</code>, you must add the <code>-t {certs-proxy-context}</code> option to the <code>katello-certs-check</code> command.</p>
870c870
< If you set a wildcard value, you must add the <code>-t foreman-proxy</code> option when you use the <code>katello-certs-check</code> command.</p>
---
> If you set a wildcard value, you must add the <code>-t {certs-proxy-context}</code> option when you use the <code>katello-certs-check</code> command.</p>
934c934
< <p>If you set the <code>commonName=</code> to a wildcard value <code>*</code>, you must add the <code>-t foreman-proxy</code> option to the <code>katello-certs-check</code> command.</p>
---
> <p>If you set the <code>commonName=</code> to a wildcard value <code>*</code>, you must add the <code>-t {certs-proxy-context}</code> option to the <code>katello-certs-check</code> command.</p>
diff -r build-foreman-deb-cleanup-attributes/Content_Management_Guide/index-foreman-deb.html build-foreman-deb-master/Content_Management_Guide/index-foreman-deb.html
53,55c53,55
< <li><a href="#Importing_Content-Importing_Custom_SSL_Certificates">5.2. Importing SSL Certificates</a></li>
< <li><a href="#Importing_Content-Creating_a_Custom_Product">5.3. Creating a Product</a></li>
< <li><a href="#Importing_Content-Creating_a_Custom_RPM_Repository">5.4. Adding RPM Repositories</a></li>
---
> <li><a href="#Importing_Content-Importing_Custom_SSL_Certificates">5.2. Importing {customssltitle} Certificates</a></li>
> <li><a href="#Importing_Content-Creating_a_Custom_Product">5.3. Creating a {customproducttitle}</a></li>
> <li><a href="#Importing_Content-Creating_a_Custom_RPM_Repository">5.4. Adding {customrpmtitle} Repositories</a></li>
62c62
< <li><a href="#uploading-content-to-a-custom-rpm-repository">5.11. Uploading Content to RPM Repositories</a></li>
---
> <li><a href="#uploading-content-to-a-custom-rpm-repository">5.11. Uploading Content to {customrpmtitle} Repositories</a></li>
70c70
< <li><a href="#Importing_Content-Importing_a_Custom_GPG_Key">5.19. Importing a GPG Key</a></li>
---
> <li><a href="#Importing_Content-Importing_a_Custom_GPG_Key">5.19. Importing a {customgpgtitle} Key</a></li>
128c128
< <li><a href="#_importing_ostree_content">11.2. Importing OSTree Content</a></li>
---
> <li><a href="#_importing_customostreecontenttitle">11.2. Importing {customostreecontenttitle}</a></li>
146c146
< <li><a href="#Managing_Custom_File_Type_Content">14. Managing File Type Content</a>
---
> <li><a href="#Managing_Custom_File_Type_Content">14. Managing {customfiletypetitle} Content</a>
148,149c148,149
< <li><a href="#Importing_Content-Creating_a_Custom_File_Type_Repository">14.1. Creating a File Type Repository in Foreman</a></li>
< <li><a href="#Importing_Content-Creating_a_Custom_File_Type_Repository_Local_Directory">14.2. Creating a File Type Repository in a Local Directory</a></li>
---
> <li><a href="#Importing_Content-Creating_a_Custom_File_Type_Repository">14.1. Creating a {customfiletypetitle} Repository in Foreman</a></li>
> <li><a href="#Importing_Content-Creating_a_Custom_File_Type_Repository_Local_Directory">14.2. Creating a {customfiletypetitle} Repository in a Local Directory</a></li>
151,152c151,152
< <li><a href="#Importing_Content-Uploading_Files_To_a_Custom_File_Type_Repository">14.4. Uploading Files To a File Type Repository in Foreman</a></li>
< <li><a href="#Importing_Content-Downloading_Files_From_a_Custom_File_Type_Repository">14.5. Downloading Files to a Host From a File Type Repository in Foreman</a></li>
---
> <li><a href="#Importing_Content-Uploading_Files_To_a_Custom_File_Type_Repository">14.4. Uploading Files To a {customfiletypetitle} Repository in Foreman</a></li>
> <li><a href="#Importing_Content-Downloading_Files_From_a_Custom_File_Type_Repository">14.5. Downloading Files to a Host From a {customfiletypetitle} Repository in Foreman</a></li>
206c206
< <p>You can also manage other types of content in Foreman.
---
> <p>You can also manage other types of {customcontent} in Foreman.
228c228
< <p>You can use the procedure to add content for any type of content you require, for example, SSL certificates and OVAL files.</p>
---
> <p>You can use the procedure to add {customcontent} for any type of content you require, for example, SSL certificates and OVAL files.</p>
1074c1074
< <p>This chapter outlines how you can import different types of content to Foreman.
---
> <p>This chapter outlines how you can import different types of {customcontent} to Foreman.
1078c1078
< <p>For example, you can use the following chapters for information on specific types of content but the underlying procedures are the same:</p>
---
> <p>For example, you can use the following chapters for information on specific types of {customcontent} but the underlying procedures are the same:</p>
1089c1089
< <p><a href="#Managing_Custom_File_Type_Content">Managing File Type Content</a></p>
---
> <p><a href="#Managing_Custom_File_Type_Content">Managing {customfiletypetitle} Content</a></p>
1096c1096
< <p>Both Red Hat content and non-RH content in Foreman have similarities:</p>
---
> <p>Both Red Hat content and non-RH {customcontent} in Foreman have similarities:</p>
1104,1105c1104,1105
< <p>products require a subscription for clients to access, similar to subscriptions to Red Hat Products.
< Foreman creates a subscription for each product you create.</p>
---
> <p>{customproduct}s require a subscription for clients to access, similar to subscriptions to Red Hat Products.
> Foreman creates a subscription for each {customproduct} you create.</p>
1114c1114
< <p>Other content can be organized into Products however you want.  For example, you might create an EPEL (Extra Packages for Enterprise Linux) product and add an "EPEL 7 x86_64" repository to it.</p>
---
> <p>Other {customcontent} can be organized into Products however you want.  For example, you might create an EPEL (Extra Packages for Enterprise Linux) product and add an "EPEL 7 x86_64" repository to it.</p>
1121c1121
< <h3 id="Importing_Content-Importing_Custom_SSL_Certificates">5.2. Importing SSL Certificates</h3>
---
> <h3 id="Importing_Content-Importing_Custom_SSL_Certificates">5.2. Importing {customssltitle} Certificates</h3>
1123c1123
< <p>Before you synchronize content from an external source, you might need to import SSL certificates into your product. This might include client certs and keys or CA certificates for the upstream repositories you want to synchronize.</p>
---
> <p>Before you synchronize {customcontent} from an external source, you might need to import SSL certificates into your {customproduct}. This might include client certs and keys or CA certificates for the upstream repositories you want to synchronize.</p>
1150c1150
< <h3 id="Importing_Content-Creating_a_Custom_Product">5.3. Creating a Product</h3>
---
> <h3 id="Importing_Content-Creating_a_Custom_Product">5.3. Creating a {customproducttitle}</h3>
1152c1152
< <p>Use this procedure to create a product that you can then add repositories to.
---
> <p>Use this procedure to create a {customproduct} that you can then add repositories to.
1203c1203
< <h3 id="Importing_Content-Creating_a_Custom_RPM_Repository">5.4. Adding RPM Repositories</h3>
---
> <h3 id="Importing_Content-Creating_a_Custom_RPM_Repository">5.4. Adding {customrpmtitle} Repositories</h3>
1205c1205
< <p>Use this procedure to add RPM repositories in Foreman.
---
> <p>Use this procedure to add {customrpm} repositories in Foreman.
1209c1209
< <p>The Products window in the Foreman web UI also provides a <strong>Repo Discovery</strong> function that finds all repositories from a URL and you can select which ones to add to your product.
---
> <p>The Products window in the Foreman web UI also provides a <strong>Repo Discovery</strong> function that finds all repositories from a URL and you can select which ones to add to your {customproduct}.
1534c1534
< <p>Depending on whether it is a Red Hat or non-Red Hat repository, Foreman uses separate settings.
---
> <p>Depending on whether it is a Red Hat or non-Red Hat {customrepo}, Foreman uses separate settings.
1557c1557
< <p>To change the default download policy for a non-Red Hat repository, change the value of the <strong>Default Custom Repository download policy</strong> setting.</p>
---
> <p>To change the default download policy for a non-Red Hat {customrepo}, change the value of the <strong>Default Custom Repository download policy</strong> setting.</p>
1578c1578
< <p>To change the default download policy for a repository to one of <code>immediate</code> or <code>on_demand</code>, enter the following command:</p>
---
> <p>To change the default download policy for a {customrepo} to one of <code>immediate</code> or <code>on_demand</code>, enter the following command:</p>
1637c1637
< <h3 id="uploading-content-to-a-custom-rpm-repository">5.11. Uploading Content to RPM Repositories</h3>
---
> <h3 id="uploading-content-to-a-custom-rpm-repository">5.11. Uploading Content to {customrpmtitle} Repositories</h3>
1639c1639
< <p>You can upload individual RPMs and source RPMs to RPM repositories.
---
> <p>You can upload individual RPMs and source RPMs to {customrpm} repositories.
1650c1650
< <p>Click the name of the product.</p>
---
> <p>Click the name of the {customproduct}.</p>
1653c1653
< <p>In the <strong>Repositories</strong> tab, click the name of the RPM repository.</p>
---
> <p>In the <strong>Repositories</strong> tab, click the name of the {customrpm} repository.</p>
2100c2100
< <h3 id="Importing_Content-Importing_a_Custom_GPG_Key">5.19. Importing a GPG Key</h3>
---
> <h3 id="Importing_Content-Importing_a_Custom_GPG_Key">5.19. Importing a {customgpgtitle} Key</h3>
2102c2102
< <p>When clients are consuming signed content, ensure that the clients are configured to validate the installation of RPMs with the appropriate GPG Key.  This helps to ensure that only packages from authorized sources can be installed.</p>
---
> <p>When clients are consuming signed {customcontent}, ensure that the clients are configured to validate the installation of RPMs with the appropriate GPG Key.  This helps to ensure that only packages from authorized sources can be installed.</p>
2149c2149
< <p>If your repository contains content signed by multiple GPG keys, you must enter all required GPG keys in the <strong>Content Credential Contents</strong> field with new lines between each key, for example:</p>
---
> <p>If your {customrepo} contains content signed by multiple GPG keys, you must enter all required GPG keys in the <strong>Content Credential Contents</strong> field with new lines between each key, for example:</p>
2992c2992
< <p>Before you begin, upload the required Puppet module to a repository within a product.
---
> <p>Before you begin, upload the required Puppet module to a repository within a {customproduct}.
4141,4142c4141,4142
< <div class="title">Products</div>
< <p>If a product, typically containing content not provided by Red Hat, is assigned to an activation key, this product is always enabled for the registered content host regardless of the auto-attach setting.</p>
---
> <div class="title">{customproducttitle}s</div>
> <p>If a {customproduct}, typically containing content not provided by Red Hat, is assigned to an activation key, this product is always enabled for the registered content host regardless of the auto-attach setting.</p>
5324c5324
< You can use OSTree content on a build system, then export an OSTree
---
> You can use {customostreecontent} on a build system, then export an OSTree
5455c5455
< <h3 id="_importing_ostree_content">11.2. Importing OSTree Content</h3>
---
> <h3 id="_importing_customostreecontenttitle">11.2. Importing {customostreecontenttitle}</h3>
5533c5533
< <p>Create a product for your OSTree content:</p>
---
> <p>Create a product for your {customostreecontent}:</p>
5537c5537
< --name "<em>OSTree Content</em>" \
---
> --name "<em>{customostreecontenttitle}</em>" \
6059c6059
< <p>Create the product:</p>
---
> <p>Create the {customproduct}:</p>
6099c6099
< <h2 id="Managing_Custom_File_Type_Content">14. Managing File Type Content</h2>
---
> <h2 id="Managing_Custom_File_Type_Content">14. Managing {customfiletypetitle} Content</h2>
6103c6103
< To achieve this, (customproduct)s in Foreman include repositories for file types.
---
> To achieve this, (customproduct)s in Foreman include repositories for {customfiletype}s.
6108c6108
< When you add files to a file type repository, you can use the normal Foreman management functions such as adding a specific version to a Content View to provide version control and making the repository of files available on various Smart&#160;Proxy&#160;servers.
---
> When you add files to a {customfiletype} repository, you can use the normal Foreman management functions such as adding a specific version to a Content View to provide version control and making the repository of files available on various Smart&#160;Proxy&#160;servers.
6117c6117
< <h3 id="Importing_Content-Creating_a_Custom_File_Type_Repository">14.1. Creating a File Type Repository in Foreman</h3>
---
> <h3 id="Importing_Content-Creating_a_Custom_File_Type_Repository">14.1. Creating a {customfiletypetitle} Repository in Foreman</h3>
6119,6120c6119,6120
< <p>The procedure for creating a file type repository is the same as the procedure for creating any content, except that when you create the repository, you select the <strong>file</strong> type.
< You must create a product and then add a repository.</p>
---
> <p>The procedure for creating a {customfiletype} repository is the same as the procedure for creating any {customcontent}, except that when you create the repository, you select the <strong>file</strong> type.
> You must create a product and then add a {customrepo}.</p>
6127c6127
< <p>To create a product, complete the following procedure:</p>
---
> <p>To create a {customproduct}, complete the following procedure:</p>
6150c6150
< <p>To create a repository for your product, complete the following procedure:</p>
---
> <p>To create a repository for your {customproduct}, complete the following procedure:</p>
6190c6190
< <p>Create a product</p>
---
> <p>Create a {customproduct}</p>
6303c6303
< <h3 id="Importing_Content-Creating_a_Custom_File_Type_Repository_Local_Directory">14.2. Creating a File Type Repository in a Local Directory</h3>
---
> <h3 id="Importing_Content-Creating_a_Custom_File_Type_Repository_Local_Directory">14.2. Creating a {customfiletypetitle} Repository in a Local Directory</h3>
6305c6305
< <p>You can create a file type repository, from a directory of files, on the base system where Foreman is installed using the <code>pulp-manifest</code> command.
---
> <p>You can create a {customfiletype} repository, from a directory of files, on the base system where Foreman is installed using the <code>pulp-manifest</code> command.
6372c6372
< <p>Ensure a product exists in Foreman&#160;server.</p>
---
> <p>Ensure a {customproduct} exists in Foreman&#160;server.</p>
6434c6434
< <p>You can create a file type repository from a directory of files that is external to Foreman&#160;server using the <code>pulp-manifest</code> command.
---
> <p>You can create a {customfiletype} repository from a directory of files that is external to Foreman&#160;server using the <code>pulp-manifest</code> command.
6440c6440
< To create a file type repository in a directory on the base system where Foreman&#160;server is installed, see <a href="#Importing_Content-Creating_a_Custom_File_Type_Repository_Local_Directory">Creating a File Type Repository in a Local Directory</a>.</p>
---
> To create a file type repository in a directory on the base system where Foreman&#160;server is installed, see <a href="#Importing_Content-Creating_a_Custom_File_Type_Repository_Local_Directory">Creating a {customfiletypetitle} Repository in a Local Directory</a>.</p>
6516,6517c6516,6517
< <p>Ensure a product exists in Foreman&#160;server, or create a product.
< For more information see <a href="#Importing_Content-Creating_a_Custom_File_Type_Repository">Creating a File Type Repository in Foreman</a></p>
---
> <p>Ensure a {customproduct} exists in Foreman&#160;server, or create a {customproduct}.
> For more information see <a href="#Importing_Content-Creating_a_Custom_File_Type_Repository">Creating a {customfiletypetitle} Repository in Foreman</a></p>
6569c6569
< <h3 id="Importing_Content-Uploading_Files_To_a_Custom_File_Type_Repository">14.4. Uploading Files To a File Type Repository in Foreman</h3>
---
> <h3 id="Importing_Content-Uploading_Files_To_a_Custom_File_Type_Repository">14.4. Uploading Files To a {customfiletypetitle} Repository in Foreman</h3>
6571c6571
< <p>Use this procedure to upload files to a file type repository.</p>
---
> <p>Use this procedure to upload files to a {customfiletype} repository.</p>
6580c6580
< <p>Select a product by name.</p>
---
> <p>Select a {customproduct} by name.</p>
6611c6611
< <h3 id="Importing_Content-Downloading_Files_From_a_Custom_File_Type_Repository">14.5. Downloading Files to a Host From a File Type Repository in Foreman</h3>
---
> <h3 id="Importing_Content-Downloading_Files_From_a_Custom_File_Type_Repository">14.5. Downloading Files to a Host From a {customfiletypetitle} Repository in Foreman</h3>
6619,6620c6619,6620
< <p>You have a file type repository.
< See <a href="#Importing_Content-Creating_a_Custom_File_Type_Repository">Creating a File Type Repository in Foreman</a> for more information.</p>
---
> <p>You have a {customfiletype} repository.
> See <a href="#Importing_Content-Creating_a_Custom_File_Type_Repository">Creating a {customfiletypetitle} Repository in Foreman</a> for more information.</p>
6649c6649
< <p>Select a product by name.</p>
---
> <p>Select a {customproduct} by name.</p>
diff -r build-foreman-deb-cleanup-attributes/Deploying_on_AWS/index-foreman-deb.html build-foreman-deb-master/Deploying_on_AWS/index-foreman-deb.html
234c234
< <p>To install Foreman in an AWS environment, you must ensure that your AWS set up meets the <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#system-requirements_foreman">System Requirements</a> in <em>Installing Foreman server on Debian</em>.</p>
---
> <p>To install Foreman in an AWS environment, you must ensure that your AWS set up meets the <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#system-requirements_foreman">System Requirements</a> in <em>Installing Foreman 2.4 server on Enterprise Linux</em>.</p>
251c251
< <p>Ensure that the Amazon EC2 instance type meets or exceeds the <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#system-requirements_foreman">System Requirements</a> in <em>Installing Foreman server on Debian</em>.
---
> <p>Ensure that the Amazon EC2 instance type meets or exceeds the <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#system-requirements_foreman">System Requirements</a> in <em>Installing Foreman 2.4 server on Enterprise Linux</em>.
274c274
< <p>Use <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#index">Installing Foreman server on Debian</a> to install Foreman&#160;server.</p>
---
> <p>Use <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#index">Installing Foreman 2.4 server on Enterprise Linux</a> to install Foreman&#160;server.</p>
diff -r build-foreman-deb-cleanup-attributes/Installing_Proxy_on_Debian/index-foreman-deb.html build-foreman-deb-master/Installing_Proxy_on_Debian/index-foreman-deb.html
368c368
< For more information, see <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#enabling-client-connections-to-satellite_foreman">Enabling Connections from a Client to Foreman&#160;server</a> in <em>Installing Foreman server on Debian</em>.</p>
---
> For more information, see <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#enabling-client-connections-to-satellite_foreman">Enabling Connections from a Client to Foreman&#160;server</a> in <em>Installing Foreman 2.4 server on Enterprise Linux</em>.</p>
diff -r build-foreman-deb-cleanup-attributes/Installing_Proxy_on_Red_Hat/index-foreman-deb.html build-foreman-deb-master/Installing_Proxy_on_Red_Hat/index-foreman-deb.html
376c376
< For more information, see <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#enabling-client-connections-to-satellite_foreman">Enabling Connections from a Client to Foreman&#160;server</a> in <em>Installing Foreman server on Debian</em>.</p>
---
> For more information, see <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#enabling-client-connections-to-satellite_foreman">Enabling Connections from a Client to Foreman&#160;server</a> in <em>Installing Foreman 2.4 server on Enterprise Linux</em>.</p>
diff -r build-foreman-deb-cleanup-attributes/Installing_Server_on_Debian/index-foreman-deb.html build-foreman-deb-master/Installing_Server_on_Debian/index-foreman-deb.html
8c8
< <title>Installing Foreman server on Debian</title>
---
> <title>Installing Foreman 2.4 server on Enterprise Linux</title>
13c13
< <h1>Installing Foreman server on Debian</h1>
---
> <h1>Installing Foreman 2.4 server on Enterprise Linux</h1>
diff -r build-foreman-deb-cleanup-attributes/Installing_Server_on_Red_Hat/index-foreman-deb.html build-foreman-deb-master/Installing_Server_on_Red_Hat/index-foreman-deb.html
8c8
< <title>Installing Foreman server on Debian</title>
---
> <title>Installing Foreman 2.4 server on Enterprise Linux</title>
13c13
< <h1>Installing Foreman server on Debian</h1>
---
> <h1>Installing Foreman 2.4 server on Enterprise Linux</h1>
diff -r build-foreman-deb-cleanup-attributes/Managing_Hosts/index-foreman-deb.html build-foreman-deb-master/Managing_Hosts/index-foreman-deb.html
1096c1096
< For more information about BMC, see <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#enabling-power-management-on-managed-hosts_foreman">Enabling Power Management on Managed Hosts</a> in <em>Installing Foreman server on Debian</em>.
---
> For more information about BMC, see <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#enabling-power-management-on-managed-hosts_foreman">Enabling Power Management on Managed Hosts</a> in <em>Installing Foreman 2.4 server on Enterprise Linux</em>.
diff -r build-foreman-deb-cleanup-attributes/Planning_Guide/index-foreman-deb.html build-foreman-deb-master/Planning_Guide/index-foreman-deb.html
501c501
< <p>The <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#satellite-ports-and-firewalls-requirements_foreman">Ports and Firewalls Requirements</a> in <em>Installing Foreman server on Debian</em> and <a href="https://docs.theforeman.org/nightly/Installing_Proxy_on_Red_Hat/index-foreman-deb.html#capsule-ports-and-firewalls-requirements_smart-proxy">Ports and Firewalls Requirements</a> in <em>Installing an External Smart Proxy Server on Debian</em> contain complete instructions for configuring the host-based firewall to open the ports required.</p>
---
> <p>The <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#satellite-ports-and-firewalls-requirements_foreman">Ports and Firewalls Requirements</a> in <em>Installing Foreman 2.4 server on Enterprise Linux</em> and <a href="https://docs.theforeman.org/nightly/Installing_Proxy_on_Red_Hat/index-foreman-deb.html#capsule-ports-and-firewalls-requirements_smart-proxy">Ports and Firewalls Requirements</a> in <em>Installing an External Smart Proxy Server on Debian</em> contain complete instructions for configuring the host-based firewall to open the ports required.</p>
908c908
< <p>For more information about using an external database, see <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#using-external-databases_foreman">Using External Databases with Foreman</a> in <em>Installing Foreman server on Debian</em>.</p>
---
> <p>For more information about using an external database, see <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#using-external-databases_foreman">Using External Databases with Foreman</a> in <em>Installing Foreman 2.4 server on Enterprise Linux</em>.</p>
diff -r build-foreman-deb-cleanup-attributes/Provisioning_Guide/index-foreman-deb.html build-foreman-deb-master/Provisioning_Guide/index-foreman-deb.html
1302c1302
< To configure Foreman to use external DHCP, DNS, and TFTP services, see <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#configuring-external-services">Configuring External Services</a> in <em>Installing Foreman server on Debian</em>.</p>
---
> To configure Foreman to use external DHCP, DNS, and TFTP services, see <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#configuring-external-services">Configuring External Services</a> in <em>Installing Foreman 2.4 server on Enterprise Linux</em>.</p>
6963c6963
< <p>For more information about synchronizing time in Foreman, see <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#synchronizing-the-system-clock-with-chronyd_foreman">Synchronizing Time</a> in <em>Installing Foreman server on Debian</em>.</p>
---
> <p>For more information about synchronizing time in Foreman, see <a href="https://docs.theforeman.org/nightly/Installing_Server_on_Red_Hat/index-foreman-deb.html#synchronizing-the-system-clock-with-chronyd_foreman">Synchronizing Time</a> in <em>Installing Foreman 2.4 server on Enterprise Linux</em>.</p>
diff -r build-foreman-deb-cleanup-attributes/Quickstart_Guide/index-foreman-deb.html build-foreman-deb-master/Quickstart_Guide/index-foreman-deb.html
8c8
< <title>Installing Foreman server on Debian</title>
---
> <title>Installing Foreman 2.4 server on Enterprise Linux</title>
13c13
< <h1>Installing Foreman server on Debian</h1>
---
> <h1>Installing Foreman 2.4 server on Enterprise Linux</h1>
```

Everything else is okay now:

```
[lzap@nuc guides]$ LC_ALL=C ls -lah *diff
-rw-r--r--. 1 lzap lzap 39K Apr  1 12:25 foreman-deb-cleanup-attributes-vs-master.diff
-rw-r--r--. 1 lzap lzap   0 Apr  1 12:25 foreman-el-cleanup-attributes-vs-master.diff
-rw-r--r--. 1 lzap lzap   0 Apr  1 12:25 katello-cleanup-attributes-vs-master.diff
-rw-r--r--. 1 lzap lzap   0 Apr  1 12:25 satellite-cleanup-attributes-vs-master.diff
```

melcorr (approved):
LGTM

melcorr:
I'm unsure whether we wait for other PRs to merge. Wdyt?

lzap:
I will be commenting on opened PRs explaining what just happened, thanks.

PR #462, opened 2021-03-24
Fixes #32169 - use yum.theforeman.org for Katello

evgeni:
Cherry-pick into:

### DIFF
--- a/guides/common/modules/proc_configuring-repositories-proxy.adoc
+++ b/guides/common/modules/proc_configuring-repositories-proxy.adoc
@@ -72,7 +72,7 @@ ifdef::foreman-el,katello[]
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# yum localinstall https://fedorapeople.org/groups/katello/releases/yum/{KatelloVersion}/katello/el7/x86_64/katello-repos-latest.rpm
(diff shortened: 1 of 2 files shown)


ehelms (approved):

guides/common/modules/proc_configuring-repositories-proxy.adoc
```diff
@@ -72,7 +72,7 @@ ifdef::foreman-el,katello[]
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# yum localinstall https://fedorapeople.org/groups/katello/releases/yum/{KatelloVersion}/katello/el7/x86_64/katello-repos-latest.rpm
+# yum localinstall https://yum.theforeman.org/katello/{KatelloVersion}/katello/el7/x86_64/katello-repos-latest.rpm
```
ekohl, line 75: Looking at the other file, this entire file should use `el{distribution-major-version}` instead of `el7`.

evgeni: probably correct, but it also talks about epel and scls, so needs more than just that for el8 awareness

ekohl: Yes, it needs a lot of love.

PR #465, opened 2021-03-31
Remove "the" before "{SmartProxyServer}"

maximiliankolb:
This PR removes all instances of `the` before `{SmartProxyServer}`:

`git grep -i -n "the {SmartProxyServer}"`

This came up during the review of the application centric deployment PR.

I think `the {ProjectServer}` occurs just as often for which I would create a separate PR if desired.


melcorr (approved):
Thank you @maximiliankolb

PR #469, opened 2021-04-01
Fail on asciidoc error and fix one context

lzap:
This patch removes unmatched endif statement which caused asciidoc to print an ERROR.

Because asciidoc still returns zero code, the build continued. The makefile now checks for ERROR strings in asciidoc output and fail if one is found.

This patch also removes one warning - reused id. This is to serve as an example how to fix those. Once we get rid of all warnings we can start failing when warnings are found so everything can be even more strict.

### DIFF
--- a/guides/common/modules/proc_installing-the-satellite-server-packages.adoc
+++ b/guides/common/modules/proc_installing-the-satellite-server-packages.adoc
@@ -1,3 +1,5 @@
+ifdef::context[:parent-context: {context}]
+
 [id="installing-server-packages_{context}"]
 [id="installing-the-satellite-server-packages_{context}"]
 = Installing the {ProjectServer} Packages
@@ -19,6 +21,7 @@ endif::[]
 ifdef::foreman-el,katello[]
 == [[centos-7]]CentOS 7
 
+:context: centos7
 :package-manager: yum
 include::proc_installing-server-packages-el.adoc[]
 
@@ -27,6 +30,7 @@ endif::[]
 ifdef::foreman-el,katello[]
 == [[centos-8]]CentOS 8
 
+:context: centos8
 :package-manager: dnf
 include::proc_installing-server-packages-el.adoc[]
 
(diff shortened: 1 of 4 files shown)


guides/common/modules/proc_installing-the-satellite-server-packages.adoc
```diff
@@ -43,12 +48,17 @@ endif::[]
...
 
 ifdef::foreman-deb[]
 == [[ubuntu-1804]]Ubuntu 18.04 (Bionic)
 
+:context: ubuntu-bionic
 include::proc_installing-server-packages-deb.adoc[]
 endif::[]
+
+ifdef::parent-context[:context: {parent-context}]
+ifndef::parent-context[:!context
```
melcorr, line 64: should this be closed?

lzap: Yeah weird. Accident.

lzap:
Rebased.

lzap:
Build fails because of two unrecognized UTF-8 codes, I will remove them:

```
asciidoctor: ERROR: failed to parse formatted text: ​<strong>Install</strong> &mdash; to install a new package using the default method. Alternatively, select the drop-down icon to the right of the button and select a method to use. Selecting the <strong>via remote execution - customize first</strong> menu entry will take you to the <strong>Job invocation</strong> page where you can customize the action.
asciidoctor: ERROR: failed to parse formatted text: ​<strong>Update</strong> &mdash; to update an existing package in the host collection using the default method. Alternatively, select the drop-down icon to the right of the button and select a method to use. Selecting the <strong>via remote execution - customize first</strong> menu entry will take you to the <strong>Job invocation</strong> page where you can customize the action.
```

lzap:
For the record, there is a long discussion about implementing `--` and `---` for em/en dashes in asciidoc, steering committee needs to decide because it is a breaking change: https://github.com/asciidoctor/asciidoctor/issues/1174

I am gonna replace these with a minus for now to get builds rolling.

melcorr (approved):

Issue #470, opened 2021-04-01
Get rid of "id ... already in use" warnings


lzap:
There is a trick on how to fix this.

Example patch: https://github.com/theforeman/foreman-documentation/pull/469/files

(Ignore the ifded and makefile change).

We have many of these so I am creating this issue:

```
asciidoctor: WARNING: common/modules/proc_creating-network-or-image-based-hosts.adoc: line 58: id assigned to block already in use: cli-creating-network-or-image-based-hosts
asciidoctor: WARNING: common/modules/proc_adding-images-to-server.adoc: line 84: id assigned to block already in use: cli-adding-images-to-server
asciidoctor: WARNING: common/modules/proc_adding-images-to-server.adoc: line 99: id assigned to block already in use: cli-adding-images-to-server
asciidoctor: WARNING: common/modules/proc_creating-image-only-hosts.adoc: line 51: id assigned to block already in use: cli-creating-image-only-hosts
asciidoctor: WARNING: common/modules/proc_adding-images-to-server.adoc: line 114: id assigned to block already in use: cli-adding-images-to-server
asciidoctor: WARNING: common/modules/proc_creating-image-only-hosts.adoc: line 51: id assigned to block already in use: cli-creating-image-only-hosts
... more
```

Once we get rid of all, we can change the grep in the makefile to also fail on warnings! :-)

tahliar:
Ah, the ones with the cli_ IDs were my fault. If we add _{context} to the IDs and the xrefs, that will solve it. I'd thought they didn't need it since the xref was only to an ID in the same file, but of course file reuse would create duplicate IDs! 🤦 

Big +1 to failing on these warnings; in the downstream tools, broken xrefs and duplicate IDs cause even a local test build to fail, so they're easy to catch. 

Since I caused the problem, I'll fix it.

tahliar:
Opened PR #471 for the cli- IDs. Will open a separate PR for the install guide duplicate IDs.

melcorr:
Thank you @tahliar !

lzap:
Thank you so much! I will look into treating WARNINGS as ERRORS.

PR #471, opened 2021-04-02
Fix duplicate ID errors with {context} IDs

tahliar:
For Issue #470

"It's only an xref to an ID within the same file", I thought when I added the CLI procedure IDs. "It doesn't need a complex ID", I thought. But no, if the file is being reused, the CLI procedure still needs the complex {context} ID (as does the xref to it), so now I've given all of them the appropriate ID format, which I should have just done in the first place.

There are still some duplicate IDs in the install guides, which need a different fix as shown in PR #469.

Cherry-pick into:


melcorr (approved):
Looks good!

lzap:
Thank you for this, we can now treat warnings as errors.

PR #487, opened 2021-04-14
Escape http links in inline code

maximiliankolb:
This PR escapes HTTP links in inline code with a backslash.

In asciidoc, you can escape links using a backslash in inline codes as follows:

```
clickable            vs. not clickable
`http://example.com` vs. `\http://example.com`

```

This makes sense if the docs link to "foreman.example.com/pub/something" etc., as the URL such never be clicked on/resolve.


melcorr:
Hey @maximiliankolb I haven't seen this as an issue in codeblocks. 
Does it cause problems in your docs? 
I've no objection to this, but don't think I saw the problem anywhere.

maximiliankolb:
It happens with inline code, for example [here](https://docs.theforeman.org/nightly/Provisioning_Guide/index-foreman-el.html#adding-openstack-connection_openstack-provisioning). Adding a backslash between "`" and "http" solves this.

I think this creates a slightly better UX as `*.example.com` is never meant to be clicked on anyway.

It does not cause problems on our side, it's just a small annoyance I guess.

melcorr (approved):

PR #489, opened 2021-04-15
Adapt remote execution guide

maximiliankolb:
This PR adapts remote execution files for usage in the documentation of the downstream product orcharhino.

Do you like the 30%-70% width distribution in the table? I feel like it's a small UX improvement. [original 50-50](https://docs.theforeman.org/nightly/Managing_Hosts/index-foreman-el.html#default-job-template-categories_managing-hosts)


melcorr (approved):
Yes! I like the change to the table!

PR #498, opened 2021-04-21
Allow foreman-proxy user to read /etc/rndc.key

maximiliankolb:
Setting `root:named` as the owner and `0640` as the permissions is not sufficient, `foreman-proxy` has to be able to read `/etc/rndc.key`.

Cherry-pick into:

### DIFF
--- a/guides/common/modules/proc_configuring-external-dns.adoc
+++ b/guides/common/modules/proc_configuring-external-dns.adoc
@@ -46,6 +46,8 @@ endif::[]
 # chown -v root:named /etc/rndc.key
 # chmod -v 640 /etc/rndc.key
 ----
++
+Ensure the `foreman-proxy` user can read `/etc/rndc.key`.
 
 . To test the `nsupdate` utility, add a host remotely:
 +


melcorr:
@ekohl  would you mind taking a quick look?

guides/common/modules/proc_configuring-external-dns.adoc
```diff
@@ -46,6 +46,8 @@ endif::[]
 # chown -v root:named /etc/rndc.key
 # chmod -v 640 /etc/rndc.key
 ----
++
+Ensure the `foreman-proxy` user can read `/etc/rndc.key`.
```
ekohl, line 50: This is redundant and why the block on line 65 is there.

ekohl (changes requested):

guides/common/modules/proc_configuring-external-dns.adoc
maximiliankolb: I had information that this is not sufficient.

ekohl: I certainly think this is the wrong place but I do think the two blocks can be merged. Make it roughly something like
> To ensure the `foreman-proxy` user can read `/etc/rndc.key`, assign the `foreman-proxy` user ...

This makes it clear to any reader what the purpose of the action is which can help when something doesn't work.

ekohl (commented):
Btw, this entire procedure is needlessly complicated. What I would suggest is:
```console
# scp root@dns.example.com:/etc/rndc.key /etc/foreman-proxy/rndc.key
# restorecon /etc/foreman-proxy/rndc.key
# chown root:foreman-proxy /etc/foreman-proxy/rndc.key
# chmod 640 /etc/foreman-proxy/rndc.key
# {foreman-installer} --foreman-proxy-dns=true \
  --foreman-proxy-dns-managed=false \
  --foreman-proxy-dns-provider=nsupdate \
  --foreman-proxy-dns-server="DNS_IP_Address" \
  --foreman-proxy-keyfile=/etc/foreman-proxy/rndc.key \
  --foreman-proxy-dns-ttl=86400
```

By using a file in `/etc/foreman-proxy`, you can use it just for this service and make `foreman-proxy` the group. That also allows the service to read the content without needing the `named` group.

Now step 1 is not needed at all. You may need `bind-utils` for `nsupdate` to test, but bind itself is certainly not needed. In the current guide it's installed just for the named group but that's really overkill.
Step 7 is not needed at all and was never needed. The installer will restart the service if needed.
Step 4 (testing) is a good step but I didn't add it here because it can remain unchanged.

ekohl:
For what it's worth, https://bugzilla.redhat.com/show_bug.cgi?id=1776930 is a Bugzilla that was opened for the same thing.

ekohl:
Submitted a PR as https://github.com/theforeman/foreman-documentation/pull/595

PR #508, opened 2021-04-22
Correct IBM POWER client support

lzap:
There was a question raised about this on an internal list, I believe we do support everything except ARM64. Do not merge until a PM confirms this please. @ares or @ehelms ?

Cherry-pick into:

* [x] Foreman 2.4


maximiliankolb:
Any updates @lzap @ares @ehelms ?

lzap:
Well this is a clear case, Little Encoding is the only IBM POWER mode supported in EL8...

ares:
As per my understanding, RHEL 8 only supports LE and this is the right name of the architecture, so merging. Thanks @lzap!

PR #535, opened 2021-05-19
Use anchors and internal crossreferences with {Project_Link}

maximiliankolb:
Hi everyone!

This PR makes anchors and internal crossreferences (`xref:`) more downstream friendly by removing the `Red Hat Satellite` branding.

```
find guides -type f -iname "*.adoc" | xargs sed -i "" "s/{Project_Link}-Administering/Administering/g"
find guides -type f -iname "*.adoc" | xargs sed -i "" "s/Red_Hat_Satellite-Administering_Red_Hat_Satellite/Administering/g"
find guides -type f -iname "*.adoc" | xargs sed -i "" "s/Documentation-Administering_Red_Hat_Satellite/Administering/g"
find guides -type f -iname "*.adoc" | xargs sed -i "" "s/Administering_{Project_Link}/Administering/g"
```

Note that I ran this on macOS; using GNU sed, you will have to remove the `""` after the `-i` option.

It this PR is successful, we could have a look at doing the same for other guides as well.


melcorr:
Hey @maximiliankolb 
Thank you for this PR. 
I have a concern that this might break the Red Hat guide building tool. 
I will try and get confirmation on this. 
I agree in general with what you're doing, and under normal circumstances, your approach is perfect. 
If we can't use your approach, I still want to remove all the references from the links and will worth with you to  do so via another method. 
Bear with me while I try and find someone who might help me.

melcorr:
As it stands, something here breaks the html build for the admin guide.

lzap:
@melcorr is you concern that downstream there is a different attributes.adoc file? In that case we could just add this newly created attribute before we merge this so it will not break?

Anyway, that Red Hat in links is annoying, yeah. We should be able to get rid of it as long as we keep the old ids in pace so existing links will continue working (those which are pointing to "latest" docs).

melcorr:
> @melcorr is you concern that downstream there is a different attributes.adoc file? In that case we could just add this newly created attribute before we merge this so it will not break?
> 
> Anyway, that Red Hat in links is annoying, yeah. We should be able to get rid of it as long as we keep the old ids in pace so existing links will continue working (those which are pointing to "latest" docs).

My concern is that a double variable in the id breaks the build in Pantheon downstream. A total nonsense problem but one that will make Satellite docs impossible to build all the same. I have to ascertain this but I'm pretty sure it's the truth.

maximiliankolb:
> My concern is that a double variable in the id breaks the build in Pantheon downstream.

Should I look into dropping the first `{Project_Link}` in both anchors and links?

melcorr:
> > My concern is that a double variable in the id breaks the build in Pantheon downstream.
> 
> Should I look into dropping the first `{Project_Link}` in both anchors and links?

If you leave it with me for a few days @maximiliankolb I will get an actual answer, so I can add it to our read me and I will help come up with a solution with you. Is that OK?

maximiliankolb:
> > > My concern is that a double variable in the id breaks the build in Pantheon downstream.
> > 
> > 
> > Should I look into dropping the first `{Project_Link}` in both anchors and links?
> 
> If you leave it with me for a few days @maximiliankolb I will get an actual answer, so I can add it to our read me and I will help come up with a solution with you. Is that OK?

That's great. Thanks for looking into it.

melcorr (commented):
Hi @maximiliankolb 
I can confirm that unfortunately two variables in ids will break internal tooling in Red Hat used for Publication. 
I agree that we need to remove these. Your initiative is appreciated and under normal circumstances, your proposed solution would be fine. 
I suggest that we delete the term red hat satellite from the links entirely and just leave at 
"Administering", for example. 
What do you think?

melcorr:
> @melcorr is you concern that downstream there is a different attributes.adoc file? In that case we could just add this newly created attribute before we merge this so it will not break?
> 
> Anyway, that Red Hat in links is annoying, yeah. We should be able to get rid of it as long as we keep the old ids in pace so existing links will continue working (those which are pointing to "latest" docs).

the double variable in the id causes Pantheon to break downstream @lzap - I confirmed this is the case unfortunately for the newer version of Pantheon that they plan to use. Sub optimal.

maximiliankolb:
**draft**

No need to re-review this yet. I am currently trying different approaches ...

maximiliankolb:
These substitutions seem to do the trick, so this PR is now rfr.

melcorr:
This LGTM but will test tomorrow @maximiliankolb thank you !

maximiliankolb:
*ping*

melcorr:
> _ping_

@maximiliankolb I was out sick last week. So sorry for the delay. Will get to this, and your other PRs now!

guides/doc-Administering_Red_Hat_Satellite/topics/Monitoring_Capsules.adoc
```diff
@@ -31,15 +31,15 @@ Click the number to view the details of associated hosts.
...
 
-[[sect-Documentation-Administering_Red_Hat_Satellite-Monitoring_Services]]
+[[sect-Administering-Monitoring_Services]]
 ==== Monitoring Services
 
 Navigate to *Infrastructure* > *{SmartProxies}* and click the name of the selected {SmartProxy}.
 At the *Services* tab, you can find basic information on {SmartProxy} services, such as the list of DNS domains, or the number of Pulp workers.
 The appearance of the page depends on what services are enabled on {SmartProxyServer}.
-Services providing more detailed status information can have dedicated tabs at the {SmartProxy} page (see xref:sect-Documentation-Administering_{Project_Link}-Monitoring_Puppet[]).
+Services providing more detailed status information can have dedicated tabs at the {SmartProxy} page (see xref:sect-Documentation-Administering-Monitoring_Puppet[]).
```
melcorr, line 40: This link is broken. It appears as: 

the Smart Proxy page (see [sect-Documentation-Administering-Monitoring_Puppet]).

melcorr (changes requested):
Hey @maximiliankolb just one issue!
Thank you!

guides/doc-Administering_Red_Hat_Satellite/topics/Monitoring_Capsules.adoc
maximiliankolb: Rebased to master & fixed.

melcorr (approved):
LGTM thank you

PR #538, opened 2021-05-21
Disable unresolvable links

maximiliankolb:
This PR disables some more unresolvable links.

Reason: `X.example.com` won't ever/most likely not resolve, i.e. users cannot/should not click these kinds of links.

See also https://github.com/theforeman/foreman-documentation/pull/487

Question: is the bold and italic on purpose or "should" this be inline code?


ekohl (approved):
It makes sense so me. Any `example.com` link, even if it does resolve, is not meant to be clicked since it's (as the name says) is an example.

However, I'll let some docs person weigh in on this and answer the question.

lzap:
I am not sure why you see any error, we do have `example.com` whitelisted. Oh does this actually fix the link in the rendered page so users cannot click it anymore?

lzap (approved):
Yeah that was it. Thanks.

image

PR #539, opened 2021-05-21
Rename notifications icon file

maximiliankolb:
Minor PR to fix the file name of the `notifications` icon. This image is only used once: `git grep -i -n "notfications1"`


lzap (approved):
TY

PR #543, opened 2021-06-02
Fix highlighting for submit button

maximiliankolb:
Buttons are ~usually~ by convention in bold font and capitalized, i.e. surrounded by `*`.
I've used `git grep -i -n "click submit"` & `git grep "click \*submit"` to find this.


lzap (approved):
Yeah thanks.

PR #544, opened 2021-06-02
Add orcharhino to readme as downstream product

maximiliankolb:
This PR adds orcharhino (Foreman downstream product by ATIX) to the readme.

It also updates the versions of Foreman and Satellite. Note that this is just in the readme, not the attributes which are used throughout the documentation.

### DIFF
--- a/guides/README.md
+++ b/guides/README.md
@@ -56,7 +56,7 @@ Currently there are three different versions:
 
 `make BUILD=katello` - this generates a katello build of the guide
 
-`make BUILD=foreman-deb` - This generates  a version for Foreman installed on Debian.
+`make BUILD=foreman-deb` - This generates a version for Foreman installed on Debian.
 
 `make BUILD=orcharhino` - This generates a downstream preview of the guide for orcharhino.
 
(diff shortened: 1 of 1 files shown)


guides/README.md
```diff
@@ -91,16 +91,16 @@ Please read these guidelines before opening a Pull Request. For more information
...
-| -------- | -------------- | ---------------- |
-| {ProjectNameXY} | Foreman 1.22 | Red Hat Satellite 6.5 |
-| {ProjectNameX} | Foreman | Red Hat Satellite 6 |
-| {ProjectName} | Foreman | Red Hat Satellite |
-| {ProjectXY} | Foreman 1.22 | Satellite 6.5 |
-| {ProjectX} | Foreman | Satellite 6 |
-| {Project} | Foreman | Satellite |
-| {SmartProxyServer} | Smart Proxy server | Capsule Server |
-| {SmartProxy} | Smart Proxy | Capsule |
+| Variable | Upstream value | Downstream by Red Hat | Downstream by ATIX |
```
melcorr, line 94: Do you think we should have the corresponding versions here? 
We'll forget to update it, won't we?

melcorr (commented):
Should we think about how best to manage these references to versions, or just delete them? or say x.y or something?

guides/README.md
maximiliankolb: Yes! I am all for removing it.

If you want to remove the whole table, then we should extend how the attributes work. I am fine with both.

Rebased to master.

guides/README.md
```diff
@@ -91,28 +91,25 @@ Please read these guidelines before opening a Pull Request. For more information
...
 Therefore, never write "Foreman", "Satellite", or "orcharhino" words directly, but use the following variables:
 
-| Variable | Upstream value | Downstream value |
-| -------- | -------------- | ---------------- |
-| {ProjectNameXY} | Foreman 1.22 | Red Hat Satellite 6.5 |
-| {ProjectNameX} | Foreman | Red Hat Satellite 6 |
-| {ProjectName} | Foreman | Red Hat Satellite |
-| {ProjectXY} | Foreman 1.22 | Satellite 6.5 |
-| {ProjectX} | Foreman | Satellite 6 |
-| {Project} | Foreman | Satellite |
```
melcorr, line 101: I think that we should keep form {Project} downwards, if that is OK with you?

melcorr (approved):
Ignore that last comment lol

PR #547, opened 2021-06-03
Set versions to nightly

ekohl:
This guide is for nightly and it should point to that. Without this, it falsely gives the impression it's for 2.4.

### DIFF
--- a/guides/common/attributes.adoc
+++ b/guides/common/attributes.adoc
@@ -3,9 +3,9 @@
 :DocState: nightly
 
 // Version numbers
-:ProjectVersion: 2.4
-:ProjectVersionPrevious: 2.3
-:KatelloVersion: 4.0
+:ProjectVersion: nightly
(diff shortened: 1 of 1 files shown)


ekohl:
https://github.com/theforeman/foreman-documentation/pull/549 fixes it for the 2.5 stable branch.

lzap:
There is one issue tho, this variable is used across the documentation in texts. This then renders as "Foreman nightly" or "Satellite nightly" possibly. I think if you mean to correct installation URLs we should probably introduce a special variable only used for URLs?

ekohl:
I don't see why "Foreman nightly" is a problem. I think it's a feature and the right thing to do. There is no such thing as "Foreman 2.6" until we branch.

lzap (approved):
Generally, it is not a problem. I had some concerns for downstream but Satellite 6 is no longer a marketing thing, Satellite is now just Satellite.

PR #548, opened 2021-06-03
Correct 2.5 version numbers

ekohl:
This is the 2.5 version branch so version numbers should match. It's similar to https://github.com/theforeman/foreman-documentation/pull/547.


ekohl:
Somehow I set the wrong target branch.

PR #549, opened 2021-06-03
Correct 2.5 version numbers

ekohl:
This is the 2.5 version branch so version numbers should match. This is https://github.com/theforeman/foreman-documentation/pull/547 but for stable.


lzap (approved):
Thanks.

PR #552, opened 2021-06-08
Cleared misunderstanding between major and minor versions

ianf77:
Removed the word 'major' so that the text now states you should restore from the same version you backed up. Restoring
from an alternate minor version, for example 6.7 to 6.8 is not supported and may cause problems.

Bug 1846077 - misunderstanding between major and minor version

https://bugzilla.redhat.com/show_bug.cgi?id=1846077

Cherry-pick into:

* [X] Foreman 2.5 (Satellite 6.10)
* [X ] Foreman 2.4
* [X ] Foreman 2.3 (Satellite 6.9)
* [X ] Foreman 2.1 (Satellite 6.8)

### DIFF
--- a/guides/doc-Administering_Red_Hat_Satellite/topics/restoring-from-a-full-backup.adoc
+++ b/guides/doc-Administering_Red_Hat_Satellite/topics/restoring-from-a-full-backup.adoc
@@ -6,7 +6,7 @@ When the restore process completes, all processes are online, and all databases 
 
 .Prerequisites
 * Ensure that you are restoring to the correct instance.
-The {ProjectName} instance must have the same host name, configuration, and be the same version as the original system.
+The {ProjectName} instance must have the same host name, configuration, and be the same minor version (X.Y) as the original system.
 * Ensure that you have an existing target directory.
 The target directory is read from the configuration files contained within the archive.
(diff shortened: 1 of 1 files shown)


guides/doc-Administering_Red_Hat_Satellite/topics/restoring-from-a-full-backup.adoc
```diff
@@ -6,7 +6,7 @@ When the restore process completes, all processes are online, and all databases
 
 .Prerequisites
 * Ensure that you are restoring to the correct instance.
-The {ProjectName} instance must have the same host name, configuration, and be the same major version as the original system.
+The {ProjectName} instance must have the same host name, configuration, and be the same version as the original system.
```
ekohl, line 9: Major was certainly correct, but I think for both upstream we can say it must be the same minor version. Thoughts on:
```suggestion
The {ProjectName} instance must have the same host name, configuration, and be the same minor version (X.Y) as the original system.
```

Otherwise I suspect someone might think it needs to be the same patch version and we get questions about that.

melcorr: +1 thank you @ekohl

melcorr (approved):
LGTM

ianf77:
Hi @melcorr 
I ticked the version checkboxes for cherry-pick but forgot to add the label for this.
Can this be cherry-picked to 2.3 and 2.1?

melcorr:
If you look at this PR, everything is cherry-picked.

PR #568, opened 2021-06-23
Remove link to closed BZ

maximiliankolb:
This PR removes a link to a closed BZ entry. It has been closed with `won't fix`, so I've reworded the sentence accordingly. IMHO it does not make sense to provide this link in our docs.

This PR is **very much up for discussion**, as I don't know how you want to handle such things.


melcorr (approved):
You're so good, @maximiliankolb

melcorr:
I think what you've done here is fair.
If there is a known issue in RH docs, we format with the reference, but in this case, I think you've doe all that you can.

PR #557, opened 2021-06-10
Added port 8443 to Table 1.5. Ports for Satellite to Capsule Communic…

suyogsainkar:
…ation

Bug 1921423 - Documentation missing : Ports for Client to Capsule Communication
https://bugzilla.redhat.com/show_bug.cgi?id=1921423

Cherry-pick into:

* [x] Foreman 2.5 (Satellite 6.10)
* [x] Foreman 2.4
* [x] Foreman 2.3 (Satellite 6.9)
* [x] Foreman 2.1 (Satellite 6.8)

PR #575, opened 2021-06-24
Correct the 'userdata' url in Provisioning Guide

ryandeussing:
Bug 1929003 - Incorrect 'userdata' url in the Satellite 6.8 docs
https://bugzilla.redhat.com/show_bug.cgi?id=1929003

Cherry-pick into:

* [x] Foreman 2.4

### DIFF
--- a/guides/doc-Provisioning_Guide/topics/Virt-VMWare.adoc
+++ b/guides/doc-Provisioning_Guide/topics/Virt-VMWare.adoc
@@ -364,7 +364,7 @@ endif::[]
 datasource_list: [NoCloud]
 datasource:
   NoCloud:
-    seedfrom: https://{foreman-example-com}/userdata/
+    seedfrom: https://{foreman-example-com}/unattended/userdata/
 EOF
 ----
 +


ryandeussing:
@melcorr 🤦 I jumped the gun in creating this PR, please ignore it until I get verification that the change belongs in all the branches I believe it does. 

If all goes to plan, and my permissions are correct, I can make this my first PR merge as well.

melcorr:
@ezr-ondrej does this look right to you?

lzap (changes requested):
This is incorrect, the cloud-init endpoint really is exposed at `/userdata` and not `/unattended`. It is a completely different controller.

```
$ be rake routes | grep userdata
 userdata_meta_data GET      /userdata/meta-data(.:format)                                                                                                                          userdata#metadata {:format=>/text/}
 userdata_user_data GET      /userdata/user-data(.:format)                                                                                                                          userdata#userdata {:format=>/text/}
```

When you test this, this might be misleading:

```
[lzap@nuc foreman]$ curl http://localhost:5000/userdata/meta-data
Could not find host for request ::ffff:127.0.0.1
[lzap@nuc foreman]$ curl http://localhost:5000/unattended/userdata/meta-data
unattended: unable to find a host that matches the request from ::ffff:127.0.0.1
```

Both endpoints do perform search in the database against REMOTE_IP, but the first one is correct while the other request is trying to find a host to apply `userdata/meta-data` template kind which does not exist and it would reply with error.

melcorr:
@ryandeussing we had a phrase in docs when I was on the team: "Trust No One" :o) 
We often have items reported that turn out not to be true when an engineer looks at it. 
Here is another example from Suyog from receently: https://github.com/theforeman/foreman-documentation/pull/557#discussion_r650055515 
Sadly it's not just a matter of taking suggestions, but also investigating each claim.

lzap:
Feel free to ping us beforehand to save time! In this case, there is a reason why the endpoint has a different path, it's because this is what cloud-init expects. Cheers!

PR #581, opened 2021-07-01
FIPS is only tested on RHEL

lzap:
SSIA FAIK only Red Hat Satellite QA team does that.

### DIFF
--- a/guides/common/modules/ref_system-requirements.adoc
+++ b/guides/common/modules/ref_system-requirements.adoc
@@ -91,7 +91,7 @@ Installation with disabled SELinux is not supported.
 .FIPS Mode
 You can install {ProductName} on a {RHEL} system that is operating in FIPS mode.
 ifndef::satellite[]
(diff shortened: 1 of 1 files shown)


guides/common/modules/ref_system-requirements.adoc
```diff
@@ -90,5 +90,8 @@ Installation with disabled SELinux is not supported.
 
 .FIPS Mode
 You can install {ProductName} on a {RHEL} system that is operating in FIPS mode.
+ifndef::satellite[]
+{RHEL} clones are not being actively tested in FIPS mode, we do recommend to use {RHEL}.
```
melcorr, line 94:
```suggestion
{RHEL} clones are not being actively tested in FIPS mode. If you require FIPS, consider using {RHEL}.
```

melcorr (changes requested):
Let's not recommend RHEL so strongly.

lzap:
"I do recommend" is my phrase, alongside with "you know" and several others :-)

PR #585, opened 2021-07-07
use HTTP not a placeholder servicename when registering for IdM

evgeni:
our installer hard-codes HTTP and using _servicename_ here would imply
that the user can change this, while they can not.

Cherry-pick into:

* [x] Foreman 2.4


melcorr (approved):
Thank you!

PR #586, opened 2021-07-08
Copy landing page as index.html and improve 404

lzap:
This will copy landing version page (e.g. `2.5.html`) also as `2.5/index.html`. So if readers manually delete URL to:

https://docs.theforeman.org/2.5/

instead of "Page not found!" with nightly links, there will be proper 2.5 landing page instead.

To test this, download foreman-docs-nightly.zip and unpack the artifact, there should be nightly/index.html as a copy of nightly.html.

The change also removes list of nightly docs from the 404 page and only provide a link to the root for less confusion.

Cherry-pick into:

* [x] Foreman 2.4
* other versions at will

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -182,6 +182,16 @@ jobs:
         run: hugo -v --ignoreCache
         working-directory: web
 
+      - name: Copy nightly landing page as index.html
+        if: github.repository_owner == 'theforeman' && github.ref == 'refs/heads/master'
+        run: cp -f nightly.html nightly/index.html
+        working-directory: web/public
+
+      - name: Copy stable landing page as index.html
+        if: github.repository_owner == 'theforeman' && (startsWith(github.ref, 'refs/heads/2.') || startsWith(github.ref, 'refs/heads/3.') || startsWith(github.ref, 'refs/heads/4.'))
+        run: cp -f ${{ env.BRANCH_NAME }}.html ${{ env.BRANCH_NAME }}/index.html
+        working-directory: web/public
+
       - name: Upload web
         uses: actions/upload-artifact@v2
         with:
--- a/web/content/404/index.md
+++ b/web/content/404/index.md
@@ -3,4 +3,4 @@ title: 404
 headless: true
 ---
 
-{{% guide-list version="nightly" %}}
+Continue to the [front page](https://docs.theforeman.org) to find available documentation.
--- a/web/layouts/404.html
+++ b/web/layouts/404.html
@@ -19,8 +19,6 @@
 </script>
 </p>
 
-<p>Available guides for the latest nightly build:</p>
-
 {{ with site.GetPage "/404" }}
 {{ .Content }}
 {{ end }}


.github/workflows/deploy.yml
```diff
@@ -144,6 +144,16 @@ jobs:
...
         working-directory: web
 
+      - name: Copy nightly landing page as index.html
+        if: github.repository_owner == 'theforeman' && github.ref == 'refs/heads/master'
+        run: cp -f nightly.html nightly/index.html
+        working-directory: web/public
+
+      - name: Copy stable landing page as index.html
+        if: github.repository_owner == 'theforeman' && (startsWith(github.ref, 'refs/heads/2.') || startsWith(github.ref, 'refs/heads/3.') || startsWith(github.ref, 'refs/heads/4.'))
+        run: cp -f ${{ env.BRANCH_NAME }}.html ${{ env.BRANCH_NAME }}/index.html
```
ekohl, line 154: I checked out the 2.5 branch and ran `hugo` but there I don't see a `2.5.html` file. How would that be built?

ekohl (commented):
Overall this looks sensible to me but something appears to be off in the 2.5 branch.

.github/workflows/deploy.yml
lzap: The whole web/ directory is a static site, it should pick the `web/content/2.5.md` file and generate the HTML.

Note it is always built from master. Just use master. I think our GHA is confusing - we build from the branch which is being built and copy files over not deleting contents of the result directory, that's why we still see 2.5 even after GHA with an older contents is triggered. I think we should change the action to always pull `master` to build the web portion of the site.

lzap: Yeah let's fix this: https://github.com/theforeman/foreman-documentation/pull/590

ekohl: That is now merged. Does that mean changes to this PR are now needed? Since we now always build from master, does it mean we can somehow teach hugo to do this for us instead of manually working around it?

lzap: This is still needed, the problem here is that hugo is used to build the whole site (main index, css, pages etc) while in the root directory we also have `/X.Y` directories which are generated by a separate workflow (`asciidoctor`). And these directories are not managed by hugo, I would like not to go there as I found hugo quite complex, way more complex than jekyll. I was hoping it was more simple, it advertises itself as a lightweight but its not.

One day I was even thinking ditching it completely for just Ruby ERB, we have a such a simple use case that it is probably an overkill.

melcorr:
I'll wait for the green light from @ekohl before I merge.

lzap:
We should be all set.

lzap:
@ekohl bump?

lzap:
If we could land this before 3.1 branching that would be awesome, thanks! Maybe @maximiliankolb could help since @ekohl is busy with getting the bits out.

maximiliankolb:
I cannot verify this: I cannot find the `nightly` directory. I've downloaded the `foreman-docs-web-master.zip` artifact.

lzap:
Rebased.

maximiliankolb:
Again, @lzap , I'm sorry but I don't really know how to review this. I've downloaded both artifacts but did not know what to look for.

lzap:
Let's keep the status quo for now, migrate to Nanoc to simplify all of this and then migrate to Pantheon later on.

maximiliankolb:
Just for the record:

> and then migrate to Pantheon later on.

This did not go by unnoticed.

If there are any serious plans, please shed some light on the community.

melcorr:
I don't understand the Pantheon comment either.

lzap:
Sorry messed up wording, I mean Antora of course! :)

maximiliankolb:
Ah, now it makes more sense. All good @lzap

PR #584, opened 2021-07-07
Added a new chapter/assembly about using foreman webhooks

suyogsainkar:
https://issues.redhat.com/browse/SAT-3106
[doc] Foreman Webhooks

Cherry-pick into:

* [ ] Foreman 2.5 (Satellite 6.10)
* [ ] Foreman 2.4
* [ ] Foreman 2.3 (Satellite 6.9)
* [ ] Foreman 2.1 (Satellite 6.8)

PR #588, opened 2021-07-08
Correct index + footer

ekohl:
See the individual commits for details. In short:
* It changes the link order on the index
* It links to latest official docs rather than nightly
* Corrects the source link
* Add a full license text to the repo

### DIFF
--- a/web/content/_index.md
+++ b/web/content/_index.md
@@ -4,14 +4,14 @@ title: "Foreman and Katello documentation"
 
 This is work-in-progress documentation site for <a href="https://www.theforeman.org">Foreman</a> open-source software.
 
-Official documentation [is available here](https://theforeman.org/manuals/nightly/index.html).
+Official documentation [is available here](https://theforeman.org/manuals/latest/index.html).
 
 Available upstream documentation sets are:
 
 * [Foreman and Katello Nightly](/nightly.html)
-* [Foreman 2.3 and Katello 3.18](/2.3.html)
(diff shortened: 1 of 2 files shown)


melcorr (approved):
Thanks @ekohl

lzap:
Right on time, I raised one concern here: https://github.com/theforeman/foreman-documentation/pull/584#discussion_r667666461

We copy a paragraph from Wikipedia, IANAL but we should give them attribution: https://creativecommons.org/use-remix/attribution/

I think the reasonable approach would be to create AUTHORS file programmatically from git history + adding something like "Wikipedia, https://www.wikipedia.org". That would be the bare minimum and I think we could implement it in this PR.

We could go little further and put Wikipedia on the front page of the guide, but for upstream we only generate HTML versions and there is not much there. Which is actually not correct, I think the HTML versions of upstream docs should contain copyright information, perhaps with a link to the AUTHORS file here on github. I think we can do this by modifying `revdate` in `guides/common/header.adoc` or we can use revremark: https://docs.asciidoctor.org/asciidoc/latest/document/revision-information/

ekohl:
Shall I split off the license bit in a separate PR?

ekohl:
So that's https://github.com/theforeman/foreman-documentation/pull/599.

lzap (approved):
No objections. I can add AUTHORS file.

lzap:
AUTHORS: https://github.com/theforeman/foreman-documentation/pull/601

PR #590, opened 2021-07-08
Always build web from the master

lzap:
Ewoud noticed that we actually build the web portion from the topic/stable branch, but since we do not overwrite files (due to documentation portion of the site) files do not get deleted and we haven't noticed. This fixes it, everything under `/web` should be built from `master`.

There is another solution to create a dedicated branch just for the web/ portion, however I do not like having two different content sets in two branches. This creates unecessary long checkouts when switching over, we have this problem with foreman-packaging (rpm/develop vs deb/develop) and I ended up having two separate checkouts for that. I would rather keep this approach when all content is in the single branch - master. This is where most of the development/writing takes place.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -124,16 +124,10 @@ jobs:
       run:
         shell: bash
     steps:
-      - name: Get branch name (merge)
-        if: github.event_name != 'pull_request'
-        run: echo "BRANCH_NAME=$(echo ${GITHUB_REF#refs/heads/} | tr / -)" >> $GITHUB_ENV
-
-      - name: Get branch name (pull request)
-        if: github.event_name == 'pull_request'
-        run: echo "BRANCH_NAME=$(echo ${GITHUB_HEAD_REF} | tr / -)" >> $GITHUB_ENV
-
       - name: Checkout
         uses: actions/checkout@v2
+        with:
+          ref: master
 
       - name: Setup Hugo
         uses: peaceiris/actions-hugo@v2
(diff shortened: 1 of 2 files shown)


ekohl (commented):
On line 151 it refers to `env.BRANCH_NAME` but it feels like that is no longer accurate then. I also wonder what should be done about the `Get branch name` step at line 131.

lzap:
Right, these actions can be deleted, they are not used when building web artifact. Removed. Also added a line into the README about this.

.github/workflows/deploy.yml
```diff
@@ -124,16 +124,9 @@ jobs:
...
-        if: github.event_name != 'pull_request'
-        run: echo "BRANCH_NAME=$(echo ${GITHUB_REF#refs/heads/} | tr / -)" >> $GITHUB_ENV
-
-      - name: Get branch name (pull request)
-        if: github.event_name == 'pull_request'
-        run: echo "BRANCH_NAME=$(echo ${GITHUB_HEAD_REF} | tr / -)" >> $GITHUB_ENV
-
       - name: Checkout
         uses: actions/checkout@v2
+        ref: master
```
ekohl, line 129:
```
Invalid workflow file : .github/workflows/deploy.yml#L129
The workflow is not valid. .github/workflows/deploy.yml (Line: 129, Col: 9): Unexpected value 'ref'
```

lzap: Forgot to add `with:`, fixed.

ekohl (approved):

melcorr (approved):

PR #595, opened 2021-07-09
BZ1776930: Simplify the remote DNS instructions

ekohl:
By using a file in /etc/foreman-proxy, you can use it just for this service and make foreman-proxy the group. That also allows the service to read the content without needing the named group.

Now step 1 is not needed at all. You may need bind-utils for nsupdate to test, but bind itself is certainly not needed. In the current guide it's installed just for the named group but that's really overkill.

Step 7 is not needed at all and was never needed. The installer will restart the service if needed.

Step 4 (testing) is a good step but I think example.com is a better domain than virtual.lan.

Note that this instruction now only works well if you have an existing installation where the foreman-proxy group is present. So carefully review this.

I also removed restorecon on Debian where we SELinux is typically not present.

Based on https://github.com/theforeman/foreman-documentation/pull/498#pullrequestreview-640968680

### DIFF
--- a/guides/common/modules/proc_configuring-external-dns.adoc
+++ b/guides/common/modules/proc_configuring-external-dns.adoc
@@ -9,42 +9,26 @@ To make any changes persistent, you must enter the `{foreman-installer}` command
 .Prerequisites
 
 * You must have a configured external DNS server.
+* This guide assumes you have an existing installation.
 
 .Procedure
-
-ifndef::foreman-deb[]
-. Install the `bind-utils` package:
-+
-[options="nowrap" subs="+quotes,attributes"]
-----
-# {package-install} bind bind-utils
-----
-endif::[]
-
-ifdef::foreman-deb[]
-
-. Install the `bind-utils` package:
-+
-[options="nowrap" subs="+quotes,attributes"]
-----
-# {package-install} bind9
-----
-endif::[]
 
 . Copy the `/etc/rndc.key` file from the external DNS server to {ProductName}:
 +
 [options="nowrap" subs="+quotes"]
 ----
-# scp root@_dns.example.com_:/etc/rndc.key /etc/rndc.key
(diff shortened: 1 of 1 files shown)


lzap (approved):
Nice.

guides/common/modules/proc_configuring-external-dns.adoc
```diff
@@ -12,60 +12,35 @@ To make any changes persistent, you must enter the `{foreman-installer}` command
...
 ----
 
 . Configure the ownership, permissions, and SELinux context:
 +
 [options="nowrap"]
 ----
-# restorecon -v /etc/rndc.key
-# chown -v root:named /etc/rndc.key
-# chmod -v 640 /etc/rndc.key
+ifndef::foreman-deb[]
```
lzap, line 26: Error outs with `asciidoctor: WARNING: common/modules/proc_configuring-external-dns.adoc: line 25: unterminated listing block`

guides/common/modules/proc_configuring-external-dns.adoc
```diff
@@ -12,60 +12,35 @@ To make any changes persistent, you must enter the `{foreman-installer}` command
...
+# restorecon -v /etc/foreman-proxy/rndc.key
+endif::
```
melcorr, line 28:
```suggestion
endif::[]
```
Should fix that

ekohl:
> Note that this instruction now only works well if you have an existing installation where the foreman-proxy group is present. So carefully review this.

Is this acceptable? Should we note this to the reader?

ekohl:
Updated:
* Added a prerequisite
* Fixed the syntax

melcorr (approved):
This looks good. 
@ekohl should I cherry pick for the 2.5 release?

ekohl:
I'm not sure. It's a pretty big change and I don't know how much risk we want to take. On the other hand, the current instructions also have issues. For upstream I'd be fine with this, can't answer for downstream.

PR #597, opened 2021-07-12
Remove Satellite 6.8 from GHT

lzap:
That's gone for us, RHD team should file PRs for "master" and cherry pick into older releases themselves. We will continue doing 6.9 but long term, I would like only to cherry pick into last two upstream release. What you think?


melcorr:
Let's wait until 6.10!

PR #598, opened 2021-07-12
Make Nightly ribbon a link

lzap:
I think it is a good idea to make this a link so users can actually click to github for possible contributions.


melcorr (approved):
Even better.

ekohl:
Maybe not the right place, but it would be good to include this in the branching that it should be modified so in the stable releases we have the same thing.

PR #600, opened 2021-07-12
Deprecations for 6.10

ianf77:
Marked items as being deprecated in the next version.

[doc] 6.10 deprecations

https://issues.redhat.com/browse/SAT-4179

Cherry-pick into:

* [x] Foreman 2.4


melcorr:
You've added this to the downstream release notes too ?

ianf77:
Not yet. It's on the list.

ianf77:
Actually it looks like Suyog did it already.

guides/doc-Upgrading_and_Updating/topics/upgrading_satellite_server_prerequisites.adoc
```diff
@@ -9,6 +9,9 @@ For more information, see {InstallingProjectDocURL}preparing-environment-for-sat
 * Ensure that you do not delete the manifest from the Customer Portal or in the {Project} Web UI because this removes all the entitlements of your content hosts.
 * Back up and remove all Foreman hooks before upgrading.
 Restore any hooks only after {Project} is known to be working after the upgrade is complete.
++
+Note that Foreman Hooks functionality is deprecated and will be removed in the next {Project} version.
```
melcorr, line 13: No need for the blank lines before and after this.

melcorr (changes requested):
One small requested change. 
Any questions, let me know.

guides/doc-Upgrading_and_Updating/topics/upgrading_satellite_server_prerequisites.adoc
ianf77: Changes made.

guides/doc-Planning_Guide/topics/Glossary.adoc
```diff
@@ -130,7 +130,9 @@ In relationship to a host, erratum is *applicable* if it updates a package insta
 [[varl-Glossary_of_Terms-External_Node_Classifier]]
 *External Node Classifier*:: A Puppet construct that provides additional data for a Puppet server to use when configuring hosts.
 {ProjectName} acts as an External Node Classifier to Puppet servers in a {Project} deployment.
-
++
+Note that the Puppet External Node Classifier is deprecated and will be removed in the next {Project} version.
```
ekohl, line 134: This is a bit more nuanced. In upstream it'll be moved to [foreman_puppet](https://github.com/theforeman/foreman_puppet) (currently undecided if it'll be included by default or not, also undecided how we'll handle it in upgrades: https://github.com/theforeman/foreman-installer/pull/695).

In downstream the foreman_puppet won't be shipped so there it will be removed.

To make it even more confusing, https://github.com/theforeman/foreman/pull/8510 (as I understand it now) will remove `/node/:certname` but not `/api/hosts/:id/enc` which is the same data.

The crucial part is that the environment and classes are removed from the ENC.

guides/doc-Planning_Guide/topics/Glossary.adoc
```diff
@@ -156,7 +158,9 @@ To see the full list of services, enter the `{foreman-maintain} service list` co
 
 [[varl-Glossary_of_Terms-Foreman_Hooks]]
 *Foreman Hook*:: An executable that is automatically triggered when an orchestration event occurs, such as when a host is created or when provisioning of a host has completed.
-
++
+Note that Foreman Hook functionality is deprecated and will be removed in the next {Project} version.
```
ekohl, line 162: I think this is true for downstream only. While I think @lzap considers the classic hooks deprecated (in favor of webhooks), we don't ship it by default in upstream. That means it's not really removed as such. Also, you're still able to install it in upstream.

guides/doc-Planning_Guide/topics/Glossary.adoc
```diff
@@ -130,7 +130,9 @@ In relationship to a host, erratum is *applicable* if it updates a package insta
 [[varl-Glossary_of_Terms-External_Node_Classifier]]
 *External Node Classifier*:: A Puppet construct that provides additional data for a Puppet server to use when configuring hosts.
```
ekohl, line 131: I still disagree with this. While Puppet introduced it and we follow its format, I still think that the parameters it provides are also very useful in other cases. However, it appears I can't really get through to people so I've given up.

melcorr: I have no context for this comment @ekohl 
I'd gladly help get to where you think this needs to be, but have no idea about Puppet.

ekohl: I've been trying to convince people that Foreman should adopt the term ENC and not consider it a Puppet-only concept. However, this appears to fall on deaf ears. Maybe at this point I'm just ranting.

guides/doc-Content_Management_Guide/topics/Managing_OSTree_Branches.adoc
```diff
@@ -8,6 +8,17 @@ Red Hat Enterprise Linux Atomic Server uses OSTree content composed from RPM fil
...
+[NOTE]
+====
+The following management operations of OSTree and Puppet content types are deprecated and will be removed in the next release:
+
+* Creating OSTree and puppet repositories.
+* Synchronizing OSTree and puppet repositories.
+* Adding repositories of OSTree and puppet content types to content view.
+* Publishing and promoting the content views containing repositories of OSTree and puppet content types across life cycle environments.
+
+====
```
melcorr, line 20: You can remove this entire note as all of this content has been removed. It was deprecated in the previous release. 
This should be added release notes about removals.

guides/doc-Planning_Guide/topics/Glossary.adoc
melcorr: This is also removed rather than deprecated.

guides/doc-Planning_Guide/topics/Glossary.adoc
melcorr: And how would that look? 
```suggestion
*External Node Classifier*:: A construct that provides additional data for a server to use when configuring hosts.
```
Like this?

guides/doc-Planning_Guide/topics/Glossary.adoc
melcorr: If it's for downstream, wrap like this: 
```suggestion
ifdef::satellite[]
Note that Foreman Hook functionality is deprecated and will be removed in the next {Project} version.
endif::[]
```

guides/doc-Upgrading_and_Updating/topics/upgrading_satellite_server_prerequisites.adoc
melcorr: As per Ewoud's feedback, wrap this in a Satellite ifdef statement.

melcorr (changes requested):

guides/doc-Upgrading_and_Updating/topics/upgrading_satellite_server_prerequisites.adoc
ianf77: Changed as per comment

guides/doc-Planning_Guide/topics/Glossary.adoc
ianf77: Changed as per comment.

guides/doc-Planning_Guide/topics/Glossary.adoc
ianf77: Changed as per comment.

guides/doc-Content_Management_Guide/topics/Managing_OSTree_Branches.adoc
ianf77: Changed as per comment.

guides/doc-Planning_Guide/topics/Glossary.adoc
ianf77: Changed as per comment.

PR #603, opened 2021-07-13
Remove use of port 5000 for running containers

ekohl:
I suspect this was previously used for Crane (which ran on port 5000) but Crane is gone and replaced by Pulp 3 which uses port 443.


ehelms (approved):

melcorr (commented):
Does this need cherry-picking to 2.5 then?

melcorr:
@ekohl happy to merge this but it looks like it should be in 2.5 also?

ekohl:
Yes, it also applies to 2.5.

PR #616, opened 2021-07-19
Fix incorrect markup on FAM attribute

melcorr:
@evgeni pointed out on IRC that a FAM example
was rendering incorrectly in our docs.
https://docs.theforeman.org/nightly/Administering_Red_Hat_Satellite/index-foreman-el.html#using-satellite-ansible-content-collections_admin

The attributes files included backticks
that also appeared in the codeblocks.

Best to format on the namespace as needed 
as the formatting limited its usage. 

@maximiliankolb - I didn't touch orcharhino as you have not exposed this section in your docs yet: https://docs.theforeman.org/nightly/Administering_Red_Hat_Satellite/index-orcharhino.html#using-satellite-ansible-content-collections_admin 
I'm adding you as a reviewer as you might need me to adjust the orcharhino attributes also. 

Cherry-pick into:

* [ ] Foreman 2.5 (Satellite 6.10)
* [ ] Foreman 2.4
* [ ] Foreman 2.3 (Satellite 6.9)
* [ ] Foreman 2.1 (Satellite 6.8)

PR #617, opened 2021-07-19
Remove unused orcharhino attributes

maximiliankolb:
This PR removes unused attributes specifically for build target 'orcharhino'.

See also #616.


melcorr (approved):
Thanks for the updates!

PR #515, opened 2021-04-28
Branch docs for Foreman 2.5

upadhyeammit:

PR #618, opened 2021-07-20
Proposal: Add Upstream release notes guide

melcorr:
**Problem Description**

The way that Foreman and Katello release notes are handled on docs.theforeman.org has some problems: 

1. Difficult to find: you have to know to scroll past all the guide links to view them: 
2. Not possible to link to: there are no anchors with which you can share the list of headlines, deprecations etc
3. Katello at the moment are just dumping their release notes under the Foreman notes. This requires even more scrolling for someone interested in Katello

**Solution**

I propose to create dedicated upstream release notes guides for both Foreman and Katello. 
I created a new guide with two files, one for Katello release notes and one for Foreman. I copied in the 2.5 release notes to test. 

## Issues that need to be resolved before this can be merged 

1. Our Asciidoc that we use here is close to but not fully compatible with markdown. The main issue that I encountered was that the Foreman release notes have links to redmine issues formatted in the following way ([name](link)) - for example - ([#28311](https://projects.theforeman.org/issues/28311)) - these outer parentheses are not parsed and break the formatting of the links. 
Would it be possible to update the script to remove these parentheses, or add a sed to the end of the script to remove those parentheses? @upadhyeammit ?
2. When I look at the release notes added from the branching of 2.5: https://github.com/theforeman/foreman-documentation/pull/515/files I notice that after deprecations, everything is left at a level 4 header level and makes it look like everything afterwards is a deprecation, for example: https://github.com/theforeman/foreman-documentation/pull/515/files#diff-9030df5060ebc6f9c46fd4c5a6b9a75b6e00b1132e7ba570f4dfd1768da2fae9R64 - is it easy to edit so that these become a heading level higher? 
3. When we autogenerate Katello release notes, is it possible to include the redmine issues? That would be great. 

**Next steps**

If this proposal is accepted I will add a link to the two sets of Release Notes at the very top of the guides list. 

Let me know what you (plural) think. 

PS: Handling Satellite release notes is out of the scope of this PR but is very easy to include them in a future migration.


ekohl:
>  Would it be possible to update the script to remove these parentheses, or add a sed to the end of the script to remove those parentheses? @upadhyeammit ?

That should be easy:

https://github.com/theforeman/theforeman.org/blob/0820f26900f197e2c290f5a667027f9b4333055f/scripts/release_notes.rb#L81

> When we autogenerate Katello release notes, is it possible to include the redmine issues? That would be great.

The Foreman script has an option "projects" so calling it as `./release_notes.rb katello 4.1.1` should generate the same report.

What I've been wondering if for Katello we can actually generate a changelog that includes Foreman. That way we don't have to say "also read Foreman's notes at that location" but have a single document.

melcorr:
> What I've been wondering if for Katello we can actually generate a changelog that includes Foreman. That way we don't have to say "also read Foreman's notes at that location" but have a single document.

This could be handled quite easily with an if statement actually
in our master.adoc for Katello builds, we could print Katello first and then add a line "And here are the corresponding Foreman release notes" (or whatever we want) and print the Foreman release notes beneath.

ekohl:
Should we then include the logs in a chronological order? So:
* Katello 4.1.1
* Foreman 2.5.2
* Foreman 2.5.1
* Katello 4.1.0
* Foreman 2.5.0

Or first all Katello, then all Foreman? I'm leaning to mixing it because that's the order it's released. It also means that Foreman releases must publish new Katello changelogs, but it's a nice service to users.

melcorr:
> Should we then include the logs in a chronological order? So:
> 
>     * Katello 4.1.1
> 
>     * Foreman 2.5.2
> 
>     * Foreman 2.5.1
> 
>     * Katello 4.1.0
> 
>     * Foreman 2.5.0
> 
> 
> Or first all Katello, then all Foreman? I'm leaning to mixing it because that's the order it's released. It also means that Foreman releases must publish new Katello changelogs, but it's a nice service to users.

I'm unsure if I fully follow here. I will need to think about this a bit.

melcorr:
> > Should we then include the logs in a chronological order? So:
> > ```
> > * Katello 4.1.1
> > 
> > * Foreman 2.5.2
> > 
> > * Foreman 2.5.1
> > 
> > * Katello 4.1.0
> > 
> > * Foreman 2.5.0
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Or first all Katello, then all Foreman? I'm leaning to mixing it because that's the order it's released. It also means that Foreman releases must publish new Katello changelogs, but it's a nice service to users.
> 
> I'm unsure if I fully follow here. I will need to think about this a bit.

To implement this, we would need a separate .adoc file for each z-stream release, but it is possible.

ekohl:
> To implement this, we would need a separate .adoc file for each z-stream release, but it is possible.

I think that would be a good idea regardless. It means you can run `./changelog.rb $project $version > $version.adoc` and don't have to merge manually.

melcorr:
> > To implement this, we would need a separate .adoc file for each z-stream release, but it is possible.
> 
> I think that would be a good idea regardless. It means you can run `./changelog.rb $project $version > $version.adoc` and don't have to merge manually.

Can I clarify one thing @ekohl 

We publish two separate release notes guides, one for Foreman and one for Katello? 

The guides would build from a single source, but would have the Foreman or Katello release notes up front and then the corresponding Foreman or Katello release info after it. We would just link to both the foreman-el.html and the katello.html from docs.theforeman.org 

For example, for Katello it would look like this:

```
ifdef::katello[]
include:: Katello 4.1.1.adoc
<Some text to introduce the Foreman release notes>
include:: Foreman 2.5.2.adoc
include:: Foreman 2.5.1.adoc
## Previous versions
<some text to introduce previous z-stream versions>
include:: Katello 4.1.0.adoc
include:: Katello Foreman 2.5.0.adoc
endif::[]
```

This is what I have in my mind, just want to make sure we are on the same page.

ekohl:
Pretty much that yes. The only thing I wonder is how it'll look like in between. So the time line is:

* Foreman 2.5.0-rc1 is released
At this point having Katello release notes doesn't make a lot of sense since there is no Katello
* Katello 4.1.0-rc1 is released

Now the katello release notes look like:
```asciidoc
include:: Katello 4.1.0.adoc
include:: Foreman 2.5.0.adoc
```

Then over time the sections will be updated until both Foreman and Katello are GA. Our current process is that we don't post notes for individual RCs, we just update them all. If users really want to, they can read the git diff.

So it'll still look the same. Then Foreman 2.5.1 is released. Now it looks like this?

```asciidoc
include:: Foreman 2.5.1.adoc
include:: Katello 4.1.0.adoc
include:: Foreman 2.5.0.adoc
```

And chronologically we then released Foreman 2.5.2 and then Katello 4.1.1.

melcorr:
I was hoping to merge this today with the release candidate but I need a slight change in this script: 

https://github.com/theforeman/theforeman.org/blob/0820f26900f197e2c290f5a667027f9b4333055f/scripts/release_notes.rb#L81

One change I already asked for has been completed, but now I realise I need the script to be updated for urls to be changed from the following: 

For example: 

[#32139](https://projects.theforeman.org/issues/32139)

To: 

https://projects.theforeman.org/issues/32139[#32139]

lzap:
https://github.com/theforeman/theforeman.org/pull/1880

guides/Release_Notes/master.adoc
```diff
@@ -0,0 +1,29 @@
+include::common/attributes.adoc[]
```
ekohl, line 1: The directory here isn't named `doc-Release_Notes`. Is that intentional?

ekohl (changes requested):
This now needs an update. In https://github.com/theforeman/foreman-documentation/pull/698 I've submitted a skeleton version of this.

ekohl:
I think https://github.com/theforeman/foreman-documentation/pull/698 and https://github.com/theforeman/foreman-documentation/pull/699 implement the changes. What's then missing is the update to the actual `web/content/*.md`, depending on how far back we cherry pick.

Note that the former PR includes a script to generate the notes. That should also be cherry picked for convenience once it's done.

guides/Release_Notes/master.adoc
```diff
@@ -0,0 +1,29 @@
...
+endif::[]
+
+:upstream-release-notes:
+:context: {project-context}
+:ProductName: {ProjectServer}
+
+ifdef::foreman-el[]
+include::common/modules/ref_foreman-3-0-release-notes.adoc[]
+
+## Corresponding Katello release notes
```
maximiliankolb, line 16:
```suggestion
== Corresponding Katello release notes
```

🙃

guides/Release_Notes/master.adoc
```diff
@@ -0,0 +1,29 @@
...
+
+
+include::common/modules/ref_katello-release-notes.adoc[]
+endif::[]
+
+
+ifdef::katello[]
+include::common/modules/ref_katello-release-notes.adoc[]
+
+## Corresponding Foreman release notes
```
maximiliankolb, line 26:
```suggestion
== Corresponding Foreman release notes
```

guides/common/modules/ref_katello-4-0-release-notes.adoc
```diff
@@ -0,0 +1,58 @@
+[id="katello-release-notes_{context}"]
+
+# Katello Release Notes
```
maximiliankolb, line 3:
```suggestion
= Katello Release Notes
```

guides/common/modules/ref_katello-4-0-release-notes.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+// Use `//` to comment headings out as necessary.
+
+## Features
```
maximiliankolb, line 7:
```suggestion
== Features
```

guides/common/modules/ref_katello-4-0-release-notes.adoc
```diff
@@ -0,0 +1,58 @@
...
+  - Automatically create content view on import
+  - Import/export now supports File and Ansible repo types
+  - `hammer content-import list` command now available
+* Improved handling of Activation Keys and other Katello-related items for Global Host Registration feature
+* More updates to new Content View preview UI at /labs/content_views
+* Upgraded to pulpcore 3.11
+* Added global_download_concurrency setting (defaults to 5)
+* Katello now supports logging correlation_id for pulp3 operations to make it easier to debug across different log files
+
+## Bug Fixes
```
maximiliankolb, line 21:
```suggestion
== Bug Fixes
```

guides/common/modules/ref_katello-4-0-release-notes.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+## Deprecations
```
maximiliankolb, line 23:
```suggestion
== Deprecations
```

guides/common/modules/ref_katello-4-0-release-notes.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+* API endpoints for katello-agent actions have been deprecated. Consider switching to remote execution.
+* New deprecation warning on the Register Content Host page - Use Global Registration instead (/hosts/register)
+
+## Removals
```
maximiliankolb, line 28:
```suggestion
== Removals
```

guides/common/modules/ref_katello-4-0-release-notes.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+## Contributors
```
maximiliankolb, line 30:
```suggestion
== Contributors
```

guides/common/modules/ref_katello-4-0-release-notes.adoc
```diff
@@ -0,0 +1,58 @@
+[id="katello-release-notes_{context}"]
+
```
maximiliankolb, line 2:
```suggestion
```

guides/Release_Notes/master.adoc
```diff
@@ -0,0 +1,29 @@
...
+:context: {project-context}
+:ProductName: {ProjectServer}
+
+ifdef::foreman-el[]
+include::common/modules/ref_foreman-3-0-release-notes.adoc[]
+
+## Corresponding Katello release notes
+
+
+include::common/modules/ref_katello-release-notes.adoc[]
```
maximiliankolb, line 19: I don't fully understand the logic behind this. Is it mixed up? Only show katello release notes "ifdef katello"?

maximiliankolb (commented):
Lot of nitpicks on asciidoc headings.

lzap:
So where we are with this one? Looks like we are missing 3.0 with this, so does this mean release notes will be published on the old place? We can still finish and merge this one I think it is worth the effort.

ekohl:
In #699 I started an alternative for this. IMHO the 3.0 release notes should not be added to master since master is now 3.1.

That said, I don't think we're going to make these docs the primary source of documentation for 3.0. There's too much work to do. Right now I'd qualify docs.theforeman.org for vanilla Foreman (so without Katello) as pre-alpha quality.

lzap:
I think we can't go wrong adding release notes for 3.0 into nightly and learn along the way the process, structure and everything. It won't be advertised as the real source, however, it is a good starting point.

Thanks for explanation, after my PTO I am trying to navigate through this storm of PRs which were caused by trying to hit 3.0 deadline I think.

ekohl:
> I think we can't go wrong adding release notes for 3.0 into nightly and learn along the way the process, structure and everything. It won't be advertised as the real source, however, it is a good starting point.

I think we can go wrong that way. Today we only have the release notes for a specific version in that manual. I wasn't thinking about making one changelog guide that contained all changes for all versions. Rather stick only 3.0.x changes in the 3.0 manual. And thus only what will become 3.1.0 changes in current nightly.

> Thanks for explanation, after my PTO I am trying to navigate through this storm of PRs which were caused by trying to hit 3.0 deadline I think.

Correct. I was trying to get it ready in time for 3.0 since that would have been nice moment to align. However, after working on it I've come to the conclusion it's not feasible in time for 3.0 GA. We should have started earlier.

ekohl:
This has been implemented and merged in other PRs so I'm going to close this.

PR #622, opened 2021-07-22
Use {ProjectWebUI} instead of "web UI" or "{Project} web UI"

maximiliankolb:
This PR replaces `web UI` and `{Project} web UI` with `{ProjectWebUI}`.

The first two commits were done using `find` & `sed`; the last commit was done by hand.
We can squash this before merging if you prefer. Splitting the commits should be easier to review.


melcorr (approved):
I really appreciate this @maximiliankolb !!

PR #623, opened 2021-07-22
Rework managing hosts guide

maximiliankolb:
* Remove various unnecessary whitespace
* Set first heading to '=' for each file and adjust the leveloffset
  accordingly
* Use level 2 and higher headings instead of asciidoc titles


melcorr (commented):
I appreciate this, but it looks like there's one heading or so that needs reworking from the build output? 
I always wanted to do this when I was in docs, but never had the time. So I am excited!

maximiliankolb:
> I appreciate this, but it looks like there's one heading or so that needs reworking from the build output?
> I always wanted to do this when I was in docs, but never had the time. So I am excited!

Checks are green, seems to work.

Rebased to master.

melcorr (approved):
Thank you so much for this @maximiliankolb 
I am delighted with this!

PR #624, opened 2021-07-22
ACD: allow Ansible parameters in yaml

maximiliankolb:
ACD supports now adding and editing arbitrary YAML data instead of only `key:value`.


melcorr (approved):
LGTM

PR #627, opened 2021-07-27
Fix various whitespace

maximiliankolb:
This PR fixes various whitespace: it remove unnecessary empty lines, adds lines above and below including other `.adoc` files, and removes one instance of two spaces in inline text.


melcorr (approved):
Hey @maximiliankolb 
LGTM

PR #628, opened 2021-07-27
Extend list of supported compute resources

maximiliankolb:
This PR extends the list of supported compute resource providers and sorts it alphabetically.

### DIFF
--- a/guides/doc-Managing_Hosts/topics/con_overview-of-hosts-in-satellite.adoc
+++ b/guides/doc-Managing_Hosts/topics/con_overview-of-hosts-in-satellite.adoc
@@ -4,12 +4,14 @@
 
 A host is any Linux client that {ProjectName} manages.
 Hosts can be physical or virtual.
-ifndef::orcharhino[]
-Virtual hosts can be deployed on any platform supported by {ProjectName}, such as KVM, VMware vSphere, OpenStack, Amazon EC2, Rackspace Cloud Services or Google Compute Engine.
+ifdef::foreman-el,foreman-deb,katello[]
+Virtual hosts can be deployed on any platform supported by {ProjectName}, such as Amazon EC2, Google Compute Engine, KVM, libvirt, Microsoft Azure, OpenStack, {oVirt}, Proxmox, Rackspace Cloud Services, or VMware vSphere.
 endif::[]
-
+ifdef::satellite[]
+Virtual hosts can be deployed on any platform supported by {ProjectName}, such as Amazon EC2, Google Compute Engine, KVM, libvirt, Microsoft Azure, OpenStack, {oVirt}, Rackspace Cloud Services, or VMware vSphere.
+endif::[]
 ifdef::orcharhino[]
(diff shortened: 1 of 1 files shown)


guides/doc-Managing_Hosts/topics/con_overview-of-hosts-in-satellite.adoc
```diff
@@ -4,7 +4,7 @@
 
 A host is any Linux client that {ProjectName} manages.
 Hosts can be physical or virtual.
-Virtual hosts can be deployed on any platform supported by {ProjectName}, such as KVM, VMware vSphere, OpenStack, Amazon EC2, Rackspace Cloud Services or Google Compute Engine.
+Virtual hosts can be deployed on any platform supported by {ProjectName}, such as Amazon EC2, Google Compute Engine, KVM, libvirt, Microsoft Azure, OpenStack, oVirt, Proxmox, Rackspace Cloud Services, RHV, or VMware vSphere.
```
melcorr, line 7:
```suggestion
Virtual hosts can be deployed on any platform supported by {ProjectName}, such as Amazon EC2, Google Compute Engine, KVM, libvirt, Microsoft Azure, OpenStack, {oVirt}, Proxmox, Rackspace Cloud Services, or VMware vSphere.
```

I'm unsure about Proxmox in Red Hat Satellite @maximiliankolb - let me try and confirm. 
Also, oVirt = RHV https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-satellite.adoc#L53

So if we are to do this, it'll have to be ifdef to remove some of them. Sorry !

melcorr (commented):
Hey @maximiliankolb 
I'll investigate how best to do this.

guides/doc-Managing_Hosts/topics/con_overview-of-hosts-in-satellite.adoc
maximiliankolb: Great. One more hint if you allow: theforeman.org still shows `RHEV` instead of `RHV`. just fyi.

Rebased with proper attribute for `oVirt`. Also, what would happen if community users use both "oVirt" and "RHV"?

melcorr: > Great. One more hint if you allow: theforeman.org still shows `RHEV` instead of `RHV`. just fyi.

Lol. They've changed their branding and I'm unsure which is the most up to date.
> 
> Rebased with proper attribute for `oVirt`. Also, what would happen if community users use both "oVirt" and "RHV"?

That's a good point. In the docs we only have oVirt for the actual procedures. For the actual computer resource docs, we could add in a note there saying oVirt and RHV are the same things for non-Satellite scenarios. 

For this sentence, we should probably just duplicate with an ifdef here for Satellite with the original sentence, and then ifndef Satellite for the rest :-)

maximiliankolb: I think a note somewhere would not hurt.

Regarding this sentence: I am currently working on another PR to add a sentence for `orcharhino` only. I can rebase afterwards of course.

maximiliankolb:
Rebased to master & removed `Proxmox` for Satellite.

melcorr (approved):
This slipped through the cracks. All good?

PR #630, opened 2021-07-27
Remove bash highlighting in code blocks

maximiliankolb:
This PR fixes the source highlighting for code blocks in the 'puppet for configuration management' documentation.


melcorr (approved):
I've no issue with this but what problem was it causing?

maximiliankolb:
This issue comes from `#` in combination with `bash`. This results in grey cursive text because it's treated as a comment.

melcorr:
> This issue comes from `#` in combination with `bash`. This results in grey cursive text because it's treated as a comment.

This is good to know. Thank you for explaining. I found that the "quotes" specification in the asciidoc can highlight parts of the text yellow sometimes...

maximiliankolb:
> > This issue comes from `#` in combination with `bash`. This results in grey cursive text because it's treated as a comment.
> 
> This is good to know. Thank you for explaining. I found that the "quotes" specification in the asciidoc can highlight parts of the text yellow sometimes...

Did not know that! Do you have a general recommendation? Let's discuss this here: https://github.com/theforeman/foreman-documentation/issues/631

PR #635, opened 2021-07-27
BZ1983101 deploying capsule cert is confusing

ryandeussing:
Need to add "-t capsule" if the katello-certs-check is called with custom ssl certificates for Capsule server

Bug 1983101 - Deploying capsule certificates is confusing as per the documentation

https://bugzilla.redhat.com/show_bug.cgi?id=1983101

Cherry-pick into:

* [x] Foreman 2.4

### DIFF
--- a/guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
+++ b/guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
@@ -22,7 +22,7 @@ For more information, see {InstallingSmartProxyDocURL}installing-capsule-server-
 [options="nowrap", subs="+quotes,attributes"]
 ----
 # katello-certs-check \
--c __/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem__ \      <1>
+-t capsule -c __/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem__ \      <1>
 -k __/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert_key.pem__ \  <2>
 -b __/root/{smart-proxy-context}_cert/ca_cert_bundle.pem__      <3>
 ----
--- a/guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
+++ b/guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
@@ -22,7 +22,7 @@ For more information, see {InstallingSmartProxyDocURL}installing-capsule-server-
 [options="nowrap", subs="+quotes,attributes"]
 ----
 # katello-certs-check \
--t {smart-proxy-context} -c __/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem__ \      <1>
+-t {certs-proxy-context} -c __/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem__ \      <1>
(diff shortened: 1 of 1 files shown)


ryandeussing:
Link to PR shared with QE for review.

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
```diff
@@ -22,7 +22,7 @@ For more information, see {InstallingSmartProxyDocURL}installing-capsule-server-
 [options="nowrap", subs="+quotes,attributes"]
 ----
 # katello-certs-check \
--c __/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem__ \      <1>
+-t capsule -c __/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem__ \      <1>
```
melcorr, line 25:
```suggestion
-t {smart-proxy-context} -c __/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem__ \      <1>
```

We don't use capsule in the upstream so we need to add attributes.

melcorr (commented):
Lukas is on PTO.

melcorr:
@ekohl would you mind taking a look?

ekohl (commented):
For reference, this was introduced as part of https://projects.theforeman.org/issues/25564. It's only the help text that matters, but in this case it's correct. https://projects.theforeman.org/issues/30251 then also introduced the option to brand it

ryandeussing:
Thank you @ekohl changes applied

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
```diff
@@ -22,7 +22,7 @@ For more information, see {InstallingSmartProxyDocURL}installing-capsule-server-
 [options="nowrap", subs="+quotes,attributes"]
 ----
 # katello-certs-check \
--c __/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem__ \      <1>
+-t {smart-proxy-context} -c __/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem__ \      <1>
```
ekohl, line 25: In upstream `smart-proxy-context` is `smart-proxy`. I looked at it and the help text suggests (in upstream) it's either `katello` or `foreman-proxy` but the code says that anything other than `katello` will trigger the non-server code. So `smart-proxy` and `foreman-proxy` will be handled the same.

The only thing I can think of is that a user may run `katello-certs-check -h` and be confused.

Turns out there is already a variable for this:
```suggestion
-t {certs-proxy-context} -c __/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem__ \      <1>
```

See https://github.com/theforeman/foreman-documentation/blob/master/guides/common/modules/proc_creating-a-custom-certificate.adoc as well:

> If you set a wildcard value, you must add the `-t {certs-proxy-context}` option when you use the `katello-certs-check` command.

This may be a bad example since that particular line is plain wrong. Given the `ifeval` it should state to **always** pass `-t {certs-proxy-context}` instead of only when a wildcard value is set, just like in this change.

melcorr: Thanks so much, @ekohl

ryandeussing:
Thank you x2 @ekohl changes applied

ekohl (approved):

PR #641, opened 2021-07-30
Mark Foreman 2.4 as unsupported

upadhyeammit:


upadhyeammit:
This should be merged once we GA then Foreman 3.0.

PR #643, opened 2021-08-04
Use installer options to set up DHCP multi homing

ekohl:
The installer can set this up since [Foreman 1.18](https://github.com/theforeman/puppet-foreman_proxy/commit/fcf6d82fde29527bf063d6cb60199339fd4a6266).

Technically it should be safe to cherry pick this into all branches, but I must admit I haven't tested this. That's why I'd only pick it into upstream supported branches, but if the docs team is confident, I won't hold you back :)

Cherry-pick into:

* [x] Foreman 2.4


melcorr (approved):
Just like that! :-) 
We've made such large changes that I agree with your CP selections !

Issue #647, opened 2021-08-04
Fix the "Installing an External Smart Proxy Server on Debian" guide


ekohl:
https://docs.theforeman.org/nightly/Installing_Proxy_on_Debian/index-foreman-deb.html contains many thing that are incorrect. I think some of these also affect Foreman without Katello installations.

* https://docs.theforeman.org/nightly/Installing_Proxy_on_Debian/index-foreman-deb.html#system-requirements_smart-proxy

The users apache, foreman and postgres will not be needed. `apache` doesn't even exist on Debian (it's `www-data`), but no Apache is installed. `foreman` can also be dropped from the RPM-based guides since it doesn't matter. `postgres` is only relevant on content proxies.

* https://docs.theforeman.org/nightly/Installing_Proxy_on_Debian/index-foreman-deb.html#capsule-ports-and-firewalls-requirements_smart-proxy

Port 5646 is irrelevant on non-Katello installs.

* https://docs.theforeman.org/nightly/Installing_Proxy_on_Debian/index-foreman-deb.html#capsule-ports-and-firewalls-requirements_smart-proxy

This whole chapter is irrelevant on non-Katello since port 5646 isn't used.

* https://docs.theforeman.org/nightly/Installing_Proxy_on_Debian/index-foreman-deb.html#enabling-connections-to-capsule_smart-proxy

Ports need updating

* https://docs.theforeman.org/nightly/Installing_Proxy_on_Debian/index-foreman-deb.html#configuring-repositories-proxy-deb_smart-proxy

Always lists `buster` as the distro, but that doesn't work on Ubuntu. I'd recommend using what's on https://docs.theforeman.org/nightly/Quickstart_Guide/index-foreman-deb.html#configuring-repositories_foreman since that also sets up Puppet, which is required.

* https://docs.theforeman.org/nightly/Installing_Proxy_on_Debian/index-foreman-deb.html#installing-the-satellite-capsule-server-packages_smart-proxy

Is wrong, needs to be what's @ https://docs.theforeman.org/nightly/Quickstart_Guide/index-foreman-deb.html#installing-the-satellite-server-packages_foreman

* https://docs.theforeman.org/nightly/Installing_Proxy_on_Debian/index-foreman-deb.html#synchronizing-the-system-clock-with-chronyd_smart-proxy

It's odd to refer to RHEL docs for Debian users.

[...]

PR #649, opened 2021-08-04
Update Debian's repository setup to avoid apt-key

ekohl:
Similar to the change made on deb.theforeman.org's header[1], this
should also avoid using apt-key.

It also removes the separate proxy repository setup since that's wrong.
It hardcoded buster which broke Ubuntu. It also skips over Puppet repo
setup.

[1]: https://github.com/theforeman/foreman-infra/commit/5f5bf8f5f4c9297e2c9639b8ed91f836adcb1dd5

While this applies to all branches, the whole Debian installation needs a lot of work. This works towards fixing https://github.com/theforeman/foreman-documentation/issues/647.

Cherry-pick into:

* [ ] Foreman 2.4


melcorr (approved):
Tested this locally and all looks good.

PR #664, opened 2021-08-16
Set 3.0/4.2 versions

jlsherrill:
Cherry-pick into:

* [ ] Foreman 2.4

### DIFF
--- a/guides/common/attributes.adoc
+++ b/guides/common/attributes.adoc
@@ -3,9 +3,9 @@
 :DocState: nightly
 
 // Version numbers
-:ProjectVersion: nightly
+:ProjectVersion: 3.0
 :ProjectVersionPrevious: 2.5
-:KatelloVersion: nightly
+:KatelloVersion: 4.2
 :TargetVersion: 2.4
 :TargetVersionMaintainUpgrade: 2.4
 // The above attribute should point to the GA version number (x.y) for all releases including Beta


guides/common/attributes.adoc
```diff
@@ -3,9 +3,9 @@
 :DocState: nightly
 
 // Version numbers
-:ProjectVersion: nightly
+:ProjectVersion: 3.0
 :ProjectVersionPrevious: 2.5
-:KatelloVersion: nightly
+:KatelloVersion: 4.2
 :TargetVersion: 2.4
```
ekohl, line 9: This looks horribly out of date. Why do we have both a TargetVersion and a ProjectVersion? Then again, it's not like foreman-maintain actually works in upstream, right?

jlsherrill: right, i don't think those versions are relevant for upstream docs, and probably aren't relevant at all for a releaes that doesn't have a downstream version (like this one).   I'm happy to set these to something like 6.9  or 3.0 if its preferred

ekohl: I'll merge it for now, but IMHO something we should address in master first.

melcorr: Noooo these are for the upgrade guide :-( 
These were necessary

melcorr: OK - good. From the email this wasn't clear. 
We need a target version to reference upgrades in upgrade guides.

jlsherrill: @melcorr hey, to clarify do we need to change anything here in the 3.0 branch?    I wasn't suggesting we get rid of these values, i just didn't think they had any relevance to the  upstream docs for 3.0

ekohl: @melcorr out of interest, why can't we use the ProjectVersion as the TargetVersion? And `2.4` doesn't make any sense, even if it doesn't end up being used, it's just confusing, right?

PR #666, opened 2021-08-17
Update platform links

melcorr:
RHEL Platform documentation went through an upgrade during the weekend, and this upgrade changed several urls. I have been assured that eventually redirects will work over the next week or so, but the linkchecker is causing the links to fail, which is slowing progress. It is less time consuming just to fix and move on. 

Cherry-pick into:

* [x] Foreman 3.0
* [ ] Foreman 2.4


ekohl (approved):
It's really a monumental failure on their side. You really don't want to do that for SEO etc.

That said, shouldn't we pick this to all branches?

melcorr:
> It's really a monumental failure on their side. You really don't want to do that for SEO etc.
> 
> That said, shouldn't we pick this to all branches?

They are promising redirects in a few days for sure. 
But for the supported branches, we definitely should.

melcorr:
Because there's been a lot of work done post-2.5, this is very messy to cherry-pick. 
I'll wait a few days and see do they redirect elsewhere. In the meantime, I can work on 3.0 and also the migration of the manual without things breaking.

PR #670, opened 2021-08-17
Update readme for instructions on previewing the site locally

melcorr:
* Thanks to ekohl

Cherry-pick into:

* [ ] Foreman 3.0
* [ ] Foreman 2.4

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -8,6 +8,14 @@ Install `hugo` utility.
 
 ## Building
 
+While the `make` commands outliined in the [guides readme](https://github.com/theforeman/foreman-documentation/blob/master/guides/README.md) output previous of the guides, for a local preview of the full site, and links, do the following:
+
+```console
+$ cd web
+$ hugo server
+```
+Then open the URL it mentions: (http://localhost:1313/).
+
 Use `make` or `hugo` to build contents from markdown and layouts in `public/` directory.
 
 ## Deployment
--- a/web/README.md
+++ b/web/README.md
@@ -8,7 +8,7 @@ Install `hugo` utility.
 
 ## Building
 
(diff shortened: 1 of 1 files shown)


web/README.md
```diff
@@ -8,6 +8,14 @@ Install `hugo` utility.
 
 ## Building
 
+While the `make` commands outliined in the [guides readme](https://github.com/theforeman/foreman-documentation/blob/master/guides/README.md) output previous of the guides, for a local preview of the full site, and links, do the following:
```
ekohl, line 11:
```suggestion
While the `make` commands outlined in the [guides readme](https://github.com/theforeman/foreman-documentation/blob/master/guides/README.md) output previous of the guides, for a local preview of the full site, and links, do the following:
```

ekohl (commented):
There is a `Makefile` in the `web` directory but I'm not sure how useful it is.

You may also want to mention that in Fedora you need to run `dnf install hugo`.

web/README.md
```diff
@@ -6,6 +6,12 @@ Landing page for [docs.theforeman.org](https://docs.theforeman.org) is available
 
 Install `hugo` utility.
 
+On a Fedora system:
```
ekohl, line 9: How about shortening it a bit:
```suggestion
Install `hugo`, for example on a Fedora system:
```

web/README.md
```diff
@@ -14,6 +20,9 @@ While the `make` commands outlined in the [guides readme](https://github.com/the
 $ cd web
 $ hugo server
 ```
+
+Alternatively, you can use `make server`
```
ekohl, line 24: I realized there is one trick that makes it easier: you can run `make -C web server` from the root. Not sure if we want to document this.

melcorr: I am worried about the LCD of contributors, so would avoid.

ekohl (approved):

melcorr:
Thanks for your suggestions!

Issue #396, opened 2021-02-12
Only generate guides when it makes sense for the given scenario.


ehelms:
For example, do not generate content guides for Foreman, only for Katello and Satellite.

lzap:
Eric, I took the opportunity to turn this ticket into discussion. I edited your post and added a table where we can put our thoughts, Yes meaning it makes sense to build that version. Feel free to correct my thoughts.

The breakup into `foreman-el` and `foreman-deb` made things a little bit worse for some guides, for example in the Provisioning Guide we only have if-statements for "satellite" and "foreman-el". Therefore it would probably make more sense to just pick one of the `foreman-el` and `foreman-deb` and build that. That would result in: `foreman-el`, `katello` and `satellite`. Thoughts? @ehelms @ekohl @melcorr @maximiliankolb

lzap:
Once we land on something here, I will implement the necessary Makefile changes.

maximiliankolb:
> The breakup into `foreman-el` and `foreman-deb` made things a little bit worse for some guides, for example in the Provisioning Guide we only have if-statements for "satellite" and "foreman-el". 

I am not up to date with this approach (is this a suggestion/has this been decided on/already been implemented?), but it makes sense to me.

A follow up should be going through all guides and checking if individual blocks are missing. I would certainly help doing this.

ehelms:
I would agree that some guides, largely those that deal with using Foreman/Katello/Satellite should not be generated per OS but rather once where applicable. For example, there should be a Content Management guide for Katello users. For installation and administration guides, things get trickier and I currently find my leaning to the current approach of doing it per major OS as the commands and locations of things tends to differ. If we look at different versions of an OS, things do get a little bit harry there as sometimes commands are different (yum vs dnf) and sometimes packages are different (scl vs non-scl) and locations on disk differ (/var/lib/pgsql vs /var/opt/rh/rh-postgresql12/var/lib/pgsql). They may vary enough that having dedicated guides for each just feels cleaner than having both every time within the same page.

For installation this would mean doing:

 * Foreman Install on Enterprise Linux (CentOS + RHEL, 7 & 8)
 * Foreman Install on Debian (Debian + Ubuntu)
 * Katello Install on Enterprise Linux (CentOS + RHEL, 7 & 8)
 * Satellite Install on Enterprise Linux (RHEL, 7 & 8)

Or:

 * Foreman Install on Enterprise Linux (CentOS + RHEL 7)
[...]

PR #672, opened 2021-08-17
Adding various bits of feedback given by ekohl

melcorr:
Ewoud made a number of criticisms on 
https://github.com/theforeman/theforeman.org/pull/1879

In this PR, I flag RHEL and Katello only instructions. 
I also remove for the upstream a notice telling everyone to contact RH Support with their logs. 

This is all that I can implement for now. 
I will open Issues for the others.  
All are welcome to contribute some of these changes. 

Cherry-pick into:

* [ ] Foreman 3.0
* [ ] Foreman 2.4

### DIFF
--- a/guides/doc-Administering_Red_Hat_Satellite/topics/Logging_in_to_Red_Hat_Satellite.adoc
+++ b/guides/doc-Administering_Red_Hat_Satellite/topics/Logging_in_to_Red_Hat_Satellite.adoc
@@ -1,6 +1,10 @@
 [[sect-Administering-Installing_the_Katello_Root_CA_Certificate]]
 === Installing the Katello Root CA Certificate
 
+ifndef::satellite[]
+include::snip_katello-only.adoc[]
+endif::[]
+
 The first time you log on to {Project}, you might see a warning informing you that you are using the default self-signed certificate and you might not be able to connect this browser to {Project} until the root CA certificate is installed in the browser.
 Use the following procedure to locate the root CA certificate on {Project} and to install it in your browser.
 
--- a/guides/doc-Administering_Red_Hat_Satellite/topics/Maintenance.adoc
+++ b/guides/doc-Administering_Red_Hat_Satellite/topics/Maintenance.adoc
@@ -74,6 +74,11 @@ For example, to schedule the cron job to run every day at 15:00, enter the follo
 [[sect-Administering-Recovering_from_a_Full_Disk]]
 === Recovering from a Full Disk
 
+ifndef::satellite[]
+include::snip_katello-only.adoc[]
+endif::[]
+
+
 The following procedure describes how to resolve the situation when a logical volume (LV) with the Pulp database on it has no free space.
 
 [[proc-Administering-to_Recover_from_a_Full_Disk]]
--- a/guides/doc-Administering_Red_Hat_Satellite/topics/Renaming_a_Server.adoc
+++ b/guides/doc-Administering_Red_Hat_Satellite/topics/Renaming_a_Server.adoc
@@ -1,6 +1,10 @@
 [[sect-Administering-Renaming_a_Server]]
 == Renaming {ProjectServer} or {SmartProxyServer}
 
+ifndef::satellite[]
+include::snip_katello-only.adoc[]
+endif::[]
+
 To rename {ProjectServer} or {SmartProxyServer}, you must use the `{project-change-hostname}` script.
 
 If you rename {ProjectServer}, you must reregister all {Project} clients and configure each {SmartProxyServer} to point them to the new {Project} host name.
--- a/guides/doc-Administering_Red_Hat_Satellite/topics/Security_Compliance_Management.adoc
+++ b/guides/doc-Administering_Red_Hat_Satellite/topics/Security_Compliance_Management.adoc
@@ -5,6 +5,11 @@ Security compliance management is the ongoing process of defining security polic
 Any non-compliance is managed according to the organization's configuration management policies.
 Security policies range in scope from host-specific to industry-wide, therefore, flexibility in their definition is required.
 
+ifndef::satellite[]
+Note that the following chapter refers only to RPM-based systems.
+We welcome any and all equivalent documentation for Debian users.
+endif::[]
+
 [[sect-Administering-OpenSCAP_Concepts]]
 
 include::security-content-automation-protocol.adoc[leveloffset=+2]
--- a/guides/doc-Administering_Red_Hat_Satellite/topics/Starting_and_Stopping_Satellite.adoc
+++ b/guides/doc-Administering_Red_Hat_Satellite/topics/Starting_and_Stopping_Satellite.adoc
@@ -1,6 +1,10 @@
 [[chap-Administering-Starting_and_Stopping_Red_Hat_Satellite]]
 == Starting and Stopping {ProjectName}
 
+ifndef::satellite[]
+Note that the following procedure uses {foreman-maintain}, which is not available for Debian users.
+endif::[]
+
 {Project} provides the `{foreman-maintain} service` command to manage {Project} services from the command line.
 This is useful when creating a backup of {Project}.
 For more information on creating backups, see xref:backing-up-satellite-server-and-capsule-server[].
--- a/guides/doc-Administering_Red_Hat_Satellite/topics/log-files-provided-by-satellite.adoc
+++ b/guides/doc-Administering_Red_Hat_Satellite/topics/log-files-provided-by-satellite.adoc
@@ -4,6 +4,11 @@
 
 {ProjectName} provides system information in the form of notifications and log files.
 
+ifndef::satellite[]
+Note that depending on your specific configuration, your log directories vary.
+For example, if you do not use the Katello plugin, some of the entries in this table are not present in your deployment. 
+endif::[]
+
 [[tabl-Administering-Logging_and_Reporting-Log_Files_for_Reporting_and_Troubleshooting]]
 
 .Log File Directories for Reporting and Troubleshooting
--- a/guides/doc-Administering_Red_Hat_Satellite/topics/snip_katello-only.adoc
+++ b/guides/doc-Administering_Red_Hat_Satellite/topics/snip_katello-only.adoc
@@ -1,0 +1,2 @@
+[NOTE]
+This section is only for Katello plugin users.
--- a/guides/common/assembly_migrating-from-internal-databases-to-external-databases.adoc
+++ b/guides/common/assembly_migrating-from-internal-databases-to-external-databases.adoc
@@ -4,7 +4,7 @@ ifdef::context[:parent-context: {context}]
 = Migrating from Internal {Project} Databases to External Databases
 
 ifdef::foreman-el,katello[]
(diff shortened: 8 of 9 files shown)


guides/doc-Administering_Red_Hat_Satellite/topics/Logging_in_to_Red_Hat_Satellite.adoc
```diff
@@ -1,6 +1,10 @@
 [[sect-Administering-Installing_the_Katello_Root_CA_Certificate]]
 === Installing the Katello Root CA Certificate
 
+ifndef::satellite[]
+include::snip_katello-only.adoc[]
+endif::[]
```
ekohl, line 6: The reason this section is Katello-only is that `https://_{foreman-example-com}_/pub` does not exist on non-Katello.

The manual import that is below actually does work for non-Katello but you need to modify the path a bit. I think what you can do is get the file that's mentioned in `/etc/httpd/conf.d/05-foreman-ssl.conf` (`apache2` instead of `httpd` on Debian) as `SSLCertificateChainFile` and import that.

If anything, you can say `/etc/puppetlabs/puppet/ssl/certs/ca.pem` for non-Katello (which is the default) can be used.

There is actually a PR that was recently merged (https://github.com/theforeman/foreman/commit/0d08d7a6ffeb267e4c38755bb68a48ae3f57da01) which at least introduces a setting. With that we can define an endpoint to introduce common functionality.

guides/doc-Administering_Red_Hat_Satellite/topics/Maintenance.adoc
```diff
@@ -74,6 +74,11 @@ For example, to schedule the cron job to run every day at 15:00, enter the follo
 [[sect-Administering-Recovering_from_a_Full_Disk]]
 === Recovering from a Full Disk
 
+ifndef::satellite[]
```
ekohl, line 77: Can't we hide this on non-Katello builds? I think it's terrible UX to have to skip over content and we have the ability to generate multiple documents.

Also, isn't there a Katello tag? That's probably nicer for Orcharhino as well.

guides/doc-Administering_Red_Hat_Satellite/topics/Renaming_a_Server.adoc
```diff
@@ -1,6 +1,10 @@
 [[sect-Administering-Renaming_a_Server]]
 == Renaming {ProjectServer} or {SmartProxyServer}
 
+ifndef::satellite[]
+include::snip_katello-only.adoc[]
+endif::[]
```
ekohl, line 6: Can't we only include the whole section if Katello is defined?

guides/doc-Administering_Red_Hat_Satellite/topics/Security_Compliance_Management.adoc
```diff
@@ -5,6 +5,11 @@ Security compliance management is the ongoing process of defining security polic
 Any non-compliance is managed according to the organization's configuration management policies.
 Security policies range in scope from host-specific to industry-wide, therefore, flexibility in their definition is required.
 
+ifndef::satellite[]
+Note that the following chapter refers only to RPM-based systems.
+We welcome any and all equivalent documentation for Debian users.
+endif::[]
+
 [[sect-Administering-OpenSCAP_Concepts]]
```
ekohl, line 13: Doesn't this work? It probably needs to be done a layer above, but you get the idea.
```suggestion
ifndef::foreman-deb[]
[[sect-Administering-OpenSCAP_Concepts]]
endif::[]
```

guides/doc-Administering_Red_Hat_Satellite/topics/Security_Compliance_Management.adoc
```diff
@@ -5,6 +5,11 @@ Security compliance management is the ongoing process of defining security polic
 Any non-compliance is managed according to the organization's configuration management policies.
 Security policies range in scope from host-specific to industry-wide, therefore, flexibility in their definition is required.
 
+ifndef::satellite[]
+Note that the following chapter refers only to RPM-based systems.
+We welcome any and all equivalent documentation for Debian users.
```
ekohl, line 10: That can't work since we haven't packaged the plugins.

guides/doc-Administering_Red_Hat_Satellite/topics/Starting_and_Stopping_Satellite.adoc
```diff
@@ -1,6 +1,10 @@
 [[chap-Administering-Starting_and_Stopping_Red_Hat_Satellite]]
 == Starting and Stopping {ProjectName}
 
+ifndef::satellite[]
```
ekohl, line 4: What about Orcharhino?

guides/doc-Administering_Red_Hat_Satellite/topics/log-files-provided-by-satellite.adoc
```diff
@@ -4,6 +4,11 @@
 
 {ProjectName} provides system information in the form of notifications and log files.
 
+ifndef::satellite[]
+Note that depending on your specific configuration, your log directories vary.
+For example, if you do not use the Katello plugin, some of the entries in this table are not present in your deployment. 
+endif::[]
```
ekohl, line 10: I'd prefer to:
* hide `candlepin` and `tomcat` on non-Katello
* `/var/log/rhsm` should be hidden on `foreman-deb`
* `/var/log/httpd` should be `/var/log/apach2` on `foreman-deb`
* `/var/log/messages` should be `/var/log/daemon.log` on `foreman-deb`.

ekohl (changes requested):

melcorr:
I have no idea if we want to build out a separate Katello admin guide or not. 
I remember there was some discussion here: https://github.com/theforeman/foreman-documentation/issues/396

We can just continue that conversation. 
I am closing this PR as I don't think anything in this PR will satisfy anyone.

guides/doc-Administering_Red_Hat_Satellite/topics/Logging_in_to_Red_Hat_Satellite.adoc
lzap: Can we break up things that Mel cannot do into separate issues so engineers can do these changes? This smell like something that is a bit more complex. Then we could merge this PR which contains bunch of meaningful changes and improve the guide.

guides/doc-Administering_Red_Hat_Satellite/topics/Maintenance.adoc
lzap: We discussed this in the past and we thought that a note is better than hiding whole chapters because it presents the users whats possible if they had installed Katello. At least this is what we were doing everywhere, if we want to change the strategy then we should probably do it everywhere.

I kinda lean towards showing it and advocating Katello deployments.

guides/doc-Administering_Red_Hat_Satellite/topics/snip_katello-only.adoc
```diff
@@ -0,0 +1,2 @@
+[NOTE]
+This section is only for Katello plugin users.
```
lzap, line 2: I'd swear we already have something like that. Interesting.

guides/doc-Administering_Red_Hat_Satellite/topics/Renaming_a_Server.adoc
lzap: Ditto.

guides/doc-Administering_Red_Hat_Satellite/topics/snip_katello-only.adoc
melcorr: We had for the provisioning guide but then that was changed by work done to use one set of prereqs for all Compute Resources!

guides/doc-Administering_Red_Hat_Satellite/topics/log-files-provided-by-satellite.adoc
lzap: We should extract this into a separate PR done by an engineer as well.

guides/doc-Administering_Red_Hat_Satellite/topics/Starting_and_Stopping_Satellite.adoc
lzap: What do you mean exactly?

guides/doc-Administering_Red_Hat_Satellite/topics/Security_Compliance_Management.adoc
lzap: We need to drop a note or hide this yeah. This won't work at the moment and unfortunately we need to hide.

This is a typical case when hiding whole chapters is not the best approach, see, a technically skilled user could see what Foreman offers here and it's not that difficult for someone with experience to deploy this on Debian even without packages available. This is why I prefer to drop a note and still show all the text.

guides/doc-Administering_Red_Hat_Satellite/topics/Starting_and_Stopping_Satellite.adoc
melcorr: He meant that I hadn't taken them into consideration - whether they would use this or not. 
My plan was that I was going to ask @maximiliankolb  for a review to ensure I didn't cut them out. 
This PR is dead, and won't be resurrected, but appreciate your feedback!

lzap:
So, to fork or not to fork.

I think we should try to not have a separate guide until it's reasonable. Meaning, unless there are so many quirks and notes in the guide for other builds, we should keep trying. This one is on the edge I think, initially I thought we would only maintain separate guides for installations.

I believe this PR is worth reopening and finishing, contains bunch of good changes, I feel like Mel should not be the one writing more advanced technical changes without good context and I see two of the hunks here that should be extracted and worked separately.

There is the question on how to approach content that is irrelevant. We have several options:

* Hide whole content or even chapter.
* Provide a note that the content is not relevant and keep the content.
* Provide a note that the content is not relevant and hide the content.
* Provide a note that the content is not relevant and hide the content and make a link to katello guide that has it.

I think we do not have a hard rule and our docs is a mix and perhaps we should keep it like that. Sometimes it is good to show the content, sometimes it is just completely irrelevant and hiding is better. I think we should work on case to case basis, what you think?

ekohl:
I think it's unreasonable to have a single guide. If you insist that we go this route, we might as well stop with docs.theforeman.org as an effort since I think it'll never work until Katello truly is a Foreman plugin that can be added after the fact. I'm guessing that's at least a year in the future but probably more.

I can't emphasize enough how much I think that's a bad idea. IMHO the instructions are really garbage for vanilla Foreman users. I know that's strong words, but that is not an understatement. Every single guide I read through needs entire chapters ripped out or rewritten.

lzap:
Assume good intentions, your words indeed are strong. What we are doing here is with minimum resources and we are far away from being done. But I think we share the same goal, its just a detail on how are we going to maintain the guts.

I am not against forking this guide, I think we should be able to find a good middleground: we need to include common chapters via `include` asciidoc feature from the day one so maintenance is easier (no copy-pasting).

PR #673, opened 2021-08-18
Install foreman-console instead of all plugins

ekohl:
This removes the extremely dangerous instruction to install all plugins. Users who do this can easily effectively break their installation. We have no way to safely remove plugins and there is no way back.

The correct thing to do is ensure the foreman-console package is installed. That's the only thing that is needed.

This also has the benefit that it works on all supported platforms rather than just EL7.

I consider this very important and would suggest to fix it where possible.

Cherry-pick into:

* [x] Foreman 3.0
* [x] Foreman 2.4


maximiliankolb (approved):

melcorr:
Cherry picked eeeeeeeeeeeverywhere

PR #677, opened 2021-08-19
Replace recursive make

ekohl:
[Recursive make is considered harmful](https://web.archive.org/web/20150330111905/http://miller.emu.id.au/pmiller/books/rmch/). This declares exact rules which means a single Makefile is sufficient.

Right now this is incomplete and there's a weird issue that it doesn't create all builds at once. For some reason, I need to run `make` multiple times. Maybe @lzap has an idea.

### DIFF
--- a/guides/Makefile
+++ b/guides/Makefile
@@ -1,37 +1,51 @@
 SHELL := /bin/bash
-ALL_DIRS = $(shell ls -d doc-*)
-GUIDES-HTML = $(ALL_DIRS)
+ALL_DIRS = $(wildcard doc-*)
+ALL_GUIDES = $(patsubst doc-%,%,$(ALL_DIRS))
+BUILDS=foreman-el foreman-deb katello
+BUILD_DIR := build
+COMMON_DIR := common
+DEPENDENCIES := $(shell find $(COMMON_DIR) -type f -name \*.adoc)
+DATE_MDY := $(shell LC_ALL=C date "+%b %d %Y")
+DATE_MY := $(shell LC_ALL=C date "+%b %Y")
 
-.PHONY: all html pdf linkchecker all-html all-pdf clean all-clean linkchecker-tryer $(GUIDES-HTML)
+HTML_INDEXES := $(BUILDS:%=index-%.html)
+PDFS := $(BUILDS:%=index-%.pdf)
+GUIDE_BUILDS := $(addprefix $(BUILD_DIR)/,$(ALL_GUIDES))
+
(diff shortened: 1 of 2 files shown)


guides/Makefile
```diff
@@ -1,37 +1,51 @@
...
 
-html: all-html
+html: $(GUIDES-HTML)
 
-pdf: all-pdf
+pdf: $(GUIDES-PDF)
 
-clean: all-clean
+clean:
+	rm -rf build
```
lzap, line 27:
```suggestion
	rm -rf $(BUILD_DIR)
```

ekohl: I swear I had that somewhere. So many branches.

guides/Makefile
```diff
@@ -1,37 +1,89 @@
...
-pdf: all-pdf
+pdf: $(GUIDES-PDF)
 
-clean: all-clean
+clean:
+	rm -rf $(BUILD_DIR)
 
 linkchecker:
-	find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern
+	find $(BUILD_DIR) -type f -name \*.html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern
```
ekohl, line 30: Note I changed `index-$build` to `$build`. However, I haven't modified the web parts yet.

lzap:
The paper presents the following problems which I marked with (X) as irrelevant to this project.

* Order of modules. (X)
* Multiple passes needed. (X)
* Slow build time.
* Too much information per module. (X)
* Incapability to build cleanly. (X)
* Not possible to parallelize.

Now, I am not against removing recursive make, but your proposal copies the actual build command (asciidoc) from the guide directory in here. What is the plan? Do you plan to end up with just a single makefile in the root folder deleting all the others (symlinks and the common/Makefile)?

What I very much like in my workflow is ability to build just a single guide by issuing `make` command in a directory with a guide. How that is supposed to work?

ekohl:
> What I very much like in my workflow is ability to build just a single guide by issuing `make` command in a directory with a guide. How that is supposed to work?

That's indeed why I haven't removed the common Makefiles. I wasn't sure how to deal with that.

That said, my main goal was to build all guides by default every time. https://github.com/theforeman/foreman-documentation/pull/678 would be the reason for that. That can also be done with a common makefile setup though.

If anything, with this exercise I learned a lot more about the build process.

lzap:
I am not against this, if you want to carry on just do it. I would like to be able to compile just a single guide when in `guides/xyz` directory tho.

ekohl:
I agree with your points, but right now I don't have time to finish this.

ekohl:
Closing because I don't have time to finish this now.

PR #678, opened 2021-08-19
Split documentation into multiple guides

ekohl:
There really are 3 different flavors of guides:

* Foreman on RHEL/CentOS
* Foreman on Debian/Ubuntu
* Katello on RHEL/CentOS

This takes that to the documentation listing.

This is web so no need to cherry pick.


melcorr:
I was hoping that we would get someone to make a page with a better design. 
I think this is premature as we haven't reviewed most of the guides for Debian.

ekohl:
> I was hoping that we would get someone to make a page with a better design.

Perhaps this should be considered an intermediate step.

>  I think this is premature as we haven't reviewed most of the guides for Debian.

I kind of agree, but that was on my list. Perhaps we can even use it to clearly say "this is reviewed, this is WIP".

melcorr:
IMO, this just adds more mess, more things to click on that aren't yet right. Exposing more unreviewed content than before.  
I've no problem with this in the near future, when more work is done, but I am against merging this now.

ekohl:
So let's start with that. Please have a look at https://github.com/theforeman/foreman-documentation/pull/679. I'm still a bit new to the whole documentation tech stack so I'd like some guidance.

ekohl:
Updated. I now only modify nightly docs. The commits to remove broken guides link to issues/PRs where they exist.

I wonder, should we merge the the guides `doc-Installing_Proxy_on_Debian` and `doc-Installing_Proxy_on_Red_Hat` into a single `Installing Proxy` directory? The same for the server bits. We generate multiple flavors anyway.

maximiliankolb:
I did not come up with something better, so maybe `Day 1` and `Day 2`? I also don't want to stall this PR.

ekohl:
I looked at https://docs.orcharhino.com and I'm wondering if we should use your example: `Installation guides` and `Usage guides`. While not entirely correct, I feel like it's closer than what's there now.

While discussing this with @ehelms he also mentioned that at some point we'll have plugins. Today we have ACD and Ansible as "plugin guides" We need to consider how that'll take shape. Perhaps not today, but it'll come up at some point.

maximiliankolb:
> I looked at https://docs.orcharhino.com and I'm wondering if we should use your example: `Installation guides` and `Usage guides`. While not entirely correct, I feel like it's closer than what's there now.
> 
> While discussing this with @ehelms he also mentioned that at some point we'll have plugins. Today we have ACD and Ansible as "plugin guides" We need to consider how that'll take shape. Perhaps not today, but it'll come up at some point.

Thanks for looking at our downstream product, but spoiler alert: we are in the process of restructuring orcharhino documentation. We face the same issue of users not immediately finding what they're looking for. Docs have expanded a lot lately, so reorganizing them becomes necessary. I will keep you posted.

lzap:
> some point we'll have plugins. Today we have ACD and Ansible as "plugin guides"

Overall, we should push against this and create rather consistent experience for users who are not aware of plugins. ACD is special case because what it provides is rather unique. But I can imagine Ansible, Puppet and Chef being in a single guide named Configuration Management Guide.

ekohl:
I wasn't saying we need exactly one guide per plugin, but at some point plugins will need documentation and we need a plan to deal with that.

We'll probably need some guide on writing documentation. While reflecting on it, I also came to the conclusion that there's more. For example, we can make a list of katello-isms. It can be as simple as a list with items like "if you see `/root/ssl-build`, it's Katello-only". I recently also make a list of the biggest differences between Debian and RH (httpd vs apache2, ...).

lzap:
I suggest to mark the suggested guides via `WIP!` ribbon with link and further explanation. I updated my PR: https://github.com/theforeman/foreman-documentation/pull/710

ekohl:
I split off commits into separate PRs:
* https://github.com/theforeman/foreman-documentation/pull/921
* https://github.com/theforeman/foreman-documentation/pull/922

As for the index: I think it still needs work but given the effort of #857 I'm closing this now.

PR #679, opened 2021-08-20
Planning guide: Hide Katello parts in non-Katello

ekohl:
This works towards making the planning guide available on vanilla Foreman, both EL and Debian/Ubuntu. It's still heavy work in progress and I've only managed to go through the first 4 chapters.

It first has one commit to fix the description of a Smart Proxy. I think this makes sense in itself and I intend to submit it as a separate PR. It can also spark some discussion.

I do think System Architecture and System Components can make sense in a vanilla Foreman setup, but they require a complete rewrite. I do wonder how to best deal with it. Do we split out those chapters to separate files and include the correct one?

So please have a look at the rough changes and comment on the direction. It's written with https://github.com/theforeman/foreman-documentation/pull/678 in mind.

Issue #676, opened 2021-08-18
Adapt Administering Foreman guide for different scenarios


melcorr:
As part of #396, we and the latest discussion there, we need to stream the admin guide to generate versions for 

* Satellite (the status quo but ensure nothing leaks that shouldn't etc)
* Foreman-EL
* Katello
* Debian users
* (I know orcharhino work has already been done, maybe more is needed)

This was inspired by https://github.com/theforeman/theforeman.org/pull/1879#discussion_r690316318

I (@melcorr) will work on the foreman-el and katello scenarios.

melcorr:
@ekohl can you help me formulate a prerequisite for chapter two for Foreman EL and Katello users. 

To use this section, you must have foreman-maintain installed (how do they do that?) 
And is there an alternative method? I can add that in? 
At the moment, i've just added an ifndef for foreman-deb builds for Chapter 2.

ekohl:
> To use this section, you must have foreman-maintain installed (how do they do that?)

For Foreman on EL, you can `yum/dnf install` it. For Debian you can't. The software doesn't support it and is also not packaged.

Personally I always use `systemctl` directly.

`systemctl stop foreman.service foreman.socket` to stop Foreman itself. For something like (re)start it's sufficient to only restart `foreman.service`. There's also `dynflow-sidekiq@*` (always recommend to use `\*` if it's a shell command to prevent globbing files).

You should also be aware that `httpd` (EL) or `apache2` (Debian) is a service that's used on the Foreman server. There's also `rh-redis5-redis` (EL7) or `redis` (EL8 / Debian).

If you're using a DB on the same server, there's also `postgresql.service`.

If I want to manage the Foreman Proxy, it's `systemctl $action foreman-proxy.service`. `$action` can be `status`, `stop`, `start` and `restart`.

maximiliankolb:
orcharhino is currently using a version of "Administering Foreman" somewhere between foreman-el and satellite. For us, Katello is a non-optional plugin and orcharhino only runs on EL (7) currently.

Please tag me @melcorr  whenever there are changes regarding the versions for foreman-el, satellite, and orcharhino.

melcorr:
> There's also `dynflow-sidekiq@*` (always recommend to use `\*` if it's a shell command to prevent globbing files).

[...]

PR #680, opened 2021-08-20
Define katello for satellite and orcharhino

ekohl:
This allows ifdef::katello for Katello-plugin-only installs while still including those in Satellite and Orcharhino builds. It also allows ifndef::katello for vanilla Foreman installations.

Right now this PR is incomplete, but I want to submit it for consideration just as an idea (rather than the exact implementation). What do you think?

### DIFF
--- a/guides/common/assembly_configuring-an-alternate-cname.adoc
+++ b/guides/common/assembly_configuring-an-alternate-cname.adoc
@@ -7,7 +7,7 @@ You can configure an alternate CNAME for {Project}.
 This might be useful if you want to deploy the {Project} web interface on a different domain name than the one that is used by client systems to connect to {Project}.
 You must plan the alternate CNAME configuration in advance prior to installing {SmartProxies} and registering hosts to {Project} to avoid redeploying new certificates to hosts.
 
-ifndef::satellite[]
+ifndef::katello[]
 This procedure is only for Katello plug-in users.
 endif::[]
 
--- a/guides/common/assembly_configuring-satellite-custom-server-certificate.adoc
+++ b/guides/common/assembly_configuring-satellite-custom-server-certificate.adoc
@@ -6,7 +6,7 @@ ifdef::context[:parent-context: {context}]
 
 = Configuring {ProjectServer} with a Custom SSL Certificate
 
-ifndef::satellite[]
+ifndef::katello[]
 This procedure is only for Katello plug-in users.
 endif::[]
 
--- a/guides/common/assembly_installing-capsule-server.adoc
+++ b/guides/common/assembly_installing-capsule-server.adoc
@@ -7,7 +7,7 @@ ifdef::context[:parent-context: {context}]
 Before you install {SmartProxyServer}, you must ensure that your environment meets the requirements for installation.
 For more information, see {InstallingSmartProxyDocURL}preparing-environment-for-capsule-installation[Preparing your Environment for Installation].
 
-ifdef::satellite,katello[]
+ifdef::katello[]
 // Registering {SmartProxyServer} to {ProjectServer}
 include::modules/proc_registering-capsule-to-satellite-server.adoc[leveloffset=+1]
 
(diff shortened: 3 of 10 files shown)


ehelms:
If the intent is to try to say Katello is a subset of Satellite, I think that is an incorrect approach as its not strictly true and could lead to confusing situations. I don't entirely grasp what the problem being encountered is so I can only try to give that thought for now.

ekohl:
> If the intent is to try to say Katello is a subset of Satellite, I think that is an incorrect approach as its not strictly true and could lead to confusing situations. I don't entirely grasp what the problem being encountered is so I can only try to give that thought for now.

I think that where it isn't true, we should have additional `ifdef::satellite[]` or `ifndef::satellite[]`. More of an opt-out than opt-in. The problem I encountered is that in https://github.com/theforeman/foreman-documentation/pull/679 I end up with a **lot** of `ifdef::katello`.

melcorr (commented):
I understand what you're trying to do. 
I've been sitting here for a while going around in circles with this. 
I'm trying to think from all perspectives; katello ifdefs are frequent at the moment. As we expand the docset `ifndefs` for satellite will be as prevalent if we introduce this, and definitely if ATIX add their Katello content here, which I would love.
If I weigh the consequences of this, we come out with an equal amount of conditionalising in the long term, just it's `satellite` rather than `katello`. 
I do think that the change makes it more confusing than the declaration of the relevant scenarios in the ifdef. When I say this, I am thinking about the level of contributions we've had from people less familiar with Foreman over the past few months who seem to genuinely struggle. We've new people joining at some point in the near future, allegedly. 
I also think that there is some merit to considering all our potential scenarios in a conscious way when declaring the ifdef. I know that is not how engineers view the world, but there's a merit in being forced to think about who precisely will be reading it. 
All that said, I am not against this. 
I'd really like to hear opinions from @lzap  and @maximiliankolb and we'll make a decision. 
I hope that's OK. 
PS: If we do merge this, we'll need to add this to the readme.

melcorr:
Looking at the errors in the build, this also breaks our linking conventions?

melcorr:
In light of your comment on #676 @maximiliankolb and since this breaks how we build our links, I'm leaning towards thinking this might bring about more problems than it solves.

maximiliankolb (approved):
LGTM.

maximiliankolb:
> In light of your comment on #676 @maximiliankolb and since this breaks how we build our links, I'm leaning towards thinking this might bring about more problems than it solves.

Your call. For us, the only change would be to add `:katello:` to each document using upstream files, so that's ok for me. As both orcharhino and Satellite uses Katello, I think it's fairly obvious where I would need to make adjustments/ add `,orcharhino` to `ifdef::satellite[]`.

(I've reviewed this before I saw your comment; the changes look good to me.)

melcorr:
Will it be easy to get around the link building problems that are now occurring because of this change?

lzap:
Let me recap what we have today. It all started when we created two build flags: `foreman` and `satellite`. In the beginning, everything was very clear - it is either foreman-core or satellite/katello. As we were adding more and more content and Katello project started using this repo for upstream docs, we have ended up with the following build attributes:

* foreman
* foreman-el
* foreman-deb
* katello
* satellite
* orcharhino

These are disjoint groups, except two cases (let me use `<` for "includes")

* foreman-el < foreman
* foreman-deb < foreman

This only exists so its easier to type `ifdef::foreman[]` instead of `ifdef::foreman-el,foreman-deb[]`. Ewoud suggest to create a new one:

* satellite < katello

I am trying to imagine what happens when we get to the point when Katello becomes really a plugin that can be installed on top of Foreman. We cannot do this anymore:

* satellite < katello < foreman

Why exactly you need this when you can use either `ifdef::katello,satellite[]` or one or the other?

I am not entirely sure what are the differences between satellite and katello deployments, to me it's the same thing except branding. But if we need to differentiate between them, we should totally use just `katello` and `satellite` attributes or both.

In the PR you refer to, I see nesting of ifdefs. I highly disencourage from this, this can lead into issues, we should treat them as disjoint attributes (foreman is special, we could also create `katello_or_satellite` one that would serve the same purpose). Specifically, `katello` cannot be nested into `satellite` and vice versa. This approach sometimes lead to repeating a sentence or a paragraph, but at least we are not copying whole guides. I think this is worth it.

So, I do not understand why would we need to merge this proposal. To me, making attributes including other attributes is dangerous.

ekohl:
> In the PR you refer to, I see nesting of ifdefs. I highly disencourage from this

I did not intend to nest. I intended "this should be shown if katello is present". The further nesting is then if there's anything specific for Satellite, like mention of Red Hat support vs community support within the chapter.

What you didn't mention is that orcharhino also includes katello. So in most cases we'll write `ifdef::katello,satellite,orcharhino[]`. That is very long (and easy to forget).

So in short: I want some definition when the katello plugin is present. This PR's goal is to make that just `katello`.

If we have a specific need for upstream only documentation (that shouldn't make it into satellite nor orcharhino) we can either use `ifndef::satellite,orcharhino[]` or introduce a new `katello-upstream` tag. However, I think the former is good enough.

PR #681, opened 2021-08-20
Correct Home link in navigation

ekohl:
The Home button should behave the same in small as in wide views.

### DIFF
--- a/web/layouts/partials/header.html
+++ b/web/layouts/partials/header.html
@@ -3,7 +3,7 @@
       <nav class="navbar navbar-default navbar-static-top">
         <ul class="nav navbar-nav nav-justified">
           <li class="visible-xs">
-            <a class="index-link" href="https://theforeman.org/">Home</a>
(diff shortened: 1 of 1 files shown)


melcorr (commented):
While I've no problem with this, and will merge, what do you mean by small and wide views?

ekohl:
It's hard because I can't link to exact sections, but it has to do with responsive views that depends on the browser width. You have:
```html
          <li class="visible-xs">
            <a class="index-link" href="https://theforeman.org/">Home</a>
          </li>
```

This is the block that's only visible on eXtra Small views. Then there's this:
```
          <li class="hidden-xs">
            <a class="index-link" href="/"><i class="logo"><img style="width: 90px" src=
            "/index_assets/foremanhelmet.svg" alt="Foreman home" /></i></a>
          </li>
```

That's visible when it's not eXtra Small.

Both blocks are essentially the same, just for a different browser width. They should also link to the same target.

melcorr (approved):

PR #705, opened 2021-09-01
Add checkbox to select all hosts when deleting an app instance

maximiliankolb:
This PR documents a new checkbox to select/deselect all hosts when deleting an application instance in ACD.


lzap:
Thanks.

PR #687, opened 2021-08-23
Modified procedures for Ansible

ianf77:
Modifed procedures and added an additional procedure due to changes in Ansible

[doc] Improve Ansible integration to ease a Puppet replacement

https://issues.redhat.com/browse/SAT-3104

Cherry-pick into:

* [x] Foreman 3.0
* [x] Foreman 2.5 (Satellite 6.10)
* [ ] Foreman 2.4
* [ ] Foreman 2.3 (Satellite 6.9)
* [ ] Foreman 2.1 (Satellite 6.8)

PR #709, opened 2021-09-03
several smaller improvements to the debian docs

evgeni:
Cherry-pick into:

* [ ] Foreman 3.0
* [ ] Foreman 2.4

### DIFF
--- a/guides/common/modules/proc_configuring-repositories-deb.adoc
+++ b/guides/common/modules/proc_configuring-repositories-deb.adoc
@@ -11,7 +11,7 @@
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-cd /tmp && wget https://apt.puppet.com/puppet6-release-{distribution-codename}.deb
+# cd /tmp && wget https://apt.puppet.com/puppet6-release-{distribution-codename}.deb
 ----
 
 . Install the `puppet6-release-{distribution-codename}.deb` package:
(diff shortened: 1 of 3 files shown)


evgeni:
the linkchecker failed, on an unrelated file, and I actually think it's redhat.com that's broken.

@melcorr somehow https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/upgrading_from_rhel_7_to_rhel_8/preparing-a-rhel-7-system-for-the-upgrade_upgrading-from-rhel-7-to-rhel-8 became https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/upgrading_from_rhel_7_to_rhel_8/assembly_preparing-for-the-upgrade_upgrading-from-rhel-7-to-rhel-8 (not the *assembly* in front of the filename) and that looks wrong to me?

lzap (approved):
LGTM can you simply fix the link in our docs? This happens all the time.

evgeni:
> can you simply fix the link in our docs? This happens all the time.

technically? sure. I just think that the link should not have had moved in the first place, but don't know who to ask besides Mel.

maximiliankolb (approved):
LGTM.

lzap:
> I just think that the link should not have had moved in the first place

Yup, it happens all the time. I tried to file a ticket but IIRC it is still opened, we just need to catch up with them that's all we can do at this point.

Feel free to file a BZ, does not look right indeed but we need to get tests green sooner :-)

ekohl (approved):
I didn't know `apt-get` could install local files.

melcorr:
Hey @evgeni 
A small bit of background....
I could merge this, but it wouldn't build because of the broken link. 
In pr #666 recently I had to update a lot of links broken by the RHEL platform group as part of a Drupal upgrade otherwise we couldn't get our docs to build.  
I have written to docs management, like the upper echelons of their group to request redirects related to #666 as that isn't cherry-picked, but redirects haven't come in the last two weeks at least. 
Lzap tried to address this before with the Red Hat Insights team but got no where either.

I know it's not your responsibility, but I also wonder whose responsibility is it :o) ? 
If you wouldn't mind fixing it, I would be extremely grateful.

ekohl:
@ianf77 did make a change in that area in https://github.com/theforeman/foreman-documentation/pull/687/files#diff-f89681be2802614f1f40b5cd2842f009e1d227fdeeb6bf536a324bd7f3eb1640 so either we should cherry pick that and merge it to unblock or merge his entire PR.

evgeni:
> I know it's not your responsibility, but I also wonder whose responsibility is it :o) ?
> If you wouldn't mind fixing it, I would be extremely grateful.

I don't mind fixing this, if this is the link as it should be, but it felt (to me) like this is a bug which will change again in a week. And I hate playing cat and mice.

melcorr:
> > I know it's not your responsibility, but I also wonder whose responsibility is it :o) ?
> > If you wouldn't mind fixing it, I would be extremely grateful.
> 
> I don't mind fixing this, if this is the link as it should be, but it felt (to me) like this is a bug which will change again in a week. And I hate playing cat and mice.

Yeah, you're not wrong there

PR #720, opened 2021-09-16
Reword salt

maximiliankolb:
A few rewordings for Salt plus exented Salt terminology.


melcorr (approved):
Lol, I feel bad now. I should have called that out. 
I assumed that people using Salt would understand the terminology, but this is definitely an enhancement.

maximiliankolb:
It _is_ git after all, so I think extending stuff is better than waiting forever for the "perfect" PR. Anyway, thanks for merging @melcorr

PR #728, opened 2021-09-23
Add check for orphan xrefs

lzap:
Adds a new check that will fail the build if any of the HTML output contains the following pattern: `<a href="#xyz">[xyz]</a>`

That is a result of asciidoc being unable to find a HTML anchor. In the document, it renders as a link named `[xyz]` leading to nowhere.

The PR will fail tests because there are some cases we need to fix first.

https://gist.github.com/lzap/06d9aca052b793b0ec7f8e2653ba70d7


lzap:
Aaaand it fails, that is a very nice error right there:

```
CHECKING FOR ORPHAN XREFS
<p>To use external DNS, DHCP, and TFTP services instead, see <a href="#configuring-external-services">[configuring-external-services]</a>.</p>
make[1]: *** [../common/Makefile:63: ../build/Installing_Proxy_on_Debian/index-foreman-el.html] Error 1
```

Looks like @melcorr is planning fixing those so let me know once I can rebase this one.

guides/doc-Installing_Proxy_on_Debian/master.adoc
```diff
@@ -3,12 +3,13 @@
 :numbered:
 :toc:
 :toc-placement: left
-
 include::common/attributes.adoc[]
-
 :context: {smart-proxy-context}
 :ProductName: {SmartProxyServer}
```
lzap, line 8: In asciidoc header must not contain empty lines. I fixed this here, althought, I haven't seen any differences in the output.

melcorr:
If the disconnected Satellite installation guide could only build for Satellite (we don't expose it anywhere upstream) - then there would be no xrefs to report.

lzap:
Done, rebased.

lzap:
I believe I fixed the last one (bond interface xref) it should be green now.

lzap:
I fixed this one more problem: `For more information, see <a href="#Importing_Content">[Importing_Content]</a>.</p>`

I also rebased on top of master. Can you give me blessing @maximiliankolb 

Also, would you mind being nominated for push permissions? You are closing the gap for the 3rd place and I am little worried about my gold: https://github.com/theforeman/foreman-documentation/graphs/contributors :-)

lzap:
Ooooh right! Makes sense, I should be able to figure this out.

I guess the heart was yes: https://community.theforeman.org/t/nominating-maximilian-kolb-for-foreman-documentation/25753

PR #738, opened 2021-10-07
Include pulp3 migration upstream

ryandeussing:
This PR takes certain recent downstream changes to the Upgrading and Updating guide and introduces them upstream, so that the guide can be 'upstreamed' and synchronized via the satellite synch script.

Cherry-pick into:

* [x] Foreman 3.0
* [x] Foreman 2.5 (Satellite 6.10)
* [ ] Foreman 2.4
* [ ] Foreman 2.3 (Satellite 6.9)

PR #739, opened 2021-10-07
Update product number

ryandeussing:
updates the productnumber for the Upgrading & Updating guide

see https://github.com/theforeman/foreman-documentation/pull/738 for context


ryandeussing:
that didn't work

PR #740, opened 2021-10-07
updtates productnumber

ryandeussing:
Updates the productnumber for the Upgrading and Updating guide, which will soon be 'upstreamed', see https://github.com/theforeman/foreman-documentation/pull/738 for context.


lzap:
Merged, let me know if you need to backport it, but I feel like I already did this in 2.5 branch.

PR #642, opened 2021-08-02
Use Global Registration instead of  rpm

stejskalleos:
For https://github.com/theforeman/foreman/pull/8563

PR #745, opened 2021-10-12
Use Global Registration instead of katello_ca_consumer rpm

stejskalleos:
https://github.com/theforeman/foreman-documentation/pull/642 is merged & we should propagate global registration as default method


lzap:
There are conflicts, please rebase.

@ares do you have some time to review this? LGTM

stejskalleos:
Rebased & updated with some tweaks:
* Created `common/modules/proc_host_registration_steps.adoc` with registration steps that are same for other guides
* `proc_registering-a-host-to-satellite-manually.adoc` merged to `guides/doc-Managing_Hosts/topics/Registering_Hosts.adoc` as it had almost same steps

Affected guides:
* doc-Configuring_Load_Balancer
* doc-Content_Management_Guide
* doc-Managing_Hosts

guides/common/modules/proc_host_registration_steps.adoc
```diff
@@ -0,0 +1,44 @@
+.Procedure
+
+. Generate registration command
++
+UI: In the {ProjectWebUI}, navigate to *Hosts > Register Host* and generate the registration command.
```
lzap, line 5: This smells, @melcorr help! :-)

guides/common/modules/proc_host_registration_steps.adoc
```diff
@@ -0,0 +1,44 @@
...
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# hammer host-registration generate-command
+
+# or manually:
+
+# curl -X POST https://{foreman-example-com}/api/registration_commands \
+    --user username:password \
+    -H 'Content-Type: application/json'
```
lzap, line 18: Observation: This leaves your Satellite username and password in shell history. That could have pretty interesting security consequences.

It looks like we don't have this anywhere else in the documentation, so my suggestion would be to remove `:password` and according to the manual page: _If you simply specify the user name, curl will prompt for a password._

Alternatively, keep it this way but drop a sentence below explaining this threat. I personally like the opposite - having the main copy-paste bit secure by default and instructing user below about `username:password` possibility with all the consequences.

Final solution is to unset shell history for the session, you can do this with `unset HISTFILE`. This makes it a two liner, there is a trick to do this on a single line:

    HISTFILE=/dev/null curl -X POST https://{foreman-example-com}/api/registration_commands ...

lzap: Correction: `HISTFILE=/dev/null` on a single line does not work for me, so perhaps just use `unset HISTFILE` on another line.

guides/common/modules/proc_host_registration_steps.adoc
```diff
@@ -0,0 +1,44 @@
...
+[options="nowrap" subs="+quotes,attributes"]
+----
+# hammer host-registration generate-command --activation-keys your-key
+
+# or manually:
+
+# curl -X POST https://{foreman-example-com}/api/registration_commands \
+    --user username:password \
+    -H 'Content-Type: application/json' \
+    -d '{\"activation_keys\":\"your-key\"}'"
```
lzap, line 32: Ditto.

guides/doc-Configuring_Load_Balancer/topics/Registering_Clients.adoc
```diff
@@ -6,11 +6,17 @@ For more information about registering clients and configuring them to use Puppe
...
 * xref:registering-clients-manually[]
 
+[id='registering-clients-using-the-host-registration']
+== {ProjectName} host registration
+
+include::../common/modules/proc_host_registration_steps.adoc[leveloffset=+1]
+
 [id='registering-clients-using-the-bootstrap-script']
-== Registering Clients Using the Bootstrap Script
+== **(Deprecated)** Registering Clients Using the Bootstrap Script
```
lzap, line 19: This also smells, @melcorr ?

guides/common/modules/proc_host_registration_steps.adoc
ares: it may not be such an issue if we also mention the personal access token can be used as a password and that is can have expiration set. It seems we don't have a documentation for PAT yet, otherwise I'd say just link it from here.

ares (approved):
All my comments were addressed. It needs rebase now and a response to @lzap's comments.

guides/common/modules/proc_host_registration_steps.adoc
stejskalleos: When I will finish documentation for API, with examples & with described all different authentication methods we can link it here, as @ares says.

lzap: Can you just make the option without password the default one and ask the user in the text to change "username" to "username:passwordortoken" explaining the threat? Then we can merge I think.

stejskalleos: password removed

lzap (approved):
Damn sorry my review was in pending state. Sure, let's go. We can optionally tune the security concern.

guides/common/modules/proc_host_registration_steps.adoc
lzap: Guys wants this, if you feel this needs a correction please file a PR I know you're busy.

guides/common/modules/proc_host_registration_steps.adoc
lzap: I still see it but I don't insist, let's merge.

guides/common/modules/proc_host_registration_steps.adoc
```diff
@@ -0,0 +1,44 @@
...
+ifdef::katello,satellite[]
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# hammer host-registration generate-command --activation-keys your-key
+
+# or manually:
+
+# curl -X POST https://{foreman-example-com}/api/registration_commands \
+    --user username \
```
lzap, line 30: Actually it is only removed here. Can you either keep or remove for both? To keep consistency.

lzap: If you remove it, I suggest to add a sentence below telling the user what is the format: `username:password`.

lzap (approved):
Thank you!

guides/common/modules/proc_host_registration_steps.adoc
maximiliankolb: Filed a PR: https://github.com/theforeman/foreman-documentation/pull/828

guides/doc-Configuring_Load_Balancer/topics/Registering_Clients.adoc
maximiliankolb: I've opened an issue for this: https://github.com/theforeman/foreman-documentation/issues/829

PR #752, opened 2021-10-15
CMG for Foreman 2.5 - fixed broken bullet list in section Configuring…

mdolezelova:
… definitions under Synchronizing content between Foreman servers.

Cherry-pick into:

* [ ] Foreman 3.0
* [ ] Foreman 2.4


mdolezelova:
Thank you so much, @melcorr !  I was thinking that the cherry-pick flag may not be needed in this case, but was not sure, so thank you also for answering this question :-)

PR #773, opened 2021-10-25
Hide broken/unfinished variants

lzap:
In https://github.com/theforeman/foreman-documentation/pull/728 I had to hide variants which had broken xrefs. The pages still render, but it renders to an empty page with no content, this allows our xrefs validator to pass. I forgot that we link these guides from the main menu.

Therefore I am proposing to remove these links to avoid confusion. We will add those back as soon as we fix links and in some cases also update its content.

### DIFF
--- a/web/layouts/shortcodes/guide-list.html
+++ b/web/layouts/shortcodes/guide-list.html
@@ -1,16 +1,14 @@
 {{ $path := default "nightly" (.Get "version") }}
 
-* [Planning for Foreman](/{{ $path }}/Planning_Guide/index-foreman-el.html)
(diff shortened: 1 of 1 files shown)


lzap:
> Is hiding the content really necessary at all?

Due to how building is implemented (simple makefiles) we can currently only build all combinations, even those which do not make sense. That "hiding" was supposed to only hide irrelevant things. If there is any content that should be shown, let's show it back that was not what I meant.

lzap:
This was fixed in: https://github.com/theforeman/foreman-documentation/pull/781

PR #779, opened 2021-10-26
Set ISS acronym to Inter-Server Synchronization

maximiliankolb:
This PR sets the `ISS` acronym from "Inter-Satellite Synchronization" to "Inter-Server Synchronization".

Example: [Issue in upstream guide](https://docs.theforeman.org/nightly/Content_Management_Guide/index-katello.html#Synchronizing_Content_Between_Servers_content-management)

### DIFF
--- a/guides/common/assembly_synchronizing-content-between-foreman.adoc
+++ b/guides/common/assembly_synchronizing-content-between-foreman.adoc
@@ -1,7 +1,7 @@
 [id="Synchronizing_Content_Between_Servers_{context}"]
 = Synchronizing Content Between {ProjectServer}s
 
-{ProjectName} uses Inter-{Project} Synchronization (ISS) to synchronize content between two {ProjectServer}s including those that are airgapped.
+{ProjectName} uses Inter-Server Synchronization (ISS) to synchronize content between two {ProjectServer}s including those that are airgapped.
 
 include::modules/con_iss-scenarios.adoc[leveloffset=+1]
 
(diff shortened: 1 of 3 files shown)


melcorr:
@parthaa @bbuckingham would you mind weighing in with your opinion on this? 
For me, this works and I +1 it, but ultimately you might have more insight. The ISS acronym doesn't work for Foreman and orcarhino.

lzap:
I like this.

bbuckingham:
@melcorr , @maximiliankolb , I like it; however, somewhere we should probably define what a "Server" is. 

In other words, we would not want someone to assume that it a server could be a Smart Proxy or Capsule that isn't running the Katello software.  Valid examples of server would be Foreman/Katello server upstream or Satellite server on Red Hat's downstream.

melcorr:
> @melcorr , @maximiliankolb , I like it; however, somewhere we should probably define what a "Server" is.
> 
> In other words, we would not want someone to assume that it a server could be a Smart Proxy or Capsule that isn't running the Katello software. Valid examples of server would be Foreman/Katello server upstream or Satellite server on Red Hat's downstream.

Thank you @bbuckingham 
Do you think this sentence needs expanding? https://github.com/theforeman/foreman-documentation/pull/779/files#diff-aa9c58a9f236b92654693599a2cef809a41bde4449de72209228aeb9826d4725R4 
This works for all three scenarios.

bbuckingham:
@melcorr , ACK, given ProjectServer provides the distinction. Thanks!

melcorr (approved):
I approve :-) 
Bless your brains @maximiliankolb !

PR #781, opened 2021-10-26
Some guides in nightly are empty

maximiliankolb:
Three guides are currently empty:

- [Planning for Foreman](https://docs.theforeman.org/nightly/Planning_Guide/index-foreman-el.html)
- [QuickStart Guide](https://docs.theforeman.org/nightly/Quickstart_Guide/index-katello.html)
- [Upgrading and Updating Foreman](https://docs.theforeman.org/nightly/Upgrading_and_Updating/index-foreman-el.html)

This PR removes the conditionals to show content again.

### DIFF
--- a/guides/doc-Planning_Guide/master.adoc
+++ b/guides/doc-Planning_Guide/master.adoc
@@ -2,9 +2,6 @@ include::common/attributes.adoc[]
 :imagesdir: images
 include::common/header.adoc[]
 
-// Render only for relevant and finished contexts
-ifdef::katello,satellite,orcharhino[]
-
 = Planning for {ProjectName}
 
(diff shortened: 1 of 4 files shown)


lzap:
This started in the xref checker, this "hiding" allowed me to get rid of irrelevant combinations.

While I was working on that, I reviewed all guides and removed all that could be relevant but are not yet ready for publication.

Only then I forgot that we still link them and proposed https://github.com/theforeman/foreman-documentation/pull/773

If you think it is better not to remove them, I am good with this. You fixed the xhref failure so we are all good.

maximiliankolb:
Thanks for merging. I think we should -for now- leave it as is.

PR #774, opened 2021-10-25
BZ 2008947: Procedure to rename Satellite Server

Lennonka:
Add parameter to update host name URL in Puppet after renaming Satellite Server

https://bugzilla.redhat.com/show_bug.cgi?id=2008947

Cherry-pick into:

* [ ] Foreman 3.0
* [x] Foreman 2.5 (Satellite 6.10)
* [ ] Foreman 2.4
* [x] Foreman 2.3 (Satellite 6.9)
* [x] Foreman 2.1 (Satellite 6.8)

PR #795, opened 2021-11-01
BZ 2008947: Condition host name update in Puppet config…

Lennonka:
… after renaming Satellite Server (Puppet possibly disabled)

Follow-up to #774

Cherry-pick into:

* [x] Foreman 3.0


ekohl (commented):
Technically disabling Puppet is not supported for Satellite until Satellite 7.0 and entirely undocumented in the current documentation. I'd suggest we hold off on this and keep this work for a much larger "how to deal with a disabled Puppet" story. As we speak, we're working hard as engineering about how to deal with this.

PR #809, opened 2021-11-04
Build and check with local HTML links

lzap:
Our Github Action builds HTML pages for all variants and then it calls linkchecker on them. So far so good.

Now, we build with BaseURL that points to real sites:

* docs.foreman.org
* Red Hat Docs
* ATIX docs

Problem is, real sites are often out of date.

This patch solves it, it builds all HTML pages with BaseURL set to `file:///absolute/path/to/build/dir/` which then correctly maps all links. Link validation is more relevant as it checks links of the actual pull request. It is also faster.

And I also found that we call the linkchecker 5 times instead of just one. This will dramatically speed up pull request tests.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -52,8 +52,21 @@ jobs:
       - name: Clean the environment
         run: make clean
 
-      - name: Build HTML
+      - name: Build HTML with local links
         run: |
+          make html BUILD=foreman-el LINKS=LOCAL
+          make html BUILD=foreman-deb LINKS=LOCAL
+          make html BUILD=katello LINKS=LOCAL
+          make html BUILD=satellite LINKS=LOCAL
+          make html BUILD=orcharhino LINKS=LOCAL
+
+      - name: Check HTML links of all builds
+        run: |
+          make linkchecker-tryer
+
+      - name: Rebuild HTML with real links
+        run: |
+          make clean
           make html BUILD=foreman-el
           make html BUILD=foreman-deb
           make html BUILD=katello
@@ -65,14 +78,6 @@ jobs:
         with:
           name: foreman-docs-html-${{ env.BRANCH_NAME }}
           path: guides/build/
-
-      - name: Check HTML links
-        run: |
-          make linkchecker-tryer BUILD=foreman-el
-          make linkchecker-tryer BUILD=foreman-deb
-          make linkchecker-tryer BUILD=katello
-          make linkchecker-tryer BUILD=satellite
-          make linkchecker-tryer BUILD=orcharhino
 
   build-pdf:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -54,25 +54,21 @@ jobs:
...
 
-      - name: Check HTML links
+      - name: Check HTML links of all builds
         run: |
-          make linkchecker-tryer BUILD=foreman-el
-          make linkchecker-tryer BUILD=foreman-deb
-          make linkchecker-tryer BUILD=katello
-          make linkchecker-tryer BUILD=satellite
-          make linkchecker-tryer BUILD=orcharhino
+          make linkchecker-tryer
```
lzap, line 71: This was not necessary, this makefile target simply finds all HTML pages and checks them. We can only call this once, BUILD variable is unused for this case.

guides/Makefile
```diff
@@ -13,10 +13,10 @@ pdf: all-pdf
 clean: all-clean
 
 linkchecker:
-	find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern
+	find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern --verbose
 
 linkchecker-tryer:
-	find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern | linkchecker-tryer
+	find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern --verbose | linkchecker-tryer
```
lzap, line 20: I am adding verbosity to be able to verify links are actually rendered as expected (locally).

guides/common/Makefile
```diff
@@ -24,6 +25,10 @@ ifeq ($(UNAME), Darwin)
 BROWSER_OPEN = open
 endif
 
+ifeq ($(LINKS), LOCAL)
+BASEURL_ATTR = "-a BaseURL=file://$(BUILD_DIRA)"
+endif
+
```
lzap, line 31: When `LINKS=LOCAL` is passed, it adds the local BaseURL. Otherwise it builds as usual.

lzap:
To check links, compare before/after:

```
[lzap@nuc foreman-documentation]$ ag file:// guides/build/ | wc -l
224
```

lzap:
HTML build is now faster, previously about 6 minutes, now about 2 minutes. The bottleneck is now PDF building.

Since we never had a PDF build error, maybe we don't need to run it for pull request actions?

maximiliankolb (approved):
I've checked the output and it looks good to me! I am not super familiar with GHA, but it looks reasonable 👍

lzap:
Let's merge it then, I can speed up PDF building in another PR. Need to figure out how to share variables.

lzap:
https://github.com/theforeman/foreman-documentation/pull/813

PR #811, opened 2021-11-09
Visually separate the abort and reset commands in the Pulp migration p…

ianf77:
…rocess

Separated using notes.

Bug 2013682 - Remove satellite-maintain content migration-reset as a step

https://issues.redhat.com/browse/SATDOC-490

Cherry-pick into:

* [ ] Foreman 3.0

### DIFF
--- a/guides/doc-Upgrading_and_Updating/topics/preparing_to_migrate_pulp_content.adoc
+++ b/guides/doc-Upgrading_and_Updating/topics/preparing_to_migrate_pulp_content.adoc
@@ -51,21 +51,29 @@ This might take some time on high latency systems.
 # {foreman-maintain} content prepare
 ----
 +
-You cannot use `*Ctrl*` *+* `*C*` to terminate this process.
-If a user attempts to halt the process using CTRL + C or by disconnecting their SSH session, the process does not terminate but continues in the background.
+[Note]
+====
+You cannot use *`Ctrl` + `C`* to terminate this process.
+If you attempt to halt the process using *`Ctrl` + `C`* or by disconnecting your SSH session, the process does not terminate but continues in the background.
 You can use the following command to terminate the process gracefully, whenever necessary, so that you can continue later.
-+
+
 [options="nowrap", subs="verbatim,quotes,attributes"]
 ----
 # {foreman-maintain} content prepare-abort
 ----
-+
+
 Note that `{foreman-maintain} content-prepare abort` can take several minutes to terminate the process.
-. Continue the migration process using `{foreman-maintain} content-prepare` whenever it is convenient.
-. If problems occur, you can restart the pre-migration process from the beginning:
+You can continue the migration process using `{foreman-maintain} content-prepare` whenever it is convenient.
+====
 +
+. The migration process is not fully complete until the upgrade of {ProjectServer} from {ProjectVersionPrevious} to {ProjectVersion} is complete.
++
+[Note]
+====
+If problems occur, you can restart the pre-migration process from the beginning using the following command:
+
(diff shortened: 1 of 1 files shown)


lzap:
From the issue: _Is being shown a next step for step 4. if a person does not pay close attention they might actually perform this as a successive step aborting the preparation for the migration._

Generally we are against these NOTE boxes, but I agree we should do something with this:

Snímek obrazovky 2021-11-10 v 15 23 11

@maximiliankolb what do you suggest? I would be perhaps fine with this.

lzap:
Oh wait, this renders incorrectly, please fix this:

Snímek obrazovky 2021-11-10 v 15 27 11

What would be really helpful if you could attach image just like me so we can asses this more quickly.

lzap:
Don't set the labels please, this is for us. Also, I will ask you to file this against 2.5 too, we have a new policy of cherry picking. We will require PRs from now on to ensure all changes in 2.5 are passing builds.

lzap:
Thank you, don't start until we land on a good solution. I am not sold on using NOTE, @maximiliankolb help me here?

I think just making this a subchapter would be better, those note boxes are not mean for structure content.

guides/doc-Upgrading_and_Updating/topics/preparing_to_migrate_pulp_content.adoc
```diff
@@ -51,21 +51,30 @@ This might take some time on high latency systems.
...
 # {foreman-maintain} content prepare-abort
 ----
-+
+
 Note that `{foreman-maintain} content-prepare abort` can take several minutes to terminate the process.
-. Continue the migration process using `{foreman-maintain} content-prepare` whenever it is convenient.
-. If problems occur, you can restart the pre-migration process from the beginning:
+You can continue the migration process using `{foreman-maintain} content-prepare` whenever it is convenient.
+====
 +
```
maximiliankolb, line 69:
```suggestion
```

guides/doc-Upgrading_and_Updating/topics/preparing_to_migrate_pulp_content.adoc
```diff
@@ -51,21 +51,30 @@ This might take some time on high latency systems.
...
-You cannot use `*Ctrl*` *+* `*C*` to terminate this process.
-If a user attempts to halt the process using CTRL + C or by disconnecting their SSH session, the process does not terminate but continues in the background.
+[NOTE]
+====
+You cannot use *`Ctrl` + `C`* to terminate this process.
+If you attempt to halt the process using *`Ctrl` + `C`* or by disconnecting your SSH session, the process does not terminate but continues in the background.
 You can use the following command to terminate the process gracefully, whenever necessary, so that you can continue later.
-+
+
+
```
maximiliankolb, line 60:
```suggestion
```

maximiliankolb (commented):
Two small nitpicks; but the built docs look good to me!

I am a fan of using `NOTE|WARNING|CAUTION|TIP` [admonitions](https://docs.asciidoctor.org/asciidoc/latest/blocks/admonitions/) but I know not everyone shares my excitement. If you want to go the extra mile and create a separate procedure to stop the content migrate process, that would be great as well.

Is content migration something that is still relevant for Foreman 3.x+ ? If not, maybe we don't have to spend too much time with this.

I would be fine with your PR.

maximiliankolb (approved):

lzap (approved):
Allright then, I have no problems with these notes. We can easily search them if we decide its a problem later.

lzap:
Merged, thanks. @ianf77 please do the same for 2.5.

PR #812, opened 2021-11-09
Added statement that Pulp migration increases upgade time.

ianf77:
Bug 2013267 - Need to update time required to complete the upgrade in Prerequisites section of Sat6.10

https://issues.redhat.com/browse/SATDOC-477

Cherry-pick into:

* [ ] Foreman 3.0


ekohl (changes requested):
I think this only really belongs in the 2.5-stable branch. master follows what 3.1 will be and in upstream we're way beyond the point where Pulp 2 was a thing. In fact, we're expecting at this point that the user has removed Pulp 2 altogether and things will break if they haven't.

ianf77:
Removed the cherry-pick to v3.

ekohl:
It still shouldn't be a PR to the master branch, only against 2.5. You can change the base branch of a PR but it may be easier to open a new PR.

ianf77:
Ok

PR #813, opened 2021-11-09
Build PDFs only on deploy

lzap:
I tried but this is the best I am able to achieve with GHA.

This makes PDFs to be built only when we deploy (merge into master). It is slow and we have never had a build failure for PDF. It is safe to do.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -10,9 +10,12 @@ on:
     branches:
       - master
       - "[0-9]+.[0-9]+"
+env:
+  RUBY_VERSION: 2.7
+  PYTHON_VERSION: 2.x
 
 jobs:
-  build-html:
+  check-html:
     runs-on: ubuntu-latest
     defaults:
       run:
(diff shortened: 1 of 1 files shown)


lzap:
Do not merge, fighting with GHA syntax. YAML is good, BUT FOR DATA NOT CODE! :-)

https://github.community/t/accessing-environment-variable/211159

ekohl:
> Do not merge, fighting with GHA syntax. YAML is good, BUT FOR DATA NOT CODE! :-)

I'm not sure I agree with "YAML is good" but I certainly agree that it's not good for code.

Overall :+1: for not building PDFs. I'm not sure anyone ever uses them.

lzap:
> I'm not sure anyone ever uses them.

Yeah, the plan is to make sure they are working until we figure out how to publish them in the future. We can do it after merge, it is not critical now if they break.

lzap:
Done, I could not figure it so I copy pasted the block twice. I at least broke it into multiple lines and shortened it a bit.

Also another change is that now we run HTML build and HTML local build+check in parallel. This speeds things even up. Total time is down from approx. 6 minutes to 2 minutes and that counts! This is now ready.

.github/workflows/deploy.yml
```diff
@@ -10,9 +10,12 @@ on:
     branches:
       - master
       - "[0-9]+.[0-9]+"
+env:
+  RUBY_VERSION: 2.7
+  PYTHON_VERSION: 2.x
```
lzap, line 15: I also extracted these into variables and bumped Ruby to 2.7 version.

maximiliankolb (approved):
Fine for me. I would even +1 dropping the pdf document altogether.

lzap:
For the record here is the deploy run that is supposed to build PDF actually: https://github.com/theforeman/foreman-documentation/actions/runs/1444557713

PR #816, opened 2021-11-09
Fix 2.5 gha

maximiliankolb:
Trying to figure out why 2.5 does not build & why there was no GHA workflow.

PR #819, opened 2021-11-10
Require PRs for backporting

lzap:
We had issues with backporting some of the changes. Cherry picking is mostly seamless process, but due to amount of changes recently and also due to the fact that Red Hat Documentation team requires backporting into one year old branch, we very often encounter conflicts breaking the builds.

This patch removes the checkbox from the PR template and asks authors to file PRs on their own. This way we can ensure good quality by running builds via Github Actions.

I would suggest that Red Hat Documentation team would work earier in the process, e.g. when features are being completed rather than after beta Satellite release to avoid extra work of backporting changes.

https://github.com/theforeman/foreman-documentation/pull/816

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,7 +2,7 @@
 Cherry-pick into:
 
 * [ ] Foreman 3.0
-* [ ] Foreman 2.5 (Satellite 6.10)
+* For Foreman 2.5, file a separate PR request
 
 <!---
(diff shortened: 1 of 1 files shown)


melcorr (approved):

melcorr:
Yeah. The maintenance of this branch with such low participation from Red Hat docs in the upstream is an unreasonable burden on our time. I am merging this.

ekohl (commented):
Based on experience in other repositories I've been wondering if we should apply the policy altogether: require the author to open cherry pick PRs.

Anyway, retroactive :+1: for this PR.

PR #674, opened 2021-08-18
Adding chapter to Content Management Guide on Convert2RHEL process

ianf77:
Adding files to 3086

doc] Satellite Convert2RHEL support

https://issues.redhat.com/browse/SAT-3086

Cherry-pick into:

* [x] Foreman 3.0
* [x] Foreman 2.5 (Satellite 6.10)
* [ ] Foreman 2.4
* [ ] Foreman 2.3 (Satellite 6.9)
* [ ] Foreman 2.1 (Satellite 6.8)

PR #821, opened 2021-11-10
adds convert2RHEL content

ryandeussing:
PR #674 was not cherry-picked into 2.5, and as it has been a while
I queued up the changes in a new PR to avoid conflicts.


lzap (approved):
Thanks, we are still working on the process. Looks clean, tests are green.

PR #822, opened 2021-11-10
BZ 2021473: Specify disk space requirement for Pulp content migration

Lennonka:
https://bugzilla.redhat.com/show_bug.cgi?id=2021472

Cherry-pick into:

* [x] Foreman 3.0
* [x] Foreman 2.5 (Satellite 6.10)

PR #825, opened 2021-11-10
BZ 2021473: Specify disk space requirement for Pulp content migration on 2.5

Lennonka:
Follow-up to #822

### DIFF
--- a/guides/doc-Upgrading_and_Updating/topics/upgrading_satellite_server_prerequisites.adoc
+++ b/guides/doc-Upgrading_and_Updating/topics/upgrading_satellite_server_prerequisites.adoc
@@ -19,6 +19,13 @@ In the {ProjectWebUI}, navigate to *Monitor* > *Audits* and search for the templ
 If you use the export method, restore your changes by comparing the exported template and the default template, manually applying your changes.
 ifdef::satellite[]
 * Pulp content migration occurs at {ProjectVersionPrevious}, before the upgrade to {ProjectVersion}.
+* Ensure there is sufficient storage space for `/var/lib/pulp/published` to double in size before starting the migration process.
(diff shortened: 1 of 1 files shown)


lzap (approved):
And thank you again!

guides/doc-Upgrading_and_Updating/topics/upgrading_satellite_server_prerequisites.adoc
```diff
@@ -19,6 +19,13 @@ In the {ProjectWebUI}, navigate to *Monitor* > *Audits* and search for the templ
 If you use the export method, restore your changes by comparing the exported template and the default template, manually applying your changes.
 ifdef::satellite[]
 * Pulp content migration occurs at {ProjectVersionPrevious}, before the upgrade to {ProjectVersion}.
+* Ensure there is sufficient storage space for `/var/lib/pulp/published` to double in size before starting the migration process.
```
ekohl, line 22: While this is already merged, I'd suggest to reduce this to `/var/lib/pulp`. `/var/lib/pulp/published` is only for Pulp 2 while Pulp 3 uses other directories under `/var/lib/pulp`.

Another thing to note is that `/var/lib/pulp` should not be split into multiple partitions: the migration wants to hardlink files between Pulp 2 and Pulp 3, which it can't do over multiple partitions. Then the migration will copy files which is a LOT slower. So `/var/lib/pulp` can be a separate partition, but `/var/lib/pulp/published` can't.

I'm also wondering if this is really correct. I think it's mostly the PostgreSQL database that will greatly increase in size during the migration, not the actual files on disk. The Bugzilla number doesn't explain why and looks completely unrelated.

PR #832, opened 2021-11-11
updates Satellite versions in 2.5

ryandeussing:
Updates attributes and docinfo files changing 6.10-beta to 6.10 where appropriate.

### DIFF
--- a/guides/common/attributes-base.adoc
+++ b/guides/common/attributes-base.adoc
@@ -17,11 +17,11 @@
 :RepoRHEL7ServerOptional: rhel-7-server-optional-rpms
 :RepoRHEL7ServerAnsible: rhel-7-server-ansible-2.9-rpms
 // For Beta, change to "Beta". For GA releases, change to, for example, "6.8".
-:RepoRHEL7ServerSatelliteServerProductVersion: rhel-server-7-satellite-6-beta-rpms
+:RepoRHEL7ServerSatelliteServerProductVersion: rhel-server-7-satellite-6-rpms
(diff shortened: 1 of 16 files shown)


ryandeussing:
@lzap found help with a patch error in a [corner of the internet I've never visited before](https://lukas.zapletalovi.com/2011/11/gits-crying-patch-format-detection.html) ;)

maximiliankolb:
I think this PR is missing https://github.com/theforeman/foreman-documentation/blame/2.5/guides/common/attributes.adoc#L18

ryandeussing:
@maximiliankolb thank you but in this case 6.10 is the value we want in the 2.5 branch, so this PR includes that change.

The advice Ian got, and passed on to me, is that the AccessRedHatComVersion attribute in master should always represent and existing, published version of Red Hat docs. That's why the change to 6.10 didn't get submitted to master first.

If that is poor advice, let me know and I'll spread that information.

lzap:
That is correct, I mean, we don't want to break links in our nightly documentation by referring to documentation set that hasn't been published yet. Nightly is actually our the most visited documentation version as Google prefers it the most. Let me take a look.

lzap (approved):
LGTM.

lzap:
Merged.

FYI I created a PR for the remaining attribute, we can merge it only after 6.10 docs are out: https://github.com/theforeman/foreman-documentation/pull/838

maximiliankolb:
OK, my bad @ryandeussing

PR #833, opened 2021-11-12
Add procedure to create operating system entry for Debian

maximiliankolb:
This PR adds two procedures to provision hosts running Debian and Ubuntu.

### DIFF
--- a/guides/common/modules/proc_creating-an-installation-medium-for-debian.adoc
+++ b/guides/common/modules/proc_creating-an-installation-medium-for-debian.adoc
@@ -1,0 +1,23 @@
+[id="Creating_an_Installation_Medium_for_Debian_{context}"]
+= Creating an Installation Medium for Debian
+
+Create an installation medium in {Project} to provision hosts running Debian or Ubuntu.
+
+ifdef::orcharhino[]
+[TIP]
+====
+You can use an Ansible role to configure operating systems to skip this step.
+The Ansible role is located on your orcharhino Server at `/usr/share/orcharhino-ansible/roles/or_operating_systems/`.
+====
+endif::[]
+
+.Procedure
+. In the {ProjectWebUI}, navigate to *Hosts > Installation Media*.
+. Click *Create Medium* to create an installation medium entry.
+. Enter a *Name* for the installation medium.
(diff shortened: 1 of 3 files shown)


guides/common/modules/proc_creating-an-installation-medium-for-debian.adoc
```diff
@@ -0,0 +1,23 @@
...
+
+Create an installation medium in {Project} to provision hosts running Debian or Ubuntu.
+
+ifdef::orcharhino[]
+[TIP]
+====
+You can use an Ansible role to configure operating systems to skip this step.
+The Ansible role is located on your orcharhino Server at `/usr/share/orcharhino-ansible/roles/or_operating_systems/`.
+====
+endif::[]
```
lzap, line 12: Oh I want this in Foreman core someday too :-)

guides/common/modules/proc_creating-an-installation-medium-for-debian.adoc
```diff
@@ -0,0 +1,23 @@
...
+
+.Procedure
+. In the {ProjectWebUI}, navigate to *Hosts > Installation Media*.
+. Click *Create Medium* to create an installation medium entry.
+. Enter a *Name* for the installation medium.
+. Set the *Path* to the upstream URL `\http://ftp.debian.org/debian/`.
++
+Alternatively, you can also use a Debian Mirror.
+Note that synchronized content from {Project} does not work for two reasons: first, the `linux` and `initrd.gz` files are not synchronized; second, the `Release` file is not signed with the official Debian GPG private key.
```
lzap, line 21: You could be more specific on what exactly does not work. Perhaps instead stating that something does not work, which generally is something that RHD does not do, we could tell what needs to be done in order to provision the host. I would assume if files were `wget/curled` on the TFTP server it could work. Anyway, not blocking.

lzap (approved):
Merging.

guides/common/modules/proc_creating-an-installation-medium-for-debian.adoc
quba42: It would probably need to be made quite a bit more generic, to add it to the upstream collections.
The next role, I do want to add to the collection is the one to configure Katello to sign deb content.

PR #834, opened 2021-11-12
Extend ACD procedures

maximiliankolb:
This PR adds four more ACD procedures:

* Add procedure to change host parameters for ACD
* Add procedure to overwrite Ansible group variables in ACD
* Add procedure to view application instance reports for ACD
* Add procedure to view last deployment task for ACD


lzap (approved):
Could not spot anything, this brought me to the idea to use aspell actually :-)

PR #838, opened 2021-11-12
Update Satellite links to 6.10

lzap:
Do not merge yet until 6.10 is announced.


ryandeussing:
Thanks for this!

lzap:
For the record, tests are failing - that is expected. We need to re-run them as soon as 6.10 is published online.

PR #844, opened 2021-11-15
Fix spelling of authentication

maximiliankolb:


maximiliankolb:
I took some liberties here.

PR #794, opened 2021-11-01
Removed information from Satellite docs which is no longer supported

ianf77:
Where relevant ifndef and ifdef statement s have been used to remove material from Satellite documentation

Bug 1938633 - We have to remove those topic from documentation, which we removed functionality from new satellite version

https://issues.redhat.com/browse/SATDOC-473

Cherry-pick into:

* [X] Foreman 3.0
* [X] Foreman 2.5 (Satellite 6.10)

PR #849, opened 2021-11-15
Removed additional PXE-less provisioning entry

ianf77:
Removed entry missing from original commit

Bug 1938633 - We have to remove those topic from documentation, which we removed functionality from new satellite version

https://issues.redhat.com/browse/SATDOC-473

Cherry-pick into:

* [ ] Foreman 3.0


lzap:
I do not see a single word about PXE-less discovery in here:

https://access.redhat.com/documentation/en-us/red_hat_satellite/6.9/html/release_notes/release_information#removed_functionality

Also, I am PXE-less discovery maintainer and I have no information about this being dropped form Satellite. This is based on incorrect information. We are dropping support for some bootdisk types, that one I can confirm.

I am not merging this into docs, there must be some misinformation somewhere.

ianf77:
If this is the case, why was #794 merged. This is just a sentence that escaped the previous change.

lzap:
This change sneaked through my review, it should not have been there.

PR #852, opened 2021-11-16
Fix broken xref

maximiliankolb:


maximiliankolb:
Found another broken xref.

lzap (approved):
Not picked by our stupid grep checker :-)

PR #761, opened 2021-10-20
Reformatted Port Requirement Tables

ianf77:
The port requirement tables have been redesigned and now incorporate more information on
the functions of each port. They are divided into outgoing and incoming traffic for
Satellite and Capsule

[doc] Reformat the port requirements tables

https://issues.redhat.com/browse/SAT-3277

Cherry-pick into:

* [X] Foreman 3.0
* [X] Foreman 2.5 (Satellite 6.10)
* [ ] Foreman 2.4
* [ ] Foreman 2.3 (Satellite 6.9)

PR #856, opened 2021-11-16
Replace links to access.redhat.com with an attr

ekohl:
This introduces an attribute for the access.redhat.com for multi-guides. The question is if this should really be linking to the base URL. That also makes them usable in upstream and Orcharhino. Introducing a single attribute makes it easier to replace it in the future.


guides/common/attributes.adoc
```diff
@@ -15,6 +15,8 @@
 
 // Use current RH Satellite release version for links to access.redhat.com
 :AccessRedHatComVersion: 6.10-beta
+// TODO: Should this be removed and really always be BaseURL?
+:AccessMultiUrl: https://access.redhat.com/documentation/en-us/red_hat_satellite/{AccessRedHatComVersion}/html/
```
lzap, line 19: Isn't this exactly what we have here:

```
$ ag :BaseURL: -A2 -B1
guides/common/attributes.adoc
51-:SiteURL: https://docs.theforeman.org
52::BaseURL:  {SiteURL}/{ProjectVersion}/
53-include::attributes-base.adoc[]
54-
--
68-ifdef::satellite[]
69::BaseURL: https://access.redhat.com/documentation/en-us/red_hat_satellite/{AccessRedHatComVersion}/html-single/
70-include::attributes-satellite.adoc[]
71-endif::[]
--
73-ifdef::orcharhino[]
74::BaseURL: https://docs.orcharhino.com/
75-include::attributes-orcharhino.adoc[]
76-endif::[]
```

lzap:
I think we should just use BaseURL instead. Or is there a reason to use html/ instead of html-single? In that case move it to the same file and I would suggest some better name like `MultiBaseURL` perhaps.

guides/common/attributes.adoc
ekohl: Notice it's `/html/` vs `/html-single/`. I can't judge if we can align on one or the other though I'd strongly prefer that.

ekohl:
I feel that I'm unqualified to answer that. I'd love to align on one or the other so we end up with just `BaseURL`.

However, I think your `MultiBaseURL` may be a good one. We can even set those correctly for upstream and Orcharhino so we're a step closer to fully making it a more integrated documentation site.

@ianf77 or @ryandeussing can you shed some light on the Satellite side of `html` vs `html-single`?

ryandeussing:
I always use html-single because command-F is the only way I can find what I'm looking for quickly or confirm that it's not in the doc. This is true of most doc sets, and I find it particularly true of Red Hat doc sets.

My hypothesis is that this is true of most customers who are in a `find` or `do` mode vs a `learn` mode.

ekohl:
I think I share that preference for the reasons you mention. Should we make all documents link to `html-single` (which is now `BaseURL`)?

ryandeussing:
My vote is yes, but @mdolezelova is the new program manager for Satellite docs and I will let her decide.

mdolezelova:
Hello,  @ryandeussing , thank you for looping me in! @ekohl , I have no major concerns regarding your suggestion. However, what I'd like to point out is that from my experience with practices on RHEL, we don't have any strict guidelines for using just html (multi page) or html-single (single page). In some cases, when we expect the reader to follow up by searching through the whole doc, it makes sense to use the single-page link. While in some other cases, when we really want to reference only a specific section, and don't expect the reader to go through the whole doc, it makes more sense to use the multi-page link. So not knowing all the background about the logic behind the format of the links in Satellite docs, I would be hesitant to jsut confirm right away. @ryandeussing could you maybe share more details about the logic used so far? Does the whole doc set really use the html-sinlge as a default option? Thank you!

melcorr:
@mdolezelova  historically, we always used `html`, given the fact that we factored in potential accessibility issues with sysadmins accessing docs from areas of low connectivity. However, I believe @adahms informed @spetrosi in the last year that there is a move away from this thinking, and html-single is going to become the standard. I do not have the thread to hand but can find it if necessary. Over the last year or so, html-single has been creeping in, and probably should replace html.

spetrosi:
To provide some context on :BaseURL:, at the time I was introducing it, I needed a way to create a way to make URLs us attributes that fit both upstream and downstream. I have chosen html-single format because it is consistent between Foreman and Satellite docs, because Foreman docs are html-single only. Building Foreman-docs to multi-paged html would have been a large mess. I discussed it with @adahms and we agreed that using html-single is fine.
To give an example, see URL to the same section in multi-paged SatDocs, html-single SatDocs, and Foreman docs:
https://access.redhat.com/documentation/en-us/red_hat_satellite/6.10/html/content_management_guide/importing_content#changing_the_download_policy_for_a_repository
https://access.redhat.com/documentation/en-us/red_hat_satellite/6.10/html-single/content_management_guide/index#changing_the_download_policy_for_a_repository
https://docs.theforeman.org/nightly/Content_Management_Guide/index-katello.html#changing_the_download_policy_for_a_repository
 
Satdocs html-single and Foreman UTL formats are the same, while SatDocs multi-paged html has the additional parent-section name in it's URL `importing_content#`. That's why html-single was chosen.

melcorr:
I appreciate the clarification @spetrosi

ekohl:
Thank you all for the clarifications. It sounds like we can't just use `BaseURL` everywhere. We need to at least check all section links.

Now the question is how to move forward. I'm proposing that I rebase this to resolve the merge conflicts. Then we can merge this because I still think a common variable is better than duplicating it everywhere. If you agree, I'll do that.

The next steps should be to decide how to move forward. I see some options. I'm not sure how to refer to it, but when I write projects, I mean foreman, satellite and orcharhino
* All projects agree on a single page format to link to (probably `html-single`). Then everything can use BaseURL again and everything is unified
* There is still a desire for `html` vs `html-single`
  * All projects have both formats and there are 2 variables
  * Some projects prefer a single, some multi. This means we need a good technical solution to properly cross link. That may be to define variables for all projects for all URLs (as we already have a bunch). That's probably better than if/else everywhere in the codebase. Perhaps there's also some macro/function we could use

Does that make sense? Am I missing something?

guides/common/attributes.adoc
spetrosi: Did you check if this works with html? I don't think it would because html link requires linking a section and then a subsection. I think BaseURL is doing what you are trying to do here.

guides/doc-Upgrading_and_Updating/topics/upgrading_a_disconnected_satellite.adoc
```diff
@@ -265,7 +265,7 @@ In such a case, change directory, for example to the *_root_* user's home direct
 ====
 +
 If the script fails due to missing or outdated packages, you must download and install these separately.
-For more information, see the https://access.redhat.com/documentation/en-us/red_hat_satellite/{AccessRedHatComVersion}/html-single/installing_satellite_server_from_a_disconnected_network/installing-satellite-server-disconnected#resolving-package-dependency-errors_satellite[Resolving Package Dependency Errors] section in the _Installing {ProjectServer} from a Disconnected Network_ guide.
+For more information, see the {BaseURL}/installing_satellite_server_from_a_disconnected_network/installing-satellite-server-disconnected#resolving-package-dependency-errors_satellite[Resolving Package Dependency Errors] section in the _Installing {ProjectServer} from a Disconnected Network_ guide.
```
maximiliankolb, line 268: Does this for all all projects? IMHO this would need a "InstallingServerDisconnectedDocURL" attribute.

maximiliankolb:
By `html-single` you mean "one guide is on one HTML page", for example [Foreman Provisioning Guide](https://docs.theforeman.org/nightly/Provisioning_Guide/index-foreman-el.html)?

Downstream, we do the same. I think Red Hat offers both version: [single](https://access.redhat.com/documentation/en-us/red_hat_satellite/6.10/html-single/provisioning_guide/index) and [multi](https://access.redhat.com/documentation/en-us/red_hat_satellite/6.10/html/provisioning_guide/index).

I prefer what you call "html-single" (which is a very confusing term because it could also mean one adoc page becomes one html page; as in " a single topic per page"^^). But I don't need a say in what you do when you link to access.redhat.com, as long as it's easy to read in adoc and easy to maintain.

IMHO the future of Foreman Documentation hopefully brings more separated guides which are in turn not as excessively long as they are right now.

guides/common/attributes.adoc
ekohl: Note I'm not using it everywhere, just replaced the hardcoded substring with a variable. All links I'm replacing were already `html` so if it's broken, it was broken before.

guides/doc-Upgrading_and_Updating/topics/upgrading_a_disconnected_satellite.adoc
ekohl: I'm not too familiar with the conventions. Is that something we could agree on and write down somewhere? That would also be a good thing for people new to this repository.

ekohl:
> By `html-single` you mean "one guide is on one HTML page", for example [Foreman Provisioning Guide](https://docs.theforeman.org/nightly/Provisioning_Guide/index-foreman-el.html)?

Yes, I do indeed.

> IMHO the future of Foreman Documentation hopefully brings more separated guides which are in turn not as excessively long as they are right now.

I agree.

guides/doc-Upgrading_and_Updating/topics/upgrading_a_disconnected_satellite.adoc
maximiliankolb: No convention, but there's already a small list; basically each guide has its own `DocURL` attribute: https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-base.adoc#L1

IMHO: one directory in `guides/` == one attribute.

ekohl: I think that makes sense. Especially since in upstream we need to take the build into account (foreman-el, foreman-deb, katello) which is reflected in the index name.

lzap:
Okay to sum this up:

* Use BaseURL to link to Foreman itself or single-html pages on RH Portal
* Use AccessMultiURL to link to multi-html pages on RH Portal

Why do we need to differentiate between the two? I would expect just here to agree on one or another and update BaseURL with that. I also prefer single-page, although some huge guides can be tough to open (rendering of the sidebar can be slower and download can be big). I must be missing something - why are you introducing a new attribute. I prefer simple things. In case this is wanted change, can you drop a comment in the README explaining which one to pick and when?

People favoring single-page versions is a symptom of lack of any search on our portal. I think this should be fixed and we should then default to (single page) html versions in the future.

mdolezelova:
Hello, thank you for these efforts in simplifying linking in Foreman and Satellite docs. However, for the downstream docs published by Red Hat, I would still be very hesitant to say "yes" to the replacement of all links by the html-single version. In general, we use both html-single and html-multi based on the context. In many cases it's more logical to link to the html-multi version.  Moreover, our current publication tool supports the "direct documentation feedback"  ONLY on the multi-page version. By using the DDF feature, customers are able to provide their feedback directly - by highlighting a certain part of the text, clicking the "Add feedback" button, a BZ is opened automatically. I'd like to bring this topic to the attention our Satellite content strategist Shweta Jalgaonkar to ask for her opinion before we make any conclusions for downstream docs. Apologies and thanks for your patience!

ekohl:
Rebased to resolve conflicts. It now introduces an `InstallingProjectDisconnectedDocUrl` attribute. To keep links easier, I've based it on #921 (and included that PR here). It still keeps the links to the multi guides, but it uses an attribute to reduce the duplication. I'd taken @lzap's suggestion to name it `MultiBaseURL` and added a note in the attributes to explain why it's there.

maximiliankolb (approved):
LGTM.
I suggest we merge https://github.com/theforeman/foreman-documentation/pull/921 first, then you can rebase to master again.

guides/doc-Administering_Red_Hat_Satellite/topics/Renaming_a_Satellite_Server.adoc
```diff
@@ -10,7 +10,7 @@ For more information about configuring external authentication, see xref:Configu
 
 If you use virt-who, you must update the virt-who configuration files with the new host name after you run the `{project-change-hostname}` script.
 ifndef::orcharhino[]
-For more information, see https://access.redhat.com/documentation/en-us/red_hat_satellite/{AccessRedHatComVersion}/html/configuring_virtual_machine_subscriptions_in_red_hat_satellite/troubleshooting-virt-who#modifying-virt-who-configuration_vm-subs-satellite[Modifying a virt-who Configuration] in _Configuring Virtual Machine Subscriptions in {ProjectName}_.
+For more information, see {MultiBaseURL}/configuring_virtual_machine_subscriptions_in_red_hat_satellite/troubleshooting-virt-who#modifying-virt-who-configuration_vm-subs-satellite[Modifying a virt-who Configuration] in _Configuring Virtual Machine Subscriptions in {ProjectName}_.
```
lzap, line 13: In Foreman/Katello context, this does get rendered as a link to RH Portal, is this intended? I would expect linking the upstream guide instead.

In other words, I think the best solution would be also to override `MultiBaseURL` in non-Satellite context with `BaseURL`. Upstream there is no multi and single-page we just have one version for all.

guides/common/attributes.adoc
```diff
@@ -15,6 +15,9 @@
 
 // Use current RH Satellite release version for links to access.redhat.com
 :AccessRedHatComVersion: 6.10
+// Red Hat sometimes prefers the multi page (as opposed to html-single)
+// The URL structure is different and there's no upstream equivalent.
+:MultiBaseURL: https://access.redhat.com/documentation/en-us/red_hat_satellite/{AccessRedHatComVersion}/html/
```
lzap, line 20: Can you move this few lines below to the section where we also define the `BaseURL`? I suggest also to override it for non-satellite context to be actually `BaseURL` so it works seamlessly.

lzap (changes requested):
Generally I am good, but there is one nitpick about non-satellite context.

We could also additionally define Multi versions of all guide URLs to prevent repeating ourselves even more, but this is not required for this PR:

```
:AdministeringMultiDocURL: {BaseURL}administering_red_hat_satellite/index#
:ConfiguringAnsibleMultiDocURL: {BaseURL}configuring_satellite_to_use_ansible/index#
:ConfiguringLoadBalancerMultiDocURL:  ...
```

ekohl:
> We could also additionally define Multi versions of all guide URLs to prevent repeating ourselves even more, but this is not required for this PR:

That's the ideal case, but https://github.com/theforeman/foreman-documentation/pull/856#issuecomment-976393531 explains why this can't be done trivially. That's why I started with this approach as a first step.

guides/doc-Administering_Red_Hat_Satellite/topics/Renaming_a_Satellite_Server.adoc
ekohl: > In Foreman/Katello context, this does get rendered as a link to RH Portal, is this intended? I would expect linking the upstream guide instead.

I would expect the same thing, but note that it was already there prior to this PR. I don't have a good solution to this now but you're completely right that we should link to an upstream document.

Looking at the documentation, we may be able to use a custom macro. https://dzone.com/articles/asciidoctor-creating-a-macro describes how it works with asciidoctor and https://asciidoc-py.github.io/userguide.html#_macros has more. However, this may not work with the other systems (AFAIK Orcharhino uses Antora and RH uses something else).

> In other words, I think the best solution would be also to override MultiBaseURL in non-Satellite context with BaseURL. Upstream there is no multi and single-page we just have one version for all.

I agree, but I'm not sure how to technically achieve that now.

guides/common/attributes.adoc
ekohl: > Can you move this few lines below to the section where we also define the BaseURL?

Sure

> I suggest also to override it for non-satellite context to be actually BaseURL so it works seamlessly.

I don't know which value.

ekohl:
Rebased to resolve conflicts and moved the attribute a few down line per https://github.com/theforeman/foreman-documentation/pull/856#discussion_r773865895.

guides/doc-Administering_Red_Hat_Satellite/topics/Renaming_a_Satellite_Server.adoc
maximiliankolb: My thoughts: Let's not overcomplicate it for now with a macro if we can use an attribute instead. On the other hand, if it's only for links to access.redhat.com, I am fine because those should only be visible in Satellite builds anyway.

But yes, please keep the downstream solutions Antora and pantheon in mind.

ekohl: AFAIK most guides have upstream equivalents. We just don't build the multi guide. For example:

```
guides/doc-Configuring_Load_Balancer/topics/Before_You_Begin.adoc
29:To upgrade {SmartProxyServer}s from {ProductVersionPrevious} to {ProductVersion}, complete the {MultiBaseURL}/upgrading_and_updating_red_hat_satellite/upgrading_red_hat_satellite#upgrading_capsule_server[Upgrading {SmartProxyServer}s] procedure in _Upgrading and Updating {ProjectName}_.
```

This refers to what we build from `guides/doc-Upgrading_and_Updating` but then a specific chapter of it. So essentially what you have is:
```asciidoc
ifdef::satellite[]
To upgrade {SmartProxyServer}s from {ProductVersionPrevious} to {ProductVersion}, complete the {MultiBaseURL}/upgrading_and_updating_red_hat_satellite/upgrading_red_hat_satellite#upgrading_capsule_server[Upgrading {SmartProxyServer}s] procedure in _Upgrading and Updating {ProjectName}_.
endif::[]
ifndef::satellite[]
To upgrade {SmartProxyServer}s from {ProductVersionPrevious} to {ProductVersion}, complete the {UpgradingDocURL}#upgrading_capsule_server[Upgrading {SmartProxyServer}s] procedure in _Upgrading and Updating {ProjectName}_.
endif::[]
```

Or in short: `{MultiBaseURL}/upgrading_and_updating_red_hat_satellite/upgrading_red_hat_satellite#upgrading_capsule_server` vs `{UpgradingDocURL}#upgrading_capsule_server`.

Then when you take in mind that `UpgradingDocURL` is defined as:
```
guides/common/attributes-base.adoc
14::UpgradingDocURL: {BaseURL}Upgrading_and_Updating/{BaseFilenameURL}#

12::UpgradingDocURL: {BaseURL}upgrading_and_updating_red_hat_satellite/index#
```

You see that it's even a smaller difference. It's really the base URL and `upgrading_red_hat_satellite` vs `{BaseFilenameURL}` (you could say `index` is `{BaseFilenameURL}` for satellite).

maximiliankolb: I thought this PR was purely about links to docs that are no (yet) part of upstream.

Of course we should keep our attributes the way they are; i.e. one attribute per guide. You can decide to use single vs multi per build, e.g. in Satellite attributes.

The way it is right now if IMHO great because it's short and you only need to know which guide it is in.

ekohl: > I thought this PR was purely about links to docs that are no (yet) part of upstream.

In the beginning I thought so too but over time I found out the real issues. For example, 3daa833e3381c95517de9594d634c9f0d26b4b93 shows that things never linked to the upstream guide that was already there. Then there are still a bunch of links it doesn't replace since I don't know how to deal with multi vs single.

For example, `guides/doc-Planning_Guide/topics/Capsule_Server_Overview.adoc` has:
```asciidoc
ifdef::satellite[]
The {InstallingProjectDocURL}satellite-ports-and-firewalls-requirements_{project-context}[Ports and Firewalls Requirements] in _{project-installation-guide-title}_, https://access.redhat.com/documentation/en-us/red_hat_satellite/{AccessRedHatComVersion}/html/installing_satellite_server_from_a_disconnected_network/preparing-environment-for-satellite-installation#satellite-ports-and-firewalls-requirements_{project-context}[Ports and Firewalls Requirements] in _Installing {ProjectServer} from a Disconnected Network_ and {InstallingSmartProxyDocURL}capsule-ports-and-firewalls-requirements_{smart-proxy-context}[Ports and Firewalls Requirements] in _{smart-proxy-installation-guide-title}_ contain complete instructions for configuring the host-based firewall to open the ports required.

A matrix table of ports is available in the Red{nbsp}Hat Knowledgebase solution https://access.redhat.com/solutions/5627751[Red Hat Satellite List of Network Ports].
endif::[]
```

IMHO this is wrong since it's not Satellite specific and all guides that it mentions are upstream. The matrix of ports was actually introduced upstream in #761 and the link to the knowledge base that remains is probably an oversight.

However, we don't have an attribute that can replace `https://access.redhat.com/documentation/en-us/red_hat_satellite/{AccessRedHatComVersion}/html/installing_satellite_server_from_a_disconnected_network/preparing-environment-for-satellite-installation#satellite-ports-and-firewalls-requirements_{project-context}`. You can end up with an ugly solution such as https://github.com/theforeman/foreman-documentation/pull/924 but this may not work for sentences that can't be split up.

What really needs to be done is to create a tracker to further eliminate all links to downstream documentation in upstream and figure out a solution. However, I won't have time to work on that.

ShwetaNaresh:
Thank you all for contributing to this effort for a better customer experience upstream and downstream.
I am a bit less qualified right now to visualize a bigger picture or pros and cons of BaseURL | html | html-single | html-multi. I agree that based on the context, in certain cases, html-single can be a better alternative over html-multi. 
I am fine with whatever helps better customer experience.
A couple of Qs though: 
- Can we not use BaseURL both upstream and downstream?
- @mdolezelova will BaseURL not work for DDFs? To make the URL usage consistent across upstream and downstream, if we have to update the sources, how much impact will it have in terms of resource capacity and time?

mdolezelova:
Hi @ShwetaNaresh , yes, DDF currently works for html-multi only (I can't tell how it will look like in the future with the Jupiter tool), that's why we direct customers to html-multi. Also, for huge documents, html-single is not the best experience because rendering the doc takes  long time.  I don't know how much time exactly would that need to adapt to this. If my understanding is correct, it can be done fully automatically, so no additional efforts needed for the Satellite repo. But I'm not sure whether the change applied retroactively (in the already release docs) or would apply for the upcoming release onward. @ekohl please could you clarify this for us? Thank you!

ShwetaNaresh:
Thanks, @mdolezelova. I understand we aren't yet sure what changes Jupiter would demand. Nice to know that we can automate the changes required if any. 
If we want to have similar URL usage for Satellite and non-Satellite content, do you all think it would be a good idea to weigh the pros and cons in terms of customer experience for BaseURL | html | html-single | html-multi, and then finalize the option? 
If using attributes for Satellite and non-Satellite content is a quick fix that does not impact the customer experience, I am fine with that too. But at the same time, IMO, keeping it simple for the writers will avoid human error.

guides/common/attributes.adoc
lzap: My thinking was to set `MultiBaseURL` to something like `https://docs.theforeman.org/{Version}/` so it is rendered correctly in non-Satellite contexts too.

ekohl: I think that will result into 404s.

From the example in https://github.com/theforeman/foreman-documentation/pull/856#discussion_r773898246:
```
{MultiBaseURL}/upgrading_and_updating_red_hat_satellite/upgrading_red_hat_satellite
```

This would then result in:
```
https://docs.theforeman.org/{Version}/upgrading_and_updating_red_hat_satellite/upgrading_red_hat_satellite
```

However, we only have:
```
https://docs.theforeman.org/{Version}/upgrading_and_updating_red_hat_satellite/index-foreman-el.html
https://docs.theforeman.org/{Version}/upgrading_and_updating_red_hat_satellite/index-foreman-deb.html
https://docs.theforeman.org/{Version}/upgrading_and_updating_red_hat_satellite/index-katello.html
```

We also don't have rewrite rules so it's not like we can configure Apache to solve it for us. We already use fragments so can't solve it there either.

lzap:
The GHA is so cool :)

lzap (approved):

ekohl:
Oh yes, the diff actually fixes the broken links. I'll wait and rebase on #1585.

maximiliankolb (approved):
I haven't checked all the links manually, but your replacements look reasonable.

ekohl:
Rebased now that #1585 has been merged.

ekohl:
3.2 had a merge conflict, but I solved that and pushed it:

   a3cae3bb8..8be97bce0  3.2-pick-more-attributes -> 3.2
   36d80f747..cc8df619c  3.3-pick-more-attributes -> 3.3
   9497c8b53..c039da19e  3.4-pick-more-attributes -> 3.4

3.1 also had a merge conflict but to be safe I opened a PR for that: https://github.com/theforeman/foreman-documentation/pull/856

PR #858, opened 2021-11-16
Pulpcore database in PostgreSQL should have the 'pulp' user

wbclark:
Cherry-pick into:

* [x] Foreman 3.0

### DIFF
--- a/guides/common/modules/proc_installing-postgresql.adoc
+++ b/guides/common/modules/proc_installing-postgresql.adoc
@@ -81,10 +81,10 @@ $ su - postgres -c psql
 ----
 CREATE USER "foreman" WITH PASSWORD '_Foreman_Password_';
 CREATE USER "candlepin" WITH PASSWORD '_Candlepin_Password_';
-CREATE USER "pulpcore" WITH PASSWORD '_Pulpcore_Password_';
+CREATE USER "pulp" WITH PASSWORD '_Pulpcore_Password_';
 CREATE DATABASE foreman OWNER foreman;
 CREATE DATABASE candlepin OWNER candlepin;
-CREATE DATABASE pulpcore OWNER pulpcore;
+CREATE DATABASE pulpcore OWNER pulp;
 ----
 +
 . Exit the `postgres` user:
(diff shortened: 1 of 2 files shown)


wbclark:
For reference: https://github.com/theforeman/puppet-foreman_proxy_content/blob/76e2a6852d1d2ca33935ccf8a6ab69992c32ec1d/manifests/init.pp#L132

ekohl (approved):
It's good to match the defaults, but should it explicitly set `--foreman-proxy-content-pulpcore-postgresql-user pulp` as well to indicate it can be changed? We already set `...-db-name` even though it matches the default. On the other hand, we don't set the user for other systems. This is inconsistent.

Still :+1: for this as a starter.

wbclark:
> It's good to match the defaults, but should it explicitly set `--foreman-proxy-content-pulpcore-postgresql-user pulp` as well to indicate it can be changed? We already set `...-db-name` even though it matches the default. On the other hand, we don't set the user for other systems. This is inconsistent.
> 
> Still +1 for this as a starter.

I could see an argument either way. It's good to be comprehensive and inform the user that the option exists. On the other hand, I can't really conceive of a reason why it would need to be changed ever.

If you have strong feelings that it should be included, I could update this PR to add that as well. Let me know what you think @ekohl

ekohl:
No strong feelings from my side. As I said: :+1: for this. If we do want to include username, it should be done for all 3 (foreman, candlepin and pulp) which should likely be its own PR.

evgeni:
+1 on explicitly calling out the username in the installer call for all DBs as a separate PR

ehelms:
Merging this as the resolution appears to be open a new PR with the other changes. I will leave that to you @wbclark to track it

ekohl:
Don't forget cherry picks either.

wbclark:
PRs for cherry picking:

- [ ] 3.0 - https://github.com/theforeman/foreman-documentation/pull/866
- [ ] 2.5 - https://github.com/theforeman/foreman-documentation/pull/867

PR #863, opened 2021-11-17
Add "In the {ProjectWebUI}" to each instance of "Navigate to X > Y"

maximiliankolb:
Replace `Navigate to` with `In the {ProjectWebUI}, navigate to`.


lzap (approved):
Looks sane. Looks like you are focusing very much on consistency and style, feel free to update our README with the most up to date information on style. It has some random thoughts at the moment, more formal guideline would be appreciated.

PR #846, opened 2021-11-15
BZ 1996571: Note the default to destroy associated VM on host delete

Lennonka:
[BZ1996571](https://bugzilla.redhat.com/show_bug.cgi?id=1996571)

Cherry-pick into:

* [x] Foreman 3.0
* For Foreman 2.3 and 2.5, separate PR requests will be filed

PR #869, opened 2021-11-18
BZ 1996571: Note the default to destroy associated VM on host delete on 2.3

Lennonka:
Follow-up to #846


maximiliankolb:
IMHO we don't backport to 2.3 any more.

PR #871, opened 2021-11-22
Add two procedures for Puppet

maximiliankolb:
This PR add two procedures to "using Puppet for configuration management".


melcorr (approved):
Thank you for this!

PR #875, opened 2021-11-23
Fix title and an anchor in upgrading guide on 2.5

Lennonka:
No cherry picks.
This may be an intermediate fix that I need to verify in our internal Pantheon build and I don't have another way of verification than after merge. Unless we make local builds with bccutil work. Sorry!


lzap:
Just checking, was this closed on purpose? Thanks!

PR #877, opened 2021-11-23
Fix attribute used in title in Upgrading Guide

Lennonka:
No cherry pick.


melcorr:
So did that fix the issue as we discussed yesterday?

PR #888, opened 2021-11-30
Fix spelling of separate

maximiliankolb:
`git blame`: @maximiliankolb  🙈


melcorr:
Well, I reviewed this, so I am to blame. 
Thanks for the fix.

lzap:
I have a work-in-progress script that runs all adoc files through `aspell` spelling program, but I do not want to rip things apart now. I can finish the script if anyone is interested running it and doing the actual work tho, many words need to be put into the ignore list :-)

Or maybe check - spellers are not recommended by the IBM Bible, then I will vanish off with this.

PR #902, opened 2021-12-13
Update global reg procedure text

swadeley:
Hello

for Sat7, minor text change as the menu is now greyed to prevent this bug:

Bug 2014251 - Global Registration: Selecting Satellite URL as the proxy fails to register hosts with default config

Using the internal Capsule is the default option, as it always was, its just that if you explicitly selected it you would have port 9090  appended. 

Thank you

Cherry-pick into:

* [ ] Foreman 3.0


melcorr:
Thank you @swadeley

melcorr (approved):

PR #905, opened 2021-12-15
Add content filter for APT content & various minor improvements

maximiliankolb:
This PR adds a procedure to use content filters for APT content & contains various other smaller fixes and improvements.

### DIFF
--- a/guides/common/modules/proc_creating-a-content-filter-for-yum-content.adoc
+++ b/guides/common/modules/proc_creating-a-content-filter-for-yum-content.adoc
@@ -1,8 +1,8 @@
-[id="Creating_a_Content_Filter_{context}"]
-= Creating a Content Filter
+[id="Creating_a_Content_Filter_for_Yum_Content_{context}"]
+= Creating a Content Filter for Yum Content
 
-Use this procedure to create a content filter.
-To use the CLI instead of the {ProjectWebUI}, see the xref:cli-creating-a-content-filter_{context}[].
+Use this procedure to create a content filter for Yum content.
+To use the CLI instead of the {ProjectWebUI}, see the xref:cli-creating-a-content-filter-yum_{context}[].
 
 For examples of how to build a filter, see xref:Content_Filter_Examples_{context}[].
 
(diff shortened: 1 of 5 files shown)


guides/common/modules/proc_creating-a-content-filter-for-yum-content.adoc
```diff
@@ -1,8 +1,8 @@
-[id="Creating_a_Content_Filter_{context}"]
-= Creating a Content Filter
+[id="Creating_a_Content_Filter_for_YUM_Content_{context}"]
+= Creating a Content Filter for YUM Content
```
melcorr, line 2: This might raise some eyebrows for Satellite folks. Can we leave the ID but do this... 
```suggestion
ifdef::satellite[]
= Creating a Content Filter
endif::[]
ifndef::satellite[]
= Creating a Content Filter for YUM Content
endif::[]
```

guides/common/modules/proc_creating-a-content-filter-for-yum-content.adoc
```diff
@@ -1,8 +1,8 @@
...
 
-Use this procedure to create a content filter.
-To use the CLI instead of the {ProjectWebUI}, see the xref:cli-creating-a-content-filter_{context}[].
+Use this procedure to create a content filter for YUM content.
```
melcorr, line 4:
```suggestion
Use this procedure to create a content filter for Yum content.
```
I think we chatted about this before. The capitalization gives the impression it is an acronym, which it isn't. The standard is `Yum`

melcorr (changes requested):
I build and read. Looks fine. 
I've one request regarding the change in the title, and one small comment re formatting of Yum.

guides/common/modules/proc_creating-a-content-filter-for-yum-content.adoc
maximiliankolb: You can also have content filters for container images as far as I can see.

melcorr: Ahh, if that's the case, you can ignore this, if you just s/YUM/Yum :)

guides/common/modules/proc_creating-a-content-filter-for-yum-content.adoc
maximiliankolb: Yes, now I remember. Applied.

guides/common/modules/proc_creating-a-content-filter-for-yum-content.adoc
maximiliankolb: That has been fixed, yes.

maximiliankolb:
Rebased to master and ready for re-review.

PR #906, opened 2021-12-15
Add procedure to clone existing host & minor improvements

maximiliankolb:
Hey.
This PR adds a procedure to clone existing hosts and contains various other minor improvements and fixes.


guides/common/modules/proc_cloning-hosts.adoc
```diff
@@ -0,0 +1,13 @@
+[id="Cloning_Hosts_{context}"]
+= Cloning Hosts
+
+You can clone existing hosts.
```
melcorr, line 4: Some might say that here we should explain **Why** someone would want to do this.

You can clone hosts to ...

Not a blocker though :)

melcorr (approved):
One non-blocking comment. 
You can merge without implementing it if you like!
This reads well!

guides/common/modules/proc_cloning-hosts.adoc
maximiliankolb: I cannot come up with something useful. We can fix this in the future.

PR #910, opened 2021-12-16
Temporary disable link checker

maximiliankolb:
cc @melcorr 

If this does not work, we should revert the last commit to master. my bad!

### DIFF
--- a/guides/Makefile
+++ b/guides/Makefile
@@ -13,7 +13,8 @@ pdf: all-pdf
 clean: all-clean
 
 linkchecker:
-	find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern
+	echo "do not check links"
+	# find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern
 
 linkchecker-tryer:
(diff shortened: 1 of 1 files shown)


melcorr (approved):
let's go. 
I'll revert if necessary.

maximiliankolb:
I've missed on item in the Makefile. Give me another chance! ;)

melcorr:
oh damn

melcorr:
@maximiliankolb - do you use IRC at all? I feel we should have some faster way of pinging if necessary.

maximiliankolb:
> @maximiliankolb - do you use IRC at all? I feel we should have some faster way of pinging if necessary.

Maybe I should? It's just so weird that you have to stay connected all the time. I'll think about it. Right now, mail is your best chance. I also use the Github app, so you can literally reach me any time.

melcorr:
> > @maximiliankolb - do you use IRC at all? I feel we should have some faster way of pinging if necessary.
> 
> Maybe I should? It's just so weird that you have to stay connected all the time. I'll think about it. Right now, mail is your best chance. I also use the Github app, so you can literally reach me any time.

I use Matrix and sign into Libera through that. 
It's no problem really - just at times like this it would have been handier!

PR #913, opened 2021-12-16
Removed Self-Registered Satellites section after confirming with Suppo…

bangelic:
…rt that the link SATDOC-505 requested to be updated was an outdated workflow.

Cherry-pick into:

* [ ] Foreman 3.0


melcorr (changes requested):
Please remove all the text rather than comment out. 
We don't really comment anything out in this doc set. 
We'll never need this again so it is safe to do so.

melcorr (approved):
Great!

PR #916, opened 2021-12-17
Fix spelling of mutually

maximiliankolb:
git blame @maximiliankolb  🤷


maximiliankolb:
Took the liberty to just merge this.

PR #920, opened 2021-12-20
Restore linkchecker

melcorr:
We temporarily disabled linkchecker because adding a new guide
causes the linkchecker to fail the new builds.
The new guide builds just fine, so now re-enabling our checker.

Cherry-pick into:

* [ ] Foreman 3.0

### DIFF
--- a/guides/Makefile
+++ b/guides/Makefile
@@ -13,12 +13,10 @@ pdf: all-pdf
 clean: all-clean
 
 linkchecker:
-	echo "do not check links"
-	# find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern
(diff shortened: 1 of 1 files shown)


lzap:
This appears to use two tabs instead of one? Can you use just one?

Keep in mind this is a Makefile, you MUST use tabs and not spaces.

melcorr:
> This appears to use two tabs instead of one? Can you use just one?
> 
> Keep in mind this is a Makefile, you MUST use tabs and not spaces.

I'll try with one! I just restored to what it looked like before we cancelled it out.

lzap (approved):
Thanks, after this commit I am branching 3.1.

maximiliankolb:
Thanks for restoring!

PR #921, opened 2021-12-21
Always use the build to refer to documents

ekohl:
This unifies the linking structure and prevents documents from linking back to foreman-el instructions when using Debian or Katello.


maximiliankolb (approved):
This sounds like a useful change. :+1:

lzap:
Thanks merged.

PR #923, opened 2021-12-21
Update support information on bootdisks

lzap:
There are two PXE less methods in Satellite: one is bootdisk which was limited in 6.10 (only supported bootdisks are Full Host and Subnet disk), that is the change that was supposed to be made. Then we have Discovery PXE less which is still present in Satellite, however, it is technical preview and the feature is not leaving this state until it is fully removed and replaced by something more reliable.

This PR fixes what was introduced in

https://github.com/theforeman/foreman-documentation/pull/794

I should have been more careful when doing the review.

Cherry-pick into:

* [x] Foreman 3.1


melcorr (approved):
LGTM

melcorr:
041d7021..42b858c5  3.1 -> 3.1

PR #932, opened 2022-01-03
Added missing ifdef for Satellite

mdolezelova:
Cherry-pick into:

* [ ] Foreman 3.1


mdolezelova:
Hi @lzap , in the previous MR I included wrong syntax for one ifdef, which made half of the Provisioning guide for Satellite disappear. I've fixed it right now. Please, could you review and merge as quick as possible so that customers are not exposed to this issue for a long time? Apologies.

lzap (approved):
My bad. Thanks.

Are you going to file respective changes into master as well? Thanks.

mdolezelova:
@lzap Thank you! On master, everything is already set correctly. The point was that the change had been done only on master, and not on 2.5, so it was not visible in Satellite 6.10 docs, so I've created the previous PR to get it on the branch 2.5, but I made a mistake in the ifdef syntax. So now, everything should be correct. I just need to wait for the change to propagate to the downstream repo and republish the Provisioning guide for Satellite 6.10 afterwards. Thank you!

lzap:
I see, thanks for clarification.

PR #933, opened 2022-01-04
Add Puppet disablement procedure (unplaced)

Lennonka:
Disabling Puppet -- currently unincluded, but will be included only for Satellite, Katello and orcharhino in future.

Cherry-pick into:

* [x] Foreman 3.1

### DIFF
--- a/guides/common/modules/proc_disabling-puppet.adoc
+++ b/guides/common/modules/proc_disabling-puppet.adoc
@@ -1,0 +1,24 @@
+[id="Disabling_Puppet_Integration_{context}"]
+= Disabling Puppet Integration with {Project}
+
+To discontinue using Puppet in your {Project}, follow this procedure.
+
+.Prerequisite
+* Puppet is enabled on {Project}.
+
+.Procedure
+. Disable Puppet server on all external {SmartProxies} before you disable it on {ProjectServer}:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foreman-maintain} plugin purge-puppet
+----
+. Disable Puppet server on {ProjectServer}:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foreman-maintain} plugin purge-puppet [--remove-all-data]
+----
++
+The command without arguments will remove all Puppet-related data in {Project}.
+To remove Puppet server data files, including Puppet environments, add the `--remove-all-data` argument to the command.


guides/common/modules/proc_disabling-puppet.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+.Procedure
+. Disable Puppet server on {SmartProxies} before you disable it on {ProjectServer}:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foreman-maintain} plugin purge-puppet
+----
++
+If your deployment includes a database on {SmartProxy}, the command will also remove internal data from the Puppet database.
```
ekohl, line 17: I don't think database on SmartProxy is really a thing. While there is PuppetDB, it's an entirely optional project and we don't support it. The removal procedure thus doesn't remove PuppetDB.

It really removes `/etc/puppetlabs` and `/opt/puppetlabs/server/data`. So in the phrasing I'd keep it to `data`, not `database`.

However, looking at the code now it also calls `foreman-rake purge:puppet` which certainly does NOT work on a SmartProxy so right now it can't work on a SmarProxy.

We'd need to look into that but for now I think this line should be dropped:
```suggestion
```

guides/common/modules/proc_disabling-puppet.adoc
```diff
@@ -0,0 +1,26 @@
...
+. Disable Puppet server on {ProjectServer}:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foreman-maintain} plugin purge-puppet [--remove-all-data]
+----
++
+The command without arguments will also remove internal data from the Puppet database on {ProjectServer}.
+To remove external Puppet data as well, add the `--remove-all-data` argument to the command.
```
ekohl, line 26: This is a bit vague to me. Again because Puppet database makes me (as a Puppet user) think of PuppetDB.

@ezr-ondrej how can we phrase this correctly?

guides/common/modules/proc_disabling-puppet.adoc
```diff
@@ -0,0 +1,26 @@
...
++
+If your deployment includes a database on {SmartProxy}, the command will also remove internal data from the Puppet database.
+. Disable Puppet server on {ProjectServer}:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foreman-maintain} plugin purge-puppet [--remove-all-data]
+----
++
+The command without arguments will also remove internal data from the Puppet database on {ProjectServer}.
```
ezr-ondrej, line 25:
```suggestion
The command without arguments will remove all internal Puppet data on {ProjectServer}.
```
What about this @ekohl ?

guides/common/modules/proc_disabling-puppet.adoc
```diff
@@ -0,0 +1,26 @@
+[id="Disabling_Puppet_Integration_{context}"]
+= Disabling Puppet Integration with {Project}
+
+To discontinue using Puppet in your {Project}, follow this procedure.
+
+.Prerequisites
+* You upgraded {Project} from the previous version with Puppet server enabled.
+
+.Procedure
+. Disable Puppet server on {SmartProxies} before you disable it on {ProjectServer}:
```
ezr-ondrej, line 10: Maybe we should be specific here, someone might realize there is also an internal Smart proxy and might want to try disabling it on that one :)
```suggestion
. Disable Puppet server on external {SmartProxies} before you disable it on {ProjectServer}:
```

guides/common/modules/proc_disabling-puppet.adoc
ezr-ondrej: I agree, the command will not remove the database, because there is no "internal" data (as in internal to Foreman/Satellite) held here.

ezr-ondrej (commented):
I think it's all we need. Simple, but on point :)

If Ewoud is right and the command does not work on smart proxy, we should fix the command not the docs.

guides/common/modules/proc_disabling-puppet.adoc
Lennonka: Thank you both! Dropping line 17.

guides/common/modules/proc_disabling-puppet.adoc
Lennonka: Rephrased.

guides/common/modules/proc_disabling-puppet.adoc
Lennonka: Changed, thx

guides/common/modules/proc_disabling-puppet.adoc
Lennonka: Changed.

Lennonka:
Your suggestions incorporated :) How about now?

guides/common/modules/proc_disabling-puppet.adoc
```diff
@@ -0,0 +1,24 @@
+[id="Disabling_Puppet_Integration_{context}"]
+= Disabling Puppet Integration with {Project}
+
+To discontinue using Puppet in your {Project}, follow this procedure.
+
+.Prerequisites
```
maximiliankolb, line 6:
```suggestion
.Prerequisite
```

maximiliankolb (approved):
One small nitpick; the rest LGTM.

guides/common/modules/proc_disabling-puppet.adoc
```diff
@@ -0,0 +1,24 @@
...
+* You upgraded {Project} from the previous version with Puppet server enabled.
```
ekohl, line 7: Do you need to have upgraded from a previous version? Another path to get here is to install Satellite 7.0 without Puppet and then follow the enabling procedure. I'm not sure how to best phrase it but essentially it can be generalized as "Puppet is enabled on {Project}".

guides/common/modules/proc_disabling-puppet.adoc
```diff
@@ -0,0 +1,24 @@
...
+
+.Procedure
+. Disable Puppet server on external {SmartProxies} before you disable it on {ProjectServer}:
```
ekohl, line 10: Perhaps emphasize it's on all? Though we have coded a safe guard, it doesn't hurt to be explicit.
```suggestion
. Disable Puppet server on all external {SmartProxies} before you disable it on {ProjectServer}:
```

guides/common/modules/proc_disabling-puppet.adoc
```diff
@@ -0,0 +1,24 @@
...
+----
+. Disable Puppet server on {ProjectServer}:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foreman-maintain} plugin purge-puppet [--remove-all-data]
+----
++
+The command without arguments will remove all internal Puppet data on {ProjectServer}.
+To remove external Puppet data as well, add the `--remove-all-data` argument to the command.
```
ekohl, line 24: I'm not sure about the internal vs external data. Perhaps instead of internal we should say "Puppet related data in {Project}" and instead of external data "Puppetserver data files, including Puppet environments"

Lennonka:
This should be ready for merge.

ekohl:
Post merge ACK here.

maximiliankolb:
Cherry-pick: https://github.com/theforeman/foreman-documentation/pull/948 ✅

PR #935, opened 2022-01-04
Update version numbers and branching instructions

ekohl:
The branching instructions now first happen by creating the branch. This means the guides will be available by the time the index is created. It also explicitly mentions some steps that were previously missed.

The Satellite versions are updated and now grouped together. It reuses macros to reduce duplication.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,15 +37,17 @@ When a commit is pushed into `X.Y`:
 
 ## Branching new release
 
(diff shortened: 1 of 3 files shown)


maximiliankolb (approved):
LGTM!

ekohl:
Can you merge it then?

PR #940, opened 2022-01-10
Fix spelling

maximiliankolb:
This PR fixes three spelling mistakes: `Complicance`, `Paramter`, and `Uplodaing`.

Cherry-pick into:

* [x] Foreman 3.1

### DIFF
--- a/guides/common/modules/proc_creating-a-compliance-policy.adoc
+++ b/guides/common/modules/proc_creating-a-compliance-policy.adoc
@@ -1,4 +1,4 @@
-[id="Creating_a_Complicance_Policy_{context}"]
+[id="Creating_a_Compliance_Policy_{context}"]
 = Creating a Compliance Policy
 
 With {Project}, you can create a compliance policy to scan your content hosts to ensure that the hosts remain compliant to your security requirements.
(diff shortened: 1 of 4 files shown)


guides/common/modules/proc_creating-a-compliance-policy.adoc
```diff
@@ -1,4 +1,4 @@
-[id="Creating_a_Complicance_Policy_{context}"]
+[id="Creating_a_Compliance_Policy_{context}"]
```
ekohl, line 1: Should the old ID stick around so links aren't broken or will that create problems?

maximiliankolb: No, we never keep old IDs around just for the sake for links. We always ensure master is in a working state; which it is according to the GHA pipeline. I also did not find any broken xrefs by hand.

ekohl: I suppose it's OK if you don't cherry pick it. For upstream I'd also say it's safe for now since the documentation is still experimental and we're not too concerned with stability. xrefs are also not too critical since users still end up on the right page. And for Red Hat downstream 3.1 is still considered to be under development so picking to 3.1 is fine.

ekohl (approved):

maximiliankolb:
cherry-picked to 3.1 ✅

PR #945, opened 2022-01-10
Modularize administering foreman guide

maximiliankolb:
This PR moves content from an assembly to an included file. No links are broken & no content is changed.

Cherry-pick into:

* [x] Foreman 3.1


Lennonka (commented):
Is the include working correctly without a level offset? Otherwise LGTM.

maximiliankolb:
> Is the include working correctly without a level offset? Otherwise LGTM.

Yes, because the level of the heading does not need to be changed.

The offset is only necessary when you have content like `== my second level` and move it to a new file; then you'd go with `include::my-file.adoc[leveloffset=+1]` and `= my second level` as its first line.

maximiliankolb:
Cherry-picked: https://github.com/theforeman/foreman-documentation/pull/947 ✅

PR #964, opened 2022-01-17
Update to Ruby 3.1 in building

ekohl:
Nanoc has started to require Psych 3.1 or newer. This is shipped in Ruby 3. Ruby 3.1 actually has 4.0 and nanoc master has started to require that.


lzap:
All green except the HTML validator, we can fix this later.

PR #966, opened 2022-01-17
Fix OpenSCAP Workbench link error

lzap:
* [ ] Foreman 3.1

### DIFF
--- a/guides/common/modules/con_tailoring-files.adoc
+++ b/guides/common/modules/con_tailoring-files.adoc
@@ -4,5 +4,5 @@
 Tailoring Files allow existing OpenSCAP policies to be customized without forking or rewriting the policy.
 You can assign a Tailoring File to a policy when creating or updating a policy.
 
-You can create a Tailoring File using the https://www.open-scap.org/tools/scap-workbench/[SCAP Workbench].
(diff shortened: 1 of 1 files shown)


guides/common/modules/con_tailoring-files.adoc
```diff
@@ -4,5 +4,5 @@
 Tailoring Files allow existing OpenSCAP policies to be customized without forking or rewriting the policy.
 You can assign a Tailoring File to a policy when creating or updating a policy.
 
-You can create a Tailoring File using the https://www.open-scap.org/tools/scap-workbench/[SCAP Workbench].
+You can create a Tailoring File using the https://www.open-scap.org/tools/[SCAP Workbench].
```
ekohl, line 7: I've opened https://github.com/OpenSCAP/website/issues/210. Should we give it a moment to let them respond?

lzap: I would rather have our CI green, if you clink on it, you will land on a nice page where there is a huge title SCAP Workbench, from there users can work it out somehow.

lzap:
Ok its green, lets fix our CI we can put it back. We could also link: https://github.com/OpenSCAP/scap-workbench but Google will find that site quickly too.

PR #968, opened 2022-01-17
Use three columns menu to separate EL/DEB

lzap:
I think this actually looks cleaner.

Also the landing page where we have a list of guides we can essentially kill now and use that space to explain key concepts:

* What is Foreman
* What platforms are supported
* What is Katello and why it is different

I also made the following changes:

* Added Release Notes missing guide.
* Moved Planning guide to Katello menu (it's not very relevant in Foreman context).

### DIFF
--- a/web/content/js/nav.js
+++ b/web/content/js/nav.js
@@ -19,37 +19,22 @@ let navVersions = [
 
 let navItems = [
 {
-	title: "Foreman documentation", items: [
+	title: "Foreman on EL", items: [
 	{
-		title: "Planning for Foreman",
-		href: "/Planning_Guide/index-foreman-el.html"
+		title: "Release Notes",
+		href: "/Release_notes/index-foreman-el.html"
 	},{
-		title: "Quickstart Guide on RHEL/CentOS",
+		title: "Quickstart Guide",
 		href: "/Quickstart_Guide/index-foreman-el.html"
 	},{
-		title: "Quickstart Guide on Debian/Ubuntu",
-		href: "/Quickstart_Guide/index-foreman-deb.html"
-	},{
-		title: "Installing on RHEL/CentOS",
+		title: "Installing Foreman Server",
 		href: "/Installing_Server_on_Red_Hat/index-foreman-el.html"
 	},{
-		title: "Installing Smart Proxy on RHEL/CentOS",
+		title: "Installing Smart Proxy",
 		href: "/Installing_Proxy_on_Red_Hat/index-foreman-el.html"
-	},{
-		title: "Installing on Debian/Ubuntu",
-		href: "/Installing_Server_on_Debian/index-foreman-deb.html"
-	},{
-		title: "Installing Smart Proxy on Debian/Ubuntu",
-		href: "/Installing_Proxy_on_Debian/index-foreman-deb.html"
-	},{
-		title: "Upgrading and Updating",
-		href: "/Upgrading_and_Updating/index-foreman-el.html"
 	},{
 		title: "Deploying Foreman on AWS",
 		href: "/Deploying_on_AWS/index-foreman-el.html"
-	},{
-		title: "Configuring Smart Proxies with a Load Balancer",
-		href: "/Configuring_Load_Balancer/index-foreman-el.html"
 	},{
 		title: "Provisioning Guide",
 		href: "/Provisioning_Guide/index-foreman-el.html"
@@ -67,22 +52,76 @@ let navItems = [
 		href: "/Application_Centric_Deployment/index-foreman-el.html"
 	}]
 },{
-	title: "Katello documentation", items: [
+	title: "Foreman on Debian", items: [
 	{
-		title: "Quickstart Guide on RHEL/CentOS",
+		title: "Release Notes",
+		href: "/Release_notes/index-foreman-deb.html"
+	},{
+		title: "Quickstart Guide",
+		href: "/Quickstart_Guide/index-foreman-deb.html"
+	},{
+		title: "Installing Foreman Server",
+		href: "/Installing_Server_on_Red_Hat/index-foreman-deb.html"
+	},{
+		title: "Deploying Foreman on AWS",
+		href: "/Deploying_on_AWS/index-foreman-deb.html"
+	},{
+		title: "Provisioning Guide",
+		href: "/Provisioning_Guide/index-foreman-deb.html"
+	},{
+		title: "Configuring Foreman to use Ansible",
+		href: "/Configuring_Ansible/index-foreman-deb.html"
+	},{
+		title: "Administering Foreman",
+		href: "/Administering_Red_Hat_Satellite/index-foreman-deb.html"
+	},{
+		title: "Application Centric Deployment",
+		href: "/Application_Centric_Deployment/index-foreman-deb.html"
+	}]
+},{
+	title: "Katello on EL", items: [
+	{
+		title: "Release Notes",
+		href: "/Release_notes/index-foreman-el.html"
+	},{
+		title: "Planning Guide",
+		href: "/Planning_Guide/index-katello.html"
+	},{
+		title: "Quickstart Guide",
 		href: "/Quickstart_Guide/index-katello.html"
 	},{
-		title: "Installing on RHEL/CentOS",
+		title: "Installing Katello Server",
 		href: "/Installing_Server_on_Red_Hat/index-katello.html"
 	},{
-		title: "Installing Smart Proxy with Content on RHEL/CentOS",
+		title: "Installing Smart Proxy with Content",
 		href: "/Installing_Proxy_on_Red_Hat/index-katello.html"
 	},{
 		title: "Upgrading and Updating",
 		href: "/Upgrading_and_Updating/index-katello.html"
 	},{
+		title: "Configuring Smart Proxies with a Load Balancer",
+		href: "/Configuring_Load_Balancer/index-katello.html"
+	},{
 		title: "Content Management Guide",
 		href: "/Content_Management_Guide/index-katello.html"
+	},{
+		title: "Deploying Foreman on AWS",
+		href: "/Deploying_on_AWS/index-katello.html"
+	},{
+		title: "Provisioning Guide",
+		href: "/Provisioning_Guide/index-katello.html"
+	},{
+		title: "Configuring Foreman to use Ansible",
+		href: "/Configuring_Ansible/index-katello.html"
+	},{
+		title: "Managing Hosts Guide",
+		href: "/Managing_Hosts/index-katello.html"
+	},{
+		title: "Administering Foreman",
+		href: "/Administering_Red_Hat_Satellite/index-katello.html"
+	},{
+		title: "Application Centric Deployment",
+		href: "/Application_Centric_Deployment/index-katello.html"
 	}]
     }
 ]


lzap:
Snímek obrazovky 2022-01-17 v 15 19 28
Snímek obrazovky 2022-01-17 v 15 19 32
Snímek obrazovky 2022-01-17 v 15 19 38
Snímek obrazovky 2022-01-17 v 15 19 43

web/content/js/nav.js
```diff
@@ -67,15 +61,57 @@ let navItems = [
...
+		href: "/Planning_Guide/index-foreman-deb.html"
+	},{
+		title: "Release Notes",
+		href: "/Release_notes/index-foreman-deb.html"
+	},{
+		title: "Quickstart Guide",
+		href: "/Quickstart_Guide/index-foreman-deb.html"
+	},{
+		title: "Installation Guide",
+		href: "/Installing_Server_on_Red_Hat/index-foreman-deb.html"
```
ekohl, line 70: I said it before, but I think we should get rid of the `on_Red_Hat` for this and proxy below. Just merge the two guides and use the flavors we have now.

Also, for me when I build this locally the page is empty?

web/content/js/nav.js
```diff
@@ -67,15 +61,57 @@ let navItems = [
 		href: "/Application_Centric_Deployment/index-foreman-el.html"
 	}]
 },{
-	title: "Katello documentation", items: [
+	title: "Foreman on Debian", items: [
+	{
+		title: "Planning Guide",
+		href: "/Planning_Guide/index-foreman-deb.html"
```
ekohl: https://github.com/theforeman/foreman-documentation/pull/679

lzap:
One more thought - I have my conversion script ready to convert our official guide into Asciidoc.

Now, where and how do we publish the converted manual?

I think a good option would be something like "Complete Guide" that would be enlisted as the first entry in both Foreman on EL and Debian menus. Then we would link it from all the places where we think the new guides are not yet ready for users - most of these issues are in the installation and upgrade Guides.

web/content/js/nav.js
```diff
@@ -67,15 +58,57 @@ let navItems = [
...
 	}]
 },{
-	title: "Katello documentation", items: [
+	title: "Foreman on Debian", items: [
+	{
+		title: "Release Notes",
+		href: "/Release_notes/index-foreman-deb.html"
+	},{
+		title: "Quickstart Guide",
+		href: "/Quickstart_Guide/index-foreman-deb.html"
```
ekohl, line 67: Looks like we need to reflect that Ubuntu 18.04 was dropped in Foreman 3.1.

It would be good to from from Puppet 6 to 7, but both are supported for now.

web/content/js/nav.js
```diff
@@ -67,15 +58,57 @@ let navItems = [
...
+	},{
+		title: "Installation Guide",
+		href: "/Installing_Server_on_Red_Hat/index-foreman-deb.html"
+	},{
+		title: "Installing Smart Proxy",
+		href: "/Installing_Proxy_on_Red_Hat/index-foreman-deb.html"
```
ekohl, line 73: This page also looks empty.

web/content/js/nav.js
```diff
@@ -67,15 +58,57 @@ let navItems = [
...
+	},{
+		title: "Upgrading and Updating",
+		href: "/Upgrading_and_Updating/index-foreman-deb.html"
```
ekohl, line 76: Needs a complete overhaul before this is usable. 100% RPM focused and a lot of Katello specifics.

web/content/js/nav.js
```diff
@@ -67,15 +58,57 @@ let navItems = [
...
+	},{
+		title: "Deploying Foreman on AWS",
+		href: "/Deploying_on_AWS/index-foreman-deb.html"
+	},{
+		title: "Configuring Smart Proxies with a Load Balancer",
+		href: "/Configuring_Load_Balancer/index-foreman-deb.html"
```
ekohl, line 82: Katello specific guide in all the installer commands and certificate handling. I'd recommend to not link this (nor in the foreman-el flavor).

web/content/js/nav.js
```diff
@@ -67,15 +58,57 @@ let navItems = [
...
+	},{
+		title: "Provisioning Guide",
+		href: "/Provisioning_Guide/index-foreman-deb.html"
```
ekohl, line 85: Not related to Debian, but in this guide I noticed it has `\http://smartproxy.example.com:8000/httpboot/ipxe-x64.efi` so something's wrong with the escaping.

web/content/js/nav.js
```diff
@@ -67,15 +58,57 @@ let navItems = [
...
+	},{
+		title: "Configuring Foreman to use Ansible",
+		href: "/Configuring_Ansible/index-foreman-deb.html"
+	},{
+		title: "Managing Hosts Guide",
+		href: "/Managing_Hosts/index-foreman-deb.html"
```
ekohl, line 91: This page looks empty to me. Is that local to me?

web/content/js/nav.js
```diff
@@ -67,15 +58,57 @@ let navItems = [
...
+	},{
+		title: "Administering Foreman",
+		href: "/Administering_Red_Hat_Satellite/index-foreman-deb.html"
```
ekohl, line 94: https://github.com/theforeman/foreman-documentation/issues/676 lists the various issues in this page. For now I'd not link it with Debian since it's not usable.

web/content/js/nav.js
```diff
@@ -67,15 +58,57 @@ let navItems = [
...
+	},{
+		title: "Application Centric Deployment",
+		href: "/Application_Centric_Deployment/index-foreman-deb.html"
```
ekohl, line 97: This renders as empty to me.

ekohl (commented):
I like this as a start, but how do we deal with some guides being ready while others aren't? I left inline comments on various guides. I'm leaning to not linking them until they're ready. We should create issues/PRs with the work that needs to be completed.

web/content/js/nav.js
ekohl: > Looks like we need to reflect that Ubuntu 18.04 was dropped in Foreman 3.1.
https://github.com/theforeman/foreman-documentation/pull/969 does that

web/content/js/nav.js
lzap: Agreed, created a ticket for @melcorr to arrange this downstream, I have a feeling that it might break downstream builds. Once we are all set, I can do the change: https://github.com/theforeman/foreman-documentation/issues/970

web/content/js/nav.js
lzap: I think this guide we should definitely fix as it will be heavily used. Once we convert the original manual, the first change would be to remove quickstart section from there and link these guides (as javascript was originally used). If you can help with this one this would be great.

web/content/js/nav.js
lzap: Yeah, I made a patch that does render an empty page on guides which are not correct.

How bad is Installing Smart Proxy for both EL and Debian? I think we might hide this one for now.

web/content/js/nav.js
lzap: Agreed.

web/content/js/nav.js
lzap: That's what I thought.

web/content/js/nav.js
lzap: I actually moved this only to Katello (after I tipped the screenshots). When this is ready, we can put it back.

web/content/js/nav.js
lzap: I am hiding this for now.

web/content/js/nav.js
ekohl: I see this is indeed outdated. Marking as resolved.

web/content/js/nav.js
lzap: Can I actually move it to Katello? I don't want to completely remove a guide from all menus, in Katello this might be pretty accurate I think.

web/content/js/nav.js
lzap: I will tag @maximiliankolb I am sure he is interested. Let's keep this one, I would like to maintain the Debian version too.

web/content/js/nav.js
lzap: Again, perhaps result of my patch that hides guides which were rendering incorrectly in some contexts. I will remove this Debian menu entry and if @maximiliankolb is interested in bringing it back some time, let's do it!

web/content/js/nav.js
lzap: Since there is a PR opened, I suggest to keep it that way. @melcorr and @maximiliankolb thoughts?

web/content/js/nav.js
ekohl: I had a very quick look through it and nothing jumped out to me as wrong.

web/content/js/nav.js
lzap: @maximiliankolb I would expect this to be a fully working combination? Do you want to fix this? I might have accidentally hidden this, sorry in that case just remove the if statement from the master.adoc:

https://docs.theforeman.org/nightly/Application_Centric_Deployment/index-foreman-deb.html

lzap:
> I'm leaning to not linking them until they're ready.

I agree for some where it's clear they are in really bad shape. But I want every single guide to be linked at least from Katello, which I think the current proposal (including the changes from your review) fulfill.

I implemented the changes:

```diff
[lzap@nuc foreman-documentation]$ gd
diff --git a/web/content/js/nav.js b/web/content/js/nav.js
index 3aa791cd..043e6cc4 100644
--- a/web/content/js/nav.js
+++ b/web/content/js/nav.js
@@ -38,9 +38,6 @@ let navItems = [
        },{
                title: "Deploying Foreman on AWS",
                href: "/Deploying_on_AWS/index-foreman-el.html"
-       },{
-               title: "Configuring Smart Proxies with a Load Balancer",
-               href: "/Configuring_Load_Balancer/index-foreman-el.html"
        },{
                title: "Provisioning Guide",
                href: "/Provisioning_Guide/index-foreman-el.html"
@@ -68,27 +65,15 @@ let navItems = [
        },{
                title: "Installation Guide",
                href: "/Installing_Server_on_Red_Hat/index-foreman-deb.html"
-       },{
-               title: "Installing Smart Proxy",
-               href: "/Installing_Proxy_on_Red_Hat/index-foreman-deb.html"
-       },{
-               title: "Upgrading and Updating",
-               href: "/Upgrading_and_Updating/index-foreman-deb.html"
        },{
                title: "Deploying Foreman on AWS",
                href: "/Deploying_on_AWS/index-foreman-deb.html"
-       },{
-               title: "Configuring Smart Proxies with a Load Balancer",
-               href: "/Configuring_Load_Balancer/index-foreman-deb.html"
        },{
                title: "Provisioning Guide",
                href: "/Provisioning_Guide/index-foreman-deb.html"
        },{
                title: "Configuring Foreman to use Ansible",
                href: "/Configuring_Ansible/index-foreman-deb.html"
-       },{
-               title: "Managing Hosts Guide",
-               href: "/Managing_Hosts/index-foreman-deb.html"
        },{
                title: "Administering Foreman",
                href: "/Administering_Red_Hat_Satellite/index-foreman-deb.html"
@@ -113,6 +98,9 @@ let navItems = [
        },{
                title: "Upgrading and Updating",
                href: "/Upgrading_and_Updating/index-katello.html"
+       },{
+               title: "Configuring Smart Proxies with a Load Balancer",
+               href: "/Configuring_Load_Balancer/index-katello.html"
        },{
                title: "Content Management Guide",
                href: "/Content_Management_Guide/index-katello.html"
```

In conclusion:

* Moved LB to Katello
* Removed Debian menu items: Installing SP, Upgrading, Managing hosts
* Kept Debian items: Provisioning, Administrating (WIP), Application Centric (probably just a bug)

web/content/js/nav.js
ekohl: It needs a bunch of work so I'm good with hiding for now.

ekohl:
I think Katello should list every single guide we have. If they're in foreman-el/foreman-deb, they should also be in Katello. That's because you then get a consistent overview. Also, guides in foreman-el have Katello specific info hidden.

web/content/js/nav.js
maximiliankolb: Trying to fix this: https://github.com/theforeman/foreman-documentation/pull/971

web/content/js/nav.js
maximiliankolb: Running Foreman on Debian/Ubuntu is not currently on my plate. However, I would support making the guides/structure more generic/ disconnected from "only" running on RHEL7/8.

lzap:
> I think Katello should list every single guide we have

Good observation, I kinda forgot about this. Added:

```diff
[lzap@nuc foreman-documentation]$ gd
diff --git a/web/content/js/nav.js b/web/content/js/nav.js
index 3aa791cd..e50070c6 100644
--- a/web/content/js/nav.js
+++ b/web/content/js/nav.js
@@ -38,9 +38,6 @@ let navItems = [
        },{
                title: "Deploying Foreman on AWS",
                href: "/Deploying_on_AWS/index-foreman-el.html"
-       },{
-               title: "Configuring Smart Proxies with a Load Balancer",
-               href: "/Configuring_Load_Balancer/index-foreman-el.html"
        },{
                title: "Provisioning Guide",
                href: "/Provisioning_Guide/index-foreman-el.html"
@@ -68,27 +65,15 @@ let navItems = [
        },{
                title: "Installation Guide",
                href: "/Installing_Server_on_Red_Hat/index-foreman-deb.html"
-       },{
-               title: "Installing Smart Proxy",
-               href: "/Installing_Proxy_on_Red_Hat/index-foreman-deb.html"
-       },{
-               title: "Upgrading and Updating",
-               href: "/Upgrading_and_Updating/index-foreman-deb.html"
        },{
                title: "Deploying Foreman on AWS",
                href: "/Deploying_on_AWS/index-foreman-deb.html"
-       },{
-               title: "Configuring Smart Proxies with a Load Balancer",
-               href: "/Configuring_Load_Balancer/index-foreman-deb.html"
        },{
                title: "Provisioning Guide",
                href: "/Provisioning_Guide/index-foreman-deb.html"
        },{
                title: "Configuring Foreman to use Ansible",
                href: "/Configuring_Ansible/index-foreman-deb.html"
-       },{
-               title: "Managing Hosts Guide",
-               href: "/Managing_Hosts/index-foreman-deb.html"
        },{
                title: "Administering Foreman",
                href: "/Administering_Red_Hat_Satellite/index-foreman-deb.html"
@@ -99,6 +84,9 @@ let navItems = [
 },{
        title: "Katello on EL", items: [
        {
+               title: "Release Notes",
+               href: "/Release_notes/index-foreman-el.html"
+       },{
                title: "Planning Guide",
                href: "/Planning_Guide/index-katello.html"
        },{
@@ -113,9 +101,30 @@ let navItems = [
        },{
                title: "Upgrading and Updating",
                href: "/Upgrading_and_Updating/index-katello.html"
+       },{
+               title: "Configuring Smart Proxies with a Load Balancer",
+               href: "/Configuring_Load_Balancer/index-katello.html"
        },{
                title: "Content Management Guide",
                href: "/Content_Management_Guide/index-katello.html"
+       },{
+               title: "Deploying Foreman on AWS",
+               href: "/Deploying_on_AWS/index-katello.html"
+       },{
+               title: "Provisioning Guide",
+               href: "/Provisioning_Guide/index-katello.html"
+       },{
+               title: "Configuring Foreman to use Ansible",
+               href: "/Configuring_Ansible/index-katello.html"
+       },{
+               title: "Managing Hosts Guide",
+               href: "/Managing_Hosts/index-katello.html"
+       },{
+               title: "Administering Foreman",
+               href: "/Administering_Red_Hat_Satellite/index-katello.html"
+       },{
+               title: "Application Centric Deployment",
+               href: "/Application_Centric_Deployment/index-katello.html"
        }]
     }
 ]
```

lzap:
I think this is ready, @maximiliankolb please comment specifically on the debian related stuff, thanks!

web/content/js/nav.js
maximiliankolb: Merged & cherry-picking to 3.1.

web/content/js/nav.js
maximiliankolb: Fine for me to **not** link it for now. We can improve it in the future and then link it accordingly.

web/content/js/nav.js
maximiliankolb: There are currently no packages to install ACD plugin on Debian/Ubuntu. This may change in the future, but IMHO we can safely not link it for now.

maximiliankolb (commented):
Sometimes is says "Installing", sometimes, "Installation", sometimes with "guide", sometimes without.

Also, maybe we can rename `Installation Guide` to "Installing (Foreman) Server", in contrast to "Installing Smart Proxy"?

web/content/js/nav.js
```diff
@@ -19,37 +19,25 @@ let navVersions = [
...
 		href: "/Installing_Proxy_on_Red_Hat/index-foreman-el.html"
-	},{
-		title: "Installing on Debian/Ubuntu",
-		href: "/Installing_Server_on_Debian/index-foreman-deb.html"
-	},{
-		title: "Installing Smart Proxy on Debian/Ubuntu",
-		href: "/Installing_Proxy_on_Debian/index-foreman-deb.html"
 	},{
 		title: "Upgrading and Updating",
 		href: "/Upgrading_and_Updating/index-foreman-el.html"
```
ekohl, line 37: I don't think this guide is ready for public consumption. It is very Katello centric. It's really crucial once we go live, but  at this point it's not usable.

ekohl: After https://github.com/theforeman/foreman-documentation/pull/974 was merged, I looked at more chapters that can be skipped: https://github.com/theforeman/foreman-documentation/pull/976 takes a first stab at it. However, I haven't gotten to the meat of the upgrade: the actual upgrade commands. I think they may be broken for all of upstream, but at least for Debian you can't use foreman-maintain.

lzap: I will only keep Katello version for now then.

lzap:
Good point, I have reworded just to 'Installing' to match directory names. We are all good, let's merge and progress further.

lzap:
Allright, merged. Next step - I am going to convert the old manual. Need to figure out a good menu item name for it.

PR #973, opened 2022-01-17
List CentOS 8 before CentOS 7

ekohl:
It is better to list the newest first since we want to encourage users to use the latest version.

I also noticed we don't list RHEL 8 support while in upstream I think it should work. In general I've been thinking about merging some sections into just EL7 or EL8 but that can change the scope quite a bit so I won't do it in this PR.

### DIFF
--- a/guides/common/modules/proc_configuring-repositories.adoc
+++ b/guides/common/modules/proc_configuring-repositories.adoc
@@ -5,8 +5,8 @@
 Use this procedure to enable the repositories that are required to install {ProductName}. Choose from the available list which operating system and version you are installing on:
 
 ifdef::foreman-el,katello[]
-* xref:#repositories-centos-7[CentOS 7]
 * xref:#repositories-centos-8[CentOS 8]
+* xref:#repositories-centos-7[CentOS 7]
 endif::[]
 ifdef::foreman-el,katello,satellite[]
 * xref:#repositories-rhel-7[Red Hat Enterprise Linux 7]
@@ -17,29 +17,6 @@ ifdef::foreman-deb[]
 endif::[]
 
 ifdef::foreman-el,katello[]
-== [[repositories-centos-7]]CentOS 7
-
-:distribution-major-version: 7
-:package-manager: yum
-include::proc_configuring-repositories-el.adoc[]
-
-+
-. Install the `epel-release` package:
-+
-[options="nowrap" subs="+quotes,attributes"]
-----
-# {package-manager} install epel-release
-----
-+
-. Install the `centos-release-scl-rh` package:
-+
-[options="nowrap" subs="+quotes,attributes"]
-----
-# {package-manager} install centos-release-scl-rh
-----
-endif::[]
-
-ifdef::foreman-el,katello[]
 == [[repositories-centos-8]]CentOS 8
 
 :distribution-major-version: 8
@@ -47,9 +24,6 @@ ifdef::foreman-el,katello[]
 
 include::proc_configuring-repositories-el.adoc[]
 
-endif::[]
-
-ifdef::foreman-el,katello[]
 +
 . Enable Ruby 2.7 module:
 +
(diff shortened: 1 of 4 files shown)


lzap:
I was thinking, this particular guide could have been simply broken down in a different way - each platform could have been a separate module and we would simply include modules like centos7, centos8, debian, ubuntu so we would have relevant commands together. This is good enough for now I believe.

ekohl:
I have a follow up PR ready that may spark some discussion. I did indeed go with "good enough for now".

lzap:
I am going to review what we have on the upstream site and this patch:

Snímek obrazovky 2022-01-18 v 13 24 19
Snímek obrazovky 2022-01-18 v 13 24 26
Snímek obrazovky 2022-01-18 v 13 24 34
Snímek obrazovky 2022-01-18 v 13 24 57
Snímek obrazovky 2022-01-18 v 13 25 06

lzap:
First observation, we are missing CentOS 8 Stream and RHEL 8. The old Ubuntu is already gone which is correct (TBD upstream).

lzap:
Differences (just confirm if these are not issues):

* CentOS 8: we do not have that `postgresql enable` upstream
* CentOS 7: we do not carry `sudo yum-config-manager --enable extras` over
* RHEL 7: repositories are different in the `sudo yum-config-manager --enable rhel-7-server-optional-rpms rhel-server-rhscl-7-rpms` command but I think this is because we instruct users to disable all first (upstream we do not do this)
* upstream has an extra `sudo apt-get -y install ca-certificates gpg` command that install previously installed ca-certificates (?) but also gpg (perhaps we want to install that as well here)
* ubnutu 20.04 - the same problem but this time `gpg` is not installed upstream
*

ekohl:
> CentOS 8: we do not have that `postgresql enable` upstream

Technically it' supported to use PostgreSQL 10 (default module) but for Katello I think you need to use 12 due to the `-evr` package. Ideally we'd tell users to install 12 since it provides better performance but we may need to provide upgrade instructions.

Since it's the install guide, I'd say it's safe and preferred to add the enable PG 12 module part.

> CentOS 7: we do not carry `sudo yum-config-manager --enable extras` over

This is for safety. It's enabled by default but we do depend on it

> RHEL 7: repositories are different in the `sudo yum-config-manager --enable rhel-7-server-optional-rpms rhel-server-rhscl-7-rpms` command but I think this is because we instruct users to disable all first (upstream we do not do this)

I'm not sure what's wise here since I rarely use RHEL.

> upstream has an extra `sudo apt-get -y install ca-certificates gpg` command that install previously installed ca-certificates (?) but also gpg (perhaps we want to install that as well here)

`ca-certificates` is needed for `wget`. It's a recommended package so probably already present. Like extras on CentOS 7: it's explicit in what's needed, but likely redundant.

As for `gpg`, since bfb5cd8e1ec6105594531bf81058882fee8efaff it's no longer needed. We still have some `sudo apt-key add -` instructions in the upstream guide, but that's no longer the preferred solution. The simple wget solution does not require `gpg` to be installed and is easier overall.

> * CentOS 8 Stream is missing

CentOS Stream 8 & CentOS Linux 8 have the same instructions. Also, CentOS Linux 8 is EOL. Let's standardize on CentOS Stream 8.

> * RHEL 8 is missing

Agreed, this should be added. However, the whole guide needs a major overhaul to properly support RHEL 8. IMHO it's fine to start adding it to the quickstart guide, but we should go over everything. For downstream it's also needed since Satellite 7.0 (based on 3.1) will need it too.

lzap:
> I'm not sure what's wise here since I rarely use RHEL.

I think I am good. We like to recommend to disable all and then re-enable exactly those which are needed, good practice also for upstream. Let's keep it that way.

> Let's standardize on CentOS Stream 8.

Ok.

Any plans to push or are we good now? I don't mind merging this as is, or if you want also to take a look on mentioning Stream or creating RHEL 8 too.

ekohl:
> Any plans to push or are we good now? I don't mind merging this as is, or if you want also to take a look on mentioning Stream or creating RHEL 8 too.

I'd prefer to merge it as is. Consider the above discussion as an issue to work on. If you have time to pick it up, please do. Otherwise, should we create a new issue?

lzap:
Merged.

> I'd prefer to merge it as is. Consider the above discussion as an issue to work on. If you have time to pick it up, please do. Otherwise, should we create a new issue?

I will work on this right away. Will incorporate changes from https://github.com/theforeman/foreman-documentation/pull/982 too.

PR #974, opened 2022-01-17
Hide tuning profiles on foreman-{el,deb}

ekohl:
The tuning profiles only exist within Katello so they must be otherwise.

I started going through the rest of the document, but it needs a LOT of work. This is just the tip of the ice berg.


maximiliankolb (approved):
Thanks, LGTM.

ekohl:
FYI: I didn't specify a cherry pick to 3.1 since I think we should first see what's needed to make the guide usable.

PR #976, opened 2022-01-17
Partial cleanup of the upgrading guide

ekohl:
This hides a lot of the Katello-specifics in the upgrade guide. There is still much more to do, but this moves it a step forward.

See the individual commits for explanations and reasoning.

I'm unsure what the state of this document is in the Orcharhino context but I'm getting the impression it's completely unused.


guides/doc-Upgrading_and_Updating/topics/upgrading_process_overview.adoc
```diff
@@ -91,8 +85,10 @@ ifdef::satellite[]
 endif::[]
 . Upgrade {ProjectServer} and all {SmartProxyServer}s to {Project} {ProjectVersion}.
```
ekohl, line 86: This is now the only step remaining in the "high level overview" on non-Katello. It feels really weird but I'm not sure what to do with it.

I've been thinking about splitting it up:
```suggestion
. Upgrade {ProjectServer} to {ProjectVersion}.
. Upgrade all {SmartProxyServer}s to {ProjectVersion}.
```

This would be correct in all scenarios and matches the headers. On the other hand, perhaps it's better to just skip the whole section to keep things short.

guides/doc-Upgrading_and_Updating/topics/upgrading_clients.adoc
```diff
@@ -68,6 +70,16 @@ For more information, see {ManagingHostsDocURL}configuring-and-setting-up-remote
...
 --enable={project-client-RHEL7-url}
 ----
+endif::[]
+
+ifdef::katello[]
+. Enable the {project-client-name} repository for this version of {Project}.
++
+[options="nowrap" subs="attributes"]
+----
+# yum install {project-client-RHEL7-url}
```
maximiliankolb, line 80: **NOT PART OF YOUR PR**: IMHO this is something we should tackle in the future: the guide is basically made for EL7, but not EL8, SLES, or Debian/Ubuntu.

guides/doc-Upgrading_and_Updating/topics/upgrading_process_overview.adoc
maximiliankolb: I think splitting it up is the way to go. Users can then still do both procedures sequentially.

maximiliankolb (commented):
All "ifdef Katello/orcharhino/Satellite" look good; however, we currently don't use upgrade docs for our downstream product.

guides/doc-Upgrading_and_Updating/topics/upgrading_clients.adoc
ekohl: Oh absolutely, still a massive concern I have. Also why I think the new docs still need a huge amount of work. It may be a good thing to focus the Red Hat docs writers on since for Satellite 7.0 it will be needed to have RHEL 8 support.

ekohl:
Updated:
* Fix build issue
* Split high level steps
* Hide FIPS
* Hide irrelevant section on non-Katello

Note that this is still incomplete. The real upgrade instructions are still plain broken. I don't get it: Katello users should be using this guide to upgrade, right? How can they use this? It literally doesn't contain the instructions to update the yum repository files.

lzap (approved):
LGTM

@maximiliankolb please merge when you are done.

maximiliankolb (approved):
Thanks for cleaning up.

PR #977, opened 2022-01-18
Add legacy Foreman manual

lzap:
This adds an unchanged legacy manual as a new guide, this was converted from:

https://github.com/theforeman/theforeman.org/pull/1916

I did not do any changes except removing two broken images (settings) these were missing ini the images directory.

I added a new menu item called "Complete Guide" into both EL and Debian menus to state this is the complete experience at the moment.

I had to modify how images are build, instead of symlinking them I now hard-link them into the destination recursively, this allows us to use directories in the images folders. The makefile now explicitly expects images/ directory to be present in the guide directory, it holds images only specific to that guide, shared images are in ../common/images as previously. This was done to be able to copy unchanged directory structure for the legacy guide. We can iterate from here and start taking (or updating) screenshots and moving them to the common directory.

I do not want to fix bugs in this PR, there are some minor problems which needs to be fixed. I would rather break this up into separate PRs. Examples:

* Break up a single page into one file per chapter (for better navigation)
* Links in 5.2 Compute Resources table are broken (needs to be fixed manually)
* This (C) is rendered as Copyright unicode character (needs to be fixed manually)
* Modularize content
* Create references
* Fix bugs
* 
If you want to see the current version head over to: https://lzap.fedorapeople.org/temp/foreman_manual/foreman-manual-nightly.html


ekohl:
I'm not happy with this if I'm honest. This adds a massive document that's hard to maintain. Some things I think we can already do:
* Merge https://github.com/theforeman/theforeman.org/pull/1940 and use that as a base
* Drop the quickstart instructions and ensure the separate quickstart guide is actually correct
* Figure out a better way to "convert" the OS selection drop down
* Copy over release notes from the old guide to this repo's release notes guide
* Drop the release notes from the manual and link to the guide we already have or even include it via modules

lzap:
I can do some if not all of these as separate commits, my concern is that I don't want to work on this for a long time only after finding out that there was a change in the old guide that needs to be "backported" here. I want to perform a quick switch instead.

Are these your only major concerns? After those changes I will go ahead, publish the legacy guide on the new site and remove it from the original site linking it from there.

ekohl:
I still maintain that I don't want real users reading docs.theforeman.org without some more SERIOUS work. Almost every guide contains major flaws. It gives the impression documentation for Foreman 3.1, 3.0 and 2.5 is of production quality but I'd say it's highly experimental. IMHO we should not link Foreman on EL and Foreman on deb for those releases - limit it to nightly and only target it for 3.2.

maximiliankolb (commented):
I did not read all 13.000 lines, but two things:

I am unsure about how to feel about all the screenshots. I would be fine with merging them and then gradually removing them, esp. when we have adequate procedures. However, this would mean we have to live with a bigger git history.

Second, I ask myself which is smarter/less work to proceed: convert everything and then remove individual procedures/sections/chapters because we already have better content; or selectively copy/convert content from the manual to the docs.

ekohl:
> I am unsure about how to feel about all the screenshots. I would be fine with merging them and then gradually removing them, esp. when we have adequate procedures. However, this would mean we have to live with a bigger git history.

I'm also concerned about this. That's why I'd prefer to clean up more things in the current guides and reuse as much as possible.

lzap:
> I am unsure about how to feel about all the screenshots. I would be fine with merging them and then gradually removing them, esp. when we have adequate procedures. However, this would mean we have to live with a bigger git history.

I made sure they are kept separate, the screenshots from the old guide are in the `images/` subfolder while most of our images are in the `common/images` folder where it gets hardlinked (previusly symlinked) during build. Quite frankly, our images are currently a mess, a lot of files named simply `1.png` etc, it definitely needs a cleanup.

Not sure about that git history concern, they are less than 8 MiB in size, not big deal I think. But if you want to review and give me list of screenshots that you think we should definitely remove prior merging, then feel free and I will rip them off.

> Second, I ask myself which is smarter/less work to proceed: convert everything and then remove individual procedures/sections/chapters because we already have better content; or selectively copy/convert content from the manual to the docs.

I don't think we need to be formal about it, we can simply evolve in smaller steps. Something like:

* Introduction is literally just one paragraph, we can probably use what we have in the modules today.
* The Diagram - perhaps we keep it and use it in other guides as well.
* Release notes - to be deleted and replaced with our new guide
* Quickstart - ditto
* Installation and upgrade sections - I suggest to leave them as-is for now, until we are satisfied with our guides
* Installer generated table - we can extract into a separate adoc file and update the generator, perhaps this can be appendix or even a separate guide (this is by the way about 1/20th of the whole file)
* Smart proxy "scenarios" - these should be moved into provisioning/puppet guides I think as they are examples how to setup environment
* Install from source - modularize, move it into our install guide perhaps (if we want to keep it)
* Configuration - describes our Administer - Settings options - that is probably material for other guide
* Configuration - ENC - describes ENC setup - perhaps Puppet guide
* Upgrade guide - do not touch until we improve upgrade guide so we can link it instead
* General Foreman - this is the funky part - lots of various bits
* LDAP - we cover in a much better way, we can just link
* Roles and Premissions, Audit, Searching - ditto (after review)
* Puppet classes and Parameters - there is a lot of good content that needs to be reviewed and incorporated into Puppet guide
* Smart proxy installation and setting - the same as with installation section
* DHCP, DNS, TFTP, Reams - I will review and move existing bits into Provisioning Guide
* Libvirt proxy module - should be deleted, unmaintained
* SSL - I think this could be moved to Architecture guide or somewhere else
* Provisioning - I will review but 90% of this is in Provisioning Guide (and much more)
* Examples of partition tables - outdated, should be deleted
* Workflow examples - this is actually the only provisioning bit I will consider migrating over
* FreeBSD, Windows - we should move over into provisioning guide
* Hammer - we should have this in a guide
* Email, Reports - review, incorporate into other guides
* Advanced Foreman
* API and examples - review, incorporate into other guides
* Dynflow Scaling - Tuning Guide should cover this
* Compute Resources - fully covered in Prov Guide (will review)
* SSL for Smart Proxy - again we talk about this for the 3rd time now, we need to review and provide a consistent content
* Backup and restore - review, incorporate into other guides
* Kerberos and PAM, SSO - review, incorporate into other guides
* HTTPS Proxy - for the 4rd time a topic around smart proxies and SSL, this is not really great
* Troubleshooting, Getting help  - review, incorporate into maybe a new guide, maybe we can move this to wiki or somewhere else

maximiliankolb:
It looks like you have a plan @lzap, so I won't stand in the way of merging this and moving forward. Once it builds, I can have a quick look at the changes to the makefile (or whatever needs to be changed) again. Then I would also like to press to issue to delete the content on the website itself and add a fancy like to docs.theforeman.org :)

lzap:
Nobody expects the guide to be rewritten next month, we will cherry pick various chapters, compare with what we have in our new guides, incorporate everything of a value and then deleting the original content linking to the new location. In the end, the guide should be shorter and shorter full of references until we realize its just three chapters which we can move and close it.

It cannot be worse than what we have today - two sites, to completely different manuals not linked together, then there's Foreman vs Katello as well.

Anyway, if anyone has bigger concerns before I fully commit to convert once again and then do the necessary changes (breaking up manual into smaller files a bit, extracting installer table, fixing minor things) **speak up now please**! It is goonna be some work, once it builds I would like to go ahead, do the change and remove the manual from the old location.

ekohl:
> Installation and upgrade sections - I suggest to leave them as-is for now, until we are satisfied with our guides

I'm tempted to improve our current modules/procs because they are quite close.

> Installer generated table - we can extract into a separate adoc file and update the generator, perhaps this can be appendix or even a separate guide (this is by the way about 1/20th of the whole file)

:+1: for separate adoc. I'd also move it to an appendix while you're at it. We now have a good structure for it, let's use it.

> Install from source - modularize, move it into our install guide perhaps (if we want to keep it)

IMHO we can drop it. This has been discussed before, but never followed through. Let's take this chance.

> The Diagram - perhaps we keep it and use it in other guides as well.

I think most people desperately want to update it. The person who does this can become a legend because everyone copies it to their presentation ;)

Other than that I think I think it's a decent plan.

Something we must also think about is redirects. You will need to properly redirect or risk losing a lot of SEO.

Another BIG blocker is that the Foreman application itself links to the website. https://github.com/theforeman/foreman/pull/8613 was started, but never finished. That's actually a good reason to only limit the new guide to nightly and leave existing releases in place.

lzap:
> This has been discussed before, but never followed through. Let's take this chance.

Yeah actually I moved it from https://theforeman.org/contribute.html into the manual last year IIRC :) We can move it back, or drop it. I don't know to be honest. Probably let's keep it for now until we solve the whole manual... ;)

> Something we must also think about is redirects. You will need to properly redirect or risk losing a lot of SEO.

My plan is to keep all the chapters on the original place with original HTML anchors and create simply links to the chapters here. I do have a happing structure which I created during conversion actually, should work I hope. Worst case - I will create it manually.

That should ensure also that links coming from the old Foreman version would work. Until we change the links in Foreman itself, which will be easy - we can use the same anchors we use downstream.

> Another BIG blocker is that the Foreman application itself links to the website. [theforeman/foreman#8613](https://github.com/theforeman/foreman/pull/8613) was started, but never finished. That's actually a good reason to only limit the new guide to nightly and leave existing releases in place.

Oh yeah, that is exactly what I thought. Leos told me he would resurrect the effort once we will have a landing document. That is this PR I assume. I hope for a bigger change - to have a special landing page somewhere (perhaps it can be a nanoc-generated page with links to various guides about various topics and maybe even screenshots). That would be awesome, ton of work tho.

lzap:
Rebased, it is based on https://github.com/theforeman/theforeman.org/commit/76830f5820371b1d52949073405606850da31cc4 for the record.

lzap:
Closing this, but the script which I used is still available in my branch so feel free to use it if needed.

PR #979, opened 2022-01-18
Admin guide: hide invalid sections

ekohl:
Installing the Katello root CA is, as the name suggests, a Katello only thing.

The starting and stopping guide relies on foreman-maintain which is unavailable on Debian so the whole section is hidden.

The external PostgreSQL guide is EL7 only (even incorrect on EL8 which is supported). For now it's just hidden on Debian but this needs serious work.

Additionally, OpenSCAP is unavailable on Debian so the sections are hidden.

This works towards resolving https://github.com/theforeman/foreman-documentation/issues/676.

### DIFF
--- a/guides/common/assembly_accessing-server.adoc
+++ b/guides/common/assembly_accessing-server.adoc
@@ -1,6 +1,8 @@
 include::modules/con_accessing-server.adoc[]
 
+ifdef::katello,orcharhino,satellite[]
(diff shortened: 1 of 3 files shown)


maximiliankolb (approved):
Thanks for the PR. LGTM!

lzap (approved):
Thank you!

PR #980, opened 2022-01-18
Only allow the exact remote IP for Foreman

ekohl:
There is no reason to assume a whole /24 subnet needs access. This is insecure and can create security problems. By only allowing /32 it's limited to the exact IP.

### DIFF
--- a/guides/common/modules/proc_installing-postgresql.adoc
+++ b/guides/common/modules/proc_installing-postgresql.adoc
@@ -49,7 +49,7 @@ listen_addresses = '*'
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-  host  all   all   _{Project}_ip_/24   md5
+  host  all   all   _{Project}_ip_/32   md5
 ----
 
 . To start, and enable PostgreSQL service, enter the following commands:


evgeni (approved):

ekohl:
I do think this applies to all guides so I'd propose to cherry pick it to at least 3.1, but possibly 2.5 as well for RH downstream.

lzap (approved):
Had no idea you can use such syntax, thanks.

lzap:
By the way that md5, ehm. Not sure how relevant this is in 2021 but that is another story. Merged.

ekohl:
Yes, we'll need a migration to scram-sha-256 (see https://www.postgresql.org/docs/10/auth-methods.html#AUTH-PASSWORD). Perhaps we need to put that on the roadmap somewhere. However, they do note:
> To ease transition from the md5 method to the newer SCRAM method, if md5 is specified as a method in pg_hba.conf but the user's password on the server is encrypted for SCRAM (see below), then SCRAM-based authentication will automatically be chosen instead.

So the md5 bit there may not be as insecure as it looks.

PR #981, opened 2022-01-18
Install guides: Hide invalid sections

ekohl:
Alternative cname, custom certs and external DB are only valid on Katello scenarios. Eventually these should be written for non-Katello but for now this hides the invalid sections.


maximiliankolb (approved):
LGTM.

lzap (approved):
External database is something that I am sure our (Foreman core) users do, but sure let's not show it until it's ready.

ekohl:
> External database is something that I am sure our (Foreman core) users do, but sure let's not show it until it's ready.

We'll need to modify it. I've opened https://github.com/theforeman/foreman-documentation/issues/985 to track that and list what needs to be done.

PR #983, opened 2022-01-18
Make the drop downs fit the width of the content

ekohl:
Many guide names are longer than the width of the title. This resizes the nav to fix the content. According to [caniuse](https://caniuse.com/mdn-css_properties_flex-basis_max-content) it's not available on all browsers, but at least on most. On unsupported browsers I'd expect it to behave as it previously did.

### DIFF
--- a/web/content/css/_nav.scss
+++ b/web/content/css/_nav.scss
@@ -80,6 +80,7 @@ nav .dropdown-menu {
   border: 1px solid rgba(0, 0, 0, 0.3);
   border-radius: 0.5em;
   background-color: #fff;
+  width: max-content;
 }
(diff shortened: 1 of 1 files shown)


lzap:
Yeah I could not figure this one out, thanks :)

lzap:
For the record, JS/CSS changes must be pushed into stable branches too so asciidoc can rebuild guides with the new style. Except `nav.js` that one is dynamically loaded.

Pushing:

2e4f7809..7522d42a  3.1 -> 3.1
d025346a..35a1ebab  3.0 -> 3.0
bc49ffab..43f59b35  2.5 -> 2.5

PR #998, opened 2022-01-19
Fix 3.0 branch building (WIP)

lzap:
Do not merge, need to test guides building for 3.0

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -42,6 +42,9 @@ jobs:
           ruby-version: ${{ env.RUBY_VERSION }}
           bundler-cache: true
 
+      - name: Run bundle install
+        run: bundle install
+
       - name: Build HTML with real links
         run: |
           make clean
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -42,6 +42,9 @@ jobs:
           ruby-version: ${{ env.RUBY_VERSION }}
           bundler-cache: true
 
+      - name: Run bundle install
+        run: bundle install
+
```
lzap, line 47: I don't understand why but in the 3.0 and 2.5 branches build fails with "bundle exec - not found". Ruby somehow installs correctly. I tried to copy the very same `deploy.yml` over from master but it just fails.

lzap:
I am giving up, putting "bundle install" there for now, it is installing gems so apparently the ruby/setup-ruby does not somehow work or installs it incorrectly. Not sure maybe @ekohl can spot something later on.

lzap:
Must be something with Gemfile or Gemfile.lock but I doublechecked that and they were the same.

ekohl:
It can be that psych is in the gem cache. You can try https://github.com/ruby/setup-ruby#dealing-with-a-corrupted-cache

PR #772, opened 2021-10-22
Move "Convert to RHEL" to separate guide

maximiliankolb:
Hi!
I propose the following change: Let's move the "Convert a CentOS host to RHEL" section to a separate guide, because it's not really part of any procedure to manage content.

**This should not be merged without a broader discussion to see if we're moving towards the right direction.**

PR #999, opened 2022-01-19
Show convert2rhel guide

maximiliankolb:
We've moved content from the content management guide, but we currently don't show it in the navigation.

Cherry-pick into:

* [x] Foreman 3.1

### DIFF
--- a/web/content/js/nav.js
+++ b/web/content/js/nav.js
@@ -128,4 +128,3 @@ let navItems = [
 	}]
     }
 ]
-
--- a/web/content/js/nav.js
+++ b/web/content/js/nav.js
@@ -105,6 +105,9 @@ let navItems = [
 		title: "Content Management Guide",
 		href: "/Content_Management_Guide/index-katello.html"
 	},{
+		title: "Converting a Host to RHEL",
+		href: "/Converting_a_Host_to_RHEL/index-katello.html"
+	},{
 		title: "Deploying Foreman on AWS",
 		href: "/Deploying_on_AWS/index-katello.html"
 	},{


web/content/js/nav.js
```diff
@@ -104,6 +104,9 @@ let navItems = [
 	},{
 		title: "Content Management Guide",
 		href: "/Content_Management_Guide/index-katello.html"
+	},{
+		title: "Converting a Host to RHEL",
+		href: "/Converting_a_Host_to_RHEL/index-katello-el.html"
```
ekohl, line 109:
```suggestion
		href: "/Converting_a_Host_to_RHEL/index-katello.html"
```

ekohl (changes requested):
This will result in 404 links in older guides, right?

web/content/js/nav.js
maximiliankolb: Thanks, applied!

maximiliankolb:
As far as I can tell, nightly and 3.1 had included the "convert2rhel" assembly in the content management guide until today; now it is/should be show as a separate guide. See https://github.com/theforeman/foreman-documentation/pull/772.

ekohl (approved):
The 404 issue will be a thing, but I'll open an issue describing that and a possible solution.

maximiliankolb:
I don't fully understand this yet. Can you please give an example where the 404 would occur?

ekohl:
You mentioned nightly and 3.1, but https://docs.theforeman.org also lists 2.5 and 2.4. For Katello those are production level and there is no other source. The menu is the same for all versions and there is no way to show a different menu for 3.0/2.5/2.4 than 3.1/nightly. This is a major limitation since it means it's hard to add new guides. I'd also like to hide more guides for Foreman on (Deb|EL). Perhaps even hide all Foreman guides on older than nightly since they're experimental.

maximiliankolb:
This still works IMHO: https://docs.theforeman.org/2.5/Content_Management_Guide/index-katello.html#assembly_converting-a_host-with-convert2rhel_content-management

My understanding is that every branch (2.5,3.1, nightly, etc) has its own list of navigation items. So older versions can still reach the "convert2rhel" section.

ekohl:
> My understanding is that every branch (2.5,3.1, nightly, etc) has its own list of navigation items. So older versions can still reach the "convert2rhel" section.

No, nav.js now has a list of versions (2.4, 2.5, ..., nightly) and a list of navigation items. Those navigation items are shared. See https://docs.theforeman.org/js/nav.js now. So to reproduce it:
* Visit https://docs.theforeman.org
* Select 2.4
* Select Katello Guides
* Click Convert to RHEL guide
* It links to https://docs.theforeman.org/2.4/Converting_a_Host_to_RHEL/index-katello.html

You can see that's a 404.

maximiliankolb:
Got it!

I don't think we can really fix this without touching the basic "historically grown" (I mean that in the nicest way possible!) architecture. IMHO, this is a basic design flaw that the branches are not totally independent of each other.

Does that mean, in order to not break links, we should backport changes to the navigation? I don't really want to do this to be honest.

ekohl:
Today we can't. IMHO what we should implement is basically a `navigation.json` or `navigation.js` in every branch, perhaps including the landing page (`release-x.y.adoc` in the `web` directory) and use that. It means writing a bit more Javascript, but it shouldn't be very complicated.

As I said, I'll try to capture it in an issue.

PR #1002, opened 2022-01-19
Fix PDF generation, exit code from make and parallelization

lzap:
So it turns out we have several issues.

Asciidoctor PDF now has a missing dependency "matrix". Without this gem, PDF generation simply fails, probably a bug usptream.

Second problem actually hid the first one - I added `asciidoctor ... | tee LOGFILE; grep LOGFILE` to search for warnings which were never returned via exit code by asciidoctor, but I haven't realized that the pipe always returns zero exit code so even when HTML/PDF generation was broken (we had brokens links), makefile was still returning zero.

After I fixed the tee problem, a third appeared - recursive make implementation did not care about return codes from the submakes at all, therefore when targets "all-html" or "all-pdf" were called, it always returned zero even if we had issues (warnings via the grep command).

Finally, I am making small improvement - for PR CI runs, let's simply build just one PDF to speed things up just to see if it actually builds. For deployments, let's build all of them, this creates a huge ZIP file and takes much longer. I think this is a good solution, previously we were building no PDFs for PR CI runs.

The root cause for PDF generation failing was because I was giving it the wrong images directory. I am going to fix all problems in separate commits. On top of that, since I am looking into this, I reimplemented the parallel make after reading GNU Makefile manual now in a proper way so it is `-j` friendly. Now we can use it on CI to even further speedups builds, for example PDF generation on my Intel NUC 8thgen:

```
# make clean; make html
real	0m14,125s
user	0m12,568s
sys	0m1,552s

# make clean; make html -j5
real	0m2,578s
user	0m8,955s
sys	0m0,897s

# make clean; make pdf
real	0m49,084s
user	0m46,572s
sys	0m2,375s

# make clean; make pdf -j5
real	0m23,316s
user	1m16,476s
sys	0m3,601s
```

For parallelization, I had to drop the clunky images symlinking, now we use simply `cp` to copy images over to the destination dir for each guide. I actually have a hard-link option to make it even faster. I also had to solve the common/images vs images/ directories problem which I run into when I was working on https://github.com/theforeman/foreman-documentation/pull/977. We used to have images/ subdirectory for some guides, now we have it for all of them and images are copied from both directories into a single one.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -14,6 +14,7 @@ env:
   RUBY_VERSION: 3.1
   PYTHON_VERSION: 2.x
   BUNDLE_WITHOUT: "nanoc"
+  MAKE_J: 3
 
 jobs:
   check-html:
@@ -55,11 +56,11 @@ jobs:
 
       - name: Build HTML with local links
         run: |
-          make html BUILD=foreman-el LINKS=LOCAL
-          make html BUILD=foreman-deb LINKS=LOCAL
-          make html BUILD=katello LINKS=LOCAL
-          make html BUILD=satellite LINKS=LOCAL
-          make html BUILD=orcharhino LINKS=LOCAL
+          make -j ${{ env.MAKE_J }} html BUILD=foreman-el LINKS=LOCAL
+          make -j ${{ env.MAKE_J }} html BUILD=foreman-deb LINKS=LOCAL
+          make -j ${{ env.MAKE_J }} html BUILD=katello LINKS=LOCAL
+          make -j ${{ env.MAKE_J }} html BUILD=satellite LINKS=LOCAL
+          make -j ${{ env.MAKE_J }} html BUILD=orcharhino LINKS=LOCAL
 
       - name: Check HTML links of all builds
         run: |
@@ -94,11 +95,11 @@ jobs:
       - name: Build HTML with real links
         run: |
           make clean
-          make html BUILD=foreman-el
-          make html BUILD=foreman-deb
-          make html BUILD=katello
-          make html BUILD=satellite
-          make html BUILD=orcharhino
+          make -j ${{ env.MAKE_J }} html BUILD=foreman-el
+          make -j ${{ env.MAKE_J }} html BUILD=foreman-deb
+          make -j ${{ env.MAKE_J }} html BUILD=katello
+          make -j ${{ env.MAKE_J }} html BUILD=satellite
+          make -j ${{ env.MAKE_J }} html BUILD=orcharhino
 
       - name: Upload HTML
         uses: actions/upload-artifact@v2
@@ -107,11 +108,6 @@ jobs:
           path: guides/build/
 
   build-pdf:
-    if: >
-      github.repository_owner == 'theforeman' &&
-      (github.ref == 'refs/heads/master' ||
-      startsWith(github.ref, 'refs/heads/2.') ||
-      startsWith(github.ref, 'refs/heads/3.'))
     runs-on: ubuntu-latest
     defaults:
       run:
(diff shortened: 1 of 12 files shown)


lzap:
So added `make -j3` to GHA (builders do appear to have 2 cores according to documentation). Previous run was running 2m49s after the change it was 2m24s. Not a huge difference, for PDFs it is similar: 1m18s vs 52s.

Anyway, my main goal was to fix how exit codes are handled and also PDFs were not generating at all.

.github/workflows/deploy.yml
```diff
@@ -107,11 +108,6 @@ jobs:
           path: guides/build/
 
   build-pdf:
-    if: >
-      github.repository_owner == 'theforeman' &&
-      (github.ref == 'refs/heads/master' ||
-      startsWith(github.ref, 'refs/heads/2.') ||
-      startsWith(github.ref, 'refs/heads/3.'))
```
lzap, line 114: I moved this down, we now build one PDF version for PR CI and all for deploy commits.

Gemfile
```diff
@@ -2,6 +2,8 @@ source 'https://rubygems.org'
 
 gem 'asciidoctor'
 gem 'asciidoctor-pdf'
+# missing dependency for asciidoctor-pdf
+gem 'matrix'
```
lzap, line 6: This was causing Ruby to fail.

guides/Makefile
```diff
@@ -1,37 +1,29 @@
...
-clean: all-clean
+subdirs: $(SUBDIRS)
+
+$(SUBDIRS):
+	$(MAKE) -C "$@" $(MAKECMDGOALS)
+
+clean:
+	@for DIR in $(SUBDIRS) ; do \
+		$(MAKE) -s --directory="$$DIR" clean ; \
+	done
```
lzap, line 20: This is completely simplified, it now properly calls make so recursive option does work.

guides/Makefile
```diff
@@ -1,37 +1,29 @@
...
-
-$(GUIDES-PDF):
-	$(MAKE) --directory=$@ pdf
-
-all-clean:
-	for DIR in $(ALL_DIRS) ; do \
-		$(MAKE) --directory="$$DIR" clean ; \
-	done
+serve:
+	python -m http.server --directory ./build 8813
```
lzap, line 29: The same target is in the common Makefile, I use this for testing.

guides/common/Makefile
```diff
@@ -57,15 +60,11 @@ linkchecker: html
 open-pdf: pdf
 	${BROWSER_OPEN} "$(realpath $(ROOTDIR)/$(DEST_PDF))"
 
-$(BUILD_DIR)/default.css: $(CSS_SOURCES)
-	bundle exec sass -C --style $(CSS_STYLE) -I $(COMMON_DIR)/css $(COMMON_DIR)/css/default.scss $@
-
-$(DEST_DIR)/$(BUILD).css: $(BUILD_DIR)/default.css
-	cp $< $@
+$(DEST_DIR)/$(BUILD).css: $(CSS_SOURCES)
+	bundle exec sass --cache-location $(DEST_DIR)/sass-cache --style $(CSS_STYLE) -I $(COMMON_DIR)/css $(COMMON_DIR)/css/default.scss $@
```
lzap, line 64: Because sass command is a bit slow, previously it built into a common BUILD_DIR just once and other guides could simply copy the generated file quickly over to the destination. With parallel option this is no longer possible (I was hitting problems). So now sass needs to be executed for each guide separately, which is a bit slower. I implemented cache option, but that does not speeds things up on CI (where we start from a clean state anyway). Later on, we can migrate to Dart-SCSS.

guides/common/Makefile
```diff
@@ -57,15 +60,11 @@ linkchecker: html
...
 
 $(IMAGES_TS): $(IMAGES)
-	cp $? $(IMAGES_DIR)/
-	pushd "$(IMAGES_DIR)" >/dev/null && for IMG in *.*; do ln -fs ../../images/$$IMG ../common/images/$$IMG; done && popd >/dev/null
+	cp -lrf $? $(IMAGES_DIR)
 	@touch $(IMAGES_TS)
```
lzap, line 68: This was dodgy, now images are properly copied (if they are new).

guides/common/Makefile
```diff
@@ -76,6 +75,6 @@ $(DEST_HTML): $(SOURCES) $(DEPENDENCIES) $(DOCINFO_SOURCES)
 	@! grep -RE '<a href="#[a-zA-Z0-9_-]+">\[[a-zA-Z0-9_-]+\]</a>' $@
 
 $(DEST_PDF): $(SOURCES) $(IMAGES) $(DEPENDENCIES)
-	bundle exec asciidoctor-pdf -a build=$(BUILD) -d book --trace -o $@ $< 2>&1 | tee $@.log
+	bundle exec asciidoctor-pdf -a build=$(BUILD) -a imagesdir=$(IMAGES_DIR) -d book --trace -o $@ $< 2>&1 | tee $@.log
```
lzap, line 78: This was another issue with PDF generation - it could not find images in the correct directory. This fixes it.

ekohl (approved):
Looks good to me.

guides/doc-Administering_with_Ansible/master.adoc
```diff
@@ -1,5 +1,5 @@
 include::common/attributes.adoc[]
-:imagesdir: common/images
+:imagesdir: images
```
lzap, line 2: This also consolidates where images should be find - they are always copied by make into the BUILD_DIR/images so no longer we will confuse common and non-common images. They are all copied to the same destination, of course locals overriding common.

PR #1006, opened 2022-01-20
Remove _on_$os from installation guides

### LINKED ISSUE #970: Remove "Red Hat" from directories
This needs to be communicated with the Red Hat team so they know when their downstream build breaks. I would like ask @melcorr to arrange this. I can do the change when requested but merging should be synchronized with RHD.

```
$ ls guides/ | grep Red
doc-Administering_Red_Hat_Satellite
doc-Installing_Proxy_on_Red_Hat
doc-Installing_Server_on_Red_Hat
```

ekohl:
These guides were essentially the same and we already have another mechanism to handle multiple builds. This unifies things which makes it easier.

This leaves the existing guides so this can be safely cherry picked to older branches without affecting downstream.

It's implemented as 2 commits. Do keep these because the only the first should be cherry picked to 3.0, 2.5 and 2.4 while both should be picked to 3.1.

Fixes #970


ekohl:
I made the mistake of reading the guides and now I got stuck in fixing things. Marking as draft since I at least made 1 mistake.

ekohl:
It escalated a bit and there's now a bunch of other fixes in there as well. I do think it's ready now. Let me know if you want me to split it into multiple PRs.

Note that it does fix things for Install Proxy with foreman-deb, but it's not listed. That's because I'm still not happy with with everything.

guides/common/assembly_installing-capsule-server.adoc
```diff
@@ -10,18 +10,15 @@ For more information, see {InstallingSmartProxyDocURL}preparing-environment-for-
...
 // Registering {SmartProxyServer} to {ProjectServer}
 include::modules/proc_registering-capsule-to-satellite-server.adoc[leveloffset=+1]
+endif::[]
 
 ifdef::satellite[]
 // Attaching the Satellite Infrastructure Subscription
 include::modules/proc_attaching-satellite-infrastructure-subscription.adoc[leveloffset=+1]
 endif::[]
 
+ifdef::foreman-el,foreman-deb,satellite[]
```
ekohl, line 20: I'm unsure about Katello and Orcharhino here. I think it should be done, but I wasn't quite sure.

ekohl:
Link checker will obviously fail since this is building a new guide which is not online yet. I wonder if CI could somehow build everything with a different base URL and serve it locally instead.

lzap:
> Link checker will obviously fail since this is building a new guide which is not online yet

Not the case anymore, I changed linkchecking so it actually does not go online, it checks local links against local directory actually. So you probably have a broken link there.

```
        - name: Build HTML with local links
          run: |
            make -j ${{ env.MAKE_J }} html BUILD=foreman-el LINKS=LOCAL
            make -j ${{ env.MAKE_J }} html BUILD=foreman-deb LINKS=LOCAL
            make -j ${{ env.MAKE_J }} html BUILD=katello LINKS=LOCAL
            make -j ${{ env.MAKE_J }} html BUILD=satellite LINKS=LOCAL
            make -j ${{ env.MAKE_J }} html BUILD=orcharhino LINKS=LOCAL
```

lzap:
Weird, if a link utilizes BaseURL attribute, then it should not go online since we build it with:

    asciidoctor -a BaseURL=file:///absolute/path/to/the/dir

lzap:
Ha I see an extra quote, maybe it does not work? Could you try to unqote it?

    bundle exec asciidoctor -a build=foreman-el "-a BaseURL=file:///home/runner/work/foreman-documentation/foreman-documentation/guides/build/" ...

ekohl:
Added a commit at  the start of the series, but now it fails with;
```
Result     Error: URLError: <urlopen error [Errno 2] No such file or directory: '/home/runner/work/foreman-documentation/foreman-documentation/guides/build/installing_capsule_server/index'>
```

This is within the Satellite build, which works slightly different in their builds. This is getting hard.

ekohl:
I chose to exclude satellite in local link checks for now. Hopefully that works.

ekohl:
:green_apple: now.

guides/common/assembly_installing-capsule-server.adoc
```diff
@@ -10,18 +10,15 @@ For more information, see {InstallingSmartProxyDocURL}preparing-environment-for-
 ifdef::satellite,katello[]
 // Registering {SmartProxyServer} to {ProjectServer}
 include::modules/proc_registering-capsule-to-satellite-server.adoc[leveloffset=+1]
+endif::[]
```
lzap, line 13: Another missed endif? I think we need a checker for that.

guides/common/assembly_installing-capsule-server.adoc
lzap: @maximiliankolb ?

web/content/js/nav.js
```diff
@@ -91,10 +91,10 @@ let navItems = [
...
 		title: "Installing Katello Server",
-		href: "/Installing_Server_on_Red_Hat/index-katello.html"
+		href: "/Installing_Server/index-katello.html"
 	},{
 		title: "Installing Smart Proxy with Content",
-		href: "/Installing_Proxy_on_Red_Hat/index-katello.html"
+		href: "/Installing_Proxy/index-katello.html"
 	},{
 		title: "Upgrading and Updating",
 		href: "/Upgrading_and_Updating/index-katello.html"
```
lzap, line 100: Do you plan to backport it down to 2.4? Could experience a conflict or two there.

lzap (approved):
LGTM

I am gonna build it locally now and read that way. I wish we had a GHA action that would upload and extract the HTML ZIP artifact to a temporary directory on a web somewhere :)

ekohl:
> I wish we had a GHA action that would upload and extract the HTML ZIP artifact to a temporary directory on a web somewhere :)
That is already in place? Go to the `Checks` tab, click `deploy` and see. For this build it's https://github.com/theforeman/foreman-documentation/actions/runs/1728946784 and there is a zip file with the HTML files.

web/content/js/nav.js
ekohl: See the initial message: it's split into two commits so you can cleanly pick one (copy the guide) and not the other (remove the old guides).

guides/common/assembly_installing-capsule-server.adoc
ekohl: No, the endif was in the old version on line 22. However, configuring repositories is not limited to satellite and katello so the ifdef here should be closed earlier. Technically it could have been on line 18 too, but this keeps it closer to the opening ifdef.

lzap:
Nitpick: not sure how `firewall-cmd` is relevant in the Debian environment.

Debian: `apt-get install grub2-efi` this is wrong, I think we install this via installer, this is useless. Can you drop this together with "On Foreman or Smart Proxy from which you provision, update the grub2-efi package to the latest version" that is handled by the installer.

Debian: drop "4.2.2. Configuring SELinux to Ensure Access to Foreman on Custom Ports"

Debian: `apt-get install dhcp bind` looks incorrect, I think these packages are not present according to Debian online package search, but it is in very edge-case chapter External DHCP/DNS so you don't need to fix this now

Other than that, all good.

lzap:
> there is a zip file with the HTML files.

Yeah I meant an action that would actually publish it online somewhere (temporarily).

guides/common/assembly_installing-capsule-server.adoc
ekohl: Now that I read through it again, I see it was previously done for Katello. I was confused by the registering part. In theory it's possible to sync content to Katello, register the Proxy to Katello and get the content that way. I just don't know if that's a desired (primary) workflow.

ekohl:
> Nitpick: not sure how `firewall-cmd` is relevant in the Debian environment.

This is a larger issue that is not limited to this guide. I completely agree with it, but as you can see I already blew up the scope by a lot.

> Debian: `apt-get install grub2-efi` this is wrong, I think we install this via installer, this is useless. Can you drop this together with "On Foreman or Smart Proxy from which you provision, update the grub2-efi package to the latest version" that is handled by the installer.

I didn't see this, but looking at it I see it in multiple guides. It's weird because they say "update the `grub2-efi` package to the latest version" and then print "{package-install-project} grub2-efi". If it wants to update, shouldn't it be `{package-update-project}` instead?

I did a few things in a separate commit. Please have a look.

> Debian: drop "4.2.2. Configuring SELinux to Ensure Access to Foreman on Custom Ports"

Added as a separate commit.

> Debian: `apt-get install dhcp bind` looks incorrect, I think these packages are not present according to Debian online package search, but it is in very edge-case chapter External DHCP/DNS so you don't need to fix this now

I noticed this too, but I decided to skip this for now.

> Yeah I meant an action that would actually publish it online somewhere (temporarily).

I've seen some frontend developers use https://surge.sh/ but I don't have time to look at it.

lzap (approved):
I am good, please merge @maximiliankolb

guides/common/assembly_installing-capsule-server.adoc
maximiliankolb: This should not be visible for orcharhino builds. Actually, I will put it on my todo to simply make the whole install/upgrade guide blank for orcharhino, as we use a different installation and upgrade process. One big ifndef in the master files should be easier to maintain for now.

guides/common/attributes-base.adoc
```diff
@@ -4,8 +4,8 @@
 :ConfiguringAnsibleDocURL: {BaseURL}Configuring_Ansible/{BaseFilenameURL}#
 :ConfiguringLoadBalancerDocURL: {BaseURL}Configuring_Load_Balancer/{BaseFilenameURL}#
 :ContentManagementDocURL: {BaseURL}Content_Management_Guide/{BaseFilenameURL}#
-:InstallingProjectDocURL: {BaseURL}Installing_Server_on_Red_Hat/{BaseFilenameURL}#
-:InstallingSmartProxyDocURL: {BaseURL}Installing_Proxy_on_Red_Hat/{BaseFilenameURL}#
+:InstallingProjectDocURL: {BaseURL}Installing_Server/{BaseFilenameURL}#
```
maximiliankolb, line 7: 👍

guides/common/modules/proc_installing-proxy-packages.adoc
```diff
@@ -1,10 +1,10 @@
-[id="configuring-foreman-proxy-repositories-el-{package-manager}"]
+[id="installing-foreman-proxy-packages-{package-manager}"]
```
maximiliankolb, line 1: IMHO this file is either lacking a heading or should be considered as a snippet (> change the file name prefix).

guides/doc-Planning_Guide/topics/Provisioning_Concepts.adoc
```diff
@@ -96,13 +96,15 @@ For HTTP booting to work, ensure that {Project} has the following configurations
 
 * The subnet that functions as the host's provisioning interface has a DHCP {SmartProxy}, a HTTP Boot {SmartProxy},  a TFTP {SmartProxy}, and a Templates {SmartProxy}
 
+ifndef::foreman-deb[]
 * The  `grub2-efi` package is updated to the latest version. To update the `grub2-efi` package to the latest version and execute the installer to copy the recent bootloader from `/boot` into `/var/lib/tftpboot` directory, enter the following commands:
```
maximiliankolb, line 100:
```suggestion
* The `grub2-efi` package is updated to the latest version.
To update the `grub2-efi` package to the latest version and execute the installer to copy the recent bootloader from `/boot` into `/var/lib/tftpboot` directory, enter the following commands:
```

guides/doc-Planning_Guide/topics/Provisioning_Concepts.adoc
```diff
@@ -140,15 +142,17 @@ Although TFTP protocol is not used for HTTP UEFI Booting, {Project} use TFTP {Sm
...
+# {package-update-project} grub2-efi
 # {foreman-installer}
 ----
+endif::[]
 
-ifeval::["{build}" != "satellite"]
+ifdef::foreman-el,katello,orcharhino[]
 === Secure Boot
 
 When {Project} is installed on {RHEL} or compatible operating systems using  `{foreman-installer}`, grub2 and shim bootloaders that are signed by Red Hat are deployed into the TFTP and HTTP UEFI Boot directory. PXE loader options named "SecureBoot" configure hosts to load `shim.efi`.
```
maximiliankolb, line 158:
```suggestion
When {Project} is installed on {RHEL} or compatible operating systems using  `{foreman-installer}`, grub2 and shim bootloaders that are signed by Red Hat are deployed into the TFTP and HTTP UEFI Boot directory.
PXE loader options named "SecureBoot" configure hosts to load `shim.efi`.
```

maximiliankolb (approved):
All changes look reasonable to me!

I made some optional suggestion, but feel free to merge. Also, please try to backport to 3.1 if possible. The PR states that it is two commits, but now it's even more (which is good!); please consider this when cherry-picking!

Thanks for your PR & the separate commits. This makes it easier to retrace in the future!

ekohl:
> Also, please try to backport to 3.1 if possible. The PR states that it is two commits, but now it's even more (which is good!); please consider this when cherry-picking!

This was indeed my intention. It started out as 2 but quickly grew.

I think your points are good, but I'd prefer to merge it now. Feel free to pick up those additional improvements yourself.

ekohl:
https://github.com/theforeman/foreman-documentation/pull/1019 is the pick to 3.1. I'll do lighter PRs to 2.4 - 3.0.

ekohl:
And to complete the list:
* #1020 for 3.0 
* #1021 for 2.5
* #1022 for 2.4

Issue #746, opened 2021-10-14
Two asciidoc attributes in one anchor/link


maximiliankolb:
Run `git grep "{.*{" guides | grep 'id='`:

    guides/common/modules/proc_configuring-repositories-el.adoc:1:[id="configuring-repositories-el-{distribution-major-version}-{package-manager}"]
    guides/common/modules/proc_installing-server-packages-el.adoc:1:[id="configuring-foreman-repositories-el-{package-manager}-{context}"]

I've fixed all other instances in https://github.com/theforeman/foreman-documentation/pull/741 > See 8c364cbd829864547e4be96d6a4a0b8901b7157d & https://github.com/theforeman/foreman-documentation/pull/741#pullrequestreview-778895635

FYI @melcorr

melcorr:
Thanks @maximiliankolb  - I will take a look .

ekohl:
Is this still an issue?

maximiliankolb:
It's specific to RHs framework to build/deliver docs. I don't know.

PR #1011, opened 2022-01-21
Build via ccutil

lzap:
So we were asked by Red Hat Documentation team to run build with `ccutil` to prevent downstream breakages, I am fine with that as long as it will be possible to run it - and it looks it is pretty easy. RHD team maintains a container with all the tools and that is the official way to run it: https://quay.io/repository/ivanhorvath/ccutil

There is a bccutil wrapper shell script that uses podman to run the tool from the CLI, but that is not possibly to utilize in GHA, luckily docker on GHA have no problems launching the image. This patch adds then, then runs 'html-single' build on all guides which contain 'docinfo.xml' file metadata. Directories without it are skipped.

And 'ccutil' actually helped to reveal a cross-reference link error, need to look into it:

```
Unknown ID or title "rhel-7-package", used as an internal cross reference
/__w/foreman-documentation/foreman-documentation/guides/doc-Installing_Proxy_on_Red_Hat/build/en-US/master.xml fails to validate
Error: Process completed with exit code 255.
```

It turns out (thanks @Lennonka) that this was an unclosed `ifdef` which caused the `ccutil` parser to error out while `asciidoctor` just ignored it. So we can see some improvement even for upstream builds since we have missed this.

During my testing, I have experienced several errors with make which was caused by requiring the incorrect CSS file - instead the one from the build directory it required the one that was in the current directory. I fixed that, I changed linking CSS to actually embedding it since we also embed JS and it makes no difference - every guide has its own CSS file anyway.


.github/workflows/deploy.yml
```diff
@@ -17,6 +17,14 @@ env:
   MAKE_J: 3
 
 jobs:
+  build-bccutil:
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/ivanhorvath/ccutil:amazing
```
ekohl, line 23: It looks like this container hasn't been updated for 2 years. If anything, wouldn't it be better to use something maintained? Perhaps maintain it ourselves?

lzap: I was asked by RHD team if it would be possible to run `ccutil` upstream because downstream the build process breaks way too often. This is the official image that is used to build all content, I can send you an internal documentation document on how to set it up if you are interested.

I see no reason not to run it if it is going to be reasonably fast, I just need to update XML metadata files in order to actually build it.

ekohl: I'm not opposed to it, but it would be great if the final commit describes what ccutil is and why we run it.

I'd also give the feedback that they may want to update their container. Perhaps that allows Quay to index the packages and gems which would mean you can actually see which software it contains.

lzap: Sure will update description, this is a draft. FYI RHD team is actually building a NG build system that should replace ccutil so for this reason there is not any further development of the tool.

guides/common/Makefile
```diff
@@ -68,7 +68,7 @@ $(IMAGES_TS): $(IMAGES)
 	@touch $(IMAGES_TS)
 
 $(DEST_HTML): $(SOURCES) $(DEPENDENCIES) $(DOCINFO_SOURCES)
-	bundle exec asciidoctor -a build=$(BUILD) $(BASEURL_ATTR) -a docinfo=shared -a docinfodir=common -a date_mdy="$(DATE_MDY)" -a date_my="$(DATE_MY)" -a docdatetime="$(DATE_MY)" -a stylesheet=$(BUILD).css -a linkcss=true -b html5 -d book --trace -o $@ $< 2>&1 | tee $@.log
+	bundle exec asciidoctor -a build=$(BUILD) $(BASEURL_ATTR) -a docinfo=shared -a docinfodir=common -a date_mdy="$(DATE_MDY)" -a date_my="$(DATE_MY)" -a docdatetime="$(DATE_MY)" -a stylesheet=$(DEST_DIR)/$(BUILD).css -b html5 -d book --trace -o $@ $< 2>&1 | tee $@.log
```
ekohl, line 71: This makes sense but I think you should ensure it's present prior to building. #1013 does that.

ekohl (commented):
This looks good to me, but could elaborate the "Build via bccutil" commit message a bit?

guides/common/Makefile
lzap, line 71: This was a regression introduced by the parallelization patch this week - CSS file needs to be compiled for each guide separately and goes into build/GUIDENAME/ so we need to take that version.

lzap: Yeah I had this change locally but then it built fine... weird. Anyway, incorporated into my patch now to prevent conflicts.

lzap: I am going even simplify it more, the phony `css` is no longer needed actually. It was a leftover.

ekohl: Oh yes, that's even better. Always great when you can achieve more with fewer lines of code :)

lzap:
All green now!

guides/common/modules/proc_installing-capsule-server-packages.adoc
```diff
@@ -35,13 +35,14 @@ ifndef::satellite[]
 ----
 # {package-install} foreman-proxy-content
 ----
+endif::[]
```
maximiliankolb, line 38: Nice job spotting this!

melcorr:
Cherry picked:
   1049e29b..ae618bc4  3.1 -> 3.1

maximiliankolb (commented):
I realize I'm too late and I am not against building upstream docs differently.

I have some questions on the container image we now use: where does it come from; who maintains it; is tag `amazing` stable; is there a Dockerfile to recreate it; why change it?

guides/common/modules/proc_installing-capsule-server-packages.adoc
ekohl: I also found this in https://github.com/theforeman/foreman-documentation/pull/1006 along with more things. I get the impression we don't really read the guides when reviewing. I also noticed many URLs are printed now instead of a nice label.

lzap:
3.1: https://github.com/theforeman/foreman-documentation/pull/1014

guides/common/modules/proc_installing-capsule-server-packages.adoc
lzap: I also spotted that we have some `xref:#xxx` when the hash should not be really required, it might even break things, but ccutil did not complain. Did not fix it yet tho.

lzap:
> building upstream docs differently.

Oh no, we are not changing how we build stuff, we just add another job to verify that downstream does not break. It apparently uses a different parser or something. We continue to use GNU Makefile for now, no change there.

> I have some questions on the container image we now use: where does it come from; who maintains it; is tag `amazing` stable; is there a Dockerfile to recreate it; why change it?

This is an internal tool of RHD and I don't think it's even open source. Can you answer this for us @Lennonka @melcorr ?

ekohl:
This is exactly what I asked for twice in the commit message. Next time I'll use the "request changes" to make that explicit.

melcorr:
AFAIK @maximiliankolb these checks that @lzap has now added ensure that we preserve items in a way that they'll build in this closed sourced and very fragile software called Pantheon that is used by Red Hat to build and publish Red Hat documentation. 
For some reason, and I suspect it has nothing to do with any magic in the mix, the packages they use are hosted internally, so there's no available container. 
This changes nothing for our processes, except that, for example, this will now fail if a double variable is used in an ID rather than me having to spot it. 
With regards to stability, I don't think these versions were updated. I can verify this though with @jhradilek

lzap:
I have just realized that the PR title was a bit misleading, so I changed it. Sorry about the confusion, I did not give it much thinking when I filed a Draft PR this morning as I thought I won't be able to get ccutil running in GHA but it was actually quite easy.

maximiliankolb:
Regarding anchors with two variables: I doubt that it works as expected. See https://github.com/theforeman/foreman-documentation/issues/746

```
$ foreman-documentation git:(master) git grep "{.*{" guides | grep 'id='
guides/common/modules/proc_configuring-repositories-el.adoc:[id="configuring-repositories-el-{distribution-major-version}-{package-manager}"]
guides/common/modules/proc_installing-server-packages-el.adoc:[id="configuring-foreman-repositories-el-{package-manager}-{context}"]
```

Yet, the latest master GHA is green.

melcorr:
@maximiliankolb wow. And normally that fails on ccutil....

PR #1013, opened 2022-01-21
Make asciidoctor depend on the CSS file

ekohl:
This CSS file is used by asciidoctor and if it isn't built, it fails.


lzap:
Yeah, I only spotted this recently. Incorporated into my patch now to prevent conflicts.

PR #1015, opened 2022-01-21
Cleanup of images and solution for common directory

lzap:
This is a cleanup of images.

Previously situation was messy - all guides had common/images/ symlink but some guides had also images/ subdirectory. Each guide then defined its own `:imagesdir:` attribute to either one, or another. Last week I made an attempt to simplify this introducing a makefile target that copies files from both directories into the destination first and then it uses that directory for images.

This works, however, it requires images to be copied first to the destination. This step should not be necessary at all (for building), only for HTML when images are symlinked but not for PDF. It also did not work with `ccutil`.

This patch aims to solve all problems by enforcing the following:

* each guide must have `images/` directory
* each guide must have `images/common` symlink into common directory

This way, both common and non-common images can be referenced easily:

* `image::test.png`
* `image::common/test.png`

Also directories can will be preserved so subdirectories can be created within both image directories as needed. Previously only flat structure was supported which can be problematic for the future.

This patch also fixes all places where common images are referenced. For the record, this command was used to find them:

```
find . -name \*.adoc -type f | xargs perl -pi -e 's/image::?(compliance_policy_dashboard.png|notifications1.png|PXEless-mode.drawio|PXEless-mode.png|PXEless-mode.svg|PXE-mode.drawio|PXE-mode.png|PXE-mode.svg|user-data-sequence-foreman.drawio|user-data-sequence-foreman.svg|user-data-sequence-satellite.png|user-data-sequence-satellite.txt|1228.png|1229.png|1230.png|[123456789].png)/image::common\/\1/g'
```

As part of this PR, I would like also to implement some kind of image chekcing, if an image does not exists it will fail the build. All tools we use in our pipeline only issue a warning, that should be definitely an error from now on.


guides/common/modules/proc_using-the-project-content-dashboard.adoc
```diff
@@ -15,13 +15,13 @@ The following table shows the descriptions of the possible configuration states.
...
 |====
 | Icon | State | Description
-| image:1.png[] | *Hosts that had performed modifications without error* | Host that successfully performed modifications during the last reporting interval.
-| image:2.png[] | *Hosts in error state* | Hosts on which an error was detected during the last reporting interval.
-| image:3.png[] | *Good host reports in the last 35 minutes* | Hosts without error that did not perform any modifications in the last 35 minutes.
-| image:4.png[] | *Hosts that had pending changes* | Hosts on which some resources would be applied but Puppet was configured to run in the `noop` mode.
-| image:5.png[] | *Out of sync hosts* | Hosts that were not synchronized and the report was not received during the last reporting interval.
-| image:6.png[] | *Hosts with no reports* | Hosts for which no reports were collected during the last reporting interval.
-| image:7.png[] | *Hosts with alerts disabled* | Hosts which are not being monitored.
+| image::common/1.png[] | *Hosts that had performed modifications without error* | Host that successfully performed modifications during the last reporting interval.
```
ekohl, line 18: Unrelated to this PR, but that's a very generic filename.

ekohl (commented):
Can't we set the imagesdir? https://docs.asciidoctor.org/asciidoc/latest/macros/images-directory/ suggests we can set `imagesdir` to `common/images` and we wouldn't need such a large patch.

guides/common/modules/proc_using-the-project-content-dashboard.adoc
lzap: Yeah I don't want to go into this rabbithole now :)

lzap:
> Can't we set the imagesdir? https://docs.asciidoctor.org/asciidoc/latest/macros/images-directory/ suggests we can set `imagesdir` to `common/images` and we wouldn't need such a large patch.

It's set to images/ so guides can put their local-related images in there while they can still reference common images via that symlink.

```
$ ag imagesdir
doc-Converting_a_Host_to_RHEL/master.adoc
2::imagesdir: images

doc-Managing_Organizations_and_Locations/master.adoc
4::imagesdir: images

doc-Upgrading_and_Updating/master.adoc
2::imagesdir: images

doc-Quickstart_Guide/master.adoc
2::imagesdir: images

doc-Planning_Guide/master.adoc
2::imagesdir: images

doc-Provisioning_Guide/master.adoc
2::imagesdir: images

doc-Managing_Hosts/master.adoc
2::imagesdir: images

doc-Installing_Server_on_Red_Hat/master.adoc
2::imagesdir: images

doc-Installing_Satellite_Server_Disconnected/master.adoc
2::imagesdir: images

doc-Installing_Proxy_on_Debian/master.adoc
1::imagesdir: images

doc-Installing_Server_on_Debian/master.adoc
1::imagesdir: images

doc-Installing_Proxy_on_Red_Hat/master.adoc
1::imagesdir: images

doc-Configuring_Load_Balancer/master.adoc
2::imagesdir: images

doc-Deploying_on_AWS/master.adoc
2::imagesdir: images

doc-Content_Management_Guide/master.adoc
2::imagesdir: images

doc-Administering_with_Ansible/master.adoc
2::imagesdir: images

doc-Configuring_Ansible/master.adoc
2::imagesdir: images

doc-Administering_Red_Hat_Satellite/master.adoc
2::imagesdir: images

doc-Application_Centric_Deployment/master.adoc
2::imagesdir: images
```

I know that we already have the common/ symlink which contains the images/ directory but we cannot set `:imagesdir: common/images` because at that moment you loose ability to put local images (only relevant for the particular guide).

A different solution would be simply using just the common dir for all images, but the reason why I actually want to fix this is our incoming Complete Guide which contains 9MB of images and I don't think we should simply dump them all into common/images. We should gradually move them as we will be reviewing content.

.github/workflows/deploy.yml
```diff
@@ -29,10 +29,10 @@ jobs:
         run: |
           for DIR in guides/doc-*; do
             pushd $DIR &>/dev/null
+            [[ -d images/ ]] || (echo "Required images/ directory missing in $DIR"; exit 1)
+            [[ -h images/common ]] || (echo "Required images/common symlink missing in $DIR"; exit 1)
```
lzap, line 33: This will fail early if we don't have these directories.

ekohl (approved):
That's a fair point. Will this also avoid copying all common images to every build?

lzap:
> Will this also avoid copying all common images to every build?

Yes, for PDFs. I have just made a last-minute change and removed the IMAGES dependency, that is not needed now.

Also updated README with further info about images.

lzap:
Green, this looks great, only took 2 minutes (HTML link checking), all other jobs are under one minute. Huge improvements there.

lzap:
FYI I will backport this into 3.1, this is needed in order to get `ccutil` building downstream again (I broke it this week).

lzap:
> As part of this PR, I would like also to implement some kind of image chekcing

Turns out image checking is not necessary, CI job build-pdf will fail if there is a missing image! Due to parallel make errors might be little hidden, I updated the `tee` command to print to standard error but GHA for some reason does not print those lines in red color. Not sure why, perhaps I need to set XTERM.

lzap:
> We currently have 41 images in total, so IMHO one common directory would be sufficient for all guides.

Yes, I agree but as I said, this will change with the Complete Guide (our legacy manual), that one brings about 80+ new images and I don't think we will like to put them all into the common directory right away. What I think is worth doing is reviewing them chapter by chapter and only then we can decide to move, replace or remove them. That's why I actually implemented this, so this patch was actually more upstream than downstream (I already did some changes and broke downstream but the reason was the same).

Good point about dupes, so I used `fdupes` tool to find these:

```
./guides/doc-Planning_Guide/images/satellite_6_system_architecture.png
./guides/doc-Planning_Guide/images/satellite_6_topology.png

./guides/doc-Planning_Guide/images/satellite_6_four_environments.png
./guides/doc-Planning_Guide/images/satellite_6_life_cycle.png
```

I am creating one more commit and deleting one of these replacing the links to refer to only single image.

lzap:
Github is drunk, I clicked on the merge button:

* 10be324a4ae2ef6c148d59c44b4e65c2a3192dde
* 0a3e3424b82e037a739a6079da4406c4abad6987
* 9db8b4f182e707f22705f4189ef04ffe5f36975f

lzap:
Backport into 3.1: https://github.com/theforeman/foreman-documentation/pull/1017

lzap:
Not sure what happened at github but it missed the push, GHA did not run this time. I did a small change in master (README fix) to trigger GHA again: cf7f5cdd

PR #1019, opened 2022-01-24
[3.1] Cherry pick remove _on_$os from install guides

ekohl:
This picks https://github.com/theforeman/foreman-documentation/pull/1006 except the removal commit. To resolve merge conflicts, it does change the URL macros to match master.


lzap:
Some directories were left behind. I fixed this in master in 16ea8ba7e055acf689321486882756860c087e85 so perhaps cherry pick it too to resolve the build problem?

If we run into missing images directories this should fix it:

```
cd guides
for D in doc-*; do mkdir images; done
for D in doc-*; do ln -sf ../common/images $D/images/common ; done
```

ekohl:
I picked that exact commit. See the last commit in the series. That commit actually introduces the problem by removing the images directory from guides. I've amended the commit to only add the missing symlinks.

lzap:
It's green, merged, thank you. I know it's a mess, hopefully we won't need to touch this anymore, it looks like structure is stable now.

PR #1022, opened 2022-01-24
[2.4] Cherry pick remove _on_$os from installation guides

ekohl:
These guides were essentially the same and we already have another
mechanism to handle multiple builds. This unifies things which makes it
easier.

This leaves the existing guides so this can be safely cherry picked to
older branches without affecting downstream.

It also updates the URL structure macros to match master. This commit is
actually a cherry pick of a cherry pick.

(cherry picked from commit de2b45611231fb091644e350424f86a88791640b)
(cherry picked from commit 963c3cb01e1fc646daf84f7e73eff758289ac6f2)
(cherry picked from commit 08e24f22495c5e57d7978695e78fc36c000e02e1)


ekohl:
With https://github.com/theforeman/foreman-documentation/pull/1066 this would not be needed.

ekohl:
Redundant now

PR #1025, opened 2022-01-25
Fix missing images directories

lzap:
SSIA

Ask for rebase before merging please, bigger changes of the Makefile and GHA can cause issues.


ekohl (approved):
My bad. I'll update the cherry picks too.

lzap:
No problem you could not see this and it was green. We need to ask to rebase only on those PRs which add new guide directories or delete old ones.

ekohl:
Normally it's not as big a deal, but we recently had many big, structural changes in flight. I should have been aware of that.

ekohl:
I think this was only needed in 3.1 since you didn't pick the images restructure to 3.0 and older, right?

PR #1029, opened 2022-01-26
Adding info on how to enable SSL to the section about configuring the…

mdolezelova:
… product to use external databases

Cherry-pick into:

* [ ] Foreman 3.1

### DIFF
--- a/guides/common/modules/proc_configuring-satellite-to-use-external-databases.adoc
+++ b/guides/common/modules/proc_configuring-satellite-to-use-external-databases.adoc
@@ -29,9 +29,9 @@ Use the `{foreman-installer}` command to configure {Project} to connect to an ex
   --foreman-proxy-content-pulpcore-postgresql-password _Pulpcore_Password_
 ----
 +
-[NOTE]
-====
+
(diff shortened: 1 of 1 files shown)


guides/common/modules/proc_configuring-satellite-to-use-external-databases.adoc
```diff
@@ -28,3 +28,19 @@ Use the `{foreman-installer}` command to configure {Project} to connect to an ex
   --foreman-proxy-content-pulpcore-postgresql-db-name pulpcore \
   --foreman-proxy-content-pulpcore-postgresql-password _Pulpcore_Password_
 ----
++
+[NOTE]
+====
```
melcorr, line 33: Can we not just put it as running text? 
If we were to add a note every time, we would have nothing but notes. 
Everyone wants a note, but not everything is noteworthy. This just looks like information.

mdolezelova: @melcorr Thank you for the review! Sure, you're right :-) I'll fix it in the next commit. Thank you!

melcorr (approved):
LGTM thank you!

PR #1046, opened 2022-02-04
Update index and content pages for Foreman 3.2

upadhyeammit:


upadhyeammit:
@lzap @melcorr is `:AccessRedHatComVersion:` required? The 3.2 manuals are not applicable for Satellite 6.10? Maybe applicable for 7.1 however that fails as well.

ekohl (changes requested):
All changes to `web/` should be made in `master`, not `3.2`. I also think you don't need to touch the attributes. It's questionable what it should be since 3.2 doesn't align to a downstream release.

lzap:
Please drop the attribute changes, and update `content/js/nav.js` with the new version to update the main menu.

Could you also update the Branching new release section of the README? Apparently it's out of date. What you have here plus nav.js minus attributes changes is what is really needed. Thanks.

upadhyeammit:
This should require web/content/release-3.2.adoc file as well as I see that in 3.1 branch?

[release-3.1.adoc](https://github.com/theforeman/foreman-documentation/blob/3.1/web/content/release-3.1.adoc)

PR #1039, opened 2022-02-02
Remake Puppet introduction

Lennonka:
Cherry-pick into:

* [x] Foreman 3.1

PR #1049, opened 2022-02-07
Move Puppet chapter to a standalone guide

Lennonka:
Change in plans. I've created the new Puppet Guide sooner than the Intro chapter is finished, because we need to make the preview link work for the Testathon (starting Monday 14th).
This can (and is preferred to) be merged before #1039 

Cherry-pick into:

* [x] Foreman 3.1


melcorr:
@maximiliankolb  I am testing this now but would appreciate your opinion also.

guides/doc-Managing_Configurations_Puppet/docinfo.xml
```diff
@@ -0,0 +1,14 @@
+<title>Managing Configurations Using Puppet Integration in Satellite</title>
+<productname>Red Hat Satellite</productname>
+<productnumber>7.0</productnumber>
```
ekohl, line 3: Probably unrelated, but in a lot of docinfo there's still 6.10. It would be good if those were all bumped to 7.0, both here and in 3.1.

ekohl (commented):
I like you split it into two separate commits. That really makes things easier for reviewers and in the history (and easier revert in the future). What really makes things great is if you add a small paragraph to it explaining why. https://cbea.ms/git-commit/#separate has some good reasoning and examples. For example, the disable link checker commit could explain why it's needed to disable the link checker. Understanding the reasoning is easier than guessing. Worse yet is when a reviewer needs to ask because then you have a longer back and forth which in turn makes a merge slower.

In this case I guess it's due to the chicken-and-egg problem where you introduce a new guide here and link to it, but it hasn't been published in downstream and can't be until this is merged.

I wonder if we can somehow tell the link checker to ignore just a single or a small set of links which wouldn't make it so blunt all or nothing. Worth a separate issue?

And to be clear, I'm very happy that we'll be one step closer to having upstream and downstream be in sync. I hate special cases.

Since this is just moving assemblies etc I won't comment on the actual content. The assembly itself looks alright to me.

melcorr (approved):
LGTM but I will leave @maximiliankolb have final say before I merge etc.

guides/doc-Managing_Configurations_Puppet/master.adoc
```diff
@@ -0,0 +1,17 @@
+include::common/attributes.adoc[]
+include::common/header.adoc[]
+:context: managing-configurations-puppet
+
+= Managing Configurations Using Puppet Integration in {Project}
```
maximiliankolb, line 5: I somewhat dislike the title.

What about `Configuring Hosts Using Puppet`? I think its both shorter and more precise.

Also, it would give us room to create two more guides like `... using Ansible` and `... using Salt`.

maximiliankolb (changes requested):
Great job @Lennonka, your PR looks very clean and tidy. For the second commit, I wish there was a bit of an explanation to _why_ we do this.

guides/doc-Managing_Configurations_Puppet/master.adoc
melcorr: I think @maximiliankolb is onto something here. I agree.

Lennonka: 1. Good point with the various guides, I was wondering about the same
2. We would like to have product name in the main title -- can we have two titles conditioned per product?
3. second commit -- I'll elaborate in the commit message

melcorr: We can have upstream and downstream titles so if Shweta doesn't like the version that is consistent with upstream, let's make two.

Lennonka: Anyway, I'm going to use conditional titles either way, because we still want to include the product name.

Lennonka: But what we need to decide for the upstream is the guide URL and I'm presuming that we prefer it in the same wording as the actual title?

guides/common/modules/con_salt-introduction.adoc
```diff
@@ -1,7 +1,7 @@
 [id="salt_guide_introduction_{context}"]
 = Introduction to Salt
 
-Salt can be used as a configuration management tool similar to {ConfiguringAnsibleDocURL}using-ansible-roles_ansible[Ansible] or {ManagingHostsDocURL}using-puppet-for-configuration-management_managing-hosts[Puppet].
+Salt can be used as a configuration management tool similar to {ConfiguringAnsibleDocURL}using-ansible-roles_ansible[Ansible] or {ManagingConfigurationsPuppetDocURL}[Puppet].
```
ndeevy, line 4: I'd suggest switching the wording around here so that it's active and user oriented instead of passive. That'd be compliant with the IBM style guide and easier for translation too. So something like:

You can use Salt as a configuration management tool similar to {ConfiguringAnsibleDocURL}using-ansible-roles_ansible[Ansible] or {ManagingHostsDocURL}using-puppet-for-configuration-management_managing-hosts[Puppet].

guides/doc-Managing_Configurations_Puppet/docinfo.xml
```diff
@@ -0,0 +1,14 @@
...
+<subtitle>Using Puppet integration to manage configurations of hosts with Puppet modules and reporting to Satellite</subtitle>
+<abstract><para>This guide describes how to enable Puppet integration with Satellite, configure the Puppet agent on hosts, how to import Puppet modules, and how to use the Puppet modules to enforce configurations on hosts managed in your Red&nbsp;Hat Satellite infrastructure.</para></abstract>
```
ndeevy, line 5: Pretty good overall, but one thing I'd change is the self-referential wording at the start. Our editor in RH always highlights that we should avoid self references like 'this chapter discusses <x>' or 'this document details <x>' because it removes the reader from their task and bring the writer talking about their guide into the equation.
So I'd change it to something more user action oriented like:

In Red&nbsp;Hat Satellite, you can enable Puppet integration, configure the Puppet agent on hosts, import Puppet modules, and use the Puppet modules to enforce configurations on hosts managed in your Red&nbsp;Hat Satellite infrastructure.

guides/doc-Managing_Configurations_Puppet/docinfo.xml
```diff
@@ -0,0 +1,14 @@
+<title>Managing Configurations Using Puppet Integration in Satellite</title>
+<productname>Red Hat Satellite</productname>
+<productnumber>7.0</productnumber>
+<subtitle>Using Puppet integration to manage configurations of hosts with Puppet modules and reporting to Satellite</subtitle>
```
ndeevy, line 4: This is a tiny bit confusing to parse, (although I'm not super familiar with Satellite).
The 'with' is problematic for translation because it's not clear if you use Puppet modules to manage configurations, or you manage configurations that have Puppet modules. I think it's the latter but a machine translator probably wouldn't know which.

The ending 'and reporting to satellite' seems tacked on and the verb is in a different tense (present continuous instead of infinitive like the previous verb). 

This suggestion might not be accurate so take with a pinch of salt:

`Using Puppet integration to manage and report configurations and <and_whatever_is_being_reported> of hosts that have Puppet modules.`

If that suggestion is no good, I still suggest to make it clearer based on the issues I mention here.

ndeevy:
Review done @Lennonka 
any Qs, give me a shout!

guides/doc-Managing_Configurations_Puppet/docinfo.xml
Lennonka: Okay, just for the record: We've decided to keep this for now, so the wording is consistent with the other guides on RH Customer Portal. Noting for future improvements.

guides/doc-Managing_Configurations_Puppet/master.adoc
melcorr: Avoiding titles where possible in the URL is the best approach. It makes it vendor neutral and this is our desired end state.

Lennonka: @melcorr I've added Max's title and could we perhaps merge this now, please?

melcorr: Your wish is my command.

Lennonka:
@melcorr Absolutely, but that's not a priority at the mo. All we need is love and a nightly build preview link.

melcorr:
I have cherry-picked to 3.1 for the Satellite 7 release: 
   560dd00e..89b02285  3.1 -> 3.1

PR #1054, opened 2022-02-09
Updated the attributes-base.adoc file for Sat 6.11

mdolezelova:
Cherry-pick into:

* [x] Foreman 3.1


mdolezelova:
Please, don't review this PR now. It's still WIP. Thank you!

guides/common/attributes-base.adoc
```diff
@@ -29,6 +29,22 @@
...
+//:RepoRHEL7ServerSoftwareCollections: rhel-server-rhscl-7-rpms - still valid??
+:RepoRHEL7ServerSatelliteServerProductVersion: rhel-7-server-satellite-capsule-7.0-rpms
+:RepoRHEL8ServerSatelliteServerProductVersion: satellite-7.0-for-rhel-8-x86_64-rpms
+:RepoRHEL7ServerSatelliteCapsuleProductVersion: rhel-7-server-satellite-capsule-7.0-rpms
+:RepoRHEL8ServerSatelliteCapsuleProductVersion: satellite-capsule-7.0-for-rhel-8-x86_64-rpms
+:RepoRHEL7ServerSatelliteMaintenanceProductVersion: rhel-7-server-satellite-maintenance-7.0-rpms
+:RepoRHEL8ServerSatelliteMaintenanceProductVersion: satellite-maintenance-7.0-for-rhel-8-x86_64-rpms
+// Satellite-Tools will become Satellite-Client ???
+// :RepoRHEL7ServerOptional: any equivalent for Sat7???
+// Ansible ??? - not mentioned i nthe doc with repo IDs
```
melcorr, line 45: Ansible is used in other guides but defined here with the rest of them. One stop shop!

melcorr: PS: Not reviewing, just giving context.

ekohl:
> Please, don't review this PR now. It's still WIP. Thank you!

In GitHub you can open a draft PR. https://github.blog/2019-02-14-introducing-draft-pull-requests/ describes the feature and is intended to signal this. I've now converted your PR to a draft. When it's ready you can press the big "Ready for review" button.

guides/common/attributes-base.adoc
```diff
@@ -29,6 +29,24 @@
 :RepoRHEL7ServerSatelliteServerPuppetVersion: rhel-7-server-satellite-6.3-puppet4-rpms
 :RepoRHEL7ServerSatelliteCapsulePuppetVersion: rhel-7-server-satellite-capsule-6.3-puppet4-rpms
 
+//Added by Marie
+:SatelliteSub: Red Hat Satellite Infrastructure Subscription
+// :RepoRHEL7Server: rhel-7-server-satellite
+// :RepoRHEL8Server: ???
+//:RepoRHEL7ServerSoftwareCollections: rhel-server-rhscl-7-rpms - still valid??
```
ekohl, line 36: To answer the question: on RHEL 7 we'll continue to use SCLs AFAIK so RH SCL is still needed. On EL8 there are no SCLs so there it will be irrelevant.

mdolezelova: Thank you, @ekohl ! So I'll uncomment the RHSCL piece. In the meantime,  I'm also reaching out to particular stakeholders for other pieces of information that are missing. Thank you!

guides/common/attributes-base.adoc
```diff
@@ -29,6 +29,26 @@
...
+:RepoRHEL7ServerSatelliteServerProductVersion: rhel-7-server-satellite-capsule-7.0-rpms
+:RepoRHEL8ServerSatelliteServerProductVersion: satellite-7.0-for-rhel-8-x86_64-rpms
+:RepoRHEL7ServerSatelliteServerProductVersionPrevious: rhel-server-7-satellite-6.10-rpms
+:RepoRHEL7ServerSatelliteCapsuleProductVersion: rhel-7-server-satellite-capsule-7.0-rpms
+:RepoRHEL8ServerSatelliteCapsuleProductVersion: satellite-capsule-7.0-for-rhel-8-x86_64-rpms
+:RepoRHEL7ServerSatelliteMaintenanceProductVersion: rhel-7-server-satellite-maintenance-7.0-rpms
+:RepoRHEL8ServerSatelliteMaintenanceProductVersion: satellite-maintenance-7.0-for-rhel-8-x86_64-rpms
+// Satellite-Tools will become Satellite-Client ??? - I haven't found this repo mentioned anywhere except for this file. So it may not be needed.
+// Repos relevant for Puppet - The Puppet repos were not mentioned in 6.10 docs. 
+
```
lzap, line 51: We have a file attributes-satellite which is included when BUILD=satellite. I wonder if it should all have been there instead of the attributes-base.adoc.

I am assuming that ll of these attributes are only used in "if satellite" statements of course, otherwise they would be rendered directly into text (not expanded).

mdolezelova: Hi @lzap , yes, it sounds logical. TBH, I'm no sure why it was set like this. I'd need to take a closer look, which may take some time.

Lennonka: I totally agree with lzap, I was about to suggest the same.

mdolezelova: Hi @lzap , so the attributes-base.adoc file says that these attributes are used in both Satellite and Katello docs. I'm not sure  which problems it could cause if we just move them under attributes-satellite.adoc (and maybe we would need to copy also under attributes-katello.adoc). Would it be an issue to keep them in the attributes-base.adoc file? Thank you!

ekohl: AFAIK there are some common ones. I think these are:
```
:RepoRHEL7Server: rhel-7-server-rpms
:RepoRHEL7ServerOptional: rhel-7-server-optional-rpms
:RepoRHEL7ServerAnsible: rhel-7-server-ansible-2.9-rpms
```

Those make are needed to install Katello on upstream using RHEL 7.

However, the `RepoRHEL7ServerSatellite*` ones are only used by Satellite. So I'd move all those to the Satellite guide. If any upstream source uses it, that's a bug. `SatelliteSub` is another that I suspect is only needed in the Satellite case.

Does that help?

guides/common/attributes-base.adoc
```diff
@@ -29,6 +29,44 @@
...
 :RepoRHEL7ServerSatelliteCapsulePuppetVersion: rhel-7-server-satellite-capsule-6.3-puppet4-rpms
 
+//Added by Marie for GA
+:SatelliteSub: Red Hat Satellite Infrastructure Subscription
+:RepoRHEL7Server: rhel-7-server-satellite
+:RepoRHEL8Server: rhel-8-for-x86_64-baseos-rpms 
+// We would probably need to mention also "rhel-8-for-x86_64-appstream-rpms"
+:RepoRHEL7ServerSoftwareCollections: rhel-server-rhscl-7-rpms
+:RepoRHEL7ServerOptional: rhel-7-server-optional-rpms
+// :RepoRHEL7ServerOptional: any equivalent for Sat7 for RHEL 8? - "codeready-builder-for-rhel-8-x86_64-rpms" ?
```
evgeni, line 39: no equivalent, and shouldn't be required.

guides/common/attributes-base.adoc
```diff
@@ -29,6 +29,44 @@
 :RepoRHEL7ServerSatelliteServerPuppetVersion: rhel-7-server-satellite-6.3-puppet4-rpms
 :RepoRHEL7ServerSatelliteCapsulePuppetVersion: rhel-7-server-satellite-capsule-6.3-puppet4-rpms
 
+//Added by Marie for GA
+:SatelliteSub: Red Hat Satellite Infrastructure Subscription
+:RepoRHEL7Server: rhel-7-server-satellite
+:RepoRHEL8Server: rhel-8-for-x86_64-baseos-rpms 
+// We would probably need to mention also "rhel-8-for-x86_64-appstream-rpms"
```
evgeni, line 36: correct, also for EL8Server, I would call it `RepoRHEL8BaseOS` as there is no more "server" repo as such (there is no difference between server and workstation in el8).

guides/common/attributes-base.adoc
evgeni: agree, yeah, there are repos you need to deploy *any* variant on RHEL which should go to common attributes, and then there are satellite specific ones, which should go to satellite attributes.

guides/common/attributes-base.adoc
```diff
@@ -29,6 +29,44 @@
 :RepoRHEL7ServerSatelliteServerPuppetVersion: rhel-7-server-satellite-6.3-puppet4-rpms
 :RepoRHEL7ServerSatelliteCapsulePuppetVersion: rhel-7-server-satellite-capsule-6.3-puppet4-rpms
 
+//Added by Marie for GA
```
evgeni, line 32: this block does clash with what we have in lines 16 to 27 -- should it replace it instead?

mdolezelova: Hi @evgeni , you're right. I meant to remove the old attributes used for 6.10 later. I agree it looks confusing so I've gone ahead and removed them. Thank you!

guides/common/attributes-base.adoc
```diff
@@ -13,21 +13,49 @@
...
-:RepoRHEL7ServerSatelliteToolsProductVersion: rhel-7-server-satellite-tools-6.10-rpms
-:RepoRHEL7ServerSatelliteMaintenanceProductVersion: rhel-7-server-satellite-maintenance-6-rpms
+:RepoRHEL8ServerAnsible: ansible-2.9-for-rhel-8-x86_64-rpms 
+:RepoRHEL7ServerSatelliteServerProductVersion: rhel-7-server-satellite-capsule-7.0-rpms
+:RepoRHEL8ServerSatelliteServerProductVersion: satellite-7.0-for-rhel-8-x86_64-rpms
+:RepoRHEL7ServerSatelliteServerProductVersionPrevious: rhel-server-7-satellite-6.10-rpms
+:RepoRHEL7ServerSatelliteCapsuleProductVersion: rhel-7-server-satellite-capsule-7.0-rpms
+:RepoRHEL8ServerSatelliteCapsuleProductVersion: satellite-capsule-7.0-for-rhel-8-x86_64-rpms
+:RepoRHEL7ServerSatelliteMaintenanceProductVersion: rhel-7-server-satellite-maintenance-7.0-rpms
+:RepoRHEL8ServerSatelliteMaintenanceProductVersion: satellite-maintenance-7.0-for-rhel-8-x86_64-rpms
```
ekohl, line 33: Once you move them to `attributes-satellite.adoc`, could you use `{ProductVersion}` instead of `7.0`?

guides/common/attributes-base.adoc
```diff
@@ -13,21 +13,49 @@
...
+// Satellite-Tools will become Satellite-Client ??? - I haven't found this repo mentioned anywhere except for this file. So it may not be needed. The below was mentioend in 6.10 docs.
+// :RepoRHEL7ServerSatelliteToolsProductVersion: rhel-7-server-satellite-tools-6.10-rpms
+// Repos relevant for Puppet - The below was mentioend for 6.10 docs.
 // Do not update the puppet4 repo versions. They must stay at 6.3.
-:RepoRHEL7ServerSatelliteServerPuppetVersion: rhel-7-server-satellite-6.3-puppet4-rpms
-:RepoRHEL7ServerSatelliteCapsulePuppetVersion: rhel-7-server-satellite-capsule-6.3-puppet4-rpms
+// :RepoRHEL7ServerSatelliteServerPuppetVersion: rhel-7-server-satellite-6.3-puppet4-rpms
+// :RepoRHEL7ServerSatelliteCapsulePuppetVersion: rhel-7-server-satellite-capsule-6.3-puppet4-rpms
```
ekohl, line 39: This macro is not used anywhere in this repository. Is there any downstream-specific guide that uses it? I can imagine that now the Puppet guide is upstreamed it's no longer needed.

guides/common/attributes-base.adoc
```diff
@@ -13,21 +13,49 @@
...
+//:SatelliteSub: Red Hat Satellite Infrastructure Subscription
+//:RepoRHEL7Server: rhel-7-server-satellite
+//:RepoRHEL8Server: rhel-8-for-x86_64-baseos-rpms 
+// We would probably need to mention also "rhel-8-for-x86_64-appstream-rpms"
+//:RepoRHEL7ServerSoftwareCollections: rhel-server-rhscl-7-rpms
+//:RepoRHEL7ServerOptional: rhel-7-server-optional-rpms
+//:RepoRHEL7ServerOptional: any equivalent for Sat7 for RHEL 8? - "codeready-builder-for-rhel-8-x86_64-rpms" ?
+//:RepoRHEL7ServerAnsible: rhel-7-server-ansible-2.9-rpms
+//:RepoRHEL8ServerAnsible: ansible-2.9-for-rhel-8-x86_64-rpms 
+//:RepoRHEL7ServerSatelliteServerProductVersion: rhel-server-7-satellite-7-beta-rpms
```
ekohl, line 52: What I would suggest is
```suggestion
// Set to beta- when this version is in beta
:RepoSatelliteBeta: beta-
:RepoRHEL7ServerSatelliteServerProductVersion: rhel-server-7-satellite-7-{RepoSatelliteBeta}rpms
```

Then when the beta is over, you simply empty the `RepoSatelliteBeta` variable and there's no need to modify all the other macros.

guides/common/attributes-base.adoc
```diff
@@ -13,21 +13,9 @@
 :UpgradingDocURL: {BaseURL}Upgrading_and_Updating/{BaseFilenameURL}#
 :ReleaseNotesURL: {BaseURL}Release_notes/{BaseFilenameURL}#
 
+
 // Repositories and subscriptions (used both in Katello and Satellite guides)
-:SatelliteSub: Red Hat Satellite Infrastructure Subscription
-:RepoRHEL7Server: rhel-7-server-rpms
-:RepoRHEL7ServerSoftwareCollections: rhel-server-rhscl-7-rpms
-:RepoRHEL7ServerOptional: rhel-7-server-optional-rpms
-:RepoRHEL7ServerAnsible: rhel-7-server-ansible-2.9-rpms
```
ekohl, line 21: It was suggested to keep these in here since these are also used for upstream installations on RHEL 7 in all scenarios, so base is really the correct place.

guides/common/attributes-katello.adoc
```diff
@@ -5,3 +5,49 @@
 :installer-log-file: /var/log/foreman-installer/katello.log
 :project-installation-guide-title: Installing Foreman {ProjectVersion} server with Katello {KatelloVersion} plugin on Enterprise Linux
 :smartproxy_port: 9090
+
+//Added by Marie for GA
```
ekohl, line 9: I don't think you need to add these to katello since they're either unused or should be in `-base`.

guides/common/attributes-base.adoc
mdolezelova: OK, yes confirmed this attribute is no longer needed. I removed it. Thank you, @ekohl !

guides/common/attributes-base.adoc
mdolezelova: Yes. I've moved the all these attributes to "attributes-satellite.adoc" and "attributes-katello.adoc". I've tried to apply the {ProductVersion} attribute where possible. Thank you, @ekohl !

guides/common/attributes-base.adoc
mdolezelova: @evgeni , thank you! I've removed this note.

guides/common/attributes-base.adoc
mdolezelova: Thank you, @ekohl ! LGTM , but actually I'm not sure whether this would work. All beta repos include just "7"  (not 7.0) and the {ProductVersion} needs to be set to "7.0".  Also, the GA repos mostly include 7.0, so the automatic replacement would maybe require some more new attributes.

guides/common/attributes-katello.adoc
mdolezelova: Oh, I've added them because the original file used in 2.5 includes statement that these repo attributes are relevant for Satellite and Katello.

guides/common/attributes-base.adoc
mdolezelova: Ok, got it. So I've moved these to attributes-base.adoc a removed from attributes-satellite.adoc. Also removed all I've added into attributes-katello.adoc. So hope it's correct now :-) Thank you for your guidance!

guides/common/attributes-base.adoc
ekohl: I recall we still had some discussion about 7.0 vs 7 naming in repo names. @ehelms did we make a decision on that already?

ekohl: Also, looking more closely, it's really `7.0` vs `7-beta`. So you could also name it `RepoSatelliteVersion` and use that instead. My point is that you'd want a single place to modify it so the change from beta -> non-beta is as small as it can be.

ekohl: Wouldn't this work for both cases?
```suggestion
// For beta
:RepoSatelliteVersion: 7-beta
// For a release
:RepoSatelliteVersion: {ProductVersion}

:RepoRHEL7ServerSatelliteServerProductVersion: rhel-server-7-satellite-{RepoSatelliteVersion}-rpms
```

guides/common/attributes-satellite.adoc
```diff
@@ -89,26 +89,23 @@
...
-:RepoRHEL8ServerSatelliteCapsuleProductVersion: satellite-capsule-{ProductVersion}-for-rhel-8-x86_64-rpms
-:RepoRHEL7ServerSatelliteMaintenanceProductVersion: rhel-7-server-satellite-maintenance-{ProductVersion}-rpms
-:RepoRHEL8ServerSatelliteMaintenanceProductVersion: satellite-maintenance-{ProductVersion}-for-rhel-8-x86_64-rpms
+
+// For beta
+:RepoSatelliteVersion: 7-beta
+:BaseServerRepoID: server-7
+// For a release
+:RepoSatelliteVersion: {ProductVersion}
+:BaseServerRepoID: 7-server
```
ekohl, line 99: If you're going to use this, shouldn't this be
```suggestion
:BaseServerRepoID: rhel-7-server
```

guides/common/attributes-base.adoc
mdolezelova: @ekohl, nice and easy! Thank you! :-) I applied it. However, this are couple more issues. While GA repos for RHEL 7 in all cases use "7-server", the Beta repos mostly use "server-7". So I introduced one more attribute to resolve this. However, the names of Beta repos in the source document are even not consistent - some use "server-7" and some other "7-server". I assume this is just a mistake in the source document, and all Beta repos should be called consistently. I've already asked that in the source document with repo IDs. If it's just a mistake, and all Beta repos should be named consistently, then my current solution should work.  If it's not mistake and Beta repo IDs are really inconsistent, then we need to go back to the previous solution - so two blocks of text - one that would apply for Beta and another one for GA.

guides/common/attributes-satellite.adoc
mdolezelova: @ekohl, maybe yes, but I dropped this attribute in the last commit. After my alert related to the inconsistency of naming Beta and GA repos, it has been unified, so an attribute to distinguish between Beta and GA is no longer needed.

guides/common/attributes-base.adoc
mdolezelova: @evgeni , thank you! I agree! I've updated the name of the attribute for BaseOS and also added a new attribute for AppStream. This is done in the "attributes-base.adoc" file.

guides/common/attributes-satellite.adoc
ekohl: And that's why it's great to get this work going sooner rather than later. Now it was still possible to fix. Glad it got fixed, makes everyone's lives easier.

mdolezelova: Definitely :-)

upadhyeammit:
@mdolezelova can we fast track the work on this pull request? This will help others to add changes in efficient way

mdolezelova:
> @mdolezelova can we fast track the work on this pull request? This will help others to add changes in efficient way

@upadhyeammit yes, apologies. I was busy with other work. It has already been ready to go for 7.0 but then the change to 6.11 happened, so I'll need to update for 6.11 as well. Apologies. I'll take a look ASAP.

mdolezelova:
@melcorr @ekohl I've finally updated the attributes for Satellite for version 6.11. Please could you review and merge this PR if all is okay? Thank you!

PR #1057, opened 2022-02-14
Restore linkchecker but disable linkchecker for Satellite.

melcorr:
This check was disabled to allow for a new guide to build
without failing checks.

This commit restores the checks.

Cherry-pick into:

* [ ] Foreman 3.1

### DIFF
--- a/guides/Makefile
+++ b/guides/Makefile
@@ -19,13 +19,11 @@ clean:
 		$(MAKE) -s --directory="$$DIR" clean ; \
 	done
 
-linkchecker:
-	echo "do not check links"
-	# find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern
+linkchecker:	
+	find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern
 
 linkchecker-tryer:
-	echo "do not check links"
-	# find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern | linkchecker-tryer
+	find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern | linkchecker-tryer
 
 serve:
 	python -m http.server --directory ./build 8813
--- a/guides/common/attributes-base.adoc
+++ b/guides/common/attributes-base.adoc
@@ -7,7 +7,7 @@
 :InstallingProjectDocURL: {BaseURL}Installing_Server/{BaseFilenameURL}#
 :InstallingSmartProxyDocURL: {BaseURL}Installing_Proxy/{BaseFilenameURL}#
 :ManagingHostsDocURL: {BaseURL}Managing_Hosts/{BaseFilenameURL}#
-:ManagingConfigurationsPuppetDocURL: {BaseURL}Managing_Configurations_Using_Puppet/{BaseFilenameURL}#
+:ManagingConfigurationsPuppetDocURL: {BaseURL}Managing_Configurations_Puppet/{BaseFilenameURL}#
 :ManagingOrganizationsLocationsDocURL: {BaseURL}Managing_Organizations_and_Locations/{BaseFilenameURL}#
 :PlanningDocURL: {BaseURL}Planning_Guide/{BaseFilenameURL}#
 :ProvisioningDocURL: {BaseURL}Provisioning_Guide/{BaseFilenameURL}#
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -83,7 +83,6 @@ jobs:
           make -j ${{ env.MAKE_J }} html BUILD=foreman-el LINKS=LOCAL
           make -j ${{ env.MAKE_J }} html BUILD=foreman-deb LINKS=LOCAL
           make -j ${{ env.MAKE_J }} html BUILD=katello LINKS=LOCAL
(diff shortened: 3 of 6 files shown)


melcorr:
It found that we had wrong values entered for the Puppet guide in attribute base!
The 6.10 Puppet guide does not exist, so this will break the links for Satellite also. 
We need some exception for a few weeks for those links.

melcorr:
Oooh I found the issue. Linkchecker is good.

guides/common/attributes-base.adoc
```diff
@@ -7,7 +7,7 @@
 :InstallingProjectDocURL: {BaseURL}Installing_Server/{BaseFilenameURL}#
 :InstallingSmartProxyDocURL: {BaseURL}Installing_Proxy/{BaseFilenameURL}#
 :ManagingHostsDocURL: {BaseURL}Managing_Hosts/{BaseFilenameURL}#
-:ManagingConfigurationsPuppetDocURL: {BaseURL}Managing_Configurations_Using_Puppet/{BaseFilenameURL}#
+:ManagingConfigurationsPuppetDocURL: {BaseURL}Managing_Configurations_Puppet/{BaseFilenameURL}#
```
ekohl, line 10: Shouldn't it be different only in the Satellite attributes document?

melcorr: https://docs.theforeman.org/nightly/Managing_Configurations_Puppet/index-foreman-el.html - this is what is live on our site.

melcorr:
There are several broken links throughout the Managing Hosts guide with bad links to Puppet that are now gone.

melcorr:
I fixed some syntax I found and also @maximiliankolb - just an FYI - the config file examples in your Salt docs have Puppet examples. I don't know if this is what you want or not :o)

guides/common/attributes-base.adoc
ekohl: Ah ok, that makes sense.

guides/Makefile
```diff
@@ -19,13 +19,11 @@ clean:
 		$(MAKE) -s --directory="$$DIR" clean ; \
 	done
 
-linkchecker:
-	echo "do not check links"
-	# find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern
+linkchecker:	
+	find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern
```
ekohl, line 23: https://linkchecker.github.io/linkchecker/man/linkchecker.html has `linkchecker --ignore-url=/secret` as an example. I wonder if we can somehow use that. Like `--ignore-url=https://access.redhat.com/documentation/en-us/red_hat_satellite/`. What I tried:
```suggestion
	find build -type f -name index\*html | xargs linkchecker --ignore-url=${IGNORE_RH_SAT_DOCS} -r1 -f common/linkchecker.ini --check-extern
```

And at line 3 `IGNORE_RH_SAT_DOCS = https://access.redhat.com/documentation/en-us/red_hat_satellite/`. That made the build pass for me. It could even be more specific and only exclude `https://access.redhat.com/documentation/en-us/red_hat_satellite/6.10/html-single/managing_configurations_using_puppet_integration/index` of course.

Then we should remember to remove that in branches that do map to Satellite releases though.

melcorr:
The RH docs have the Puppet guide created that is not published anywhere so Satellite links will fail until it is published.
They'll also start to add Satellite attributes that will break links, such as 7.0, which doesn't exist. 
Because of that, I've disabled link checking for Satellite until after the release.

guides/Makefile
melcorr: For the linkchecker example, it is probably for a permanent issue.  Since we won't use this permanently, it will require enabling and disabling, so I see little value in figuring this out.

ekohl: Oh, we already have `guides/common/linkchecker.ini` with a list of things to ignore. The easiest would be to add `access.redhat.com/documentation/en-us/red_hat_satellite/6.10/html-single/managing_configurations_using_puppet_integration/index` there. That should be also added to the "adding a new guide"-guide because this is something that will keep coming back.

melcorr: I said in my comment below that there will be a wave of various things that will break the Satellite build in the coming weeks. 
We can either make one change here, I open a ticket and assign to RH docs to readd on GA. Or we could make several changes.
 I don't see the value. Am happy to close this PR and you can do what you see fit.

ekohl: Sounds good to me: I opened https://github.com/theforeman/foreman-documentation/pull/1063 as an alternative.

melcorr: Great that you have so much time for all these things. Thank you!

ekohl: And thanks for fixing the (un)related broken links. I think in the past I've noticed a few others where there was no "label" (is it called that?). I mean the part in `[]`.

lzap:
So here is my proposal: how about we keep checking links even for satellite builds (BUILD=local) that will check all non-absolute links (e.g. links which are not http://access.redhat.com) and since this site is constantly cause issues to us, we would exclude `access.redhat.com` via `linkchecker.ini`? I feel like it causes more pain than benefit, those cross-guide links which are done correctly (via attributes) will be still checked via the local build feature (it overrides base URL so its actually a local link `file:///working/directory`) therefore only links which are direct to access.redhat.com would be really ignored.

PR #1060, opened 2022-02-14
Fixed the restorecon command in the section about restoring from a fu…

mdolezelova:
…ll backup

Cherry-pick into:

* [ ] Foreman 3.1


mdolezelova:
> This is still in master too. It would be better to first fix it in master and then cherry pick it in all the branches. Otherwise it'll continue to be a problem in all current and future versions.

Hi @ekohl , thank you! I'll create a PR for master as well. I was told that cherry picking to old branches usually doesn't work because  they're already far behind, so a separate PR for each branch is needed. Maybe cherry-picking would work for this one, but I wasn't sure so I've created a separate PR for each branch to be on the safe side. So do you think you could merge this one, please? Thank you!

ekohl (approved):
Generally I always prefer to start with newest and then work to oldest. I'd also note that rules 2 & 5 from https://cbea.ms/git-commit/#seven-rules. For now I'll merge this one but keep that in mind for the future.

PR #1063, opened 2022-02-14
Re-enable linkchecker and fix broken URLs

ekohl:
This is an alternative to https://github.com/theforeman/foreman-documentation/pull/1057 which takes some commits, but also does something different.

Rather than disabling the linkchecker because something is unavailable in downstream, this takes the approach of ignoring those broken links.

### DIFF
--- a/guides/common/modules/proc_salt-salt-variables.adoc
+++ b/guides/common/modules/proc_salt-salt-variables.adoc
@@ -4,4 +4,4 @@
 You can configure Salt Variables within {Project}.
 The configured values are available as Salt Pillar data.
 
-For more information, see {ManagingHostsDocURL}puppet_guide_configuring_a_puppet_class_{context}[Puppet smart class parameters] section on how to configure the variables' data.
+For more information, see {ManagingConfigurationsPuppetDocURL}using-puppet-smart-parameters_managing-configurations-puppet[Configuring Smart Class Parameters] in the _Puppet Guide_ on how to configure the variables' data.
--- a/guides/Makefile
+++ b/guides/Makefile
@@ -20,12 +20,10 @@ clean:
 	done
 
 linkchecker:
-	echo "do not check links"
-	# find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern
+	find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern
 
 linkchecker-tryer:
-	echo "do not check links"
-	# find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern | linkchecker-tryer
+	find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern | linkchecker-tryer
 
 serve:
 	python -m http.server --directory ./build 8813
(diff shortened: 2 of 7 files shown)


guides/common/modules/proc_salt-salt-variables.adoc
```diff
@@ -4,4 +4,4 @@
 You can configure Salt Variables within {Project}.
 The configured values are available as Salt Pillar data.
 
-For more information, see {ManagingHostsDocURL}puppet_guide_configuring_a_puppet_class_{context}[Puppet smart class parameters] section on how to configure the variables' data.
+For more information, {ManagingConfigurationsPuppetDocURL}using-puppet-smart-parameters_managing-configurations-puppet[Configuring Smart Class Parameters] in the _Puppet Guide_ on how to configure the variables' data.
```
maximiliankolb, line 7:
```suggestion
For more information, see {ManagingConfigurationsPuppetDocURL}using-puppet-smart-parameters_managing-configurations-puppet[Configuring Smart Class Parameters] in the _Puppet Guide_ on how to configure the variables' data.
```

lzap:
So here is my proposal: how about we keep checking links even for satellite builds (BUILD=local) that will check all non-absolute links (e.g. links which are not http://access.redhat.com) and since this site is constantly cause issues to us, we would exclude `access.redhat.com` via `linkchecker.ini`? I feel like it causes more pain than benefit, those cross-guide links which are done correctly (via attributes) will be still checked via the local build feature (it overrides base URL so its actually a local link `file:///working/directory`) therefore only links which are direct to access.redhat.com would be really ignored.

lzap (approved):
I am fine with this.

ekohl:
I was actually thinking about somewhere in between.

First of all, I think we should bump the Satellite version to 7.1 in our master branch. That means the URLs become `https://access.redhat.com/documentation/en-us/red_hat_satellite/7.1`. These should be ignored in our link checker.

Then when we have a branch that becomes a Satellite release, we should obviously set the correct Satellite version there too. In this case 7.0. Once the docs have been published, we can enable the linkchecker once RH's docs writers feel it's ready.

This would allow us to truly look to the future and become an upstream-first style project.

lzap:
> First of all, I think we should bump the Satellite version to 7.1 in our master branch. That means the URLs become `https://access.redhat.com/documentation/en-us/red_hat_satellite/7.1`. These should be ignored in our link checker.

Great idea, I like that. This would actually solve all those links, well, until we bump it to 7.2 and only then we would fix all of the remaining links in a controlled manner (single PR).

> Then when we have a branch

Oh yeah you describe exactly my thoughts. Good approach.

I am fine merging this as-is or if you want to do the suggested change in this PR to.

lzap:
Would be good idea to drop a comment both in attributes and linkchecker.ini so we know these two are connected and that we would bump them together.

ekohl:
I was thinking of doing that in a separate PR. We already have https://github.com/theforeman/foreman-documentation/pull/1054 and it would conflict with that. I'll create an issue that describes the workflow so we can get on the same page and see who does what.

maximiliankolb:
Still an open issue; apart from that 👍

guides/common/modules/proc_salt-salt-variables.adoc
ekohl: Fixed

ekohl:
@maximiliankolb can this be merged now?

PR #1066, opened 2022-02-15
Differentiate guides by version

ekohl:
When a guide is added, it is not always included in all older versions. This changes the navigation to be version aware. This allows only showing certain builds per version and certain guides per build.

For example, prior to Foreman 3.1 the Foreman on EL/Deb guides were unreviewed. This chooses not to show them, but does show the Katello guides since those were considered production starting Foreman 2.4.

Another example is the Release Notes guide which was introduced in nightly and then picked to all 3.x versions. This change allows not presenting the link in the 2.x guides.

A future step is to collect the per version builds dynamically based on which branches were built. This means the version is really maintained inside that branch rather than everything inside master. That makes it possible to cherry pick a new guide including the link. Whenever the cherry pick is merged, the navigation is correctly updated. However, that is not part of this.

A note for reviewers:
* I verified the navigation still works on Firefox
* The various guides in versions.js are not carefully reviewed. I copied the current guides in a structure, then removed Foreman from 2.x and the release notes. Please review carefully.

Another option is to merge this, wait for it to be deployed and run the linkchecker against docs.theforeman.org to see which guides were actually not published.


lzap:
Nice this should have been implemented from the day one but I was very frustrated with my little JavaScript knowledge and I laded on a simplified solution. One comment: everytime we change the JS we need also to regenerate each branch (empty commit is fine) so asciidoctor includes the new JS in all guides.

Feel free to merge @ekohl if you have time to push into stables and fix possible issues. I can find some time next week tho. But this is a great change.

ekohl:
I've also been a bit occupied with other matters. I'll see what I can do.

ekohl:
Ok, I just realized that it doesn't work in my test env since it can't load `/js/nav.js` locally when I use `make html browser` in a guide. I think this is ready for another round of review.

I also updated this to include 3.2 and drop all non-Katello guides from the non-nightly landing pages. Essentially this PR now only keeps Katello guides on non-nightly and Foreman guides only show up for nightly. That means we can probably drop the experimental label soonish after merging this.

lzap (approved):
It works as expected let's merge.

We need a script/makefile in the root folder that will:

* build guides
* build web
* copy all artifacts to a single directlry
* serve the directory

That would be good for testing. I will build something this is pain to test.

lzap:
So the top bar does not render now, we need to push into stable branches to rebuild its contents.

lzap:
https://github.com/theforeman/foreman-documentation/pull/1133

lzap:
So I pushed to all stable branches, looks like there is a problem:

https://docs.theforeman.org/3.2/Release_notes/index-katello.html

The navbar does not render.

lzap:
Oh damn, we need to backport this patch to all branches in order to fix this. Damn.

lzap:
I am backporting 3.2 now.

lzap:
Fixed 3.2 and pushing into all stable branches too.

PR #1069, opened 2022-02-16
Fix Puppet link

Lennonka:
This will be useful once Puppet guide is published on RH Customer Portal.

Cherry-pick to:

* [x] Foreman 3.1

### DIFF
--- a/guides/doc-Configuring_Load_Balancer/topics/Promoting_SCAP_Content_to_Clients.adoc
+++ b/guides/doc-Configuring_Load_Balancer/topics/Promoting_SCAP_Content_to_Clients.adoc
@@ -40,9 +40,5 @@ Do not run the Puppet agent on every client manually because the Puppet agent ru
 ----
 
 .Additional Resources
-
-ifdef::satellite[]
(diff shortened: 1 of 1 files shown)


guides/doc-Configuring_Load_Balancer/topics/Promoting_SCAP_Content_to_Clients.adoc
```diff
@@ -42,7 +42,7 @@ Do not run the Puppet agent on every client manually because the Puppet agent ru
 .Additional Resources
 
 ifdef::satellite[]
```
ekohl, line 44: Is this `ifdef` still correct? Now that it's actually a guide within upstream (and also Orcharhino I think, but @maximiliankolb can answer that), this will be useful for everyone.

Lennonka: Yes, it is. We might need to fine tune these ifdefs for upstream, actually, because we've decided to use different titles for this guide.

ekohl (approved):
I'd like to see a similar section for upstream, but no objections to merging this now. @maximiliankolb?

guides/doc-Configuring_Load_Balancer/topics/Promoting_SCAP_Content_to_Clients.adoc
maximiliankolb: You can remove this ifdef now. @Lennonka

Lennonka: Oki, thanks. Done.

PR #1070, opened 2022-02-16
Fix usage of attributes (ProjectName)

Lennonka:
Cosmetic fine tuning. An attempt to remove version number from the project name, i.e. replacing {ProjectNameX} with {ProjectName}, where (hopefully) applicable.

Cherry-pick to:

* [x] Foreman 3.1

### DIFF
--- a/guides/common/modules/proc_adding-server-as-a-dynamic-inventory-item.adoc
+++ b/guides/common/modules/proc_adding-server-as-a-dynamic-inventory-item.adoc
@@ -15,7 +15,7 @@ For more information about managing users, roles, and permission filters, see {A
 .Procedure
 
 . In the {awx} web UI, create a credential for your {Project}.
-For more information about creating credentials, see http://docs.ansible.com/ansible-tower/latest/html/userguide/credentials.html#add-a-new-credential[Add a New Credential] and http://docs.ansible.com/ansible-tower/latest/html/userguide/credentials.html#red-hat-satellite-6[{ProjectNameX} Credentials] in the _{awx} User Guide_.
+For more information about creating credentials, see http://docs.ansible.com/ansible-tower/latest/html/userguide/credentials.html#add-a-new-credential[Add a New Credential] and http://docs.ansible.com/ansible-tower/latest/html/userguide/credentials.html#red-hat-satellite-6[{ProjectName} Credentials] in the _{awx} User Guide_.
 +
 [[tabl-Managing_Hosts-Integrating_Satellite_and_Ansible_Tower-Satellite_Credentials]]
 .{Project} Credentials
--- a/guides/common/modules/proc_registering-a-host-using-the-bootstrap-script.adoc
+++ b/guides/common/modules/proc_registering-a-host-using-the-bootstrap-script.adoc
@@ -4,7 +4,7 @@
 **Deprecated** Use registration via Global registration feature.
 
 Use the bootstrap script to automate content registration and Puppet configuration.
-You can use the bootstrap script to register new hosts, or to migrate existing hosts to {ProjectNameX} from {Project} 5, RHN, SAM, or RHSM.
+You can use the bootstrap script to register new hosts, or to migrate existing hosts from RHN, SAM, RHSM, or another {ProjectName} instance.
 
 The `katello-client-bootstrap` package is installed by default on {ProjectServer}'s base operating system.
 The `bootstrap.py` script is installed in the `/var/www/html/pub/` directory to make it available to hosts at `_{foreman-example-com}_/pub/bootstrap.py`.
--- a/guides/common/assembly_configuring-satellite-custom-server-certificate.adoc
+++ b/guides/common/assembly_configuring-satellite-custom-server-certificate.adoc
@@ -3,7 +3,7 @@ ifdef::context[:parent-context: {context}]
 [id="Configuring_Server_with_a_Custom_SSL_Certificate_{context}"]
 = Configuring {ProjectServer} with a Custom SSL Certificate
 
-By default, {ProjectNameX} uses a self-signed SSL certificate to enable encrypted communications between {ProjectServer}, external {SmartProxyServer}s, and all hosts.
+By default, {ProjectName} uses a self-signed SSL certificate to enable encrypted communications between {ProjectServer}, external {SmartProxyServer}s, and all hosts.
 If you cannot use a {Project} self-signed certificate, you can configure {ProjectServer} to use an SSL certificate signed by an external Certificate Authority.
 
 To configure your {ProjectServer} with a custom certificate, complete the following procedures:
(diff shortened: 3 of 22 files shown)


guides/common/modules/proc_adding-server-as-a-dynamic-inventory-item.adoc
```diff
@@ -15,13 +15,13 @@ For more information about managing users, roles, and permission filters, see {A
...
 . In the {awx} web UI, create a credential for your {Project}.
-For more information about creating credentials, see http://docs.ansible.com/ansible-tower/latest/html/userguide/credentials.html#add-a-new-credential[Add a New Credential] and http://docs.ansible.com/ansible-tower/latest/html/userguide/credentials.html#red-hat-satellite-6[{ProjectNameX} Credentials] in the _{awx} User Guide_.
+For more information about creating credentials, see http://docs.ansible.com/ansible-tower/latest/html/userguide/credentials.html#add-a-new-credential[Add a New Credential] and http://docs.ansible.com/ansible-tower/latest/html/userguide/credentials.html#red-hat-satellite-6[{ProjectName} Credentials] in the _{awx} User Guide_.
 +
 [[tabl-Managing_Hosts-Integrating_Satellite_and_Ansible_Tower-Satellite_Credentials]]
 .{Project} Credentials
 [cols="1a,2a"options="header"]
 |====
-|*Credential Type*: |*{ProjectNameX}*
+|*Credential Type*: |*{ProjectName}*
```
ekohl, line 24: This looks like the option within Ansible Tower - is that already modified to drop the version number?

guides/common/modules/proc_adding-server-as-a-dynamic-inventory-item.adoc
```diff
@@ -38,7 +38,7 @@ For more information, see https://docs.ansible.com/ansible-tower/latest/html/use
 .Inventory Source Options
 [cols="1a,2a"options="header"]
 |====
-|*Source* |*{ProjectNameX}*
+|*Source* |*{ProjectName}*
```
ekohl, line 41: Same comment about Tower here.

guides/common/modules/proc_registering-a-host-using-the-bootstrap-script.adoc
```diff
@@ -4,7 +4,7 @@
 **Deprecated** Use registration via Global registration feature.
 
 Use the bootstrap script to automate content registration and Puppet configuration.
-You can use the bootstrap script to register new hosts, or to migrate existing hosts to {ProjectNameX} from {Project} 5, RHN, SAM, or RHSM.
+You can use the bootstrap script to register new hosts, or to migrate existing hosts to {ProjectName} 6 from {ProjectName} 5, RHN, SAM, or RHSM.
```
ekohl, line 7: I wonder about `{ProjectName} 6` here. That will be odd once Satellite 7 is out. However, a bare `{ProjectName}` will also read weird. Perhaps just leave out the `to {ProjectName} 6` bit altogether?
```suggestion
You can use the bootstrap script to register new hosts, or to migrate existing hosts from {ProjectName} 5, RHN, SAM, or RHSM.
```

Lennonka: Will there be an option to migrate hosts from Sat 5 to Sat 7 directly?

guides/common/modules/proc_adding-server-as-a-dynamic-inventory-item.adoc
Lennonka: This may be replaced by mistake.

guides/common/modules/proc_registering-a-host-using-the-bootstrap-script.adoc
ekohl: I don't see why not. Satellite 6 and 7 are the same technology stack. 6.9 and 6.10 differ more than 6.10 and 7.0.

Lennonka: In that case, we will need to look deeper into the phrasing once we get to updates for 7.0. For now, I would keep only this change, if you agree.

guides/common/modules/proc_adding-server-as-a-dynamic-inventory-item.adoc
Lennonka: I see, it's not in our interface. Reverting.

guides/common/modules/proc_adding-server-as-a-dynamic-inventory-item.adoc
Lennonka: Reverting.

guides/common/modules/proc_registering-a-host-using-the-bootstrap-script.adoc
Lennonka: Hang on. I was asking about version five, not six.

ekohl: I get that, but I was saying that 6 and 7 are so much the same that the bootstrap script will work with both.

Lennonka: Oki. I'm removing the "to". And I'm thinking we should probably keep the version number in this case. Agreed?

guides/common/modules/proc_registering-a-host-using-the-bootstrap-script.adoc
```diff
@@ -4,7 +4,7 @@
 **Deprecated** Use registration via Global registration feature.
 
 Use the bootstrap script to automate content registration and Puppet configuration.
-You can use the bootstrap script to register new hosts, or to migrate existing hosts to {ProjectNameX} from {Project} 5, RHN, SAM, or RHSM.
+You can use the bootstrap script to register new hosts, or to migrate existing hosts from RHN, SAM, RHSM, or an earlier version of {ProjectName}.
```
Lennonka, line 7: @ekohl How about this?

ekohl: Technically I think it can be used on hosts registered to any other Satellite/Foreman/Katello instance. So it can also be used to migrate hosts from a Satellite 7 to another Satellite 7, for example if you had built a lab environment but then another lab/production. Perhaps `or a different {ProjectName} (instance)` would be a more generic description? Not sure about the exact phrasing, but I hope you get the idea.

Lennonka: Cool. I hope this will suffice.

lzap (approved):
LGTM

maximiliankolb (commented):
One question: where is the graphic used? Why do you not delete the "old" graphic?

All other changes look good!

Lennonka:
Oh, sorry, I committed the graphic by mistake! That's for another PR :)) Hang on

maximiliankolb (approved):
LGTM. Thanks for the PR!

maximiliankolb:
@Lennonka Please create a separate PR against 3.1 because there are merge conflicts.

PR #1072, opened 2022-02-16
Whitespace

maximiliankolb:
Cherry-pick into:

* [x] Foreman 3.1


melcorr (approved):
lol why does this matter?

melcorr:
Double space is a bad thing obviously. But the others? What is the harm ? :)

maximiliankolb:
Double whitespace is just wrong; and the rest, I'd argue consistency.

Sure, an extra empty line here and there doesn't hurt; but where do you draw the line? I'd argue that we can drop it if it's not required by asciidoc or, more importantly, for easier maintenance.

PR #1073, opened 2022-02-16
Add underscore before context in anchor

maximiliankolb:
FYI: I've checked; the anchor is not referenced anywhere.

Cherry-pick into:

* [x] Foreman 3.1


melcorr (approved):
Thanks!

PR #1081, opened 2022-02-16
Disable link checker for downstream docs

ekohl:
There is no corresponding Satellite release to Foreman 3.2 so there is no point in checking.

### DIFF
--- a/guides/common/linkchecker.ini
+++ b/guides/common/linkchecker.ini
@@ -17,3 +17,5 @@ ignore=example.com
   rhsso.com
   sources/
   atixservice.zendesk.com
+  # There are no downstream 7.1 docs
+  access.redhat.com/documentation/en-us/red_hat_satellite/7.1/


guides/common/linkchecker.ini
```diff
@@ -17,3 +17,5 @@ ignore=example.com
   rhsso.com
   sources/
   atixservice.zendesk.com
+  # There are no downstream 7.1 docs
+  access.redhat.com/documentation/en-us/red_hat_satellite/7.1/
```
maximiliankolb, line 21: Does it make sense to mention the version `7.1`? Would it be "safe" to omit it?

ekohl: My reasoning is that even for a Foreman 3.1 version we should have a period where the docs are not published. So as I said in https://github.com/theforeman/foreman-documentation/pull/1063#issuecomment-1040068743, I do think we should keep the version in explicitly, also in master. Though technically you're probably right that it doesn't matter that much.

maximiliankolb (approved):

PR #1083, opened 2022-02-16
Satdoc 606 upgrading to 6.8 causes old /var/lib/pgsql data to be left behind

bangelic:
I added two steps regarding /var/lib/pgsql/ from SATDOC-606. These are steps 4 and 5 in Reclaiming PostgreSQL Space. These steps will confirm all data under /var/lib/pgsql/ is cleaned.

@melcorr (I realized when I did the commit, I needed to actually do this for 2.5 and not the master. Can this be merged to 2.5 instead or do I need to do another commit specifically to 2.5?)
Cherry-pick into:

* [ ] Foreman 3.1


melcorr:
If it is only for 2.5 you'll need to checkout the 2.5 branch, create a branch from there, and open a pull request against 2.5.

This cannot be merged unfortunately :-(

PR #1087, opened 2022-02-18
Links with context

maximiliankolb:
This PR fixes various broken xrefs and simplifies (use xref directly without attribute containing the path to the guide) where possible.

Cherry-pick into:

* [x] Foreman 3.1


Lennonka (approved):
LGTM

Just out of curiosity -- can you please clarify what the use case is for the `project-context` attribute?

maximiliankolb:
So my understanding is that `{context}` depends on the name of the guide, so it's generally set in the `master.adoc` file (and sometimes for different assemblies, like kvm and ovirt), and `{project-context}` depends on the build target, i.e. foreman-el, foreman-deb, katello, satellite, orcharhino.

use case: I don't know.

maximiliankolb:
Merged to master & cherry-picked to 3.1: https://github.com/theforeman/foreman-documentation/commit/acea7a2dbc88423dfe1dba6355585947a103d3c3 ✅

PR #1089, opened 2022-02-21
Add OSTree content type

Lennonka:
Cherry-pick into:

* [x] Foreman 3.2
* I'll file another PR for 3.1

PR #1093, opened 2022-02-21
Add OSTree content type 3.1

Lennonka:
Follow-up to #1089


maximiliankolb:
Small tip: If you cherry-pick commits from a "stable" branch such as master, use `cherry-pick -x _SHA1_` to include the commit id of the original commit. Stable means that we don't rewrite history of this branch, i.e. the commit id still stay like this forver and never change.

maximiliankolb (approved):

PR #1098, opened 2022-02-23
don't use _something_ in escaped URLs 

evgeni:
Cherry-pick into:

* [x] Foreman 3.1

### DIFF
--- a/guides/common/modules/proc_adding-a-default-http-proxy.adoc
+++ b/guides/common/modules/proc_adding-a-default-http-proxy.adoc
@@ -12,7 +12,7 @@ To use the CLI instead of the {ProjectWebUI}, see the xref:cli-adding-a-default-
 . In the {ProjectWebUI}, navigate to *Infrastructure* > *HTTP Proxies*.
 . Click *New HTTP Proxy*.
 . In the *Name* field, enter the name for the HTTP proxy.
-. In the *Url* field, enter the URL of the HTTP proxy in the following format: `\https://_proxy.example.com_:8080`.
+. In the *Url* field, enter the URL of the HTTP proxy in the following format: `\https://proxy.example.com:8080`.
 . Optional: If authentication is required, in the *Username* field, enter the username to authenticate with.
 . Optional: If authentication is required, in the *Password* field, enter the password to authenticate with.
 . To test connection to the proxy, click the *Test Connection* button.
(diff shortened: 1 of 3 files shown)


evgeni:
interesting, this comes from https://github.com/theforeman/foreman-documentation/pull/487 by @maximiliankolb but right now this renders the `\` verbatim in the docs?

example: https://docs.theforeman.org/nightly/Installing_Server/index-katello.html#Configuring_Server_with_an_HTTP_Proxy_foreman has
Screenshot from 2022-02-23 11-36-45

maximiliankolb:
Yes, I was surprised to see the literal backslash.

This seems to work fine in `guides/common/modules/proc_preparing-suse-installation-media.adoc`:

````
 50 Use the path to the local installation media files when {ProvisioningDocURL}adding-installation-media_provisioning[creating an installation medium], for example `\http://{foreman-example-com}/pub/installation_media/sles/15sp3/`.
 ````
Screenshot 2022-02-23 at 11 40 28

[source](https://docs.theforeman.org/nightly/Content_Management_Guide/index-katello.html#Preparing_SUSE_Installation_Media_content-management)

evgeni:
it does indeed, very weird.

evgeni:
ooh, is it because here we use `_proxy.example.com_`, which is not a valid URL, so the URL detection thingy rejects it, and then there is nothing to escape?

maximiliankolb:
So the solution is to not display it in italic?

evgeni:
I think we've been using `_hostname_` to indicate something that the reader should change (so from `foreman.example.com` to `foreman.theircorp.invalid` or similar), which I think is a good thing to keep.

So it should be more correct so check that all `*.example.com` URLs use that, and all *other* use `\` to escape. Bah, more work.

maximiliankolb:
Maybe use an attribute for this? 

For example `httpproxy-example-com` like https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-base.adoc#L100

evgeni:
https://github.com/asciidoctor/asciidoctor/blob/de54df6307c2b1ad6db5a7f0b305b730b7afb096/lib/asciidoctor/rx.rb#L511-L522 confirms my trail of thought.

I don't mind which way we fix this, tbh :)

maximiliankolb:
I think it makes sense to just remove the `_` for now, and then in a second step use an attribute.

All other URLs also just use SOMETHING.example.com.

evgeni:
so, like that?

maximiliankolb (approved):
LGTM

maximiliankolb:
Cherry-picked to 3.1: https://github.com/theforeman/foreman-documentation/commit/a03f1e6ed7f33b76d47d6bd6e6367c8d939a7566 ✅

ekohl:
We should really update the template because 3.2 is also branched. We shouldn't forget that.

PR #1100, opened 2022-02-23
BZ#2056737: Configure Ansible repository (Sat only)

Lennonka:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2056737

This concerns a topic that is published for disconnected Satellite only.

Cherry-pick into:

* [x] Foreman 3.1

### DIFF
--- a/guides/doc-Upgrading_and_Updating/topics/upgrading_a_disconnected_satellite.adoc
+++ b/guides/doc-Upgrading_and_Updating/topics/upgrading_a_disconnected_satellite.adoc
@@ -120,6 +120,36 @@ Do not install or update any packages at this stage.
 baseurl=file:///media/sat6/
 ----
 
+. Configure the Ansible repository from the ISO file.
+
+.. Copy the ISO file's repository data file for Ansible packages:
++
+[options="nowrap"]
+----
+# cp /media/sat6/ansible/media.repo /etc/yum.repos.d/ansible.repo
+----
+
+.. Edit the `/etc/yum.repos.d/ansible.repo` file:
++
+[options="nowrap"]
+----
+# vi /etc/yum.repos.d/ansible.repo
+----
+
+... Change the default `InstallMedia` repository name to `Ansible`:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+[Ansible]
+----
+
+... Add the `baseurl` directive:
++
+[options="nowrap"]
+----
+baseurl=file:///media/sat6/ansible/
+----
+
 . Configure the Red Hat Software Collections repository from the ISO file.
 
 .. Copy the ISO file's repository data file for Red Hat Software Collections packages:


ezr-ondrej:
I do not know enough about the process, but FWIW LGTM

guides/doc-Upgrading_and_Updating/topics/upgrading_a_disconnected_satellite.adoc
```diff
@@ -119,6 +119,14 @@ Do not install or update any packages at this stage.
 ----
 baseurl=file:///media/sat6/
 ----
+... Optional: If you want to configure a custom path for the Ansible Runner packages, navigate to the `Ansible` repository and modify the `baseurl` directive with your custom path.
```
evgeni, line 122: the BZ is about ansible, which needs to be set up identically to the SCL below (with the caveat, that there seem not to be an ansible.repo file on the ISO). and it's not optional.

Lennonka: Thank you!

Lennonka:
@evgeni Like this? :)

evgeni:
almost, the repo file will now be missing some crucial entries (as there was no template to use from)

let me check if the missing file is intentional (then we need to properly document how to create one) or not (then we don't)

Lennonka:
@evgeni Hi! Any updates? :)

evgeni:
So far the update is "it shouldn't be like this", but also no clear path forward if that can be fixed trivially. :(

Lennonka:
@evgeni We're going to provide a KBase article to give our users the workaround for the missing file and include a Known Issue in the RN for 6.10 -- this will be a separate PR. As for 7.0, this is going to be fixed and I believe we can go ahead with this and merge it.

Lennonka:
@evgeni Oopsiedaisy, I missed that. Thanks, will do!

Lennonka:
Naming confirmed with the ISO composition team.
@evgeni Will you do the honors? :)

guides/doc-Upgrading_and_Updating/topics/upgrading_a_disconnected_satellite.adoc
```diff
@@ -106,20 +106,50 @@ Do not install or update any packages at this stage.
 # vi /etc/yum.repos.d/sat6.repo
 ----
 
-... Change the default `InstallMedia` repository name to `{Project}-{ProjectVersion}`:
+.. Change the default `InstallMedia` repository name to `{Project}-{ProjectVersion}`:
```
evgeni, line 109: why this level change? these are (technically) sub-items of the "edit the repo file" section.

evgeni (commented):
I am cool with the Ansible part, but unsure about the level change overall.

evgeni (approved):
I shall first read the comments… ACK

Lennonka:
I've kept the list structure in the end.

evgeni:
[3.2 496a28aa] BZ#2056737: Configure Ansible repository
[3.1 38b5d654] BZ#2056737: Configure Ansible repository

PR #1101, opened 2022-02-23
Add saltuser to foreman-proxy group

maximiliankolb:
See also https://github.com/theforeman/foreman_salt/pull/136#discussion_r627233869

Cherry-pick into:

* [x] Foreman 3.3
* [x] Foreman 3.2
* [x] Foreman 3.1

### DIFF
--- a/guides/common/modules/proc_salt-authenticating-salt-minions-using-host-names.adoc
+++ b/guides/common/modules/proc_salt-authenticating-salt-minions-using-host-names.adoc
@@ -11,6 +11,12 @@ This relies on the `autosign.conf` file that stores the host names of Salt Minio
 ----
 # ssh root@salt-master.example.com
 ----
+. Add the `foreman-proxy` user that is running Salt to the `root` user group:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# usermod -a -G foreman-proxy root
+----
 . Enable the `autosign.conf` file in `/etc/salt/master`:
 +
 [options="nowrap" subs="attributes"]
@@ -23,6 +29,6 @@ permissive_pki_access: True
 [options="nowrap" subs="attributes"]
 ----
 # touch /etc/salt/autosign.conf
-# chgrp foreman-proxy /etc/salt/autosign.conf
+# chown root:foreman-proxy /etc/salt/autosign.conf
 # chmod 660 /etc/salt/autosign.conf
 ----


guides/common/modules/proc_salt-authenticating-salt-minions-using-host-names.adoc
```diff
@@ -24,5 +30,5 @@ permissive_pki_access: True
 ----
 # touch /etc/salt/autosign.conf
 # chgrp foreman-proxy /etc/salt/autosign.conf
-# chmod 660 /etc/salt/autosign.conf
+# chmod 640 /etc/salt/autosign.conf
```
Lennonka, line 33: @bochi Can you please doublecheck that these permissions are fine? I'm not an expert to review this factually :) Thank you!

ekohl: I was looking at the installer, but it surprises me we don't ensure the file exists with the correct permissions.

These permissions also look a bit odd. The result is that it will be owned by `root` and the group is `foreman-proxy`. It is then only writable for `root` and `foreman-proxy` can read it. Is that correct? It implies there is either no automation to modify `autosign.conf` or there's a process running as `root` that does.

maximiliankolb:
Rebased to master & waiting for either @bochi or @sbernhard

maximiliankolb:
Rebased to master.

guides/common/modules/proc_salt-authenticating-salt-minions-using-host-names.adoc
Lennonka: @sbernhard @bochi Please, can you help us to move this forward?

sbernhard: @bochi wrote in https://github.com/theforeman/foreman_salt/pull/136#discussion_r627233869
> I actually have exactly that setup working on salt 3003 :-)
The only thing that needs to be confiured is that the user running salt needs to be in the foreman-proxy group and the configuration option permissive_pki_access: True in the salt master config needs to be set.

Means, the following works:

In /etc/salt/master the configuration need to look like this:
`autosign_file: /etc/salt/autosign.conf
permissive_pki_access: True`

The file permission (user, group) of /etc/salt/autosign.conf  need to look like this:
`-rw-rw----. 1 root foreman-proxy 37 10. Mai 22:24 /etc/salt/autosign.conf`

In /etc/group the user who runs salt (in this case 'root') need to be in the group:
`foreman-proxy:x:993:root`

maximiliankolb:
Added Bernhards suggestion and rebased to master. Ready to merge.

sbernhard:
One part is missing. In salt/master config file the following line need to exist:
`permissive_pki_access: True`

sbernhard:
> > One part is missing. In salt/master config file the following line need to exist: `permissive_pki_access: True`
> 
> This is already part of the docs, see line 25:
> 
> ```
> . Enable the `autosign.conf` file in `/etc/salt/master`:
> +
> [options="nowrap" subs="attributes"]
> ----
> autosign_file: /etc/salt/autosign.conf
> permissive_pki_access: True
> ----
> ```
> 
> My PR does _not_ need to change this, does it?

You are right. Then I would sugget to use
chown root.foreman-proxy /etc/salt/autosign.conf

maximiliankolb:
> > > One part is missing. In salt/master config file the following line need to exist: `permissive_pki_access: True`
> > 
> > 
> > This is already part of the docs, see line 25:
> > ```
> > . Enable the `autosign.conf` file in `/etc/salt/master`:
> > +
> > [options="nowrap" subs="attributes"]
> > ----
> > autosign_file: /etc/salt/autosign.conf
> > permissive_pki_access: True
> > ----
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > My PR does _not_ need to change this, does it?
> 
> You are right. Then I would sugget to use chown root.foreman-proxy /etc/salt/autosign.conf

Does that replace the two other commands?

```diff
-# chgrp foreman-proxy /etc/salt/autosign.conf
-# chmod 660 /etc/salt/autosign.conf
+# chown root:foreman-proxy /etc/salt/autosign.conf
```

sbernhard:
No, it replaces only the chgrp command

maximiliankolb:
Rebased to master and applied all feedback.

Lennonka:
Cherry-picked:

- 114f6627..8494e6b6  3.3 -> 3.3
- 571428c0..18d194ba  3.2 -> 3.2
- c62c0aa1..1955a7d8  3.1 -> 3.1

PR #1103, opened 2022-02-23
Add Foreman 3.2 links

ianballou:
I've added some changes to get the process of publishing Foreman 3.2 (and hence Katello 4.4) docs.  This process needs to happen before I can make the release announcement since Katello relies on this new documentation website.

I'm not certain I edited everything correctly, so I need some help there. Also, is it right to say 3.2 is in the "RC-phase" as I did? It didn't seem right to say it's stable yet.

### DIFF
--- a/web/content/index.adoc
+++ b/web/content/index.adoc
@@ -5,6 +5,7 @@ title: Home
 == Foreman and Katello Documentation
 
 * link:/release/nightly/[Nightly build]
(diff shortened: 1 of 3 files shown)


ekohl (approved):

ekohl (approved):
I think `Release Candidate` rather than `RC-phase` might be better but for now I'm going to merge this.

PR #1104, opened 2022-02-24
Fixed the command to install satellite-clone

mdolezelova:
Cherry-pick into:

* [ ] Foreman 3.2
* [ ] Foreman 3.1

### DIFF
--- a/guides/doc-Upgrading_and_Updating/topics/cloning_satellite_server.adoc
+++ b/guides/doc-Upgrading_and_Updating/topics/cloning_satellite_server.adoc
@@ -165,12 +165,22 @@ To clone your server, complete the following steps on your target server:
 --enable=rhel-7-server-satellite-{ProjectVersionPrevious}-rpms
 ----
 +
+ifdef::satellite[]
+. Install the `satellite-clone` package:
++
+[options="nowrap" subs="attributes"]
+----
+# yum install satellite-clone
+----
+endif::[]
+ifndef::satellite[]
 . Install the `satellite-clone` package:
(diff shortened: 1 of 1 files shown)


maximiliankolb:
Hi @mdolezelova This PR is against 2.5. If your change is also relevant for Foreman nightly, please open a PR against the master branch, and then we can cherry-pick to 3.2 and 3.1. The general workflow always (unless it's specifically _only_ for older versions) starts with a PR against master, a commit to master, and then cherry-picks to other branches.

For 2.5, you would need a standalone PR, so this is fine (even though we're missing a reference to the commit in master).

mdolezelova:
> Hi @mdolezelova This PR is against 2.5. If your change is also relevant for Foreman nightly, please open a PR against the master branch, and then we can cherry-pick to 3.2 and 3.1. The general workflow always (unless it's specifically _only_ for older versions) starts with a PR against master, a commit to master, and then cherry-picks to other branches.
> 
> For 2.5, you would need a standalone PR, so this is fine (even though we're missing a reference to the commit in master).

Hi @maximiliankolb , sure, this one is to fix a bug in Satellite 6.10 docs, so it's against the 2.5 branch. I'm not sure yet whether it's relevant also for future versions of Satellite documentation, but I'd expect that yes. So I'll create a separate PR against master requesting cherry-picking to 3.1 and 3.2. Many thanks for your guidance!
Please could you go ahead and merge this one? Thank you!

mdolezelova:
Thank you, @maximiliankolb !

PR #1114, opened 2022-03-01
Remove discovery upgrade instructions on capsules

ekohl:
This should be covered by the other instructions. If it was already installed, the yum/dnf upgrade has already happened. Then the installer should take care of the rest. When it wasn't installed, users should really follow the guide to install discovery. That shouldn't be part of the upgrade guide.

@lzap I recall you said this would be OK. Mind taking a look?

Cherry-pick into:

* [x] Foreman 3.2
* [ ] Foreman 3.1


lzap (approved):
Yeah this is all useless, installer/dnf should take care of this.

lzap:
Commit(s) 7acf3df5 were cherry-picked into: 3.2
* ea67e701 Remove discovery upgrade instructions on capsules

PR #1117, opened 2022-03-02
Fix path to postgresql in storage requirement section

mdolezelova:
Cherry-pick into:

* [ ] Foreman 3.2
* [ ] Foreman 3.1

### DIFF
--- a/guides/common/modules/ref_storage-requirements.adoc
+++ b/guides/common/modules/ref_storage-requirements.adoc
@@ -26,7 +26,7 @@ ifdef::foreman-el,katello[]
 
 |/var/log/ |10 MB |10 GB
 
-|/var/opt/rh/rh-postgresql12 |100 MB |20 GB
+|/var/opt/rh/rh-postgresql12/lib/pgsql |100 MB |20 GB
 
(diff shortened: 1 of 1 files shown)


ekohl (approved):
This looks OK to me, but it should also be submitted to master and cherry picked to keep it consistent.

mdolezelova:
@ekohl , thank you! Yes, I'm thinking the same. I'm just for the confirmation about the relevance for future version in the BZ. So once it's confirmed, I'll submit a PR against master and request cherry-picking to the later versions. Many thanks!

ekohl:
I can confirm for you that this will be relevant on all versions where EL7 is supported. It will be different on EL8, but EL8 is a massive change that needs to happen throughout the documentation.

mdolezelova:
> I can confirm for you that this will be relevant on all versions where EL7 is supported. It will be different on EL8, but EL8 is a massive change that needs to happen throughout the documentation.

Thank you, @ekohl ! So the corresponding PR for master and future version is https://github.com/theforeman/foreman-documentation/pull/1119.

PR #1120, opened 2022-03-03
Allow building on Darwin

lzap:
It was reported to me that guides do not build on MacOS because there are no hardlinks. Poor MacOS users, let's fix that :)

### DIFF
--- a/guides/common/Makefile
+++ b/guides/common/Makefile
@@ -22,9 +22,11 @@ CSS_STYLE = compressed
 
 ifeq ($(UNAME), Linux)
 BROWSER_OPEN = xdg-open
+CP_ARGS = -l
 endif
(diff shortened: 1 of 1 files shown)


ekohl (approved):

melcorr:
I'm pretty sure @maximiliankolb is using Mac.

maximiliankolb:
Yes, technically correct, but I think this was done because of Sarah.

For local builds, I still rely on docker:

    rm -rf guides/build && docker run --rm -v $(pwd):/foreman-documentation foreman_documentation make html && open -a Chromium guides/build

lzap:
I can confirm, yeah! I sometimes build on Mac too but this bug slipped in.

PR #1121, opened 2022-03-03
Set pagination value for Hammer CLI

maximiliankolb:
Maybe there is an even better place to put this? I know that you need to set the value at least there if you have more than 20 CVs.

Cherry-pick into:

* [x] Foreman 3.2
* [x] Foreman 3.1

### DIFF
--- a/guides/common/modules/proc_exporting-a-content-view-version.adoc
+++ b/guides/common/modules/proc_exporting-a-content-view-version.adoc
@@ -26,6 +26,12 @@ To export a Content View, ensure that {ProjectServer} where you want to export m
 For more information, see xref:Download_Policies_Overview[].
 * Ensure that you synchronize Products that you export to the required date.
 * Ensure that the user exporting the content has the `Content Exporter` role.
+* If you have more than 20 Content Views, set a higher value for `entries_per_page` to avoid pagination of Hammer output:
++
+[subs="+quotes"]
+----
+# hammer settings set --name entries_per_page --value 1000
+----
 
 .To Export a Content View Version
 . List versions of the Content View that are available for export:


melcorr (approved):
I am happy with the location of this. 
Will wait to see if anyone else has an opinion.

guides/common/modules/proc_exporting-a-content-view-version.adoc
```diff
@@ -26,6 +26,12 @@ To export a Content View, ensure that {ProjectServer} where you want to export m
 For more information, see xref:Download_Policies_Overview[].
 * Ensure that you synchronize Products that you export to the required date.
 * Ensure that the user exporting the content has the `Content Exporter` role.
+* If you have more than 20 Content Views, set a higher value for `entries_per_page` to avoid pagination of Hammer output:
++
+[subs="+quotes"]
+----
+# hammer settings set --name entries_per_page --value 1000
```
ekohl, line 33: This will affect the UI too. Is this really wise? Isn't there an option for every hammer command to set a page size? We should also have the special value `all` for a page size if you want to disable pagination, at least in the API. I don't know if Hammer exposes this.

maximiliankolb: I didn't know that, thanks for pointing it out! I doubt that `1000` is a good value for the Web UI.

What about a solution where you temporarily store the "original" value?

    CURRENT_ENTRIES_PER_PAGE=$(hammer settings info --name entries_per_page --fields Value | grep "Value:" | awk '{print $2}')
    hammer settings set --name entries_per_page --value 1000
    # do something
    hammer settings set --name entries_per_page --value $CURRENT_ENTRIES_PER_PAGE

ekohl: I didn't check with `hammer content-view version list`, but with all other list views in hammer you can pass `--per-page`:
```console
# hammer location list --help
Usage:
    hammer location <list|index> [OPTIONS]

Options:
 ...
 --page NUMBER                           Page number, starting at 1
 --per-page VALUE                        Number of results per page to return, 'all' to return all results
```

So I think it'd be better to change the list command to add `--per-page all`.

maximiliankolb: Thanks for your input @ekohl 

I've tested this again and I did not run into the issue I was trying to solve. I've set the `Entries per page` to 15, and even though I had 26 CVs, I saw all of them using `hammer content-view version list`. I am closing this PR.

PR #1124, opened 2022-03-03
Reword path of the installation medium

maximiliankolb:
IMHO using "medium's path" is not possible because medium is not a person.

Cherry-pick into:

* [x] Foreman 3.2
* [x] Foreman 3.1


lzap (approved):
OK TY

PR #1126, opened 2022-03-03
ISS CDN configuration

Lennonka:
Cherry-pick into:

* [x] Foreman 3.2
* [x] Foreman 3.1

### DIFF
--- a/guides/common/modules/proc_configuring-satellite-to-synchronize-content-with-a-local-cdn-server.adoc
+++ b/guides/common/modules/proc_configuring-satellite-to-synchronize-content-with-a-local-cdn-server.adoc
@@ -47,8 +47,10 @@ For example, `/var/www/html/pub/sat-import/`:
 # restorecon -rv /var/www/html/pub/sat-import/
 ----
 . In the {ProjectWebUI}, navigate to *Content* > *Subscriptions*.
-. Click *Manage Manifest*.
-. Edit the *Red Hat CDN URL* field to point to the host name of the system that you use as a local CDN server with the newly created directory, for example:
+. Click the *Manage Manifest* button.
+. Switch to the *CDN Configuration* tab.
+. Select *Red Hat CDN*.
+. Edit the *URL* field to point to the host name of the system that you use as a local CDN server with the newly created directory, for example:
 +
 `\http://server.example.com/pub/sat-import/`
 . Click *Update* and then upload your manifest into {Project}.
--- a/guides/common/modules/proc_configuring-server-for-air-gapped-operation.adoc
+++ b/guides/common/modules/proc_configuring-server-for-air-gapped-operation.adoc
@@ -1,0 +1,11 @@
+[id="configuring-server-for-air-gapped-operation_{context}"]
+= Configuring {Project} Server for Air-Gapped Operation
+
+If you deployed your {ProjectServer} as air-gapped, disable automatic communication with the Red Hat CDN to avoid checking what content is available to save system resources.
+
+.Procedure
+. In the {ProjectWebUI}, navigate to *Content* > *Subscriptions*.
+. Click the *Manage Manifest* button.
+. Switch to the *CDN Configuration* tab.
+. Select *Air-gapped*.
+. Click *Update*.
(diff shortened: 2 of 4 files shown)


guides/common/modules/proc_configuring-satellite-for-air-gapped-operation.adoc
```diff
@@ -0,0 +1,11 @@
+[id="configuring-satellite-for-air-gapped-operation_{context}"]
+= Configuring Satellite for Air-Gapped Operation
```
Lennonka, line 2: I'm pretty sure that this title doesn't say what we need to say here. Any ideas?

guides/common/modules/proc_configuring-satellite-for-air-gapped-operation.adoc
```diff
@@ -0,0 +1,11 @@
...
+
+If you deployed your {ProjectServer} as  air-gapped, disable automatic communication with the Red Hat CDN to avoid checking what content is available to save system resources.
```
melcorr, line 4: Double space here.

jturel:
I need to chat w/ @parthaa but I'm seeing https://github.com/theforeman/foreman-documentation/blob/master/guides/common/modules/proc_configuring-satellite-to-synchronize-content-with-a-local-cdn-server.adoc for the first time. This workflow ("local cdn server") is no longer applicable in combination with the two new ISS workflows. It should (imo) use ISS Export Sync which implies that they'd need to use a second system to export the content. Which means that doc ultimately goes away. Considering content ISOs are going away at some point in the future this is especially true. Partha can correct me if I'm forgetting something here

jturel:
Update on my previous comment: The Red Hat CDN url can still be changed via the UI but those docs should likely be re-worked soon to reflect the deprecation of content ISOs. A docs bug seems appropriate as I'm not sure of the timeline of that deprecation

guides/common/modules/proc_configuring-satellite-for-air-gapped-operation.adoc
melcorr: We will not really call it Satellite but {Project} right?

Lennonka:
@melcorr I appreciate your early feedback. This PR is in the DRAFT mode. I was hoping that it was enough to say that as such, it is not yet ready for community review. Should I communicate this state of art differently to make it clearer next time?

melcorr:
> @melcorr I appreciate your early feedback. This PR is in the DRAFT mode. I was hoping that it was enough to say that as such, it is not yet ready for community review. Should I communicate this state of art differently to make it clearer next time?

I would prefer if we didn't use vendor terms in the upstream at any point, if they can be avoided @Lennonka

Lennonka:
@melcorr Oh, sure. Noted.

Lennonka:
> I need to chat w/ @parthaa but I'm seeing https://github.com/theforeman/foreman-documentation/blob/master/guides/common/modules/proc_configuring-satellite-to-synchronize-content-with-a-local-cdn-server.adoc for the first time. This workflow ("local cdn server") is no longer applicable in combination with the two new ISS workflows. It should (imo) use ISS Export Sync which implies that they'd need to use a second system to export the content. Which means that doc ultimately goes away. Considering content ISOs are going away at some point in the future this is especially true. Partha can correct me if I'm forgetting something here

@jturel Let's consider the bigger picture:

1. I'm wondering if the sections about disconnected/air-gapped scenarios and about configuring SatServer for ISS Network Sync should actually be in this guide instead of this section and not in the Content Management Guide (CMG), because it looks more like a one-time configuration task than a repeated content-management task.
2. If so, the ISS chapter in the CMG could stil be only about ISS Export Sync with some tweaks (also of the heading) and definitely hyperlinked from the air-gapped config section in this "Installing Disconnected" Guide.

This is, of course, a part of the bigger problem that we've discussed, regarding guidance on multi-server setups in general.
I'll discuss it with our Content Strategist, but I'd like to know what you think as well.

jturel:
> 1. I'm wondering if the sections about disconnected/air-gapped scenarios and about configuring SatServer for ISS Network Sync should actually be in this guide instead of this section and not in the Content Management Guide (CMG), because it looks more like a one-time configuration task than a repeated content-management task.
> 2. If so, the ISS chapter in the CMG could stil be only about ISS Export Sync with some tweaks (also of the heading) and definitely hyperlinked from the air-gapped config section in this "Installing Disconnected" Guide.
> 
> This is, of course, a part of the bigger problem that we've discussed, regarding guidance on multi-server setups in general. I'll discuss it with our Content Strategist, but I'd like to know what you think as well.

I think this was the missing piece as to what was bothering me about the existing docs - it mixes both satellite configuration/setups and also the content management aspects of those setups. Splitting the two seems absolutely correct.

PR #1131, opened 2022-03-07
Use inline images in tables

maximiliankolb:
In tables, you cannot use `image::`, but instead, must use inline images like `image:`.

Cherry-pick into:

* [x] Foreman 3.2
* [x] Foreman 3.1


lzap (approved):
Thank you.

lzap:
Commit(s) e0dd279b were cherry-picked into: 3.2 3.1
* dc62e744 Use inline images in tables
* 970f3d77 Use inline images in tables

PR #1142, opened 2022-03-08
Fixed postgresql entry in the storage requirements section

mdolezelova:
Cherry-pick into:

* [ ] Foreman 3.2
* [ ] Foreman 3.1


mdolezelova:
Hi, this is the same fix as was merged via https://github.com/theforeman/foreman-documentation/pull/1117. In 1117 I've fixed only a piece that is conditioned for Foreman as I haven't noticed the condition and thought this is the only place where it needs to be changed. Apologies for this .

PR #1149, opened 2022-03-10
Document new APT download policy "on-demand"

maximiliankolb:
With Katello 4.1, it is also possible to download APT content "on-demand".

I am unsure about the second commit. The whole wording sounds off to me but I didn't come up with something better, so I just added `deb`.

Cherry-pick into:

* [x] Foreman 3.2
* [x] Foreman 3.1

### DIFF
--- a/guides/common/modules/con_download-policies-overview.adoc
+++ b/guides/common/modules/con_download-policies-overview.adoc
@@ -20,7 +20,12 @@ On Demand::
 This setting has no effect if you set a corresponding repository on a {SmartProxy} to *Immediate* because {ProjectServer} is forced to download all the packages.
 
 The *On Demand* policy acts as a _Lazy Synchronization_ feature because they save time synchronizing content.
+ifdef::satellite[]
 The lazy synchronization feature must be used only for `yum` repositories.
+endif::[]
+ifndef::satellite[]
+The lazy synchronization feature must be used only for `deb` and `yum` repositories.
+endif::[]
 You can add the packages to Content Views and promote to life cycle environments as normal.
 
 {SmartProxyServer} has the following policies:
--- a/guides/common/modules/con_download-policies-overview.adoc
+++ b/guides/common/modules/con_download-policies-overview.adoc
@@ -28,6 +28,15 @@ The lazy synchronization feature must be used only for `deb` and `yum` repositor
 endif::[]
 You can add the packages to Content Views and promote to life cycle environments as normal.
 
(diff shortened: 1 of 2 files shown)


guides/common/modules/con_download-policies-overview.adoc
```diff
@@ -20,7 +20,12 @@ On Demand::
 This setting has no effect if you set a corresponding repository on a {SmartProxy} to *Immediate* because {ProjectServer} is forced to download all the packages.
 
 The *On Demand* policy acts as a _Lazy Synchronization_ feature because they save time synchronizing content.
+ifdef::satellite[]
 The lazy synchronization feature must be used only for `yum` repositories.
+endif::[]
+ifndef::satellite[]
+The lazy synchronization feature must be used only for `deb` and `yum` repositories.
```
quba42, line 27: It may be worth adding that the feature must only be used for additive upstream repos. If the upstream repos is in the habbit of chucking out packages it previously offered, you are going to have a bad time trying to sync it "lazy"...

maximiliankolb: I have added an explanation and used "EPEL" as an example.

maximiliankolb:
Do you mind having a look @lzap ?

lzap (approved):
Apologies for the delay, perfect!

lzap:
Commit(s) b345231a 0ada0a41 ebfebd99 were cherry-picked into: 3.2 3.1
* 4fadcc10 Do not use lazy sync for fast moving repositories
* d5ed0fa6 Do not use lazy sync for fast moving repositories

guides/common/modules/con_download-policies-overview.adoc
```diff
@@ -20,9 +20,23 @@ On Demand::
...
+endif::[]
 You can add the packages to Content Views and promote to life cycle environments as normal.
 
+Note that some repositories such as EPEL not only add packages to their upstream repository, but also remove older versions of packages.
+You should enable *Mirror on Sync* and use download policy *immediate* for such repositories.
```
m-bucher, line 32: `Mirror on Sync`- setting is replaced by `Mirroring Policy` in Katello-4.4 and latest -4.3, but I guess this will be a separate PR.

maximiliankolb: Seconds too late to the party... But I can create a separate PR. We can track this in https://github.com/theforeman/foreman-documentation/issues/1184 for now.

PR #1150, opened 2022-03-11
Expand logging sections

maximiliankolb:
See individual commits.

Cherry-pick into:

* [x] Foreman 3.2
* [x] Foreman 3.1

### DIFF
--- a/guides/common/modules/proc_increasing-the-logging-levels-to-help-with-debugging.adoc
+++ b/guides/common/modules/proc_increasing-the-logging-levels-to-help-with-debugging.adoc
@@ -53,7 +53,7 @@ Edit `/etc/hammer/cli_config.yml` and set the `:log_level:`:
 :log_level: 'debug'
 ----
 
-== Increasing the logging level On {SmartProxy}
+== Increasing the Logging Level On {SmartProxy}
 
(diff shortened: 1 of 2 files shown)


maximiliankolb:
> Thanks, I see `yaml` language highlighting is only used for upstream, I am wondering if it breaks things downstream. If not maybe we can add this everywhere someday.

Thanks for merging.

Yes, syntax highlighting is really neat but for most blocks impossible because we always start shell commands with `# `. If we did this with CSS, then it would make sense to enable syntax highlighting. Source: https://css-tricks.com/to-or-not-to-displaying-terminal-code-snippets/

ekohl:
> Yes, syntax highlighting is really neat but for most blocks impossible because we always start shell commands with `# `. If we did this with CSS, then it would make sense to enable syntax highlighting. Source: https://css-tricks.com/to-or-not-to-displaying-terminal-code-snippets/

In GitHub's markdown flavor you have `console` as highlighting. For example,
```console
$ echo This runs as a normal user
This runs as a normal user
# echo and as root
and as root
```

 Isn't there an equivalent in asciidoctor?

lzap:
Thank you I had an issue in my cherry picking script it did silently ignore merge conflict. Sorry, will fix it.

PR #1154, opened 2022-03-15
Do not use link dereferencing on MacOS

lzap:
Turns out MacOS/BSD does not support dereferencing. It is not a hard requirement, it looks like recursive copy does dereference `images/common` symlink anyway when making hard links on Linux. On MacOS (tested) it also dereference by default so it works as expected.

Cherry-pick into:

* [x] Foreman 3.2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ guides/common/default.css
 /web/crash.log
 /web/.byebug_history
 /web/output/
+/vendor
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -13,3 +13,4 @@ guides/common/default.css
 /web/crash.log
 /web/.byebug_history
 /web/output/
+/vendor
```
lzap, line 16: Closed a PR yesterday when a contributor committed this, making sure this does not stand in the way.

PR #1111, opened 2022-02-28
Use modularity on EL8

ekohl:
Now that there's a proper module on EL8 this should be used.

In practice a user may see conflicts if the Ruby 2.5 or PostgreSQL module has been enabled. Those are the default modules but foreman (and thus katello) pulls in ruby:2.7 and postgresql:12. This means a user must run that version.

In case of PostgreSQL it can be that there's already a version 10 database. In that case an upgrade must be performed, but this is not documented. This can be considered unsupported for installations.

However, there are essentially no real upgrade instructions anyway. It's all part of foreman-maintain but that may result in a chicken-and-egg situation where you would need updated modules to update. So for now I'm not touching that.

This is only relevant for nightly now. Ideally speaking this goes into 3.2 but there are some blockers on the infra side to be resolved.

PR #1163, opened 2022-03-18
Add fixes section to katello quickstart guide + Katello 4.4.0 GA updates

ianballou:
Adds manual 

Cherry-pick into:

* [x] Foreman 3.2
* [x] Foreman 3.1
  * NOTE: The fix related to Debian shouldn't be mentioned for 3.1


guides/doc-Quickstart_Guide/master.adoc
```diff
@@ -42,3 +42,31 @@ ifdef::katello[]
 endif::[]
 
 When the installer has completed, details will be printed about where to find Foreman and the Smart Proxy.
+
+ifdef::katello[]
+== Manual Fixes
```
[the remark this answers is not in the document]
ianballou: I moved it to the end of the main server installation guide.

ekohl (commented):
Normally I prefer to promote 3.2/4.4 to GA as a separate commit. For example, we may also need to include the last release notes. Can we pull that out to a separate PR?

guides/common/modules/con_troubleshooting-katello.adoc
```diff
@@ -0,0 +1,25 @@
+[id="troubleshooting-katello"]
+= Troubleshooting Katello
+
+== "[Errno 1] Operation not permitted: ..." during repository syncing:
+[options="nowrap" subs="+quotes,attributes"]
+----
+# chown -hR pulp.pulp /var/lib/pulp/media/artifact/
```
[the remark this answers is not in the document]
ianballou: Hm yeah `-h` may not be necessary because,  by this point, everything should be `pulp.pulp`.

ianballou:
Sure, I can do the 4.4.0 GA update separately

ekohl (approved):
I'm thinking a bit wider and in #1111 I kind of run into the same problem where I want to document some troubleshooting. The only difference is that there it's common for all upstream builds. That makes me wonder if we should make it generic, so just Troubleshooting instead of Troubleshooting Katello.

For now I'm going merge this and refactor it later. Thanks!

guides/common/modules/con_troubleshooting-katello.adoc
```diff
@@ -0,0 +1,26 @@
+[id="troubleshooting-katello"]
```
ekohl, line 1: My only concern here is that I don't want the ID to change later on. @maximiliankolb thoughts?

maximiliankolb: Is this a special case where we don't want the ID to change?

Other than that, I recommend the following:
```suggestion
[id="Troubleshooting_Katello_{context}"]
```

ekohl: Well, as I said in https://github.com/theforeman/foreman-documentation/pull/1163#pullrequestreview-918448100 I think we should generalize it to just "troubleshooting".

maximiliankolb: All three versions are basically fine for me; right now, I have no strong preference (because we don't include this file in downstream docs).

ianballou:
I changed it to just be "Troubleshooting".  I'm still only allowing it for Katello docs since it will be empty otherwise.

ianballou:
If this is all set, can I have some help merging? :)

maximiliankolb:
`This branch has not been deployed` < I don't know why.

Please add empty lines between the headings & paragraphs and paragraphs & code blocks.

ianballou:
Should be all set now @maximiliankolb. I'm not sure about the `This branch has not been deployed` bit though! Not sure where you're seeing it, maybe because I'm not the one merging.

ianballou:
Oh, that is weird. We'll have to get some help.

lzap:
Can you rebase and push again? This will trigger tests again. Github Actions had an outage yesterday: https://www.githubstatus.com/history

ekohl (approved):
Thanks for the changes. They look good. As said, please rebase for CI but other than that :+1:

ianballou:
Okay, I rebased and pushed.

ianballou:
> Oh, I didn't notice there was no CI at all. The green button can be misleading :(

I'll blame that on github!

PR #1167, opened 2022-03-21
Correct Satellite versions

ekohl:
Current development is tracking what will become Satellite 6.12.

https://github.com/theforeman/foreman-documentation/pull/1166 already corrects 3.1 but I wasn't quite sure what to with 3.2. Technically it doesn't map to any Satellite. You could do something like 3.12-pre but it's probably best to ignore it altogether.

Cherry-pick into:

* [ ] Foreman 3.2
* [ ] Foreman 3.1


ekohl:
> Actually: does this mean we have to (temporarily) disable the link checker?

I ended up changing the linkchecker config. It's now green.

PR #1170, opened 2022-03-21
Correct the default organization name & hide debug certs in non-Katello context

ekohl:
The default organization is named Default Organization, which results in Default_Organization as a label but this doesn't refer to labels.

I contemplated mentioning that it's affecter by the --initial-organization installer argument, but wasn't sure how to mention this clearly.

Debug certificates are a content feature, which only exists with Katello.  This means the content can be removed.

Technically this can be cherry picked but we hide all guides for non-Katello releases from the navigation so it doesn't matter that much. The other change is also pretty trivial.

Cherry-pick into:

* [ ] Foreman 3.2
* [ ] Foreman 3.1


maximiliankolb (approved):
LGTM.

PR #1172, opened 2022-03-21
Bump up stable version to 3.2

ianballou:


ekohl (approved):
I'm wondering about the whole dance of stable vs supported. I'm leaning to 3 possible states:
* RC
* (nothing)
* EOL

I also don't mind merging this now and iterating on it further. Given @upadhyeammit is out that may be a good choice.

ianballou:
Yeah stable vs supported does seem like an unnecessary differentiation. I could also see it being RC - Supported - EOL.

PR #1174, opened 2022-03-21
Added content to a section that was blank

mdolezelova:
Cherry-pick into:

* [ ] Foreman 3.2
* [ ] Foreman 3.1

### DIFF
--- a/guides/common/assembly_installing-capsule-server.adoc
+++ b/guides/common/assembly_installing-capsule-server.adoc
@@ -36,12 +36,6 @@ ifdef::katello,satellite[]
 [id="configuring-capsule-server-with-ssl-certificates"]
 == Configuring {SmartProxyServer} with SSL Certificates
 
-
-ifdef::foreman-deb[]
-This procedure is only for Katello plug-in users.
-endif::[]
-
-ifndef::foreman-deb[]
 {ProjectName} uses SSL certificates to enable encrypted communications between {ProjectServer}, external {SmartProxyServer}s, and all hosts.
 Depending on the requirements of your organization, you must configure your {SmartProxyServer} with a default or custom certificate.
 
@@ -49,7 +43,6 @@ Depending on the requirements of your organization, you must configure your {Sma
 For more information, see xref:configuring-capsule-default-certificate_{smart-proxy-context}[].
 * If you use a custom SSL certificate, you must also configure each external {SmartProxyServer} with a distinct custom SSL certificate.
 For more information, see xref:configuring-capsule-custom-server-certificate_{smart-proxy-context}[].
-endif::[]
 endif::[]
 
 // Assigning the Correct Organization and Location to {SmartProxyServer} in the {ProjectWebUI}
--- a/guides/common/assembly_installing-capsule-server.adoc
+++ b/guides/common/assembly_installing-capsule-server.adoc
@@ -37,6 +37,19 @@ ifdef::katello,satellite[]
 == Configuring {SmartProxyServer} with SSL Certificates
 endif::[]
 
+ifeval::["{build}" == "foreman"]
+This procedure is only for Katello plug-in users.
+endif::[]
+
+ifeval::["{build}" != "foreman-deb"]
+{ProjectName} uses SSL certificates to enable encrypted communications between {ProjectServer}, external {SmartProxyServer}s, and all hosts.
+Depending on the requirements of your organization, you must configure your {SmartProxyServer} with a default or custom certificate.
+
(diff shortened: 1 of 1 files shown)


guides/common/assembly_installing-capsule-server.adoc
```diff
@@ -37,6 +37,19 @@ ifdef::katello,satellite[]
 == Configuring {SmartProxyServer} with SSL Certificates
 endif::[]
 
+ifeval::["{build}" == "foreman"]
```
Lennonka, line 40: I think that you can use a simpler condition here to make it more readable, I'm guessing `ifdef::foreman-deb[]`.
@melcorr Can you please confirm that this suggestion is correct?

guides/common/assembly_installing-capsule-server.adoc
```diff
@@ -37,6 +37,19 @@ ifdef::katello,satellite[]
...
+This procedure is only for Katello plug-in users.
+endif::[]
+
+ifeval::["{build}" != "foreman-deb"]
```
Lennonka, line 44: Similar here: `ifndef::foreman-deb[]`

Lennonka (changes requested):
The text looks good, I just have a minor suggestion about the conditions.

guides/common/assembly_installing-capsule-server.adoc
Lennonka: Missing `endif::[]`.

guides/common/assembly_installing-capsule-server.adoc
```diff
@@ -37,6 +37,19 @@ ifdef::katello,satellite[]
 == Configuring {SmartProxyServer} with SSL Certificates
 endif::[]
```
Lennonka, line 38: This endif is looking weird. Maybe it would be better to leave out both conditions below and add the text of the second one under the condition above?

guides/common/assembly_installing-capsule-server.adoc
melcorr: AFAICT yep. This is probably leftover of super old text somewhere...

mdolezelova:
Thank you for the comments, @Lennonka and @melcorr ! I've fixed that - hope I understood your suggestions well. Really looks like a leftover from some very old text. Please could you check again and merge it if all good? Thank you!

lzap:
Can you rebase and push again? This will trigger tests again. Github Actions had an outage yesterday: https://www.githubstatus.com/history

guides/common/assembly_installing-capsule-server.adoc
```diff
@@ -35,6 +35,21 @@ include::modules/proc_synchronizing-the-system-clock-with-chronyd.adoc[leveloffs
 ifdef::katello,satellite[]
 [id="configuring-capsule-server-with-ssl-certificates"]
 == Configuring {SmartProxyServer} with SSL Certificates
+
+
+ifdef::foreman-deb[]
+This procedure is only for Katello plug-in users.
+endif::[]
```
Lennonka, line 42: I think that this sentence is not necessary in this case, because the whole section will not be generated for foreman-deb sice it's nested under the condition to be output only for katello and satellite.

guides/common/assembly_installing-capsule-server.adoc
```diff
@@ -35,6 +35,21 @@ include::modules/proc_synchronizing-the-system-clock-with-chronyd.adoc[leveloffs
...
+
+ifndef::foreman-deb[]
```
Lennonka, line 44: And for the same reason, this content doesn't need this condition either.

Lennonka (changes requested):
Almost there. :)

Lennonka (approved):
You're welcome. Yes, this works great!

melcorr:
@Lennonka I can squash on merge :)

mdolezelova:
Hi @melcorr , thank you for merging this PR! :-) So I just wanted to make about what is the preferred process - should Satellite doc team members do the squashing or do you prefer upstream stakeholders to do this on merge as you did? I somehow remember you've mentioned Sat docs team shouldn't do the squashing, but maybe I'm just confusing things or do not remember it correctly. Please can you let me know your preference? Thank you! :-)

melcorr:
We can squash on merge @mdolezelova - this has been the standard and we didn't come to consensus, so just like this is fine unless we decide some change.

mdolezelova:
> We can squash on merge @mdolezelova - this has been the standard and we didn't come to consensus, so just like this is fine unless we decide some change.

Okay. Thank you for confirmation and your assistance, @melcorr ! :-)

PR #1179, opened 2022-03-24
Hide Katello only note

maximiliankolb:
This note is unnecessary because we only show these procedures in guides for Katello users.

Cherry-pick into:

* [x] Foreman 3.2
* [x] Foreman 3.1


lzap (approved):

lzap:
Commit(s) 26aa5677 were cherry-picked into: 3.2 3.1
* 2005d56c Hide Katello only note
* 32c45741 Hide Katello only note

PR #1183, opened 2022-03-25
Rework content management

maximiliankolb:
Various small commits to fix various small issues like broken links.

Cherry-pick into:

* [x] Foreman 3.2
* [x] Foreman 3.1

### DIFF
--- a/guides/common/modules/con_content-management-types-overview.adoc
+++ b/guides/common/modules/con_content-management-types-overview.adoc
@@ -1,7 +1,7 @@
 [id="Content_Management_Types_Overview_{context}"]
 = Content Management Types Overview
 
-With {ProjectName}, you can manage the following Red Hat content types:
+With {ProjectName}, you can manage the following content types:
 
 ifdef::satellite[]
 RPM Packages::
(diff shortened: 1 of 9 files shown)


guides/common/modules/con_content-management-types-overview.adoc
```diff
@@ -1,7 +1,12 @@
 [id="Content_Management_Types_Overview_{context}"]
 = Content Management Types Overview
 
+ifdef::satellite[]
 With {ProjectName}, you can manage the following Red Hat content types:
```
ekohl, line 5: I'll let others weigh in, but IMHO it can always manage non-Red Hat content types. Line 47 is wrong and it can manage "files" Pulp 2 called it the ISO store, but Pulp 3 just names it files because that's what it is. Since arbitrary files aren't always Red Hat content, I think you should keep only the `following content types` version. Also, containers are certainly part of Pulp 3 and I wouldn't call that a Red Hat content type.

ekohl (commented):
@Lennonka mind taking a look?

guides/common/modules/con_content-management-types-overview.adoc
Lennonka: I agree with Ewoud, I would keep it generic.

maximiliankolb: Applied.

Lennonka:
Otherwise, looks good!

maximiliankolb:
Rebased to master. cc @Lennonka

Lennonka (approved):

maximiliankolb:
Thanks to all for reviewing this.

Merged and cherry-picked: ✅ 
   402a6350..37cc60e0  3.2 -> 3.2
   4da1472c..38d77998  3.1 -> 3.1

PR #1192, opened 2022-03-29
Cherry-pick: Add troubleshooting section for Katello

ianballou:


lzap (approved):
Thank you, you can simply request CPicks of small PRs directly via a comment in the original PR, we can do this for you.

PR #1196, opened 2022-03-30
Fix include path and reference 2.5

Lennonka:
No cherry-picks.
We need this fix for ccutil build.


Lennonka:
Didn't get it right. Hang on.

PR #1197, opened 2022-03-31
Fix broken links

maximiliankolb:
Cherry-pick into:

* [x] Foreman 3.2
* [ ] Foreman 3.1

For 3.1, I already cherry-picked this which needed conflict resolution. > I will open a separate PR for this.


Lennonka:
LGTM. The linkchecker tests these links during the build, right?

Lennonka (approved):
Looks good to go.

maximiliankolb:
> LGTM. The linkchecker tests these links during the build, right? Can it recognize when an anchor in the URL does not exist at the destination?

In theory, it should. IMHO the issue is that it only checks if the `html` file exists, but not whether there is an anchor as expected.

maximiliankolb:
Thanks for your quick review @Lennonka 

Merged and cherry-picked to 3.2 and 3.1: ✅ 

   be26b1f0..6bda0414  3.2 -> 3.2
   81990f6e..7219ef04  3.1 -> 3.1

PR #1200, opened 2022-04-01
Remove single endif

maximiliankolb:
I broke 3.1 with https://github.com/theforeman/foreman-documentation/commit/7219ef0441321cc73a2de8228e92d7f25e996337 🙈


maximiliankolb:
Merged. @maximiliankolb needs more coffee this early.

PR #1204, opened 2022-04-07
Improvements to local serving

ekohl:
See individual commits for details, but the important part is that with this the cross document links are correctly linking to the local site rather than production.


Makefile
```diff
@@ -22,10 +25,16 @@ clean:
 	$(MAKE) -C guides/ clean
 	rm -rf $(DEST) web/output/
 
-html: prep
-	$(MAKE) -C guides/ html BUILD=foreman-el
-	$(MAKE) -C guides/ html BUILD=foreman-deb
-	$(MAKE) -C guides/ html BUILD=katello
+html: build-foreman-el build-foreman-deb build-katello
```
evgeni, line 28: these new targets should be `.PHONY`

guides/common/Makefile
```diff
@@ -29,14 +29,17 @@ BROWSER_OPEN = open
...
 # The Satellite build uses links which do not resolve. This does mean the link
 # checker has issues with verification of guides that have not yet been built
 # by downstream, but there really isn't a way around this unless we start
 # building with downstream filenames.
 ifneq ($(BUILD), satellite)
+ifeq ($(LINKS), LOCAL)
 BASEURL_ATTR = -a "BaseURL=file://$(BUILD_DIRA)"
 endif
+ifneq ($(BASEURL), "")
+BASEURL_ATTR = -a "BaseURL=$(BASEURL)"
```
evgeni, line 41: doesn't this mean that production builds will also contain a localhost baseurl? I don't see anything in `.github/workflows/deploy.yml` that would unset `BASEURL` or set it to `docs.theforeman.org`?

evgeni: Uh, not sure I like abusing this implementation detail as a differentiator.

Maybe document the var in the readme and explicitly set it to "" in the deployment pipeline?

Lennonka: @evgeni @ekohl Hiii! Is Ewoud still available to help you finish this? Or should we try to ask someone else? :)

ekohl: @evgeni how about this: you do need to pass an env var to opt-in, but at least the behavior is consistent

guides/common/Makefile
```diff
@@ -29,14 +29,24 @@ BROWSER_OPEN = open
...
 BASEURL_ATTR = -a "BaseURL=file://$(BUILD_DIRA)"
 endif
+
+ifdef BASEURL
+	ifdef BASEPATH
+		BASEURL_ATTR = -a "BaseURL=$(BASEURL)$(BASEPATH)"
+	else
+		BASEURL_ATTR = -a "BaseURL=$(BASEURL)"
+	endif
+endif
```
ekohl, line 48: Indenting appears to work, but syntax highlighting doesn't recognize it. Is this valid?

guides/common/Makefile
```diff
@@ -29,14 +29,24 @@ BROWSER_OPEN = open
 CP_ARGS =
 endif
 
-ifeq ($(LINKS), LOCAL)
 # The Satellite build uses links which do not resolve. This does mean the link
 # checker has issues with verification of guides that have not yet been built
 # by downstream, but there really isn't a way around this unless we start
 # building with downstream filenames.
 ifneq ($(BUILD), satellite)
```
ekohl, line 36: I wonder if this is still valid.

guides/common/Makefile
evgeni: sounds good to me

ekohl:
@evgeni does that mean you approve and it can be merged?

evgeni:
I didn't test it, so can't really approve, but the concern I had is solved.

Makefile
```diff
@@ -1,15 +1,22 @@
 SHELL := /bin/bash
 DEST := result
 PORT := 5000
+ifeq ($(LINKS), SERVE)
+BASEURL := http://localhost:$(PORT)/
+endif
 VERSION_LINKS := 3.4 3.3 3.2 3.1 3.0 2.5 2.4
```
maximiliankolb, line 7: We should probably extend the procedure "how to add a new Foreman version" to ensure this list is also extended with 3.5.

ekohl:
While testing this out I found it's annoying that it's not really live. Currently investigating nanoc with live serving.

ekohl:
Rebased to resolve some conflicts, but no real change. https://github.com/theforeman/foreman-documentation/pull/2840 is related work that I want to build on.

PR #1095, opened 2022-02-22
Fix repo Id

mdolezelova:
Cherry-pick into:

* [ x] Foreman 3.1
* For Foreman 3.0 or older, file a separate PR request

PR #1207, opened 2022-04-07
Fixed repo name to reflect that it's a Capsule server

mdolezelova:
Cherry-pick into:

* [ ] Foreman 3.2
* [ ] Foreman 3.1

### DIFF
--- a/guides/common/assembly_installing-capsule-server.adoc
+++ b/guides/common/assembly_installing-capsule-server.adoc
@@ -19,7 +19,7 @@ endif::[]
 
 ifdef::foreman-el,foreman-deb,satellite[]
 // Configuring Repositories
-include::modules/proc_configuring-repositories.adoc[leveloffset=+1]
+include::modules/proc_configuring-repositories-proxy.adoc[leveloffset=+1]
 endif::[]
 
 // Installing {SmartProxyServer} Packages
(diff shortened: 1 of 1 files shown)


melcorr:
You're fixing a file name rather than a repo name, right?

mdolezelova:
Hi @melcorr , this is really a confusing case. If you search for the BZ#2056143, it reports that there is a wrong repository name in one section. However, I was already fixing it some time ago in https://github.com/theforeman/foreman-documentation/pull/1095. At that time, I checked the repo ID was fixed correctly, republished the guide and closed the BZ. Yesterday, the BZ was reopened as it was reported that the issue is still there. Not sure how this could happen. I saw there are two similar modules -proc_configuring-repositories.adoc and proc_configuring-repositories-proxy.adoc. So I assumed the first one would be relevant for the Installing Satellite Server guide ans the second one would be relevant for Installing Capsule Server guide. So I though wrong file is included and that's why in this PR I've replaced it with the file relevant for the Capsule (proxy).  However, this still doesn't explain how the change done via https://github.com/theforeman/foreman-documentation/pull/1095 could be reverted. Apologies for the mess, but please do you have any suggestion what to do here?

ekohl:
I think the quickest way now is to keep the current code for upstream, but drop the `satellite` from `ifdef` and make a new block. Then you can also drop the upstream instructions from the proxy document.

mdolezelova:
> I think the quickest way now is to keep the current code for upstream, but drop the `satellite` from `ifdef` and make a new block. Then you can also drop the upstream instructions from the proxy document.

@ekohl , yes, I've created a separate condition for satellite and kept the original file for upstream. Please check my last commit. If it's fine like this, can we merge it, please?
Thank you!

ekohl (approved):
Thanks!

mdolezelova:
Thank you, @ekohl ! :-)

PR #1208, opened 2022-04-07
Fixed repo ID to reflect that it's Capsule Server

mdolezelova:
Cherry-pick into:

* [x] Foreman 3.2
* [x] Foreman 3.1

### DIFF
--- a/guides/common/assembly_installing-capsule-server.adoc
+++ b/guides/common/assembly_installing-capsule-server.adoc
@@ -19,7 +19,7 @@ endif::[]
 
 ifdef::foreman-el,foreman-deb,satellite[]
 // Configuring Repositories
-include::modules/proc_configuring-repositories.adoc[leveloffset=+1]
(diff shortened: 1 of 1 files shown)


ekohl (changes requested):
For upstream, I'd rather not do this. The content of `modules/proc_configuring-repositories-proxy.adoc` is messy and (at least for upstream) `modules/proc_configuring-repositories.adoc` is the correct way.

ekohl:
So the context is that in https://github.com/theforeman/foreman-documentation/commit/bfb5cd8e1ec6105594531bf81058882fee8efaff I changed it because it was clearly wrong for upstream. I missed that it was correct in downstream.

evgeni:
Hmm, and then we ended up "breaking" it again anyways in https://github.com/theforeman/foreman-documentation/commit/3c590ac7b5b1d14a3b746c263553fdf70b167252 :(

evgeni:
Fixed in #1327

PR #1213, opened 2022-04-11
Introduce macros of the pgsql lib paths on EL7 & EL8

ekohl:
Currently the installer is very EL7 focused while in nightly it's about to be dropped. The idea behind this is to make it clear what EL 7 and 8 is, thus preparing for the future.

It also makes one other change. In the CentOS 7 storage requirements the path is aligned with RHEL 7.

Cherry-pick into:

* [ ] Foreman 3.2
* [ ] Foreman 3.1


ekohl:
Given we're now actively dropping this, I think it can be closed.

@upadhyeammit this may be relevant for you if you want to properly move to EL8 though.

PR #982, opened 2022-01-18
Merge installing on CentOS and RHEL

ekohl:
These sections are the same. Merging this makes the entire guide shorter.

This builds on https://github.com/theforeman/foreman-documentation/pull/973 (included). I expect there to be some discussion, so marking it as a draft for now. Please don't look at the specifics right now but rather on the concept, which is:

* Use Enterprise Linux when refering to RHEL and derivatives upstream
* Use Red Hat Enterprise Linux in Satellite
* Use specific distro names when the instructions differ (like setting up repos)

Is this a direction we can agree on?

PR #1215, opened 2022-04-11
Drop EL7 from the nightly quickstart installation guide

upadhyeammit:
Cherry-pick into:
None

### DIFF
--- a/guides/common/modules/proc_configuring-repositories.adoc
+++ b/guides/common/modules/proc_configuring-repositories.adoc
@@ -6,10 +6,6 @@ Use this procedure to enable the repositories that are required to install {Prod
 
 ifdef::foreman-el,katello[]
 * xref:#repositories-centos-8[CentOS 8]
-* xref:#repositories-centos-7[CentOS 7]
-endif::[]
-ifdef::foreman-el,katello,satellite[]
-* xref:#repositories-rhel-7[{RHEL} 7]
 endif::[]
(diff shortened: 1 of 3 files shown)


guides/common/modules/proc_configuring-repositories.adoc
```diff
@@ -6,9 +6,8 @@ Use this procedure to enable the repositories that are required to install {Prod
 
 ifdef::foreman-el,katello[]
 * xref:#repositories-centos-8[CentOS 8]
-* xref:#repositories-centos-7[CentOS 7]
 endif::[]
-ifdef::foreman-el,katello,satellite[]
+ifdef::satellite[]
```
ekohl, line 10: I'm not sure this is needed. It's not like Satellite is going to support EL7 either.

ekohl (commented):
I had hoped https://github.com/theforeman/foreman-documentation/pull/982 would first get attention since it makes this easier. I also think that we should assume Satellite also drops RHEL 7 support. This does imply that everywhere it must be RHEL 8.

It may make https://github.com/theforeman/foreman-documentation/pull/1213 obsolete and we can simplify it by always using the EL 8 path.

guides/common/modules/proc_configuring-repositories.adoc
upadhyeammit: Removed!

melcorr (approved):
LGTM!

PR #1214, opened 2022-04-11
Changed attributes for Satellite to version 6.11

mdolezelova:
Cherry-pick into:

* [ ] Foreman 3.2
* [x] Foreman 3.1
* For Foreman 3.0 or older, file a separate PR request

PR #1238, opened 2022-04-22
Document how to use linkchecker

maximiliankolb:
See https://github.com/theforeman/foreman-documentation/pull/1214#issuecomment-1106443054


melcorr:
I really appreciate you doing this @maximiliankolb

PR #1244, opened 2022-04-26
Replaced the new option in Recovering a Repository topic

mjivraja:
A new option was replaced i.e "Verify Content Checksum" with
Validate Content Sync as it is no longer an option for an advanced
synchronization in Satellite 6.10

"Validate Content Sync" is no longer an option for "Recovering a
Repository"

https://issues.redhat.com/browse/SATDOC-728

Cherry-pick into:

* [X] Foreman 3.2
* [X] Foreman 3.1

### DIFF
--- a/guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
+++ b/guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
@@ -436,7 +436,6 @@ Complete Sync::
 Synchronizes all RPMs regardless of detected changes.
 Use this option if specific RPMs could not be downloaded to the local repository even though they exist in the upstream repository.
 
-//Validate Content Sync::
 Verify Content Checksum::
 Synchronizes all RPMs and then verifies the checksum of all RPMs locally.
 If the checksum of an RPM differs from the upstream, it re-downloads the RPM.
@@ -453,7 +452,10 @@ Use this option if you have one of the following errors:
 . Select the product containing the corrupted repository.
 . Select the name of a repository you want to synchronize.
 . To verify the checksum, click *Verify Content Checksum* from the *Action* menu.
-. To perform optimised sync and complete sync, select *Advanced Sync* from the *Select Action* menu. And then, select the option and click *Sync*.
+. To perform optimised sync and complete sync, select *Advanced Sync*, perform the following:
+.. Select *Advanced Sync* from the *Select Action* menu.
+.. Select the desired option.
+. Click *Sync*.
 
 .CLI procedure
 
--- a/guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
+++ b/guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
@@ -451,11 +451,9 @@ Use this option if you have one of the following errors:
 . In the {Project} web UI, navigate to *Content* > *Products*.
 . Select the product containing the corrupted repository.
 . Select the name of a repository you want to synchronize.
-. To verify the checksum, click *Verify Content Checksum* from the *Action* menu.
-. To perform optimised sync and complete sync, select *Advanced Sync*, perform the following:
-.. Select *Advanced Sync* from the *Select Action* menu.
-.. Select the desired option.
-. Click *Sync*.
+. To perform optimized sync or complete sync, select *Advanced Sync* from the *Select Action* menu.
+. Select the required option and click *Sync*.
+. To verify the checksum, click *Verify Content Checksum* from the *Select Action* menu. (optional)
 
 .CLI procedure
 
--- a/guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
+++ b/guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
@@ -436,7 +436,8 @@ Complete Sync::
 Synchronizes all RPMs regardless of detected changes.
 Use this option if specific RPMs could not be downloaded to the local repository even though they exist in the upstream repository.
 
-Validate Content Sync::
+//Validate Content Sync::
+Verify Content Checksum::
 Synchronizes all RPMs and then verifies the checksum of all RPMs locally.
 If the checksum of an RPM differs from the upstream, it re-downloads the RPM.
 This option is relevant only for `yum` repositories.
@@ -451,8 +452,8 @@ Use this option if you have one of the following errors:
 . In the {Project} web UI, navigate to *Content* > *Products*.
 . Select the product containing the corrupted repository.
 . Select the name of a repository you want to synchronize.
-. From the *Select Action* menu, select *Advanced Sync*.
-. Select the option and click *Sync*.
+. To verify the checksum, click *Verify Content Checksum* from the *Action* menu.
+. To perform optimised sync and complete sync, select *Advanced Sync* from the *Select Action* menu. And then, select the option and click *Sync*.
 
 .CLI procedure
 


guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
```diff
@@ -451,8 +452,8 @@ Use this option if you have one of the following errors:
 . In the {Project} web UI, navigate to *Content* > *Products*.
 . Select the product containing the corrupted repository.
 . Select the name of a repository you want to synchronize.
-. From the *Select Action* menu, select *Advanced Sync*.
-. Select the option and click *Sync*.
+. To verify the checksum, click *Verify Content Checksum* from the *Action* menu.
```
AkshayGadhaveRH, line 455: . To verify the checksum, click *Verify Content Checksum* from the *Select Action* menu.

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
```diff
@@ -436,7 +436,8 @@ Complete Sync::
 Synchronizes all RPMs regardless of detected changes.
 Use this option if specific RPMs could not be downloaded to the local repository even though they exist in the upstream repository.
 
-Validate Content Sync::
+//Validate Content Sync::
```
melcorr, line 439: old content should be removed rather than commented out.

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
melcorr: Orient the user first as per IBM style, so @mjivraja is slightly more correct here. Depending on the style guide, this rule changes, but we follow IBM here.

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
```diff
@@ -451,8 +452,8 @@ Use this option if you have one of the following errors:
 . In the {Project} web UI, navigate to *Content* > *Products*.
 . Select the product containing the corrupted repository.
 . Select the name of a repository you want to synchronize.
-. From the *Select Action* menu, select *Advanced Sync*.
-. Select the option and click *Sync*.
+. To verify the checksum, click *Verify Content Checksum* from the *Action* menu.
+. To perform optimised sync and complete sync, select *Advanced Sync* from the *Select Action* menu. And then, select the option and click *Sync*.
```
melcorr, line 456: We use American English in the docs so ized rather than ised. 
We don't start paragraphs with And. 
What option? we should tell them what to select.

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
```diff
@@ -451,8 +452,8 @@ Use this option if you have one of the following errors:
 . In the {Project} web UI, navigate to *Content* > *Products*.
 . Select the product containing the corrupted repository.
 . Select the name of a repository you want to synchronize.
-. From the *Select Action* menu, select *Advanced Sync*.
```
AkshayGadhaveRH, line 454: I think this is still relevant, as optimized sync and complete sync are still accessible after clicking on Advanced sync. The option here refers to what the user might want to select, either optimized or complete. The section after this elaborates on the meaning of the different options. It probably might be better to add another section for "Verify Content Checksum" outside of the advanced sync section.

mjivraja:
Hi @AkshayGadhaveRH @melcorr 

I have taken the changes by considering your points and Shweta's as well. Please consider my PR further and also let me know if there is anything from my side.

Thank you.

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
```diff
@@ -451,8 +451,11 @@ Use this option if you have one of the following errors:
...
-. Select the option and click *Sync*.
+. To verify the checksum, click *Verify Content Checksum* from the *Action* menu.
+. To perform optimised sync and complete sync, select *Advanced Sync*, perform the following:
```
melcorr, line 455: We don't use subsections and I don't see the value of a subsection here as this is as clear when written in a linear procedure. Are you performing one or two syncs? If you are selecting one of two, you need to use or rather than and. 
```suggestion
. To perform optimised sync or complete sync, select *Advanced Sync*.
```

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
```diff
@@ -451,8 +451,11 @@ Use this option if you have one of the following errors:
...
+.. Select *Advanced Sync* from the *Select Action* menu.
```
melcorr, line 456: seems like a duplicate of the above line?

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
```diff
@@ -451,8 +451,11 @@ Use this option if you have one of the following errors:
...
+.. Select the desired option.
```
melcorr, line 457: Do not use sub-steps. There's no reason as the text reads as well and isn't complicated. 
We avoid sub steps if possible. 

IBM Style says never ever use desire in technical writing. 

desired adj
Do not use. Use “required” or “the <item> that you want.”

Is this the point where the user selects from optimised or complete sync? Say that.

melcorr (commented):
I have left some questions. 

IBM Style points to learn from this: 

Never use desire. 

Orient the user. Pp 87 of the IBM Style Guide.

If you orient the user to the location where the action is performed or explain the purpose of the
action, do so before describing the action.
Examples (incorrect)
1. Select a feature from the list in the Installation window.
1. Click Replace to overwrite the previous version.
Examples (correct)
1. In the Installation window, select a feature from the list.
1. To overwrite the previous version, click Replace.

We also avoid substeps except in very complicated workflows. This is not a complicated workflow and the steps can move in a linear fashion from one action to the next without need for indentation. 

I hope this is helpful.

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
mjivraja: Yes, it is the same as above for which you have a suggestion.

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
mjivraja: Hi @melcorr,

Before committing these changes, I had a session with Shweta. Although she wasn't confident about sub-sections, she felt that we are left with no other options and hence, we opted for this. Writing style and word selection are kept the same except for the changes that we took (the last 2 points).

I will replace the word 'desire' with 'required' as I can see the same on IBM Style Guide Quick Reference.

I will also take the changes on the above suggestion where the 'and' can be replaced with 'or'.

At last, I have not taken any decision on the style guide or sentence structure in this commit as it had minor modifications & this can be the probable reason why I am assigned this type of ticket at the current level.

I can understand from your suggestions that there can be a better writing style and sentence structure. But, the advanced sync and verify content checksum are two different actions that are included in the same drop-down menu which can be the probable reason for confusion, as per my understanding.

Please guide me further to make this commit perfect.

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
melcorr: Then you don't need to say it twice.

guides/doc-Content_Management_Guide/topics/Importing_Content.adoc
melcorr: > Hi @melcorr,
> 
> Before committing these changes, I had a session with Shweta. Although she wasn't confident about sub-sections, she felt that we are left with no other options and hence, we opted for this. Writing style and word selection are kept the same except for the changes that we took (the last 2 points).
I don't understand the "left with no other options" comment because I cannot see the complication that this PR is introducing. 
> 
> I will replace the word 'desire' with 'required' as I can see the same on IBM Style Guide Quick Reference.

If you are selecting between two options, this or that, you could jjust say that. "required" is ok but less descriptive.
My main issue is that desired/required aren't descriptive enough. Do you know what that menu reveals at this point?  
> 
> I will also take the changes on the above suggestion where the 'and' can be replaced with 'or'.
> 
> At last, I have not taken any decision on the style guide or sentence structure in this commit as it had minor modifications & this can be the probable reason why I am assigned this type of ticket at the current level.

That's fair. Everything is a learning opportunity all the same. 
> 
> I can understand from your suggestions that there can be a better writing style and sentence structure. But, the advanced sync and verify content checksum are two different actions that are included in the same drop-down menu which can be the probable reason for confusion, as per my understanding.
> 
> Please guide me further to make this commit perfect.

We've avoided sub-procedures everywhere so far. If all the steps are happening in a sequence, then they can remain in that same sequence. If it is an optional step, which means it mightn't happen at all, it would need to be marked as optional. If the user must choose something, tell the user what the options are. 

I hope this is helping a bit, Malhar.  I understand that you're learning and eager to learn.

mjivraja: > If you are selecting between two options, this or that, you could jjust say that. "required" is ok but less descriptive.
> My main issue is that desired/required aren't descriptive enough. Do you know what that menu reveals at this point?

I think the description of all 3 options are there just before the procedure steps. So my question is - the description will still require in the procedure step as well?

> We've avoided sub-procedures everywhere so far. If all the steps are happening in a sequence, then they can remain in that same sequence. If it is an optional step, which means it mightn't happen at all, it would need to be marked as optional. If the user must choose something, tell the user what the options are.
> 
> I hope this is helping a bit, Malhar. I understand that you're learning and eager to learn.

Verify Content Checksum can be the optional step as per my understanding. We can mention it as " (optional) " in procedure steps as per your suggestion. However, I will clear these queries once again with peers.

So, at the end:

1) I will replace require with desire.
2) I will re-verify the options which are mandatory and optional..
3) Sub-steps will be removed and will cover those steps in a single point, even if it requires one more point to be added.

Please let me know if I am missing something.

mjivraja:
> Can you check is this also relevant to all newer versions?

I checked with SME Matt Jia, the same changes will be applicable to v6.11 as well.

mjivraja:
Hi @melcorr 

I have taken all the changes suggested and pushed them. Please review it at your convenience.

Thank you.

melcorr (approved):
LGTM @mjivraja 
Thank you for your patience in working through the issues. 
Can you now raise a PR with this content for the master branch also?

mjivraja:
> LGTM @mjivraja 
> Thank you for your patience in working through the issues. 
> Can you now raise a PR with this content for the master branch also? 
> 

Sure, @melcorr 

Soon, I will raise the PR for Master as well.

In fact, I am thankful to you for your guidance. :)

PR #1248, opened 2022-05-04
Created module for RHEL 7 to RHEL 8 upgrade

bangelic:
Created a new module for the master.adoc for Upgrading and Updating Foreman. This module is titled proc_upgrading-el7-to-el8.adoc. The content came from an existing document and was cleaned up for minimalist purposes and grammar. The content was pulled from this link: https://hackmd.io/@evgeni/B1AT256Q5

@melcorr This is the one you needed a PR for. Needed to clean up the content to fit minimalist principles. Did not see any attributes for RHEL so let me know if I missed that.

Cherry-pick into:

* [ ] Foreman 3.2
* [ ] Foreman 3.1

### DIFF
--- a/guides/common/modules/proc_upgrading-el7-to-el8.adoc
+++ b/guides/common/modules/proc_upgrading-el7-to-el8.adoc
@@ -1,4 +1,4 @@
-[id="Upgrading-Project-from-EL7-to-EL8{context}"]
+[id="Upgrading-Project-from-EL7-to-EL8_{context}"]
 ifdef::satellite[]
 = Upgrading {Project} from {RHEL} 7 to {RHEL} 8
 endif::[]
@@ -6,38 +6,72 @@ ifdef::katello,foreman-el[]
 = Upgrading {Project} from Enterprise Linux 7 to Enterprise Linux 8
 endif::[]
 
+ifdef::satellite[]
 Use this procedure to upgrade your {Project} installation from {RHEL} 7 to {RHEL} 8.
 LEAPP is the supported tool for {RHEL} 7 to {RHEL} 8 upgrades.
-
-.Prerequisites
-ifdef::katello,foreman-el[]
-* {Project} {ProjectVersionPrevious} running on CentOS Linux 7.
-- {Project} installations running on {RHEL} 7 cannot upgrade to {RHEL} 8. You must have a {RHEL} compatible LEAPP build to upgrade.
 endif::[]
 
+ifdef::katello,foreman-el[]
+Use this procedure to upgrade your {Project} installation from Enterprise Linux 7 to Enterprise Linux 8.
+LEAPP is the supported tool for {RHEL} 7 to {RHEL} 8 upgrades.
+endif::[]
+
+
+ifdef::katello,foreman-el[]
+.Prerequisites
+* {Project} {ProjectVersionPrevious} running on CentOS Linux 7.
+- {Project} installations running on {RHEL} 7 cannot upgrade to {RHEL} 8.
+You must have a {RHEL} compatible LEAPP build to upgrade.
 * Access to available repositories or a local mirror of repositories.
 * No outstanding updates.
-ifdef::katello,foreman-el[]
 * {Project} installations can be upgraded to CentOS Stream 8 and a {RHEL} rebuild.
 endif::[]
 
+ifdef::satellite[]
+.Prerequisites
+* Access to available repositories or a local mirror of repositories.
+* No outstanding updates.
+
 .Procedure
+[start=1]
+. Ensure `leapp preupgrade` has no issues.
+[start=2]
+. Run:
+----
+# leapp upgrade
+----
+[start=3]
+. Reboot the system.
+[start=4]
+. After the system reboots, a live system conducts the upgrade, reboots to fix SELinux labels, then reboots into the final {RHEL} 8 system.
+[start=5]
+. After the system reboots, LEAPP finishes the upgrade, watch it with:
+----
+# journalctl -u leapp_resume -f
+----
+
+[NOTE]
+====
+If you install the system and need to use `--disable-system-checks`, the last step of the upgrade is going to fail, and you need to call `{foreman-installer} --disable-system-checks` manually once the system reboots.
+====
+endif::[]
+
 ifdef::katello,foreman-el[]
+.Procedure
+[start=1]
 . Configure the https://copr.fedorainfracloud.org/coprs/g/theforeman/leapp/[@theforeman/leapp COPR Repository], which contains newer LEAPP packages than those shipped by https://wiki.almalinux.org/elevate/[AlmaLinux/ELevate], and support {Project} upgrades:
 
-+
 ----
 # curl -o /etc/yum.repos.d/theforeman-leapp.repo https://copr.fedorainfracloud.org/coprs/g/theforeman/leapp/repo/epel-7/group_theforeman-leapp-epel-7.repo
 ----
+[start=2]
 . Install required packages:
-+
 ----
 # yum install leapp leapp-repository leapp-data-centos
 ----
 
-ifdef::katello,foreman-el[]
+[start=3]
 . Add {Project} specific repositories to `/etc/leapp/files/leapp_upgrade_repositories.repo`:
-+
 ----
 [leapp-foreman]
 name=Foreman 3.2
@@ -98,46 +132,40 @@ gpgcheck=1
 ----
 endif::[]
 
+ifdef::katello,foreman-el[]
 * If you are using Puppet 6 instead of Puppet 7, replace the `7` with a `6` in the `leapp-puppet7` entry.
 
 * You need a Puppet repository for the Puppet agent that the installer is using.
 endif::[]
 
 ifdef::katello,foreman-el[]
+[start=4]
 . We do not support Enterprise Linux 8 installations with EPEL 8 enabled, so remove `epel-release`:
-+
 ----
 # yum remove epel-release
 ----
-
+[start=5]
 . Remove `centos-release-scl` and `centos-release-scl-rh` repositories:
-+
 ----
 # yum remove centos-release-scl centos-release-scl-rh
 ----
+[start=6]
 . Let LEAPP analyze your system:
-+
 ----
 # leapp preupgrade
 ----
-
 The first run is expected to fail but report issues.
 Continue to the next step for remediation.
 
 [start=7]
 . Fix issues found by LEAPP (check `/var/log/leapp/leapp-report.txt` for details):
-+
 ----
 # rmmod pata_acpi
 # echo PermitRootLogin yes | tee -a /etc/ssh/sshd_config
 # leapp answer --section remove_pam_pkcs11_module_check.confirm=True
 ----
-+
-endif::[]
-
 ifdef::katello,foreman-el[]
 ** The `preupgrade` might fail with a dependency resolution error such as:
-+
 
 * "package rubygem-fx-0.5.0-2.el8.noarch requires rubygem(railties) >= 4.0.0, but none of the providers can be installed"
 * "package rubygem-railties-6.0.4.7-1.el8.noarch requires rubygem(thor) < 2.0, but none of the providers can be installed"
@@ -156,30 +184,32 @@ Remove `rubygem-thor`:
 ----
 endif::[]
 
+ifdef::katello,foreman-el[]
 [start=8]
 . Ensure `leapp preupgrade` has no issues.
-+
+endif::[]
+
+ifdef::katello,foreman-el[]
+[start=9]
 . Run:
-+
 ----
 # leapp upgrade
 ----
-
-+
-
-. Reboot the system.
-+
-ifdef::satellite[]
-. After the system reboots, a live system conducts the upgrade, reboots to fix SELinux labels, then reboots into the final {RHEL} 8 system.
-+
 endif::[]
+
+
 ifdef::katello,foreman-el[]
-. After the system reboots, a live system conducts the upgrade, reboots to fix SELinux labels, then reboots into the final Enterprise Linux 8 system.
-+
+[start=10]
+. Reboot the system.
 endif::[]
 
+ifdef::katello,foreman-el[]
+[start=11]
+. After the system reboots, a live system conducts the upgrade, reboots to fix SELinux labels, then reboots into the final Enterprise Linux 8 system.
+
+
+[start=12]
 . After the system reboots, LEAPP finishes the upgrade, watch it with:
-+
 ----
 # journalctl -u leapp_resume -f
 ----
@@ -188,3 +218,4 @@ endif::[]
 ====
 If you install the system and need to use `--disable-system-checks`, the last step of the upgrade is going to fail, and you need to call `{foreman-installer} --disable-system-checks` manually once the system reboots.
 ====
+endif::[]
--- a/guides/common/modules/proc_upgrading-el7-to-el8.adoc
+++ b/guides/common/modules/proc_upgrading-el7-to-el8.adoc
@@ -8,13 +8,21 @@ LEAPP is the supported tool for Enterprise Linux 7 to Enterprise Linux 8 upgrade
 * {Project} {ProjectVersion} running on Enterprise Linux 7.
 * Access to available repositories or a local mirror of repositories.
 * No outstanding updates.
-* {Project} installations can be upgraded to CentOS Stream 8 and a {RHEL} rebuild.
+* {Project} installations running on CentOS 7 can be upgraded to CentOS Stream 8 or a {RHEL} rebuild.
 
 .Procedure
-. Do not perform this step on {RHEL} as this is for CentOS. Configure the https://copr.fedorainfracloud.org/coprs/g/theforeman/leapp/[@theforeman/leapp COPR Repository], which contains newer LEAPP packages than those shipped by https://wiki.almalinux.org/elevate/[AlmaLinux/ELevate], and support {Project} upgrades:
+. Configure the repositories to obtain LEAPP.
++
+On CentOS, configure the https://copr.fedorainfracloud.org/coprs/g/theforeman/leapp/[@theforeman/leapp COPR Repository], which contains newer LEAPP packages than those shipped by https://wiki.almalinux.org/elevate/[AlmaLinux/ELevate], and support {Project} upgrades:
 +
 ----
 # curl -o /etc/yum.repos.d/theforeman-leapp.repo https://copr.fedorainfracloud.org/coprs/g/theforeman/leapp/repo/epel-7/group_theforeman-leapp-epel-7.repo
+----
++
+On {RHEL}, enable the `rhel-7-server-extras-rpms` repository:
++
+----
+# subscription-manager repos --enable=rhel-7-server-extras-rpms
 ----
 
 . Install required packages:
@@ -23,20 +31,20 @@ LEAPP is the supported tool for Enterprise Linux 7 to Enterprise Linux 8 upgrade
 # yum install leapp leapp-repository
 ----
 
-. Install additional OS specific packages (`leapp-data-centos` for CentOS Stream, `leapp-data-rocky` for Rocky Linux, `leapp-data-almalinux` for AlmaLinux`). Not required for {RHEL}.
+. Install additional OS specific packages (`leapp-data-centos` for CentOS Stream, `leapp-data-rocky` for Rocky Linux, `leapp-data-almalinux` for AlmaLinux). Not required for {RHEL}.
 +
 ----
 # yum install leapp-data-centos
 ----
 
 +
-[options="nowrap", subs="+quotes,verbatim,attributes"]
 . Add {Project} specific repositories to `/etc/leapp/files/leapp_upgrade_repositories.repo`:
 +
+[options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
 [leapp-foreman]
 name=Foreman {ProjectVersion}
-baseurl=https://yum.theforeman.org/releases/{ProjectVersion}/3.2/el8/$basearch
+baseurl=https://yum.theforeman.org/releases/{ProjectVersion}/el8/$basearch
 gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman
 enabled=1
 gpgcheck=1
@@ -117,7 +125,8 @@ For more information, see link:https://docs.theforeman.org/nightly/Managing_Conf
 The first run is expected to fail but report issues.
 Continue to the next step for remediation.
 
-. Fix issues found by LEAPP (check `/var/log/leapp/leapp-report.txt` for details):
+. Examine the report in the `/var/log/leapp/leapp-report.txt` file, answer all questions (using `leapp answer`) and manually resolve the other reported problems.
+The following commands show the most common steps required:
 +
 ----
 # rmmod pata_acpi
--- a/guides/common/modules/proc_upgrading-el7-to-el8.adoc
+++ b/guides/common/modules/proc_upgrading-el7-to-el8.adoc
@@ -1,23 +1,9 @@
 [id="Upgrading-Project-from-EL7-to-EL8_{context}"]
-ifndef::satellite[]
-= Upgrading {Project} from {RHEL} 7 to {RHEL} 8
-endif::[]
-ifdef::katello,foreman-el[]
 = Upgrading {Project} from Enterprise Linux 7 to Enterprise Linux 8
-endif::[]
 
-ifdef::satellite[]
-Use this procedure to upgrade your {Project} installation from {RHEL} 7 to {RHEL} 8.
-LEAPP is the supported tool for {RHEL} 7 to {RHEL} 8 upgrades.
-endif::[]
-
-ifdef::katello,foreman-el[]
 Use this procedure to upgrade your {Project} installation from Enterprise Linux 7 to Enterprise Linux 8.
-LEAPP is the supported tool for {RHEL} 7 to {RHEL} 8 upgrades.
-endif::[]
+LEAPP is the supported tool for Enterprise Linux 7 to Enterprise Linux 8 upgrades.
 
-
-ifdef::katello,foreman-el[]
 .Prerequisites
 * {Project} {ProjectVersionPrevious} running on CentOS Linux 7.
 * {Project} installations running on {RHEL} 7 cannot upgrade to {RHEL} 8.
@@ -25,40 +11,7 @@ You must have a {RHEL} compatible LEAPP build to upgrade.
 * Access to available repositories or a local mirror of repositories.
 * No outstanding updates.
 * {Project} installations can be upgraded to CentOS Stream 8 and a {RHEL} rebuild.
-endif::[]
 
-ifdef::satellite[]
-.Prerequisites
-* Access to available repositories or a local mirror of repositories.
-* No outstanding updates.
-
-.Procedure
-[start=1]
-. Ensure `leapp preupgrade` has no issues.
-[start=2]
-. Run:
-+
-----
-# leapp upgrade
-----
-[start=3]
-. Reboot the system.
-[start=4]
-. After the system reboots, a live system conducts the upgrade, reboots to fix SELinux labels, then reboots into the final {RHEL} 8 system.
-[start=5]
-. After the system reboots, LEAPP finishes the upgrade, watch it with:
-+
-----
-# journalctl -u leapp_resume -f
-----
-
-[NOTE]
-====
-If you install the system and need to use `--disable-system-checks`, the last step of the upgrade is going to fail, and you need to call `{foreman-installer} --disable-system-checks` manually once the system reboots.
-====
-endif::[]
-
-ifdef::katello,foreman-el[]
 .Procedure
 [start=1]
 . Configure the https://copr.fedorainfracloud.org/coprs/g/theforeman/leapp/[@theforeman/leapp COPR Repository], which contains newer LEAPP packages than those shipped by https://wiki.almalinux.org/elevate/[AlmaLinux/ELevate], and support {Project} upgrades:
@@ -82,7 +35,7 @@ ifdef::katello,foreman-el[]
 name=Foreman 3.2
 baseurl=https://yum.theforeman.org/releases/3.2/el8/$basearch
 enabled=1
-gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-puppet7-release
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman
 enabled=1
 gpgcheck=1
 
@@ -94,7 +47,6 @@ gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman
 enabled=1
 gpgcheck=1
 module_hotfixes=1
-endif::[]
 
 [leapp-katello-candlepin]
 name=Candlepin: an open source entitlement management system.
@@ -111,6 +63,7 @@ gpgkey=https://yum.theforeman.org/pulpcore/3.16/GPG-RPM-KEY-pulpcore
 enabled=1
 gpgcheck=1
 module_hotfixes=1
+endif::[]
 
 [leapp-foreman-plugins]
 name=Foreman plugins 3.2
@@ -135,15 +88,11 @@ gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-puppet7-release
 enabled=1
 gpgcheck=1
 ----
-endif::[]
 
-ifdef::katello,foreman-el[]
 * If you are using Puppet 6 instead of Puppet 7, replace the `7` with a `6` in the `leapp-puppet7` entry.
 
 * You need a Puppet repository for the Puppet agent that the installer is using.
-endif::[]
 
-ifdef::katello,foreman-el[]
 [start=4]
 . We do not support Enterprise Linux 8 installations with EPEL 8 enabled, so remove `epel-release`:
 +
@@ -173,14 +122,14 @@ Continue to the next step for remediation.
 # echo PermitRootLogin yes | tee -a /etc/ssh/sshd_config
 # leapp answer --section remove_pam_pkcs11_module_check.confirm=True
 ----
-ifdef::katello,foreman-el[]
+
 The `preupgrade` might fail with a dependency resolution error such as:
 
 * "package rubygem-fx-0.5.0-2.el8.noarch requires rubygem(railties) >= 4.0.0, but none of the providers can be installed"
 * "package rubygem-railties-6.0.4.7-1.el8.noarch requires rubygem(thor) < 2.0, but none of the providers can be installed"
 
 If this happens, do the following to clean up packages that cannot automatically upgrade:
-
++
 ----
 # dnf remove $(dnf repoquery --extras --exclude '*katello*' --exclude '*pulp*' --exclude '*localhost*' --exclude "*$HOSTNAME*" --exclude libmodulemd)
 ----
@@ -188,36 +137,26 @@ If this happens, do the following to clean up packages that cannot automatically
 Check the output if it does not contain any local packages.
 
 Remove `rubygem-thor`:
-
++
 ----
 # yum remove rubygem-thor
 ----
-endif::[]
 
-ifdef::katello,foreman-el[]
 [start=8]
 . Ensure `leapp preupgrade` has no issues.
-endif::[]
 
-ifdef::katello,foreman-el[]
 [start=9]
 . Run:
 +
 ----
 # leapp upgrade
 ----
-endif::[]
 
-
-ifdef::katello,foreman-el[]
 [start=10]
 . Reboot the system.
-endif::[]
 
-ifdef::katello,foreman-el[]
 [start=11]
 . After the system reboots, a live system conducts the upgrade, reboots to fix SELinux labels, then reboots into the final Enterprise Linux 8 system.
-
 
 [start=12]
 . After the system reboots, LEAPP finishes the upgrade, watch it with:
@@ -230,4 +169,3 @@ ifdef::katello,foreman-el[]
 ====
 If you install the system and need to use `--disable-system-checks`, the last step of the upgrade is going to fail, and you need to call `{foreman-installer} --disable-system-checks` manually once the system reboots.
 ====
-endif::[]
--- a/guides/common/modules/proc_upgrading-el7-to-el8.adoc
+++ b/guides/common/modules/proc_upgrading-el7-to-el8.adoc
@@ -6,8 +6,6 @@ LEAPP is the supported tool for Enterprise Linux 7 to Enterprise Linux 8 upgrade
 
 .Prerequisites
 * {Project} {ProjectVersion} running on Enterprise Linux 7.
-* {Project} installations running on {RHEL} 7 cannot upgrade to {RHEL} 8.
-You must have a {RHEL} compatible LEAPP build to upgrade.
 * Access to available repositories or a local mirror of repositories.
 * No outstanding updates.
 * {Project} installations can be upgraded to CentOS Stream 8 and a {RHEL} rebuild.
@@ -37,8 +35,8 @@ You must have a {RHEL} compatible LEAPP build to upgrade.
 +
 ----
 [leapp-foreman]
-name={ProjectVersion}
-baseurl=https://yum.theforeman.org/releases/3.2/el8/$basearch
+name=Foreman {ProjectVersion}
+baseurl=https://yum.theforeman.org/releases/{ProjectVersion}/3.2/el8/$basearch
 gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman
 enabled=1
 gpgcheck=1
@@ -46,16 +44,16 @@ module_hotfixes=1
 
 ifdef::katello[]
 [leapp-katello]
-name={KatelloVersion}
-baseurl=https://yum.theforeman.org/katello/4.4/katello/el8/$basearch/
+name=Katello {KatelloVersion}
+baseurl=https://yum.theforeman.org/katello/{KatelloVersion}/katello/el8/$basearch/
 gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman
 enabled=1
 gpgcheck=1
 module_hotfixes=1
 
 [leapp-katello-candlepin]
-name={KatelloVersion}
-baseurl=https://yum.theforeman.org/katello/4.4/candlepin/el8/$basearch/
+name=Candlepin: an open source entitlement management system.
+baseurl=https://yum.theforeman.org/katello/{KatelloVersion}/candlepin/el8/$basearch/
 gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman
 enabled=1
 gpgcheck=1
@@ -71,16 +69,16 @@ module_hotfixes=1
 endif::[]
 
 [leapp-foreman-plugins]
-name={ProjectVersion}
-baseurl=https://yum.theforeman.org/plugins/3.2/el8/$basearch
+name=Foreman plugins {ProjectVersion}
+baseurl=https://yum.theforeman.org/plugins/{ProjectVersion}/el8/$basearch
 enabled=1
 gpgcheck=0
 gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman
 module_hotfixes=1
 
 [leapp-foreman-client]
-name=Foreman client 3.2
-baseurl=https://yum.theforeman.org/client/3.2/el8/$basearch
+name=Foreman client {ProjectVersion}
+baseurl=https://yum.theforeman.org/client/{ProjectVersion}/el8/$basearch
 enabled=1
 gpgcheck=1
 gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman-client
@@ -130,24 +128,18 @@ Continue to the next step for remediation.
 +
 The `preupgrade` might fail with a dependency resolution error such as:
 +
-----
+
 * "package rubygem-fx-0.5.0-2.el8.noarch requires rubygem(railties) >= 4.0.0, but none of the providers can be installed"
 * "package rubygem-railties-6.0.4.7-1.el8.noarch requires rubygem(thor) < 2.0, but none of the providers can be installed"
-----
+
 +
 If this happens, do the following to clean up packages that cannot automatically upgrade (`rubygem(thor)` and `rubygem(railties)` in the example above):
+
 +
 ----
 # yum remove rubygem-thor rubygem-railties
 ----
-+
-Check the output if it does not contain any local packages.
-+
-Remove `rubygem-thor`:
-+
-----
-# yum remove rubygem-thor
-----
+
 . Ensure `leapp preupgrade` has no issues.
 
 . Run:
@@ -158,7 +150,7 @@ Remove `rubygem-thor`:
 . Reboot the system.
 
 . After the system reboots, a live system conducts the upgrade, reboots to fix SELinux labels, then reboots into the final Enterprise Linux 8 system.
-. After the system reboots, LEAPP finishes the upgrade, watch it with:
+. LEAPP finishes the upgrade, watch it with:
 +
 ----
 # journalctl -u leapp_resume -f
--- a/guides/common/modules/proc_upgrading-el7-to-el8.adoc
+++ b/guides/common/modules/proc_upgrading-el7-to-el8.adoc
@@ -9,6 +9,7 @@ LEAPP is the supported tool for Enterprise Linux 7 to Enterprise Linux 8 upgrade
 * Access to available repositories or a local mirror of repositories.
 * No outstanding updates.
 * {Project} installations running on CentOS 7 can be upgraded to CentOS Stream 8 or a {RHEL} rebuild.
+* {Project} installations running on {RHEL} 7 can be upgraded to {RHEL} 8.
 
 .Procedure
 . Configure the repositories to obtain LEAPP.
@@ -103,7 +104,6 @@ gpgcheck=1
 * If you are using Puppet 6 instead of Puppet 7, replace the `7` with a `6` in the `leapp-puppet7` entry.
 
 * You need a Puppet repository for the Puppet agent that the installer is using.
-For more information, see link:https://docs.theforeman.org/nightly/Managing_Configurations_Puppet/index-katello.html[Configuring Hosts Using Puppet].
 
 . We do not support Enterprise Linux 8 installations with EPEL 8 enabled, so remove `epel-release`:
 +
--- a/guides/common/modules/proc_upgrading-el7-to-el8.adoc
+++ b/guides/common/modules/proc_upgrading-el7-to-el8.adoc
@@ -5,14 +5,21 @@ Use this procedure to upgrade your {Project} installation from {RHEL} 7 to {RHEL
 LEAPP is the supported tool for {RHEL} 7 to {RHEL} 8 upgrades.
 
 .Prerequisites
-* {Project} {ProjectVersionPrevious} running on CentOS 7.
-- Upgrading does not work on {RHEL}. You must have a {RHEL}-compatible LEAPP build to upgrade.
-* Access to a local repository server mirroring all required repositories.
+* {Project} {ProjectVersionPrevious} running on CentOS Linux 7.
+ifdef::katello,foreman-el[]
+- {Project} installations running on {RHEL} 7 cannot upgrade to {RHEL} 8. You must have a {RHEL} compatible LEAPP build to upgrade.
+endif::[]
+* Access to available repositories or a local mirror of repositories.
 * No outstanding updates.
-* You can upgrade a CentOS7 system before you upgrade {Project}.
+* {Project} installations can be upgraded to CentOS Stream 8 and a {RHEL} rebuild.
+ifdef::satellite[]
+* {Project} installations upgrade from {RHEL} 7 to {RHEL} 8, not from CentOS Linux 7 and {RHEL} rebuilds.
+endif::[]
 
 .Procedure
+ifdef::katello,foreman-el[]
 . Configure the https://copr.fedorainfracloud.org/coprs/g/theforeman/leapp/[@theforeman/leapp COPR Repository], which contains newer LEAPP packages than those shipped by https://wiki.almalinux.org/elevate/[AlmaLinux/ELevate], and support {Project} upgrades:
+
 +
 ----
 # curl -o /etc/yum.repos.d/theforeman-leapp.repo https://copr.fedorainfracloud.org/coprs/g/theforeman/leapp/repo/epel-7/group_theforeman-leapp-epel-7.repo
@@ -41,6 +48,13 @@ enabled=1
 gpgcheck=0
 gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman
 module_hotfixes=1
+
+[leapp-foreman-client]
+name=Foreman client 3.2
+baseurl=https://yum.theforeman.org/client/3.2/el8/$basearch
+enabled=1
+gpgcheck=1
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman-client
 
 [leapp-puppet7]
 name=Puppet 7 Repository el 8 - $basearch
@@ -74,6 +88,7 @@ enabled=1
 gpgcheck=1
 module_hotfixes=1
 ----
+endif::[]
 
 * If you are not using {Project}, you can leave out `leapp-katello`, `leapp-katello-candlepin` and `leapp-pulpcore`.
 
@@ -81,10 +96,18 @@ module_hotfixes=1
 
 * You need a Puppet repository for the Puppet agent that the installer is using.
 
-. We do not support EL8 installations with EPEL8 enabled, so remove `epel-release`:
+ifdef::katello,foreman-el[]
+. We do not support {RHEL} 8 installations with EPEL 8 enabled, so remove `epel-release`:
 +
 ----
 # yum remove epel-release
+----
+endif::[]
+
+. Remove `centos-release-scl` and `centos-release-scl-rh` repositories:
++
+----
+# yum remove centos-release-scl centos-release-scl-rh
 ----
 . Let LEAPP analyze your system:
 +
@@ -106,28 +129,33 @@ Continue to the next step for remediation.
 ----
 +
 
+ifdef::katello,foreman-el[]
 ** If `preupgrade` aborts with a dependency resolution error such as:
 +
 ----
-# package rubygem-fx-0.5.0-2.el8.noarch requires rubygem(railties) >= 4.0.0, but none of the providers can install
+# package rubygem-fx-0.5.0-2.el8.noarch requires rubygem(railties) >= 4.0.0,
 ----
-or:
+but none of the providers can install or:
 +
 ----
- # package rubygem-railties-6.0.4.7-1.el8.noarch requires rubygem(thor) < 2.0, but none of the providers can install
+# package rubygem-railties-6.0.4.7-1.el8.noarch requires rubygem(thor) < 2.0
 ----
+but none of the providers can install:
+endif::[]
 
-** Remove packages that are no longer available from any repository:
+ifdef::katello,foreman-el[]
+. Remove packages that are no longer available from any repository:
 +
 ----
 # dnf remove $(dnf repoquery --extras --exclude '*katello*' --exclude '*pulp*' --exclude '*localhost*' --exclude "*$HOSTNAME*" --exclude libmodulemd) –
 ----
+endif::[]
 
 +
 
-** Check the output if it does not contain any local packages.
+. Check the output if it does not contain any local packages.
 
-** Remove `rubygem-thor`:
+. Remove `rubygem-thor`:
 +
 ----
 # yum remove rubygem-thor
@@ -143,17 +171,14 @@ or:
 +
 
 . Reboot the system.
-. After the system reboots, a live system conducts the upgrade.
-. Reboot to fix SELinux labels.
-. Reboot into the final {RHEL} 8 system.
-. After the system reboots, LEAPP finishes the upgrade.
-. You can watch LEAPP finish the upgrade with:
+. After the system reboots, a live system conducts the upgrade, reboots to fix SELinux labels, then reboots into the final {RHEL} 8 system.
+. After the system reboots, LEAPP finishes the upgrade, watch it with:
 +
 ----
 # journalctl -u leapp_resume -f
 ----
 
-[WARNING]
+[NOTE]
 ====
-If you installed the system and had to use `--disable-system-checks`, the last step of the upgrade will fail, and you will need to call `foreman-installer --disable-system-checks` manually once the system is booted up.
+If you install the system and need to use `--disable-system-checks`, the last step of the upgrade is going to fail, and you need to call `{foreman-installer} --disable-system-checks` manually once the system reboots.
 ====
--- a/guides/common/modules/proc_upgrading-el7-to-el8.adoc
+++ b/guides/common/modules/proc_upgrading-el7-to-el8.adoc
@@ -97,8 +97,6 @@ gpgcheck=1
 * If you are using Puppet 6 instead of Puppet 7, replace the `7` with a `6` in the `leapp-puppet7` entry.
 
 * You need a Puppet repository for the Puppet agent that the installer is using.
-
-+
 For more information, see link:https://docs.theforeman.org/nightly/Managing_Configurations_Puppet/index-katello.html[Configuring Hosts Using Puppet].
 
 . We do not support Enterprise Linux 8 installations with EPEL 8 enabled, so remove `epel-release`:
@@ -131,36 +129,35 @@ Continue to the next step for remediation.
 
 +
 The `preupgrade` might fail with a dependency resolution error such as:
-
++
+----
 * "package rubygem-fx-0.5.0-2.el8.noarch requires rubygem(railties) >= 4.0.0, but none of the providers can be installed"
 * "package rubygem-railties-6.0.4.7-1.el8.noarch requires rubygem(thor) < 2.0, but none of the providers can be installed"
-
+----
++
 If this happens, do the following to clean up packages that cannot automatically upgrade (`rubygem(thor)` and `rubygem(railties)` in the example above):
++
 ----
 # yum remove rubygem-thor rubygem-railties
 ----
-
++
 Check the output if it does not contain any local packages.
-
++
 Remove `rubygem-thor`:
++
 ----
 # yum remove rubygem-thor
 ----
-[start=9]
 . Ensure `leapp preupgrade` has no issues.
-[start=10]
+
 . Run:
 +
 ----
 # leapp upgrade
 ----
-
-[start=11]
 . Reboot the system.
 
-[start=12]
 . After the system reboots, a live system conducts the upgrade, reboots to fix SELinux labels, then reboots into the final Enterprise Linux 8 system.
-[start=13]
 . After the system reboots, LEAPP finishes the upgrade, watch it with:
 +
 ----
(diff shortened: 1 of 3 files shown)


melcorr:
The attribute for RHEL is {RHEL}

bangelic:
@melcorr Updated with {RHEL} attribute.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,159 @@
+[id="Upgrading-Project-from-EL7-to-EL8{context}"]
+= Upgrading {Project} from {RHEL} 7 to {RHEL} 8
+
+Use this procedure to upgrade your {Project} installation from {RHEL} 7 to {RHEL} 8.
+LEAPP is the supported tool for {RHEL} 7 to {RHEL} 8 upgrades.
+
+.Prerequisites
+* {Project} {ProjectVersionPrevious} running on CentOS 7.
+- Upgrading does not work on {RHEL}. You must have a {RHEL}-compatible LEAPP build to upgrade.
```
melcorr, line 9: I do not understand the use of a dash here. 
We have a strict one sentence per line rule and have always had. Please adhere everywhere and always.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,159 @@
...
+* Access to a local repository server mirroring all required repositories.
+* No outstanding updates.
+* You can upgrade a CentOS7 system before you upgrade {Project}.
```
melcorr, line 12: Please follow conventions when referring to an operating system. This is not how it is written.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,159 @@
...
+
+** If `preupgrade` aborts with a dependency resolution error such as:
++
+----
+# package rubygem-fx-0.5.0-2.el8.noarch requires rubygem(railties) >= 4.0.0, but none of the providers can install
+----
+or:
++
+----
+ # package rubygem-railties-6.0.4.7-1.el8.noarch requires rubygem(thor) < 2.0, but none of the providers can install
```
melcorr, line 117: why is there text in the code block?

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,159 @@
...
+----
+
+** Remove packages that are no longer available from any repository:
```
melcorr, line 120: I'm pretty sure that an ordered list would be better if these are a sequence of steps.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,159 @@
...
++
+----
+# dnf remove $(dnf repoquery --extras --exclude '*katello*' --exclude '*pulp*' --exclude '*localhost*' --exclude "*$HOSTNAME*" --exclude libmodulemd) –
+----
+
++
```
melcorr, line 126: What is this for?

guides/doc-Upgrading_and_Updating/master.adoc
```diff
@@ -56,6 +56,11 @@ include::topics/upgrading_discovery_satellite.adoc[leveloffset=+3]
 // Configuring Subnets with a Template Capsule
 include::topics/upgrading_discovery_subnet_with_a_template_capsule.adoc[leveloffset=+3]
 
+ifdef::katello,satellite[]
```
melcorr, line 59: @evgeni 
would you mind advising which parts are not applicable for Satellite? in the procedure in general?

melcorr (changes requested):
Some comments.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,159 @@
+[id="Upgrading-Project-from-EL7-to-EL8{context}"]
+= Upgrading {Project} from {RHEL} 7 to {RHEL} 8
+
+Use this procedure to upgrade your {Project} installation from {RHEL} 7 to {RHEL} 8.
+LEAPP is the supported tool for {RHEL} 7 to {RHEL} 8 upgrades.
```
evgeni, line 5: Here you say upgrading from RHEL7 to RHEL8, later you write it needs to run on CentOS 7 and upgrades on RHEL are not possible. This will confuse all readers :)

Just to make it clear:
* Foreman and Katello installations running on CentOS 7 can be upgraded to CentOS Stream 8 (and probably any other EL8-rebuild like Rocky and Alma, even tho I didn't test that explicitly).
* Foreman and Katello installations running on RHEL7 can't be upgraded (right now).
* Satellite installations will be upgradeable from RHEL7 to RHEL8, no CentOS/EL-rebuilds.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,159 @@
...
+
+.Prerequisites
+* {Project} {ProjectVersionPrevious} running on CentOS 7.
+- Upgrading does not work on {RHEL}. You must have a {RHEL}-compatible LEAPP build to upgrade.
+* Access to a local repository server mirroring all required repositories.
```
evgeni, line 10: No. The system needs access to the repositories we offer, *OR* a local mirror.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,159 @@
...
+----
+# curl -o /etc/yum.repos.d/theforeman-leapp.repo https://copr.fedorainfracloud.org/coprs/g/theforeman/leapp/repo/epel-7/group_theforeman-leapp-epel-7.repo
+----
+. Install required packages:
++
+----
+# yum install leapp leapp-repository leapp-data-centos
+----
+
+. Add {Project} specific repositories to `/etc/leapp/files/leapp_upgrade_repositories.repo`:
```
evgeni, line 26: please have a look at https://github.com/theforeman/theforeman.org/pull/1977, I've updated this section with the "client" repo, which should be reflected here

guides/common/modules/proc_upgrading-el7-to-el8.adoc
evgeni, line 117: a `#` usually indicates a command, but this is just a text block

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,159 @@
...
+----
+# leapp upgrade
+----
+
++
+
+. Reboot the system.
+. After the system reboots, a live system conducts the upgrade.
+. Reboot to fix SELinux labels.
+. Reboot into the final {RHEL} 8 system.
```
evgeni, line 148: these steps happen automaticaly, the user doesn't need to do anything

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,159 @@
...
+. After the system reboots, LEAPP finishes the upgrade.
+. You can watch LEAPP finish the upgrade with:
```
evgeni, line 150: this looks like two steps, but the "watch" part is more of a comment to the "leapp does finish"

guides/doc-Upgrading_and_Updating/master.adoc
evgeni, line 59: this applies to foreman on EL too, so should be `ifdef::foreman-el,katello,satellite` I think

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,159 @@
...
+
+.Prerequisites
+* {Project} {ProjectVersionPrevious} running on CentOS 7.
+- Upgrading does not work on {RHEL}. You must have a {RHEL}-compatible LEAPP build to upgrade.
+* Access to a local repository server mirroring all required repositories.
+* No outstanding updates.
+* You can upgrade a CentOS7 system before you upgrade {Project}.
+
+.Procedure
+. Configure the https://copr.fedorainfracloud.org/coprs/g/theforeman/leapp/[@theforeman/leapp COPR Repository], which contains newer LEAPP packages than those shipped by https://wiki.almalinux.org/elevate/[AlmaLinux/ELevate], and support {Project} upgrades:
```
evgeni, line 15: this step is not required for satellite

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,159 @@
...
++
+----
+# curl -o /etc/yum.repos.d/theforeman-leapp.repo https://copr.fedorainfracloud.org/coprs/g/theforeman/leapp/repo/epel-7/group_theforeman-leapp-epel-7.repo
+----
+. Install required packages:
```
evgeni, line 20: for satellite, the `leapp-data-centos` package is not required

guides/common/modules/proc_upgrading-el7-to-el8.adoc
evgeni, line 26: this step is not required for satellite

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,159 @@
...
+* If you are not using {Project}, you can leave out `leapp-katello`, `leapp-katello-candlepin` and `leapp-pulpcore`.
+
+* If you are using Puppet 6 instead of Puppet 7, replace the `7` with a `6` in the `leapp-puppet7` entry.
+
+* You need a Puppet repository for the Puppet agent that the installer is using.
+
+. We do not support EL8 installations with EPEL8 enabled, so remove `epel-release`:
++
+----
+# yum remove epel-release
```
evgeni, line 87: this step is not required for satellite

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,159 @@
...
+. Fix issues found by LEAPP (check `/var/log/leapp/leapp-report.txt` for details):
++
+----
+# rmmod pata_acpi
+# echo PermitRootLogin yes | tee -a /etc/ssh/sshd_config
+# leapp answer --section remove_pam_pkcs11_module_check.confirm=True
+----
++
+
+** If `preupgrade` aborts with a dependency resolution error such as:
```
evgeni, line 109: this step shouldn't be required with satellite

guides/common/modules/proc_upgrading-el7-to-el8.adoc
evgeni, line 120: this step is not required with satellite

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,159 @@
...
+. After the system reboots, LEAPP finishes the upgrade.
+. You can watch LEAPP finish the upgrade with:
++
+----
+# journalctl -u leapp_resume -f
+----
+
+[WARNING]
+====
+If you installed the system and had to use `--disable-system-checks`, the last step of the upgrade will fail, and you will need to call `foreman-installer --disable-system-checks` manually once the system is booted up.
```
evgeni, line 158: this should use the `{foreman-installer}` macro

melcorr: Also, we don't use language like `booted up` in official docs. 
IBM Style Guide has specifications for when and if you ever use the word "once". 
We do  not use the future tense in our docs. I've mentioned this several times by now. 
We do not use warnings in our docs unless the action with irreparably damage the deployment or cause harm to human life. This does not seem to be the case here.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
bangelic: It appears in the code block in the document that was written to be converted to Asciidoc.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
bangelic: The order of the procedure was steps 1-12 and I had to add the + to keep the continuing list from restarting back at 1 in the middle of the steps.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
bangelic: Some of the other writers mentioned that they have used CentOS7 when documenting. What is the proper naming convention for an operating system?

guides/common/modules/proc_upgrading-el7-to-el8.adoc
melcorr: I think that @evgeni meant this as a comment

guides/common/modules/proc_upgrading-el7-to-el8.adoc
melcorr: If you google and find the CentOS 7 documentation or website, you'll find how to write it.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
bangelic: Understood @melcorr T

There are other instances in the existing documentation where **NOTE** and **WARNING** have future tense in the text so I wasn't sure if that was the only time it was allowed.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
bangelic: Thanks!

evgeni: I meant this as "you could see errors like … or …" and didn't know how to format it better as "this is the error output"

guides/common/modules/proc_upgrading-el7-to-el8.adoc
melcorr: Brian, I am sorry. I think that I came across quite short here in retrospect and it wasn't my intention. All I meant is that there are certain ways that all operating systems, projects, third party infrastructure etc can and should be referred to specifically and it's easy enough to find them. Their project or product websites are the authority on how it should be spelled rather than myself or any writer.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,184 @@
+[id="Upgrading-Project-from-EL7-to-EL8{context}"]
+= Upgrading {Project} from {RHEL} 7 to {RHEL} 8
```
evgeni, line 2: For Foreman, this renders as "Upgrading Foreman from Red Hat Enterprise Linux 7 to Red Hat Enterprise Linux 8", which is exactly the combination we do not support today.

Can we make it render "Upgrading Satellite from Red Hat Enterprise Linux 7 to Red Hat Enterprise Linux 8", but for the others it'd just be "Enterprise Linux" without the "Red Hat", as that would include CentOS and friends.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+Use this procedure to upgrade your {Project} installation from {RHEL} 7 to {RHEL} 8.
+LEAPP is the supported tool for {RHEL} 7 to {RHEL} 8 upgrades.
+
+.Prerequisites
+* {Project} {ProjectVersionPrevious} running on CentOS Linux 7.
```
evgeni, line 8: This won't be true for Satellite, which runs on RHEL 7 only.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,184 @@
...
+. Install required packages:
++
+----
+# yum install leapp leapp-repository leapp-data-centos
+----
+
+. Add {Project} specific repositories to `/etc/leapp/files/leapp_upgrade_repositories.repo`:
++
+----
+# [leapp-foreman]
```
evgeni, line 36:
```suggestion
[leapp-foreman]
```

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+[leapp-puppet7]
+name=Puppet 7 Repository el 8 - $basearch
+baseurl=http://yum.puppetlabs.com/puppet7/el/8/$basearch
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-puppet7-release
+       file:///etc/pki/rpm-gpg/RPM-GPG-KEY-2025-04-06-puppet7-release
+enabled=1
+gpgcheck=1
+
+[leapp-katello]
```
evgeni, line 67: can we use `ifdef::katello` here inside the code block? so we hide the katello repos in the foreman docs?

allows to drop the silly "If you are not using " comment below

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+[leapp-pulpcore]
+name=pulpcore: Fetch, Upload, Organize, and Distribute Software Packages.
+baseurl=https://yum.theforeman.org/pulpcore/3.16/el8/$basearch/
+gpgkey=https://yum.theforeman.org/pulpcore/3.16/GPG-RPM-KEY-pulpcore
+enabled=1
+gpgcheck=1
+module_hotfixes=1
+----
+endif::[]
```
evgeni, line 91: why does this endif here? the points below still apply to katello/foreman only

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+ifdef::katello,foreman-el[]
+. We do not support {RHEL} 8 installations with EPEL 8 enabled, so remove `epel-release`:
++
+----
+# yum remove epel-release
+----
+endif::[]
+
+. Remove `centos-release-scl` and `centos-release-scl-rh` repositories:
```
evgeni, line 107: this should also be inside `ifdef::katello,foreman-el`

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,184 @@
...
+----
++
+
+ifdef::katello,foreman-el[]
+** If `preupgrade` aborts with a dependency resolution error such as:
++
+----
+# package rubygem-fx-0.5.0-2.el8.noarch requires rubygem(railties) >= 4.0.0,
+----
+but none of the providers can install or:
```
evgeni, line 138: the errors we want to highlight are  "package rubygem-fx-0.5.0-2.el8.noarch requires rubygem(railties) >= 4.0.0, but none of the providers providers can be installed" and "package rubygem-railties-6.0.4.7-1.el8.noarch requires rubygem(thor) < 2.0, but none of the providers providers can be installed"

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,184 @@
...
++
+----
+# package rubygem-railties-6.0.4.7-1.el8.noarch requires rubygem(thor) < 2.0
+----
+but none of the providers can install:
+endif::[]
+
+ifdef::katello,foreman-el[]
+. Remove packages that are no longer available from any repository:
```
evgeni, line 147: this needs to be a sub-item of the "dependency resolution error" step

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,184 @@
...
++
+----
+# dnf remove $(dnf repoquery --extras --exclude '*katello*' --exclude '*pulp*' --exclude '*localhost*' --exclude "*$HOSTNAME*" --exclude libmodulemd) –
+----
+endif::[]
+
++
+
+. Check the output if it does not contain any local packages.
```
evgeni, line 156: this needs to be a sub-item of the "dependency resolution error" step

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,184 @@
...
++
+----
+# rmmod pata_acpi
+# echo PermitRootLogin yes | tee -a /etc/ssh/sshd_config
+# leapp answer --section remove_pam_pkcs11_module_check.confirm=True
+----
++
+
+ifdef::katello,foreman-el[]
+** If `preupgrade` aborts with a dependency resolution error such as:
```
evgeni, line 133: this whole step should look like this (step wise, not text wise):


`preupgrade` might fail with a dependency resolution error such as:

* "package rubygem-fx-0.5.0-2.el8.noarch requires rubygem(railties) >= 4.0.0, but none of the providers providers can be installed"
* "package rubygem-railties-6.0.4.7-1.el8.noarch requires rubygem(thor) < 2.0, but none of the providers providers can be installed"

if this happens, you need to clean up packages that can't be automatically upgraded, please try:
*  `dnf remove $(dnf repoquery --extras --exclude '*katello*' --exclude '*pulp*' --exclude '*localhost*' --exclude "*$HOSTNAME*" --exclude libmodulemd)`
    *  Check the output if it does not contain any local packages.
* remove `rubygem-thor`

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,184 @@
...
+# package rubygem-railties-6.0.4.7-1.el8.noarch requires rubygem(thor) < 2.0
+----
+but none of the providers can install:
+endif::[]
+
+ifdef::katello,foreman-el[]
+. Remove packages that are no longer available from any repository:
++
+----
+# dnf remove $(dnf repoquery --extras --exclude '*katello*' --exclude '*pulp*' --exclude '*localhost*' --exclude "*$HOSTNAME*" --exclude libmodulemd) –
```
evgeni, line 150:
```suggestion
# dnf remove $(dnf repoquery --extras --exclude '*katello*' --exclude '*pulp*' --exclude '*localhost*' --exclude "*$HOSTNAME*" --exclude libmodulemd)
```

guides/common/modules/proc_upgrading-el7-to-el8.adoc
evgeni: I used the following, and it worked:

```patch
diff --git guides/common/modules/proc_upgrading-el7-to-el8.adoc guides/common/modules/proc_upgrading-el7-to-el8.adoc
index ab89ff26..541e4aec 100644
--- guides/common/modules/proc_upgrading-el7-to-el8.adoc
+++ guides/common/modules/proc_upgrading-el7-to-el8.adoc
@@ -33,7 +33,7 @@ ifdef::katello,foreman-el[]
 . Add {Project} specific repositories to `/etc/leapp/files/leapp_upgrade_repositories.repo`:
 +
 ----
-# [leapp-foreman]
+[leapp-foreman]
 name=Foreman 3.2
 baseurl=https://yum.theforeman.org/releases/3.2/el8/$basearch
 enabled=1
@@ -64,6 +64,7 @@ gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-puppet7-release
 enabled=1
 gpgcheck=1
 
+ifdef::katello[]
 [leapp-katello]
 name=Katello 4.4
 baseurl=https://yum.theforeman.org/katello/4.4/katello/el8/$basearch/
@@ -87,11 +88,10 @@ gpgkey=https://yum.theforeman.org/pulpcore/3.16/GPG-RPM-KEY-pulpcore
 enabled=1
 gpgcheck=1
 module_hotfixes=1
+endif::[]
 ----
 endif::[]
 
-* If you are not using {Project}, you can leave out `leapp-katello`, `leapp-katello-candlepin` and `leapp-pulpcore`.
-
 * If you are using Puppet 6 instead of Puppet 7, replace the `7` with a `6` in the `leapp-puppet7` entry.
 
 * You need a Puppet repository for the Puppet agent that the installer is using.
```

bangelic:
@evgeni Addressed your latest comments. Let me know if this is better.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,203 @@
...
+LEAPP is the supported tool for {RHEL} 7 to {RHEL} 8 upgrades.
+
+.Prerequisites
+ifdef::katello,foreman-el[]
+* {Project} {ProjectVersionPrevious} running on CentOS Linux 7.
+endif::[]
+
+ifdef::satellite[]
+- {Project} installations running on {RHEL} 7 cannot upgrade to {RHEL} 8. You must have a {RHEL} compatible LEAPP build to upgrade.
+endif::[]
```
evgeni, line 19: This is only relevant for Foreman/Katello.
Satellite only runs on RHEL and RHEL users have the right LEAPP builds.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,203 @@
...
+
+* Access to available repositories or a local mirror of repositories.
+* No outstanding updates.
+ifdef::katello,foreman-el[]
+* {Project} installations can be upgraded to CentOS Stream 8 and a {RHEL} rebuild.
+endif::[]
+
+ifdef::satellite[]
+* {Project} installations upgrade from {RHEL} 7 to {RHEL} 8, not from CentOS Linux 7 and {RHEL} rebuilds.
+endif::[]
```
evgeni, line 29: Satellite is not supported on non-RHEL, so this can go away completely.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,203 @@
...
+gpgcheck=1
+module_hotfixes=1
+
+[leapp-pulpcore]
+name=pulpcore: Fetch, Upload, Organize, and Distribute Software Packages.
+baseurl=https://yum.theforeman.org/pulpcore/3.16/el8/$basearch/
+gpgkey=https://yum.theforeman.org/pulpcore/3.16/GPG-RPM-KEY-pulpcore
+enabled=1
+gpgcheck=1
+module_hotfixes=1
```
evgeni, line 109: these two should also go into the `ifdef::katello` section above.

bangelic:
@evgeni Addressed your latest 3 suggested changes. Is this good to be merged?

guides/doc-Upgrading_and_Updating/master.adoc
```diff
@@ -56,6 +56,11 @@ include::topics/upgrading_discovery_satellite.adoc[leveloffset=+3]
 // Configuring Subnets with a Template Capsule
 include::topics/upgrading_discovery_subnet_with_a_template_capsule.adoc[leveloffset=+3]
 
+ifdef::foreman-el,katello,satellite[]
+// Upgrading Foreman/Katello from EL7 to EL8
+include::common/modules/proc_upgrading-el7-to-el8.adoc[leveloffset=+2]
```
ehelms, line 61: Only thought I had is that given this is a big deal to upgrade the OS and that it is required to be able to update to the following version of Foreman/katello/Satellite if it should have its own top level section header to ensure it stands out. 

Right now we have:

```
[1. Upgrade Overview]
[2. Upgrading Foreman]
[3. Updating Foreman server and Content Hosts]
```

Wondering if we need:

```
[1. Upgrade Overview]
[2. Upgrading Foreman]
[3. Upgrading Foreman from Enterprise Linux 7 to Enterprise Linux 8]
[4. Updating Foreman server and Content Hosts]
```

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,195 @@
...
+----
+endif::[]
+
++
+----
```
evgeni, line 79: this should be in one codeblock with the other repos in the lines above

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,195 @@
...
+
+[start=8]
+. Check the output if it does not contain any local packages.
++
+. Remove `rubygem-thor`:
++
+----
+# yum remove rubygem-thor
+----
```
evgeni, line 160: this is not an own step, it still belongs to "The `preupgrade` might fail with a dependency resolution error such as"

ehelms:
I am totally good with doing small updates and tackling areas. Two I wanted to ask about:

 * Will we do a separate guide for Smart Proxy / Capsule upgrades?
 * Will we do a separate guide / section for disconnected to call out the differences there?

guides/common/modules/proc_upgrading-el7-to-el8.adoc
bangelic: @evgeni What specifically should be in one code block? The + sign?

evgeni: All entries for the repositories should go into one

bangelic:
Thanks @evgeni. Pushed those changes through. Take a look and let me know if this can be merged.

evgeni:
I won't have time before Tuesday to further review this.

@ehelms @melcorr feel free to merge if you think it's good

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,190 @@
+[id="Upgrading-Project-from-EL7-to-EL8{context}"]
```
maximiliankolb, line 1:
```suggestion
[id="Upgrading-Project-from-EL7-to-EL8_{context}"]
```

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,190 @@
...
+= Upgrading {Project} from Enterprise Linux 7 to Enterprise Linux 8
+endif::[]
+
+Use this procedure to upgrade your {Project} installation from {RHEL} 7 to {RHEL} 8.
+LEAPP is the supported tool for {RHEL} 7 to {RHEL} 8 upgrades.
+
+.Prerequisites
+ifdef::katello,foreman-el[]
+* {Project} {ProjectVersionPrevious} running on CentOS Linux 7.
+- {Project} installations running on {RHEL} 7 cannot upgrade to {RHEL} 8. You must have a {RHEL} compatible LEAPP build to upgrade.
```
maximiliankolb, line 15:
```suggestion
* {Project} installations running on {RHEL} 7 cannot upgrade to {RHEL} 8.
You must have a {RHEL} compatible LEAPP build to upgrade.
```

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,190 @@
...
+* Access to available repositories or a local mirror of repositories.
+* No outstanding updates.
+ifdef::katello,foreman-el[]
+* {Project} installations can be upgraded to CentOS Stream 8 and a {RHEL} rebuild.
+endif::[]
+
+.Procedure
+ifdef::katello,foreman-el[]
+. Configure the https://copr.fedorainfracloud.org/coprs/g/theforeman/leapp/[@theforeman/leapp COPR Repository], which contains newer LEAPP packages than those shipped by https://wiki.almalinux.org/elevate/[AlmaLinux/ELevate], and support {Project} upgrades:
+
```
maximiliankolb, line 27:
```suggestion
```

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,190 @@
...
+ifdef::katello,foreman-el[]
+. Add {Project} specific repositories to `/etc/leapp/files/leapp_upgrade_repositories.repo`:
++
+----
+[leapp-foreman]
+name=Foreman 3.2
+baseurl=https://yum.theforeman.org/releases/3.2/el8/$basearch
+enabled=1
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-puppet7-release
```
maximiliankolb, line 46: This sounds off to me. Is is meant to be the same GPG as below?

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,190 @@
...
+* You need a Puppet repository for the Puppet agent that the installer is using.
+endif::[]
+
+ifdef::katello,foreman-el[]
+. We do not support Enterprise Linux 8 installations with EPEL 8 enabled, so remove `epel-release`:
++
+----
+# yum remove epel-release
+----
+
```
maximiliankolb, line 112:
```suggestion
```

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,190 @@
...
++
+----
+# yum remove centos-release-scl centos-release-scl-rh
+----
+. Let LEAPP analyze your system:
++
+----
+# leapp preupgrade
+----
+
```
maximiliankolb, line 123:
```suggestion
+
```

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,190 @@
...
+The first run is expected to fail but report issues.
+Continue to the next step for remediation.
+
+[start=7]
```
maximiliankolb, line 127:
```suggestion
```

If you apply my suggestion in line 123, then defining the start should not be necessary.

maximiliankolb (commented):
Some small comments.

If there is anything mentioned in the review of your PR that you could not yet fix/add, please open an issue within github and reference the comment. Thanks!

melcorr:
Hey @bangelic this PR now has build errors since the last push. 
Can you please take a look and resolve any errors.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,221 @@
+[id="Upgrading-Project-from-EL7-to-EL8_{context}"]
+ifdef::satellite[]
+= Upgrading {Project} from {RHEL} 7 to {RHEL} 8
+endif::[]
+ifdef::katello,foreman-el[]
```
Lennonka, line 5: @maximiliankolb Would it be better to use `ifndef:satellite[]` instead?

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,221 @@
...
+.Prerequisites
+* Access to available repositories or a local mirror of repositories.
+* No outstanding updates.
+
+.Procedure
+[start=1]
+. Ensure `leapp preupgrade` has no issues.
+[start=2]
+. Run:
+----
```
Lennonka, line 40: Insert a line with `+` between the list item and the code.

Lennonka: Also all other occurences.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,221 @@
...
+ifdef::katello,foreman-el[]
+** The `preupgrade` might fail with a dependency resolution error such as:
+
+* "package rubygem-fx-0.5.0-2.el8.noarch requires rubygem(railties) >= 4.0.0, but none of the providers can be installed"
+* "package rubygem-railties-6.0.4.7-1.el8.noarch requires rubygem(thor) < 2.0, but none of the providers can be installed"
+
+If this happens, do the following to clean up packages that cannot automatically upgrade:
+----
+# dnf remove $(dnf repoquery --extras --exclude '*katello*' --exclude '*pulp*' --exclude '*localhost*' --exclude "*$HOSTNAME*" --exclude libmodulemd)
+----
```
Lennonka, line 176:
```
+
The `preupgrade` might fail with a dependency resolution error such as:

* "package rubygem-fx-0.5.0-2.el8.noarch requires rubygem(railties) >= 4.0.0, but none of the providers can be installed"
* "package rubygem-railties-6.0.4.7-1.el8.noarch requires rubygem(thor) < 2.0, but none of the providers can be installed"

+
If this happens, do the following to clean up packages that cannot automatically upgrade:
+
----
# dnf remove $(dnf repoquery --extras --exclude '*katello*' --exclude '*pulp*' --exclude '*localhost*' --exclude "*$HOSTNAME*" --exclude libmodulemd)
----
```

Lennonka (changes requested):
Tips to fix the syntax

guides/common/modules/proc_upgrading-el7-to-el8.adoc
bangelic: @Lennonka I did a build preview with the `ifndef::satellite[]`and got this
image
Looks like 2.5 is what would be meant for Satellite but after seeing that there is a ticket for **Upgrading Satellite from RHEL 7 to RHEL 8 with LEAPP** with accompanying documentation, can we make it so that none of this content including anything with `ifdef::{}`gets built into Satellite documentation? It just needs to be for Foreman/Katello.

maximiliankolb: Yes, please apply Lena's suggestion. Something like:

```
if satellite:
  heading 1
if not satellite
  heading 2
```
Neither build targets should display empty chapters or redundant headings.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
Lennonka: @bangelic Can you verify this with a SME?

Lennonka (commented):
Okay, so the build looks good now.
I've found mentions of connected and diconnected Server -- it could use a better structure, but that's for another PR.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,233 @@
...
+# echo PermitRootLogin yes | tee -a /etc/ssh/sshd_config
+# leapp answer --section remove_pam_pkcs11_module_check.confirm=True
+----
+ifdef::katello,foreman-el[]
+The `preupgrade` might fail with a dependency resolution error such as:
+
+* "package rubygem-fx-0.5.0-2.el8.noarch requires rubygem(railties) >= 4.0.0, but none of the providers can be installed"
+* "package rubygem-railties-6.0.4.7-1.el8.noarch requires rubygem(thor) < 2.0, but none of the providers can be installed"
+
+If this happens, do the following to clean up packages that cannot automatically upgrade:
```
Lennonka, line 182: Add a line with a `+` before this so that it is inside the list item.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
evgeni: I can verify that this is completely wrong and should be `gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman`

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,233 @@
...
+gpgcheck=1
+module_hotfixes=1
+
+[leapp-pulpcore]
+name=pulpcore: Fetch, Upload, Organize, and Distribute Software Packages.
+baseurl=https://yum.theforeman.org/pulpcore/3.16/el8/$basearch/
+gpgkey=https://yum.theforeman.org/pulpcore/3.16/GPG-RPM-KEY-pulpcore
+enabled=1
+gpgcheck=1
+module_hotfixes=1
```
evgeni, line 113: the `ifdef::katello[]` needs to end here.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
bangelic: @evgeni That's the verification needed haha. Thank you!

bangelic:
@evgeni @maximiliankolb Let me know if this can be merged. I cleaned up the conditional statements and updated the master.adoc. @Lennonka helped me with build issues that I was having and now that the conditional statements have been cleaned up, this should be good to go unless there is more content needed or needs to be updated. Thank you.

guides/doc-Upgrading_and_Updating/master.adoc
Lennonka: @bangelic I will review after you address and resolve this comment.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,171 @@
...
+baseurl=http://yum.puppetlabs.com/puppet7/el/8/$basearch
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-puppet7-release
+       file:///etc/pki/rpm-gpg/RPM-GPG-KEY-2025-04-06-puppet7-release
+enabled=1
+gpgcheck=1
+----
+
+* If you are using Puppet 6 instead of Puppet 7, replace the `7` with a `6` in the `leapp-puppet7` entry.
+
+* You need a Puppet repository for the Puppet agent that the installer is using.
```
Lennonka, line 94: I'd add a link to the Puppet guide here.
Note that you need 2 links - sat and non-sat, and condition them.

bangelic: @Lennonka Since this section is specific for Foreman/Katello and not Satellite, I will use the link for non Satellite.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,173 @@
+[id="Upgrading-Project-from-EL7-to-EL8_{context}"]
+= Upgrading {Project} from Enterprise Linux 7 to Enterprise Linux 8
+
+Use this procedure to upgrade your {Project} installation from Enterprise Linux 7 to Enterprise Linux 8.
+LEAPP is the supported tool for Enterprise Linux 7 to Enterprise Linux 8 upgrades.
+
+.Prerequisites
+* {Project} {ProjectVersionPrevious} running on CentOS Linux 7.
```
maximiliankolb, line 8: Maybe it's best to say "Enterprise Linux 7" again?

If this procedure is used for both Foreman and Satellite, then maybe your can also use an attribute to differentiate between RHEL and Enterprise Linux 7. Alternatively, we can discuss setting `RHEL` to CentOS for Foreman build targets?

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,173 @@
...
+. Add {Project} specific repositories to `/etc/leapp/files/leapp_upgrade_repositories.repo`:
++
+----
+[leapp-foreman]
+name=Foreman 3.2
+baseurl=https://yum.theforeman.org/releases/3.2/el8/$basearch
+enabled=1
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman
+enabled=1
+gpgcheck=1
```
evgeni, line 40:
```suggestion
gpgcheck=1
module_hotfixes=1
```

This line was also lost somehow in all the rebases

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,173 @@
...
+----
+
+[start=3]
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+. Add {Project} specific repositories to `/etc/leapp/files/leapp_upgrade_repositories.repo`:
++
+----
+[leapp-foreman]
+name=Foreman 3.2
+baseurl=https://yum.theforeman.org/releases/3.2/el8/$basearch
```
evgeni, line 36: This document will work for Foreman 3.2 and 3.3, so I think we should use `{ProjectVersion}` here.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,173 @@
...
+enabled=1
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman
+enabled=1
+gpgcheck=1
+
+ifdef::katello[]
+[leapp-katello]
+name=Katello 4.4
+baseurl=https://yum.theforeman.org/katello/4.4/katello/el8/$basearch/
```
evgeni, line 45: This document will work for Katello 4.4 and 4.5, so I think we should use `{KatelloVersion}` here.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,173 @@
...
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman
+enabled=1
+gpgcheck=1
+module_hotfixes=1
+
+[leapp-katello-candlepin]
+name=Candlepin: an open source entitlement management system.
+baseurl=https://yum.theforeman.org/katello/4.4/candlepin/el8/$basearch/
```
evgeni, line 53: `{KatelloVersion}`

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,173 @@
...
+baseurl=https://yum.theforeman.org/pulpcore/3.16/el8/$basearch/
+gpgkey=https://yum.theforeman.org/pulpcore/3.16/GPG-RPM-KEY-pulpcore
+enabled=1
+gpgcheck=1
+module_hotfixes=1
+endif::[]
+
+[leapp-foreman-plugins]
+name=Foreman plugins 3.2
+baseurl=https://yum.theforeman.org/plugins/3.2/el8/$basearch
```
evgeni, line 70: `{ProjectVersion}`

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,173 @@
...
+enabled=1
+gpgcheck=0
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman
+module_hotfixes=1
+
+[leapp-foreman-client]
+name=Foreman client 3.2
+baseurl=https://yum.theforeman.org/client/3.2/el8/$basearch
```
evgeni, line 78: `{ProjectVersion}`

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,173 @@
...
+baseurl=https://yum.theforeman.org/katello/4.4/candlepin/el8/$basearch/
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman
+enabled=1
+gpgcheck=1
+module_hotfixes=1
+
+[leapp-pulpcore]
+name=pulpcore: Fetch, Upload, Organize, and Distribute Software Packages.
+baseurl=https://yum.theforeman.org/pulpcore/3.16/el8/$basearch/
+gpgkey=https://yum.theforeman.org/pulpcore/3.16/GPG-RPM-KEY-pulpcore
```
evgeni, line 62: we don't have a macro for the pulpcore version (the user usually doesn't have to set it themself).
Katello 4.4 is using 3.16, 4.5 is using 3.17 (well, soon 3.18 I guess) -- should we introduce a macro or have an if/else here directly?

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,173 @@
...
+       file:///etc/pki/rpm-gpg/RPM-GPG-KEY-2025-04-06-puppet7-release
+enabled=1
+gpgcheck=1
+----
+
+* If you are using Puppet 6 instead of Puppet 7, replace the `7` with a `6` in the `leapp-puppet7` entry.
+
+* You need a Puppet repository for the Puppet agent that the installer is using.
+
+For more information, see link:https://docs.theforeman.org/nightly/Managing_Configurations_Puppet/index-katello.html[Configuring Hosts Using Puppet].
```
evgeni, line 96: I don't see how this link is relevant to the upgrade?

guides/common/modules/proc_upgrading-el7-to-el8.adoc
evgeni, line 8:
```suggestion
* {Project} {ProjectVersion} running on CentOS Linux 7.
```

We're upgrading the *current* version, not the previous one.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,173 @@
...
+
+[start=3]
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+. Add {Project} specific repositories to `/etc/leapp/files/leapp_upgrade_repositories.repo`:
++
+----
+[leapp-foreman]
+name=Foreman 3.2
+baseurl=https://yum.theforeman.org/releases/3.2/el8/$basearch
+enabled=1
```
evgeni, line 37:
```suggestion
```

this line is duplicated, accidentally

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,173 @@
...
+# leapp answer --section remove_pam_pkcs11_module_check.confirm=True
+----
+
+The `preupgrade` might fail with a dependency resolution error such as:
+
+* "package rubygem-fx-0.5.0-2.el8.noarch requires rubygem(railties) >= 4.0.0, but none of the providers can be installed"
+* "package rubygem-railties-6.0.4.7-1.el8.noarch requires rubygem(thor) < 2.0, but none of the providers can be installed"
+
+If this happens, do the following to clean up packages that cannot automatically upgrade:
++
```
evgeni, line 134: these `+` render weirdly for me:

image

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,173 @@
...
+# curl -o /etc/yum.repos.d/theforeman-leapp.repo https://copr.fedorainfracloud.org/coprs/g/theforeman/leapp/repo/epel-7/group_theforeman-leapp-epel-7.repo
+----
+[start=2]
+. Install required packages:
++
+----
+# yum install leapp leapp-repository leapp-data-centos
+----
+
+[start=3]
```
maximiliankolb, line 29: I am still not convinced we really need this. It looks very error-prone in the long run. Usually, when you exclude certain steps for certain built targets, the lists in asciidoc still work as expected.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,173 @@
...
+. Run:
++
+----
+# leapp upgrade
+----
+
+[start=10]
+. Reboot the system.
+
+[start=11]
```
maximiliankolb, line 160:
```suggestion
```

You don't need any empty lines between two list elements.

maximiliankolb (changes requested):

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,173 @@
...
+# dnf remove $(dnf repoquery --extras --exclude '*katello*' --exclude '*pulp*' --exclude '*localhost*' --exclude "*$HOSTNAME*" --exclude libmodulemd)
+----
+
+Check the output if it does not contain any local packages.
+
+Remove `rubygem-thor`:
++
+----
+# yum remove rubygem-thor
+----
```
evgeni, line 145: After a bit of thinking, the `dnf remove $(dnf repoquery …)` step is too dangerous. Let's make 

```suggestion
If this happens, do the following to clean up packages that cannot automatically upgrade (`rubygem(thor)` and `rubygem(railties)` in the example above):
+
----
# yum remove rubygem-thor rubygem-railties
----
```

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,173 @@
...
+.Prerequisites
+* {Project} {ProjectVersionPrevious} running on CentOS Linux 7.
+* {Project} installations running on {RHEL} 7 cannot upgrade to {RHEL} 8.
+You must have a {RHEL} compatible LEAPP build to upgrade.
+* Access to available repositories or a local mirror of repositories.
+* No outstanding updates.
+* {Project} installations can be upgraded to CentOS Stream 8 and a {RHEL} rebuild.
+
+.Procedure
+[start=1]
```
evgeni, line 16: is there a way to avoid those `start=X` lines? it's horrible to read and will make us miss updating them if we ever change the procedure

evgeni (changes requested):

evgeni (commented):
Let's support RHEL upgrades too.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
bangelic: Same. I will delete the +.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
bangelic: @evgeni I had to insert the `start=x` lines because the steps kept restarting at 1 throughout the document due to code blocks and other sentences interrupting the flow of the procedure.

maximiliankolb: I super rarely use this. IMHO it's not needed here. Please remove all empty lines between list elements and remove all instances of "start=X".

bangelic: Can we compromise? I will only use `start=x` starting at the 8th step because I have tried playing around with the + sign and attempted dozens of build previews and it will not continue the numbering after the 7th step. The + signs end up in the text when I do a build preview and it doesn't matter where I put them. The numbering starts over at 1 after number 7.

bangelic: Another step was recommended and added. Now the `start=x` will begin at the 9th step if you guys are cool with it.

bangelic: Never mind. Vendula helped me fix the issue I was having. I will take out all `start=x` references.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,173 @@
+[id="Upgrading-Project-from-EL7-to-EL8_{context}"]
+= Upgrading {Project} from Enterprise Linux 7 to Enterprise Linux 8
+
+Use this procedure to upgrade your {Project} installation from Enterprise Linux 7 to Enterprise Linux 8.
+LEAPP is the supported tool for Enterprise Linux 7 to Enterprise Linux 8 upgrades.
+
+.Prerequisites
+* {Project} {ProjectVersionPrevious} running on CentOS Linux 7.
+* {Project} installations running on {RHEL} 7 cannot upgrade to {RHEL} 8.
+You must have a {RHEL} compatible LEAPP build to upgrade.
```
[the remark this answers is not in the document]
evgeni: Why was this marked as resolved?

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,170 @@
...
+# yum remove rubygem-thor rubygem-railties
+----
++
+Check the output if it does not contain any local packages.
++
+Remove `rubygem-thor`:
++
+----
+# yum remove rubygem-thor
+----
```
evgeni, line 150:
```suggestion
```

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,170 @@
...
+
+. Run:
++
+----
+# leapp upgrade
+----
+. Reboot the system.
+
+. After the system reboots, a live system conducts the upgrade, reboots to fix SELinux labels, then reboots into the final Enterprise Linux 8 system.
+. After the system reboots, LEAPP finishes the upgrade, watch it with:
```
evgeni, line 161: starting two points with "After the system reboots" sounds odd to me, but I am not sure how one could re-word this better.

evgeni (changes requested):

guides/common/modules/proc_upgrading-el7-to-el8.adoc
bangelic: @evgeni I made it easy. Just removed "After the system reboots" and it still makes sense.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,170 @@
...
+# leapp answer --section remove_pam_pkcs11_module_check.confirm=True
+----
+
++
+The `preupgrade` might fail with a dependency resolution error such as:
++
+----
+* "package rubygem-fx-0.5.0-2.el8.noarch requires rubygem(railties) >= 4.0.0, but none of the providers can be installed"
+* "package rubygem-railties-6.0.4.7-1.el8.noarch requires rubygem(thor) < 2.0, but none of the providers can be installed"
+----
```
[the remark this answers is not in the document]
bangelic: Fixed

guides/common/modules/proc_upgrading-el7-to-el8.adoc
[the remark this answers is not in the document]
bangelic: Shouldn't have been. It's resolved for real now.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,162 @@
...
+. Install additional OS specific packages (`leapp-data-centos` for CentOS Stream, `leapp-data-rocky` for Rocky Linux, `leapp-data-almalinux` for AlmaLinux`). Not required for {RHEL}.
++
+----
+# yum install leapp-data-centos
+----
+
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+. Add {Project} specific repositories to `/etc/leapp/files/leapp_upgrade_repositories.repo`:
++
```
evgeni, line 35:
```suggestion
. Add {Project} specific repositories to `/etc/leapp/files/leapp_upgrade_repositories.repo`:
+
[options="nowrap", subs="+quotes,verbatim,attributes"]
```

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,162 @@
...
+----
+[leapp-foreman]
+name=Foreman {ProjectVersion}
+baseurl=https://yum.theforeman.org/releases/{ProjectVersion}/3.2/el8/$basearch
```
evgeni, line 39:
```suggestion
baseurl=https://yum.theforeman.org/releases/{ProjectVersion}/el8/$basearch
```

guides/common/modules/proc_upgrading-el7-to-el8.adoc
evgeni: This is still outstanding, I think adding a new macro is the simplest?

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,162 @@
...
+
+. Let LEAPP analyze your system:
++
+----
+# leapp preupgrade
+----
+The first run is expected to fail but report issues.
+Continue to the next step for remediation.
+
+. Fix issues found by LEAPP (check `/var/log/leapp/leapp-report.txt` for details):
```
evgeni, line 120:
```suggestion
. Examine the report in the `/var/log/leapp/leapp-report.txt` file, answer all questions (using `leapp answer`) and manually resolve the other reported problems. The following commands show the most common steps required:
```

evgeni (changes requested):

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,162 @@
...
+= Upgrading {Project} from Enterprise Linux 7 to Enterprise Linux 8
+
+Use this procedure to upgrade your {Project} installation from Enterprise Linux 7 to Enterprise Linux 8.
+LEAPP is the supported tool for Enterprise Linux 7 to Enterprise Linux 8 upgrades.
+
+.Prerequisites
+* {Project} {ProjectVersion} running on Enterprise Linux 7.
+* Access to available repositories or a local mirror of repositories.
+* No outstanding updates.
+* {Project} installations can be upgraded to CentOS Stream 8 and a {RHEL} rebuild.
```
evgeni, line 11:
```suggestion
* {Project} installations running on CentOS 7 can be upgraded to CentOS Stream 8 or a {RHEL} rebuild.
* {Project} installations running on {RHEL} 7 can be upgraded to {RHEL} 8.
```

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,162 @@
...
+
+.Procedure
+. Do not perform this step on {RHEL} as this is for CentOS. Configure the https://copr.fedorainfracloud.org/coprs/g/theforeman/leapp/[@theforeman/leapp COPR Repository], which contains newer LEAPP packages than those shipped by https://wiki.almalinux.org/elevate/[AlmaLinux/ELevate], and support {Project} upgrades:
++
+----
+# curl -o /etc/yum.repos.d/theforeman-leapp.repo https://copr.fedorainfracloud.org/coprs/g/theforeman/leapp/repo/epel-7/group_theforeman-leapp-epel-7.repo
+----
```
evgeni, line 18:
```suggestion
. Configure the repositories to obtain LEAPP.
+
On CentOS, configure the https://copr.fedorainfracloud.org/coprs/g/theforeman/leapp/[@theforeman/leapp COPR Repository], which contains newer LEAPP packages than those shipped by https://wiki.almalinux.org/elevate/[AlmaLinux/ELevate], and support {Project} upgrades:
+
----
# curl -o /etc/yum.repos.d/theforeman-leapp.repo https://copr.fedorainfracloud.org/coprs/g/theforeman/leapp/repo/epel-7/group_theforeman-leapp-epel-7.repo
----
+
On {RHEL}, enable the `rhel-7-server-extras-rpms` repository:
+
----
# subscription-manager repos --enable=rhel-7-server-extras-rpms
----
```

guides/common/modules/proc_upgrading-el7-to-el8.adoc
evgeni:
```diff
diff --git guides/common/attributes.adoc guides/common/attributes.adoc
index 65eec3bb..f2413a79 100644
--- guides/common/attributes.adoc
+++ guides/common/attributes.adoc
@@ -6,6 +6,7 @@
 :ProjectVersion: nightly
 :ProjectVersionPrevious: 3.3
 :KatelloVersion: nightly
+:PulpcoreVersion: 3.18
 :TargetVersion: {ProjectVersion}
 :TargetVersionMaintainUpgrade: {ProjectVersion}
 // The above attribute should point to the GA version number (x.y) for all releases including Beta
diff --git guides/common/modules/proc_upgrading-el7-to-el8.adoc guides/common/modules/proc_upgrading-el7-to-el8.adoc
index 1940d7d2..45988e87 100644
--- guides/common/modules/proc_upgrading-el7-to-el8.adoc
+++ guides/common/modules/proc_upgrading-el7-to-el8.adoc
@@ -61,8 +61,8 @@ module_hotfixes=1
 
 [leapp-pulpcore]
 name=pulpcore: Fetch, Upload, Organize, and Distribute Software Packages.
-baseurl=https://yum.theforeman.org/pulpcore/3.16/el8/$basearch/
-gpgkey=https://yum.theforeman.org/pulpcore/3.16/GPG-RPM-KEY-pulpcore
+baseurl=https://yum.theforeman.org/pulpcore/{PulpcoreVersion}/el8/$basearch/
+gpgkey=https://yum.theforeman.org/pulpcore/{PulpcoreVersion}/GPG-RPM-KEY-pulpcore
 enabled=1
 gpgcheck=1
```

something like this should do (we then need to adjust the attribute differently in the branches, but that's fine, that's what attributes are for)

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,162 @@
...
+
+. Install required packages:
++
+----
+# yum install leapp leapp-repository
+----
+
+. Install additional OS specific packages (`leapp-data-centos` for CentOS Stream, `leapp-data-rocky` for Rocky Linux, `leapp-data-almalinux` for AlmaLinux`). Not required for {RHEL}.
```
evgeni, line 26:
```suggestion
. Install additional OS specific packages (`leapp-data-centos` for CentOS Stream, `leapp-data-rocky` for Rocky Linux, `leapp-data-almalinux` for AlmaLinux). Not required for {RHEL}.
```

bangelic:
"This is still outstanding, I think adding a new macro is the simplest?"
@evgeni Do you have a macro you want to use?

evgeni:
> "This is still outstanding, I think adding a new macro is the simplest?"
> @evgeni Do you have a macro you want to use?

I could swear I posted a follow up comment with an example patch. Can't find it. Stupid GitHub UI folding comments left and right.

I have the following patch applied locally:

```diff
--- guides/common/attributes.adoc
+++ guides/common/attributes.adoc
@@ -6,6 +6,7 @@
 :ProjectVersion: nightly
 :ProjectVersionPrevious: 3.3
 :KatelloVersion: nightly
+:PulpcoreVersion: 3.18
 :TargetVersion: {ProjectVersion}
 :TargetVersionMaintainUpgrade: {ProjectVersion}
 // The above attribute should point to the GA version number (x.y) for all releases including Beta
diff --git guides/common/modules/proc_upgrading-el7-to-el8.adoc guides/common/modules/proc_upgrading-el7-to-el8.adoc
index 1940d7d2..45988e87 100644
--- guides/common/modules/proc_upgrading-el7-to-el8.adoc
+++ guides/common/modules/proc_upgrading-el7-to-el8.adoc
@@ -61,8 +61,8 @@ module_hotfixes=1
 
 [leapp-pulpcore]
 name=pulpcore: Fetch, Upload, Organize, and Distribute Software Packages.
-baseurl=https://yum.theforeman.org/pulpcore/3.16/el8/$basearch/
-gpgkey=https://yum.theforeman.org/pulpcore/3.16/GPG-RPM-KEY-pulpcore
+baseurl=https://yum.theforeman.org/pulpcore/{PulpcoreVersion}/el8/$basearch/
+gpgkey=https://yum.theforeman.org/pulpcore/{PulpcoreVersion}/GPG-RPM-KEY-pulpcore
 enabled=1
 gpgcheck=1
 module_hotfixes=1
```

Read: add `:PulpcoreVersion: 3.18` to `attributes.adoc` and then use `{PulpcoreVersion}` in the code block.

guides/common/modules/proc_upgrading-el7-to-el8.adoc
```diff
@@ -0,0 +1,171 @@
...
+= Upgrading {Project} from Enterprise Linux 7 to Enterprise Linux 8
+
+Use this procedure to upgrade your {Project} installation from Enterprise Linux 7 to Enterprise Linux 8.
+LEAPP is the supported tool for Enterprise Linux 7 to Enterprise Linux 8 upgrades.
+
+.Prerequisites
+* {Project} {ProjectVersion} running on Enterprise Linux 7.
+* Access to available repositories or a local mirror of repositories.
+* No outstanding updates.
+* {Project} installations running on CentOS 7 can be upgraded to CentOS Stream 8 or a {RHEL} rebuild.
```
evgeni, line 11:
```suggestion
* {Project} installations running on CentOS 7 can be upgraded to CentOS Stream 8 or a {RHEL} rebuild.
* {Project} installations running on {RHEL} 7 can be upgraded to {RHEL} 8.
```

evgeni (approved):
There are two outstanding things, which I do not consider blocking:
* the link to the Puppet guide doesn't seem to add value to the uprgrade process? https://github.com/theforeman/foreman-documentation/pull/1248/files#r892002478
* we should mention that RHEL7→RHEL8 does work now. https://github.com/theforeman/foreman-documentation/pull/1248/files#r894551939

@Lennonka did you want to do any further review on this one?

evgeni (approved):

evgeni:
I am hearing Lena has no time today, so I'll just merge this, and then we see from there.

maximiliankolb:
Is there anything left to do here? Do we need to create issue? @bangelic 

In general, I feel like this PR needs some more love.
For example, the title does not match the anchor and file name; and some lines don't really match our style, like `* No outstanding updates.`.

cc @melcorr

PR #1252, opened 2022-05-06
Replace the command in the topic Configuring Satellite to use Externa…

mjivraja:
Replace the command in the topic Configuring Satellite to use External Databases

The installer flags need to be updated as they should have the
correct username. The command replacement is applicable to
connected and disconnected networks both.

External database documentation needs modification for user

https://issues.redhat.com/browse/SATDOC-751

Cherry-pick into:

* [ ] Foreman 3.2
* [ ] Foreman 3.1


maximiliankolb (approved):
LGTM.

Please limit your commit message to 50 characters if possible. You can always add more (and should, if it's anything non-trivial) to why you made the change after an empty line.

mjivraja:
> LGTM.
> 
> Please limit your commit message to 50 characters if possible. You can always add more (and should, if it's anything non-trivial) to why you made the change after an empty line.

Noted @maximiliankolb 
I will take care of the character limit in the title of a commit.

Thank you for notifying.  :)

PR #1254, opened 2022-05-06
Update master for Foreman 3.3

upadhyeammit:
Cherry-pick into:
None

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,6 +1,7 @@
 
 Cherry-pick into:
 
+* [ ] Foreman 3.3
 * [ ] Foreman 3.2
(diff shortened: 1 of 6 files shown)


maximiliankolb (commented):
Thanks for the PR. Everything looks good to me. However, I am unsure if we could simply save the hassle and go straight from "3.2 is the current stable release" to "3.3 is the current stable release" without the RC phase.

upadhyeammit:
> Thanks for the PR. Everything looks good to me. However, I am unsure if we could simply save the hassle and go straight from "3.2 is the current stable release" to "3.3 is the current stable release" without the RC phase.

It use to be like what you have mentioned in former theforeman.org , however there are few challenges doing this,
1. We can't say 3.3 is new stable unless we GA it?
2. This means updated docs of 3.3 are not available in RC phase?

PR #1258, opened 2022-05-09
Added missing ports

ianf77:
Ports had been removed from table 1.2 after Satellite 6.5. This change puts them back.
The information is current for Satellite 6.9. Note that the information in satellite 6.1/Foreman 2.5
is expanded and corrected.

Bug 1958092 - [Doc] Firewall documentation missing ports

https://bugzilla.redhat.com/show_bug.cgi?id=1958092

Cherry-pick into:

* [ ] Foreman 3.2
* [ ] Foreman 3.1

### DIFF
--- a/guides/common/modules/ref_capsule-ports-and-firewalls-requirements.adoc
+++ b/guides/common/modules/ref_capsule-ports-and-firewalls-requirements.adoc
@@ -27,11 +27,15 @@ endif::[]
 
 The following tables indicate the destination port and the direction of network traffic:
 
-.Ports for {SmartProxy} to {Project} Communication
+Ports for {SmartProxy} to {Project} Communication
 [cols="24%,20%,18%,38%",options="header"]
 |====
 | Port | Protocol | Service | Required For
+| 80 | TCP | HTTP | Anaconda, yum, and for obtaining Katello certificate updates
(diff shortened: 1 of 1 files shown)


ianf77:
This change is only for 2.3 (Satellite 6.9)

guides/common/modules/ref_capsule-ports-and-firewalls-requirements.adoc
```diff
@@ -27,11 +27,15 @@ endif::[]
 
 The following tables indicate the destination port and the direction of network traffic:
 
-.Ports for {SmartProxy} to {Project} Communication
+Ports for {SmartProxy} to {Project} Communication
```
maximiliankolb, line 30: just FYI: the dot without space makes this line the title of the table.

Lennonka: I wonder if it was meant as a heading? Then perhaps adding a newline after the heading instead of removing the dot?

Lennonka (changes requested):
LGTM, let's just try to fine tune the formatting

guides/common/modules/ref_capsule-ports-and-firewalls-requirements.adoc
ianf77: It's the table title.

PR #1259, opened 2022-05-10
Remove context from xref

maximiliankolb:
For xrefs, we should not hard-code the context but use the `{context}` attribute instead.

Cherry-pick into:

* [x] Foreman 3.3
* [x] Foreman 3.2
* [x] Foreman 3.1


maximiliankolb:
Cherry-picked: ✅ 
   e895cbdf..4c21f28d  3.1 -> 3.1
   29db0a96..33ffae55  3.2 -> 3.2
   37db7be7..74ccf49d  3.3 -> 3.3

PR #1260, opened 2022-05-10
Remove double whitespace

maximiliankolb:
TIP: Review using `git diff master -w` :)

Cherry-pick into:

* [x] Foreman 3.3
* [x] Foreman 3.2
* [x] Foreman 3.1

### DIFF
--- a/guides/common/modules/con_applying-custom-configuration.adoc
+++ b/guides/common/modules/con_applying-custom-configuration.adoc
@@ -6,7 +6,7 @@ If these flags are not specified during the initial installer run, rerunning of 
 If changes are overwritten, you must run the restore procedure to restore the manual changes.
 For more information, see {InstallingProjectDocURL}restoring-manual-changes-overwritten-by-a-puppet-run_{project-context}[Restoring Manual Changes Overwritten by a Puppet Run].
 
-To view all installer flags available for custom  configuration, run `{installer-scenario} --full-help`.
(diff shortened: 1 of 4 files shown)


upadhyeammit (commented):
Few questions,
1. Do we need to fix white space at all? because if I see rendered docs then it does not include extra white space, as example,
https://docs.theforeman.org/3.2/Installing_Server/index-katello.html#applying-custom-configuration_foreman

2. If we need to do that for any reason, then there are many more in drafts directory, are we intending to clean up that as well?

maximiliankolb:
> Few questions,
> 
> 1. Do we need to fix white space at all? because if I see rendered docs then it does not include extra white space, as example,
>    https://docs.theforeman.org/3.2/Installing_Server/index-katello.html#applying-custom-configuration_foreman
> 2. If we need to do that for any reason, then there are many more in drafts directory, are we intending to clean up that as well?

1. We don't _have_ to, but it makes maintaining and reading code easier. If you think about it the other way: would you accept contributions that add endless whitespace, for example multiple new lines? Also, `  ` is visible in rendered asciidoc. See https://docs.theforeman.org/nightly/Installing_Server/index-katello.html#Installing_Server_Connected_foreman
2. I don't want to spend time cleaning up drafts right now because there is so much more "wrong" with its content that just whitespace issues. It's missing attributes, structure, etc.

Issue #1184, opened 2022-03-25
Katello 4.4+: Rework mirroring policy for Yum and APT content


maximiliankolb:
New mirroring policy with Katello 4.4+:

* [Foreman issue 33939](https://projects.theforeman.org/issues/33939)
* https://github.com/Katello/katello/commit/281e9ee968d46107556f00da86a8fc3a2f836ccb
* https://github.com/Katello/katello/commit/2fe15f43488e9b5f984f77a3ebc4451b2db9554c

See also https://github.com/theforeman/foreman-documentation/pull/1149#discussion_r835215351

PR #1266, opened 2022-05-11
Rework various procedures to add repositories

maximiliankolb:
Resolves https://github.com/theforeman/foreman-documentation/issues/1184 and some other issues. See individual commits for an easier review.

Cherry-pick into:

* [x] Foreman 3.3
* [x] Foreman 3.2
* [x] Foreman 3.1

### DIFF
--- a/guides/common/modules/proc_adding-custom-deb-repositories.adoc
+++ b/guides/common/modules/proc_adding-custom-deb-repositories.adoc
@@ -39,9 +39,9 @@ ifdef::orcharhino[]
 . Optional: In the *Errata URL* field, enter the URL of an errata service.
 endif::[]
 . Select the *Verify SSL* check box if you want to verify that the upstream repository's SSL certificates are signed by a trusted CA.
-. In the *Upstream Username* field, enter the user name for the upstream repository if required for authentication.
+. Optional: In the *Upstream Username* field, enter the user name for the upstream repository if required for authentication.
 Clear this field if the repository does not require authentication.
-. In the *Upstream Password* field, enter the corresponding password for the upstream repository.
+. Optional: In the *Upstream Password* field, enter the corresponding password for the upstream repository.
 Clear this field if the repository does not require authentication.
 . Optional: In the *Upstream Authentication Token* field, provide the token of the upstream repository user for authentication.
 Leave this field empty if the repository does not require authentication.
--- a/guides/common/modules/proc_adding-custom-deb-repositories.adoc
+++ b/guides/common/modules/proc_adding-custom-deb-repositories.adoc
@@ -43,6 +43,8 @@ endif::[]
 Clear this field if the repository does not require authentication.
 . In the *Upstream Password* field, enter the corresponding password for the upstream repository.
 Clear this field if the repository does not require authentication.
+. Optional: In the *Upstream Authentication Token* field, provide the token of the upstream repository user for authentication.
+Leave this field empty if the repository does not require authentication.
 . From the *Download Policy* list, select the type of synchronization {ProjectServer} performs.
 For more information, see xref:Download_Policies_Overview_{context}[].
 . From the *Mirroring Policy* list, select the type of content synchronization {ProjectServer} performs.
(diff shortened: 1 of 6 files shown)


guides/common/modules/proc_adding-custom-deb-repositories.adoc
```diff
@@ -50,7 +50,8 @@ For more information, see xref:Download_Policies_Overview_{context}[].
 . From the *Mirroring Policy* list, select the type of content synchronization {ProjectServer} performs.
 For more information, see xref:Mirroring_Policies_Overview_{context}[].
 . Optional: In the *HTTP Proxy Policy* field, select an HTTP proxy.
-. Optional: If you want, you can clear the *Publish via HTTP* check box to disable this repository from publishing through HTTP.
+. Optional: You can clear the *Unprotected* check box to require a subscription entitlement certificate for accessing this repository.
+By default, the repository is published through HTTP.
```
Lennonka, line 56: Do we allow both HTTP and HTTPS? I noticed in another place that only HTTPS was allowed for security reasons and I wonder what the case is here.

guides/common/modules/proc_creating-a-custom-file-type-repository.adoc
```diff
@@ -7,24 +7,13 @@ You must create a product and then add a {customrepo}.
...
-{Project} automatically completes the *Label* field based on what you have entered for *Name*.
-. Optional: From the *GPG Key* list, select the GPG key for the product.
-. Optional: From the *Sync Plan* list, select a synchronization plan for the product.
-. In the *Description* field, enter a description of the product, and then click *Save*.
-
-To create a repository for your {customproduct}, complete the following procedure:
-
-. In the Products window, select the name of a product that you want to create a repository for.
-. Click the *Repositories* tab, and then click *New Repository*.
+. In the {ProjectWebUI}, navigate to *Content > Products* and select a product that you want to create a repository for.
```
Lennonka, line 10: One list item for each step, pls

guides/common/modules/proc_adding-custom-deb-repositories.adoc
```diff
@@ -10,8 +10,10 @@ To use the CLI instead of the {ProjectWebUI}, see the xref:cli-adding-custom-deb
 {Project} automatically completes the *Label* field based on what you have entered for *Name*.
 . Optional: In the *Description* field, enter a description for the repository.
 . From the *Type* list, select `deb` as type of repository.
-. In the *Upstream URL* field, enter the URL of the external repository to use as a source.
+. Optional: In the *Upstream URL* field, enter the URL of the external repository to use as a source.
 You can find the upstream URLs on Debian-based systems in `/etc/apt/sources.list`.
++
+If you do not enter a upstream URL, you can manually upload packages.
```
Lennonka, line 16: **an** upstream URL (it goes by pronunciation, not spelling) -- also below

Lennonka (changes requested):
Just a couple of details and it will be perfect to go.
Btw, thanks for the tip to go by commit, most useful! :)

guides/common/modules/proc_adding-custom-deb-repositories.adoc
maximiliankolb: I have tested this on Foreman 3.2 and it's also available through HTTP & I found no setting to prohibit downloads over HTTP. Do you know more @m-bucher ?

guides/common/modules/proc_adding-custom-deb-repositories.adoc
maximiliankolb: Fixed both times.

guides/common/modules/proc_creating-a-custom-file-type-repository.adoc
maximiliankolb: Fixed.

guides/common/modules/proc_adding-custom-deb-repositories.adoc
m-bucher: I do not know if there is a special handling for RedHat repos, but the repository will always be distributed via HTTPS where clients can authenticate using subscription-manager.
But if you want/have to access the repository without authentication (e.g. because the OS needs a special repository to install subscription-manager in the first place), it has to be offered via HTTP, because HTTPS-repos always require client-atuhentication using SSL-certs.
For Debian repositories distributing over HTTP is perfectly save as far as authenticity is concerned (as long as a secure GPG-key has been used for signature).

Lennonka: So maybe we should mention both and recommend HTTPS as preferred? Or mention cases when HTTP is necessary?

maximiliankolb: I have tested this again. It's possible to access content through HTTP.

maximiliankolb:

   e2585381..301c2463  3.1 -> 3.1
   7c1256c4..9f76ea42  3.2 -> 3.2
   0d63628f..c9943080  3.3 -> 3.3

PR #1268, opened 2022-05-11
Fixed asciidoc syntax in table entries

mdolezelova:
Cherry-pick into:

* [x] Foreman 3.3
* [x] Foreman 3.2
* [x] Foreman 3.1

### DIFF
--- a/guides/common/modules/ref_facts-settings.adoc
+++ b/guides/common/modules/ref_facts-settings.adoc
@@ -15,10 +15,10 @@ The content of this fact should be the full label of the organization.
 | *Ignore facts for domain* | No | Stop updating domain values from facts.
 | *Update subnets from facts* | None | {Project} updates a host's subnet from its facts.
 | *Ignore interfaces facts for provisioning* | No | Stop updating IP and MAC address values from facts (affects all interfaces).
-| *Ignore interfaces with matching identifier* | `lo`, `en*v*`, `usb*`, `vnet*`, `macvtap*`, `;vdsmdummy;`, `veth*`, `tap*`, `qbr*`, `qvb*`, `qvo*`, `qr*`, `qg*`, `vlinuxbr*`, `vovsbr*`, `br-int` | Skip creating or updating host network interfaces objects with identifiers matching these values from incoming facts.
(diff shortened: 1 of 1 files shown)


mdolezelova:
Hi @melcorr , I had to do this fix because the Administration guide for Satellite 6.11 that I've set up in Pantheon has a build issue. I troubleshooted it downstream to make sure what must be changed in order to make the build work. And with this fix, it works.  Please could you review the PR and merge it if no issues? Thank you!

melcorr:
How on earth is this the cause of the problem? 
Does removing the - signs change the information?

melcorr:
How on earth is this the cause of the problem? 
Does removing the - signs change the information?

evgeni (changes requested):
those settings literally contain `qr-*`, so if you change it here, the rendered doc won't match the reality.
Would using something like:
```
`['lo', 'en*v*', 'usb*', 'vnet*', 'macvtap*', ';vdsmdummy;', 'veth*', 'tap*', 'qbr*', 'qvb*', 'qvo*', 'qr-*', 'qg-*', 'vlinuxbr*', 'vovsbr*', 'br-int']`
```

be an alternative?

mdolezelova:
> those settings literally contain `qr-*`, so if you change it here, the rendered doc won't match the reality. Would using something like:
> 
> ```
> `['lo', 'en*v*', 'usb*', 'vnet*', 'macvtap*', ';vdsmdummy;', 'veth*', 'tap*', 'qbr*', 'qvb*', 'qvo*', 'qr-*', 'qg-*', 'vlinuxbr*', 'vovsbr*', 'br-int']`
> ```
> 
> be an alternative?

I'm sorry, this does not work. I've tried multiple  things, but the only thing I could get it building is removing the dash. The Asciidoc syntax looks correct. Also Asciidoctor itself builds the title correctly. However, the issue is that Pantheon, our publication tool, first translates it to DocBook and then the title is built. And here the translation to Docbook for some reason fails because of the syntax including -*. Honestly, I though this is a mistake as the dash isn't included with the other options. But if it needs to be there, I need to continue to investigate how to make this working.

mdolezelova:
> If you're curious, the data comes from here: https://github.com/theforeman/foreman/blob/75e7c0bf6d3fa1049931ec9000a0819e72201a05/app/registries/foreman/settings/facts.rb#L21-L40

Thank you, @evgeni !

mdolezelova:
This PR is now obsolete. A new one will be created. Closing.

PR #1292, opened 2022-05-26
Update MacOS build instructions

lzap:
It looks our `bundle install` must be done both in the root and web/ directories. This moves this into a new make target. Also Ruby is marked as deprecated on MacOS now and will not be updated from the legacy 2.6 version, so use brew Ruby instead.

Also serve is the default makefile target, that is confusing. Perhaps html is a better target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ ifeq ($(UNAME), Darwin)
 CP_ARGS =
 endif
 
-all: serve
+all: html
 
 prep:
 	bundle install
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ DEST := result
 PORT := 5000
 VERSION_LINKS := 3.3 3.2 3.1 3.0 2.5 2.4
 
-.PHONY: all clean html web compile serve
+.PHONY: all clean html web compile serve prep
 
 UNAME = $(shell uname)
 ifeq ($(UNAME), Linux)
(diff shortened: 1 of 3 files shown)


maximiliankolb:
Just for reference without looking at your change: You can also use the Docker container that is present in this repo. This works on my machine running macOS and should also run on all other OS (with a small tweak for SELinux that is documented in the README).

on macos: `podman build --tag foreman_documentation .` & `rm -rf guides/build && podman run --rm -v $(pwd):/foreman-documentation foreman_documentation make html`

I use an alias to run this command: `udk='rm -rf guides/build && docker run --rm -v $(pwd):/foreman-documentation foreman_documentation make html BUILD=katello && chr guides/build'` with "chr" being an alias for opening a directory with my browser.

Thanks for the effort @lzap. Your change looks reasonable, but I did not test it. Can someone using the Makefile please confirm?

lzap:
Sarah from the RHD team is on MacOS I believe, I changed this for her. It is not our primary platform anyway, so perhaps let's just trust me? I know its very bold... :)

maximiliankolb:
> Sarah from the RHD team is on MacOS I believe, I changed this for her. It is not our primary platform anyway, so perhaps let's just trust me? I know its very bold... :)

very bold indeed.

maximiliankolb:
Merged to master. 🚀

PR #1294, opened 2022-05-30
Add procedure to remove task by ID

maximiliankolb:
This PR adds a procedure to delete tasks by ID.

Cherry-pick into:

* [x] Foreman 3.3
* [x] Foreman 3.2
* [x] Foreman 3.1

### DIFF
--- a/guides/common/modules/proc_deleting_a_task_by_id.adoc
+++ b/guides/common/modules/proc_deleting_a_task_by_id.adoc
@@ -1,0 +1,32 @@
+[id="Deleting_a_Task_by_ID_{context}"]
+= Deleting a Task by ID
+
+You can delete tasks by ID, for example if you have submitted confidential data by mistake.
+
+.Procedure
+. Connect to your {ProjectServer} using SSH:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# ssh root@{foreman-example-com}
+----
+. Optional: View the task:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# hammer task info --id _My_Task_ID_
+----
+. Delete the task:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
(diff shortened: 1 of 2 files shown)


guides/common/modules/proc_deleting_a_task_by_id.adoc
```diff
@@ -0,0 +1,32 @@
...
+# foreman-rake foreman_tasks:cleanup TASK_SEARCH="id=_My_Task_ID_"
+----
+. Optional: Ensure the task has been removed from {ProjectServer}:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# hammer task info --id _My_Task_ID_
+----
++
+Note that this command returns a non-zero exit code.
```
melcorr, line 32: What does this mean to the user?

maximiliankolb: So potentially, your terminal emulator aka shell bleeps or marks your command prompt red if the previous command does not have zero as return code:

Screenshot 2022-05-30 at 17 01 27

false returns 1, true returns zero, `$?` is the return code of the previously executed command.

By definition, anything non-zero means it failed. Users should always stop their procedure and investigate if a command returns anything else than zero.

melcorr:
```suggestion
Note that because the task is deleted, this command returns a non-zero exit code.
```
Can we change to something like this?

maximiliankolb: Applied.

melcorr (approved):
Thank you for this!

melcorr:
Please cherry-pick!

maximiliankolb:
Cherry-picked: ✅ 
   9d929804..1fdefac9  3.1 -> 3.1
   ef9bf74b..d397b381  3.2 -> 3.2
   62103073..a3360b27  3.3 -> 3.3

PR #1296, opened 2022-05-31
Find unused modules

maximiliankolb:
Provide a script to find potentially unused source files.


maximiliankolb:
Sure. This was a first PoC. I might be able to create a PR Friday or next week to have a look at all files that the script returns. Should be rather straight forward.

PR #1299, opened 2022-06-02
Setup authentication token from SCC Manager

maximiliankolb:
This is necessary on instances running Katello 4.3+ and Foreman SCC
Manager plugin 1.8.20+.

cc @nadjaheitmann

Cherry-pick into:

* [x] Foreman 3.3

### DIFF
--- a/guides/common/modules/proc_installing-the-scc-manager.adoc
+++ b/guides/common/modules/proc_installing-the-scc-manager.adoc
@@ -22,6 +22,7 @@ Perform the following steps to install the _SCC Manager_ plug-in on your {Projec
 ----
 # foreman-rake db:migrate
 # foreman-rake db:seed
+# foreman-rake foreman_scc_manager:setup_authentication_tokens
 ----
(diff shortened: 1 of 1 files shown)


guides/common/modules/proc_installing-the-scc-manager.adoc
```diff
@@ -22,6 +22,7 @@ Perform the following steps to install the _SCC Manager_ plug-in on your {Projec
 ----
 # foreman-rake db:migrate
 # foreman-rake db:seed
+# foreman-rake foreman_scc_manager:setup_authentication_tokens
```
nadjaheitmann, line 25: Actually, it is only necessary to run the command if you are already running Katello < 4.3 with SCC manager version 1.8.20 and THEN upgrade Katello to 4.3 and up.

maximiliankolb:
Nadja und I decided that the best place for this is probably the readme in foreman_scc_manager.

PR #1305, opened 2022-06-03
Updated Product Version attribute in steps 7 & 9

bangelic:
SATDOC-919 specified that the Target version in the Upgrading and Updating Guide in Section 3.1.1 was incorrect in steps 7 and 9. The Target version was 6.10 and needed to be 6.11. I updated the attribute to reflect version 6.11.

Cherry-pick into:

* [X] Foreman 3.3
* [X] Foreman 3.2
* [ ] Foreman 3.1

### DIFF
--- a/guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
+++ b/guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
@@ -60,7 +60,7 @@ These changes are applied to the `/etc/foreman-maintain/foreman-maintain-hammer.
 +
 [options="nowrap" subs="attributes"]
 ----
-# {foreman-maintain} upgrade check --target-version {TargetVersionMaintainUpgrade}
+# {foreman-maintain} upgrade check --target-version {ProductVersion}
 ----
 +
 Review the results and address any highlighted error conditions before performing the upgrade.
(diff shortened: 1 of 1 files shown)


melcorr:
This change doesn't make sense to me. 
ProductVersion is set as TargetVersionMaintainUpgrade in the attributes file, so the change looks redundant?

guides/common/attributes.adoc::TargetVersionMaintainUpgrade: {ProjectVersion}
guides/common/attributes-satellite.adoc::TargetVersionMaintainUpgrade: {ProductVersion}
guides/common/attributes-satellite.adoc::TargetVersionMaintainUpgrade: {ProductVersionPrevious}

guides/common/attributes-satellite.adoc::ProductVersion: 6.11

Please explain more why this change needs to happen?

bangelic:
Hi @melcorr ,

Per SATDOC-919: https://issues.redhat.com/browse/SATDOC-919, the issue is that in the [Upgrading and Updating Foreman Guide](https://docs.theforeman.org/3.1/Upgrading_and_Updating/index-satellite.html#Upgrading_Server_upgrade-guide), in 3.1.1. Upgrading a Connected Satellite Server, step 7. `# satellite-maintain upgrade check --target-version 6.10`and  step 9 `satellite-maintain upgrade run --target-version 6.10` we mention that the target version is 6.10 when it needs to be 6.11. According to the comments in the ticket, the attribute **{ProductVersion}** is defined in the **attributes-satellite.adoc** file and needs to be used for the target version in the 6.11 release. I used the **{ProductVersion}** attribute to update the target version.

PR #1310, opened 2022-06-08
Added necessary syntax (backslashes) so that downstream Admin guide b…

mdolezelova:
…uilds correctly

Cherry-pick into:

* [x] Foreman 3.3
* [x] Foreman 3.2
* [x] Foreman 3.1

### DIFF
--- a/guides/common/modules/ref_facts-settings.adoc
+++ b/guides/common/modules/ref_facts-settings.adoc
@@ -15,10 +15,10 @@ The content of this fact should be the full label of the organization.
 | *Ignore facts for domain* | No | Stop updating domain values from facts.
 | *Update subnets from facts* | None | {Project} updates a host's subnet from its facts.
 | *Ignore interfaces facts for provisioning* | No | Stop updating IP and MAC address values from facts (affects all interfaces).
-| *Ignore interfaces with matching identifier* | [`lo`, `en*v*`, `usb*`, `vnet*`, `macvtap*`, `;vdsmdummy;`, `veth*`, `tap*`, `qbr*`, `qvb*`, `qvo*`, `qr-*`, `qg-*`, `vlinuxbr*`, `vovsbr*`, `br-int`] | Skip creating or updating host network interfaces objects with identifiers matching these values from incoming facts.
+| *Ignore interfaces with matching identifier* | [`lo`, `en*v*`, `usb*`, `vnet*`, `macvtap*`, `;vdsmdummy;`, `veth*`, `tap*`, `qbr*`, `qvb*`, `qvo*`, `qr-\*`, `qg-\*`, `vlinuxbr*`, `vovsbr*`, `br-int`] | Skip creating or updating host network interfaces objects with identifiers matching these values from incoming facts.
 You can use a * wildcard to match identifiers with indexes, e.g. `macvtap*`.
 The ignored interface raw facts are still stored in the database, see the *Exclude pattern* setting for more details.
-| *Exclude pattern for facts stored in satellite* | [`lo`, `en*v*`, `usb*`, `vnet*`, `macvtap*`, `;vdsmdummy;`, `veth*`, `tap*`, `qbr*`, `qvb*`, `qvo*`, `qr-*`, `qg-*`, `vlinuxbr*`, `vovsbr*`, `br-int`, `load_averages::*`, `memory::swap::available*`, `memory::swap::capacity`, `memory::swap::used*`, `memory::system::available*`, `memory::system::capacity`, `memory::system::used*`, `memoryfree`, `memoryfree_mb`, `swapfree`, `swapfree_mb`, `uptime_hours`, `uptime_days`] | Exclude pattern for all types of imported facts (Puppet, Ansible, rhsm).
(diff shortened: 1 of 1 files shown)


mdolezelova:
Hi @Lennonka , please could you check this one and merge it for me? Thank you!

guides/common/modules/ref_facts-settings.adoc
```diff
@@ -15,10 +15,10 @@ The content of this fact should be the full label of the organization.
 | *Ignore facts for domain* | No | Stop updating domain values from facts.
 | *Update subnets from facts* | None | {Project} updates a host's subnet from its facts.
 | *Ignore interfaces facts for provisioning* | No | Stop updating IP and MAC address values from facts (affects all interfaces).
-| *Ignore interfaces with matching identifier* | [`lo`, `en*v*`, `usb*`, `vnet*`, `macvtap*`, `;vdsmdummy;`, `veth*`, `tap*`, `qbr*`, `qvb*`, `qvo*`, `qr-*`, `qg-*`, `vlinuxbr*`, `vovsbr*`, `br-int`] | Skip creating or updating host network interfaces objects with identifiers matching these values from incoming facts.
+| *Ignore interfaces with matching identifier* | [`lo`, `en*v*`, `usb*`, `vnet*`, `macvtap*`, `;vdsmdummy;`, `veth*`, `tap*`, `qbr*`, `qvb*`, `qvo*`, `qr-\*`, `qg-\*`, `vlinuxbr*`, `vovsbr*`, `br-int`] | Skip creating or updating host network interfaces objects with identifiers matching these values from incoming facts.
```
Lennonka, line 18: For some reason, `qg-\*` renders with the backslash in the output on my computer. Can you please double check on yours if it's necessary?

Lennonka (commented):
LGTM, just one weird thing in the comment below...

guides/common/modules/ref_facts-settings.adoc
mdolezelova: Hi @Lennonka , I 've checked this change in downstream and there it renders correctly. Actually it was of my highest interest to get it proceed correctly in Pantheon. So I've made this change on 6.11 branch downstream and the guide in Pantheon started building correctly. It was just a test as I knew it will be soon overwritten by the automatic sync from upstream. So after the sync, the Administration guide stopped building again. So to summarize, in downstream, it renders correctly just with this syntax. Wasn't able to find any other syntax that would not break the build in Pantheon.

maximiliankolb: Can you please check if it renders correctly in _upstream_ docs?

mdolezelova: Hi @Lennonka , @maximiliankolb in upstream docs the first part renders correctly, but the second has a backslash.
Like this: qr-*, qg-\*
And should be: qr-*, qg-*
In dowstream docs, both render correctly: qr-*, qg-*
Not sure why it is and what to do with it.
The only certain thing is that if we don't correct it, the build doesn't proceed in downstream and we can't publish the whole guide for 6.11. 
Lena, please could you help to figure out how to make this rendering correctly in both upstream and downstream?
Thank you!

mdolezelova: @maximiliankolb okay, perfect!  I will use the condition for Satellite then. As i really need to get the rendering resolved for downstream. Thank you!

mdolezelova: Hi @maximiliankolb , I've discovered another problem. Actually, I think the syntax might be wrong in the upstream docs. 
I'm checking the branch 3.1 (as Satellite 6.11 docs are generated from it) and master.
On both branches, there is this in the ref_fact-settings.adoc file: `qr-*`, `qg-*`,
However, if you check Foreman docs, that * sign, which is causing the issue with downstream build is not visible anyway - please check https://docs.theforeman.org/3.1/Administering_Red_Hat_Satellite/index-foreman-el.html#facts_settings_admin and https://docs.theforeman.org/nightly/Administering_Red_Hat_Satellite/index-foreman-el.html#facts_settings_admin
It's rendered just as qr- and qg-.
The question is whether it's correct like this or it should be displayed as qr-* and qg-*?

maximiliankolb: I see what you mean. I have no idea, but I assume it should be displayed as expected from the source file. Maybe a backslash fixes the issue?

While you're at it: Please remove `satellite` and use an attribute here: `Exclude pattern for facts stored in satellite`.

mdolezelova: Hi @maximiliankolb , apologies, but the backslashes is what I suggested by my commit. But you didn't want to accept it. I believe the backlash is the correct Asciidoc syntax here.  I really don't understand why in upstream, it does work for the first occurrence, but doesn't work for the second one (qg-* is not rendered correctly). In downstream, the backlashes fix both occurrences (I've checked by a commit to the downstream repo). However, this was corrected just temporarily as it was overwritten by upstream branch 3.1 automatically.  @Lennonka , please could you try to help with the correct syntax here? Apologies, I've ran out of ideas and am flooded with other tasks. Thank you!

maximiliankolb: Hi @mdolezelova 

I think using a single backslash works: https://github.com/theforeman/foreman-documentation/pull/1315

You can update your PR and then check if that works for you downstream. If it improves upstream docs, I will happily merge of course.

maximiliankolb:
   f1b2ca4e..eab18b39  3.1 -> 3.1
   49e6346f..358b1d99  3.2 -> 3.2
   0dda4a57..497a05e2  3.3 -> 3.3

For future reference, please have a look at https://cbea.ms/git-commit/#seven-rules @mdolezelova

PR #1230, opened 2022-04-19
Improve title references

Lennonka:
Phase 1 for Issue #954
This PR suggests to use attributes for guide titles, renames some guide folders and improves wording of some titles.
 
Cherry-pick into:

* [x] Foreman 3.2
* [x] Foreman 3.1

PR #1318, opened 2022-06-15
Rename Planning Guide in Satellite

Lennonka:
Cherry-pick into:

* [x] Foreman 3.1
* [x] Foreman 3.3

### DIFF
--- a/guides/common/attributes-satellite.adoc
+++ b/guides/common/attributes-satellite.adoc
@@ -8,7 +8,7 @@
 :InstallingProjectDocURL: {BaseURL}installing_satellite_server_from_a_connected_network/index#
 :InstallingDisconnectedDocURL: {BaseURL}installing_satellite_server_from_a_disconnected_network/index#
 :InstallingSmartProxyDocURL: {BaseURL}installing_capsule_server/index#
-:PlanningDocURL: {BaseURL}planning_for_red_hat_satellite/index#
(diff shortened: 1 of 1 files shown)


Lennonka:
@mdolezelova Can you give this an ack? :)

maximiliankolb:
Quick question: are you sure you want to merge this to master and then only cherry-pick to 3.1? Your call, but sounds strange.

Lennonka:
@maximiliankolb I'm not sure, I'm suggesting it because it concerns Satellite only. We can cherrypick to all newer as well if you prefer. LMK

Lennonka:
Note: Cherrypick has to wait until #1230 is cherrypicked.

maximiliankolb:
> @maximiliankolb I'm not sure, I'm suggesting it because it concerns Satellite only. We can cherrypick to all newer as well if you prefer. LMK

@Lennonka I have no preference, but I assume you want it in 3.1 and then 3.5 (will be branched from master), but it would be weird to _not_ want it in 3.3. I would for sure forget about it when 3.5 arrives.

I either recommend either cherry-picking to all versions in between, or open a PR against 3.1 directly if you're unsure about upcoming versions. But totally your call.

Lennonka:
@melcorr Melanie, what's your opinion on the cherrypicks? I think it's not necessary for Foreman higher than 3.1 when it will be included in master.

PR #1330, opened 2022-06-17
Added Port 443 to Capsule incoming traffic table

bangelic:
Per SAT-6405, I added port 443 in the Destination Port of the Capsule incoming traffic table where port 8443 is mentioned.

Cherry-pick into:

* [x] Foreman 3.3
* [x] Foreman 3.2
* [x] Foreman 3.1

### DIFF
--- a/guides/common/modules/ref_capsule-ports-and-firewalls-requirements.adoc
+++ b/guides/common/modules/ref_capsule-ports-and-firewalls-requirements.adoc
@@ -35,6 +35,11 @@ ifdef::katello,satellite,orcharhino[]
 | 443, 80 | TCP | HTTPS, HTTP| Client | Content Host Registration | {SmartProxy} CA RPM installation
 | 443 | TCP | HTTPS | {ProjectName} | Content Mirroring | Management
 | 443 | TCP | HTTPS | {ProjectName} | {SmartProxy} API | Smart Proxy functionality
+| 443 | TCP | HTTPS | Client | Content Host registration | Initiation
+
+Uploading facts
+
+Sending installed packages and traces
 | 5647 | TCP | AMQP | Client | goferd message bus | Forward message to client (optional)
 
 Katello agent to communicate with Qpid dispatcher
@@ -47,11 +52,7 @@ ifndef::katello,satellite,orcharhino[]
 | 8443 | TCP | HTTPS | {ProjectName} | {SmartProxy} API | Smart Proxy functionality
 endif::[]
 ifdef::katello,satellite,orcharhino[]
-| 8443 | TCP | HTTPS | Client | Content Host registration | Initiation
-
-Uploading facts
-
-Sending installed packages and traces
+| 8443 | TCP | HTTPS | Client | Content Host registration | Deprecated and only needed for Client hosts deployed before upgrades
 | 9090 | TCP | HTTPS | Client | OpenSCAP | Configure Client
 | 9090 | TCP | HTTPS | Discovered Node|Discovery |Host discovery and provisioning
 | 9090 | TCP | HTTPS | {ProjectName} | {SmartProxy} API | {SmartProxy} functionality


evgeni:
If you install a fresh Katello/Satellite/Orcharino, you won't need port 8443 - all newly deployed systems will use 443 exclusively.
However, if you have client hosts that have been upgraded since before Foreman 3.1, those will be still talking via 8443.

Thus it would be nice if we could somehow point out that 8443 is legacy/deprecated and only really needed in existing setups.

I *think* this is done best by having two lines, one for 443 with the current text, and one for 8443 with some additional sentence explaining the situation.

bangelic:
@evgeni Let me now if that sentence makes sense for Port 8443.

maximiliankolb:
Is this related to #1344?

evgeni:
> Is this related to #1344?

It is related as in "the same code change triggered both updates to the docs", yes.

guides/common/modules/ref_capsule-ports-and-firewalls-requirements.adoc
```diff
@@ -47,11 +47,12 @@ ifndef::katello,satellite,orcharhino[]
 | 8443 | TCP | HTTPS | {ProjectName} | {SmartProxy} API | Smart Proxy functionality
 endif::[]
 ifdef::katello,satellite,orcharhino[]
-| 8443 | TCP | HTTPS | Client | Content Host registration | Initiation
+| 443 | TCP | HTTPS | Client | Content Host registration | Initiation
 
 Uploading facts
 
 Sending installed packages and traces
+| 8443 | TCP | HTTPS | Client | Content Host registration | Deprecated and only needed for Client hosts deployed before upgrades
```
evgeni, line 55: I'm torn on the wording, also partially because I'm unhappy with the *old* description (which you didn't write).

The purpose of the port in this scenario is to allow the client to talk to the RHSM API for obtaining and updating subscription data. This happens during *registration* but also afterwards (subscription-manager regularly updates subscriptions of the content host).

Now, with the changes we implemented, new hosts will register via `443` and store that in their configs, so that later communication also happens via `443`. However, existing hosts (those, that are already registered) don't get a configuration update and continue talking via `8443`.

Thus saying it's used for *registration* seems odd to me -- those hosts aren't doing any registration anymore as they are already registered. They do, however, keep on using that port for communication until reconfigured by the admin.

I wouldn't block this PR on improving this wording, but that improvement should happen!

bangelic: @evgeni If you have something specific you want me to change from "Content Host registration", let me know, otherwise we can keep as is and address in a future PR and we can merge this one. Up to you.

evgeni (approved):
I'm good with the change (as it documents the current state of deployments).

We should def improve the wording in a later PR tho

guides/common/modules/ref_capsule-ports-and-firewalls-requirements.adoc
bangelic: @evgeni All I could think of was something related to "Content Host communication" but I'm not a subject matter expert so I defer to all of you who know this stuff better than me. :-)

ehelms: No idea, so :roller_skate: on

bangelic:
@evgeni Sounds good. One of the checks was unsuccessful. Do I need to push this again before it gets merged?

evgeni:
You can rebase and the red check will go away.

bangelic:
> You can rebase and the red check will go away.

@evgeni Done!

bangelic:
@evgeni Is this good to be merged?

bangelic:
> @bangelic Please rebase to master and squash your commits. Reword your commit message according to https://cbea.ms/git-commit/#seven-rules as stated in the contributing guidelines and please avoid downstream terminology (at least in the subject).

@maximiliankolb Done!

maximiliankolb:
@bangelic It's still past tense and says "capsule" in the commit message subject line.

maximiliankolb:
   96920db07..a3df0205b  3.3 -> 3.3
   2b1b47f21..09b0c000c  3.2 -> 3.2
   dd984e2d5..7b5b77bef  3.1 -> 3.1

PR #1333, opened 2022-06-20
drop RepoRHEL8ServerAnsible definition

evgeni:
we don't use this variable in the docs anymore

Cherry-pick into:

* [x] Foreman 3.3
* [x] Foreman 3.2
* [x] Foreman 3.1


maximiliankolb (approved):
LGTM

PR #1340, opened 2022-06-22
Remove puppet options from loadbalancing guide for puppet disabled sc…

AkshayGadhaveRH:
…enarios

Dropped puppet opions from the installer when setting up the loadbalancer, in scenarios where puppet is disabled.

Cherry-pick into:

* [x] Foreman 3.3
* [x] Foreman 3.2
* [x] Foreman 3.1

Hi @ekohl, Ron Lavi has suggested that I reach out to you and check if these changes are valid. Could you give it a look over once?

### DIFF
--- a/guides/doc-Configuring_Load_Balancer/topics/Configuring_Capsule_Server_with_Custom_SSL_Certificates_and_without_Puppet_for_Load_Balancing.adoc
+++ b/guides/doc-Configuring_Load_Balancer/topics/Configuring_Capsule_Server_with_Custom_SSL_Certificates_and_without_Puppet_for_Load_Balancing.adoc
@@ -141,17 +141,10 @@ root@_{smartproxy-example-com}_:__{smartproxy-example-com}__-certs.tar
 ----
 
 . Append the following options to the `{foreman-installer}` command that you obtain from the output of the `{certs-generate}` command.
-Set the `--puppet-ca-server` option to point to {SmartProxyServer} where you enter the command.
-You must install Puppet CA on your {SmartProxyServer}s, regardless of whether you intend to use it or not.
-Puppet is configured in its default single-node configuration.
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
 --certs-cname "_loadbalancer.example.com_" \
---puppet-dns-alt-names "_loadbalancer.example.com_" \
---puppet-ca-server "_{smartproxy-example-com}_" \
---foreman-proxy-puppetca "true" \
---puppet-server-ca "true" \
 --enable-foreman-proxy-plugin-remote-execution-script
 ----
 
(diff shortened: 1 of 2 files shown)


ekohl:
Content wise I think this is correct, but @melcorr is also correct in that we expect a certain standard for git commit messages.

ekohl:
In this case it's clear that the subject line is too long. See https://cbea.ms/git-commit/#limit-50 for that.

The more important part is realizing what this change is and why it was made (https://cbea.ms/git-commit/#why-not-how). In this particular case it is needed because Puppet is turned off by default and passing the options in the guide will actually turn Puppet back on. Knowing this is useful in determining how far back it needs to be cherry picked: namely to the point where Puppet was turned off by default.

This may be harder for someone who mostly works with downstream. The release notes on docs.theforeman.org are also incomplete (Katello 4.3 headline features are missing: https://docs.theforeman.org/3.1/Release_notes/index-katello.html#katello-headline-features) so finding out when this exactly changed is not trivial. In that case you can ask other people or put it in the commit message / PR description. In this particular case it's https://projects.theforeman.org/issues/33337 (found under https://docs.theforeman.org/3.1/Release_notes/index-katello.html#_installer_4 (unrelated note: we should find a better anchor for that, this isn't stable)) so Foreman 3.1, but only in Katello scenarios.

So for a short summary I'd say something like `Prevent enabling Puppet in Puppetless load balanced setups`. That's 58 characters, so past 50 but still well below 72. I'll leave it up to you to decide if this is clear enough.

In the body I'd describe what brought up this change (namely turning off Puppet by default), why it is a problem (Puppet is turned on by the current instructions, optionally when it was introduced (Foreman 3.1) and how the commit solves it (not passing the options maintains the current answers which means Puppet isn't turned on).

So these are how I would apply the article.

Another way of looking at it is to imagine the reader of the commit message. Imagine they don't know anything, but you want to summarize the research you did to get the issue resolved. That way they don't have to repeat that work to review the change. I certainly have used `git log` and `git blame` to find out why a change was made years ago (sometimes even by myself) and found it very useful to understand what went through the author's head at the time.

AkshayGadhaveRH:
Hi @melcorr apologies for the incorrect format in the commit message, I'll make sure to stick to the guidelines here on.

@ekohl Thanks for reviewing this as well as for the commit message instructions, should I push another commit with a better message?

maximiliankolb:
I think git commit messages are always a judgement call, but the rules linked above can be viewed as guidelines. Making your commit message 58 characters long is IMHO perfectly fine if the alternative is "short but useless".

You can ask yourself the question "what happens if I apply this commit to my branch?".

Why am I a stickler for good commit messages? Because I use `git blame` to find out when and why something was changed. This really helps to understand the progress of a branch.

If you're looking for examples, run `gitk &` on your terminal on the master branch. Then you can scroll through all commits and view the commit & commit message to get a feeling.

AkshayGadhaveRH:
Hey @ekohl @maximiliankolb, can this be merged?

maximiliankolb:
I've shortened your commit message.

   9f0938558..65b5b8e1d  3.1 -> 3.1
   09b0c000c..81b45796e  3.2 -> 3.2
   a3df0205b..2a4e4d670  3.3 -> 3.3

PR #1348, opened 2022-06-23
Add missing ansible repo in the upgrade guide

AkshayGadhaveRH:
Added ansible repo to the list of repos enabled on the clone system, without which the cloning will fail.

Cherry-pick into:

* [x] Foreman 3.3
* [x] Foreman 3.2
* [x] Foreman 3.1


evgeni (approved):
It would be cool if the other repos would also use attributes, but it's not something this PR introduced, so not required to be fixed in here either.

AkshayGadhaveRH:
Hey @evgeni, thanks for checking this. I'll push another PR for replacing the attributes for the other repos.

evgeni:
Thanks!

Merged and picked:
   25e3fe57..a871ebfd  3.1 -> 3.1
   38ffb743..ad44add3  3.2 -> 3.2
   fd171c52..913598b8  3.3 -> 3.3

PR #1314, opened 2022-06-10
Added condition for Satellite Upgrading to RHEL8

bangelic:
Added conditional information for upgrading Satellite from RHEL 7 to RHEL 8 per SAT-6752. Removed 'No outstanding updates" bullet in Prerequisites. Updated file name and anchor to be consistent.
@evgeni please review at your earliest convenience

Cherry-pick into:

* [X] Foreman 3.3
* [x] Foreman 3.2
* [ ] Foreman 3.1
* For Foreman 3.0 or older, file a separate PR request

PR #1311, opened 2022-06-08
Add upgrade via cloning info

sabuchan07:
Cherry-pick into:

* [ x] Foreman 3.3
* [ ] Foreman 3.2
* [ ] Foreman 3.1
* For Foreman 3.0 or older, file a separate PR request

PR #1350, opened 2022-06-23
Migrating/cloning - upgrading guide

Lennonka:
Cherry-pick into:

* [x] Foreman 3.3
* [x] Foreman 3.2
* [x] Foreman 3.1

### DIFF
--- a/guides/common/modules/con_migrating-project-to-el8.adoc
+++ b/guides/common/modules/con_migrating-project-to-el8.adoc
@@ -1,0 +1,35 @@
+[id="migrating-project-to-el8_{context}"]
+ifdef::satellite[]
+= Migrating {ProjectName} to {RHEL} 8
+endif::[]
+ifndef::satellite[]
+= Migrating {ProjectName} to Enterprise Linux 8
+endif::[]
+
+When you migrate your {Project}, you create a backup of your {ProjectServer} and your {SmartProxy}, install a fresh instance on a new server, and restore your backup on the new server.
+After your migration is complete, you can then decommission the earlier instance of {ProjectServer} and {SmartProxy}.
+
+.Terminology
+Ensure that you understand the following terms:
+
+*Source server*:: The origin of migration on which you create a backup.
+
+*Target server*:: The new server on which you restore the backup.
+
+.High-Level Procedure
+To migrate your {Project} to new hardware, follow these high-level steps:
+
+. Create a backup of the {ProjectServer} or {SmartProxyServer} on the source server.
+. Perform a fresh installation of the {ProjectServer} or {SmartProxyServer} on a target server.
+ifdef::satellite[]
+* Install a minimal {RHEL} 8 instance with the capacity to store backup files.
+endif::[]
+ifndef::satellite[]
+* Install a minimal Enterprise Linux 8 (CentOS Stream, {RHEL} or a {RHEL} rebuild) instance with the capacity to store backup files.
+endif::[]
+* Do not install any operating system software groups or third-party applications.
+ifdef::satellite[]
++
+For more information, see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/performing_a_standard_rhel_installation/index[Performing a standard {RHEL} installation].
+endif::[]
+. Restore the backup on the target server.
--- a/guides/common/modules/proc_creating-a-backup-of-a-server-on-el7.adoc
+++ b/guides/common/modules/proc_creating-a-backup-of-a-server-on-el7.adoc
@@ -1,0 +1,18 @@
+[id="Creating_a_Backup_of_a_Server_on_el7_{context}"]
+ifdef::satellite[]
+= Creating a Backup of a Server on {RHEL} 7
+
+Before you perform a fresh installation of the {ProjectServer} or {SmartProxyServer} on the {RHEL} 8 system, back up your {ProjectServer} or {SmartProxyServer} data on the {RHEL} 7 system by creating an offline backup.
+endif::[]
+ifndef::satellite[]
+= Creating a Backup of a Server on Enterprise Linux 7
+
+Before you perform a fresh installation of the {ProjectServer} or {SmartProxyServer} on the Enterprise Linux 8 system, back up your {ProjectServer} or {SmartProxyServer} data on the Enterprise Linux 7 system by creating an offline backup.
+endif::[]
+
+If you recently created an offline backup, you can perform an incremental backup to update the existing backup.
+
+.Procedure
+* Perform a backup on the source server:
+** To perform a full backup, see {AdministeringDocURL}Backing_Up_Server_and_Proxy_{project-context}[Backing Up {ProjectServer} and {SmartProxyServer}] in _{AdministeringDocTitle}_.
+** To perform an incremental backup, see {AdministeringDocURL}Performing_an_Incremental_Backup_{project-context}[Performing an Incremental Backup] in _{AdministeringDocTitle}_.
--- a/guides/common/modules/ref_performing-a-fresh-installation-of-a-server-on-el8.adoc
+++ b/guides/common/modules/ref_performing-a-fresh-installation-of-a-server-on-el8.adoc
@@ -1,0 +1,15 @@
+[id="Performing_a_Fresh_Installation_of_a_Server_on_el8_{context}"]
+ifdef::satellite[]
+= Performing a Fresh Installation of a Server on {RHEL} 8
+endif::[]
+ifndef::satellite[]
+= Performing a Fresh Installation of a Server on Enterprise Linux 8
+endif::[]
+
+After you have created a backup of the {ProjectServer} or {SmartProxyServer} on the source server, you can install {ProjectServer} or {SmartProxyServer} on the target server.
+
+* To install connected {ProjectServer}, see {InstallingProjectDocURL}[_{InstallingServerDocTitle}_].
+ifdef::satellite[]
+* To install disconnected {ProjectServer}, see {InstallingDisconnectedDocURL}[_{InstallingServerDisconnectedDocTitle}_].
+endif::[]
+* To install {SmartProxyServer}, see {InstallingSmartProxyDocURL}[_{InstallingSmartProxyDocTitle}_].
--- a/guides/common/attributes.adoc
+++ b/guides/common/attributes.adoc
@@ -50,7 +50,6 @@ endif::[]
 :SiteURL: https://docs.theforeman.org
 :BaseURL:  {SiteURL}/{ProjectVersion}/
 include::attributes-base.adoc[]
-include::attributes-titles-base.adoc[]
 
 // Load overrides for specific scenarios
 ifdef::foreman-el[]
@@ -74,3 +73,6 @@ ifdef::orcharhino[]
 :BaseURL: https://docs.orcharhino.com/
 include::attributes-orcharhino.adoc[]
 endif::[]
+
+// Must be loaded last to get attribute values correctly because of ifdefs above
+include::attributes-titles-base.adoc[]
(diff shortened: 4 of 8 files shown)


Lennonka:
Okay, it's failing because the Migrating chapter is conditioned only for Satellite whereas the xref is not. Two options:

1. We'll make the chapter available for other vendors as well.
2. We'll condition the xref for Satellite only.

LMK

evgeni:
I think "improve terms" and "move titles" commits can be their own PR(s), as they seem to be useful outside of this PR and would get in quicker that way?

Lennonka:
@evgeni Please, do not bother me with bureaucracy... I have enough on my plate for the release

evgeni:
> Okay, it's failing because the Migrating chapter is conditioned only for Satellite whereas the xref is not. Two options:
> 
> 1. We'll make the chapter available for other vendors as well.
> 2. We'll condition the xref for Satellite only.
> 
> LMK

In the end (que some Linkin Park 🎶) this chapter will/should be available at least for foreman-el, Katello and Satellite, but the last time I looked at the actual text, it had some Satellite specifics not yet ifdef'ed out (I didn't look in detail at your latest version).

So I guess it depends on how you want to proceed: your 1 means we need to get that chapter in a state that's suitable for all in this pr, your 2 means we can focus on satellite and add a FIXME comment to change that xref once the chapter is ready for foreman/Katello too.

Lennonka:
@evgeni Point absolutely taken. So please check out my text and LMK if it's worth to adapt it for multiple vendors

guides/common/modules/con_migrating-project-to-rhel8.adoc
```diff
@@ -0,0 +1,23 @@
...
+
+.High-Level Procedure
+To migrate your {Project} from {RHEL} 7 to {RHEL} 8 on new hardware, follow these high-level steps:
+
+. Create a backup of the {ProjectServer} or {SmartProxyServer} on the {RHEL} 7 source server.
+. Perform a fresh installation of the {ProjectServer} or {SmartProxyServer} on a target server.
+* Install a minimal {RHEL} 8 instance with the capacity to store backup files.
+* Do not install any {RHEL} 8 software groups or third-party applications.
++
+For more information, see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/performing_a_standard_rhel_installation/index[Performing a standard {RHEL} installation].
```
evgeni, line 22: this link should be `ifdef::satellite`

evgeni:
@Lennonka so I think the biggest difference (and honestly, the only I could spot from a quick read) is the fact that the current procedure always talks about "source on RHEL7" and "target on RHEL8" while for upstream it can be *any* Enterprise Linux 7 (so CentOS, Oracle, RHEL) and any Enterprise Linux 8 (so CentOS Stream 8, RHEL 8, Oracle, Alma, Rocky).

When working on the LEAPP version of the upgrade in https://github.com/theforeman/foreman-documentation/pull/1248/, we essentially settled on: mention once what the different names can be, and in all other places use a generic "Enterprise Linux" naming. If you could incorporate something similar in here too, we can get the `ifdef` widened to `foreman-el,katello,satellite` instead.

guides/common/modules/con_migrating-project-to-rhel8.adoc
```diff
@@ -0,0 +1,23 @@
+[id="migrating-project-to-rhel8_{context}"]
+= Migrating {ProjectName} to {RHEL} 8
```
evgeni, line 2: so, example for what I said earlier:
```suggestion
ifdef::satellite[]
= Migrating {ProjectName} to {RHEL} 8
endif::[]
ifndef::satellite[]
= Migrating {ProjectName} to Enterprise Linux 8
endif::[]
```

guides/common/modules/con_migrating-project-to-rhel8.adoc
```diff
@@ -0,0 +1,23 @@
...
+*Source server*:: The origin of migration on which you create a backup.
+
+*Target server*:: The new server on which you restore the backup.
+
+.High-Level Procedure
+To migrate your {Project} from {RHEL} 7 to {RHEL} 8 on new hardware, follow these high-level steps:
+
+. Create a backup of the {ProjectServer} or {SmartProxyServer} on the {RHEL} 7 source server.
+. Perform a fresh installation of the {ProjectServer} or {SmartProxyServer} on a target server.
+* Install a minimal {RHEL} 8 instance with the capacity to store backup files.
```
evgeni, line 19:
```suggestion
ifdef::satellite[]
* Install a minimal {RHEL} 8 instance with the capacity to store backup files.
endif::[]
ifndef::satellite[]
* Install a minimal Enterprise Linux 8 (CentOS Stream 8, {RHEL} or a {RHEL} rebuild) instance with the capacity to store backup files.
endif::[]
```

guides/common/modules/proc_restoring-a-backup-of-a-server-on-rhel8.adoc
```diff
@@ -0,0 +1,10 @@
+[id="Restoring_a_Backup_of_a_Server_on_rhel8_{context}"]
+= Restoring a Backup of a Server on {RHEL} 8
+
+After you perform a fresh installation of {ProjectServer} or {SmartProxyServer} on the {RHEL} 8 instance of the target machine, you can restore the backup you previously created.
+
+.Procedure
+. Update your network configuration to match the target server’s IP address and new host name.
```
evgeni, line 7: This is something I commented on in https://github.com/theforeman/foreman-documentation/pull/1311#pullrequestreview-1013052395, but did want to make sure it won't get lost.

1. the IP address doesn't need to match, only the host name
2. the fact that the host name needs to match is already mentioned in the linked "Restoring from Backup" documents below

So please drop the "IP address" part. And I am honestly torn whether the repetition of the other prereq is a good thing or not.

Lennonka: I decided to remove the dup.

Lennonka:
This might work :)

maximiliankolb:
How about setting the `RHEL` attribute to "Enterprise Linux" for Foreman and Katello? Maybe add a simple note, for example to the installing & updating guides, to say "We currently support X Y and Z as Enterprise Linux distributions".

This is a bit unrelated to your PR, but IMHO worth while discussing.

evgeni:
> How about setting the `RHEL` attribute to "Enterprise Linux" for Foreman and Katello? Maybe add a simple note, for example to the installing & updating guides, to say "We currently support X Y and Z as Enterprise Linux distributions".
> 
> This is a bit unrelated to your PR, but IMHO worth while discussing.

I thought about this too, but given we actually do differentiate in a few places between "running on RHEL" vs "running on RHEL-compatible" I don't think this would be a good idea per se. The alternative would be a `{EL-or-RHEL}` attribute, that renders as "Enterprise Linux" by default, and as "{RHEL}" for Satellite builds - a bit cumbersome to type, I agree, but would convey the intention better?

Certainly something we should explore, but outside of this PR.

guides/common/modules/con_migrating-project-to-el8.adoc
```diff
@@ -0,0 +1,35 @@
...
+.High-Level Procedure
+To migrate your {Project} to new hardware, follow these high-level steps:
+
+. Create a backup of the {ProjectServer} or {SmartProxyServer} on the source server.
+. Perform a fresh installation of the {ProjectServer} or {SmartProxyServer} on a target server.
+ifdef::satellite[]
+* Install a minimal {RHEL} 8 instance with the capacity to store backup files.
+endif::[]
+ifndef::satellite[]
+* Install a minimal Enterprise Linux 8 (CentOS Stream 8, {RHEL} or a {RHEL} rebuild) instance with the capacity to store backup files.
```
evgeni, line 28:
```suggestion
* Install a minimal Enterprise Linux 8 (CentOS Stream, {RHEL} or a {RHEL} rebuild) instance with the capacity to store backup files.
```

It's weird to say "EL 8 (CentOS Stream 8, RHEL without version, rebuild without version)". I think it's sufficient to have the 8 outside the parenthesis (or you need to say "CentOS Stream 8, RHEL8, RHEL8 rebuild").

guides/common/modules/proc_creating-a-backup-of-a-server-on-el7.adoc
```diff
@@ -0,0 +1,16 @@
+[id="Creating_a_Backup_of_a_Server_on_el7_{context}"]
+ifdef::satellite[]
+= Creating a Backup of a Server on {RHEL} 7
+endif::[]
+ifndef::satellite[]
+= Creating a Backup of a Server on Enterprise Linux 7
+endif::[]
+
+Before you perform a fresh installation of the {ProjectServer} or {SmartProxyServer} on the {RHEL} 8 system, back up your {ProjectServer} or {SmartProxyServer} data on the {RHEL} 7 system by creating an offline backup.
```
evgeni, line 9: This still says `{RHEL}` for non-Satellite.

evgeni (changes requested):
Thanks @Lennonka!

I've left a few (minor) content comments inline and have one last thing to discuss: placement.

When rendered, the docs look like this now:

Upstream:
1. Upgrade Overview
2. Migrating Foreman to Enterprise Linux 8
3. Upgrading Foreman
4. Updating Foreman server and Content Hosts

Satellite:
1. Upgrade Overview
2. Cloning Satellite Server
3. Migrating Red Hat Satellite to Red Hat Enterprise Linux 8
4. Upgrading Red Hat Satellite
5. Updating Satellite Server, Capsule Server, and Content Hosts

Now, seeing this as a user, the current placement would imply I need to perform the EL8 migration *before* I upgrade the system, which is wrong. One needs to first upgrade to a release that supports the shift to EL8 (Foreman 3.1, 3.2 or 3.3 upstream and Satellite 6.11 downstream) and only then perform the migration. So at the minimum I think this should move one step later in `master.adoc`.

However, there is the additional detail that we offer two ways to upgrade the underlying operating system:
1. the backup/restore one, which is what you're describing in this PR
2. the in-place one using LEAPP, which is documented in `guides/common/modules/proc_upgrading-project-from-el7-to-el8.adoc` and is currently (IMHO wrongly) part of the "Upgrading {Project}" chapter.

So I am going to repeat what I said in https://github.com/theforeman/foreman-documentation/pull/1311#pullrequestreview-1010147144 and https://github.com/theforeman/foreman-documentation/pull/1314#pullrequestreview-1003980603 : we need a chapter that first discusses the fact that the migration to EL8 is absolutely necessary and that we provide two different ways to perform that, and *then* include the details of the two ways.

I don't think it *needs* to be in this PR (moving the chapter one down is probably sufficient to make this one make sufficient sense on its own), but it needs to happen before we say "we have documentation how to upgrade to EL8".

evgeni (approved):
I'm :+1: on the changes, thanks so much @Lennonka!

guides/common/modules/ref_performing-a-fresh-installation-of-a-server-on-el8.adoc
```diff
@@ -0,0 +1,15 @@
+[id="Performing_a_Fresh_Installation_of_a_Server_on_el8_{context}"]
+ifdef::satellite[]
+= Performing a Fresh Installation of a Server on {RHEL} 8
```
maximiliankolb, line 3: Is this supposed to be "Server" in the title? Maybe `{ProjectServer}`?

maximiliankolb (approved):
One small suggestion, but LGTM!

guides/common/modules/ref_performing-a-fresh-installation-of-a-server-on-el8.adoc
evgeni: Oh yeah, `ProjectServer` makes a lot of sense here!

Lennonka: Actually, we should keep it generic, because later on, it speaks about both ProjectServer and Proxy Server.

evgeni:
   a7ff0ad2..2531715e  3.1 -> 3.1
   a4564f06..5300cd7e  3.2 -> 3.2
   e759eb4e..246e445c  3.3 -> 3.3

PR #1337, opened 2022-06-21
fix links to guides after renaming

evgeni:
in f2edd6ab45cddb26f57e892f613666b15979721b several guides were renamed,
but the links in our overview page were not updated. this change
corrects that.

Fixes: f2edd6ab45cddb26f57e892f613666b15979721b

Cherry-pick into:

* [ ] Foreman 3.3
* [ ] Foreman 3.2
* [ ] Foreman 3.1
* For Foreman 3.0 or older, file a separate PR request

PR #1352, opened 2022-06-24
deploy PR preview to surge.sh

evgeni:
Cherry-pick into:

* [ ] Foreman 3.3
* [ ] Foreman 3.2
* [ ] Foreman 3.1

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -207,6 +207,48 @@ jobs:
           name: foreman-docs-web-master
           path: web/output/
 
+  preview:
+    if: github.event_name == 'pull_request'
+    needs:
+      - check-html
+      - build-html
+      - build-web
+      - build-ccutil
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get source branch name
+        run: echo "BRANCH_NAME=$(echo ${GITHUB_HEAD_REF} | tr / -)" >> $GITHUB_ENV
+
+      - name: Set target name (master)
+        if: github.base_ref == 'master'
+        run: echo "TARGET_NAME=nightly" >> $GITHUB_ENV
+
+      - name: Set target name (stable branch)
+        if: github.base_ref != 'master'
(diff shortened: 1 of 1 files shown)


maximiliankolb:
This is super cool!

Can you please share some insights on what you did to set it up? What did you do on the side of "surge.sh"? Who has credentials for that project?

evgeni:
"nothing". We (@theforeman) have a shared account on surge.sh (I think it was originally created to host documentation for our JavaScript stack. I just took the API token of that account, added it to the secrets of this repo and configured the preview action (see the changed workflow file).

maximiliankolb:
Awesome, sounds good to me. Is there anything that anyone has to know, i.e. do we need to extend the README, contributing guidelines, or PR template? It's pretty obvious from the message from the bot, isn't it?

evgeni:
I think the only not-totally-obvious thing is: it deploys only the version of the docs, the PR is against. like this one is for `master`, so it only deployed `nightly` docs and all other will 404.

So:
* https://theforeman-foreman-documentation-preview-pr-1352.surge.sh/3.3/Installing_Server/index-foreman-el.html is 404
* https://theforeman-foreman-documentation-preview-pr-1352.surge.sh/nightly/Installing_Server/index-foreman-el.html is working

maximiliankolb:
OK, can we append `/release/nightly/` to the URL the bot posts?

evgeni:
No, I don't think we can adjust the message posted. But we could adjust what is deployed?
Right now we deploy both "web" (which is the "cover page with the links") and "html" which is the actual content.

We could deploy only "html" at the root, then there is no ambiguity, but then you'd not notice if we break the cover page (like #1337)

maximiliankolb:
I think the frontpage of the docs needs some love. So I suggest we merge this as is because it's great. I hope, some day in the future, someone will improve the UI/UX.

maximiliankolb (approved):

ehelms (approved):
Awesomeness

PR #1357, opened 2022-06-25
Improve titles references (issue 954) - Phase 1 on 3.1

Lennonka:
Follow-up to #1230


Lennonka:
I'm merging this one, because we need to fix our internal builds. Emergency :D

PR #1360, opened 2022-06-27
don't try to comment the surge link

evgeni:
Turns out, one can't post comments from an action if the PR is opened from a fork, which breaks the workflow :(

* [x] I am familiar with the [contributing](CONTRIBUTING.md) guidelines.

* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -216,6 +216,9 @@ jobs:
       - build-ccutil
     runs-on: ubuntu-latest
     steps:
+      - name: Install surge
+        run: npm install surge
+
       - name: Get source branch name
         run: echo "BRANCH_NAME=$(echo ${GITHUB_HEAD_REF} | tr / -)" >> $GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


maximiliankolb:
Does this mean we still deploy to surge.sh but don't have the bot comment the URL?

evgeni:
That's correct. It deploys, but doesn't comment.

evgeni:
Now at least the overview page of the actions run has a link? https://github.com/theforeman/foreman-documentation/actions/runs/2571126791

better than nothing.

maximiliankolb (approved):
LGTM!

A assume a lot of colleagues would appreciate it if the preview works for forks as well.

evgeni:
Yeah. I have some ideas how we could bring back the full commenting glory back, but that will need a bit of tinkering and I at least wanted to make it somewhat usable for now.

Thanks for reviewing!

PR #1403, opened 2022-07-05
Cherry-picks from master to 3.1

maximiliankolb:
Some cherry-picks from master to 3.1 that got lost along the way.

I'd appreciate a quick glance cc @evgeni


evgeni (approved):

evgeni:
@maximiliankolb should we also pick the whole "here is a nice and shiny preview" workflow code to the stable branches, so that they also get previews etc? (I can prep that, just lazy)

evgeni:
#1404 #1405 #1406 :)

PR #1408, opened 2022-07-05
Drop httpd prefork tuning instructions

ekohl:
On all non-EL7 the event MPM is used. This means that any tuning for the prefork MPM is unused and ignored. This only affects nightly.


evgeni:
Should this instead tell how to tune event mpm?

evgeni (approved):

ekohl:
After discussion in the platform team it was suggested to prefer the built in tuning profiles. If manual tuning is needed in the future, it should then be written. Ideally speaking users don't have to think about this.

PR #1370, opened 2022-06-27
Leapp upgrading and improvements

Lennonka:
I focused on improvements in the TOC, overall structure and content of upgrade paths.
We intend proper modularization later on.

* [x] I am familiar with the [contributing](CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3

PR #1409, opened 2022-07-05
Leapp upgrading and improvements 3.1

Lennonka:
Follow-up to #1370.

It's not building, we need to fine tune ifdefs, which are probably the reason why it's not building. I could use some help here!


guides/doc-Upgrading_and_Updating/master.adoc
```diff
@@ -5,8 +5,26 @@ include::common/header.adoc[]
...
 
+In this guide, the terms upgrade, update, and migrate have the following meanings:
+
+Upgrading::
+The process of advancing your {ProjectServer} and {SmartProxyServer} installations from a y-stream release to the next, for example {Project} {ProjectVersionPrevious} to {Project} {ProjectVersion}.
+For more information, see xref:upgrading_overview[].
+
+Updating::
+The process of advancing your {ProjectServer} and {SmartProxyServer} installations from a z-stream release to the next, for example {Project} {ProjectVersion}.0 to {Project} {ProjectVersion}.1.
+ifndef::foreman-deb,orcharhino[]
```
evgeni, line 16: In 3.1, `updating_project` seems only to be defined for Satellite, so that helps:
```suggestion
ifdef::satellite[]
```

guides/doc-Upgrading_and_Updating/master.adoc
```diff
@@ -5,8 +5,26 @@ include::common/header.adoc[]
...
+For more information, see xref:updating_project[].
+endif::[]
+
+Migrating::
+The process of moving an existing {Project} installation to a new instance.
+ifdef::foreman-el,katello,satellite[]
+For more information, see xref:migrating-project-to-a-new-el-system_{context}[].
```
evgeni, line 23: It says it can't find `migrating-project-to-a-new-el-system_upgrade-guide`, yet it is exactly what the file should generate in https://github.com/theforeman/foreman-documentation/pull/1409/files#diff-734b146c05567ccc07fb30f4cf7522edd41e20b96bfd8a01256dc695aa78d506R1? I'm lost.

Lennonka: Yep, so am I... :D

evgeni:
I *think* (but didn't find it yet), that there is an open `ifdef::satellite` that makes everything after it being satellite only.

PR #1410, opened 2022-07-05
tuning: Hide Katello-only sections in non-Katello setups

ekohl:
qpid and content serving are Katello-only features.

This probably affects older versions as well, but for now the guide isn't linked anyway and it doesn't affect downstreams so no picks are needed.

* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3


maximiliankolb (approved):
LGTM!

PR #1413, opened 2022-07-05
Changed upgrade command to reflect current version

bangelic:
Step 7 and Step 9 of the Upgrading a connected Satellite section had a command reference that was pointing to the previous target version for 6.10. Updated the command reference and built a preview that shows it properly references 6.11.

* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/attributes-orcharhino.adoc
+++ b/guides/common/attributes-orcharhino.adoc
@@ -21,7 +21,6 @@
 :SmartProxy: orcharhino Proxy
 :SmartProxyServer: orcharhino Proxy
 :TargetVersion: 6.0
-:TargetVersionMaintainUpgrade: 6.0
 :Team: ATIX AG
 :customcontent: custom content
 :customcontenttitle: Custom Content
--- a/guides/common/attributes-satellite.adoc
+++ b/guides/common/attributes-satellite.adoc
@@ -17,7 +17,6 @@
 :ProductVersion: 6.11
 :ProductVersionPrevious: 6.10
 :TargetVersion: {ProductVersion}
-:TargetVersionMaintainUpgrade: {ProductVersion}
 // Add -beta for Beta releases
 :ProductVersionRepoTitle: {ProductVersion}
 
--- a/guides/common/attributes.adoc
+++ b/guides/common/attributes.adoc
@@ -8,7 +8,6 @@
 :KatelloVersion: nightly
 :PulpcoreVersion: 3.18
 :TargetVersion: {ProjectVersion}
-:TargetVersionMaintainUpgrade: {ProjectVersion}
 // The above attribute should point to the GA version number (x.y) for all releases including Beta
 :SatelliteAnsibleVersion: 2.9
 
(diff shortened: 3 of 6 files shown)


maximiliankolb (commented):
LGTM; please reword your commit message and remove the (now unused attribute).

evgeni:
Both `TargetVersionMaintainUpgrade` and `TargetVersion` point at `ProductVersion` for Satellite, so this change - while cleaning up things - results in no visible change in the rendered output (as seen in the empty diff in the preview comment above).

Was the intent to clean up the attributes? Or was a change in the output expected too?

evgeni:
Ah, in the `3.1` branch, the `attributes-satellite.adoc` actually has:

```
guides/common/attributes-satellite.adoc::TargetVersionMaintainUpgrade: {ProductVersion}
guides/common/attributes-satellite.adoc::TargetVersionMaintainUpgrade: {ProductVersionPrevious}
```

And I think *that* is what you wanted to clean up? Drop the line saying `ProductVersionPrevious`?

maximiliankolb:
> 

I only checked for master 🙈

ekohl (commented):
I'm in favor of this, but should the whole attribute be dropped? It isn't used anymore after this patch:
```console
$ rg TargetVersionMaintainUpgrade
20::TargetVersionMaintainUpgrade: {ProductVersion}

guides/common/attributes-orcharhino.adoc
24::TargetVersionMaintainUpgrade: 6.0

guides/common/attributes.adoc
11::TargetVersionMaintainUpgrade: {ProjectVersion}
```

bangelic:
@evgeni I just fund out you and @Lennonka were fixing some attributes and that any bugs I had pertaining to attributes didn't need to be addressed. @Lennonka is this attribute connected to the ones you fixed?

bangelic:
So are we dropping the {TargetVersionMaintainUpgrade} attribute per @ekohl suggestion?

Lennonka:
I'm assuming that TargetVersionMaintainUpgrade was in use because of upgrades to Beta. Since we don't do Beta releases anymore, yes, I believe that we can drop it altogether.
I would appreciate thumbs-up on this reasoning, though. @ekohl @evgeni ?

evgeni:
Yeah, that sounds reasonable and as Ewoud said rather downstream specific

Lennonka:
@bangelic Yes, you can go ahead and drop it :)

bangelic:
Went ahead and dropped the attributes.

maximiliankolb (approved):

ekohl (approved):
Thanks!

evgeni (approved):
I am absolutely happy with the change.

But I repeat what I said before: it will not fix the issue you're trying to fix (satellite docs rendering the upgrade commands as "to 6.10"), that one needs to be fixed in 3.1 branch with a different patch!

bangelic:
> I am absolutely happy with the change.
> 
> But I repeat what I said before: it will not fix the issue you're trying to fix (satellite docs rendering the upgrade commands as "to 6.10"), that one needs to be fixed in 3.1 branch with a different patch!

So cherry picking won't work for 3.1, I assume? Should I checkout of the 3.1 branch and remove the attributes from there and submit a PR?

evgeni:
Yes

bangelic:
A PR was created for addressing this same change in 3.1:
https://github.com/theforeman/foreman-documentation/pull/1422

maximiliankolb:
   756d7a702..96920db07  3.3 -> 3.3
   73d1b5108..2b1b47f21  3.2 -> 3.2
   3.1: see https://github.com/theforeman/foreman-documentation/pull/1422

PR #1433, opened 2022-07-12
use installer-log-file everywhere possible

evgeni:

* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/assembly_installing-server-connected.adoc
+++ b/guides/common/assembly_installing-server-connected.adoc
@@ -25,7 +25,7 @@ Note that the {Project} installation script is based on Puppet, which means that
 ⁠
 To avoid this and determine which future changes apply, use the `--noop` argument when you run the installation script.
 This argument ensures that no actual changes are made.
-Potential changes are written to `/var/log/foreman-installer/{project-context}.log`.
+Potential changes are written to `{installer-log-file}`.
 
 Files are always backed up and so you can revert any unwanted changes.
 For example, in the foreman-installer logs, you can see an entry similar to the following about Filebucket:
--- a/guides/common/modules/proc_configuring-installation.adoc
+++ b/guides/common/modules/proc_configuring-installation.adoc
@@ -9,7 +9,7 @@ ifdef::satellite[]
 The installation process can take tens of minutes to complete.
 If you are connecting remotely to the system, use a utility such as `screen` or `tmux` that allows suspending and reattaching a communication session so that you can check the installation progress in case you become disconnected from the remote system.
 For more information, see https://access.redhat.com/articles/5247[How to use the screen command] or alternately the `screen` manual page.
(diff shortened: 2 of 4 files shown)


guides/common/assembly_installing-server-connected.adoc
```diff
@@ -25,7 +25,7 @@ Note that the {Project} installation script is based on Puppet, which means that
 ⁠
 To avoid this and determine which future changes apply, use the `--noop` argument when you run the installation script.
 This argument ensures that no actual changes are made.
-Potential changes are written to `/var/log/foreman-installer/{project-context}.log`.
+Potential changes are written to `{installer-log-file}`.
```
evgeni, line 28: I wonder whether this is correct for Orcharino? `project-context` is defined as `orcharino` there, but in other places of the docs you use `--scenario katello`.

evgeni: It def changes the rendering from `/var/log/foreman-installer/orcharhino.log` to `/var/log/foreman-installer/foreman.log`, and I bet the later is totally wrong, but is the former right? Or should it be `katello.log`?

cc @maximiliankolb

maximiliankolb: We use both `/var/log/foreman-installer/foreman.log` and `/var/log/foreman-installer/katello.log` with much much more content in the Katello log... 🤷 

Thanks for adding an attribute for this! I think `katello.log` is the right call!

evgeni: *how* do you manage to get both? The installer should be very vocal not to do such things ;)

Or do you have setups with non-content proxies, that are deployed with `foreman-installer --scenario foreman --no-enable-foreman`?

ekohl: Well, if you have the `foreman` scenario present it does run the migrations and that generates logs IIRC, even if you only use the scenario `katello`.

guides/common/modules/proc_configuring-installation.adoc
```diff
@@ -20,7 +20,7 @@ ifdef::foreman-el,katello[]
 For example, on Red Hat-based operating systems, use a utility such as `tmux` or `screen`.
```
ekohl, line 20: Unrelated, but `tmux` and `screen` are not Red Hat specific. The whole sections needs a look. Also, `screen` is no longer in EL8 so also the Satellite-specific parts need some attention.

evgeni:
Yes, I had the same thoughts, but didn't want to blow the scope of the PR ;)

guides/common/assembly_installing-server-connected.adoc
maximiliankolb: I was specifically checking orcharhino Server; not an orcharhino Proxy. To my knowledge, we use scenario `katello`.

PR #1435, opened 2022-07-12
Corrected the log file name in the upgrade guides

dubewarsagar:
Log file name was incorrect in the Upgrading {SmartProxyServer}s guide.
The name has been corrected from satellite.log to capsule.log.
This issue was reported in BZ#2104682

https://bugzilla.redhat.com/show_bug.cgi?id=2104682

* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
+++ b/guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
@@ -123,7 +123,7 @@ Ensure {SmartProxy} has access to `{RepoRHEL7ServerSatelliteMaintenanceProductVe
 . Because of the lengthy upgrade time, use a utility such as `tmux` to suspend and reattach a communication session.
 You can then check the upgrade progress without staying connected to the command shell continuously.
 +
-If you lose connection to the command shell where the upgrade command is running you can see the logged messages in the `{installer-log-file}` file to check if the process completed successfully.
+If you lose connection to the command shell where the upgrade command is running you can see the logged messages in the `{SmartProxy}.log` file to check if the process completed successfully.
 
 . Use the health check option to determine if the system is ready for upgrade:
 +
--- a/guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
+++ b/guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
@@ -123,7 +123,7 @@ Ensure {SmartProxy} has access to `{RepoRHEL7ServerSatelliteMaintenanceProductVe
 . Because of the lengthy upgrade time, use a utility such as `tmux` to suspend and reattach a communication session.
 You can then check the upgrade progress without staying connected to the command shell continuously.
 +
-If you lose connection to the command shell where the upgrade command is running you can see the logged messages in the `{SmartProxy}.log` file to check if the process completed successfully.
+If you lose connection to the command shell where the upgrade command is running you can see the logged messages in the `{installer-smartproxy-log-file}` file to check if the process completed successfully.
 
 . Use the health check option to determine if the system is ready for upgrade:
 +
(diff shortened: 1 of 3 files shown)


guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
```diff
@@ -124,7 +124,7 @@ Ensure {SmartProxy} has access to `{RepoRHEL7ServerSatelliteMaintenanceProductVe
 You can then check the upgrade progress without staying connected to the command shell continuously.
 For more information about using the screen command, see link:https://access.redhat.com/articles/5247[How do I use the screen command?] article in the _Red{nbsp}Hat Knowledge{nbsp}Base_.
 +
-If you lose connection to the command shell where the upgrade command is running you can see the logged messages in the `{installer-log-file}` file to check if the process completed successfully.
+If you lose connection to the command shell where the upgrade command is running you can see the logged messages in the `{SmartProxy}.log` file to check if the process completed successfully.
```
ekohl, line 127: This isn't correct in the upstream context. There (in the Katello context) the file is named `foreman-proxy-content.log`. I think a new attribute would be the best solution.

Note that in the pure Foreman scenario (without Katello) there is no equivalent scenario so `{installer-log-file}` is the closest, but the whole guide is not used there so you can leave it undefined.

ekohl (changes requested):

guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
evgeni: But this section is not rendered at all for upstream?

dubewarsagar: @ekohl @evgeni 

It is only applicable for satellite and there are no impacts for forman or Katello docs, and as per the current update, the final text getting rendered is `capsule.log`. Do we still need to add new variable for this or this is fine?

Also, kindly confirm if the same change is applicable for:
guides/doc-Upgrading_and_Updating/topics/updating_capsule_server_to_next_minor_version.adoc

Thanks!

ekohl: > It is only applicable for satellite and there are no impacts for forman or Katello docs, and as per the current update, the final text getting rendered is `capsule.log`. Do we still need to add new variable for this or this is fine?

I'd ok with this version then

> Also, kindly confirm if the same change is applicable for:
> guides/doc-Upgrading_and_Updating/topics/updating_capsule_server_to_next_minor_version.adoc

Yes it is. Please include it in this PR.

ekohl (commented):
I think something went wrong, because you pushed a lot more commits.

dubewarsagar:
> I think something went wrong, because you pushed a lot more commits.

Yes, checking now.

dubewarsagar:
@ekohl @evgeni

Changes are done now. We can ask to merge into master and close this issue.

ekohl:
Pretty much, just I'd use `/var/log/foreman-installer/capsule.log` instead of a relative path.

dubewarsagar:
> Pretty much, just I'd use `/var/log/foreman-installer/capsule.log` instead of a relative path.

Done. Added a new attribute as suggested with fill path to the log file. 

Thanks.

ekohl (approved):
Verified the diff renders well. Thanks. I'll squash the commits to follow https://cbea.ms/git-commit/ but please keep this in mind in the future.

ekohl:
3c50df248..20353b5c4  3.1 -> 3.1
   db91eb95c..c20ef2aa9  3.2 -> 3.2
   2dc0ab5b0..b2a1868cc  3.3 -> 3.3

PR #1439, opened 2022-07-12
drop reference to screen, only document tmux

evgeni:
screen has been deprecated in EL 7.6 and is not present in EL8+ at all

simplify documentation by only mentioning one tool

* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/doc-Upgrading_and_Updating/topics/upgrading_a_disconnected_satellite.adoc
+++ b/guides/doc-Upgrading_and_Updating/topics/upgrading_a_disconnected_satellite.adoc
@@ -253,9 +253,8 @@ Although entering the `{foreman-installer}` command with the `--noop` option doe
 # {foreman-maintain} service stop
 ----
 
-. Because of the lengthy upgrade time, use a utility such as `screen` to suspend and reattach a communication session.
+. Because of the lengthy upgrade time, use a utility such as `tmux` to suspend and reattach a communication session.
 You can then check the upgrade progress without staying connected to the command shell continuously.
-For more information about using the screen command, see link:https://access.redhat.com/articles/5247[How do I use the screen command?] article in the _Red{nbsp}Hat Knowledge{nbsp}Base_.
 +
 If you lose connection to the command shell where the upgrade command is running you can see the logs in `{installer-log-file}` to check if the process completed successfully.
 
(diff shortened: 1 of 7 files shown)


evgeni:
I found more places where we referenced `screen` and updated all of them to point at `tmux`

ekohl (approved):

guides/doc-Upgrading_and_Updating/topics/upgrading_a_disconnected_satellite.adoc
```diff
@@ -253,9 +253,8 @@ Although entering the `{foreman-installer}` command with the `--noop` option doe
 # {foreman-maintain} service stop
 ----
 
-. Because of the lengthy upgrade time, use a utility such as `screen` to suspend and reattach a communication session.
+. Because of the lengthy upgrade time, use a utility such as `tmux` to suspend and reattach a communication session.
 You can then check the upgrade progress without staying connected to the command shell continuously.
-For more information about using the screen command, see link:https://access.redhat.com/articles/5247[How do I use the screen command?] article in the _Red{nbsp}Hat Knowledge{nbsp}Base_.
 +
 If you lose connection to the command shell where the upgrade command is running you can see the logs in `{installer-log-file}` to check if the process completed successfully.
```
evgeni, line 259: this block repeats in every of the touched files. Do we have a good way to extract it? Put it in a common file and include from here?

ekohl: I thought the same thing, but didn't want to blow up the scope. However, since you asked: we have `include::...` in every `master.doc` (and probably more places). This can also reference to the `common/modules` directory where we already have various shared bits. I'd try that.

maximiliankolb: So the general rule is that you can include snippets in any file because they don't have a heading. If you want, you can do this; but IMHO this PR is also fine as it is for now.

evgeni: I know that we have `include` as a directive, but something tells me we had a rule not to have include in files that get included, or something like that.

evgeni: Ah! Thanks Maximilian!

maximiliankolb: time travelling is real!

@evgeni Yes, you can include a snippet in any file you want. `proc_` and `con_` can only be included in assemblies because they have headings.

PR #1440, opened 2022-07-12
Integrate DHCP/DNS into Smart Proxies

maximiliankolb:

* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/con_managing-dhcp-using-smart-proxy.adoc
+++ b/guides/common/modules/con_managing-dhcp-using-smart-proxy.adoc
@@ -1,0 +1,20 @@
+[id="Managing_DHCP_Using_Smart_Proxy_{context}"]
+= Managing DHCP Using {SmartProxy}
+
+{Project} can integrate with a DHCP service using your {SmartProxy}.
+A {SmartProxy} has multiple DHCP providers that you can use to integrate {Project} with your existing DHCP infrastructure or deploy a new one.
+You can use the DHCP module of {SmartProxy} to query for available IP addresses, add new, and delete existing reservations.
+Note that your {SmartProxy} cannot manage subnet declarations.
+
+.Available DHCP providers
+* `dhcp_infoblox` - For more information, see {ProvisioningDocURL}Using_Infoblox_as_DHCP_and_DNS_Providers_provisioning[Using Infoblox as DHCP and DNS Providers] in _{ProvisioningDocTitle}_.
+* `dhcp_isc` - ISC DHCP server over OMAPI.
+For more information, see {InstallingSmartProxyDocURL}configuring-dns-dhcp-and-tftp_smart-proxy[Configuring DNS, DHCP, and TFTP on {SmartProxyServer}] in _{InstallingSmartProxyDocTitle}_.
+* `dhcp_remote_isc` - ISC DHCP server over OMAPI with leases mounted through networking.
+For more information, see {InstallingSmartProxyDocURL}configuring-external-dhcp_smart-proxy[Configuring an External DHCP Server to Use with {SmartProxyServer}] in _{InstallingSmartProxyDocTitle}_.
+ifndef::satellite[]
+* `dhcp_libvirt` - dnsmasq DHCP via libvirt API
+endif::[]
+ifdef::orcharhino[]
+* `dhcp_native_ms` - Microsoft Active Directory using API
+endif::[]
--- a/guides/common/modules/con_managing-dns-using-smart-proxy.adoc
+++ b/guides/common/modules/con_managing-dns-using-smart-proxy.adoc
@@ -1,0 +1,27 @@
+[id="Managing_DNS_Using_Smart_Proxy_{context}"]
+= Managing DNS Using {SmartProxy}
+
+{Project} can manage DNS records using your {SmartProxy}.
+DNS management contains updating and removing DNS records from existing DNS zones.
+A {SmartProxy} has multiple DNS providers that you can use to integrate {Project} with your existing DNS infrastructure or deploy a new one.
+
+After you have enabled DNS, your {SmartProxy} can manipulate any DNS server that complies with RFC 2136 using the `dns_nsupdate` provider.
+Other providers provide more direct integration, such as `dns_infoblox` for https://www.infoblox.com/[Infoblox].
+
+.Available DNS Providers
+ifdef::orcharhino[]
+* `dns_dnscmd` - Static DNS records in Microsoft Active Directory.
+endif::[]
+* `dns_infoblox` - For more information, see {ProvisioningDocURL}Using_Infoblox_as_DHCP_and_DNS_Providers_provisioning[Using Infoblox as DHCP and DNS Providers] in _{ProvisioningDocTitle}_.
+ifndef::satellite[]
+* `dns_libvirt` - Dnsmasq DNS via libvirt API.
+For more information, see xref:configuring-dynamic-dns-update-with-gss-tsig-authentication_{context}[].
+endif::[]
+* `dns_nsupdate` - Dynamic DNS update using nsupdate.
+For more information, see {ProvisioningDocURL}Using_Infoblox_as_DHCP_and_DNS_Providers_provisioning[Using Infoblox as DHCP and DNS Providers] in _{ProvisioningDocTitle}_.
+* `dns_nsupdate_gss` - Dynamic DNS update with GSS-TSIG.
+For more information, see xref:configuring-dynamic-dns-update-with-gss-tsig-authentication_{context}[].
+
+ifdef::foreman-el,foreman-deb,katello[]
+For more information, see https://projects.theforeman.org/projects/foreman/wiki/List_of_Smart-Proxy_Plugins#DNS-plugins[List of DNS plugins]
+endif::[]
--- a/guides/common/modules/proc_configuring-dns-libvirt.adoc
+++ b/guides/common/modules/proc_configuring-dns-libvirt.adoc
@@ -1,0 +1,19 @@
+[id="Configuring_dns_libvirt_{context}"]
+= Configuring dns_libvirt
+
+The _dns_libvirt_ DNS provider manages IP reservations and leases using dnsmasq through the libvirt API.
+It uses `ruby-libvirt` gem to connect to the local or a remote instance of libvirt daemon.
+
+.Procedure
+* You can use `{foreman-installer}` to configure `dns_libvirt`:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# {foreman-installer} \
+--foreman-proxy-dns true \
+--foreman-proxy-dns-provider libvirt \
+--foreman-proxy-libvirt-network default \
+--foreman-proxy-libvirt-url qemu:///system
+----
++
+Note that you can only use one network and URL for both _dns_libvirt_ and _dhcp_libvirt_.
--- a/guides/common/modules/proc_securing-the-dhcp-api.adoc
+++ b/guides/common/modules/proc_securing-the-dhcp-api.adoc
@@ -1,0 +1,36 @@
+[id="Securing_the_dhcpd_API_{context}"]
+= Securing the dhcpd API
+
+{SmartProxy} interacts with DHCP daemon using the dhcpd API to manage DHCP.
+By default, the dhcpd API listens to any host without access control.
+You can add an `omapi_key` to provide basic security.
+
+.Procedure
+. Install the required packages:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+ifndef::foreman-deb[]
+----
+# {package-install} install bind-utils
+----
+endif::[]
+ifdef::foreman-deb[]
+----
+# {package-install} install bind9-utils
+----
+endif::[]
+. Generate a key:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# dnssec-keygen -r /dev/urandom -a HMAC-MD5 -b 512 -n HOST omapi_key
+# cat Komapi_key.+*.private | grep ^Key|cut -d ' ' -f2-
+----
+. Use `{foreman-installer}` to secure the dhcpd API:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# {foreman-installer} \
+--foreman-proxy-dhcp-key-name "_My_Name_" \
+--foreman-proxy-dhcp-key-secret "_My_Secret_"
+----
--- a/guides/common/modules/ref_dhcp-isc-settings.adoc
+++ b/guides/common/modules/ref_dhcp-isc-settings.adoc
@@ -1,0 +1,43 @@
+[id="DHCP_ISC_Settings_{context}"]
+= dhcp_isc Settings
+
+The _dhcp_isc_ provider uses a combination of the ISC DHCP server OMAPI management interface and parsing of configuration and lease files.
+This requires it to be run on the same host as the DHCP server.
+The following settings are defined in `dhcp_isc.yml`:
+
+.Configuring the path to the _config_ and _leases_ files:
+ifndef::foreman-deb[]
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+:config: /etc/dhcp/dhcpd.conf
+:leases: /var/lib/dhcpd/dhcpd.leases
+----
+endif::[]
+ifdef::foreman-deb[]
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+:dhcp_config: /etc/dhcp3/dhcpd.conf
+:dhcp_leases: /var/lib/dhcp3/dhcpd.leases
+----
+endif::[]
+
+.Securing the DHCP server with an _omapi_key_
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+:key_name: _My_OMAPI_Key_
+:key_secret: _My_Key_Secret_
+----
+
+.Setting a port on which the DHCP server listens
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+:omapi_port: _My_DHCP_Server_Port_  # default: 7911
+----
+
+The server is defined in `dhcp.yml`:
+
+.Setting the host on which the DHCP server runs on
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+:server: _My_DHCP_Server_FQDN_
+----
--- a/guides/common/modules/ref_dhcp-options-for-network-configuration.adoc
+++ b/guides/common/modules/ref_dhcp-options-for-network-configuration.adoc
@@ -27,4 +27,10 @@ Set this to the address for {ProjectServer} on `eth1`.
 --foreman-proxy-dhcp-server::
   Sets the address of the DHCP server to manage.
 
+ifdef::orcharhino[]
+--foreman-proxy-dhcp-subnets::
+  Sets the subnets of the DHCP server to manage.
+Example: `--foreman-proxy-dhcp-subnets 192.168.205.0/255.255.255.128` or `--foreman-proxy-dhcp-subnets 192.168.205.128/255.255.255.128`
+endif::[]
+
 Run `{foreman-installer} --help` to view more options related to DHCP and other {SmartProxy} services.
--- a/guides/doc-Installing_Proxy/master.adoc
+++ b/guides/doc-Installing_Proxy/master.adoc
@@ -25,6 +25,8 @@ include::common/assembly_configuring-external-services.adoc[leveloffset=+1]
 
 include::common/assembly_managing-dhcp-on-smart-proxies.adoc[leveloffset=+1]
 
+include::common/assembly_managing-dns-on-smart-proxies.adoc[leveloffset=+1]
+
 :numbered!:
 
 // {SmartProxyServer} Scalability Considerations
--- a/guides/doc-Installing_Proxy/master.adoc
+++ b/guides/doc-Installing_Proxy/master.adoc
@@ -23,7 +23,17 @@ include::common/assembly_performing-additional-configuration-on-capsule-server.a
 // Configuring {SmartProxyServer} with External Services
 include::common/assembly_configuring-external-services.adoc[leveloffset=+1]
 
+include::common/assembly_managing-dhcp-on-smart-proxies.adoc[leveloffset=+1]
+
 :numbered!:
 
 // {SmartProxyServer} Scalability Considerations
 include::common/modules/ref_capsule-server-scalability-considerations.adoc[leveloffset=+1]
+
+ifndef::satellite[]
+[appendix]
+include::common/modules/ref_dhcp-isc-settings.adoc[leveloffset=+1]
+
+[appendix]
+include::common/modules/ref_dhcp-options-for-network-configuration.adoc[leveloffset=+1]
+endif::[]
--- a/guides/common/assembly_managing-dhcp-on-smart-proxies.adoc
+++ b/guides/common/assembly_managing-dhcp-on-smart-proxies.adoc
@@ -1,0 +1,7 @@
+include::modules/con_managing-dhcp-using-smart-proxy.adoc[]
+
+ifndef::satellite[]
+include::modules/proc_configuring-dhcp-libvirt.adoc[leveloffset=+1]
+endif::[]
+
+include::modules/proc_securing-the-dhcp-api.adoc[leveloffset=+1]
--- a/guides/common/assembly_managing-dns-on-smart-proxies.adoc
+++ b/guides/common/assembly_managing-dns-on-smart-proxies.adoc
@@ -1,0 +1,5 @@
+include::modules/con_managing-dns-using-smart-proxy.adoc[]
+
+ifndef::satellite[]
+include::modules/proc_configuring-dns-libvirt.adoc[leveloffset=+1]
+endif::[]
--- a/guides/common/modules/proc_configuring-dhcp-libvirt.adoc
+++ b/guides/common/modules/proc_configuring-dhcp-libvirt.adoc
@@ -1,0 +1,17 @@
+[id="Configuring_dhcp_libvirt_{context}"]
+= Configuring dhcp_libvirt
+
+The _dhcp_libvirt_ plugin manages IP reservations and leases using `dnsmasq` through the libvirt API.
+It uses `ruby-libvirt` to connect to the local or remote instance of libvirt daemon.
+
+.Procedure
+* You can use `{foreman-installer}` to configure `dhcp_libvirt`:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+foreman-installer \
+--foreman-proxy-dhcp true \
+--foreman-proxy-dhcp-provider libvirt \
+--foreman-proxy-libvirt-network default \
+--foreman-proxy-libvirt-network qemu:///system
+----


maximiliankolb:
See also https://github.com/theforeman/theforeman.org/pull/1916 & https://github.com/theforeman/foreman-documentation/pull/977

guides/common/modules/proc_configuring-dns-on-smart-proxies.adoc
```diff
@@ -0,0 +1,257 @@
...
+
+* `dns_nsupdate` - dynamic DNS update using nsupdate
+* `dns_nsupdate_gss` - dynamic DNS update with GSS-TSIG
+* `dns_libvirt` - dnsmasq DNS via libvirt API
+* `dns_dnscmd` - static DNS records in Microsoft Active Directory
+
+Extra providers are available as plugins and can be installed through packages. See the following pages for more information:
+
+* http://projects.theforeman.org/projects/foreman/wiki/List_of_Smart-Proxy_Plugins[List of smart proxy plugins]
+* link:%7B%7B%20site.baseurl%20%7D%7Dplugins/[Plugin installation documentation]
```
ekohl, line 19: This looks like something went wrong in the conversion.

guides/common/modules/proc_configuring-dns-on-smart-proxies.adoc
```diff
@@ -0,0 +1,257 @@
...
+Builtin providers are:
+
+* `dns_nsupdate` - dynamic DNS update using nsupdate
+* `dns_nsupdate_gss` - dynamic DNS update with GSS-TSIG
+* `dns_libvirt` - dnsmasq DNS via libvirt API
+* `dns_dnscmd` - static DNS records in Microsoft Active Directory
+
+Extra providers are available as plugins and can be installed through packages. See the following pages for more information:
+
+* http://projects.theforeman.org/projects/foreman/wiki/List_of_Smart-Proxy_Plugins[List of smart proxy plugins]
```
ekohl, line 18:
```suggestion
* https://projects.theforeman.org/projects/foreman/wiki/List_of_Smart-Proxy_Plugins[List of smart proxy plugins]
```

ekohl (commented):
Thanks for picking this up again. I do wonder if we should continue to document the manual steps. I like having it documented, but I feel like it should be a reference guide instead. Most users should use the installer instead. AFAIK we already have some guides for that.

guides/common/modules/proc_configuring-dns-on-smart-proxies.adoc
maximiliankolb: Fixed.

guides/common/assembly_performing-additional-configuration-on-capsule-server.adoc
```diff
@@ -27,5 +27,9 @@ include::modules/proc_enabling-power-management-on-managed-hosts.adoc[leveloffse
 // Configuring DNS, DHCP, and TFTP on {SmartProxyServer}
 include::modules/proc_configuring-dns-dhcp-and-tftp.adoc[leveloffset=+1]
 
+include:modules/proc_configuring-dhcp-on-smart-proxies.adoc[leveloffset=+2]
+
+include:modules/proc_configuring-dns-on-smart-proxies.adoc[leveloffset=+2]
```
ekohl, line 32: I still think these are reference guides, not to be consumed by most users since they don't use the installer. We have `guides/common/assembly_using-infoblox-as-dhcp-and-dns-providers.adoc` and `guides/common/modules/proc_configuring-dns-dhcp-and-tftp.adoc` as more generic instructions. Ideally we expand those with the additional providers.

If you want, I can work with you to make that happen.

guides/common/modules/proc_configuring-dhcp-on-smart-proxies.adoc
```diff
@@ -0,0 +1,248 @@
...
+. Optional: Adjust the configuration in the DHCP-provider specific configuration file.
+
+[id="Configuring_dhcp_isc_{context}"]
+= Configuring dhcp_isc
+
+The _dhcp_isc_ provider uses a combination of the ISC DHCP server OMAPI management interface and parsing of configuration and lease files.
+This requires it to be run either on the same host as the DHCP server or to have network filesystem access to these files.
+
+.Procedure
+. Set the path fo the _config_ and _leases_ files in the `dhcp_isc.yml` configuration file.
```
mjivraja, line 54: Possible typo 'fo'

guides/common/modules/proc_configuring-dhcp-on-smart-proxies.adoc
```diff
@@ -0,0 +1,248 @@
...
+
+.Procedure
+* You can configure *dhcp_native_ms* in `dhcp_native_ms.yml`:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+:disable_ddns: true
+----
++
+By default, `disable_ddns` is set to true and dynamic DNS updates are disabled for all hosts that the {SmartProxy} creates.
```
mjivraja, line 107:
```suggestion
For all the hosts that {SmartProxy} creates, “disable_ddns” is set to true, and dynamic DNS updates are disabled by default.
```
Trying to make it more simple.

guides/common/modules/proc_configuring-dns-on-smart-proxies.adoc
```diff
@@ -0,0 +1,267 @@
+[id="Managing_DNS_Using_Smart_Proxies_{context}"]
+= Managing DNS Using {SmartProxies}
+
+Your {SmartProxy} can manage DNS within its network.
+DNS management contains updating and removing DNS records from existing DNS zones.
+
+After your have enabled DNS, your {SmartProxy} can manipulate any DNS server that complies with the ISC Dynamic DNS Update standard.
```
mjivraja, line 7:
```suggestion
After you have enabled DNS, your {SmartProxy} can manipulate any DNS server that complies with the ISC Dynamic DNS Update standard.
```
Possible typo.

guides/common/modules/proc_configuring-dhcp-on-smart-proxies.adoc
maximiliankolb: Fixed!

guides/common/modules/proc_configuring-dhcp-on-smart-proxies.adoc
maximiliankolb: Slightly adjusted and applied.

guides/common/modules/proc_configuring-dns-on-smart-proxies.adoc
maximiliankolb: Fixed!

maximiliankolb (commented):
Applied your feedback and rebased to master.

mjivraja:
> Applied your feedback and rebased to master.

Thank you and done from my side. :+1:

maximiliankolb:
@ekohl Do you have concerns about the content itself or the structure? Before merging, I definitely need to split up the files into "one precedure per file"; then we're much more flexible on how/where to include them.

ekohl (changes requested):
I wanted to read through the preview, but it doesn't render.

As for the content. I still think this is not the right way. We should not be documenting the direct modification of config files in our primary guides. Any time a primary guide tells users to write a YAML file, it's wrong IMHO.

For example, `dhcp_isc.yml` is managed by the installer. Any local changes will be overwritten. https://docs.theforeman.org/nightly/Installing_Proxy/index-katello.html#configuring-dns-dhcp-and-tftp_smart-proxy already covers the default case (which is `dhcp_isc`). For background: the installer has `--foreman-proxy-dhcp-provider`, which defaults to `isc` and results in `:use_provider: dhcp_isc` in `dhcp.yml`.

However, you can use `--foreman-proxy-dhcp-provider native_ms`. Specifically, `native_ms` is not supported by the installer so you would indeed need to manage `dhcp_native_ms.yml` yourself. In my view this means that the `native_ms` provider is unsupported from a documentation perspective. If it would be supported, the installer has options to manage it. Note that there is `--foreman-proxy-dhcp-server` to manage the `server` value in `dhcp.yml`.

Then for `dhcp_libvirt` it would be `--foreman-proxy-dhcp-provider libvirt`. To configure network and url you use `--foreman-proxy-libvirt-network` and `--foreman-proxy-libvirt-connection`.

For DNS it's pretty much the same. There is `--foreman-proxy-dns-provider`, defaulting to `nsupdate`. You can probably predict that it results in `:use_provider: dns_nsupdate` in `dns.yml`.

You may wonder how to find this out. I usually find it easier to look in the Puppet module we use. We have [templates](https://github.com/theforeman/puppet-foreman_proxy/tree/master/templates) and follow the convention where the filenames match. A crude way to translate things is that `foreman_proxy::dns_tsig_principal` translates to the variable `dns_tsig_principal` on the `foreman_proxy` class. The installer exposes class parameters (which are variables but a variable doesn't have to a parameter). So you can read [manifests/init.pp](https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/init.pp). We don't generate a reference (for technical reasons), but in this case you can also read `man foreman-installer` (which is the man page for the `foreman` scenario, faster than `foreman-installer --help`.

I hope this helps a bit. It's not complete, but hopefully it helps.

As for document structure: I don't like that https://docs.theforeman.org/nightly/Installing_Proxy/index-katello.html#configuring-dns-dhcp-and-tftp_smart-proxy is one huge section. Personally I think it would be better to document:
* Configuring DHCP
  * dhcp_isc
  * dhcp_libvirt
  * dhcp_native_ms
* Configuring DNS
  * dns_nsupdate
  * dns_nsupdate_gss
  * dns_libvirt
  * dns_dnscmd
  * dns_infoblox
  * dns_route53
  * dns_powerdns
* Configuring TFTP

The exact providers I mentioned are just examples. I'm not sure they all make sense in our documentation. For example, dns_dnscmd must run on Windows and we don't document Windows support. That's exactly why I think a reference manual in addition to our other manuals makes sense.

guides/common/assembly_performing-additional-configuration-on-capsule-server.adoc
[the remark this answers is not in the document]
maximiliankolb: 🤦 

Thanks, applied!

ekohl:
Since I'm not sure where else to write this, I'll give a quick high level overview.

The [Smart Proxy](https://github.com/theforeman/smart-proxy) is a small webservice that's really a framework. The real functionality is provided by modules. Each module implements what Foreman calls a Feature. If you navigate to the Smart Proxies detail page, you can see features such as `DNS`, `DHCP`, `Logs` and more. A bunch of these modules are shipped inside the `foreman-proxy` package (like `DNS`, `DHCP`), others are provided by plugins (like `Pulp`). The installer abstracts the packaging away for you, but this is important for a complete understanding.

Additionally, modules can have 0 or more providers. Providers are used when there are multiple implementations. For example, the DNS module is always the same: Foreman always uses the same API. Then each implementation is a provider. Providers can be explicitly configured (such as DNS) or programmatically (Puppet CA uses the `puppet_version` setting to determine the implementation). By convention we've named providers by their module: `$module_$provider`. This is only a convention and not hard enforced. Like modules, providers are shipped both in the `foreman-proxy` package and in plugin packages. `dns_nsupdate` is built in while `dns_infoblox` is provided by a separate package.

An example of a module without providers is TFTP. There is just a single implementation and it's inside the module.

As for configuration, modules and plugins can have configuration files. If so, they are named the same. So the `dns` module reads `dns.yml` while the `dns_nsupdate` provider reads `dns_nsupdate.yml`.

The installer abstracts this away from the user, which may make it harder to follow. However, we can let this return in the manual. This is why I suggested the structure in https://github.com/theforeman/foreman-documentation/pull/1440#pullrequestreview-1046221228. Does this help with a more fundamental understanding?

guides/common/assembly_performing-additional-configuration-on-capsule-server.adoc
maximiliankolb: I've split the procedures.

guides/common/modules/proc_configuring-dhcp-native-ms.adoc
```diff
@@ -0,0 +1,16 @@
+[id="Configuring_dhcp_native_ms_{context}"]
+= Configuring dhcp_native_ms
+
+The *dhcp_native_ms* plugin manages IP reservations in Microsoft Active Directory using its native API.
```
ekohl, line 4: Perhaps it's not clear from this, but this must run on Windows. Given we only document Linux I wonder if we should document this.

In an ideal world we would have a guide "Configuring Smart Proxy on Windows" which would document how to run this and dns_dnscmd. Until then I'm leaning to dropping it.

guides/common/modules/proc_configuring-dns-dnscmd.adoc
```diff
@@ -0,0 +1,15 @@
+[id="Configuring_dns_dnscmd_{context}"]
+= Configuring dns_dnscmd
+
+While the `dns_nsupdate` provider creates dynamic records in Active Directory, the `dns_dnscmd` provider uses the `dnscmd` tool to create static DNS records in AD, which are not affected by scavenging.
+This requires that the {SmartProxy} is installed on a Windows server with `dnscmd` available.
```
ekohl, line 5: Since we only document Linux setups today, should we also drop this?

guides/common/modules/proc_configuring-dns-libvirt.adoc
```diff
@@ -0,0 +1,19 @@
...
+. Edit `dns_libvirt.yml`:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# Libvirt network; only one network is supported.
+:network: default
+
+# Libvirt connection; ensure the 'foreman-proxy' user has permissions to connect.
+:url: qemu:///system
+----
```
ekohl, line 17: Similar to `dhcp_libvirt`, we can use the installer:
```
foreman-installer \
  --foreman-proxy-dns true \
  --foreman-proxy-dns-provider libvirt \
  --foreman-proxy-libvirt-network default \
  --foreman-proxy-libvirt-url qemu:///system
```

And yes, you notice that you can only use a single network + URL if you use both dns_libvirt and dhcp_libvirt.

guides/common/modules/proc_configuring-dns-nsupdate-gss.adoc
```diff
@@ -0,0 +1,25 @@
+[id="Configuring_dns_nsupdate_gss_{context}"]
```
ekohl, line 1: Doesn't this duplicate `guides/common/modules/proc_configuring-dynamic-dns-update-with-gss-tsig-authentication.adoc` (which itself needs a thorough readthrough)?

guides/common/modules/proc_configuring-dns-nsupdate.adoc
```diff
@@ -0,0 +1,25 @@
+[id="Configuring_dns_nsupdate_{context}"]
```
ekohl, line 1: I'd say this is already covered in the existing guide.

guides/common/modules/proc_configuring-isc-dhcp.adoc
```diff
@@ -0,0 +1,51 @@
+[id="Configuring_isc_dhcp_{context}"]
```
ekohl, line 1: This should be covered by the installer and the existing guides. I don't think we should document manual installation in the primary guide.

I've been wondering if we should document the reference bits on https://docs.theforeman.org (similar to how today it's on theforeman.org) or rather in [smart-proxy.git](https://github.com/theforeman/smart-proxy). The latter makes sense to me because it's more focused on developers. You can easily apply the same in a development setup.

guides/common/modules/proc_configuring-ms-dhcp.adoc
```diff
@@ -0,0 +1,34 @@
+[id="Configuring_MS_DHCP_{context}"]
```
ekohl, line 1: As noted before: do we want to document Windows? Right now I'm leaning to moving it over to smart-proxy.git.

guides/common/modules/proc_enabling-the-dhcp-module.adoc
```diff
@@ -0,0 +1,28 @@
+[id="Enabling_the_DHCP_Module_{context}"]
```
ekohl, line 1: I think this should be covered by our other guides already, by relying on the installer.

guides/common/modules/proc_enabling-the-dhcp-module.adoc
```diff
@@ -0,0 +1,28 @@
...
+Note that if the DHCP server is running remotely, some providers (notably ISC) require that the configuration files must be accessible to the {SmartProxyServer}.
+You can use network file systems to achive this.
+. Optional: By default, you cna manage all available subnets.
+To mitigate the performance penalty from loading all available subnets, restrict DHCP to a list of subnets to optimize performace in `_My_Network_Address_/_My_Network_Mask_` notation:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+:subnets: [192.168.205.0/255.255.255.128, 192.168.205.128/255.255.255.128]
+----
+. Optional: Adjust the configuration in the DHCP-provider specific configuration file.
```
ekohl, line 28: Interestingly enough `guides/common/modules/proc_configuring-satellite-deployment-with-an-external-dhcp-server.adoc` doesn't document `--foreman-proxy-dhcp-subnets`. The equivalent of above would be:
```
--foreman-proxy-dhcp-subnets 192.168.205.0/255.255.255.128 \
--foreman-proxy-dhcp-subnets 192.168.205.128/255.255.255.128
```

I'm also very big on using RFC5735 IP ranges in documentation. See https://community.theforeman.org/t/using-ips-and-domains-in-tests-documentation/16600

guides/common/modules/proc_securing-the-dhcp-api.adoc
```diff
@@ -0,0 +1,31 @@
...
+
+By default, the dhcpd API listens to any host.
+You can add a `omapi_key` to provide basic security.
+
+.Procedure
+. Generate a key:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# yum install -y bind
```
ekohl, line 12: At least on EL8 you need bind-utils, not bind for `dnssec-keygen`. IIRC it's different on Debian.

guides/common/modules/proc_securing-the-dhcp-api.adoc
```diff
@@ -0,0 +1,31 @@
...
+algorithm HMAC-MD5;
+  secret "XXXXXXXXX"; #<-The output from the generated key above.
+};
+omapi-key omapi_key;
+----
+. Ensure to add the omapi_key to `dhcp_isc.yml` on your {SmartProxyServer}.
+. Restart the `dhcpd` and `foreman-proxy` services.
+. In the {ProjectWebUI}, navigate to *Infrastructure > Smart Proxies*.
+. Select your {SmartProxy}.
+. On the *Overview* tab, click *Refresh* in the *Actions* menu.
```
ekohl, line 31: The proper way to do this is using the installer:
```
foreman-installer \
  --foreman-proxy-dhcp-key-name "myname" \
  --foreman-proxy-dhcp-key-secret "mysecret"
```

We don't do this by default (because generating a secret and parsing that was hard) but surprisingly enough we don't document it either.

guides/common/modules/proc_using-gss-tsig-dns.adoc
```diff
@@ -0,0 +1,98 @@
+[id="Using_GSS_TSIG_DNS_{context}"]
```
ekohl, line 1: This duplicates the `guides/common/modules/proc_configuring-dynamic-dns-update-with-gss-tsig-authentication.adoc` guide.

ekohl (commented):
Thanks for working on this. I did a first pass on this.

guides/common/modules/proc_configuring-dns-on-smart-proxies.adoc
```diff
@@ -0,0 +1,22 @@
+[id="Configuring_DNS_on_Smart_Proxies_{context}"]
```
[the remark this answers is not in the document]
maximiliankolb: Removed as duplicate.

ekohl:
The build looks red.

guides/common/modules/proc_configuring-isc-dhcp.adoc
maximiliankolb: Removed as duplicate.

guides/common/modules/proc_configuring-ms-dhcp.adoc
maximiliankolb: Moved for now; but as said before, I understand if we decide not to add this to the official Foreman documentation! (adding this is most certainly not my own idea...)

guides/common/modules/proc_using-gss-tsig-dns.adoc
maximiliankolb: Removed as duplicate.

guides/common/modules/proc_enabling-the-dhcp-module.adoc
maximiliankolb: Removed as duplicate.

guides/common/modules/proc_enabling-the-dhcp-module.adoc
maximiliankolb: Added to "DHCP Options for Network Configuration" reference.

guides/common/modules/proc_securing-the-dhcp-api.adoc
maximiliankolb: Fixed.

guides/common/modules/proc_securing-the-dhcp-api.adoc
maximiliankolb: Applied!

maximiliankolb:
Thank you so much @ekohl I really appreciate your review. I've tried to incorporate your feedback as much as possible.

maximiliankolb:
How can I proceed with my PR? Any other input? Is this good to merge?

ekohl:
I'll be honest and say that I haven't had time to read through the last update. Right now I'm working on getting the installer modules released so we can branch 3.4. However, it looks like I need to work on our CI nodes first because they have started to fail.

maximiliankolb:
Thanks for all the input so far! Do you have a rough timeline or would you like me to look for someone else to do a final review? content wise, not much/nothing has changed. I only restructured procedures to separate files.

maximiliankolb:
How can I proceed? Are there any objections to merging this? Can someone please review my PR?

guides/common/modules/proc_securing-the-dhcp-api.adoc
```diff
@@ -0,0 +1,26 @@
...
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# {foreman-installer} \
+--foreman-proxy-dhcp-key-name "_My_Name_" \
+--foreman-proxy-dhcp-key-secret "_My_Secret_"
+----
+. In the {ProjectWebUI}, navigate to *Infrastructure > Smart Proxies*.
+. Select your {SmartProxy}.
+. On the *Overview* tab, click *Refresh* in the *Actions* menu.
```
ekohl, line 26: The installer will refresh if needed. In this particular case it's not even needed since it only affects dhcpd.
```suggestion
```

guides/common/modules/ref_dhcp-options-for-network-configuration.adoc
```diff
@@ -27,4 +27,8 @@ Set this to the address for {ProjectServer} on `eth1`.
 --foreman-proxy-dhcp-server::
   Sets the address of the DHCP server to manage.
 
+--foreman-proxy-dhcp-subnets::
+  Sets the subnets of the DHCP server to manage.
```
ekohl, line 31: Should this mention that by default all subnets are managed? This helps to speed up things on large DHCP servers with many subnets, but the installer by default only supports a single subnet so I'm not sure this adds anything for most users. It'll only confuse them.

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
```diff
@@ -0,0 +1,17 @@
...
+ifdef::foreman-el,foreman-deb,katello[]
+For more information on managing subnet declarations, see https://github.com/theforeman/puppet-dhcp[puppet-dhcp].
+endif::[]
+
+The DHCP module can manage various available DHCP providers:
+
+* `dhcp_infoblox` - Infoblox DHCP server
+* `dhcp_isc` - ISC DHCP server over OMAPI
+* `dhcp_remote_isc` - ISC DHCP server over OMAPI with leases mounted through networking
+* `dhcp_libvirt` - dnsmasq DHCP via libvirt API
```
ekohl, line 16:
```suggestion
ifndef::satellite[]
* `dhcp_libvirt` - dnsmasq DHCP via libvirt API
endif::[]
```

guides/common/assembly_managing-dhcp-and-dns-using-smart-proxies.adoc
```diff
@@ -0,0 +1,11 @@
+include::modules/con_managing-dhcp-using-smart-proxies.adoc[]
+
+include::modules/proc_configuring-dhcp-libvirt.adoc[leveloffset=+1]
```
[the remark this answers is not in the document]
maximiliankolb: Applied.

guides/common/assembly_managing-dhcp-and-dns-using-smart-proxies.adoc
```diff
@@ -0,0 +1,11 @@
...
+
+include::modules/proc_securing-the-dhcp-api.adoc[leveloffset=+1]
+
+include::modules/ref_dhcp-isc-settings.adoc[leveloffset=+1]
```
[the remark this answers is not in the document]
maximiliankolb: It's a reference file, but I've moved it to the appendix.

guides/common/modules/con_managing-dns-using-smart-proxies.adoc
```diff
@@ -0,0 +1,19 @@
...
+= Managing DNS Using {SmartProxies}
+
+Your {SmartProxy} can manage DNS records.
+DNS management contains updating and removing DNS records from existing DNS zones.
+
+After you have enabled DNS, your {SmartProxy} can manipulate any DNS server that complies with RFC 2136.
+You can use it to manage both Microsoft Active Directory and BIND servers.
+
+.Available DNS Providers
+* `dns_dnscmd` - static DNS records in Microsoft Active Directory
```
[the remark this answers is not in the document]
maximiliankolb: Made orcharhino-only.

guides/common/modules/con_managing-dns-using-smart-proxies.adoc
```diff
@@ -0,0 +1,19 @@
...
+* `dns_infoblox` - For more information, see {ProvisioningDocURL}Using_Infoblox_as_DHCP_and_DNS_Providers_provisioning[Using Infoblox as DHCP and DNS Providers] in _{ProvisioningDocTitle}_.
+* `dns_libvirt` - dnsmasq DNS via libvirt API
```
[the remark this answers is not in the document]
maximiliankolb: Applied!

guides/common/modules/proc_securing-the-dhcp-api.adoc
```diff
@@ -0,0 +1,26 @@
+[id="Securing_the_dhcpd_API_{context}"]
+= Securing the dhcpd API
+
+By default, the dhcpd API listens to any host.
+You can add a `omapi_key` to provide basic security.
+
+.Procedure
+. On Enterprise Linux 8, generate a key:
```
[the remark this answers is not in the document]
maximiliankolb: Applied.

guides/common/modules/proc_securing-the-dhcp-api.adoc
maximiliankolb: Removed.

guides/common/modules/ref_dhcp-options-for-network-configuration.adoc
maximiliankolb: I don't really know. Maybe someone else can weigh in?

guides/doc-Installing_Proxy/master.adoc
```diff
@@ -23,6 +23,10 @@ include::common/assembly_performing-additional-configuration-on-capsule-server.a
 // Configuring {SmartProxyServer} with External Services
 include::common/assembly_configuring-external-services.adoc[leveloffset=+1]
 
+include::common/assembly_managing-dhcp-and-dns-using-smart-proxies.adoc[leveloffset=+1]
+
+include::common/assembly_configuring-smart-proxy-on-microsoft-windows.adoc[leveloffset=+1]
```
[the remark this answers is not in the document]
maximiliankolb: Made orcharhinon only.

maximiliankolb (commented):
Please have a look at my last commit.

guides/common/modules/proc_securing-the-dhcp-api.adoc
```diff
@@ -5,11 +5,23 @@ By default, the dhcpd API listens to any host.
 You can add a `omapi_key` to provide basic security.
 
 .Procedure
-. On Enterprise Linux 8, generate a key:
+. Install the required packages:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+ifndef::foreman-deb[]
+----
+# dnf install bind-utils
```
ekohl, line 13: Would this work?
```suggestion
# {package-install} bind-utils
```

guides/common/modules/proc_securing-the-dhcp-api.adoc
```diff
@@ -0,0 +1,35 @@
...
+----
+endif::[]
+ifdef::foreman-deb[]
+----
+# apt-get install bind9-utils
```
ekohl, line 18:
```suggestion
# {package-install} bind9-utils
```

guides/common/modules/proc_securing-the-dhcp-api.adoc
maximiliankolb: Applied.

guides/common/modules/proc_securing-the-dhcp-api.adoc
maximiliankolb: Applied.

maximiliankolb (commented):
Thanks again! Applied your suggestions.

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
```diff
@@ -0,0 +1,16 @@
+[id="Managing_DHCP_Using_Smart_Proxies_{context}"]
+= Managing DHCP Using {SmartProxies}
+
+{SmartProxies} can manage DHCP within their network.
+You can use the DHCP modules for {SmartProxy} to query for available IP addresses, add new, and delete existing reservations.
+Note that your {SmartProxy} cannot manage subnet declarations.
+ifdef::foreman-el,foreman-deb,katello[]
+For more information on managing subnet declarations, see https://github.com/theforeman/puppet-dhcp[puppet-dhcp].
+endif::[]
```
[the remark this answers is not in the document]
ekohl: This ifdef block really belongs in a part of the manual, if any at all. Mostly because we don't have instructions about how to map the DHCP options in that module to what you can use in the installer. Right now I'm leaning to dropping it

ekohl (commented):
Structure wise this becomes chapter 5 and 6, but we also have chapter 3.3 which does the combined thing. Should become chapters 3.4(.z) and 3.5(.z) instead?

Structure wise there are some choices you can make.

First of all, go with feature as top level, as you did now. So
* DHCP
  * dhcp_libvirt
  * dhcp_isc
    * Securing the dhcpd api
  * dhcp_remote_isc
    * Set up external server
    * Configure Foreman server
  * ..
* DNS
  * dns_nsupdate
    * local DNS server
    * remote DNS server

The other is what's in the current guide, which focuses more on the scenario.

I like your new style, but it brings confusion to have both styles (as the PR currently does).

@Lennonka any thoughts on this?

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
maximiliankolb: Applied for libirt.

dhcp_native_ms is orcharhino only:
```
git grep -i -n -C 1 assembly_configuring-smart-proxy-on-microsoft-windows.adoc
guides/doc-Installing_Proxy/master.adoc-28-ifdef::orcharhino[]
guides/doc-Installing_Proxy/master.adoc:29:include::common/assembly_configuring-smart-proxy-on-microsoft-windows.adoc[leveloffset=+1]
guides/doc-Installing_Proxy/master.adoc-30-endif::[]
```

guides/common/modules/ref_dhcp-options-for-network-configuration.adoc
Lennonka: @ekohl What is the purpose of this argument in the first place? What is the scenario in which the user would prefer to use it?

ekohl: Here is the example:
https://github.com/theforeman/smart-proxy/blob/9268440e7ea33d3fd0cae22896c403338e049cb9/config/settings.d/dhcp.yml.example#L11-L12
It mostly benefits the `dhcp_native_ms` provider (which the installer doesn't manage). I'm looking through the code, but can't find the code that actually ensures the whole setting is used.

I'd say that for most, if not all, users of the installer this parameter is not needed. Documenting it here will confuse users and make them do additional work. My suggestion is to leave it undocumented.

maximiliankolb: Maybe I can just put it behind `ifdef::orcharhino[]`?

maximiliankolb:
Rebased to master. I don't really have a strong opinion on the structure. If you have a good idea, please let me know. Alternatively, we can reorder content later on in a separate PR.

ekohl:
IMHO the structure is important if we cherry pick this, in particular to 3.1. This now has effects on the Satellite documentation

maximiliankolb:
@ekohl Your suggestion is to not create a new assembly, but rather include the content, where appropriate, in existing assemblies, isn't it?

guides/common/modules/ref_dhcp-options-for-network-configuration.adoc
ekohl: What is the benefit of including it in the Orcharhino manual? I'm quite sure it only affects `dhcp_native_ms` and that requires running on Windows. Has Orcharhino ported the installer to Windows?

I never looked at this before, but now as an installer maintainer I'm tempted to remove the option.

guides/common/assembly_managing-dhcp-and-dns-using-smart-proxies.adoc
```diff
@@ -0,0 +1,11 @@
...
+
+ifndef::satellite[]
+include::modules/proc_configuring-dhcp-libvirt.adoc[leveloffset=+1]
+endif::[]
+
+include::modules/proc_securing-the-dhcp-api.adoc[leveloffset=+1]
+
+include::modules/con_managing-dns-using-smart-proxies.adoc[]
+
+include::modules/proc_configuring-dns-libvirt.adoc[leveloffset=+1]
```
ekohl, line 11:
```suggestion
ifndef::satellite[]
include::modules/proc_configuring-dns-libvirt.adoc[leveloffset=+1]
endif::[]
```

ekohl:
I'd create an assembly per Smart Proxy feature with a provider. So you'd get the following assemblies:

* Preparing your Environment for Installation
* Installing `{SmartProxyServer}`
* Performing Additional Configuration on `{SmartProxyServer}` (but (possibly later) remove the DNS/DHCP part; TFTP should become its own chapter just like OpenSCAP and Power Management)
* Managing DHCP on `{SmartProxyServer}`
* Managing DNS on `{SmartProxyServer}`
* `{SmartProxyServer}` Scalability Considerations

So in short: if you see an installer argument `--foreman-proxy-$feature-provider` it should become its own assembly where each provider is a chapter(s). Simple features without providers belong under Additional Configuration.

Does that make sense? I know I'm suggesting to restructure a quite fundamental document so I'd like some buy in from actual writers.

guides/common/assembly_managing-dhcp-and-dns-using-smart-proxies.adoc
maximiliankolb: Applied.

maximiliankolb:
Restructured everything. Please see the last two commits.

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
```diff
@@ -0,0 +1,21 @@
...
+{SmartProxies} can manage DHCP within their network.
+You can use the DHCP module of {SmartProxy} to query for available IP addresses, add new, and delete existing reservations.
+Note that your {SmartProxy} cannot manage subnet declarations.
+ifdef::foreman-el,foreman-deb,katello[]
+For more information on managing subnet declarations, see https://github.com/theforeman/puppet-dhcp[puppet-dhcp].
+endif::[]
+
+The DHCP module can manage various available DHCP providers:
+
+* `dhcp_infoblox` - Infoblox DHCP server
```
ekohl, line 13: I like what you did with with DNS and link to the "using ..." part. Could you do that here too?

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
```diff
@@ -0,0 +1,21 @@
...
+* `dhcp_isc` - ISC DHCP server over OMAPI
+* `dhcp_remote_isc` - ISC DHCP server over OMAPI with leases mounted through networking
```
ekohl, line 15: This should link to the "Configuring ... with external DHCP" section since that documents it now.

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
```diff
@@ -0,0 +1,21 @@
...
+You can use the DHCP module of {SmartProxy} to query for available IP addresses, add new, and delete existing reservations.
+Note that your {SmartProxy} cannot manage subnet declarations.
+ifdef::foreman-el,foreman-deb,katello[]
+For more information on managing subnet declarations, see https://github.com/theforeman/puppet-dhcp[puppet-dhcp].
+endif::[]
+
+The DHCP module can manage various available DHCP providers:
+
+* `dhcp_infoblox` - Infoblox DHCP server
+* `dhcp_isc` - ISC DHCP server over OMAPI
```
ekohl, line 14: This should link to "Configuring DNS, DHCP, and TFTP on ..."

guides/common/modules/con_managing-dns-using-smart-proxies.adoc
```diff
@@ -0,0 +1,23 @@
...
+.Available DNS Providers
+ifdef::orcharhino[]
+* `dns_dnscmd` - static DNS records in Microsoft Active Directory
+endif::[]
+* `dns_infoblox` - For more information, see {ProvisioningDocURL}Using_Infoblox_as_DHCP_and_DNS_Providers_provisioning[Using Infoblox as DHCP and DNS Providers] in _{ProvisioningDocTitle}_.
+ifndef::satellite[]
+* `dns_libvirt` - dnsmasq DNS via libvirt API
+endif::[]
+* `dns_nsupdate_gss` - dynamic DNS update with GSS-TSIG
+* `dns_nsupdate` - dynamic DNS update using nsupdate
```
ekohl, line 19: Can you swap these?
```suggestion
* `dns_nsupdate` - dynamic DNS update using nsupdate
* `dns_nsupdate_gss` - dynamic DNS update with GSS-TSIG
```

guides/common/modules/con_managing-dns-using-smart-proxies.adoc
```diff
@@ -0,0 +1,23 @@
+[id="Managing_DNS_Using_Smart_Proxies_{context}"]
+= Managing DNS Using {SmartProxies}
+
+Your {SmartProxy} can manage DNS records.
+DNS management contains updating and removing DNS records from existing DNS zones.
+
+After you have enabled DNS, your {SmartProxy} can manipulate any DNS server that complies with RFC 2136.
+You can use it to manage both Microsoft Active Directory and BIND servers.
```
ekohl, line 8: This specifically talks about `dns_nsupdate` and `dns_nsupdate_gss`. Would it make sense to move it there?

guides/common/modules/con_managing-dns-using-smart-proxies.adoc
```diff
@@ -0,0 +1,23 @@
+[id="Managing_DNS_Using_Smart_Proxies_{context}"]
+= Managing DNS Using {SmartProxies}
+
+Your {SmartProxy} can manage DNS records.
+DNS management contains updating and removing DNS records from existing DNS zones.
```
ekohl, line 5: I was thinking about a generic bit of text. Something like:
```asciidoc
{Project} can manage DNS records using your {SmartProxy}.
DNS management contains updating and removing DNS records from existing DNS zones.
The {SmartProxy} has multiple providers, allowing you to integrate with your existing DNS infrastructure or deploy a new one.
```

A similar text can be used for the DHCP introduction.

ekohl (commented):
I noticed that the last link (currently an appendix) in the navigation is barely readable. Not limited to this guide, but this requires some CSS changes in our navigation.

On this PR: I like where it's going. The listing should be a bit more actionable, especially since some providers are documented somewhere else. See inline comments.

guides/common/modules/ref_dhcp-options-for-network-configuration.adoc
Lennonka: Unless any product handles Windows, I'd go with undocumented.

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
maximiliankolb: Applied!

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
maximiliankolb: Applied.

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
maximiliankolb: Applied.

guides/common/modules/con_managing-dns-using-smart-proxies.adoc
maximiliankolb: Applied and added.

guides/common/modules/con_managing-dns-using-smart-proxies.adoc
maximiliankolb: moved.

guides/common/modules/con_managing-dns-using-smart-proxies.adoc
maximiliankolb: Swapped.

maximiliankolb (commented):
Rebased to master and applied all feedback possible in the last commit. (needs to be squashed before merging).

maximiliankolb:
Rebased to master and removed anything related to SmartProxy on Windows. Note that this branch needs to be squashed before merging.

Please re-review.

guides/common/modules/con_managing-dns-using-smart-proxies.adoc
```diff
@@ -0,0 +1,23 @@
+[id="Managing_DNS_Using_Smart_Proxies_{context}"]
+= Managing DNS Using {SmartProxies}
+
+{Project} can manage DNS records using your {SmartProxy}.
+DNS management contains updating and removing DNS records from existing DNS zones.
+Your {SmartProxy} can use multiple providers, allowing you to integrate with your existing DNS infrastructure or deploy a new one.
+
+After you have enabled DNS, your {SmartProxy} can manipulate any DNS server that complies with RFC 2136.
```
ekohl, line 8: This is only with the default provider.

background: [nsupdate](https://en.wikipedia.org/wiki/Nsupdate) is the tool used to implement the client side of [RFC 2136](https://datatracker.ietf.org/doc/html/rfc2136). That's why the provider is named `dns_nsupdate`.

Probably not the ideal text, but have you considered something like:
```suggestion
After you have enabled DNS, your {SmartProxy} can manipulate any DNS server that complies with RFC 2136 using the `dns_nsupdate` provider.
Other providers provide more direct integration, such as `dns_infoblox` for [Infoblox](https://www.infoblox.com/).
```

This would tie in to the last sentence of the previous paragraph.

guides/common/modules/con_managing-dns-using-smart-proxies.adoc
```diff
@@ -0,0 +1,23 @@
...
+.Available DNS Providers
+ifdef::orcharhino[]
+* `dns_dnscmd` - static DNS records in Microsoft Active Directory
+endif::[]
+* `dns_infoblox` - For more information, see {ProvisioningDocURL}Using_Infoblox_as_DHCP_and_DNS_Providers_provisioning[Using Infoblox as DHCP and DNS Providers] in _{ProvisioningDocTitle}_.
+ifndef::satellite[]
+* `dns_libvirt` - dnsmasq DNS via libvirt API
+endif::[]
+* `dns_nsupdate` - dynamic DNS update using nsupdate
+* `dns_nsupdate_gss` - dynamic DNS update with GSS-TSIG
```
ekohl, line 19: This provider is documented in `guides/common/modules/proc_configuring-dynamic-dns-update-with-gss-tsig-authentication.adoc`. Would be good to link back.

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
```diff
@@ -0,0 +1,24 @@
...
+ifdef::foreman-el,foreman-deb,katello[]
+For more information on managing subnet declarations, see https://github.com/theforeman/puppet-dhcp[puppet-dhcp].
+endif::[]
+
+The DHCP module can manage various available DHCP providers:
+
+* `dhcp_infoblox` - For more information, see {ProvisioningDocURL}Using_Infoblox_as_DHCP_and_DNS_Providers_provisioning[Using Infoblox as DHCP and DNS Providers] in _{ProvisioningDocTitle}_.
+* `dhcp_isc` - ISC DHCP server over OMAPI.
+For more information, see {InstallingSmartProxyDocURL}configuring-dns-dhcp-and-tftp_smart-proxy[Configuring DNS, DHCP, and TFTP on {SmartProxyServer}] in _{InstallingSmartProxyDocTitle}_.
+* `dhcp_remote_isc` - ISC DHCP server over OMAPI with leases mounted through networking
```
ekohl, line 17:
```suggestion
* `dhcp_remote_isc` - ISC DHCP server over OMAPI with leases mounted through networking.
```

guides/common/modules/proc_configuring-dns-libvirt.adoc
```diff
@@ -0,0 +1,19 @@
+[id="Configuring_dns_libvirt_{context}"]
+= Configuring dns_libvirt
+
+The _dns_libvirt_ DNS provider manages IP reservations and leases using dnsmasq through the libvirt API.
+It uses `ruby-libvirt` gem to connect to the local or remote instance of libvirt daemon.
```
ekohl, line 5:
```suggestion
It uses `ruby-libvirt` gem to connect to the local or a remote instance of libvirt daemon.
```

guides/common/modules/ref_dhcp-options-for-network-configuration.adoc
```diff
@@ -27,4 +27,8 @@ Set this to the address for {ProjectServer} on `eth1`.
 --foreman-proxy-dhcp-server::
```
ekohl, line 27: I think this whole reference section needs redesign. It all only applies to the `dhcp_nsupdate` provider (except `--foreman-proxy-dhcp`), but the section doesn't mention that at all. Not a blocker for this PR, but this PR opens the path to make that sensible.

guides/doc-Installing_Proxy/master.adoc
```diff
@@ -23,7 +23,17 @@ include::common/assembly_performing-additional-configuration-on-capsule-server.a
...
 :numbered!:
 
 // {SmartProxyServer} Scalability Considerations
 include::common/modules/ref_capsule-server-scalability-considerations.adoc[leveloffset=+1]
+
+[appendix]
+include::common/modules/ref_dhcp-isc-settings.adoc[leveloffset=+1]
+
+[appendix]
+include::common/modules/ref_dhcp-options-for-network-configuration.adoc[leveloffset=+1]
```
ekohl, line 39: I'm guessing this shouldn't be shown for satellite
```suggestion
ifndef::satellite[]
[appendix]
include::common/modules/ref_dhcp-isc-settings.adoc[leveloffset=+1]

[appendix]
include::common/modules/ref_dhcp-options-for-network-configuration.adoc[leveloffset=+1]
endif::[]
```

guides/doc-Installing_Proxy/master.adoc
```diff
@@ -23,7 +23,17 @@ include::common/assembly_performing-additional-configuration-on-capsule-server.a
 // Configuring {SmartProxyServer} with External Services
 include::common/assembly_configuring-external-services.adoc[leveloffset=+1]
 
+include::common/assembly_managing-dhcp-on-smart-proxies.adoc[leveloffset=+1]
+
+include::common/assembly_managing-dns-on-smart-proxies.adoc[leveloffset=+1]
```
ekohl, line 28: Technically these options all also apply to the `doc-Installing_Server` side. I'd be OK with doing that in a follow up, but it's something to think about.

guides/common/modules/con_managing-dns-using-smart-proxies.adoc
```diff
@@ -0,0 +1,23 @@
...
+DNS management contains updating and removing DNS records from existing DNS zones.
+Your {SmartProxy} can use multiple providers, allowing you to integrate with your existing DNS infrastructure or deploy a new one.
+
+After you have enabled DNS, your {SmartProxy} can manipulate any DNS server that complies with RFC 2136.
+
+.Available DNS Providers
+ifdef::orcharhino[]
+* `dns_dnscmd` - static DNS records in Microsoft Active Directory
+endif::[]
+* `dns_infoblox` - For more information, see {ProvisioningDocURL}Using_Infoblox_as_DHCP_and_DNS_Providers_provisioning[Using Infoblox as DHCP and DNS Providers] in _{ProvisioningDocTitle}_.
```
ekohl, line 14: Naming thing: in the installer we don't use the `dns_` prefix, but rather automatically prefix it. To be concrete: the user passes in `--foreman-proxy-dns-provider nsupdate_gss` to use the `dns_nsupdate_gss` provider. What should we follow here?

The same applies to DHCP.

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
```diff
@@ -0,0 +1,24 @@
+[id="Managing_DHCP_Using_Smart_Proxies_{context}"]
+= Managing DHCP Using {SmartProxies}
+
+{Project} can manage DHCP service using your {SmartProxy}.
```
ekohl, line 4: I'd say `integrate with` is better. It can manage ISC DHCP but all others only talk using APIs.
```suggestion
{Project} can integrate with a DHCP service using your {SmartProxy}.
```

ekohl (commented):
When squashing, would it make sense to squash it in a DHCP and DNS commit?

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
maximiliankolb: Applied.

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
```diff
@@ -0,0 +1,24 @@
...
+The {SmartProxy} has multiple providers, allowing you to integrate with your existing DHCP infrastructure or deploy a new one.
+You can use the DHCP module of {SmartProxy} to query for available IP addresses, add new, and delete existing reservations.
+Note that your {SmartProxy} cannot manage subnet declarations.
+ifdef::foreman-el,foreman-deb,katello[]
+For more information on managing subnet declarations, see https://github.com/theforeman/puppet-dhcp[puppet-dhcp].
+endif::[]
+
+The DHCP module can manage various available DHCP providers:
+
+* `dhcp_infoblox` - For more information, see {ProvisioningDocURL}Using_Infoblox_as_DHCP_and_DNS_Providers_provisioning[Using Infoblox as DHCP and DNS Providers] in _{ProvisioningDocTitle}_.
```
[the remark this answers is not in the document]
maximiliankolb: Reading "long term", I'd say that this could be done in a follow-up PR.

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
maximiliankolb: Applied.

guides/common/modules/con_managing-dns-using-smart-proxies.adoc
maximiliankolb: Thanks, applied!

guides/common/modules/con_managing-dns-using-smart-proxies.adoc
maximiliankolb: For me, this is fine as is because there are no `foreman-installer` commands in this concept file. I have checked to ensure that all foreman-installer commands are without `[dhcp-|dns-]` prefix.

guides/common/modules/con_managing-dns-using-smart-proxies.adoc
maximiliankolb: Added a link.

guides/common/modules/proc_configuring-dns-libvirt.adoc
maximiliankolb: Applied.

guides/common/modules/ref_dhcp-options-for-network-configuration.adoc
maximiliankolb: Leaving this as is for now.

guides/common/modules/ref_dhcp-options-for-network-configuration.adoc
maximiliankolb: I have already removed anything regarding Windows for now. I'd be happy to push this upstream once a working installer exists.

guides/doc-Installing_Proxy/master.adoc
maximiliankolb: Applied.

guides/doc-Installing_Proxy/master.adoc
maximiliankolb: 👍

maximiliankolb (commented):
Rebased to master, squashed into two commits, and applied all feedback where possible.

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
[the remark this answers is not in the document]
ekohl: Exactly.

guides/common/modules/ref_dhcp-options-for-network-configuration.adoc
ekohl: Yes, that's why I suggested to leave it out of the Satellite guide. Then it's safer for a cherry pick.

guides/common/modules/ref_dhcp-isc-settings.adoc
```diff
@@ -0,0 +1,43 @@
+[id="DHCP_ISC_Settings_{context}"]
+= dhcp_isc Settings
+
+The _dhcp_isc_ provider uses a combination of the ISC DHCP server OMAPI management interface and parsing of configuration and lease files.
+This requires it to be run either on the same host as the DHCP server or to have network filesystem access to these files.
+You can define the following settings in `dhcp_isc.yml`:
```
ekohl, line 6: Rather than "you can define", should this be more generic?
```suggestion
The following settings are defined in `dhcp_isc.yml`:
```

My reasoning would be that the entire file is managed by the installer and in the manual the user is never supposed to touch these. That's also the reason why I'm very hesitant to include this in the first place. Documenting them gives the impression that it's supported to modify them, which I'd say is unsupported.

guides/common/modules/ref_dhcp-isc-settings.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+ifdef::foreman-el,katello[]
+.Configuring the path to the _config_ and _leases_ files on Enterprise Linux
```
ekohl, line 9:
```suggestion
.Configuring the path to the _config_ and _leases_ files on {EL}
```

Or just keep it generic (and apply the same below to Debian):
```suggestion
.Configuring the path to the _config_ and _leases_ files
```

guides/common/modules/ref_dhcp-isc-settings.adoc
```diff
@@ -0,0 +1,43 @@
+[id="DHCP_ISC_Settings_{context}"]
+= dhcp_isc Settings
+
+The _dhcp_isc_ provider uses a combination of the ISC DHCP server OMAPI management interface and parsing of configuration and lease files.
+This requires it to be run either on the same host as the DHCP server or to have network filesystem access to these files.
```
ekohl, line 5: We have `dhcp_remote_isc` for the network access.
```suggestion
This requires it to be run either on the same host as the DHCP server.
```

guides/common/modules/ref_dhcp-isc-settings.adoc
```diff
@@ -0,0 +1,43 @@
...
+.Setting a port on which the DHCP server listens
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+:omapi_port: _My_DHCP_Server_Port_  # default: 7911
+----
+
+.Setting the host on which the DHCP server runs on
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+:server: _My_DHCP_Server_FQDN_
```
ekohl, line 42: `server` is in `dhcp.yml`, not `dhcp_isc.yml`.

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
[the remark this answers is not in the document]
ekohl: This point is still open.

guides/common/modules/ref_dhcp-options-for-network-configuration.adoc
ekohl: Then please remove this part. I strongly insist on it.

guides/common/modules/proc_configuring-dns-libvirt.adoc
```diff
@@ -0,0 +1,19 @@
...
+
+The _dns_libvirt_ DNS provider manages IP reservations and leases using dnsmasq through the libvirt API.
+It uses `ruby-libvirt` gem to connect to the local or a remote instance of libvirt daemon.
+
+.Procedure
+* You can use `{foreman-installer}` to configure `dns_libvirt`:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+foreman-installer \
```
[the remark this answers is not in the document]
maximiliankolb: Fixed.

guides/common/modules/ref_dhcp-isc-settings.adoc
maximiliankolb: Thanks, applied!

guides/common/modules/ref_dhcp-isc-settings.adoc
maximiliankolb: Applied.

I was hoping to include this as a reference to let users know what could potentially be modified (using the installer).

guides/common/modules/ref_dhcp-isc-settings.adoc
maximiliankolb: Made it more generic.

guides/common/modules/ref_dhcp-isc-settings.adoc
maximiliankolb: Thanks, applied!

guides/common/modules/ref_dhcp-options-for-network-configuration.adoc
maximiliankolb: I have now hidden it for build target satellite.

guides/common/modules/ref_dhcp-options-for-network-configuration.adoc
maximiliankolb: OK, changed from if not satellite to if orcharhino.

maximiliankolb (commented):
Applied more feedback and rebased to master. I am unsure about one thread which I have kept open.

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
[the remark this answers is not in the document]
ekohl: I think so yes.

guides/common/modules/ref_dhcp-isc-settings.adoc
ekohl: I get that, and I do think it's a good idea to document at least somewhere. I'm questioning what the right place is though. There are a lot of examples config files (https://github.com/theforeman/smart-proxy/tree/develop/config/settings.d). Perhaps we should enhance those to have comments. The packaging simply takes these (and from plugins) and ships them by default. The installer then has its own templates, but we can base them on that.

Or we can look at creating a new reference guide altogether. I'd love to have a guide "What is the Smart Proxy" that starts with a high level overview for users/contributors who want to **understand** the concepts (rather than simply use it). We could also use those for onboarding new people.

Food for thought.

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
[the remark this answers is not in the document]
maximiliankolb: Removed.

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
```diff
@@ -0,0 +1,21 @@
+[id="Managing_DHCP_Using_Smart_Proxies_{context}"]
+= Managing DHCP Using {SmartProxies}
+
+{Project} can integrate with a DHCP service using your {SmartProxy}.
```
ekohl, line 4: Perhaps a writer could review this, but the title is plural (SmartProxies) and this sentence is singular. Should it be consistent? `your {singular}` implies there is one while you can use any number of them.

guides/common/modules/proc_securing-the-dhcp-api.adoc
```diff
@@ -0,0 +1,35 @@
+[id="Securing_the_dhcpd_API_{context}"]
+= Securing the dhcpd API
+
+By default, the dhcpd API listens to any host.
+You can add a `omapi_key` to provide basic security.
```
ekohl, line 5:
```suggestion
You can add an `omapi_key` to provide basic security.
```

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
```diff
@@ -0,0 +1,21 @@
...
+The {SmartProxy} has multiple providers, allowing you to integrate with your existing DHCP infrastructure or deploy a new one.
```
ekohl, line 5: Should be generic?
```suggestion
A {SmartProxy} has multiple providers, allowing you to integrate with your existing DHCP infrastructure or deploy a new one.
```

Also, would it make sense to mention that one `{SmartProxy}` can only use one provider active at a time?

ekohl (commented):
At this point I'm happy with it on both the technical and structural level. Some minor textual comments, but other than that I'd ACK this.

I'd really appreciate a professional writer to review the exact contents. @Lennonka could you take a look at the generated texts?

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
maximiliankolb: Changed to "SmartProxies" attribute.

maximiliankolb: Please close this thread if you agree @Lennonka

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
maximiliankolb: Applied. IMHO only one service is a given due to line 4.

guides/common/modules/proc_securing-the-dhcp-api.adoc
maximiliankolb: Applied.

maximiliankolb (commented):
Applied your feedback.

Thank you so much @ekohl for your endless patience and truely helpful comments!

Lennonka:
For some reason, the Installing Proxy guide doesn't build correctly for orcharhino on my machine. @maximiliankolb Is it working well on yours?
My situation is this: The HTML build contains only the first module of the first assembly, everything else is missing in the output.
I.e. I can only see
1. Preparing your Environment for Installation
1.1. System Requirements
Maybe an unclosed ifdef statement?

guides/common/modules/con_managing-dhcp-using-smart-proxies.adoc
```diff
@@ -0,0 +1,21 @@
+[id="Managing_DHCP_Using_Smart_Proxies_{context}"]
+= Managing DHCP Using {SmartProxies}
+
+{Project} can integrate with a DHCP service using {SmartProxies}.
```
Lennonka, line 4: Sorry to say this, but I'd prefer it the other way around - the singular incl. in the title, ID and filename.

ekohl: It was the inconsistency that bothered me. Singular is fine too.

maximiliankolb: Changed title, anchor, and file name to singular for both dhcp and dns.

maximiliankolb:
> For some reason, the Installing Proxy guide doesn't build correctly for orcharhino on my machine. @maximiliankolb Is it working well on yours? My situation is this: The HTML build contains only the first module of the first assembly, everything else is missing in the output. I.e. I can only see
> 
> 1. Preparing your Environment for Installation
>    1.1. System Requirements
>    My output ends with the list item `* qdrouterd`.
> 
> Maybe something wrong with a conditional statement?

Yes, it's broken for some reason. I will have a look at this in some other PR. Thanks for letting me know. In downstream, I don't use any installing guide at the moment, so it's not something I actively monitor.

Rebased to master. But let me know if you'd like to see additional changes or if this is good to be merged. :v:

maximiliankolb:
@Lennonka Would you mind having another look at my PR?

guides/common/modules/ref_dhcp-isc-settings.adoc
```diff
@@ -0,0 +1,45 @@
+[id="DHCP_ISC_Settings_{context}"]
+= dhcp_isc Settings
+
+The _dhcp_isc_ provider uses a combination of the ISC DHCP server OMAPI management interface and parsing of configuration and lease files.
+This requires it to be run either on the same host as the DHCP server.
```
goarsna, line 5: There is an **or** missing for the **either**.

guides/common/modules/ref_dhcp-isc-settings.adoc
```diff
@@ -0,0 +1,45 @@
...
+The following settings are defined in `dhcp_isc.yml`:
+
+ifdef::foreman-el,katello[]
```
goarsna, line 8: `orcharhino` is missing here.

```suggestion
ifdef::foreman-el,katello,orcharhino[]
```

goarsna (changes requested):
I had a look at the changes and made two suggestions. Despite that everything LGTM.

guides/common/modules/ref_dhcp-isc-settings.adoc
maximiliankolb: Reworded.

guides/common/modules/ref_dhcp-isc-settings.adoc
maximiliankolb: Fixed.

maximiliankolb:
Applied all feedback by @goarsna and rebased to master.

Still waiting for a final approval.

ekohl (approved):
I'm happy with this, but I'd ask you to submit a PR to 3.1 so we can easily see the diff there. This is mostly because there may be support implications for Satellite on a released product. In case you weren't aware, for Satellite having it documented means it's supported.

@Lennonka do you want to have a last read through before we merge it?

guides/common/modules/con_managing-dhcp-using-smart-proxy.adoc
```diff
@@ -0,0 +1,21 @@
+[id="Managing_DHCP_Using_Smart_Proxy_{context}"]
+= Managing DHCP Using {SmartProxy}
+
+{Project} can integrate with a DHCP service using {SmartProxies}.
```
Lennonka, line 4: Singular

guides/common/modules/con_managing-dhcp-using-smart-proxy.adoc
```diff
@@ -0,0 +1,21 @@
...
+A {SmartProxy} has multiple providers, allowing you to integrate with your existing DHCP infrastructure or deploy a new one.
```
Lennonka, line 5: `A {SmartProxy} has multiple DHCP service providers that you can use to integrate {Project} with your existing DHCP infrastructure or deploy a new one.` maybe?

guides/common/modules/con_managing-dhcp-using-smart-proxy.adoc
```diff
@@ -0,0 +1,21 @@
...
+You can use the DHCP module of {SmartProxy} to query for available IP addresses, add new, and delete existing reservations.
+Note that your {SmartProxy} cannot manage subnet declarations.
+
+The DHCP module can manage various available DHCP providers:
```
Lennonka, line 9: "implements" instead of "can manage"?

guides/common/modules/con_managing-dns-using-smart-proxy.adoc
```diff
@@ -0,0 +1,27 @@
+[id="Managing_DNS_Using_Smart_Proxy_{context}"]
+= Managing DNS Using {SmartProxy}
+
+{Project} can manage DNS records using your {SmartProxy}.
+DNS management contains updating and removing DNS records from existing DNS zones.
+Your {SmartProxy} can use multiple providers, allowing you to integrate with your existing DNS infrastructure or deploy a new one.
```
Lennonka, line 6: Perhaps a tiny rephrase as above? :)

guides/common/modules/con_managing-dhcp-using-smart-proxy.adoc
Lennonka: Or maybe just a phony heading like `.Available DHCP Providers` instead of this sentence? (same as for DNS below)

guides/common/modules/proc_securing-the-dhcp-api.adoc
```diff
@@ -0,0 +1,35 @@
+[id="Securing_the_dhcpd_API_{context}"]
+= Securing the dhcpd API
+
```
Lennonka, line 3: WTF is dhcpd API, how does it relate to a DHCP provider or SmartProxy and why should I care about securing it? :D

Lennonka (changes requested):
Oki, I tried to give it another review. However, when the output doesn't generate properly, that's it.

guides/common/modules/proc_securing-the-dhcp-api.adoc
ekohl: I think this is a bit below:
> By default, the dhcpd API listens to any host.
> You can add an `omapi_key` to provide basic security.

Should this be phrased stronger? Like

```
By default, the dhcpd API listens to any host without access controls.
```

maximiliankolb: I have applied your suggestion.

Lennonka (approved):
Good to go!

maximiliankolb:
Thanks for reviewing Ewound and Lena! Merged to master and cherry-picked: ✅ 
   a7ab3465d..777670ae7  3.4 -> 3.4
   f2d48fbcb..22b1d9ff6  3.3 -> 3.3
   62d1789ec..0778a7b15  3.2 -> 3.2
   b467691ee..1c434d211  3.1 -> 3.1

ekohl:
I had hoped that for 3.1 we'd have a PR so we could see the changes, but I'm glad we got this in. Thanks for the big effort @maximiliankolb! This is IMHO the base to build on.

ekohl:
> This is IMHO the base to build on.

I started with #1607.

PR #1450, opened 2022-07-13
Link to correct build target in navigation

maximiliankolb:
IMHO does not require cherry-picking because the navigation comes from master only.

WRONG: https://docs.theforeman.org/release/3.1/#
See 
Screenshot 2022-07-13 at 17 07 54

### DIFF
--- a/web/content/js/versions.js
+++ b/web/content/js/versions.js
@@ -333,7 +333,7 @@ const navVersions = [
     builds: [
       {
         title: "Katello on EL",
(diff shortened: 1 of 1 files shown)


ekohl (approved):
This may have been my copy paste error.

maximiliankolb:
I honestly can't say right now if this resolved the issue; but it probably won't make it worse. I'll merge now & then check again. 🤞

maximiliankolb:
It worked AFAICT

PR #1453, opened 2022-07-14
Document about excluding hosts

mjivraja:
Added the event where the host reporting can be excluded by setting the
parameter to false. It was required to document this because an end-user
should be aware of the whole process when the parameter is tweaked. Also,
it can work at a certain level and is not just limited to individual
hosts. The end-user should be aware that if some hosts are already
reported, they can be still removed.

Add documentation for host_registration_insights parameter usage in rh_cloud

https://bugzilla.redhat.com/show_bug.cgi?id=2083138

* [X] Foreman 3.3/Katello 4.5
* [X] Foreman 3.2/Katello 4.4
* [X] Foreman 3.1/Katello 4.3


ShimShtein:
:+1: from me

guides/common/modules/proc_using-red-hat-insights.adoc
```diff
@@ -24,6 +24,16 @@ On each recommendation line, the vertical ellipsis provide a *View in {Team} Ins
 If the host has recommendations listed, click on the number of recommendations.
 On the *Insights* tab, the vertical ellipsis next to the *Remediate* button provide a *Go To {Project} Insights page* link to information for the system, and a *View in {Team} Insights* link to host details on the console.
 
+.Excluding hosts from rh-cloud and insights-client reports
+Starting from {Project} {ProductVersion}, you can omit host reporting from _rh-cloud_ and _insight-client_ reports by setting the `host_registration_insights` parameter to *False*.
```
maximiliankolb, line 28:
```suggestion
You can set the `host_registration_insights` parameter to *False* to omit host reporting from _rh-cloud_ and _insight-client_ reports.
```

We generally don't say "starting with version x" because users will always get the docs that match their Foreman/Katello/Satellite/orcharhino version.

guides/common/modules/proc_using-red-hat-insights.adoc
```diff
@@ -24,6 +24,16 @@ On each recommendation line, the vertical ellipsis provide a *View in {Team} Ins
...
+{Project} will exclude the hosts from _rh-cloud_ reports and block _insight-client_ from uploading a report to the cloud.
+
+This parameter can also be set at the organization, hostgroup, subnet, and domain level.
```
maximiliankolb, line 31: Please make it "active", i.e. "you can set" instead of "can be set".

guides/common/modules/proc_using-red-hat-insights.adoc
```diff
@@ -24,6 +24,16 @@ On each recommendation line, the vertical ellipsis provide a *View in {Team} Ins
...
+It automatically prevents new hosts from being uploaded as long as they are associated with the entity.
+
+In the case where the host is already reported while setting the parameter to false, it can still be removed by the aging process from the cloud.
```
maximiliankolb, line 34:
```suggestion
If a host already sends reports to {Project} and you set the parameter to false, it can still be removed by the aging process from the cloud.
```

I don't understand the last part of the sentence. Which cloud? Where? What's the "aging process"?

maximiliankolb (commented):
Some suggestions.

guides/common/modules/proc_using-red-hat-insights.adoc
mjivraja: Addressed.

maximiliankolb (approved):

maximiliankolb:
   dba67d245..e82e87ada  3.3 -> 3.3
   e65b6ffd0..ff6a6d5c0  3.2 -> 3.2
   8e2bf922d..d3cf70fd4  3.1 -> 3.1

PR #1467, opened 2022-07-18
Modify the prerequisites to install Proxy

mjivraja:
The mentioning of the base operating system was not required as it does not
play any role in reaching out to Red Hat CDN. So, the second prerequisite
was corrected with accurate information about proxy server, and project 
correlation.

Should 'Red Hat CDN' be 'the Satellite'

https://bugzilla.redhat.com/show_bug.cgi?id=2104596

* [X] Foreman 3.3/Katello 4.5
* [X] Foreman 3.2/Katello 4.4
* [X] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/proc_registering-capsule-to-satellite-server.adoc
+++ b/guides/common/modules/proc_registering-capsule-to-satellite-server.adoc
@@ -16,7 +16,8 @@ For more information on manifests and repositories, see {ContentManagementDocURL
 
 .Proxy and Network Prerequisites
 * The {ProjectServer} base operating system must be able to resolve the host name of the {SmartProxy} base operating system and vice versa.
-* The base operating system on which you want to install {SmartProxyServer} must not be configured to use a proxy to connect to the Red Hat CDN.
+* Ensure HTTPS connection using client certificate authentication is possible between {SmartProxyServer} and {ProjectServer}.
+HTTP proxies between {SmartProxyServer} and {ProjectServer} are not supported.
 * You must configure the host and network-based firewalls accordingly.
 For more information, see {InstallingSmartProxyDocURL}capsule-ports-and-firewalls-requirements_{smart-proxy-context}[Ports and Firewalls Requirements].
 include::snip_host-registration-steps.adoc[]


guides/common/modules/proc_registering-capsule-to-satellite-server.adoc
```diff
@@ -16,7 +16,7 @@ For more information on manifests and repositories, see {ContentManagementDocURL
 
 .Proxy and Network Prerequisites
 * The {ProjectServer} base operating system must be able to resolve the host name of the {SmartProxy} base operating system and vice versa.
-* The base operating system on which you want to install {SmartProxyServer} must not be configured to use a proxy to connect to the Red Hat CDN.
+* There should be no proxy configured for the {SmartProxyServer} to connect to the {Project}.
```
maximiliankolb, line 19:
```suggestion
* Ensure there is no HTTP proxy required for {SmartProxyServer} to connect to {ProjectServer}.
```

taftsanders: As I understand it, {ProjectServer} will resolve to "Capsule Server", is this correct? If this is true, then this would still be incorrect. I don't believe there is any part in the documentation that instructs the user to install the Capsule server by making a connection to the Red Hat CDN at any time. All OS updates to include Capsule installation packages come from the Satellite as a client registered to the Satellite. To ensure there are no issues an http proxy cannot be configured between the Satellite and Capsule server.

maximiliankolb: No. Project Server resolves to Foreman Server, and SmartProxyServer resolves to Smart Proxy Server. See https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-base.adoc and https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-satellite.adoc

ekohl: Like @taftsanders, I also disagree with `no HTTP proxy required` because it's not strong enough. HTTP proxies often interfere with HTTPS client certificates so it is advisable (in Satellite's case you could even say unsupported) to have a proxy in place.

ekohl (commented):
I was thinking about this in a broader sense. In the installer we have a [hook](https://github.com/theforeman/foreman-installer/blob/b435f27e9a781d81afe8003cd7852d486770bb4b/hooks/boot/05-environment.rb) which cleans the environment from any HTTP proxy settings. Perhaps the [Smart Proxy](https://github.com/theforeman/smart-proxy) should apply a similar method (configurable under a setting perhaps) to do this. I believe this is a better modal than telling users "you can't do X".

Having said that: by default the Smart Proxy is started from systemd and that cleans the environment variables. I can imagine that today it's not even a problem. It used to be a problem, but in Foreman 2.5 we implemented the environment cleaning in the installer (https://projects.theforeman.org/issues/32180). Prior to this we had a check that told users to unset variables.

In short: is this even still a prerequisite or are we smart enough nowadays that it should be more something you could describe as a note.

taftsanders:
@ekohl In the experiences I have come in contact with this issue, it has mostly been a transparent proxy configured on the network. The most common use case is when using a cloud environment to deploy the Capsule server and having the Satellite deployed on-prem.

ekohl:
That sounds exactly like what I was afraid of. [Middleboxes](https://en.wikipedia.org/wiki/Middlebox) are often a source of problems, not just in Satellite.

guides/common/modules/proc_registering-capsule-to-satellite-server.adoc
maximiliankolb: Yes, now I see it too: "Ensure" + "required" is a weird combination.

guides/common/modules/proc_registering-capsule-to-satellite-server.adoc
```diff
@@ -16,7 +16,7 @@ For more information on manifests and repositories, see {ContentManagementDocURL
 
 .Proxy and Network Prerequisites
 * The {ProjectServer} base operating system must be able to resolve the host name of the {SmartProxy} base operating system and vice versa.
-* The base operating system on which you want to install {SmartProxyServer} must not be configured to use a proxy to connect to the Red Hat CDN.
+* Ensure there is no HTTP proxy configured for {SmartProxyServer} to connect to the {ProjectServer}.
```
ekohl, line 19: To summarize the discussion in this PR: it should rather be a requirement to have HTTPS client certificate authentication work from the `{SmartProxyServer}` to `{ProjectServer}`. Often HTTP proxies, including transparent ones mess this up. I'm not sure how best to phrase this, but the current text doesn't fully cover it IMHO.

mjivraja: Addressed.

guides/common/modules/proc_registering-capsule-to-satellite-server.adoc
```diff
@@ -16,7 +16,7 @@ For more information on manifests and repositories, see {ContentManagementDocURL
 
 .Proxy and Network Prerequisites
 * The {ProjectServer} base operating system must be able to resolve the host name of the {SmartProxy} base operating system and vice versa.
-* The base operating system on which you want to install {SmartProxyServer} must not be configured to use a proxy to connect to the Red Hat CDN.
+* Ensure there is HTTPS client certificate authentication configured for {SmartProxyServer} to connect to the {ProjectServer}.
```
ekohl, line 19: `configured` is wrong since the installer sets up the client certificate authentication. I think it's more like `possible`.

How about something like:

```suggestion
* Ensure HTTPS client certificate authentication is possible between {SmartProxyServer} and {ProjectServer}. HTTP proxies often interfere with this.
```

mjivraja: `Ensure` and `Possible` won't work together in a single sentence. I will prepare a new one.

guides/common/modules/proc_registering-capsule-to-satellite-server.adoc
```diff
@@ -16,7 +16,8 @@ For more information on manifests and repositories, see {ContentManagementDocURL
 
 .Proxy and Network Prerequisites
 * The {ProjectServer} base operating system must be able to resolve the host name of the {SmartProxy} base operating system and vice versa.
-* The base operating system on which you want to install {SmartProxyServer} must not be configured to use a proxy to connect to the Red Hat CDN.
+* Ensure HTTPS client certificate authentication is possible between {SmartProxyServer} and {ProjectServer}.
+HTTP proxies often interfere with this.
```
maximiliankolb, line 20: This sounds very vague. I'd recommend rewording to "Ensure HTTP proxies do not interfere with ..." or similar.

ekohl: I'm not sure that would cover it. Perhaps `HTTPS connection using client certificate authentication` is better than `HTTPS client certificate authentication`? That also implies network connectivity and all.

mjivraja: Addressed @ekohl's suggestion above.

ehelms: I wonder if our wording should be stronger to make it clear we do not support putting a proxy between a Foreman and a Smart Proxy.

ehelms: Might I suggest:

```
HTTP proxies between {SmartProxyServer} and {ProjectServer} are not supported.
```

@ekohl I'd like your opinion on the wording.

ekohl: That sounds good to me.

mjivraja: Thanks @ekohl @ehelms 

I am replacing the second sentence with the above one.

maximiliankolb (approved):
LGTM.

maximiliankolb:
Please squash when/before merging.

mjivraja:
> Please squash when/before merging.

Sure thing!

maximiliankolb:
   c8b3422a0..3e1919bef  3.3 -> 3.3
   06891ad7a..dff2bdcb6  3.2 -> 3.2
   21f61a00a..f85abfc74  3.1 -> 3.1

PR #1470, opened 2022-07-19
Restructure nightly landing page

ekohl:
This only shows a small set of guides to keep it easier for users to navigate. Copying the full navigation will be confusing for new users because the list is overwhelming.

It also removes links to guides which aren't even in the top navigation (and thus are considered not to be ready).

There is still plenty of room for improvement on this, like implementing columns, but this is a start.

If this is considered OK, the same should be applied to all releases.

* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/web/content/release-nightly.adoc
+++ b/web/content/release-nightly.adoc
@@ -1,32 +1,36 @@
+---
+title: nightly
+---
 :FOREMAN_VER: nightly
 :KATELLO_VER: nightly
 
-= Foreman {FOREMAN_VER} and Katello {KATELLO_VER}
+== Foreman {FOREMAN_VER} and Katello {KATELLO_VER}
 
-Use the top menu bar for navigation in the following guides:
+Use the top menu bar for navigation for all guides.
 
-* link:/{FOREMAN_VER}/Planning_Guide/index-foreman-el.html[Planning for Foreman]
-* link:/{FOREMAN_VER}/Release_notes/index-foreman-el.html[Release notes for Foreman]
-* link:/{FOREMAN_VER}/Release_notes/index-katello.html[Release notes for Katello]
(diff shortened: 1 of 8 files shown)


web/content/release-nightly.adoc
```diff
@@ -1,32 +1,36 @@
...
+
+== Upgrading
+
+* link:/{FOREMAN_VER}/Release_notes/index-foreman-el.html[Release notes for Foreman on RHEL/CentOS]
+* link:/{FOREMAN_VER}/Release_notes/index-foreman-deb.html[Release notes for Foreman on Debian/Ubuntu]
+* link:/{FOREMAN_VER}/Release_notes/index-katello.html[Release notes for Katello on RHEL/CentOS]
+
+// Upgrading guides are not ready for non-Katello
+//* link:/{FOREMAN_VER}/Upgrading_and_Updating/index-foreman-el.html[Upgrading and Updating Foreman on RHEL/CentOS]
+//* link:/{FOREMAN_VER}/Upgrading_and_Updating/index-foreman-deb.html[Upgrading and Updating Katello]
```
maximiliankolb, line 35:
```suggestion
//* link:/{FOREMAN_VER}/Upgrading_and_Updating/index-foreman-deb.html[Upgrading and Updating Foreman on Debian]
```

maximiliankolb (approved):
Thanks!

ekohl:
Thanks @maximiliankolb. I've now updated all other versions too. While I was at it I also updated index.adoc.

maximiliankolb (approved):
Thanks for going the extra mile here. LGTM without manually checking all links. I feel like this is best reviewed once merged.

PR #1493, opened 2022-07-27
A step added to update subscription manager config

dubewarsagar:
A step to update subscription manager config while project host
registration has been added.
This step is added after having discussion in BZ#2105995
This step is required for proper Registration method in Load-balancing 
capsule setup for a clients.

https://bugzilla.redhat.com/show_bug.cgi?id=2105995  

* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/snip_host-registration-steps.adoc
+++ b/guides/common/modules/snip_host-registration-steps.adoc
@@ -5,6 +5,14 @@ You can register hosts with {Project} using the host registration feature, the {
 . Click *Generate* to create the registration command.
 . Click on the _files_ icon to copy the command to your clipboard.
 . Log on to the host you want register and run the previously generated command.
+. Update subscription manager configuration for `rhsm.baseurl` and `server.hostname`:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# subscription-manager config \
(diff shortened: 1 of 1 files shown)


maximiliankolb (commented):
Change itself looks reasonable; but I don't understand why. Can someone else confirm?

ehelms:
The registration API does not currently have support for specifying a load-balancer when registering to a content proxy with a load-balancer in front of it. Users wanting to use the registration API to register need to take an extra step to update subscription-manager configuration after registration to point the client at the load balancer.

maximiliankolb:
Thanks @dubewarsagar 

   015ba3da4..8212f040d  3.3 -> 3.3
   214a2f6a9..2d1ecb0b9  3.2 -> 3.2
   633e067d8..605f4ff95  3.1 -> 3.1

I have added Eric's excellent description as git commit message but failed to take line breaks into account. Github UI did not help either ... :shrug:

ekohl:
I just noticed this introduced a regression in the installation instructions. In the Installing Proxy chapter it now also tell users to register with a load balancer, but a Proxy always needs to be registered to a Foreman. Foreman behind a load balancer is unsupported, so the instructions never make sense.

PR #1497, opened 2022-07-28
Introduce EL macro

ekohl:
In upstream it's common to use Enterprise Linux to mean any RHEL-like OS (RHEL, its clones and CentOS Stream (which is technically more of an upstream). In satellite only RHEL is used.

The guidance on using this is that where it doesn't matter, {EL} should be used (such as disk space requirements). In cases where it does make a difference (like setting up repositories) the actual names should be used. It may also result in multiple sections.

This takes a more conservative approach than https://github.com/theforeman/foreman-documentation/pull/982 in that it doesn't replace as many sections. I wonder if this should or shouldn't be cherry picked. It's fairly safe and picking it makes future cherry picks easier due to fewer conflicts.

* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/proc_upgrading-project-in-place-using-leapp.adoc
+++ b/guides/common/modules/proc_upgrading-project-in-place-using-leapp.adoc
@@ -1,27 +1,15 @@
 [id="upgrading-project-in-place-using-leapp_{context}"]
-ifdef::foreman-el,katello[]
-= Upgrading {Project} to Enterprise Linux 8 In-Place Using Leapp
-endif::[]
-ifdef::satellite[]
-= Upgrading {Project} to {RHEL} 8 In-Place Using Leapp
-endif::[]
+= Upgrading {Project} to {EL} 8 In-Place Using Leapp
 
-ifdef::foreman-el,katello[]
-Use this procedure to upgrade your {Project} installation from Enterprise Linux 7 to Enterprise Linux 8.
-endif::[]
-
-ifdef::satellite[]
-Use this procedure to upgrade your {Project} installation from {RHEL} 7 to {RHEL} 8.
-endif::[]
+Use this procedure to upgrade your {Project} installation from {EL} 7 to {EL} 8.
 
 .Prerequisites
+* {Project} {ProjectVersion} running on {EL} 7.
 ifdef::foreman-el,katello[]
-* {Project} {ProjectVersion} running on Enterprise Linux 7.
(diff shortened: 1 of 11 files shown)


guides/common/modules/proc_upgrading-project-in-place-using-leapp.adoc
```diff
@@ -1,27 +1,19 @@
 [id="upgrading-project-in-place-using-leapp_{context}"]
-ifdef::foreman-el,katello[]
-= Upgrading {Project} to Enterprise Linux 8 In-Place Using Leapp
-endif::[]
-ifdef::satellite[]
-= Upgrading {Project} to {RHEL} 8 In-Place Using Leapp
-endif::[]
+ifdef::satellite,foreman-el,katello[]
```
maximiliankolb, line 2: IMHO this should be moved to the assembly that includes this file.

guides/common/modules/proc_upgrading-project-in-place-using-leapp.adoc
```diff
@@ -1,27 +1,19 @@
...
-
-ifdef::satellite[]
-Use this procedure to upgrade your {Project} installation from {RHEL} 7 to {RHEL} 8.
+Use this procedure to upgrade your {Project} installation from {EL} 7 to {EL} 8.
 endif::[]
 
 .Prerequisites
+ifdef::satellite,foreman-el,katello[]
+* {Project} {ProjectVersion} running on {EL} 7.
+endif::[]
```
maximiliankolb, line 11: I don't think this is necessary if we take care of this in the assembly.

maximiliankolb (approved):
I like this PR a lot!

two small suggestions, then feel free to merge and cherry-pick.

guides/common/modules/proc_upgrading-project-in-place-using-leapp.adoc
ekohl: I think you're right and in fact, it's already there. That makes the ifdefs redundant.

maximiliankolb (approved):

ekohl:
For 3.2 and 3.1 I left out some changes due to conflicts.
   605f4ff95..d62c3a868  3.1 -> 3.1
   2d1ecb0b9..aa7b4eeb4  3.2 -> 3.2
   8212f040d..c1624e21c  3.3 -> 3.3

PR #1498, opened 2022-07-28
Make DNF module enablement consistent

ekohl:
This aligns configuring repositories instructions. The katello module should depend on foreman which makes foreman redundant. We also explicitly state the stream rather than depending on there being only one stream.

Fixes: 801d7c8676a6e330c94ac4f0199a64fbe24b6fe1

* [x] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
+++ b/guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
@@ -110,7 +110,7 @@ ifdef::katello[]
 +
 [options="nowrap" subs="attributes"]
 ----
-# dnf module enable -y foreman katello pulpcore
+# dnf module enable -y katello:el8 pulpcore:el8
 ----
 . Clean the yum cache and update the required packages:
 +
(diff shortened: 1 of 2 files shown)


maximiliankolb:
Question: If Katello depends on Foreman, does it also depend of pulp? If not, why?

ekohl:
> If Katello depends on Foreman, does it also depend of pulp?

It does not: https://github.com/theforeman/foreman-packaging/blob/ef8445efabf882f73f04c0e88e6f57438d3e6a62/modulemd/modulemd-katello-el8.yaml#L19-L25

> If not, why?

I thought we discussed this in https://github.com/theforeman/foreman-packaging/pull/7437 but I can't find the exact discussion. We may have talked about this offline, but we decided against it. IIRC it was because Katello strictly speaking doesn't depend on pulpcore. If you make it a requirement, you must always have it present, also when doing checks like repoclosure.

maximiliankolb (approved):

ekohl:
c1624e21c..662598739  3.3 -> 3.3

PR #1492, opened 2022-07-26
Drop version number for RHEL hosts

bangelic:
Removed RHEL 7 specific references and updated to reflect RHEL 8 based
installation and upgrade scenarios. The change was required because
support for installation on RHEL 7 is being dropped and only RHEL 8
installations will be supported.

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3
* For Foreman 3.0 or older, please create a separate PR.
* We do not accept PRs for Foreman 2.3 or older.

PR #1511, opened 2022-08-02
Remove EL7 to EL8 upgrade guides from docs

upadhyeammit:
Before branching the docs to Foreman 3.4 we need to remove references of EL7 to EL8 guides as there is no upgrade possible starting from 3.4


maximiliankolb:
Please also remove unused source files. You'll probably also want to check for now missing links, like `git diff master --name-only | xargs head -n 1` or similar to get a list of anchors that you've removed.

upadhyeammit:
I feel the sources can be still reused for EL8 to EL9?


> Please also remove unused source files. You'll probably also want to check for now missing links, like `git diff master --name-only | xargs head -n 1` or similar to get a list of anchors that you've removed.

I feel the sources can be still reused for EL8 to EL9 if we plan the docs which I think would be similar?

ekohl:
I don't think keeping files around is a good practice. We have git so reverting is easy. Otherwise it's much harder to figure out what is actively in use. There's also no guarantee that it can be used for 8 -> 9. Best to look at the situation then.

maximiliankolb:
I think it's a neat idea to reuse the files if possible. Can you please open an issue for that?

Once this PR get's merged, we can reference the commits that delete the source files, so future contributers can easily revert the commit and adapt the source files.

For now, please delete them for reasons stated by Ewoud.

guides/common/modules/con_migrating-project-to-a-new-el-system.adoc
```diff
@@ -1,25 +0,0 @@
-[id="migrating-project-to-a-new-el-system_{context}"]
```
maximiliankolb, line 1: I am unsure if we should really drop this procedure. I feel like this could be part of a broader backup and restore concept; maybe to clone prod systems?

upadhyeammit:
> I don't think we want to remove these sections because they were an important part of the latest release. #1492 is a PR I have submitted to remove references to EL 7 and RHEL 7 but keeping it where it is necessary such as the sections removed in this PR.

I think your pr is more extensive and we should go with it. I just want to make sure that this work makes into master before 9th August as we need to branch the foreman-documentation for 3.4.

ekohl:
> I think your pr is more extensive and we should go with it. I just want to make sure that this work makes into master before 9th August as we need to branch the foreman-documentation for 3.4.

We passed that point. Now there are merge conflicts. Could you fix those up? We can cherry pick to 3.4.

evgeni:
This has been solved in another PR, closing

PR #1512, opened 2022-08-02
Add short procedures on how to add custom RPM repositories for EL9

maximiliankolb:
Current example for DEB only: [Adding DEB Repository Example for Debian 11](https://docs.theforeman.org/nightly/Content_Management_Guide/index-katello.html#Adding_Custom_DEB_Repository_Example_for_Debian_11_content-management)

* [x] Foreman 3.3/Katello 4.5
* Do not cherry-pick to 3.2 and below because there are no clients for EL9 available.

### DIFF
--- a/guides/common/modules/proc_adding-custom-deb-repository-example-for-debian-11.adoc
+++ b/guides/common/modules/proc_adding-custom-deb-repository-example-for-debian-11.adoc
@@ -14,17 +14,17 @@ For more information, see xref:Creating_a_Custom_Product_{context}[].
 . On the *Repositories* tab, click *New Repository* to create three repositories of type `deb` as follows:
 +
 * *Debian 11 main*
-** *URL*: `\http://ftp.debian.org/debian/`
+** *Upstream URL*: `\http://ftp.debian.org/debian/`
 ** *Releases/Distributions*: `bullseye`
 ** *Component*: `main`
 ** *Architecture*: `amd64`
 * *Debian 11 updates*
-** *URL*: `\http://ftp.debian.org/debian/`
+** *Upstream URL*: `\http://ftp.debian.org/debian/`
 ** *Releases/Distributions*: `bullseye-updates`
 ** *Component*: `main`
 ** *Architecture*: `amd64`
 * *Debian 11 security*
-** *URL*: `\http://deb.debian.org/debian-security/`
+** *Upstream URL*: `\http://deb.debian.org/debian-security/`
 ** *Releases/Distributions*: `bullseye-security`
 ** *Component*: `main`
 ** *Architecture*: `amd64`
@@ -34,10 +34,10 @@ For more information, see xref:Creating_a_Custom_Product_{context}[].
 +
 * **Debian 11 client**
 ifndef::orcharhino[]
-** *URL*: `\https://apt.atix.de/Debian11/`
+** *Upstream URL*: `\https://apt.atix.de/Debian11/`
 endif::[]
 ifdef::orcharhino[]
-** *URL*: see https://atixservice.zendesk.com/hc/de/articles/360013840079[ATIX Service Portal]
+** *Upstream URL*: see https://atixservice.zendesk.com/hc/de/articles/360013840079[ATIX Service Portal]
 endif::[]
 ** *Releases/Distributions*: `stable`
 ** *Component*: `main`
--- a/guides/common/modules/proc_adding-custom-deb-repository-example-for-debian-11.adoc
+++ b/guides/common/modules/proc_adding-custom-deb-repository-example-for-debian-11.adoc
@@ -14,12 +14,12 @@ For more information, see xref:Creating_a_Custom_Product_{context}[].
 . On the *Repositories* tab, click *New Repository* to create three repositories of type `deb` as follows:
 +
 * *Debian 11 main*
-** *Upstream URL*: `\http://ftp.debian.org/debian/`
+** *Upstream URL*: `\http://deb.debian.org/debian/`
 ** *Releases/Distributions*: `bullseye`
 ** *Component*: `main`
 ** *Architecture*: `amd64`
 * *Debian 11 updates*
-** *Upstream URL*: `\http://ftp.debian.org/debian/`
+** *Upstream URL*: `\http://deb.debian.org/debian/`
 ** *Releases/Distributions*: `bullseye-updates`
 ** *Component*: `main`
 ** *Architecture*: `amd64`
--- a/guides/common/modules/proc_adding-custom-rpm-repositories-for-alma-linux-9.adoc
+++ b/guides/common/modules/proc_adding-custom-rpm-repositories-for-alma-linux-9.adoc
@@ -1,0 +1,35 @@
+[id="Adding_Custom_RPM_Repositories_for_Alma_Linux_9_{context}"]
+= Adding {customrpmtitle} Repositories for Alma Linux 9
+
+This example creates a product and repositories for Alma Linux 9.
+
+.Prerequisite
+* Download and import the http://repo.almalinux.org/almalinux/RPM-GPG-KEY-AlmaLinux-9[GPG public key] for yum repositories for Alma Linux 9 into {Project}.
+For more information, see xref:Importing_a_Custom_GPG_Key_{context}[].
+
+.Procedure
+. In the {ProjectWebUI}, navigate to *Content > Products*.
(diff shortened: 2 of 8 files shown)


maximiliankolb:
The link for CentOS Stream 9 repositories seems to be broken: `https://mirrors.centos.org/metalink?repo=centos-appstream-$stream&arch=$basearch&protocol=https,http`

Should I add `https://mirrors.centos.org/metalink?` as exception when checking links?

ekohl:
> The link for CentOS Stream 9 repositories seems to be broken: `https://mirrors.centos.org/metalink?repo=centos-appstream-$stream&arch=$basearch&protocol=https,http`

The `$stream` and `$basearch` variables are replaced by DNF at runtime by the host. So for example, 9-stream and x64_64 does work: https://mirrors.centos.org/metalink?repo=centos-appstream-9-stream&arch=x86_64&protocol=https,http

guides/common/modules/proc_adding-custom-rpm-repositories-for-centos-stream-9.adoc
```diff
@@ -0,0 +1,31 @@
+[id="Adding_Custom_RPM_Repositories_for_CentOS_Stream_9_{context}"]
+= Adding {customrpmtitle} Repositories for CentOS Stream 9
+
+This example creates a product and repositories for CentOS Stream 9.
+
+.Prerequisite
+* Download and import the https://dl.rockylinux.org/pub/rocky/RPM-GPG-KEY-rockyofficial[GPG public key] for yum repositories for Rocky Linux 9 into {Project}.
```
ekohl, line 7: This should refer to the CentOS GPG key

guides/common/modules/proc_adding-custom-rpm-repositories-for-centos-stream-9.adoc
```diff
@@ -0,0 +1,31 @@
...
+.Procedure
+. In the {ProjectWebUI}, navigate to *Content > Products*.
+. Click *Create Product* to create a product named `CentOS Stream 9`.
+For more information, see xref:Creating_a_Custom_Product_{context}[].
+. On the *Repositories* tab, click *New Repository* to create three repositories of type `yum` as follows:
++
+* *CentOS Stream 9 BaseOS*
+** *Upstream URL*: `https://mirrors.centos.org/metalink?repo=centos-baseos-$stream&arch=$basearch&protocol=https,http`
+* *CentOS Stream 9 AppStream*
+** *Upstream URL*: `https://mirrors.centos.org/metalink?repo=centos-appstream-$stream&arch=$basearch&protocol=https,http`
```
ekohl, line 19: Metalink is a different mechanism to find the right mirror. http://mirror.centos.org/ mentions http://mirror.stream.centos.org/ should be used for CentOS Stream 9. There you can find http://mirror.stream.centos.org/9-stream/BaseOS/ and http://mirror.stream.centos.org/9-stream/AppStream/

guides/common/modules/proc_adding-custom-rpm-repositories-for-centos-stream-9.adoc
```diff
@@ -0,0 +1,31 @@
...
++
+For more information, see xref:Adding_Custom_RPM_Repositories_{context}[].
+. Click *Create Product* to create a product named `CentOS Stream 9 client`.
+. On the *Repositories* tab, click *New Repository* to create a repository of type `yum` as follows:
++
+* **CentOS Stream 9 client**
+ifndef::orcharhino[]
+** *Upstream URL*: `http://yum.theforeman.org/client/{ProjectVersion}/el9/`
```
ekohl, line 27: I thought we had a macro for this. There is at least `project-client-name`. Interestingly enough, we don't use this in `project-client-RHEL7-url` and `project-client-RHEL8-url`.

So at least this simplifies things:
```suggestion
** *Upstream URL*: `{project-client-name}el9/`
```

maximiliankolb: Applied (even though I think the macro is named poorly) and resolves to garbage to orcharhino and Satellite; but that doesn't matter here because it will not be shown for orcharhino and Satellite.

maximiliankolb (commented):
Applied your suggestions and rebased to master.

guides/common/modules/proc_adding-custom-rpm-repositories-for-centos-stream-9.adoc
ekohl: > even though I think the macro is named poorly

No question about it.

guides/common/modules/proc_adding-custom-deb-repository-example-for-debian-11.adoc
```diff
@@ -14,17 +14,17 @@ For more information, see xref:Creating_a_Custom_Product_{context}[].
 . On the *Repositories* tab, click *New Repository* to create three repositories of type `deb` as follows:
 +
 * *Debian 11 main*
-** *URL*: `\http://ftp.debian.org/debian/`
+** *Upstream URL*: `\http://ftp.debian.org/debian/`
```
ekohl, line 17: Unrelated to this PR, but when I open https://ftp.debian.org/ it looks like it's really `deb.debian.org`. Perhaps it should all be consistent (now only security mentions `deb.debian.org`).

guides/common/modules/proc_adding-custom-rpm-repositories-for-alma-linux-9.adoc
```diff
@@ -0,0 +1,33 @@
...
+. On the *Repositories* tab, click *New Repository* to create three repositories of type `yum` as follows:
++
+* *Alma Linux 9 BaseOS*
+** *Upstream URL*: `http://repo.almalinux.org/almalinux/9/BaseOS/x86_64/os/`
+* *Alma Linux 9 AppStream*
+** *Upstream URL*: `http://repo.almalinux.org/almalinux/9/AppStream/x86_64/os/`
+* *Alma Linux 9 extras*
+** *Upstream URL*: `http://repo.almalinux.org/almalinux/9/extras/x86_64/os/`
++
+For more information, see xref:Adding_Custom_RPM_Repositories_{context}[].
```
ekohl, line 23: This renders a bit odd to me. It's at the indentation level of `Upstream URL` while I'd have expected it at at the top level, or at least of the numbering (3. On the Repositories ...).

guides/common/modules/proc_adding-custom-rpm-repositories-for-alma-linux-9.adoc
```diff
@@ -0,0 +1,33 @@
...
+. Click *Create Product* to create a product named `Alma Linux 9 client`.
```
ekohl, line 24: Should this be `Foreman EL9 Client` perhaps? It is the same for all EL9 flavors and if you're a shop with (for example) both RHEL and CentOS Stream, you can reuse the product, right?

ekohl (commented):
Overall it looks ok to me. I do wonder if we're not making the guide very long. There should be an easier way, rather than listing all different combinations.

Ideally Katello would have these as "well known products", but that's a larger feature. So short term I'm wondering if we should have a separate guide with all known products. For example, also EL 7 and EL 8 flavors, but also move Debian/Ubuntu there.

Perhaps even list hammer commands too, which make it easy to copy-paste. Could even suggest [foreman-ansible-modules](https://github.com/theforeman/foreman-ansible-modules) for easy automation.

All of those comments don't block this, but I'm wondering what your thoughts are.

guides/common/modules/proc_adding-custom-rpm-repositories-for-alma-linux-9.adoc
maximiliankolb: Fixed.

guides/common/modules/proc_adding-custom-deb-repository-example-for-debian-11.adoc
maximiliankolb: Good idea. Changed all instances.

maximiliankolb:
> Ideally Katello would have these as "well known products", but that's a larger feature.

I will talk to my colleagues about this; it sounds like a potential feature request for Katello to me.

ekohl (approved):

ekohl:
3e1919bef..6c743343c  3.3 -> 3.3

PR #1514, opened 2022-08-02
Make steps for Upgrade Discovery more accurate

bangelic:
Some steps in Upgrading Discovery were not completely accurate so
changes were made and a NOTE was added to better explain template
modification. These changes are required to make the steps clearer and
complete the workflow.

* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3


guides/doc-Upgrading_and_Updating/topics/upgrading_discovery_satellite.adoc
```diff
@@ -2,15 +2,24 @@
 
 = Upgrading Discovery on {ProjectServer}
 
-. Update the Discovery template in the {ProjectWebUI}:
+. Update the Discovery template in the {ProjectWebUI}:.
```
maximiliankolb, line 5: This looks like there's one to many.

guides/doc-Upgrading_and_Updating/topics/upgrading_discovery_satellite.adoc
```diff
@@ -2,15 +2,24 @@
...
 .. In the {ProjectWebUI}, navigate to *Hosts* > *Provisioning templates*.
 .. On the `PXELinux global default` line, click *Clone*.
 .. Enter a new name for the template in the *Name* field, for example `ACME PXE global default`.
 .. In the template editor field, change the line `ONTIMEOUT local` to `ONTIMEOUT discovery` and click *Submit*.
-.. In the {ProjectWebUI}, navigate to *Administer* > *Settings*.
+.. In the {ProjectWebUI}, navigate to *Administer* > *Settings* > *Provisioning*.
```
maximiliankolb, line 10: We generally don't use two `>`. Instead, you can add an extra step starting with ".. On the *Provisioning* tab, select/set/click ...".

guides/doc-Upgrading_and_Updating/topics/upgrading_discovery_satellite.adoc
```diff
@@ -2,15 +2,24 @@
...
+.. Set `Default PXE global template entry` = (custom value for environment).
```
maximiliankolb, line 11: Please reword to remove the brackets.

guides/doc-Upgrading_and_Updating/topics/upgrading_discovery_satellite.adoc
```diff
@@ -2,15 +2,24 @@
...
+.. Select the name of the newly created template from the menu and click *Submit*.
 .. In the {ProjectWebUI}, navigate to *Hosts* > *Provisioning templates*.
 .. Click *Build PXE Default*, then click *OK*.
 
++
+[NOTE]
+====
+If the template is modified, a {Project} upgrade overrides it to its default version.
+Once the PXE Default configuration is built, the template configured in the *Settings* is deployed to the TFTP.
+This can result in deploying the default template if the new template is correctly set in the *Settings*.
```
maximiliankolb, line 22: I am not 100% sure I understand the issue here; but I think we generally recommend cloning templates instead of modifying them. Because, as you said, any changes will be lost with each `foreman-installer` run.

ekohl: I'm not sure this is the same issue, but some background:

You have the TFTP template, which can be rendered and written to the TFTP server for the default boot (using the `Build PXE default` button). This is something you always do. I don't exactly recall if the installer even writes a default or doesn't have one, but in my experience it's something you always do.

Then an upgrade can change the TFTP template (foreman-x.y.rpm has version 1, foreman-x.z.rpm has version 2). When the user hasn't modified it and uses the default TFTP template, then it can also be good to regenerate the rendered template on the TFTP server.

I'm not sure this is the issue @bangelic is trying to address.

I also feel that it's poor user experience to let the user do this. Perhaps we should file an issue for the upgrade tooling to do this (foreman-maintain?). So short term document it, long term automate it.

maximiliankolb: OK, this clarifies things for me indeed! 👍 

I was confusing this with other templates in Foreman, which might be rendered just before provisioning a host.

bangelic: @ares Any comments on this?

bangelic:
Any additional changes needed or can this be merged?

maximiliankolb (approved):
LGTM

maximiliankolb:
   2ddb70492..a21087efe  3.3 -> 3.3
   d9a12b336..c70c5a00a  3.2 -> 3.2
   ad009d7df..e507d018d  3.1 -> 3.1

PR #1517, opened 2022-08-03
[3.2] Replace more uses of OS names with {EL}

ekohl:
(cherry picked from commit 434d771c89581a056faa6abc3fc3a67b3f9c4559)

### DIFF
--- a/guides/common/modules/proc_importing-kickstart-repositories.adoc
+++ b/guides/common/modules/proc_importing-kickstart-repositories.adoc
@@ -28,7 +28,7 @@ endif::[]
 ----
 
 ifeval::["{rhel-version}" == "8"]
-. Create directories for {RHEL}Linux 8 AppStream and BaseOS Kickstart repositories:
+. Create directories for {RHEL} 8 AppStream and BaseOS Kickstart repositories:
 +
 ----
 # mkdir --parents /var/www/html/pub/sat-import/content/dist/rhel8/8.1/x86_64/appstream/kickstart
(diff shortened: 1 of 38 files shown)


maximiliankolb:
Maybe @Lennonka can review this one.

ekohl:
3.2 is not used in Satellite so this is purely upstream to future make cherry picks easier.

guides/common/modules/proc_importing-kickstart-repositories.adoc
```diff
@@ -28,7 +28,7 @@ endif::[]
 ----
 
 ifeval::["{rhel-version}" == "8"]
-. Create directories for {RHEL}Linux 8 AppStream and BaseOS Kickstart repositories:
+. Create directories for {RHEL} 8 AppStream and BaseOS Kickstart repositories:
```
maximiliankolb, line 31:
```suggestion
. Create directories for {EL} 8 AppStream and BaseOS Kickstart repositories:
```

maximiliankolb:
> 3.2 is not used in Satellite so this is purely upstream to future make cherry picks easier.

just for the record, orcharhino does not use Foreman 3.2 either.

guides/common/modules/proc_importing-kickstart-repositories.adoc
ekohl: Below it is the instruction: `# mkdir --parents /var/www/html/pub/sat-import/content/dist/rhel8/8.1/x86_64/appstream/kickstart` which also mentions rhel in the path name. I think `{RHEL}` is correct here.

maximiliankolb (approved):

PR #1526, opened 2022-08-04
Correct puppet-agent log level instructions

ekohl:
The main block in puppet.conf is used for both the server and agent.  Modifying it in the agent block only affects the agent.

Restarting the server is also not needed. The puppetserver is technically a completely unrelated daemon. Combined with changing only the agent section in puppet.conf it remains at the old log level. The puppet-agent daemon automatically reloads the config if it changed, as can be seen in the logs:

    Config file /etc/puppetlabs/puppet/puppet.conf changed; triggering re-parse of all config files.

* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/proc_increasing-the-logging-levels-to-help-with-debugging.adoc
+++ b/guides/common/modules/proc_increasing-the-logging-levels-to-help-with-debugging.adoc
@@ -277,18 +277,12 @@ endif::[]
 You can increase the logging level for Puppet agent on your {ProjectServer}.
 
 .Procedure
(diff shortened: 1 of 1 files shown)


evgeni (approved):

evgeni:
   058e90c37..70e6d558f  3.1 -> 3.1
   30eefc178..806cb2b39  3.2 -> 3.2
   1212bd233..7719b1b13  3.3 -> 3.3

PR #1353, opened 2022-06-24
Unify emphasis for guides (issue 954)

maximiliankolb:
We want to unify the DocURL links and titles across all source files.

PR #1532, opened 2022-08-05
Rename attribute for Installing Server Guide

maximiliankolb:
find . -type f -iname "*.adoc" | xargs sed -i "" "s/InstallingProjectDocURL/InstallingServerDocURL/g"

Instead of cherry-picking, it's easier to just run the `find | xargs sed` command on each branch. I would like to see this change on 3.3, 3.2, and 3.1 too.

See https://github.com/theforeman/foreman-documentation/pull/1353#discussion_r915498665
cc @Lennonka


ekohl (approved):
> find . -type f -iname "*.adoc" | xargs sed -i "" "s/InstallingProjectDocURL/InstallingServerDocURL/g"

Unrelated, but you can use `-exec` in `find` to get xargs behavior. Though perhaps this is a GNU extension, not sure.
`find . -type f -iname "*.adoc" -exec sed -i "s/InstallingProjectDocURL/InstallingServerDocURL/g" {} +`

Main reason I use that more often is that it saves `--no-run-if-empty` in case `find` results in nothing.

PR #1537, opened 2022-08-06
Fix `make prep` in top-level Makefile

adamlazik1:
The `make` utility runs each line in the Makefile in a different
shell from the current working directory. Therefore lines
containing `cd web` and `cd ..` have no effect on the rest
of the commands. Bundle then installs dependencies in the `vendor`
directory which is created outside  of the `foreman-documentation`
repository. You can try this out yourself by running `make prep` in 
the `foreman-documentation` directory. Putting two commands on 
the same line results in them being executed by the same shell 
and dependencies are installed within the repository.

* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,9 +17,7 @@ all: html
 
 prep:
 	bundle install --path vendor
-	cd web
-	bundle install --path ../vendor
-	cd ..
+	cd web && bundle install --path ../vendor
 	mkdir -p $(DEST)/nightly
 
 clean:


ekohl (approved):
Thanks!

ekohl:
The cherry picks to 3.3 and older show conflicts. We also haven't kept the `web/` directory up to date in those branches so I'm not sure if we should cherry pick to older branches.

PR #1544, opened 2022-08-10
Reconsider /usr partition size for RHEL 8

mjivraja:
The partition size of /usr for RHEL 8 is 3GB at present which is not
accurate as it is more than that. It is now increased to 5GB for EL8.
Also, there is already a detailed guide on the same which helps user
to set correct size of partitions that includes /usr as well.
So, we have added a link to redirect users to the detailed guide.

Recommended /usr partition size for RHEL 8 is incorrect.

https://bugzilla.redhat.com/show_bug.cgi?id=2108615

* [X] Foreman 3.3/Katello 4.5
* [X] Foreman 3.2/Katello 4.4
* [X] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/ref_storage-requirements.adoc
+++ b/guides/common/modules/ref_storage-requirements.adoc
@@ -3,8 +3,6 @@
 
 ifdef::foreman-el,katello,satellite[]
 * xref:#storage-el-8[{EL} 8]
-endif::[]
-ifdef::satellite[]
 * xref:#storage-el-7[{EL} 7]
 endif::[]
 
@@ -22,27 +20,46 @@ ifdef::foreman-el,katello,satellite[]
 [cols="1,1,1",options="header"]
 |====
 |Directory |Installation Size |Runtime Size
-ifdef::katello,satellite,orcharhino[]
-|/var/lib/pulp |1 MB |300 GB
-endif::[]
-|/var/lib/pgsql |100 MB |10 GB
-|/usr |3 GB |Not Applicable
-|/opt/puppetlabs |500 MB |Not Applicable
-|====
-endif::[]
 
-ifdef::satellite[]
+|/var/log/ |10 MB |10 GB
+
+|/var/lib/pgsql |100 MB |20 GB
+
+|/usr | 5 GB | Not Applicable
+
+|/opt/puppetlabs | 500 MB | Not Applicable
+
+ifdef::katello,satellite,orcharhino[]
+|/var/lib/pulp/ |1 MB |300 GB
+
+|/var/lib/qpidd/ |25 MB | Not Applicable
+endif::[]
+|====
+
+For detailed information on partitioning and size, refer to the https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/system_design_guide/partitioning-reference_system-design-guide[{RHEL} 8 partitioning guide].
+
 == [[storage-el-7]]{EL} 7
 
 .Storage Requirements for a {ProjectServer} Installation
 [cols="1,1,1",options="header"]
 |====
 |Directory |Installation Size |Runtime Size
-|/var/lib/pulp |1 MB |300 GB
-|/var/opt/rh/rh-postgresql12/lib/pgsql |100 MB |10 GB
-|/usr |3 GB | Not Applicable
-|/opt |3 GB | Not Applicable
-|/opt/puppetlabs |500 MB | Not Applicable
+
+|/var/log/ |10 MB |10 GB
+
+|/var/opt/rh/rh-postgresql12 |100 MB |20 GB
+
+|/usr | 3 GB | Not Applicable
+
+|/opt | 3 GB | Not Applicable
+
+|/opt/puppetlabs | 500 MB | Not Applicable
+
+ifdef::katello,satellite,orcharhino[]
+|/var/lib/pulp/ |1 MB |300 GB
+
+|/var/lib/qpidd/ |25 MB | Not Applicable
+endif::[]
 |====
 endif::[]
 
--- a/guides/common/modules/ref_storage-requirements.adoc
+++ b/guides/common/modules/ref_storage-requirements.adoc
@@ -32,7 +32,7 @@ ifdef::foreman-el,katello,satellite[]
 ifdef::katello,satellite,orcharhino[]
 |/var/lib/pulp |1 MB |300 GB
 
-|/var/lib/qpidd |25 MB | Not Applicable
+|/var/lib/qpidd |25 MB | xref:storage-guidelines_{context}[Refer Storage Guidelines]
 endif::[]
 |====
 
@@ -58,7 +58,7 @@ For detailed information on partitioning and size, refer to the https://access.r
 ifdef::katello,satellite,orcharhino[]
 |/var/lib/pulp |1 MB |300 GB
 
-|/var/lib/qpidd |25 MB | Not Applicable
+|/var/lib/qpidd |25 MB | xref:storage-guidelines_{context}[Refer Storage Guidelines]
 endif::[]
 |====
 endif::[]
--- a/guides/common/modules/ref_storage-requirements.adoc
+++ b/guides/common/modules/ref_storage-requirements.adoc
@@ -21,7 +21,7 @@ ifdef::foreman-el,katello,satellite[]
 |====
 |Directory |Installation Size |Runtime Size
 
-|/var/log/ |10 MB |10 GB
+|/var/log |10 MB |10 GB
 
 |/var/lib/pgsql |100 MB |20 GB
 
@@ -30,9 +30,9 @@ ifdef::foreman-el,katello,satellite[]
 |/opt/puppetlabs | 500 MB | Not Applicable
 
 ifdef::katello,satellite,orcharhino[]
-|/var/lib/pulp/ |1 MB |300 GB
+|/var/lib/pulp |1 MB |300 GB
 
-|/var/lib/qpidd/ |25 MB | Not Applicable
+|/var/lib/qpidd |25 MB | Not Applicable
 endif::[]
 |====
 
@@ -45,7 +45,7 @@ For detailed information on partitioning and size, refer to the https://access.r
 |====
 |Directory |Installation Size |Runtime Size
 
-|/var/log/ |10 MB |10 GB
+|/var/log |10 MB |10 GB
 
 |/var/opt/rh/rh-postgresql12 |100 MB |20 GB
 
@@ -56,9 +56,9 @@ For detailed information on partitioning and size, refer to the https://access.r
 |/opt/puppetlabs | 500 MB | Not Applicable
 
 ifdef::katello,satellite,orcharhino[]
-|/var/lib/pulp/ |1 MB |300 GB
+|/var/lib/pulp |1 MB |300 GB
 
-|/var/lib/qpidd/ |25 MB | Not Applicable
+|/var/lib/qpidd |25 MB | Not Applicable
 endif::[]
 |====
 endif::[]
@@ -69,7 +69,7 @@ ifdef::foreman-deb[]
 |====
 |Directory |Installation Size |Runtime Size
 
-|/var/log/ |10 MB |10 GB
+|/var/log |10 MB |10 GB
 
 |/var/lib/postgresql |100 MB |20 GB
 


mjivraja:
@ehelms Need your review
@maximiliankolb @evgeni Please let me know if there is a separate partitioning guide for Foreman/Katello.

guides/common/modules/ref_storage-requirements.adoc
```diff
@@ -35,6 +33,9 @@ ifdef::katello,satellite,orcharhino[]
 |/var/lib/qpidd/ |25 MB | Not Applicable
 endif::[]
 |====
+ifdef::satellite[]
```
maximiliankolb, line 36:
```suggestion

ifdef::satellite[]
```

maximiliankolb (commented):
I am not aware of any upstream partitioning guide:

```
$ ls -d1 guides/doc-*
guides/doc-Administering_Project
guides/doc-Administering_with_Ansible
guides/doc-Configuring_Ansible
guides/doc-Configuring_Load_Balancer
guides/doc-Converting_a_Host_to_RHEL
guides/doc-Deploying_Hosts_AppCentric
guides/doc-Deploying_Project_on_AWS
guides/doc-Installing_Proxy
guides/doc-Installing_Server
guides/doc-Installing_Server_Disconnected
guides/doc-Managing_Configurations_Puppet
guides/doc-Managing_Content
guides/doc-Managing_Hosts
guides/doc-Managing_Organizations_and_Locations
guides/doc-Performance_Tuning_Guide
guides/doc-Planning_for_Project
guides/doc-Provisioning_Hosts
guides/doc-Quickstart
guides/doc-Release_Notes
guides/doc-Upgrading_and_Updating
```

mjivraja:
> I am not aware of any upstream partitioning guide:
> 
> ```
> $ ls -d1 guides/doc-*
> guides/doc-Administering_Project
> guides/doc-Administering_with_Ansible
> guides/doc-Configuring_Ansible
> guides/doc-Configuring_Load_Balancer
> guides/doc-Converting_a_Host_to_RHEL
> guides/doc-Deploying_Hosts_AppCentric
> guides/doc-Deploying_Project_on_AWS
> guides/doc-Installing_Proxy
> guides/doc-Installing_Server
> guides/doc-Installing_Server_Disconnected
> guides/doc-Managing_Configurations_Puppet
> guides/doc-Managing_Content
> guides/doc-Managing_Hosts
> guides/doc-Managing_Organizations_and_Locations
> guides/doc-Performance_Tuning_Guide
> guides/doc-Planning_for_Project
> guides/doc-Provisioning_Hosts
> guides/doc-Quickstart
> guides/doc-Release_Notes
> guides/doc-Upgrading_and_Updating
> ```

Alright. Then how can we accommodate **/usr** partition for Foreman/Katello? As we have removed it for downstream.

evgeni:
The only two files I know that are talking about storage are
* guides/common/modules/ref_storage-requirements.adoc
* guides/common/modules/ref_capsule-storage-requirements.adoc

They are included in both upstream and downstream doc builds.

This PR only changes the first.

I don't think it's a good idea to drop the `/usr` directory from the list, as we do place a lot of data there (and e.g. still document `/var/log which is also a "default" directory). But if @ehelms said that's the way, I will not block ;-)

I've always have read this table as "this is the *additional* space you will need when deploying $project", whereas people seem sometimes to think "the table said /usr 3GB, so I make that partition 3GB" but OS *plus* $project doesn't fit in 3GB…

ehelms:
> I've always have read this table as "this is the _additional_ space you will need when deploying $project", whereas people seem sometimes to think "the table said /usr 3GB, so I make that partition 3GB" but OS _plus_ $project doesn't fit in 3GB…

Could we consider changing some of the language in this guide to incorporate _minimum_ to try to get the point across better that we are saying what is special about our installation is we *know* that we have increased storage requirements than the baselines and the minimums to consider ought to be based off the numbers in this doc?



> I don't think it's a good idea to drop the `/usr` directory from the list, as we do place a lot of data there (and e.g. still document `/var/log which is also a "default" directory). But if @ehelms said that's the way, I will not block ;-)

This came out of discussions between myself, @ekohl and @upadhyeammit where we were on the fence and leaned towards referring back to the RHEL documentation on this well known locations as the RHEL docs provide more information that should be read around it. I think if we talk about minimums for Satellite/Foreman/Katello *and* reference to those storage docs since these are standard installation locations that'd be OK compromise?

Curious what the others in the conversation think.

mjivraja:
> > I've always have read this table as "this is the _additional_ space you will need when deploying $project", whereas people seem sometimes to think "the table said /usr 3GB, so I make that partition 3GB" but OS _plus_ $project doesn't fit in 3GB…
> 
> Could we consider changing some of the language in this guide to incorporate _minimum_ to try to get the point across better that we are saying what is special about our installation is we _know_ that we have increased storage requirements than the baselines and the minimums to consider ought to be based off the numbers in this doc?
> 
> > I don't think it's a good idea to drop the `/usr` directory from the list, as we do place a lot of data there (and e.g. still document `/var/log which is also a "default" directory). But if @ehelms said that's the way, I will not block ;-)
> 
> This came out of discussions between myself, @ekohl and @upadhyeammit where we were on the fence and leaned towards referring back to the RHEL documentation on this well known locations as the RHEL docs provide more information that should be read around it. I think if we talk about minimums for Satellite/Foreman/Katello _and_ reference to those storage docs since these are standard installation locations that'd be OK compromise?
> 
> Curious what the others in the conversation think.

Hi @maximiliankolb What can be your stand here?

maximiliankolb:
In downstream, we have our own tables of resource requirements. I cannot really comment on the technical stuff. Hopefully you'll come up with a solution that works for down upstream and downstream. 

👍 to making it minimum requirements.

ekohl:
> I've always have read this table as "this is the _additional_ space you will need when deploying $project", whereas people seem sometimes to think "the table said /usr 3GB, so I make that partition 3GB" but OS _plus_ $project doesn't fit in 3GB…

I'm also one of those who saw it as an absolute value, rather than a relative value. I'd also be OK with making sure everyone reads it as relative, but it may make the table too complex.

I do wonder how large our actual install is. A Foreman nightly (without Katello, so no Pulp, Candlepin etc) doesn't appear to be all that big. Feels to me that the 5G recommendation that RHEL 8 has (and 10 for graphical) should be sufficient. That's why I was in favor of dropping it.

ehelms:
My takeaway is - let's include `/usr` but as a relative value and not a maximum or absolute, and ensure we have a good link over to the RHEL docs on this.

guides/common/modules/ref_storage-requirements.adoc
```diff
@@ -32,6 +39,9 @@ endif::[]
 endif::[]
 
 ifdef::satellite[]
+For detailed information on partitioning and size, refer to the https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/system_design_guide/partitioning-reference_system-design-guide[{RHEL} 8 partitioning guide].
```
mjivraja, line 42: Is this a good-to-go @ehelms ? The link you are asking for.

maximiliankolb:
Please rebase to master and squash. Somehow Github UI has "rebase and merge" disabled for this PR.

mjivraja:
> Please rebase to master and squash. Somehow Github UI has "rebase and merge" disabled for this PR.

Sure. I will rebase and merge once the changes are good to go from Eric. :+1:

guides/common/modules/ref_storage-requirements.adoc
ehelms: The link reads good, I don't think we have to gate this on `ifdef::satellite[]` though, as this is useful to any deployment on EL8. So I would drop that ifdef wrapper around this.

ehelms:
Based on our conversation, we want to change this line: `|/usr |3 GB |Not Applicable` to `|/usr |5 GB |Not Applicable`

guides/common/modules/ref_storage-requirements.adoc
mjivraja: Addressed :+1:

mjivraja:
> Based on our conversation, we want to change this line: `|/usr |3 GB |Not Applicable` to `|/usr |5 GB |Not Applicable`

Addressed :+1:

guides/common/modules/ref_storage-requirements.adoc
```diff
@@ -22,27 +20,46 @@ ifdef::foreman-el,katello,satellite[]
...
+
+|/usr | 5 GB | Not Applicable
+
+|/opt/puppetlabs | 500 MB | Not Applicable
+
 ifdef::katello,satellite,orcharhino[]
-|/var/lib/pulp |1 MB |300 GB
+|/var/lib/pulp/ |1 MB |300 GB
+
+|/var/lib/qpidd/ |25 MB | Not Applicable
```
ehelms, line 35: Small nit - in some cases there is a trailing `/` and in others not, could you align on a single method?

mjivraja: Addressed :+1:

mjivraja:
P.S.: I will squash the commits in the end before it gets merged.

guides/common/modules/ref_storage-requirements.adoc
```diff
@@ -22,27 +20,46 @@ ifdef::foreman-el,katello,satellite[]
...
 |Directory |Installation Size |Runtime Size
+
+|/var/log |10 MB |10 GB
+
+|/var/lib/pgsql |100 MB |20 GB
+
+|/usr | 5 GB | Not Applicable
+
+|/opt/puppetlabs | 500 MB | Not Applicable
+
```
maximiliankolb, line 31: is the whitespace necessary?

mjivraja: In my POV, we should keep it even though it does not have any impact on outputs.

guides/common/modules/ref_storage-requirements.adoc
```diff
@@ -22,27 +20,46 @@ ifdef::foreman-el,katello,satellite[]
...
 ifdef::katello,satellite,orcharhino[]
 |/var/lib/pulp |1 MB |300 GB
+
+|/var/lib/qpidd |25 MB | Not Applicable
```
evgeni, line 35: the "Not Applicable" is not entirely correct, as later in `guides/common/modules/ref_storage-guidelines.adoc` we do give guidance how this grows when you add content hosts with katello-agent installed

mjivraja: Okk. Shall I go for `20 GB` instead of `Not Applicable` ?

evgeni: Can you put a reference to the chapter where it's discussed? 20G is too much for *most* users (especially as this only happens when you use the deprecated Katello agent)

mjivraja: I read [here](https://access.redhat.com/documentation/en-us/red_hat_satellite/6.11/html-single/installing_satellite_server_in_a_connected_network_environment/index#storage-guidelines_satellite). If this is correct, we can replace this link with `Not Applicable`.

mjivraja: Hi @evgeni 
There is also a suggestion of `50 MB` from [here](https://bugzilla.redhat.com/show_bug.cgi?id=2011500)

evgeni: Which doesn't make sense whatsoever (honestly, both, 20 and 50MB are nonsense values).

In the guidelines we write "The `/var/lib/qpidd/` directory uses slightly more than 2 MB per Content Host managed by the `goferd` service.", so the runtime size should either refer to that statement, or duplicate it.

evgeni: Or we could keep this PR about the original issue, `/usr`, and discuss qpidd later.

mjivraja: If we resolve `qpidd` here, it will be an addon for me. However, I am fine with either one.

mjivraja: > Which doesn't make sense whatsoever (honestly, both, 20 and 50MB are nonsense values).
> 
> In the guidelines we write "The `/var/lib/qpidd/` directory uses slightly more than 2 MB per Content Host managed by the `goferd` service.", so the runtime size should either refer to that statement, or duplicate it.

Added internal link to Storage Guidelines in the table. :+1:

evgeni (approved):

evgeni:
0b6c7c755..d02021a44  3.4 -> 3.4

@mjivraja this doesn't apply on 3.3 and older, would you mind opening cherry pick PRs directly against those branches?

PR #1546, opened 2022-08-10
enable the foreman/katello modules after leapp upgrade

evgeni:
For Satellite, Leapp will enable the right module, but not for
Foreman/Katello. However, without the module enabled, users won't get
updates past the original version they've upgraded to.

* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/proc_upgrading-project-in-place-using-leapp.adoc
+++ b/guides/common/modules/proc_upgrading-project-in-place-using-leapp.adoc
@@ -184,6 +184,22 @@ After the system reboots, a live system conducts the upgrade, reboots to fix SEL
 # journalctl -u leapp_resume -f
 ----
 
+ifdef::foreman-el[]
+. Enable the Foreman module:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# dnf module enable foreman:el8
+----
+endif::[]
+ifdef::katello[]
+. Enable the Katello and Pulpcore modules:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# dnf module enable katello:el8 pulpcore:el8
+----
+endif::[]
 ifdef::satellite[]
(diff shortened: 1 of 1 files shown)


ekohl (commented):
How do you even get the initial packages if the module isn't enabled? Does it disable modularity somehow?

evgeni:
> How do you even get the initial packages if the module isn't enabled? Does it disable modularity somehow?

Exactly ;)

https://github.com/theforeman/foreman-documentation/blob/2cfe6a111035afe0242591ffa0d67ba0cb1965aa/guides/common/modules/proc_upgrading-project-in-place-using-leapp.adoc#L66

ekohl:
Then I wonder if it can accidentally pull in Ruby 3.0 or 3.1 gems. Would `dnf distrosync` be a good suggestion?

evgeni:
> Then I wonder if it can accidentally pull in Ruby 3.0 or 3.1 gems. Would `dnf distrosync` be a good suggestion?

That only disables modularity for foreman repos, not CentOS.

ekohl (approved):
@maximiliankolb is leapp something Orcharhino supports?

maximiliankolb:
No, orcharhino does not support LEAPP yet. But this is fine for me for now because I generally don't include files regarding installing and upgrading.

evgeni:
   480310cee..07039af60  3.1 -> 3.1
   ace2f6827..b5aa7a081  3.2 -> 3.2
   8962ce391..2f8724433  3.3 -> 3.3

ekohl:
Note we've branched 3.4 docs. Something our cherry picking template doesn't deal with.

evgeni:
Picked to 3.4, but we should disable the el8-upgrade docs in 3.4+ anyways.

PR #1557, opened 2022-08-15
Change location of PostgreSQL data

adamlazik1:
The procedure in Administering Foreman, section 9.1 assumes
RHEL 7 disk layout. Added correct location of PostgreSQL data
for RHEL 8 and 9 while maintaining the original location for
RHEL 7.

Please cherry-pick 6e231a30c7525d7920c578289dc238626db9c335 into:

* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3

Please cherry-pick fb40004da94195977b87a15f35cc76a840de108f into:

* [x] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/proc_estimating-the-size-of-a-backup.adoc
+++ b/guides/common/modules/proc_estimating-the-size-of-a-backup.adoc
@@ -16,6 +16,7 @@ A full backup requires space to store the following data:
 
 .Procedure
 . Enter the `du` command to estimate the size of uncompressed directories containing {Project} database and configuration files:
+ifndef::satellite[]
 +
 .For {EL} 8:
 ----
@@ -105,6 +106,54 @@ a|`/var/lib/qpidd` +
 |85%
 |942 MB -> 141 MB
 |===
+endif::[]
+ifdef::satellite[]
++
+----
+# du -sh /var/lib/pgsql/data /var/lib/pulp
+100G    /var/lib/pgsql/data
+100G	/var/lib/pulp
+
+# du -csh /var/lib/qpidd /var/lib/tftpboot /etc /root/ssl-build \
+/var/www/html/pub /opt/puppetlabs
+886M  /var/lib/qpidd
+16M   /var/lib/tftpboot
+37M   /etc
+900K  /root/ssl-build
+100K  /var/www/html/pub
+2M    /opt/puppetlabs
+942M  total
+----
+. Calculate how much space is required to store the compressed data.
++
+The following table describes the compression ratio of all data items included in the backup:
++
+.Backup Data Compression Ratio
+[cols="4,6,3,5"]
+|===
+|Data type |Directory |Ratio |Example results
+
+|PostgreSQL database files
+|`/var/lib/pgsql/data`
+|80 - 85%
+|100 GB -> 20 GB
+
+|Pulp RPM files
+|`/var/lib/pulp`
+|(not compressed)
+|100 GB
+
+|Configuration files
+a|`/var/lib/qpidd` +
+`/var/lib/tftpboot` +
+`/etc` +
+`/root/ssl-build` +
+`/var/www/html/pub` +
+`/opt/puppetlabs`
+|85%
+|942 MB -> 141 MB
+|===
+endif::[]
 +
 In this example, the compressed backup data occupies 180 GB in total.
 . To calculate the amount of available space you require to store a backup, calculate the sum of the estimated values of compressed and uncompressed backup data, and add an extra 20% to ensure a reliable backup.
--- a/guides/common/modules/proc_estimating-the-size-of-a-backup.adoc
+++ b/guides/common/modules/proc_estimating-the-size-of-a-backup.adoc
@@ -17,6 +17,24 @@ A full backup requires space to store the following data:
 .Procedure
 . Enter the `du` command to estimate the size of uncompressed directories containing {Project} database and configuration files:
 +
+.For {EL} 8:
+----
+# du -sh /var/lib/pgsql/data /var/lib/pulp
+100G    /var/lib/pgsql/data
+100G	/var/lib/pulp
+
+# du -csh /var/lib/qpidd /var/lib/tftpboot /etc /root/ssl-build \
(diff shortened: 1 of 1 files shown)


guides/common/modules/proc_estimating-the-size-of-a-backup.adoc
```diff
@@ -32,11 +33,29 @@ A full backup requires space to store the following data:
 2M    /opt/puppetlabs
 942M  total
 ----
++
+.For RHEL 8 and 9:
```
ekohl, line 37: We don't have support for running on EL 9 yet

guides/common/modules/proc_estimating-the-size-of-a-backup.adoc
```diff
@@ -17,6 +17,7 @@ A full backup requires space to store the following data:
 .Procedure
 . Enter the `du` command to estimate the size of uncompressed directories containing {Project} database and configuration files:
 +
+.For RHEL 7:
```
ekohl, line 20: We started a convention to use `{EL}` as a macro. For Satellite this expands to `{RHEL}` which expands to the full Red Hat Enterprise Linux. In upstream it expands to Enterprise Linux.

ekohl (commented):
I know it's a draft, but 2 small notes.

evgeni:
Ewouds comments are correct. Plus, we've been trying to mention EL8 first, as that's the primary deployment platform these days.

adamlazik1:
@ekohl @evgeni Thank you both for your suggestions. 
Incorporated the suggested changes.

adamlazik1:
Also I noticed that Satellite 6.12 supports only EL 8 while Foreman 3.3 still supports EL 7 and EL 8, so I guess that means I should make a separate PR to cherry-pick into 3.3?

evgeni:
Thanks! I am happy with the changes!

Any reason not to pick them to Foreman 3.3? Or was that an oversight?

adamlazik1:
@evgeni I will make separate PR for 3.3, Because according to the milestones Foreman 3.3 supports both EL 7 and EL 8, while Satellite 6.12 only supports EL 8. I will need to put some `ifdef`s there. Please correct me if I am wrong. The alternative is to keep sections for EL 7 and EL 8 in Satellite 6.12 docs.

adamlazik1:
Or maybe the right thing to do would be to update this PR by putting `ifdef`s with version numbers? But I am not sure whether that is a correct course of action since `ifdef`s with version numbers would break in the future. Let me know if you have any suggestions on how to resolve this.

evgeni:
The OS-support difference is correct. But I would just pick it now and add the `ifdefs` in a seperate PR?

adamlazik1:
Sure thing. Updating PR then.

ekohl:
Indeed. It's fine now to write it as if we support both, pick that to all branches. Then later write it with ifdefs.

You can also write it in multiple commits:
* All changes that treat EL 7 & EL 8 as fully supported
* All changes that make it EL 7 & EL 8 upstream, EL 8 only in Satellite
* All changes that make it EL 8 only

Then we do selective cherry picks. IMHO it's fine to bundle that in a single PR to master and later raise cherry pick PRs for each individual branch.

adamlazik1:
Removed unnecessary trailing slash from `/var/lib/pgsql/data/`. Added commit for Foreman 3.3 which supports both versions if `satellite` isn't defined. I made it so if any of `orcharhino`, `katello`, `foreman-deb` or `foreman-el` are defined, then sections for both EL 7 and EL 8 are present. Is this okay or should the section with EL 8-only display in some other than satellite build? Also made third commit in which the procedure only supports EL 8 as requested.

guides/common/modules/proc_estimating-the-size-of-a-backup.adoc
```diff
@@ -42,10 +43,10 @@ The following table describes the compression ratio of all data items included i
 |Data type |Directory |Ratio |Example results
 
 |PostgreSQL database files
-|`/var/opt/rh/rh-postgresql12/lib/pgsql/data`
+|`/var/lib/pgsql/data`
 |80 - 85%
 |100 GB -> 20 GB
-
+
```
maximiliankolb, line 49: Please remove this whitespace.

guides/common/modules/proc_estimating-the-size-of-a-backup.adoc
```diff
@@ -32,6 +32,7 @@ A full backup requires space to store the following data:
 2M    /opt/puppetlabs
 942M  total
 ----
++
```
maximiliankolb, line 35: This should not be needed here. Please drop.

adamlazik1:
@maximiliankolb Thank you for your feedback.
Incorporated the suggested changes

maximiliankolb (approved):

maximiliankolb:

**Change location of PostgreSQL data**:
   8edd44b86..747c3929e  3.4 -> 3.4
   b07126902..d6cd0d768  3.3 -> 3.3
   adb34178e..a413c406b  3.2 -> 3.2
   1458d566d..d73b03f3c  3.1 -> 3.1

**Adjust the procedure for Foreman 3.3**:
   d6cd0d768..337dbdeec  3.3 -> 3.3

@adamlazik1 Please verify my cherry-picks.

adamlazik1:
@maximiliankolb my apologies, I forgot that the 3.4 branch was added. Please also cherry-pick `Adjust the procedure for Foreman 3.3` into 3.4 and then cherry-pick `Remove support for EL 7 for future releases` on top of that. I will update the PR message too, just for clarification purposes.

maximiliankolb:
@adamlazik1 I have cherry-picked "Adjust the procedure for Foreman 3.3" and "Remove support for EL 7 for future releases" to 3.4:

   cfa2b304e..05bcbbcd9  3.4 -> 3.4

adamlazik1:
@maximiliankolb Many thanks!

PR #1559, opened 2022-08-16
Add the missing SSL protocol for external database

mjivraja:
The protocol is a part of the latest version but it is required to consider
the same for earlier versions too. It is applicable to even earlier
versions while writing this commit.

Add more information on External database configuration with SSL in
Documentation for both ends.

https://bugzilla.redhat.com/show_bug.cgi?id=2079973

* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3


ekohl (changes requested):
This doesn't look right for master. The option is available in upstream since https://projects.theforeman.org/issues/34672. I just modified to properly set Fixed in version to 3.3.0.

Instead, I think you should make this change only in the 3.1 branch if Satellite has cherry picked the fix to it.

ekohl:
> I understand that I should raise the PR for the 3.1 branch of Foreman instead of Master.

That is indeed what I'm suggesting.

mjivraja:
> > I understand that I should raise the PR for the 3.1 branch of Foreman instead of Master.
> 
> That is indeed what I'm suggesting.

Thanks :+1: I am changing the branch.

ekohl:
Looks like that caused a lot of problems. Many unrelated changes and merge conflicts. I think you need to rebase it and then force push.

mjivraja:
Closing this PR. I switched the branch which later I came to know it's a mistake. Will raise the new PR for v3.1.

Sorry for the inconvenience caused.

ekohl:
https://github.com/theforeman/foreman-documentation/pull/1568 is the replacement PR.

PR #1568, opened 2022-08-17
Add the missing SSL protocol for external database

mjivraja:
The protocol is a part of latest version but it is required to consider
the same for earlier versions too. It is applicable to even earlier
versions while writing this commit.

Add more information on External database configuration with SSL in
Documentation  for both ends.

https://bugzilla.redhat.com/show_bug.cgi?id=2079973

* [ ] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/proc_configuring-satellite-to-use-external-databases.adoc
+++ b/guides/common/modules/proc_configuring-satellite-to-use-external-databases.adoc
@@ -30,6 +30,7 @@ Use the `{foreman-installer}` command to configure {Project} to connect to an ex
 
 To enable the Secure Sockets Layer (SSL) protocol for these external databases, add the following options:
 +
+ifndef::satellite[]
 [options="nowrap" subs="+quotes,attributes"]
 ----
 --foreman-db-sslmode verify-full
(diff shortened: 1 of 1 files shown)


mjivraja:
Hi @ekohl 
I have created the new PR from the v3.1 branch of Foreman. Rebasing/Merging wasn't getting possible in the last PR.

guides/common/modules/proc_configuring-satellite-to-use-external-databases.adoc
```diff
@@ -30,6 +30,7 @@ Use the `{foreman-installer}` command to configure {Project} to connect to an ex
 
 To enable the Secure Sockets Layer (SSL) protocol for these external databases, add the following options:
 +
+ifdef::foreman-el,katello,foreman-deb[]
```
ekohl, line 33: I'd make this the opposite of below so you don't exclude orcharhino
```suggestion
ifndef::satellite[]
```

ekohl (changes requested):

guides/common/modules/proc_configuring-satellite-to-use-external-databases.adoc
mjivraja: Agree :+1:

ekohl (approved):
I've confirmed that this was cherry picked to 6.11 via https://bugzilla.redhat.com/show_bug.cgi?id=2062189. So content wise :+1:. I think the commit message should reflect that the `--katello-candlepin-db-ssl-ca` option was specifically backported to Satellite 6.11, which is why we have `ifdef`. Not a blocker to merging though so let me know if you want to rephrase it.

mjivraja:
> I've confirmed that this was cherry picked to 6.11 via https://bugzilla.redhat.com/show_bug.cgi?id=2062189. So content wise +1. I think the commit message should reflect that the `--katello-candlepin-db-ssl-ca` option was specifically backported to Satellite 6.11, which is why we have `ifdef`. Not a blocker to merging though so let me know if you want to rephrase it.

Sure. I am amending the commit as per your suggestion.

ekohl (approved):

PR #1572, opened 2022-08-18
Update repo URLs in building Discovery Image

mjivraja:
For downstream, the existing paths are no longer valid due to the
pulp3. As a result, there were issues in building the images.
Therefore, the URLs of repositories need to be updated and they are
same as upstream in accordance to pupl3.

Please update Provisioning Guide (esp 7.8. Building a Discovery Image)
for pulp3 usage

https://bugzilla.redhat.com/show_bug.cgi?id=2089339

* [ ] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3


ekohl (changes requested):
This implies the image will be built on CentOS 7 instead of RHEL 7. Additionally, it uses the upstream Foreman repositories instead of Satellite ones, meaning security/bug fixes won't be there. That isn't right for Satellite since it's unsupported by Red Hat.

I'm not sure what the right paths are (Pulp 3 doesn't store it as such) or if you can consume the library content anonymously, but that is probably the correct solution.

mjivraja:
> This implies the image will be built on CentOS 7 instead of RHEL 7. Additionally, it uses the upstream Foreman repositories instead of Satellite ones, meaning security/bug fixes won't be there. That isn't right for Satellite since it's unsupported by Red Hat.
> 
> I'm not sure what the right paths are (Pulp 3 doesn't store it as such) or if you can consume the library content anonymously, but that is probably the correct solution.

Thanks @ekohl for your valuable feedback. Even I agree with this. I am connecting with SMEs for the concern and push with accurate repos.

mjivraja:
Hi @ekohl 

I had conversations with SMEs for getting the URLs to build Discovery Image. It seems that there is not a clear way to identify those URLs as they get changed with every release. Also, we can try to export RHEL and Satellite repositories into a directory but finding them will not be easy, at least for end-users.

The URLs that we removed can be optional however, end-users prefer official binaries which do not seem to be possible at this stage.

We can move forward as per your final decision. The ticket related to this need to be addressed sooner as the severity of case has been raised to 1. 

I can provide the comments of SMEs if required. 

Thank you.

ekohl:
I'd say that building on CentOS is unsupported for Satellite and the Foreman packages that it links to are end of life. No security updates are included and it could also miss crucial bugfixes.

Note that you can provide `--baseurl` and that could point to Satellite. I just don't know if there is a content view available that you could use. It may require authentication.

If this is broken, I'd suggest to start with removing the whole chapter by hiding it behind `ifndef::satellite`. No instructions is a better solution than broken instructions.

mjivraja:
> If this is broken, I'd suggest to start with removing the whole chapter by hiding it behind `ifndef::satellite`. No instructions is a better solution than broken instructions.

I went with this idea. Because there is no clear set of instructions to guide end-users as of now.

ekohl (changes requested):
This is not what I meant. I meant that you should disable the whole chapter / procedure. So effectively wrap `include::proc_building-a-discovery-image.adoc[]` in `ifndef::satellite[]`. This is because you must have a base OS while building the image.

mjivraja:
> include::proc_building-a-discovery-image.adoc[]

Addressed :+1:

mjivraja:
> This is not what I meant. I meant that you should disable the whole chapter / procedure. So effectively wrap `include::proc_building-a-discovery-image.adoc[]` in `ifndef::satellite[]`. This is because you must have a base OS while building the image.

Addressed :+1:

mjivraja:
> The build fails, but I don't know ccutil so not sure what exactly fails. Perhaps something refers to that include via some xref?

Thank you for hint :+1: 
If this is ready, I can squash the commits.

mjivraja:
Hi @ekohl 

Just a gentle reminder in case this PR is missed. It seems to move further.

guides/doc-Provisioning_Guide/topics/configuring-the-discovery-service.adoc
```diff
@@ -35,7 +35,7 @@ endif::[]
 
 The `tfm-rubygem-foreman_discovery` package contains the {Project} plug-in to handle discovered nodes, connections, and necessary database structures, and API.
 
-ifdef::satellite[]
+ifndef::satellite[]
```
ekohl, line 38: You can't just switch `ifdef` to `ifndef`. Instead, you should remove it.

ekohl (changes requested):

guides/doc-Provisioning_Guide/topics/configuring-the-discovery-service.adoc
mjivraja: Do you mean it's not applicable to any of the builds (foreman, katello ,etc.) ?

ekohl: Previously this section was only relevant to Satellite. That's what `ifdef::satellite[]` means: it only gets rendered in the Satellite build. Now we said that it no longer works for Satellite so it should be `if NEVER`, but then it's actually better to remove it.

mjivraja: Ok. I was informed it may be applicable to other builds so went with it. Removing :+1:

PR #1577, opened 2022-08-23
Move the HTTP Configuration section to installation

mjivraja:
The configuration in HTTP Proxy to connect to the Content Delivery
Network is required at the time of registering the project server to
CDN. At present, it is a part of configuring the project server with an
HTTP Proxy which is not correct. So, we moved the section here to the
installation process where the registration process takes place.

Instructions "Configuring the HTTP Proxy to Connect to CDN"
should be before "Registering to Subscription Management"
in project Documentation.

https://bugzilla.redhat.com/show_bug.cgi?id=2117691

* [X] Foreman 3.4/Katello 4.6
* [X] Foreman 3.3/Katello 4.5
* [X] Foreman 3.2/Katello 4.4
* [X] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/assembly_configuring-satellite-with-an-http-proxy.adoc
+++ b/guides/common/assembly_configuring-satellite-with-an-http-proxy.adoc
@@ -7,10 +7,6 @@ Use the following procedures to configure {Project} with an HTTP proxy.
 
 include::modules/proc_adding-a-default-http-proxy.adoc[leveloffset=+1]
 
-ifdef::satellite,katello,orcharhino[]
-include::modules/proc_configuring-http-proxy-to-connect-to-cdn.adoc[leveloffset=+1]
-endif::[]
-
 ifndef::foreman-deb[]
 include::modules/proc_configuring-selinux-to-ensure-access-on-custom-ports.adoc[leveloffset=+1]
 endif::[]
--- a/guides/common/assembly_installing-server-connected.adoc
+++ b/guides/common/assembly_installing-server-connected.adoc
@@ -43,6 +43,10 @@ You can restore the previous file as follows:
 restore /etc/dhcp/dhcpd.conf 622d9820b8e764ab124367c68f5fa3a1
 ----
 
+ifdef::satellite,katello,orcharhino[]
+include::modules/proc_configuring-http-proxy-to-connect-to-cdn.adoc[leveloffset=+1]
+endif::[]
+
 ifdef::satellite[]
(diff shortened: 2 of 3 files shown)


maximiliankolb (approved):
This is fine for me because the guide does not change.

mjivraja:
@ekohl Do you have anything to suggest for this change?

ekohl (commented):
I think the general placement is better, but the first line looks odd (taken from the Katello guide):
> Verify that Foreman can connect to the Red Hat CDN and can synchronize its repositories.

How is this actionable for the user? There is no Foreman instance yet. I don't know if Pulp also uses `rhcm.conf`, but I think not.

If the above is correct, this would only affect subscription-manager so I'd expect something like "In case an HTTP proxy is needed, configure subscription-manager in the following way". 

Also, perhaps it should be noted that this is only for RHEL installations, not for EL.

mjivraja:
> 

I would rather prefer to remove that first line. Let it directly start with `Procedure`

mjivraja:
> I think the general placement is better, but the first line looks odd (taken from the Katello guide):
> 
> > Verify that Foreman can connect to the Red Hat CDN and can synchronize its repositories.
> 
> How is this actionable for the user? There is no Foreman instance yet. I don't know if Pulp also uses `rhcm.conf`, but I think not.
> 
> If the above is correct, this would only affect subscription-manager so I'd expect something like "In case an HTTP proxy is needed, configure subscription-manager in the following way".
> 
> Also, perhaps it should be noted that this is only for RHEL installations, not for EL.

I would rather prefer to remove that first line. Let it directly start with `Procedure`

ekohl:
> I would rather prefer to remove that first line. Let it directly start with `Procedure`

That gives the user the implication some HTTP proxy must be set, but it's completely optional (depending on the user's environment).

mjivraja:
> > I would rather prefer to remove that first line. Let it directly start with `Procedure`
> 
> That gives the user the implication some HTTP proxy must be set, but it's completely optional (depending on the user's environment).

Can we move ahead if it looks good to you? @ekohl

ekohl (approved):

ekohl:
There are merge conflicts for 3.2 (didn't try 3.1), please submit a new PR for those.
   a8ebdeb69..cfe63556d  3.3 -> 3.3
   4bddc7e80..0f5168139  3.4 -> 3.4

PR #1579, opened 2022-08-23
Rename the Product Variant for RHEL 8 and above

mjivraja:
Until RHEL 7, there was Red Hat Enterprise Linux Server version as a
name in the Product variant. After the release of RHEL 8, it is changed to
Red Hat Enterprise Linux for x86_64. It is also applicable to RHEL 9
as well. Hence, renamed the Product Variant.

[DDF] This is only true for RHEL7, for RHEL8 the variant is
"Red Hat Enterprise Linux for x86_64"

https://bugzilla.redhat.com/show_bug.cgi?id=2119293

* [X] Foreman 3.4/Katello 4.6
* [X] Foreman 3.3/Katello 4.5
* [X] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/proc_downloading-the-binary-dvd-images.adoc
+++ b/guides/common/modules/proc_downloading-the-binary-dvd-images.adoc
@@ -13,11 +13,11 @@ Use this procedure to download the ISO images for {RHEL} and {ProjectName}.
 
 . Ensure that you have the correct product and version for your environment.
 +
-* *Product Variant* is set to *{RHELServer}*.
+* *Product Variant* is set to *{RHEL} for x86_64*.
 * *Version*  is set to the latest minor version of the product you plan to use as the base operating system.
 * *Architecture* is set to the 64 bit version.
 
-. On the *Product Software* tab, download the Binary DVD image for the latest {RHELServer} version.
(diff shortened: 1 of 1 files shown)


maximiliankolb (approved):
Change looks reasonable. Can someone else confirm?

evgeni:
This change is correct for EL8-only releases, but is not for those which support both EL7 *and* EL8 (read: Satellite 6.11).

I'm cool merging this and picking it to 3.2+, but 3.1 will require a different patch.

mjivraja:
> This change is correct for EL8-only releases, but is not for those which support both EL7 _and_ EL8 (read: Satellite 6.11).
> 
> I'm cool merging this and picking it to 3.2+, but 3.1 will require a different patch.

Completely agree. EL7 for downstream 6.11 is definitely supported and I am in favor to exclude 3.1 from cherry-picking. :+1:

evgeni:
Picked to 3.2+

   30d5d1b71..5dc078725  3.2 -> 3.2
   74555179d..38576233c  3.3 -> 3.3
   bcb92d4b1..7706ad015  3.4 -> 3.4

Please open a separate one for 3.1

PR #1586, opened 2022-08-25
Update the rpm command

adamlazik1:
In Upgrading and Updating guide, section 3.1.1.: Updated the rpm
command to a simpler and faster variant in step 10 of the procedure.
`rpm -q --last kernel` only shows information relevant to the user
and the shortened output is more readable compared to the output of
`rpm -qa --last | grep kernel`.

* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
+++ b/guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
@@ -133,13 +133,20 @@ ifdef::katello[]
 # {foreman-installer}
 ----
 endif::[]
-. Check when the kernel packages were last updated:
+. Determine if the system needs a reboot:
+.. Check the version of newest installed kernel:
 +
 [options="nowrap"]
 ----
-# rpm -qa --last | grep kernel
+# rpm --query --last kernel | head -n 1
 ----
-. Optional: If a kernel update occurred since the last reboot, reboot the system:
+.. Compare this to the version of currently running kernel:
++
+[options="nowrap"]
+----
+# uname --kernel-release
+----
+. Optional: If the newest kernel differs from the currently running kernel, reboot the system:
 +
 ----
 # reboot


guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
```diff
@@ -137,7 +137,7 @@ endif::[]
 +
 [options="nowrap"]
 ----
-# rpm -qa --last | grep kernel
+# rpm -q --last kernel
```
ekohl, line 140: Should we also use the long version for `-q`
```suggestion
# rpm --query --last kernel
```

guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
```diff
@@ -137,7 +137,7 @@ endif::[]
...
 ----
 . Optional: If a kernel update occurred since the last reboot, reboot the system:
```
ekohl, line 142: I wonder if we should suggest running `uname --kernel-release` to determine the running version.

ekohl (approved):
I like this. Some additional suggestions, but don't consider those blocking.

guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
adamlazik1: Thank you for your suggestion. In my opinion, however, we should keep the `-q` option to make the command shorter.

guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
adamlazik1: Good idea!
we could use two commads like this:
```
rpm -q --last kernel
uname --kernel-release
```
These are easy to type but the user has to do that extra step of finding the running release among the showed updates and confirm that it is the last update.
Or we could merge those two commands into one slighty more difficult to type but it shows exactly the information the user wants, and that is whether the newest kernel is currently running:
```
rpm -q --last kernel | head -n1 | grep -v "$(uname --kernel-release)"
```
This command outputs the newest installed kernel in a case that the kernel is not running and the system needs a reboot, or does not output anything in a case that the newest installed kernel is running.

ekohl: I'll defer to other writers, but I think we try to keep examples simple.

guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
ekohl: While in daily life I always use short options, in documentation I always try to use long options because they're self documenting. `-q` is also often used for `--quiet` and with a long option there's no confusion what it does.

adamlazik1: Noted. I will update the command with the long option.

guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
adamlazik1: So it should look something like this?

> 10. Check if the newest installed kernel is running:
> ```
> # rpm --query --last kernel | head -n 1
> # uname --kernel-release
> ```
> 11. Optional: If the kernel versions outputted from the two previous commands differ, reboot the system:
> ```
> # reboot
> ```

ekohl: I think the guidance is that there's at most one command per example. @maximiliankolb?

maximiliankolb: For better readability, we generally place each argument in it's own line. Something like `rpm -q --last kernel | head -n 1 | grep -v "$(uname --kernel-release)"` is perfectly fine for me because you cannot remove options/lines or make alternations.

However, this needs to communicate that we expect exit code 1 but no stdout (unless my test system is broken).

adamlazik1: This is true.
Do you suggest the updated section could look like this?

> 10. Check if the newest installed kernel is running:
> ```
> # rpm --query --last kernel | head -n 1 | grep -v "$(uname --kernel-release)"
> ```
> If the command produces no output, it means that the newest installed kernel is currently running. Note: In such case, 
> the command exits with exit code 1.
> 11. Optional: If the last command produced output, reboot the system:
> ```
> # reboot
> ```

maximiliankolb: Yes, something like that. You could also consider appending `|| true`: https://stackoverflow.com/a/61321426.

adamlazik1: @maximiliankolb Thank you, that seems like a valid solution. I got one more idea.
```
# rpm --query --last kernel | head -n 1 | grep $(uname --kernel-release)
```
this command produces output and exit code 0 if the newest installed kernel is currently running, or does not produce output and exits with code 1 if the system needs a reboot to switch to the newest kernel. Please let me know which alternative would be preferable to you. Either:
```
# rpm --query --last kernel | head -n 1 | grep $(uname --kernel-release)
```
Or:
```
# rpm --query --last kernel | head -n 1 | grep -v $(uname --kernel-release) || true
```

maximiliankolb: I think the first solution is more elegant and easier to understand. Either way, we should let users know what to expect if everything is working as intended.

ekohl: I am quite hesitant to essentially provide shell scripts in documentation.

What comes to mind is that I think katello-host-tools has a mechanism to determine whether a reboot is needed. Perhaps that would be a thing to look into. Looks like it's here: https://github.com/FrostyX/tracer/blob/4b8d869fb4556efaa4c618c3c7102f25239abc95/tracer/resources/tracer.py#L133-L144 (which is essentially the same logic).

For reference, on Debian it's: if `/var/run/reboot-required` exists, you need to reboot. Suse appears to have `/var/run/reboot-needed`.

@upadhyeammit perhaps this should be included in foreman-maintain instead?

guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
```diff
@@ -133,13 +133,14 @@ ifdef::katello[]
...
-. Check when the kernel packages were last updated:
+. Check if the newest installed kernel is currently running:
 +
 [options="nowrap"]
 ----
-# rpm -qa --last | grep kernel
+# rpm --query --last kernel | head -n 1 | grep $(uname --kernel-release)
 ----
-. Optional: If a kernel update occurred since the last reboot, reboot the system:
+If this command produces the kernel version as output, it means that the newest installed kernel is running. Otherwise the system needs a reboot to run the newest kernel.
```
maximiliankolb, line 142:
```suggestion
If this command produces the kernel version as output, it means that the newest installed kernel is running.
Otherwise the system needs a reboot to run the newest kernel.
```

guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
```diff
@@ -133,13 +133,15 @@ ifdef::katello[]
...
 +
 [options="nowrap"]
 ----
-# rpm -qa --last | grep kernel
+# rpm --query --last kernel | head -n 1 | grep $(uname --kernel-release)
 ----
-. Optional: If a kernel update occurred since the last reboot, reboot the system:
+If this command produces the kernel version as output, it means that the newest installed kernel is running.
+Otherwise the system needs a reboot to run the newest kernel.
+. Optional: If the previous command did not produce output, reboot the system:
```
maximiliankolb, line 144:
```suggestion
. If the previous command did not produce output, reboot the system:
```

maximiliankolb (approved):
LGTM; handing over to @ekohl to merge or request further changes.

guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
```diff
@@ -133,13 +133,15 @@ ifdef::katello[]
...
+. Check if the newest installed kernel is currently running:
 +
 [options="nowrap"]
 ----
-# rpm -qa --last | grep kernel
+# rpm --query --last kernel | head -n 1 | grep $(uname --kernel-release)
 ----
-. Optional: If a kernel update occurred since the last reboot, reboot the system:
+If this command produces the kernel version as output, it means that the newest installed kernel is running.
+Otherwise the system needs a reboot to run the newest kernel.
```
mjivraja, line 143:
```suggestion
Otherwise, the system needs a reboot to run the newest kernel.
```
Minor correction.

adamlazik1: Thanks Malhar!

guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
```diff
@@ -133,13 +133,15 @@ ifdef::katello[]
...
-# rpm -qa --last | grep kernel
+# rpm --query --last kernel | head -n 1 | grep $(uname --kernel-release)
 ----
-. Optional: If a kernel update occurred since the last reboot, reboot the system:
+If this command produces the kernel version as output, it means that the newest installed kernel is running.
+Otherwise, the system needs a reboot to run the newest kernel.
+. If the previous command did not produce output, reboot the system:
 +
 ----
 # reboot
```
ekohl, line 147: What I struggle with is the balance. I mean that we are providing a somewhat complicated command that may fail. If it does fail, how does a user recover? What can they modify to make sure they do have the right info. Taking it to the extreme, you could suggest
```
rpm --query --last kernel | head -n 1 | grep $(uname --kernel-release) || reboot
```

Or even more extreme
```
rpm --query --last kernel | head -n 1 | grep $(uname --kernel-release) && echo "Running latest kernel" || reboot
```

But is the reader really helped by this? I'm not a technical writer so I don't know how to answer this, but my feeling is that it may be better to have simple commands.

At a certain point, it may be better to submit a feature request and document that feature.

adamlazik1: Perhaps you would prefer this version?
> 10. Check the newest installed kernel package:
> ```
> # rpm --query --last kernel | head -n 1
> ```
> 11. Check the currently running kernel package:
> ```
> # uname --kernel-release
> ```
> 12. Optional: If the versions of the newest installed and currently running kernel package differ, reboot the system:
> ```
> # reboot
> ```

ekohl: I think so yes, though I'd start with explaining why you're doing it first. Something like
> Optional: to determine if a reboot is needed, check the newest installed kernel package:
> ...
> Compare this to the running kernel version:
> ...
> If the versions ....

adamlazik1: Understood. See if you like the new version. I put determining whether system needs a reboot into one step with sub-steps, and then optional rebooting in another step.

ekohl: sub-steps are a good idea and something I also thought about.

adamlazik1:
@ekohl are we good to merge?

maximiliankolb:
I would probably have done it slightly different, but this is OK for me.

   a85cbd874..3fb68e70f  3.4 -> 3.4
   84b17db71..ec2b7d09c  3.3 -> 3.3
   9b34486ba..bae48cad3  3.2 -> 3.2
   802df3d8c..fa3086c3f  3.1 -> 3.1

PR #1590, opened 2022-08-30
Replace the name of package for EL8

mjivraja:
The package name applicable to EL8 is not correct. It needs to be
replaced with the correct package name that is applicable to v3.3 and above.
For v3.1, the current package name is correct for EL7 and it will
require adding a package for EL8.

Package name for EL7 is correct and but need to update it for EL8.

https://bugzilla.redhat.com/show_bug.cgi?id=2121433

* [X] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [X] Foreman 3.4/Katello 4.6
* [X] Foreman 3.3/Katello 4.5
* [X] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3
* For Foreman 3.0 or older, please create a separate PR.
* We do not accept PRs for Foreman 2.3 or older.

PR #1591, opened 2022-08-30
Add the package for EL8

mjivraja:
For v3.1, the package applicable to EL8 is added in the Provisioning
guide.

The package name for EL7 is correct and but need to update it for EL8.

https://bugzilla.redhat.com/show_bug.cgi?id=2121433

* [ ] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [X] Foreman 3.1/Katello 4.3


mjivraja:
This PR is in reference to #1590 
The reason behind filing the separate PR for v3.1 is to keep the references of EL7 as well EL8 both in the guide.

ekohl (changes requested):
As mentioned in https://github.com/theforeman/foreman-documentation/pull/1590#pullrequestreview-1089896609, I don't think this change is correct.

PR #1378, opened 2022-06-29
Switch external database guide to install on EL8

ehelms:
* [x] I am familiar with the [contributing](CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3
* For Foreman 3.0 or older, please create a separate PR.
* We do not accept PRs for Foreman 2.3 or older.

PR #1598, opened 2022-09-01
Drop migrating internal DB to external from foreman-el

ekohl:
The instructions are very katello-centric and don't work on Foreman.

I'm unsure about cherry picks. We don't link to the guides for non-nightly so it doesn't matter that much.

* [ ] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/doc-Administering_Project/master.adoc
+++ b/guides/doc-Administering_Project/master.adoc
@@ -12,7 +12,9 @@ include::common/assembly_accessing-server.adoc[leveloffset=+1]
 
 ifndef::foreman-deb[]
 include::common/modules/proc_starting-and-stopping-server.adoc[leveloffset=+1]
+endif::[]
 
+ifdef::katello,orcharhino,satellite[]
 include::common/assembly_migrating-from-internal-databases-to-external-databases.adoc[leveloffset=+1]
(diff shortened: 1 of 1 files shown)


maximiliankolb (approved):
I cannot really comment on the _why_ because I haven't tested it, but your change looks reasonable.

ekohl:
> I cannot really comment on the _why_

Because it's broken and needs work. It's better to not give any instructions than broken ones.

`Preparing a Host for External Databases` is close. It creates 2 unused users (candlepin and pulp) and matching DBs, but that doesn't break anything. However, it does give instructions to install `postgresql-evr`, which isn't needed. There are no instructions to set up the repository though and it may not even be available at all.

However, `Migrating to External Databases` provides an installer command that simply doesn't work. The candlepin and pulp options don't exist.

In https://github.com/theforeman/foreman-documentation/pull/1378 things get more complicated. Dropping is easier.

maximiliankolb:
Thanks for explaining. Feel free to merge.

PR #1607, opened 2022-09-07
Document dns_powerdns usage

ekohl:
@maximiliankolb I'm not sure about Orcharhino here. Does it package the plugin?

* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3


maximiliankolb (approved):
Your change looks good to me.

I will create a PR in case orcharhino does not support PowerDNS as DNS provider.

PR #1614, opened 2022-09-08
Add steps for upgrading external database

bangelic:
Steps were added for upgrading external database from RHEL 7 to RHEL 8 while upgrading Satellite versions.

* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/proc_upgrading-external-database.adoc
+++ b/guides/common/modules/proc_upgrading-external-database.adoc
@@ -1,25 +1,25 @@
-[id="Upgrading_External_Database_{context}"]
-= Upgrading External Database
+[id="Upgrading_the_External_Database_{context}"]
+= Upgrading the External Database
 
 You can upgrade an external database from {EL} 7 to {EL} 8 while upgrading {Project} from {ProjectVersionPrevious} to {ProjectVersion}.
 
 .Prerequisites
-. Create a backup.
-+
-* On a virtual machine, take a snapshot.
-* On a physical machine, create a backup.
+. On a virtual machine, take a snapshot.
+. On a physical machine, create a backup.
 . Create a new {EL} 8 based host for PostgreSQL server that follows the external database on {EL} 8 documentation.
+For more information, see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/deploying_different_types_of_servers/using-databases#migrating-to-a-rhel-8-version-of-postgresql_using-postgresql[Migrating to a RHEL 8 version of PostgreSQL].
 
 .Procedure
+. Create a backup.
 . If you keep the same hostname on the new PostgreSQL server, you need to update the DNS preventing the need to do updates on {Project} itself.
 . If the new PostgreSQL server has a new host name or a different CNAME, update the reference to the external database on {Project}:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
 # {foreman-installer} \
-  --foreman-db-host newpostgres.example.com \
-  --katello-candlepin-db-host newpostgres.example.com \
-  --foreman-proxy-content-pulpcore-postgresql-host newpostgres.example.com \
+--foreman-db-host newpostgres.example.com \
+--katello-candlepin-db-host newpostgres.example.com \
+--foreman-proxy-content-pulpcore-postgresql-host newpostgres.example.com
 ----
 . Copy the backup database to the new host.
 . Restore the backup database on the new host.
--- a/guides/common/modules/proc_upgrading-the-external-database.adoc
+++ b/guides/common/modules/proc_upgrading-the-external-database.adoc
@@ -4,14 +4,13 @@
 You can upgrade an external database from {EL} 7 to {EL} 8 while upgrading {Project} from {ProjectVersionPrevious} to {ProjectVersion}.
 
 .Prerequisites
-. On a virtual machine, take a snapshot.
-. On a physical machine, create a backup.
-. Create a new {EL} 8 based host for PostgreSQL server that follows the external database on {EL} 8 documentation.
+* Create a new {EL} 8 based host for PostgreSQL server that follows the external database on {EL} 8 documentation.
 For more information, see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/deploying_different_types_of_servers/using-databases#migrating-to-a-rhel-8-version-of-postgresql_using-postgresql[Migrating to a RHEL 8 version of PostgreSQL].
 
 .Procedure
-. Create a backup.
-. If you keep the same hostname on the new PostgreSQL server, you need to update the DNS preventing the need to do updates on {Project} itself.
+. On a virtual machine, take a snapshot.
+. On a physical machine, create a backup.
+. If you keep the same hostname on the new PostgreSQL server, update the DNS record for the host to prevent the need to do updates on {Project} itself.
 . If the new PostgreSQL server has a new host name or a different CNAME, update the reference to the external database on {Project}:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
--- a/guides/common/modules/proc_upgrading-the-external-database.adoc
+++ b/guides/common/modules/proc_upgrading-the-external-database.adoc
@@ -8,8 +8,7 @@ You can upgrade an external database from {EL} 7 to {EL} 8 while upgrading {Proj
 For more information, see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/deploying_different_types_of_servers/using-databases#migrating-to-a-rhel-8-version-of-postgresql_using-postgresql[Migrating to a RHEL 8 version of PostgreSQL].
 
 .Procedure
-. On a virtual machine, take a snapshot.
-. On a physical machine, create a backup.
+. Create a backup.
 . If you keep the same hostname on the new PostgreSQL server, update the DNS record for the host to prevent the need to do updates on {Project} itself.
 . If the new PostgreSQL server has a new host name or a different CNAME, update the reference to the external database on {Project}:
 +
--- a/guides/common/modules/proc_upgrading-the-external-database.adoc
+++ b/guides/common/modules/proc_upgrading-the-external-database.adoc
@@ -5,11 +5,8 @@ You can upgrade an external database from {EL} 7 to {EL} 8 while upgrading {Proj
 
 .Prerequisites
 * Create a new {EL} 8 based host for PostgreSQL server that follows the external database on {EL} 8 documentation.
-ifdef::satellite[]
-For more information, see https://access.redhat.com/documentation/en-us/red_hat_satellite/{ProjectVersion}/html/installing_satellite_server_in_a_connected_network_environment/performing-additional-configuration#using-external-databases_satellite[Using External Databases with {Project}].
-endif::[]
-ifdef::katello[]
-For more information, see https://docs.theforeman.org/{ProjectVersion}/Installing_Server/index-katello.html#using-external-databases_foreman[Using External Databases with {Project}].
+ifdef::katello,orcharhino,satellite[]
+For more information, see {InstallingServerDocURL}using-external-databases_{project-context}[Using External Databases with {Project}].
 endif::[]
 
 .Procedure
@@ -25,5 +22,3 @@ Otherwise reconfigure {Project} to use the new name:
 --katello-candlepin-db-host _newpostgres.example.com_ \
 --foreman-proxy-content-pulpcore-postgresql-host _newpostgres.example.com_
 ----
-. Copy the backup database to the new host.
-. Restore the backup database on the new host.
--- a/guides/common/modules/proc_upgrading-the-external-database.adoc
+++ b/guides/common/modules/proc_upgrading-the-external-database.adoc
@@ -5,7 +5,6 @@ You can upgrade an external database from {EL} 7 to {EL} 8 while upgrading {Proj
 
 .Prerequisites
 * Create a new {EL} 8 based host for PostgreSQL server that follows the external database on {EL} 8 documentation.
-For more information, see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/deploying_different_types_of_servers/using-databases#migrating-to-a-rhel-8-version-of-postgresql_using-postgresql[Migrating to a RHEL 8 version of PostgreSQL].
 
 .Procedure
 . Create a backup.
@@ -15,9 +14,9 @@ For more information, see https://access.redhat.com/documentation/en-us/red_hat_
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
 # {foreman-installer} \
---foreman-db-host newpostgres.example.com \
---katello-candlepin-db-host newpostgres.example.com \
---foreman-proxy-content-pulpcore-postgresql-host newpostgres.example.com
+--foreman-db-host _newpostgres.example.com_ \
+--katello-candlepin-db-host _newpostgres.example.com_ \
+--foreman-proxy-content-pulpcore-postgresql-host _newpostgres.example.com_
 ----
 . Copy the backup database to the new host.
 . Restore the backup database on the new host.
--- a/guides/common/modules/proc_upgrading-the-external-database.adoc
+++ b/guides/common/modules/proc_upgrading-the-external-database.adoc
@@ -5,11 +5,13 @@ You can upgrade an external database from {EL} 7 to {EL} 8 while upgrading {Proj
 
 .Prerequisites
 * Create a new {EL} 8 based host for PostgreSQL server that follows the external database on {EL} 8 documentation.
+For more information, see https://access.redhat.com/documentation/en-us/red_hat_satellite/{ProjectVersion}/html/installing_satellite_server_in_a_connected_network_environment/performing-additional-configuration#using-external-databases_satellite[Using External Databases with {Project}].
 
 .Procedure
 . Create a backup.
-. If you keep the same hostname on the new PostgreSQL server, update the DNS record for the host to prevent the need to do updates on {Project} itself.
-. If the new PostgreSQL server has a new host name or a different CNAME, update the reference to the external database on {Project}:
+. Restore the backup on the new server.
+. If {Project} reaches the new database server via the old name, no further changes are required.
+Otherwise reconfigure {Project} to use the new name:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
--- a/guides/common/modules/proc_upgrading-the-external-database.adoc
+++ b/guides/common/modules/proc_upgrading-the-external-database.adoc
@@ -5,7 +5,12 @@ You can upgrade an external database from {EL} 7 to {EL} 8 while upgrading {Proj
 
 .Prerequisites
 * Create a new {EL} 8 based host for PostgreSQL server that follows the external database on {EL} 8 documentation.
+ifdef::satellite[]
 For more information, see https://access.redhat.com/documentation/en-us/red_hat_satellite/{ProjectVersion}/html/installing_satellite_server_in_a_connected_network_environment/performing-additional-configuration#using-external-databases_satellite[Using External Databases with {Project}].
+endif::[]
+ifdef::katello[]
+For more information, see https://docs.theforeman.org/{ProjectVersion}/Installing_Server/index-katello.html#using-external-databases_foreman[Using External Databases with {Project}].
+endif::[]
 
 .Procedure
 . Create a backup.
--- a/guides/doc-Upgrading_and_Updating/master.adoc
+++ b/guides/doc-Upgrading_and_Updating/master.adoc
@@ -64,9 +64,9 @@ ifdef::katello,orcharhino[]
 include::topics/upgrading_content_hosts.adoc[leveloffset=+2]
 endif::[]
 
-ifdef::foreman-el,katello,satellite[]
+ifdef::foreman-el,katello,satellite,orcharhino[]
 // Upgrading the External database
-include::common/modules/proc_upgrading-external-database.adoc[leveloffset=+2]
+include::common/modules/proc_upgrading-the-external-database.adoc[leveloffset=+2]
 endif::[]
 
 //Post-Upgrade Tasks
--- a/guides/doc-Upgrading_and_Updating/master.adoc
+++ b/guides/doc-Upgrading_and_Updating/master.adoc
@@ -43,6 +43,11 @@ include::topics/introduction_upgrading_satellite.adoc[leveloffset=+2]
 // Upgrading Satellite Server
 include::topics/upgrading_satellite_server_parent.adoc[leveloffset=+2]
 
+ifdef::foreman-el,katello,satellite[]
+// Upgrading External database
+include::topics/upgrading_external_database.adoc[leveloffset=+3]
+endif::[]
+
 // Upgrading a Connected Satellite Server
 include::topics/upgrading_a_connected_satellite_server.adoc[leveloffset=+3]
 
--- a/guides/doc-Upgrading_and_Updating/master.adoc
+++ b/guides/doc-Upgrading_and_Updating/master.adoc
@@ -43,11 +43,6 @@ include::topics/introduction_upgrading_satellite.adoc[leveloffset=+2]
 // Upgrading Satellite Server
 include::topics/upgrading_satellite_server_parent.adoc[leveloffset=+2]
 
-ifdef::foreman-el,katello,satellite[]
-// Upgrading External database
-include::topics/upgrading_external_database.adoc[leveloffset=+3]
-endif::[]
-
 // Upgrading a Connected Satellite Server
 include::topics/upgrading_a_connected_satellite_server.adoc[leveloffset=+3]
 
@@ -67,6 +62,11 @@ include::topics/upgrading_capsule_server.adoc[leveloffset=+2]
 ifdef::katello,orcharhino[]
 // Upgrading Satellite Clients
 include::topics/upgrading_content_hosts.adoc[leveloffset=+2]
+endif::[]
+
+ifdef::foreman-el,katello,satellite[]
+// Upgrading the External database
+include::common/modules/proc_upgrading-external-database.adoc[leveloffset=+2]
 endif::[]
 
 //Post-Upgrade Tasks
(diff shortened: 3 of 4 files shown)


guides/doc-Upgrading_and_Updating/topics/upgrading_satellite_server_parent.adoc
```diff
@@ -24,3 +24,23 @@ You cannot upgrade {ProjectServer} from a RHEL base system that is not operating
...
+Create a new {RHEL} 8 based host for PostgreSQL server that follows the external database on {RHEL} 8 documentation.
+
+. Copy the backup database to the new host.
+. Restore the backup database on the new host.
+
+If you keep the same hostname on the new PostgreSQL server, you need to update the DNS preventing the need to do  updates on {Project} itself.
+If the new PostgreSQL server has a new host name or a different CNAME, you need to run the following on the {Project}:
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# satellite-installer \
```
maximiliankolb, line 41: Please do not use downstream terminology. We have attributes for this.

guides/doc-Upgrading_and_Updating/topics/upgrading_satellite_server_parent.adoc
```diff
@@ -24,3 +24,23 @@ You cannot upgrade {ProjectServer} from a RHEL base system that is not operating
 To run {ProjectServer} on a {RHEL} base system operating in FIPS mode, you must install {Project} on a freshly provisioned RHEL base system operating in FIPS mode.
 For more information, see {InstallingServerDocURL}Preparing_your_Environment_for_Installation_{project-context}[Preparing your environment for installation] in _{InstallingServerDocTitle}_.
 endif::[]
+
+.Upgrade External Database
+ifdef::satellite[]
+You can upgrade an external database from {RHEL} 7 to {RHEL} 8 while upgrading {Project} from {ProjectVersionPrevious} to {ProjectVersion}.
+
+Create a new {RHEL} 8 based host for PostgreSQL server that follows the external database on {RHEL} 8 documentation.
```
maximiliankolb, line 32: This sounds like a prerequisite to me.

guides/doc-Upgrading_and_Updating/topics/upgrading_satellite_server_parent.adoc
```diff
@@ -24,3 +24,23 @@ You cannot upgrade {ProjectServer} from a RHEL base system that is not operating
...
+
+. Copy the backup database to the new host.
```
maximiliankolb, line 34: Maybe add `.Procedure` before? Maybe move this altogether to a new file.

maximiliankolb (changes requested):

guides/doc-Upgrading_and_Updating/topics/upgrading_external_database.adoc
```diff
@@ -0,0 +1,21 @@
...
+You can upgrade an external database from {RHEL} 7 to {RHEL} 8 while upgrading {Project} from {ProjectVersionPrevious} to {ProjectVersion}.
+
+.Prerequisites
+Create a new {RHEL} 8 based host for PostgreSQL server that follows the external database on {RHEL} 8 documentation.
+
+.Procedure
+. Copy the backup database to the new host.
+. Restore the backup database on the new host.
+
+If you keep the same hostname on the new PostgreSQL server, you need to update the DNS preventing the need to do  updates on {Project} itself.
```
maximiliankolb, line 13:
```suggestion
If you keep the same hostname on the new PostgreSQL server, you need to update the DNS preventing the need to do updates on {Project} itself.
```

I am unsure I understand this correctly. Is this a prerequisite?

guides/doc-Upgrading_and_Updating/topics/upgrading_external_database.adoc
```diff
@@ -0,0 +1,21 @@
...
+If the new PostgreSQL server has a new host name or a different CNAME, you need to run the following on the {Project}:
```
maximiliankolb, line 14:
```suggestion
If the new PostgreSQL server has a new host name or a different CNAME, you need to run the following on {Project}:
```

Maybe reword to "update the reference to the external database on {Project}" or similar? It should IMHO also be a step (aka. start with ". ").

maximiliankolb (commented):
Is this procedure unsupported for Foreman/Katello?

guides/doc-Upgrading_and_Updating/topics/upgrading_external_database.adoc
bangelic: Wasn't sure if this needed to be a step but it's not a prerequisite. Since it is telling the user to update the DNS as a result of keeping the host name, I am changing it to "."

bangelic:
> Is this procedure unsupported for Foreman/Katello?

@ehelms can you advise?

guides/doc-Upgrading_and_Updating/topics/upgrading_external_database.adoc
```diff
@@ -0,0 +1,20 @@
+[id="Upgrading_External_Database_{context}"]
+= Upgrading External Database
+
+You can upgrade an external database from {RHEL} 7 to {RHEL} 8 while upgrading {Project} from {ProjectVersionPrevious} to {ProjectVersion}.
+
+.Prerequisites
+Create a new {RHEL} 8 based host for PostgreSQL server that follows the external database on {RHEL} 8 documentation.
+
+.Procedure
+. Copy the backup database to the new host.
```
ehelms, line 10: I am thinking we should mention as part of prerequisites that they will need to take a backup as part of this process and should consider a maintenance window to do so.

guides/doc-Upgrading_and_Updating/topics/upgrading_external_database.adoc
```diff
@@ -0,0 +1,20 @@
...
+. Restore the backup database on the new host.
+. If you keep the same hostname on the new PostgreSQL server, you need to update the DNS preventing the need to do updates on {Project} itself.
+. If the new PostgreSQL server has a new host name or a different CNAME, update the reference to the external database on {Project}:
```
ehelms, line 13: Points 12 and 13 should come before the backup and restore part to get customers thinking about the network setup of their new host and how it affects the upgrade. Maybe even in the prerequisities?

bangelic:
@ekohl or @Lennonka could one of you take a look at this and see if it's ready to be merged? Thank you in advance!

guides/doc-Upgrading_and_Updating/topics/upgrading_external_database.adoc
```diff
@@ -0,0 +1,25 @@
...
+.Procedure
+. If you keep the same hostname on the new PostgreSQL server, you need to update the DNS preventing the need to do updates on {Project} itself.
+. If the new PostgreSQL server has a new host name or a different CNAME, update the reference to the external database on {Project}:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# {foreman-installer} \
+  --foreman-db-host newpostgres.example.com \
+  --katello-candlepin-db-host newpostgres.example.com \
+  --foreman-proxy-content-pulpcore-postgresql-host newpostgres.example.com \
```
maximiliankolb, line 22:
```suggestion
--foreman-db-host newpostgres.example.com \
--katello-candlepin-db-host newpostgres.example.com \
--foreman-proxy-content-pulpcore-postgresql-host newpostgres.example.com
```

maximiliankolb (commented):
For new content, please add `guides/doc-Upgrading_and_Updating/topics/upgrading_external_database.adoc` as module to `guides/common/modules`.

I am also thinking about rewording the title to "Upgrading the External Database" because it's only one or "Upgrading External Databases" if it's multiple DBs.

guides/doc-Upgrading_and_Updating/topics/upgrading_external_database.adoc
```diff
@@ -0,0 +1,25 @@
...
+= Upgrading External Database
+
+You can upgrade an external database from {EL} 7 to {EL} 8 while upgrading {Project} from {ProjectVersionPrevious} to {ProjectVersion}.
+
+.Prerequisites
+. Create a backup.
++
+* On a virtual machine, take a snapshot.
+* On a physical machine, create a backup.
+. Create a new {EL} 8 based host for PostgreSQL server that follows the external database on {EL} 8 documentation.
```
ekohl, line 11: Can you add a link to that section so users don't need to search for it?

guides/doc-Upgrading_and_Updating/topics/upgrading_external_database.adoc
```diff
@@ -0,0 +1,25 @@
+[id="Upgrading_External_Database_{context}"]
+= Upgrading External Database
+
+You can upgrade an external database from {EL} 7 to {EL} 8 while upgrading {Project} from {ProjectVersionPrevious} to {ProjectVersion}.
+
+.Prerequisites
+. Create a backup.
++
+* On a virtual machine, take a snapshot.
```
ekohl, line 9: Does this make sense? You're creating a backup anyway (to transfer to the new machine) and you're not making changes to the old host. So why is a snapshot needed?

In other words, move "Create a backup" from a Prerequisite to a step in the Procedure.

guides/doc-Upgrading_and_Updating/topics/upgrading_external_database.adoc
```diff
@@ -0,0 +1,25 @@
...
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# {foreman-installer} \
+  --foreman-db-host newpostgres.example.com \
+  --katello-candlepin-db-host newpostgres.example.com \
+  --foreman-proxy-content-pulpcore-postgresql-host newpostgres.example.com \
+----
+. Copy the backup database to the new host.
+. Restore the backup database on the new host.
```
ekohl, line 25: Would it make sense to give explicit instructions for this? Like how to use `pg_dump` and `pg_restore`.

ekohl (changes requested):
Should there be a recommendation around decomissioning the old machine?

Another thing I'd suggest is how to minimize data loss.

Let's say you have these steps:
1. Create a backup
2. Import backup
3. Reconfigure Foreman

Now anything that happens between step 1 and step 3 is lost. This can be prevented by stopping all services on Foreman as step 0.

guides/doc-Upgrading_and_Updating/topics/upgrading_external_database.adoc
bangelic: @ehelms your thoughts?

guides/doc-Upgrading_and_Updating/topics/upgrading_external_database.adoc
bangelic: @ekohl I am not 100% sure if this is the correct link but I am adding this:
(https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/deploying_different_types_of_servers/using-databases#migrating-to-a-rhel-8-version-of-postgresql_using-postgresql)

guides/doc-Upgrading_and_Updating/topics/upgrading_external_database.adoc
ehelms: The stance has been that if a user under takes running an external database server they are expected to take care of doing a backup and restore and the details of those procedures is out of scope of our documentation. Our intent with including an OS upgrade guide is more to give guidance the order of operations and general steps a user should take but the details are up to them based on their database.

bangelic:
Is this ready to be merged?

adamlazik1 (commented):
LGTM from a writer perpective, but I cannot confirm the actual content of the changes. Leaving to @ekohl for review.

guides/common/modules/proc_upgrading-external-database.adoc
```diff
@@ -0,0 +1,25 @@
+[id="Upgrading_the_External_Database_{context}"]
+= Upgrading the External Database
+
+You can upgrade an external database from {EL} 7 to {EL} 8 while upgrading {Project} from {ProjectVersionPrevious} to {ProjectVersion}.
+
+.Prerequisites
+. On a virtual machine, take a snapshot.
+. On a physical machine, create a backup.
+. Create a new {EL} 8 based host for PostgreSQL server that follows the external database on {EL} 8 documentation.
```
Lennonka, line 9: Use a bulleted list (`*`) for prerequisites.

guides/common/modules/proc_upgrading-external-database.adoc
```diff
@@ -0,0 +1,25 @@
...
+For more information, see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/deploying_different_types_of_servers/using-databases#migrating-to-a-rhel-8-version-of-postgresql_using-postgresql[Migrating to a RHEL 8 version of PostgreSQL].
+
+.Procedure
+. Create a backup.
+. If you keep the same hostname on the new PostgreSQL server, you need to update the DNS preventing the need to do updates on {Project} itself.
```
Lennonka, line 14:
```suggestion
. If you keep the same hostname on the new PostgreSQL server, update the DNS record for the host so that ...
```
The meaning of the second sentence is unclear to me. Can you be more specific?
Is this about updating DNS so that {Project} can still see the host as registered? Or what would happen if I didn't update the DNS?

Lennonka (changes requested):
A couple of details. + Adam's comment.

guides/common/modules/proc_upgrading-external-database.adoc
```diff
@@ -0,0 +1,25 @@
+[id="Upgrading_the_External_Database_{context}"]
```
maximiliankolb, line 1: Please adjust the file name to `proc_upgrading-the-external-database.adoc`

guides/doc-Upgrading_and_Updating/master.adoc
```diff
@@ -67,6 +67,11 @@ ifdef::katello,orcharhino,satellite[]
 include::topics/upgrading_content_hosts.adoc[leveloffset=+2]
 endif::[]
 
+ifdef::foreman-el,katello,satellite[]
```
maximiliankolb, line 70:
```suggestion
ifdef::foreman-el,katello,satellite,orcharhino[]
```

guides/doc-Upgrading_and_Updating/master.adoc
```diff
@@ -64,6 +64,11 @@ ifdef::katello,orcharhino,satellite[]
 include::topics/upgrading_content_hosts.adoc[leveloffset=+2]
 endif::[]
 
+ifdef::foreman-el,katello,satellite,orcharhino[]
+// Upgrading the External database
```
ekohl, line 68: Does this comment really add anything? I know we have it all over the place, but it's essentially repeating the self-explaining filename.

guides/common/modules/proc_upgrading-the-external-database.adoc
```diff
@@ -0,0 +1,24 @@
...
+
+.Prerequisites
+* Create a new {EL} 8 based host for PostgreSQL server that follows the external database on {EL} 8 documentation.
+For more information, see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/deploying_different_types_of_servers/using-databases#migrating-to-a-rhel-8-version-of-postgresql_using-postgresql[Migrating to a RHEL 8 version of PostgreSQL].
+
+.Procedure
+. On a virtual machine, take a snapshot.
+. On a physical machine, create a backup.
+. If you keep the same hostname on the new PostgreSQL server, update the DNS record for the host to prevent the need to do updates on {Project} itself.
+. If the new PostgreSQL server has a new host name or a different CNAME, update the reference to the external database on {Project}:
```
ekohl, line 14: I think this is confusing. Let's first start at the requirement. The requirement is that the Foreman server can reach the DB server.

There are multiple ways to achieve this.

First of all, a new hostname is used and the procedure below uses the installer to change the configs accordingly.

Then the other is that the same hostname is used. To make sure it reaches the right server, you can use two techniques.

You described changing the DNS record, but this has some potential issues. Any DNS record has a TTL (Time To Live), which allows any DNS server to cache it. If you don't plan for it and you have a high TTL value (86400 seconds, or 1 day, is common) then it can take a long time for the migration to propagate. A common technique is to plan for it, lower the TTL in advance (at least $TTL) and then change it. Once it's stable it can be raised again.

However during the migration it may still resolve to the old host if you have multiple caching servers. That's why many admins actually prefer to move the IP to the new server instead. So you use some temporary IP to completely provision & prepare the server and then add the old IP while removing it from the old server (or shutting that down).

Some also combine that and do migrate to a new IP but add both IPs to the same server during the migration so the stale DNS cache isn't an issue.

In my mind those are all fairly standard operating procedures so I wonder how much of it should be written out.

guides/common/modules/proc_upgrading-the-external-database.adoc
```diff
@@ -0,0 +1,24 @@
...
+
+You can upgrade an external database from {EL} 7 to {EL} 8 while upgrading {Project} from {ProjectVersionPrevious} to {ProjectVersion}.
+
+.Prerequisites
+* Create a new {EL} 8 based host for PostgreSQL server that follows the external database on {EL} 8 documentation.
+For more information, see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/deploying_different_types_of_servers/using-databases#migrating-to-a-rhel-8-version-of-postgresql_using-postgresql[Migrating to a RHEL 8 version of PostgreSQL].
+
+.Procedure
+. On a virtual machine, take a snapshot.
+. On a physical machine, create a backup.
```
ekohl, line 12: This should be reduced to:
```suggestion
. Create a backup.
```

This is because the procedure later states "Copy the backup database".

Additionally, many sysadmins will keep a virtual server around for a while as a backup, instead of separate backups/snapshots. In most virtualization platforms you can just shut it down anyway.

guides/doc-Upgrading_and_Updating/master.adoc
bangelic: @ekohl I have often wondered the same thing.

guides/common/modules/proc_upgrading-the-external-database.adoc
bangelic: @evgeni related to my comment above regarding this point. Your thoughts?

guides/common/modules/proc_upgrading-the-external-database.adoc
```diff
@@ -0,0 +1,23 @@
...
+You can upgrade an external database from {EL} 7 to {EL} 8 while upgrading {Project} from {ProjectVersionPrevious} to {ProjectVersion}.
+
+.Prerequisites
+* Create a new {EL} 8 based host for PostgreSQL server that follows the external database on {EL} 8 documentation.
+For more information, see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/deploying_different_types_of_servers/using-databases#migrating-to-a-rhel-8-version-of-postgresql_using-postgresql[Migrating to a RHEL 8 version of PostgreSQL].
+
+.Procedure
+. Create a backup.
+. If you keep the same hostname on the new PostgreSQL server, update the DNS record for the host to prevent the need to do updates on {Project} itself.
+. If the new PostgreSQL server has a new host name or a different CNAME, update the reference to the external database on {Project}:
```
evgeni, line 13: If you perform any of these steps *before* restoring the DB on the new system, you will see errors, which will confuse the reader (especially if you use the Installer method and the installer will fail).

guides/common/modules/proc_upgrading-the-external-database.adoc
```diff
@@ -0,0 +1,23 @@
...
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# {foreman-installer} \
+--foreman-db-host newpostgres.example.com \
+--katello-candlepin-db-host newpostgres.example.com \
+--foreman-proxy-content-pulpcore-postgresql-host newpostgres.example.com
```
evgeni, line 20:
```suggestion
--foreman-db-host _newpostgres.example.com_ \
--katello-candlepin-db-host _newpostgres.example.com_ \
--foreman-proxy-content-pulpcore-postgresql-host _newpostgres.example.com_
```

italics to indicate things that need replacement

evgeni (changes requested):

guides/common/modules/proc_upgrading-the-external-database.adoc
evgeni: I think I'd slightly reword this, making it less ambiguous:
* create a backup 
* restore the backup on the new server
* If {Project} can reach the new database server via the old name, no further changes are required. Otherwise reconfigure {Project} to use the new name: <installer invocation>

All further details are left as an exercise to the reader, as we do not know how their setup looks like.

guides/common/modules/proc_upgrading-the-external-database.adoc
```diff
@@ -0,0 +1,23 @@
+[id="Upgrading_the_External_Database_{context}"]
+= Upgrading the External Database
+
+You can upgrade an external database from {EL} 7 to {EL} 8 while upgrading {Project} from {ProjectVersionPrevious} to {ProjectVersion}.
+
+.Prerequisites
+* Create a new {EL} 8 based host for PostgreSQL server that follows the external database on {EL} 8 documentation.
```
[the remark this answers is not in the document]
bangelic: @evgeni Is there a section for using external databases or am I missing something?

guides/common/modules/proc_upgrading-the-external-database.adoc
evgeni: I'd move them towards the end, see also the wording suggestion in https://github.com/theforeman/foreman-documentation/pull/1614#discussion_r1021241684

guides/common/modules/proc_upgrading-the-external-database.adoc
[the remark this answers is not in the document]
evgeni: I was expecting "follows the external database documentation" to link to https://docs.theforeman.org/nightly/Installing_Server/index-katello.html#using-external-databases_foreman, yes.

guides/common/modules/proc_upgrading-the-external-database.adoc
ekohl: > create a backup

Should we recommend an offline backup? So stop all services and then create a backup to prevent data loss?

evgeni: Given https://github.com/theforeman/foreman-documentation/blob/a9d7313f01d8fbe0d7a4a88db72583a24d1f7385/guides/common/assembly_using-external-databases.adoc#L10-L12
I would tend to say "nah, not our pint"

(but we probably should add a similar sentence to non-Satellite builds of the documentation)

guides/common/modules/proc_upgrading-the-external-database.adoc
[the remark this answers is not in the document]
bangelic: Thanks! It's in the Installation Guide so I have to use a different link than xref.

guides/common/modules/proc_upgrading-the-external-database.adoc
bangelic: There really wasn't a place to move these since there weren't many steps in the Procedure to begin with but it has been reworded.

bangelic:
If everything looks good, can we get this merged before the release?

guides/common/modules/proc_upgrading-the-external-database.adoc
```diff
@@ -0,0 +1,29 @@
...
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# {foreman-installer} \
+--foreman-db-host _newpostgres.example.com_ \
+--katello-candlepin-db-host _newpostgres.example.com_ \
+--foreman-proxy-content-pulpcore-postgresql-host _newpostgres.example.com_
+----
+. Copy the backup database to the new host.
+. Restore the backup database on the new host.
```
evgeni, line 29: These steps are already done above
```suggestion
```

guides/common/modules/proc_upgrading-the-external-database.adoc
```diff
@@ -0,0 +1,29 @@
...
+You can upgrade an external database from {EL} 7 to {EL} 8 while upgrading {Project} from {ProjectVersionPrevious} to {ProjectVersion}.
+
+.Prerequisites
+* Create a new {EL} 8 based host for PostgreSQL server that follows the external database on {EL} 8 documentation.
+ifdef::satellite[]
+For more information, see https://access.redhat.com/documentation/en-us/red_hat_satellite/{ProjectVersion}/html/installing_satellite_server_in_a_connected_network_environment/performing-additional-configuration#using-external-databases_satellite[Using External Databases with {Project}].
+endif::[]
+ifdef::katello[]
+For more information, see https://docs.theforeman.org/{ProjectVersion}/Installing_Server/index-katello.html#using-external-databases_foreman[Using External Databases with {Project}].
+endif::[]
```
evgeni, line 13: This can be written using macros in one go:
```suggestion
ifdef::katello,orcharhino,satellite[]
For more information, see {InstallingServerDocURL}using-external-databases_{project-context}[Using External Databases with {Project}].
endif::[]
```

(we do not include external DB for foreman builds, no idea why

evgeni (changes requested):

evgeni (approved):
Ack on the content
Should be squashed on merge

evgeni:
I am away from my ssh key, so not merging. Whoever has the time feel free to!

bangelic:
> I am away from my ssh key, so not merging. Whoever has the time feel free to!

@Lennonka or @adamlazik1 or @ekohl can anyone merge this today? Thanks in advance!

maximiliankolb:
   509be8f45..83c078c15  3.5 -> 3.5
   11e75a21b..cfacd0eaa  3.4 -> 3.4
   b087eb8fe..2c5b16274  3.3 -> 3.3
   8189ce835..a91479a4a  3.2 -> 3.2
   f33edb313..fed1766cc  3.1 -> 3.1

guides/common/modules/proc_upgrading-the-external-database.adoc
ekohl: > (we do not include external DB for foreman builds, no idea why

8710bde2c7185c057d7f9a71313d7c166dae5fda. When I was reading through it, it only worked if you had Katello. If we add more conditionals to hide the Katello-specifics (like candlepin & pulp's databases) it can be included.

PR #1641, opened 2022-09-19
Replace project with project-context in anchors

adamlazik1:
In reaction to [BZ#2126829](https://bugzilla.redhat.com/show_bug.cgi?id=2126829), I wonder whether it would be a desirable change to see "foreman" or "satellite" in the respective guide links instead of "project". Let me know what you think about this. Updated all ids and xrefs containing "project".

* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3


adamlazik1:
Replaced legacy inline anchors with newer syntax because for some reason they were breaking the build.

maximiliankolb (commented):
To my knowledge, two attributes in anchors/links will break satellite downstream docs. You might want to check this first.

Also, I'm questioning if referring to "Project" is even necessary. After all, this _is_ `{Project}` documentation.
Example: `enforcing-puppet-configuration-from-{project-context}-on-managed-hosts_{context}` Surely, "Enforcing Puppet Configuration on Managed Hosts" would be just as good.

adamlazik1:
@maximiliankolb Thank you for your suggestion! It is true, {project-context} could be removed from some anchors altogether. I will do additional changes if we decide to continue with this PR.
> To my knowledge, two attributes in anchors/links will break satellite downstream docs. You might want to check this first.
This is interesting. I did not encounter such issue when building locally. @ekohl Would you happen to have any information on this?

maximiliankolb (commented):
change looks reasonable. Please consult with @Lennonka if it's really necessary/works as intended.

Lennonka:
I tested one ID in our downstream build and it looks like it does NOT break the build (anymore)! :)
I think it is desirable from the SEO perspective.
FYI In the same spirit, it could also be desirable to replace "proxy", "smartproxy" and "capsule" with a similar variable. What do you think?

adamlazik1:
@Lennonka Thanks for confirming! :slightly_smiling_face: 
Indeed, those words you suggested could be also replaced with attributes. Because it is a lot of changes, I suggest a separate PR for each attribute added into anchors for easier PR reviews. In this PR, I will focus on replacing `project` with `{project-context}`.

maximiliankolb:
Thanks @Lennonka and @adamlazik1 for investigating!

Feel free to come up with some `find | xargs sed` command to handle this & put it into your commit message for speedy reviews.

adamlazik1:
@maximiliankolb I renamed the file that you suggested but I am not sure what other files to rename. Do you have any other suggestions?

maximiliankolb:
> @maximiliankolb I renamed the file that you suggested but I am not sure what other files to rename. Do you have any other suggestions?

I assume yu've checked for `enforcing-puppet-configuration-from-project-on-managed-hosts` which has changed?

Your change so far looks good; title, anchor, and file name match 👍

adamlazik1:
@maximiliankolb yes, indeed. I was also looking into if I could drop the {project-context} from any other anchors and file names but I wasn't really sure if there are any more. So this is ready for review. Feel free to point out any other files where you would like to see {project-context} being dropped from file name, title and anchor.

maximiliankolb (approved):
LGTM

maximiliankolb:
Merged to master and cherry-picked both commits: ✅ 
   d7046e038..68f1de296  3.4 -> 3.4
   765f26b08..08ee85565  3.3 -> 3.3
   9691c82d5..dbca1b4c3  3.2 -> 3.2

@adamlazik1 Please open another PR against 3.1 because there are merge conflicts when cherry-picking.

PR #1630, opened 2022-09-13
Add pull mode to remote job execution

Lennonka:
Take-over of #1584

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

PR #1646, opened 2022-09-21
Add a case in Remote Execution Workflow

mjivraja:
We mention how project searches for the remote execution proxy of a host and we cover all the points except one i.e. the "Prefer registered through proxy for remote execution" settings. Hence, we are adding a case where project prefers to execute remote workflow that are registered through proxies. From all the other processes that are mentioned in this section, this is the preferable way and it is the reason why we kept it in the starting of section.

Document the function of "Prefer registered through proxy for remote execution" option in "Remote Execution Workflow" section

https://bugzilla.redhat.com/show_bug.cgi?id=2123976

* [X] Foreman 3.5/Katello 4.7
* [X] Foreman 3.4/Katello 4.6
* [X] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/con_remote-execution-workflow.adoc
+++ b/guides/common/modules/con_remote-execution-workflow.adoc
@@ -16,9 +16,16 @@ endif::[]
 . {Project} finds remote execution {SmartProxies} assigned to these subnets.
 . From this set of {SmartProxies}, {Project} selects the {SmartProxy} that has the least number of running jobs.
 By doing this, {Project} ensures that the jobs load is balanced between remote execution {SmartProxies}.
-. If {Project} does not find a remote execution {SmartProxy} at this stage, and if the *Fallback to Any {SmartProxy}* setting is enabled, {Project} adds another set of {SmartProxies} to select the remote execution {SmartProxy} from.
+
+If you have enabled *remote_execution_prefer_registered_through_proxy*, {Project} runs the REX job using the {SmartProxy} the host is registered to.
+
+By default, *remote_execution_prefer_registered_through_proxy* is set to *No*.
+To enable it, in the {ProjectWebUI}, navigate to *Administer* > *Settings*, and on the *Content* tab, set `remote_execution_prefer_registered_through_proxy` to *Yes*.
+This ensures that {Project} performs REX jobs on hosts by the {SmartProxy} to which they are registered to.
+
+If {Project} does not find a remote execution {SmartProxy} at this stage, and if the *Fallback to Any {SmartProxy}* setting is enabled, {Project} adds another set of {SmartProxies} to select the remote execution {SmartProxy} from.
 {Project} selects the most lightly loaded {SmartProxy} from the following types of {SmartProxies} that are assigned to the host:
-+
+
 * DHCP, DNS and TFTP {SmartProxies} assigned to the host's subnets
 * DNS {SmartProxy} assigned to the host's domain
 * Realm {SmartProxy} assigned to the host's realm
@@ -26,5 +33,4 @@ By doing this, {Project} ensures that the jobs load is balanced between remote e
 * Puppet CA {SmartProxy}
 * OpenSCAP {SmartProxy}
 
-+
-. If {Project} does not find a remote execution {SmartProxy} at this stage, and if the *Enable Global {SmartProxy}* setting is enabled, {Project} selects the most lightly loaded remote execution {SmartProxy} from the set of all {SmartProxies} in the host's organization and location to execute a remote job.
+If {Project} does not find a remote execution {SmartProxy} at this stage, and if the *Enable Global {SmartProxy}* setting is enabled, {Project} selects the most lightly loaded remote execution {SmartProxy} from the set of all {SmartProxies} in the host's organization and location to execute a remote job.


mjivraja:
This PR is in the reference to #1630

mjivraja:
This PR is referring to #1630

guides/common/modules/con_remote-execution-workflow.adoc
```diff
@@ -1,7 +1,11 @@
 [id="remote-execution-workflow_{context}"]
 = Remote Execution Workflow
 
-When you run a remote job on hosts, for every host, {Project} performs the following actions to find a remote execution {SmartProxy} to use.
+{Project} prefers remote execution on host's registered_through proxy.
+You must *enable remote_execution_prefer_registered_through_proxy* manually from *Administer > Settings > Content*.
+This will allow selecting a {SmartProxies} to perform the REX job on the hosts to which they are registered to.
```
AkshayGadhaveRH, line 6:
```suggestion
This will select the {SmartProxy} to which the hosts are registered, to perform the REX job.
```

AkshayGadhaveRH: @mjivraja I think this sounds better. What do you think @maximiliankolb ?

guides/common/modules/con_remote-execution-workflow.adoc
```diff
@@ -1,7 +1,11 @@
...
+
+For the case where the above setting is disabled, you can run a remote job on hosts, for every host, {Project} performs the following actions to find a remote execution {SmartProxy} to use.
```
AkshayGadhaveRH, line 8:
```suggestion
For the case where the above setting is disabled, for every host, {Project} performs the following actions to find a remote execution {SmartProxy} to use.
```

guides/common/modules/con_remote-execution-workflow.adoc
mjivraja: Thanks @AkshayGadhaveRH 
Wait for Maxi if he has something for us.

Lennonka:
Should this be cherry-picked to 3.2 as well? I think the same cherry-picks should be same as in #1630 , but I'm not sure about 3.2. Help, please? :)

guides/common/modules/con_remote-execution-workflow.adoc
maximiliankolb: I think the whole first paragraph should be reworded.

a) describe what REX is
b) how SmartProxies play their part to run jobs on managed hosts.

Assuming a host is always only registered to one smartproxy, either Foreman or a Smart Proxy, then describing that you _can_ use the "registered through" Smart Proxy sounds weird to me.

maximiliankolb (changes requested):

mjivraja:
> Should this be cherry-picked to 3.2 as well? I think the cherry-picks should be the same as in #1630 , but I'm not sure about 3.2. EDIT: Verified that 3.2 is correct in this PR, but it's different for #1630.

According to @adamruzicka , it should also be part of 3.1. Please correct me if  I am wrong.

guides/common/modules/con_remote-execution-workflow.adoc
```diff
@@ -1,7 +1,11 @@
 [id="remote-execution-workflow_{context}"]
 = Remote Execution Workflow
 
-When you run a remote job on hosts, for every host, {Project} performs the following actions to find a remote execution {SmartProxy} to use.
+{Project} prefers remote execution on host's registered_through proxy.
```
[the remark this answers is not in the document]
mjivraja: > what is "registered_through"? Please reword to make it less technical.

The original idea from a reporter was to keep it like this. But, we are Okay to reword it.

> describe what REX is

We have used the phrase **Remote Execution** instead of REX.

> how SmartProxies play their part to run jobs on managed hosts.

Will again reword it. :+1:

guides/common/modules/con_remote-execution-workflow.adoc
mjivraja: Addressed :+1:

guides/common/modules/con_remote-execution-workflow.adoc
mjivraja: Addressed. :+1: 
Kept `SmartProxies`

guides/common/modules/con_remote-execution-workflow.adoc
```diff
@@ -1,7 +1,11 @@
...
+You must *enable remote_execution_prefer_registered_through_proxy* manually from *Administer > Settings > Content*.
```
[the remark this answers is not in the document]
mjivraja: Addressed :+1:

guides/common/modules/con_remote-execution-workflow.adoc
```diff
@@ -1,7 +1,12 @@
 [id="remote-execution-workflow_{context}"]
 = Remote Execution Workflow
 
-When you run a remote job on hosts, for every host, {Project} performs the following actions to find a remote execution {SmartProxy} to use.
+{Project} prefers remote execution on hosts that are registered through proxy.
```
maximiliankolb, line 4:
```suggestion
{Project} prefers remote execution on hosts that are registered through a {SmartProxy}.
```

IMHO this is not true, because `Prefer registered through proxy for remote execution` defaults to "no" on Foreman 3.3.

mjivraja: Shall I go with the `Smart Proxy` that you suggested?

maximiliankolb: I still believe this is not entirely true because you want to have this change cherry-picked to 3.3 & I've checked the default behaviour on Foreman.

mjivraja: Reworded that line :+1:

guides/common/modules/con_remote-execution-workflow.adoc
```diff
@@ -1,7 +1,12 @@
 [id="remote-execution-workflow_{context}"]
 = Remote Execution Workflow
 
-When you run a remote job on hosts, for every host, {Project} performs the following actions to find a remote execution {SmartProxy} to use.
+Remote execution can be performed on the hosts that are registered through a {SmartProxy}.
```
maximiliankolb, line 4: You can also perform REX directly through `{ProjectServer}`.

adamruzicka: I'm not 100% sure about what those curly placeholders get filled with, but REX jobs always go through a smart proxy, even though it may be colocated on the same machine

mjivraja: > I'm not 100% sure about what those curly placeholders get filled with.

Curly brackets will replace the Project name as those are attributes. :slightly_smiling_face: 

> but REX jobs always go through a smart proxy, even though it may be colocated on the same machine

@maximiliankolb I am confused :smile: +

maximiliankolb: I guess what I'm trying to say: You can also attach a host to Foreman; so REX works through Foreman's internal Smart Proxy.

Foreman would be your ProjectServer, but not a "SmartProxyServer".

```suggestion
Remote execution can be performed on managed hosts through {SmartProxies}.
```

guides/common/modules/con_remote-execution-workflow.adoc
```diff
@@ -1,7 +1,12 @@
...
+You must enable *remote_execution_prefer_registered_through_proxy* manually.
```
maximiliankolb, line 5: This still needs to be reworded IMHO.

When describing in this concept _what_ REX is, we should not immediately start with "you'll probably what to use this very special feature; here's how to activate it".

Can you please provide some more details on _why_ we recommend enabling `remote_execution_prefer_registered_through_proxy`?

I'd also switch the order and add this below the procedure starting in line 11.

mjivraja: I guess we dropped the idea of `recommending`. Instead, we kept that way optional for end-user. I hope I understood correctly from earlier comments. 

For other changes that you suggested, I am Okay with that :+1:

guides/common/modules/con_remote-execution-workflow.adoc
mjivraja: Addressed :+1:

guides/common/modules/con_remote-execution-workflow.adoc
mjivraja: Hi @maximiliankolb 

After my further discussion with SME, we are in agreement to provide some more details on why we recommend `enabling remote_execution_prefer_registered_through_proxy`?

I will add more text for the same. :+1:

mjivraja: Addressed :+1:

guides/common/modules/con_remote-execution-workflow.adoc
```diff
@@ -1,7 +1,16 @@
...
 = Remote Execution Workflow
 
-When you run a remote job on hosts, for every host, {Project} performs the following actions to find a remote execution {SmartProxy} to use.
+Remote execution can be performed on managed hosts through {Project} as well as {SmartProxies}.
+Clients registered through one {SmartProxy} might not be able to communicate through other {SmartProxies} or the {project} directly and vice-versa.
+As a result, the remote execution tasks must be performed through respective {SmartProxies} only that can communicate with the client host.
+Therefore, enabling _remote_execution_prefer_registered_through_proxy_ should allow {Project} to run REX jobs on clients that are registered through their respective {SmartProxies}.
+
+You must enable *remote_execution_prefer_registered_through_proxy manually*.
+Navigate to *Administer > Settings*, enable the option from *Content* tab.
```
maximiliankolb, line 10:
```suggestion
Navigate to *Administer > Settings* and on the *Content* tab, set `remote_execution_prefer_registered_through_proxy` to true.
```

maximiliankolb (changes requested):
I am still unsure why we'd want to recommend changing the default value.

I suggest the following order: introduction to REX, then default use case/workflow, and then describe that there is a potentially very useful/required setting; what it is and how to change it.

Maybe this should be changed "in code"?

mjivraja:
> I am still unsure why we'd want to recommend changing the default value.

We are adding this new option to let end-user utilize it in-case if they do not get the expected behavior/output from default one i.e `disabled`.

> I suggest the following order: introduction to REX, then default use case/workflow, and then describe that there is a potentially very useful/required setting; what it is and how to change it.

I will follow this sequence, add/modify some parts :+1: 

> Maybe this should be changed "in code"?

I am not sure if there is anything required from the development side.

guides/common/modules/con_remote-execution-workflow.adoc
mjivraja: Addressed :+1:

guides/common/modules/con_remote-execution-workflow.adoc
```diff
@@ -1,7 +1,16 @@
 [id="remote-execution-workflow_{context}"]
 = Remote Execution Workflow
 
-When you run a remote job on hosts, for every host, {Project} performs the following actions to find a remote execution {SmartProxy} to use.
+Remote execution can be performed on managed hosts through {Project} as well as {SmartProxies}.
+Clients registered through one {SmartProxy} might not be able to communicate through other {SmartProxies} or the {project} directly and vice-versa.
+As a result, the remote execution tasks must be performed through respective {SmartProxies} only that can communicate with the client host.
+Therefore, enabling _remote_execution_prefer_registered_through_proxy_ should allow {Project} to run REX jobs on clients that are registered through their respective {SmartProxies}.
+
+You must enable *remote_execution_prefer_registered_through_proxy manually*.
```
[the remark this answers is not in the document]
mjivraja: Addressed :+1:

mjivraja:
> I suggest the following order: introduction to REX, then default use case/workflow, and then describe that there is a potentially very useful/required setting; what it is and how to change it.

`Introduction to REX` is already covered in the previous section `About Running Jobs on Hosts`. So, I started straight with the second one.

Lennonka (changes requested):
At which step do the workflows divert regarding this setting?
If set to true, does this setting circumvent the whole SmartProxy selection and Project selects the registered-through SmartProxy right away instead of searching through the subnets/SmartProxies? (You should probably ask a developer, the reporter is usually not a SME.)

If that's the case, it would be better to keep the steps as they are and add this description below the default workflow as Max suggests. (btw, by introduction, Max probably means just the introductory paragraph in this module)

mjivraja:
>  (btw, by introduction, Max probably means just the introductory paragraph in this module)

I am still wondering if we need to add more in the introduction paragraph here. That particular part has been covered in the very previous section.

> At which step do the workflows divert regarding this setting?

The second option that we added is not mandatory for end-users and hence, it's just another way. 

> If set to true, does this setting circumvent the whole SmartProxy selection and Project selects the registered-through SmartProxy right away instead of searching through the subnets/SmartProxies?

I guess, Yes. Since we document all the parameters that help with REX capsule selection, these settings need to be documented as one of them as well.

> If that's the case, it would be better to keep the steps as they are and add this description below the default workflow as Max suggests.

Sure. I am open to it but will wait for your view before taking changes.

mjivraja:
Hi @maximiliankolb 

What can be your view on the above comments from Lena?

guides/common/modules/con_remote-execution-workflow.adoc
```diff
@@ -16,15 +16,27 @@ endif::[]
 . {Project} finds remote execution {SmartProxies} assigned to these subnets.
 . From this set of {SmartProxies}, {Project} selects the {SmartProxy} that has the least number of running jobs.
 By doing this, {Project} ensures that the jobs load is balanced between remote execution {SmartProxies}.
-. If {Project} does not find a remote execution {SmartProxy} at this stage, and if the *Fallback to Any {SmartProxy}* setting is enabled, {Project} adds another set of {SmartProxies} to select the remote execution {SmartProxy} from.
+
+However, there is one more way to run REX jobs apart from the default one mentioned above.
+Remote execution can be performed on managed hosts through {Project} as well as {SmartProxies}.
+Clients registered through one {SmartProxy} might not be able to communicate through other {SmartProxies} or the {project} directly and vice-versa.
+As a result, the remote execution tasks must be performed through respective {SmartProxies} only that can communicate with the client host.
+Therefore, enabling _remote_execution_prefer_registered_through_proxy_ should allow {Project} to run REX jobs on clients that are registered through their respective {SmartProxies}.
```
maximiliankolb, line 24:
```suggestion
If you have enabled _Prefer registered through proxy for remote execution_, {Project} runs the REX job using the {SmartProxy} the host is registed to.
```

guides/common/modules/con_remote-execution-workflow.adoc
```diff
@@ -16,15 +16,27 @@ endif::[]
...
+
+For this option, you must enable *remote_execution_prefer_registered_through_proxy* manually.
+Navigate to *Administer > Settings* and on the *Content* tab, set `remote_execution_prefer_registered_through_proxy` to true.
+This will select the {SmartProxies} to which the hosts are registered, to perform the REX job.
```
maximiliankolb, line 28:
```suggestion
By default, _Prefer registered through proxy for remote execution_ is set to *No*.
To enable it, in the {ProjectWebUI}, navigate to *Administer* > *Settings*, and on the *Content* tab, set `remote_execution_prefer_registered_through_proxy` to *Yes*.
This ensures that {Project} performs REX jobs on hosts by the {SmartProxy} to which they are registered to.
```

guides/common/modules/con_remote-execution-workflow.adoc
```diff
@@ -16,15 +16,27 @@ endif::[]
...
+
+Therefore, you can use either way to perform REX jobs on hosts.
+
```
maximiliankolb, line 31:
```suggestion
```

maximiliankolb (commented):
Hey @mjivraja
I can reproduce the REX case you've described. I would kindly ask you to find out in with version of Foreman this is possible. I've only checked on Foreman 3.5.

I made some suggestions on the wording. I would try to keep it shorter. I am unsure if it makes sense to move the "enable this feature/setting" to a separate procedure.

mjivraja:
> I can reproduce the REX case you've described. I would kindly ask you to find out in with version of Foreman this is possible. I've only checked on Foreman 3.5.

I have updated the applicable versions in the description of this PR, confirmed by SME :+1:

guides/common/modules/con_remote-execution-workflow.adoc
```diff
@@ -16,15 +16,22 @@ endif::[]
 . {Project} finds remote execution {SmartProxies} assigned to these subnets.
 . From this set of {SmartProxies}, {Project} selects the {SmartProxy} that has the least number of running jobs.
 By doing this, {Project} ensures that the jobs load is balanced between remote execution {SmartProxies}.
-. If {Project} does not find a remote execution {SmartProxy} at this stage, and if the *Fallback to Any {SmartProxy}* setting is enabled, {Project} adds another set of {SmartProxies} to select the remote execution {SmartProxy} from.
+
+If you have enabled *remote_execution_prefer_registered_through_proxy*,
+{Project} runs the REX job using the {SmartProxy} the host is registered to.
```
maximiliankolb, line 21:
```suggestion
If you have enabled *remote_execution_prefer_registered_through_proxy*, {Project} runs the REX job using the {SmartProxy} the host is registered to.
```

maximiliankolb (commented):
One small comment, rest LGTM!

guides/common/modules/con_remote-execution-workflow.adoc
mjivraja: Thanks :+1:

maximiliankolb (approved):
LGTM!

maximiliankolb:
   d2aa9ae9f..9fce923b7  3.5 -> 3.5
   cb4c03b5e..0267e6ac2  3.4 -> 3.4
   1194f7849..ad2755e03  3.3 -> 3.3

PR #1650, opened 2022-09-22
Update steps for creating local/remote file repo

dubewarsagar:
Steps are updated for creating a local/remote file repository.
The tool will be provided by utils repo (not client).
Hence steps updated for RHEL 8 and 7.
Added 2 new variables for downstream docs in attribute-satellite.
Also, Pulp Manifest package name is different for RHEL 7 and 8.

https://bugzilla.redhat.com/show_bug.cgi?id=2123940

* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3


guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -14,14 +14,30 @@ To create a file type repository in a local directory, complete the following pr
...
+--enable={RepoRHEL8AppStream} \
+--enable={RepoRHEL8ServerSatelliteUtils}
 ----
 endif::[]
 . Install the Pulp Manifest package:
+** For RHEL 7:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {package-install-project} tfm-pulpcore-python-pulp_manifest
```
ekohl, line 38:
```suggestion
# {package-install-project} tfm-pulpcore-python3-pulp_manifest
```

guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -14,14 +14,30 @@ To create a file type repository in a local directory, complete the following pr
 
 . Ensure the Server and {project-client-name} repositories are enabled.
 ifdef::satellite[]
+** For RHEL 7:
```
ekohl, line 17: I'd always list RHEL 8 first. We should encourage users to use the latest version.

guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -14,14 +14,30 @@ To create a file type repository in a local directory, complete the following pr
...
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {package-install-project} tfm-pulpcore-python-pulp_manifest
+----
+** For RHEL 8:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
```
ekohl, line 43: Below the block are also instructions for RHEL 7 to install it without downtime (and with `yum install install`). Should those also be duplicated?

Also, I wonder why the instructions below disable the repo again after installing it. That looks bad.

Additionally, since it uses `foreman-maintain` and package locking only happens in Satellite (by default), the whole block should be behind `ifdef::satellite[]`.

ekohl (changes requested):

guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -34,16 +45,40 @@ Alternatively, to prevent downtime caused by stopping the service, you can use t
...
 ifdef::satellite[]
-# subscription-manager repos --enable {RepoRHEL7ServerSatelliteToolsProductVersion}
+# subscription-manager repos --enable {RepoRHEL8ServerSatelliteUtils}
 endif::[]
 # {foreman-maintain} packages unlock
-# yum install install python-pulp-manifest -y
+# yum install install python-pulp3-manifest -y
```
maximiliankolb, line 51:
```suggestion
# {package-install} python-pulp3-manifest
```

maximiliankolb (commented):
I could not find the packages on yum.theforeman.org or install them on Foreman 3.3 on EL8.

Can you please provide a link so I can reproduce this?

guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -48,7 +48,7 @@ ifdef::satellite[]
 # subscription-manager repos --enable {RepoRHEL8ServerSatelliteUtils}
 endif::[]
 # {foreman-maintain} packages unlock
-# yum install install python-pulp3-manifest -y
+# {package-install} python-pulp3-manifest
```
adamlazik1, line 51:
```suggestion
# {package-install} python-pulp3-manifest -y
```

adamlazik1: Sorry, I didn't notice `-y` option removed in the previous review. @maximiliankolb is including `-y` for automatic installation not desirable?

maximiliankolb: IMHO we rarely do that, so I'd rather say "no, not desirable".

adamlazik1: Alright. Sagar, please ignore this review.

lpramuk:
> I could not find the packages on yum.theforeman.org or install them on Foreman 3.3 on EL8.
> 
> Can you please provide a link so I can reproduce this?

This is part of PulpCore so you should be looking into pulpcore repos

guides/common/modules/proc_creating-a-local-file-repository.adoc
ekohl:
```suggestion
# {package-install} python3-pulp_manifest
```

guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -14,14 +14,25 @@ To create a file type repository in a local directory, complete the following pr
...
 [options="nowrap" subs="+quotes,attributes"]
 ----
 # subscription-manager repos --enable={RepoRHEL7Server} \
---enable={project-client-RHEL7-url}
+--enable={RepoRHEL7ServerSatelliteUtils}
 ----
 endif::[]
+ifdef::satellite,orcharhino[]
 . Install the Pulp Manifest package:
+** For RHEL 8:
```
ekohl, line 35: I'd drop this line since there's no list or anything.
```suggestion
```

guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -14,14 +14,25 @@ To create a file type repository in a local directory, complete the following pr
 
 . Ensure the Server and {project-client-name} repositories are enabled.
```
ekohl, line 15: This is not correct for upstream. There it's in the pulpcore repositories instead. I'd suggest swapping this line and the next so it's only shown for Satellite.

guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -34,16 +45,40 @@ Alternatively, to prevent downtime caused by stopping the service, you can use t
 [options="nowrap" subs="+quotes,attributes"]
 ----
 ifdef::satellite[]
```
ekohl, line 47: This ifdef should be higher. Right above "Note that this command". Also the whole block where it uses `{foreman-maintain}` is satellite-only.

guides/common/modules/proc_creating-a-remote-file-repository.adoc
```diff
@@ -27,18 +27,34 @@ To create a file type repository in a remote directory, complete the following p
...
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
 # subscription-manager repos --enable={RepoRHEL7Server} \
---enable={project-client-RHEL7-url}
+--enable={RepoRHEL7ServerSatelliteUtils}
 ----
 endif::[]
 . Install the Pulp Manifest package:
+** For RHEL 8:
```
ekohl, line 47:
```suggestion
** For {EL} 8:
```

guides/common/modules/proc_creating-a-remote-file-repository.adoc
```diff
@@ -27,18 +27,34 @@ To create a file type repository in a remote directory, complete the following p
...
 +
-[options="nowrap" subs="+quotes"]
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {package-install-project} python3-pulp_manifest
+----
+** For RHEL 7:
```
ekohl, line 53:
```suggestion
** For {EL} 7:
```

guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -34,16 +45,40 @@ Alternatively, to prevent downtime caused by stopping the service, you can use t
...
+# {package-install-project} tfm-pulpcore-python3-pulp_manifest
+----
++
+Note that this command stops the {Project} service and re-runs {foreman-installer}.
+Alternatively, to prevent downtime caused by stopping the service, you can use the following:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+ifdef::satellite[]
+# subscription-manager repos --enable {RepoRHEL8ServerSatelliteUtils}
```
ekohl, line 71: Previously the repositories were already enabled. Any reason they're enabled here again (other than that is was done previously for EL7)?

I'd suggest to drop the enable/disable inline here.

guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -34,16 +45,40 @@ Alternatively, to prevent downtime caused by stopping the service, you can use t
...
+endif::[]
+# {foreman-maintain} packages unlock
+# yum install install tfm-pulpcore-python3-pulp_manifest
+# {foreman-maintain} packages lock
+ifdef::satellite[]
+# subscription-manager repos --disable {RepoRHEL8ServerSatelliteUtils}
```
ekohl, line 77: Disabling feels like a bad idea to me. It means you can't get updated versions.

ekohl (changes requested):
I checked and we have http://koji.katello.org/koji/buildinfo?buildID=73042 but that's only in Pulpcore 3.16, but we ship 3.18. It was added in https://github.com/theforeman/pulpcore-packaging/pull/295 where it was also noted that it's missing for newer versions. There is https://github.com/theforeman/pulpcore-packaging/pull/562 but nothing for 3.18.

guides/common/modules/proc_creating-a-local-file-repository.adoc
dubewarsagar: Steps are separated here for RHEL 7 and 8. Hence this bullet points are used.

guides/common/modules/proc_creating-a-local-file-repository.adoc
dubewarsagar: Sent inside `ifdef::satellite[]`

guides/common/modules/proc_creating-a-local-file-repository.adoc
dubewarsagar: Done.

guides/common/modules/proc_creating-a-local-file-repository.adoc
dubewarsagar: dropped.

guides/common/modules/proc_creating-a-local-file-repository.adoc
dubewarsagar: dropped.

guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -12,38 +12,65 @@ To create a file type repository in a directory on a remote server, see xref:Cre
...
+** For {EL} 7:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
 # subscription-manager repos --enable={RepoRHEL7Server} \
---enable={project-client-RHEL7-url}
+--enable={RepoRHEL7ServerSatelliteUtils}
 ----
 endif::[]
+ifdef::satellite,orcharhino[]
```
ekohl, line 33: Why is this limited to satellite/orcharhino?

guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -12,38 +12,65 @@ To create a file type repository in a directory on a remote server, see xref:Cre
...
+----
++
+ifdef::satellite[]
+Note that this command stops the {Project} service and re-runs {foreman-installer}.
+Alternatively, to prevent downtime caused by stopping the service, you can use the following:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foreman-maintain} packages unlock
+# yum install install tfm-pulpcore-python3-pulp_manifest
```
ekohl, line 68:
```suggestion
# yum install tfm-pulpcore-python3-pulp_manifest
```

guides/common/modules/proc_creating-a-remote-file-repository.adoc
```diff
@@ -27,18 +27,34 @@ To create a file type repository in a remote directory, complete the following p
 
 . On your remote server, ensure that the Server and {project-client-name} repositories are enabled.
 ifdef::satellite[]
```
ekohl, line 29: I think these should be swapped.

ekohl (changes requested):
There are a lot of similarities between the two procedures. Is this something that's commonly extracted to a snippet?

Also note that in upstream we don't support EL7 anymore. I'd be OK with a PR that first adds EL7 and EL8 instructions and then a PR that kills off EL7, but please remember that right now technically it contains redundant instructions for the branch you're submitting a PR to.

guides/common/modules/proc_creating-a-local-file-repository.adoc
dubewarsagar: please see the comment above - https://github.com/theforeman/foreman-documentation/pull/1650#discussion_r977769517

ekohl: That doesn't answer the question. Why is this not shown for katello? Or rather, why is this whole `ifdef` even here?

dubewarsagar: By the response of the @maximiliankolb above, I supposed this should be applicable for satellite,orcharhino. So I put the condition there.
Do you suggest to remove that condition completely and make this step of `python3-pulp_manifest` available for all versions?

ekohl: I interpreted what @maximiliankolb said as "please make sure it's also there for orcharhino", but I think it should have been stronger and applied to all builds.

ekohl:
@dubewarsagar you don't need to request a review when you place a comment. That only generates an extra email in my inbox, which I then need to read as well. Please only use the feature when a review is forgotten or you need an additional person to take a look.

guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -12,38 +12,65 @@ To create a file type repository in a directory on a remote server, see xref:Cre
...
-endif::[]
 # {foreman-maintain} packages unlock
-# yum install install python-pulp-manifest -y
+# {package-install} python3-pulp_manifest
 # {foreman-maintain} packages lock
-ifdef::satellite[]
-# subscription-manager repos --disable {RepoRHEL7ServerSatelliteToolsProductVersion}
 endif::[]
 ----
 This installs the package without downtime.
```
Lennonka, line 53: Watch the syntax:
```
----
+
This installs the package without downtime.
```

guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -12,38 +12,65 @@ To create a file type repository in a directory on a remote server, see xref:Cre
...
+** For {EL} 7:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {package-install-project} tfm-pulpcore-python3-pulp_manifest
+----
++
```
Lennonka, line 60: This line should be inside the condition, i.e. one line down.

guides/common/modules/proc_creating-a-remote-file-repository.adoc
```diff
@@ -25,20 +25,36 @@ endif::[]
...
+----
+** For {EL} 7:
++
+[options="nowrap" subs="+quotes,attributes"]
 ----
-# yum install python3-pulp_manifest
+# {package-install-project} tfm-pulpcore-python3-pulp_manifest
 ----
 . Create a directory that you want to use as the file type repository in the HTTP server's public folder:
 +
```
Lennonka, line 60: Remove `+`.

guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -12,38 +12,65 @@ To create a file type repository in a directory on a remote server, see xref:Cre
...
 endif::[]
+ifdef::satellite,orcharhino[]
 . Install the Pulp Manifest package:
+** For {EL} 8:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
 # {package-install-project} python3-pulp_manifest
 ----
 +
```
Lennonka, line 41: This line should be inside the condition, i.e. one line down.

Lennonka (changes requested):
The build is failing because you have mistakes in the syntax. I tried to point out what might fix it.

dubewarsagar:
> @dubewarsagar you don't need to request a review when you place a comment. That only generates an extra email in my inbox, which I then need to read as well. Please only use the feature when a review is forgotten or you need an additional person to take a look.

ok

guides/common/modules/proc_creating-a-local-file-repository.adoc
dubewarsagar: ok.

guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -12,38 +12,65 @@ To create a file type repository in a directory on a remote server, see xref:Cre
...
 # {foreman-maintain} packages unlock
-# yum install install python-pulp-manifest -y
+# {package-install} python3-pulp_manifest
 # {foreman-maintain} packages lock
-ifdef::satellite[]
-# subscription-manager repos --disable {RepoRHEL7ServerSatelliteToolsProductVersion}
+----
 endif::[]
++
+This installs the package without downtime.
```
ekohl, line 53: This must be in the above `ifdef`. Or it can be dropped since on line 43 there's already "to prevent downtime".

guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -12,38 +12,65 @@ To create a file type repository in a directory on a remote server, see xref:Cre
...
+** For {EL} 7:
```
ekohl, line 54: Have you thought about the follow up? Will you submit a PR after this to drop EL7 instructions on current branches? I mostly ask because it's broken now on Katello where it tells users to run `dnf install` which doesn't work on EL7. On the other hand, we can't hardcode `yum install` since that breaks the Satellite case.

ekohl (changes requested):

guides/common/modules/proc_creating-a-local-file-repository.adoc
dubewarsagar: > Have you thought about the follow up? Will you submit a PR after this to drop EL7 instructions on current branches? I mostly ask because it's broken now on Katello where it tells users to run `dnf install` which doesn't work on EL7. On the other hand, we can't hardcode `yum install` since that breaks the Satellite case.

Then we should drop instructions for RHEL 7 in both the steps, correct? which we added as a part of this PR...

@lpramuk could you please confirm if we can get rid of the RHEL 7 instructions as suggested here?

ekohl: It depends. You specified you want to cherry pick this to 3.1, where it is supported. See https://github.com/theforeman/foreman-documentation/milestone/2 for the whole table. I think you will need to write a patch for every branch with various `ifdef` statements. At least for `master` documenting EL7 makes no sense since no platform supports it anymore. It can make sense to first submit a change to document it, cherry pick that and then drop it again but it may be faster to only document EL8 in master, cherry pick that to 3.4. Then submit a different PR to 3.3 and yet another to 3.1 (Foreman 3.2 is EOL so you don't have to do that PR).

dubewarsagar: > It depends. You specified you want to cherry pick this to 3.1, where it is supported. See https://github.com/theforeman/foreman-documentation/milestone/2 for the whole table. I think you will need to write a patch for every branch with various `ifdef` statements. At least for `master` documenting EL7 makes no sense since no platform supports it anymore. It can make sense to first submit a change to document it, cherry pick that and then drop it again but it may be faster to only document EL8 in master, cherry pick that to 3.4. Then submit a different PR to 3.3 and yet another to 3.1 (Foreman 3.2 is EOL so you don't have to do that PR).

IMO, if we are going to remove RHEL 7 support over the time, we should refrain from adding new such references. But I would wait for @lpramuk to revert on this.
Usually we keep docs for the latest release supported. In this case, RHEL 8 info should be preserved only.

lpramuk: @dubewarsagar You can't drop EL7 as you state you want to cherrypick to 3.4, 3.3, 3.2 ,3.1
And EL8 only versions are 3.5 (nightly), 3.4  

RE yum VS. dnf
you differentiate in rpm name already so you can change also command for el8 to `dnf` for el7 to `yum`

But generally it is better to refer always to yum as it work on both EL7 and EL8 (where yum is symlinked to dnf)
RHEL8 documentation does so (uses yum exclusively)
I get this is upstream so many strive to have `dnf` for EL8 - but then you are required to split all steps for each EL

lpramuk: @dubewarsagar I would proceed with keeping EL7 in order to cherrypick to older versions (`3.3`, `3.1`) and then you can do EL7 removal for `3.5 nightly` and `3.4`

dubewarsagar: > @dubewarsagar I would proceed with keeping EL7 in order to cherrypick to older versions (`3.3`, `3.1`) and then you can do EL7 removal in another PR for `3.5 nightly` and `3.4`

Sounds good to me. 
@ekohl should we merge this now and then I will create another PR for EL7 removal in another PR for `3.5 nightly` and `3.4`?

ekohl: Given how much they differ I'd personally open a new PR for every branch. There have been too many changes around macros etc so I'm not going to blindly cherry pick this. A rendered preview is really useful. If you're going to do that, you might as well do it right the first time.

Note that you can still reuse this work by using `git cherry-pick` manually. For example:
```shell
# assuming origin is your remote to this repo
git checkout -b local-repo-for-3.3 origin/3.3-stable
git cherry-pick BZ-2123940-Custom-File-Type-Repo-Utils
```

The biggest reason I'd go for the correct instructions the first time is that right after merging it will be live. That means users can see invalid instructions. I really prefer to avoid that and get into a pattern where we only merge work that's ready for consumption.

dubewarsagar: Created a new https://github.com/theforeman/foreman-documentation/pull/1671 for foreman 3.1 and 3.3. 
Once #1671 gets merged will update information in this PR#1650 to remove the EL 7 related instructions to keep it specifically for 3.4 and above. 

@ekohl pls take a look and see if the cherry-picking versions are selected correctly in #1671.

Thanks!

guides/common/modules/proc_creating-a-local-file-repository.adoc
```diff
@@ -12,38 +12,62 @@ To create a file type repository in a directory on a remote server, see xref:Cre
 
 To create a file type repository in a local directory, complete the following procedure:
 
-. Ensure the Server and {project-client-name} repositories are enabled.
 ifdef::satellite[]
+. Ensure the Utils repositories are enabled.
```
ehelms, line 16: Ought this be singular as there is only one Utils repository?

guides/common/modules/proc_creating-a-local-file-repository.adoc
ehelms: I am not following the logic for EL7 removal. This is about the Satellite Utils repository which exists and will exist for RHEL 7 and RHEL 8. Therefore I see no need to swing through and remove the EL7 part. While the server may drop EL7 support, the intent of the Utils repository is to support more OS footprints that users may want or need to install tooling on.

guides/common/modules/proc_creating-a-local-file-repository.adoc
dubewarsagar: Changed to singular and pushed changes, While pushing changes got merge conflicts were resolved. Please take a look at the commit `merge conflicts` `470abf5` that they are not breaking any existing content.

guides/common/modules/proc_creating-a-local-file-repository.adoc
dubewarsagar: Maybe @ekohl can comment on this.

ekohl: I mostly suggested it because of these two lines (from line 8):
> Use this procedure to configure a repository in a directory on the base system where {Project} is installed.
> To create a file type repository in a directory on a remote server, see xref:Creating_a_Remote_File_Type_Repository_{context}[].

Since we can only install the base system on EL8, documenting EL7 is only going to add confusion.

dubewarsagar: so are you saying we should retain EL 7 information in all the versions? 
If yes, we can merge this into master right away and there is no need of PR#1671.

guides/common/attributes-satellite.adoc
```diff
@@ -108,6 +108,8 @@
 :RepoRHEL7ServerSatelliteToolsProductVersion: rhel-7-server-satellite-client-6-rpms
 :RepoRHEL8ServerSatelliteToolsProductVersion: satellite-client-6-for-rhel-8-<arch>-rpms
 :RepoRHEL9ServerSatelliteToolsProductVersion: satellite-client-6-for-rhel-9-<arch>-rpms
+:RepoRHEL7ServerSatelliteUtils: rhel-7-server-satellite-utils-6.11-rpms
```
maximiliankolb, line 111:
```suggestion
:RepoRHEL7ServerSatelliteUtils: rhel-7-server-satellite-utils-{RepoSatelliteVersion}-rpms
```

guides/common/attributes-satellite.adoc
```diff
@@ -108,6 +108,8 @@
...
+:RepoRHEL8ServerSatelliteUtils: satellite-utils-6.11-for-rhel-8-x86_64-rpms
```
maximiliankolb, line 112:
```suggestion
:RepoRHEL8ServerSatelliteUtils: satellite-utils-{RepoSatelliteVersion}-for-rhel-8-x86_64-rpms
```

guides/common/modules/proc_creating-a-local-file-repository.adoc
ekohl: I'm saying that in this file (local installation) we only need EL8. In the remote case we can still document EL7.

dubewarsagar: > I'm saying that in this file (local installation) we only need EL8. In the remote case we can still document EL7.

Got it. Removed EL 7 related information this module. Copying the same changes in PR#1671.

guides/common/modules/proc_creating-a-remote-file-repository.adoc
```diff
@@ -25,24 +25,40 @@ endif::[]
 
 To create a file type repository in a remote directory, complete the following procedure:
 
-. On your remote server, ensure that the Server and {project-client-name} repositories are enabled.
 ifdef::satellite[]
+. On your remote server, ensure that the Server and {project-client-name} repositories are enabled.
```
ehelms, line 29: Should we change `Server` here to one of the EL macros? Server can be confusing especially given RHEL 8 calls it Base OS and AppStream  @ekohl

ekohl: Perhaps not even specify the names at all. Just say:
```suggestion
. On your machine, ensure that the right repositories are enabled.
```

This is because utils could in theory also run from a desktop (at least with EL8).

guides/common/modules/proc_creating-a-remote-file-repository.adoc
```diff
@@ -25,24 +25,40 @@ endif::[]
...
 ----
-# yum install python3-pulp_manifest
+# {package-install-project} python3-pulp_manifest
 ----
-. Create a directory that you want to use as the file type repository in the HTTP server's public folder:
+** For {EL} 7:
 +
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {package-install-project} tfm-pulpcore-python3-pulp_manifest
```
maximiliankolb, line 57: See https://github.com/theforeman/foreman-documentation/pull/1671#discussion_r991870096

maximiliankolb (changes requested):
Please rebase to master.

guides/common/modules/proc_creating-a-remote-file-repository.adoc
```diff
@@ -25,24 +25,40 @@ endif::[]
...
 ----
 endif::[]
 . Install the Pulp Manifest package:
+** For {EL} 8:
 +
-[options="nowrap" subs="+quotes"]
+[options="nowrap" subs="+quotes,attributes"]
 ----
-# yum install python3-pulp_manifest
+# {package-install-project} python3-pulp_manifest
```
Lennonka, line 51:
```suggestion
# {package-install} python3-pulp_manifest
```
The same attribute as [above](#diff-c4152de0d96ec7f6a12947f504675387578d2cfc56fa48871b82a6fbe6c1bb4fL40-R39) perhaps?

ekohl (commented):
Reading through the preview: the Katello document ends after section 5.13 so I can't really preview. I'm going to bet that there's some `endif::[]` missing after some `ifdef::satellite[]` since the Satellite document does render well.

https://docs.theforeman.org/nightly/Managing_Content/index-katello.html does render correctly so something is wrong. I'd start with rebasing your branch on master, given that there have been changes since you started your branch.

dubewarsagar:
> Reading through the preview: the Katello document ends after section 5.13 so I can't really preview. I'm going to bet that there's some `endif::[]` missing after some `ifdef::satellite[]` since the Satellite document does render well.
> 
> https://docs.theforeman.org/nightly/Managing_Content/index-katello.html does render correctly so something is wrong. I'd start with rebasing your branch on master, given that there have been changes since you started your branch.

Agree! I tried to rebase it. In a `git to do` file, it lists `50` commits to be squashed, whereas this PR has only `23` commits so far. No idea what the other commits come from and not able to decide what to do with this. Coming across such situation for the first time. So for now, I have `aborted git rebase`.

can someone guide what should be the next steps?

Thanks!

dubewarsagar:
@ekohl, 
I tried rebasing this branch and was getting into some unusual git issues. Hence i created another #1722 with the same changes. The new PR contains instructions for EL 8 only, hence should be picked in foreman 3.4 only as per the comment this [comment](https://github.com/theforeman/foreman-documentation/pull/1671#discussion_r995598356).

Regards,

maximiliankolb:
@dubewarsagar Please close this PR if you do not plan to continue working on this branch/PR.

dubewarsagar:
> @dubewarsagar Please close this PR if you do not plan to continue working on this branch/PR.

Planning to close once #1722 and #1723 which are alternate to this PR get merged to the respective branches. 

Thanks!

maximiliankolb:
Closed in favor of https://github.com/theforeman/foreman-documentation/pull/1722

PR #1651, opened 2022-09-22
Mention required version of ruby, pgsql for EL8

mjivraja:
The project needs ruby:2.7 and postgresql:12 modules to work\get installed. If end-users proceed with enabling the present el8 module in installing project and proxy, it will pop-up a huge warning related to conflict with ruby:2.5 and postgresql:10 modules. In the end, the command does enable the ruby:2.7 and postgresql:12 but the list of warnings itself is not a healthy status of the output and this occurence should be prevented. However, in the document, these warnings are mentioned as expected and can be ignored but it should not be the standard practice of user experience. Hence, this warning can be easily avoided by enabling the required dependant modules and the actual module.

https://bugzilla.redhat.com/show_bug.cgi?id=2126884

* [ ] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/proc_configuring-repositories-proxy.adoc
+++ b/guides/common/modules/proc_configuring-repositories-proxy.adoc
@@ -10,6 +10,24 @@ Select the option that corresponds with operating system and version you want to
 
 == [[repositories-rhel-8]]{RHEL} 8
 
+For configuring repositories in {EL} 8, you must enable `ruby:2.7` and `postgresql:12`.
+
+. Verify the module versions:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# dnf module list --enabled
+----
++
+If the version of modules is incompatible:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# dnf module reset ruby
+# dnf module reset postgresql
+----
++
+
 . Disable all repositories:
 +
 [options="nowrap"]
@@ -32,14 +50,11 @@ Select the option that corresponds with operating system and version you want to
 +
 [options="nowrap"]
 ----
-# dnf module enable satellite-capsule:el8
+# dnf module enable satellite-capsule:el8 postgresql:12 ruby:2.7
 ----
 +
 [NOTE]
 ====
-Enablement of the module `satellite-capsule:el8` warns about a conflict with `postgresql:10` and `ruby:2.5` as these modules are set to the default module versions on {RHEL} 8.
-The module `satellite-capsule:el8` has a dependency for the modules `postgresql:12` and `ruby:2.7` that will be enabled with the `satellite-capsule:el8` module.
-These warnings do not cause installation process failure, hence can be ignored safely.
 For more information about modules and lifecycle streams on {RHEL} 8, see https://access.redhat.com/support/policy/updates/rhel-app-streams-life-cycle[{RHEL} Application Streams Life Cycle].
 ====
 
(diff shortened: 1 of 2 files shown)


guides/common/modules/proc_configuring-repositories-proxy.adoc
```diff
@@ -10,6 +10,24 @@ Select the option that corresponds with operating system and version you want to
...
+For configuring repositories in {EL} 8, you must enable `ruby:2.7` and `postgresql:12`.
+
+. Verify the module versions:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# dnf module list --enabled
+----
++
+If the version of modules is incompatible:
```
bangelic, line 22: If the version of modules *are* incompatible.

guides/common/modules/proc_configuring-repositories.adoc
```diff
@@ -92,6 +92,24 @@ ifdef::foreman-el,katello,satellite[]
...
+For configuring repositories in {EL} 8, you must enable `ruby:2.7` and `postgresql:12`.
+
+. Verify the module versions:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# dnf module list --enabled
+----
++
+If the version of modules is incompatible:
```
bangelic, line 104: If the version of modules *are* incompatible.

guides/common/modules/proc_configuring-repositories-proxy.adoc
bangelic: > It is referring to `version` and not `modules`. slightly_smiling_face

My mistake! Now looking back at it, I see it. :)

mjivraja: > My mistake! Now looking back at it, I see it. :)

No problem @bangelic Do have a look at my other PRs :heart: :+1:

guides/common/modules/proc_configuring-repositories-proxy.adoc
```diff
@@ -10,6 +10,24 @@ Select the option that corresponds with operating system and version you want to
...
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# dnf module reset ruby
+# dnf module reset postgresql
```
ekohl, line 27: For what it's worth we already have some guidance around resetting modules in `guides/common/modules/con_troubleshooting.adoc`, but AFAIK we don't use it in Satellite. The biggest challenge with postgresql is that it could mean there's an existing PostgreSQL 10 database and migrating that to 12 is more than just resetting the module.

ekohl (changes requested):
I think this should first be applied to master and then cherry picked.

As for the document flow, I'd suggest to NOT list modules and reset, but rather change the instructions that if `dnf module enable satellite-capsule:el8 postgresql:12 ruby:2.7` fails, then direct them to the existing troubleshooting section. Also for Satellite.

guides/common/modules/proc_configuring-repositories-proxy.adoc
mjivraja: I have raised the new PR #1694 with all the changes you recommended above. :+1: 
There is a constant learning :smiley:

mjivraja: Let me know if I need to close this PR.

ekohl:
@mjivraja is this still relevant or can this be closed now?

mjivraja:
Closed this PR as it is no longer required due to #1694

Issue #1261, opened 2022-05-10
Image files cleanup


Lennonka:
The cleanup should encompass:

1. Moving images from guide folders to `common/images`
2. Renaming files to follow a convention so that they are distinguishable per product and make the file names descriptive
3. Updating include paths accordingly

We should agree on a naming convention for image files. [guides/README](https://github.com/theforeman/foreman-documentation/blob/master/guides/README.md) says:
> Image file names use dashes (`-`) and suffix a build target, e.g. `foreman`.

However, we started using prefix instead of suffix. I would suggest suffix so that the same image is "grouped" by topic when sorted alphabetically, which is more benefical for upstream workflow. And do we want to use dashes or underscores? I don't see reasons to prefer one or the other, but we should stick with one.

I guess this will have to be done manually.

Lennonka:
@maximiliankolb I'd love to have your input! :)

ekohl:
Just some infrastructure considerations. Don't let that hold you back too much, but take them into account.

> Moving images from guide folders to `common/images`

I do have a concern that the images end up in every single guide. That makes the gh-pages branch even larger than it is today. I'd love to see some deduplication here, but right now I'm not sure how to achieve that. Perhaps with symlinks or some smarter solution.

> Renaming files to follow a convention so that they are distinguishable per product and make the file names descriptive

I love conventions, but we shouldn't forget to clean the old images from the `gh-pages` branch (or figure out a solution to automatically purge old files on deployment).

maximiliankolb:
It's not really something that I have currently on my plate, but collecting *all* images in one folder seems like a good idea to me; I'd help/support doing that. This could also contain some manual deduplication.

Also, SVGs are generally smaller than PNG files.

Maybe not smylink the "common images" dir to every guide, but instead traverse back `../` to the "common images" dir?

Lennonka:
@ekohl Could we pull the renamed/deleted files from git to update/purge them on deploy? Not sure how to implement it, but git (commits) has the relevant data to do that kind of update, doesn't it?

Lennonka:
@maximiliankolb How would that help us to use backward paths instead of the symlinks? Additionally, I'm not sure how backward paths behave in our downstream build. I'm worried that the symlinks are there for a reason.

ekohl:
I've looked at that, but this would require some tweaking:
[...]

Issue #873, opened 2021-11-23
Migrate to Antora


lzap:
Hey,

this is just a tracking ticket and starting point for discussion about Antora: https://antora.org

While I am currently working on a simple navigation bar with pure HTML/JS/CSS embedded into asciidoc documents so readers can seamlessly switch guides or go back to the welcome page, I think long-term we need to think big. Antora looks like a great fit, however, this needs to be done in coordination with Red Hat Documentation team as things will need to be drastically changed downstream, and of course ATIX who also uses the documentation for their products.

Just to calm expectations: I am thinking something like end of 2022 as the earliest chance of starting this process. I would be glad for all the feedback in the ticket. Cheers!

maximiliankolb:
We are using Antora in our [downstream documentation](https://docs.orcharhino.com). It's quite a bit different compared to asciidoctor, esp. when including other files or referencing anchors in other files.

So huge +1 from my side! I already showed @stejskalleos a bit of our Antora usage and a small "behind the scenes". We also already have a solution in place to check xrefs and to add search functionality.

Feel free to message me with any questions etc.

stejskalleos:
+1 for this. @maximiliankolb gave me intro about Antora and it looked awesome. With Antora we could have search (yay!) and use their plugins and extensions to make docs really great.

I'll be glad to help you with the work, feel free to ping me.

maximiliankolb:
Some notes on recovering from a broken state in a protected branch which results in failing PRs against that branch due to GHA creating a diff:

````
## Recovering from a broken build due to the comparison with the previous commit

If you open a PR on Github, your feature branch gets built and compared to the target branch into which you want to merge.
Therefore, the Github Action builds the target branch.
This implies that even if your change works on its own, it depends on a working state in the target branch.

If the target branch does not build, your PR fails due to the comparision.
To fix this, you have to manually set the reference from "merge base of the target branch and your feature branch" to a commit ID of a working state, for example the parent commit of the faulty yet merged commit on the target branch.

Example: [REVERT ME: Use last working state as base](https://github.com/theforeman/foreman-documentation/commit/0ea9e707b6850487e60753a40fc2b2791528d175)
````

See also https://github.com/theforeman/foreman-documentation/pull/1865#discussion_r1065669720

Lennonka:
I was playing with Antora yesterday and I'm not sure if it's a good idea to implement it. But I've only started and I'm certainly not finished yet. Still getting to know the whole system. Stay tuned...

[...]

PR #1655, opened 2022-09-26
Clean up images - rename (issue 1261)

Lennonka:
Issue #1261 

* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3


guides/common/modules/proc_implementing-pxe-less-discovery.adoc
```diff
@@ -21,15 +21,15 @@ The console might freeze during the process.
 On some hardware, you might experience graphical hardware problems.
 
 ifdef::satellite[]
-image::common/PXEless-mode.png[]
+image::common/pxeless-mode.png[]
```
maximiliankolb, line 24:
```suggestion
image::common/pxeless-mode-satellite.png[]
```

Please also rename the file itself.

maximiliankolb (commented):
I would like to not loose the `.drawio` files. Maybe we can instead move them to another directly that will not be part of built docs?

maximiliankolb:
Comparing the built docs I was expecting a smaller archive:

Screenshot 2022-09-27 at 11 49 04
Screenshot 2022-09-27 at 11 48 57

guides/common/modules/proc_implementing-pxe-less-discovery.adoc
Lennonka: Done, thanks for the catch!

maximiliankolb:
Me too. I am hesitant to merging this (with good intention) when the size almost doubles. Maybe we need to introduce a way to split images by build target?

I think `guides/image-sources` is a good place.

ekohl:
This is precisely what I mentioned in https://github.com/theforeman/foreman-documentation/issues/1261#issuecomment-1253798839. Because the images now get places in every resulting guide (regardless if they're used or not) the size is much larger than if they only reside in the guide that actually uses them.

I've been wondering if we should bite the bullet and address https://github.com/theforeman/foreman-documentation/issues/873. Hopefully Antora is smarter about image placement, though I don't have any idea if this is really true.

maximiliankolb:
I've just checked on our Antora instance: Images that are included multiple times are only on there once which is nice.

guides/image-sources/README.md
```diff
@@ -0,0 +1,7 @@
+# Image sources
+
+This folder contains editable sources of images.
+
+Draw.io files belong here.
+
+Exported images in PNG and SVG formats go to `common/images`.
```
maximiliankolb, line 7:
```suggestion
Exported images in PNG and SVG formats go to `guides/common/images`.
```

Lennonka:
I've modified it to only rename the image files, but keep their location.

maximiliankolb (approved):
Thanks, LGTM!

I think it'd be helpful if you'd add a sentence or two on **where** and **how** to place images in general to our conventions/README/CONTRIBUTING (whatever you see fit best). Feel free to do this is another small follow-up PR.

Lennonka:
Cherry-picked:
- 0216b5b6f..c633f8129  3.4 -> 3.4
- 61ace1c74..6ad4b29a8  3.3 -> 3.3
- 3b389cf10..90406e5b2  3.2 -> 3.2
- c5be41cd9..9912194dd  3.1 -> 3.1

PR #1656, opened 2022-09-27
Use EL8-based instructions in upstream

ekohl:
The previous instructions were for EL7, even though they described it as EL8. The Orcharhino instructions are correct for EL8. This change moves everything to EL8 for non-Satellite and keeps EL7 for Satellite.

The only modification is that the repository file is now modified for production use. This mostly means using mirror lists and pointing to the exact release for which the documentation is generated.

* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/proc_building-a-discovery-image.adoc
+++ b/guides/common/modules/proc_building-a-discovery-image.adoc
@@ -7,13 +7,10 @@ Discovered hosts keep running the Discovery image until they are rebooted into A
 ifdef::satellite[]
 The operating system image is based on {RHEL} 7.
 endif::[]
-ifdef::orcharhino[]
-Run the following procedure on Rocky Linux 8.
+ifndef::satellite[]
+Run the following procedure on {EL} 8.
 
-The Discovery image is based on Cent OS Stream 8.
-endif::[]
-ifndef::satellite,orcharhino[]
-The operating system image is based on CentOS Stream 8.
+The Discovery image is based on CentOS Stream 8.
 endif::[]
 
 ifdef::satellite[]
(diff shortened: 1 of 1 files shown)


guides/common/modules/proc_building-a-discovery-image.adoc
```diff
@@ -44,6 +41,24 @@ $ cd foreman-discovery-image
 ----
 
 .Procedure
+. Open the `foreman-discovery-image.ks` file for editing:
++
+[options="nowrap" subs="+quotes"]
+----
+# vim 00-repos-centos8.ks
+----
```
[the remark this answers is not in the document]
ekohl: > Is this correct? I am asking because I see a different file name in the instruction than in the terminal example below it.

It's based on the old instructions (line 129 in the new version). The heading is indeed a copy-paste error.



> I also wonder if we should instruct user to explicitly use vim (which can be confusing if the user never encountered it before) instead of suggesting to simply replace the lines in the file and leaving the method to the user.

Good question. Perhaps that's better.

guides/common/modules/proc_building-a-discovery-image.adoc
```diff
@@ -44,6 +41,18 @@ $ cd foreman-discovery-image
 ----
 
 .Procedure
+. Replace content of `00-repos-centos8.ks` with the following content:
```
adamlazik1, line 44:
```suggestion
. Replace content of the `00-repos-centos8.ks` file with the following text:
```

maximiliankolb (commented):
Apart from what Adam already mentioned, this looks good to me.

guides/common/modules/proc_building-a-discovery-image.adoc
ekohl: I went with a slightly different text. How about now?

adamlazik1: Looks good to me!

adamlazik1 (commented):
LGTM

ekohl:
@maximiliankolb there are some differences for Orcharhino. I'd appreciate a formal approval/review rather than merging it myself.

maximiliankolb (approved):
LGMT!

ekohl:
I've submitted https://github.com/theforeman/foreman-documentation/pull/1681 and https://github.com/theforeman/foreman-documentation/pull/1680 since there was a minor conflict in the cherry picks.

PR #1666, opened 2022-09-29
Warn users about setting other Unicode

mjivraja:
The project only supports the UTF-8 encoding system. However, there are some users who set another encoding method in their system and as a result, they face issues in the installation process. Hence, we are asking users to set UTF-8 as an encoding method mandatorily before installing Project.

https://bugzilla.redhat.com/show_bug.cgi?id=2127567

* [X] Foreman 3.4/Katello 4.6
* [X] Foreman 3.3/Katello 4.5
* [X] Foreman 3.2/Katello 4.4
* [X] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/ref_system-requirements.adoc
+++ b/guides/common/modules/ref_system-requirements.adoc
@@ -40,6 +40,10 @@ endif::[]
 * Administrative user (root) access
 * Full forward and reverse DNS resolution using a fully-qualified domain name
 
+{Project} only supports `UTF-8` encoding.
+If your territory is USA and your language is English, set `en_US.utf-8` as the system-wide locale settings.
+For more information about configuring system locale in {EL}, see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/assembly_changing-basic-environment-settings_configuring-basic-system-settings#proc_configuring-the-system-locale_assembly_changing-basic-environment-settings[Configuring System Locale guide].
+
 {ProjectServer} and {SmartProxyServer} do not support shortnames in the hostnames.
 When using custom certificates, the Common Name (CN) of the custom certificate must be a fully qualified domain name (FQDN) instead of a shortname.
 This does not apply to the clients of a {Project}.


ekohl (commented):
Should this mention how users can check this (run `locale` for the current user and `localectl` for the system wide config), get installed locales (`localectl list-locales`) and modify it as needed (`localectl set-locale LOCALE` for the system)?

I'd also argue that this actually needs a system check to enforce this. I'm just trying to figure out where it should be. The BZ doesn't specify if it should be the user locale or system locale, so I left a comment there.

waldirio:
> Should this mention how users can check this (run `locale` for the current user and `localectl` for the system wide config), get installed locales (`localectl list-locales`) and modify it as needed (`localectl set-locale LOCALE` for the system)?
> 
> I'd also argue that this actually needs a system check to enforce this. I'm just trying to figure out where it should be. The BZ doesn't specify if it should be the user locale or system locale, so I left a comment there.

Hey my friend, yea, I believe it to be interesting to be available, quickly, how to check and set properly should be a must.

Take care!
Waldirio

mjivraja:
> > Should this mention how users can check this (run `locale` for the current user and `localectl` for the system wide config), get installed locales (`localectl list-locales`) and modify it as needed (`localectl set-locale LOCALE` for the system)?
> > I'd also argue that this actually needs a system check to enforce this. I'm just trying to figure out where it should be. The BZ doesn't specify if it should be the user locale or system locale, so I left a comment there.
> 
> Hey my friend, yea, I believe it to be interesting to be available, quickly, how to check and set properly should be a must.
> 
> Take care! Waldirio

@waldirio You shared the link which guides users to set `locale`. However, it's of very old version so I didn't consider it at this level.

ekohl:
> IMHO, from a writer's perspective, I believe that it may not fall into our responsibility in guiding users to manage their `locale` as it is OS side configuration. However, we can share the link from RHEL if they have something to serve for your view. I am open to modification though +1

I'm not too familiar with the RHEL docs, but if there is a document about how to configure it that is a big improvement for users.

@waldirio so do you know if it should be user locale, system locale or both?

waldirio:
Hello @ekohl and @mjivraja 

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/assembly_changing-basic-environment-settings_configuring-basic-system-settings#proc_configuring-the-system-locale_assembly_changing-basic-environment-settings

This is the link that I believe we are looking for. I just checked, the steps are valid for both, rhel7 and rhel8.

Please, let me know if that works for you.

Waldirio

mjivraja:
> Hello @ekohl and @mjivraja
> 
> https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/assembly_changing-basic-environment-settings_configuring-basic-system-settings#proc_configuring-the-system-locale_assembly_changing-basic-environment-settings
> 
> This is the link that I believe we are looking for. I just checked, the steps are valid for both, rhel7 and rhel8.
> 
> Please, let me know if that works for you.
> 
> Waldirio

Aah! Yeah. I should have looked for this but didn't strike my mind. Let's wait for Ewoud.

ekohl:
That guide looks like exactly what I was looking for. Linking to that makes it actionable for users and addresses my concern.

waldirio:
> That guide looks like exactly what I was looking for. Linking to that makes it actionable for users and addresses my concern.

+1

Thanks!
Waldirio

guides/common/modules/ref_system-requirements.adoc
```diff
@@ -40,6 +40,11 @@ endif::[]
 * Administrative user (root) access
 * Full forward and reverse DNS resolution using a fully-qualified domain name
 
+{Project} only supports `UTF-8` encoding system.
```
ekohl, line 43: I'd say it's either
```suggestion
{Project} only supports the `UTF-8` encoding system.
```

But more commonly
```suggestion
{Project} only supports `UTF-8` encoding.
```

guides/common/modules/ref_system-requirements.adoc
```diff
@@ -40,6 +40,11 @@ endif::[]
...
+No other encoding method is supported.
+Set `en_US.utf-8` in System-wide locale settings if your language is English and territory is USA.
```
ekohl, line 45:
```suggestion
For example, set `en_US.utf-8` as the system-wide locale if your language is English and territory is USA.
```

ekohl (commented):
Content wise I think it's correct, but language wise I think it could use a review from a writer. I made some suggestions, but @maximiliankolb or @Lennonka may be better able to review it from here.

guides/common/modules/ref_system-requirements.adoc
Lennonka: @ekohl Does the user have to specify the territory or can they use just `utf-8`?

guides/common/modules/ref_system-requirements.adoc
```diff
@@ -40,6 +40,11 @@ endif::[]
...
+For more information about configuring system locale in {EL}, see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/assembly_changing-basic-environment-settings_configuring-basic-system-settings#proc_configuring-the-system-locale_assembly_changing-basic-environment-settings[Configuring System Locale guide]
```
maximiliankolb, line 46:
```suggestion
For more information about configuring system locale in {EL}, see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/assembly_changing-basic-environment-settings_configuring-basic-system-settings#proc_configuring-the-system-locale_assembly_changing-basic-environment-settings[Configuring System Locale guide].
```

guides/common/modules/ref_system-requirements.adoc
maximiliankolb: We'd generally start with "If X, then Y", so "For example, if ..., set ...".

ekohl: @Lennonka a locale is always a combination. https://en.wikipedia.org/wiki/Locale_(computer_software)#POSIX_platforms says it's defined as `[language[_territory][.codeset][@modifier]]`.

A bit of backstory for those who don't know the potential impact of this.

Many systems use `en_US.UTF-8` and a lot of software doesn't even consider other locales may behave differently. For example, it may affect sorting:
```console
$ echo -e "a\nb\nc\nch\nd\nh\ni" | LC_ALL=en_US.UTF-8 sort
a
b
c
ch
d
h
i
$ echo -e "a\nb\nc\nch\nd\nh\ni" | LC_ALL=cs_CZ.UTF-8 sort
a
b
c
d
h
ch
i
```

Similarly, it affects how dates are printed (try `LC_TIME=... date` or `LC_ALL=...`). https://github.com/theforeman/foreman-installer/commit/926628891b02f7408cd59ce7c0a832c94709f3b4 is an example of a bug that was solved. It tried to parse a specifically formatted date, which worked in one locale but failed in another. There the easiest solution was to not store & parse the date, but rather format it just once.

The backstory of this PR is that some dependency of Pulpcore (I can't remember which one right now) would segfault if the ISO-8859-1 encoding was used instead of UTF-8.

Locales, like timezones, are a difficult subject filled with bugs because it's always more complex than you think.

guides/common/modules/ref_system-requirements.adoc
mjivraja: Addressed :+1:

guides/common/modules/ref_system-requirements.adoc
mjivraja: > 

Addressed :+1:

guides/common/modules/ref_system-requirements.adoc
```diff
@@ -40,6 +40,11 @@ endif::[]
 * Administrative user (root) access
 * Full forward and reverse DNS resolution using a fully-qualified domain name
 
+{Project} only supports `UTF-8` encoding.
+No other encoding method is supported.
```
maximiliankolb, line 44:
```suggestion
```
I'm not sure this is needed if you explicitely say "only utf8 is supported" above.

mjivraja: Addressed :+1:

maximiliankolb (approved):
LGTM

maximiliankolb:
   fa137761f..2a9fb6404  3.4 -> 3.4
   a769167c0..29d45eaa0  3.3 -> 3.3
   21a0d76b4..36e1b10f3  3.2 -> 3.2
   cb35f9ba6..eca55ac2e  3.1 -> 3.1

PR #1670, opened 2022-09-30
Rework Deleting Content View Versions module

adamlazik1:
Deleting multiple cv versions is not applicable in 3.1. You can only delete one cv version at a time. Newer versions have the bulk version selection for deletion.


adamlazik1:
I don't know whether this also applies to Foreman 3.2. perhaps someone could confirm?

adamlazik1:
For now, it is sufficient if someone reviews this so it can get merged into 3.1.

ekohl:
> I don't know whether this also applies to Foreman 3.2. Perhaps someone could confirm?

3.2 is EOL upstream and there are no downstream consumers of it, so I'm fine with ignoring it.

maximiliankolb (approved):
LGTM!

PR #1677, opened 2022-10-03
Update diagrams for Satellite

Lennonka:
Update of Sat diagrams with a new graphic style.

* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3


maximiliankolb (approved):
I can't comment on the images, but your change looks reasonable to me. Feel free to merge IMHO.

ekohl (commented):
I'm looking at this and I see a lot of it was already there, but the network ports at the top are bad IMHO. I can't comment directly inline on the PXE-mode diagram, but DHCP is over UDP port 67/68 (step 1; client asks the server on port 67 and the server responds on port 68). TFTP is UDP port 69. Then step 3 is the fact upload. That is sent to the Smart Proxy, which in Katello/Satellite runs on TCP port 9090 (8443 in vanilla Foreman). That in turn (step 4) forwards it to Foreman running on TCP port 443. Step 5 also connects to the Smart Proxy, running on port 9090. Step 6 then does connect to the discovered node on TCP port 8443.

As for load balancing: I'm not touching that. I still think it should not be supported at all because it's filled with missing parts. However, it does look like you have port 433 instead of 443.

Something I also wonder about: can we use SVG instead of PNG?

Lennonka:
I'll fix the port number in load balancer.
Regarding PXE mode, I think that adding the port numbers for each step would be an unnecessary level of detail.
And we can't use SVG because Pantheon cannot render them in PDF. (I tested it.)

Lennonka:
Cherry-picked:
- ed999687f..09f7d82a5  3.4 -> 3.4
- e7f81d824..9b5681798  3.3 -> 3.3

PR #1678, opened 2022-10-04
Add command in Proxy certificate renewal

mjivraja:
The current commands are not enough to renew the Proxy certificate which can be improved further. So, we added the --certs-update-server-ca option while deploying the custom SSL certificate to the Proxy server. In addition to this, we also added --certs-update-all to Project and Proxy both for the same section in the documentation. In the case where suggested commands do not work, --certs-update-all will help to update the internal Proxy certificates. However, it may not be needed for just the renewal process but in some cases, it is required to update all the certificates.

https://bugzilla.redhat.com/show_bug.cgi?id=2093128

* [ ] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
+++ b/guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
@@ -35,45 +35,41 @@ If you set a wildcard value `*` for the certificate's Common Name `CN =` in the 
 If the command is successful, it returns two `{certs-generate}` commands, one of which you must use to generate the certificate archive file for your {SmartProxyServer}.
 ifdef::satellite[]
 +
-.Example output of `katello-certs-check`
+.To use the certificates inside a _new-{smartproxy-example-com}_, run this command:
 [options="nowrap", subs="+quotes,attributes"]
 ----
-Validation succeeded.
-
-To use them inside a NEW $CAPSULE, run this command:
-
 capsule-certs-generate --foreman-proxy-fqdn "$CAPSULE" \
     --certs-tar  "~/$CAPSULE-certs.tar" \
     --server-cert "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem_" \
     --server-key "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert_key.pem_" \
     --server-ca-cert "_/root/{smart-proxy-context}_cert/ca_cert_bundle.pem_" \
-
-To use them inside an EXISTING $CAPSULE, run this command INSTEAD:
-
+----
++
+.To use the certificates inside an _existing-{smartproxy-example-com}_, run this command instead:
+----
   capsule-certs-generate --foreman-proxy-fqdn "$CAPSULE" \
     --certs-tar "~/$CAPSULE-certs.tar" \
     --server-cert "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem_" \
     --server-key "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert_key.pem_" \
     --server-ca-cert "_/root/{smart-proxy-context}_cert/ca_cert_bundle.pem_" \
-    --certs-update-server
+    --certs-update-all
 ----
 endif::[]
 
 ifndef::satellite[]
 +
-.Example output of `katello-certs-check`
+.To use the certificates inside a new $FOREMAN_PROXY, run this command:
 [options="nowrap", subs="+quotes,attributes"]
 ----
-Validation succeeded.
-
-To use them inside a NEW $FOREMAN_PROXY, run this command:
   foreman-proxy-certs-generate --foreman-proxy-fqdn "$FOREMAN_PROXY" \
-    --certs-tar  "~$FOREMAN_PROXY-certs.tar" \
+    --certs-tar  "~/$FOREMAN_PROXY-certs.tar" \
     --server-cert "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem_" \
     --server-key "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert_key.pem_" \
     --server-ca-cert "_/root/{smart-proxy-context}_cert/ca_cert_bundle.pem_" \
-
-To use them inside an EXISTING $FOREMAN_PROXY, run this command INSTEAD:
+----
++
+.To use the certificates inside an existing $FOREMAN_PROXY, run this command instead:
+----
   foreman-proxy-certs-generate --foreman-proxy-fqdn "\$FOREMAN_PROXY" \
     --certs-tar  "~/$FOREMAN_PROXY-certs.tar" \
     --server-cert "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem_" \
@@ -83,7 +79,10 @@ To use them inside an EXISTING $FOREMAN_PROXY, run this command INSTEAD:
 ----
 endif::[]
 
-. On {ProjectServer}, from the output of the `katello-certs-check` command, depending on your requirements, enter the `{certs-generate}` command that generates a certificate for a new or existing {SmartProxy}.
+. Depending on your requirements, enter the `{certs-generate}` command on the {ProjectServer} that generates a certificate for a new or existing {SmartProxy}.
+The output of the `katello-certs-check` command may not be accurate in some cases.
+// https://bugzilla.redhat.com/show_bug.cgi?id=2093128
+Therefore, you must follow the steps mentioned above instead of the command outputs.
 +
 ifdef::satellite[]
 In this command, change `$CAPSULE` to the FQDN of your {SmartProxyServer}.
--- a/guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
+++ b/guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
@@ -25,43 +25,38 @@ Note that for the `katello-certs-check` command to work correctly, Common Name (
 If the command is successful, it returns two `{foreman-installer}` commands, one of which you must use to deploy a certificate to {ProjectServer}.
 ifdef::satellite[]
 +
-.Example output of `katello-certs-check`
+.To install the {Team} {ProjectServer} with the custom certificates, run:
 [options="nowrap", subs="+quotes,attributes"]
 ----
-Validation succeeded.
-
-To install the Red Hat Satellite Server with the custom certificates, run:
-
   {foreman-installer} --scenario satellite \
     --certs-server-cert "_/root/satellite_cert/satellite_cert.pem_" \
     --certs-server-key "_/root/satellite_cert/satellite_cert_key.pem_" \
     --certs-server-ca-cert "_/root/satellite_cert/ca_cert_bundle.pem_"
-
-To update the certificates on a currently running Red Hat Satellite installation, run:
-
+----
++
+.To update the certificates on a currently running {Project} installation, run:
+----
   {foreman-installer} --scenario satellite \
     --certs-server-cert "_/root/satellite_cert/satellite_cert.pem_" \
     --certs-server-key "_/root/satellite_cert/satellite_cert_key.pem_" \
     --certs-server-ca-cert "_/root/satellite_cert/ca_cert_bundle.pem_" \
-    --certs-update-server --certs-update-server-ca
+    --certs-update-all
 ----
 endif::[]
 ifndef::satellite[]
 +
-.Example output of `katello-certs-check`
+.To install the Katello main server with the custom certificates, run:
 [options="nowrap", subs="+quotes,attributes"]
 ----
-Validation succeeded.
-
-To install the Katello main server with the custom certificates, run:
-
   foreman-installer --scenario katello \\
     --certs-server-cert "_/root/satellite_cert/satellite_cert.pem_" \
     --certs-server-key "_/root/satellite_cert/satellite_cert_key.pem_" \
     --certs-server-ca-cert "_/root/satellite_cert/ca_cert_bundle.pem_"
+----
 
-To update the certificates on a currently running Katello installation, run:
-
+.To update the certificates on a currently running Katello installation, run:
+[options="nowrap"]
+----
   foreman-installer --scenario katello \\
     --certs-server-cert "_/root/satellite_cert/satellite_cert.pem_" \
     --certs-server-key "_/root/satellite_cert/satellite_cert_key.pem_" \
@@ -70,7 +65,9 @@ To update the certificates on a currently running Katello installation, run:
 ----
 endif::[]
 
-. From the output of the `katello-certs-check` command, depending on your requirements, enter the `{foreman-installer}` command that installs a new {Project} with custom SSL certificates or updates certificates on a currently running {Project}.
+. Depending on your requirements, enter the `{foreman-installer}` command that installs a new {ProjectServer} with custom SSL certificates or updates certificates on a currently running {ProjectServer}.
+The output of the `katello-certs-check` command may not be accurate in some cases.
+Therefore, you must follow the steps mentioned above instead of the command outputs.
 +
 If you are unsure which command to run, you can verify that {Project} is installed by checking if the file `/etc/foreman-installer/scenarios.d/.installed` exists.
 If the file exists, run the second `{foreman-installer}` command that updates certificates.


guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
```diff
@@ -44,6 +44,7 @@ To update the certificates on a currently running Red Hat Satellite installation
     --certs-server-key "_/root/satellite_cert/satellite_cert_key.pem_" \
     --certs-server-ca-cert "_/root/satellite_cert/ca_cert_bundle.pem_" \
     --certs-update-server --certs-update-server-ca
+    --certs-update-all
```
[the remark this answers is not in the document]
mjivraja: @adamlazik1 I need to check this with the reviewers of this PR. 
Because `\` may only be applied to directories/files mentioned in double quotes.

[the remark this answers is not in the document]
ekohl: @adamlazik1 is right. This is because it's a single command. Otherwise `--certs-update-all` is executed as a standalone command that runs after.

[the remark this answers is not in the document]
mjivraja: Addressed :+1:

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
```diff
@@ -55,7 +55,8 @@ To use them inside an EXISTING $CAPSULE, run this command INSTEAD:
     --server-cert "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem_" \
     --server-key "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert_key.pem_" \
     --server-ca-cert "_/root/{smart-proxy-context}_cert/ca_cert_bundle.pem_" \
-    --certs-update-server
+    --certs-update-server --certs-update-server-ca
```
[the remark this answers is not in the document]
mjivraja: Addressed :+1:

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
```diff
@@ -55,7 +55,8 @@ To use them inside an EXISTING $CAPSULE, run this command INSTEAD:
     --server-cert "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem_" \
     --server-key "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert_key.pem_" \
     --server-ca-cert "_/root/{smart-proxy-context}_cert/ca_cert_bundle.pem_" \
-    --certs-update-server
+    --certs-update-server --certs-update-server-ca \
+    --certs-update-all
```
adamlazik1, line 59:
```suggestion
    --certs-update-server \
    --certs-update-server-ca \
    --certs-update-all
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
```diff
@@ -43,7 +43,8 @@ To update the certificates on a currently running Red Hat Satellite installation
     --certs-server-cert "_/root/satellite_cert/satellite_cert.pem_" \
     --certs-server-key "_/root/satellite_cert/satellite_cert_key.pem_" \
     --certs-server-ca-cert "_/root/satellite_cert/ca_cert_bundle.pem_" \
-    --certs-update-server --certs-update-server-ca
+    --certs-update-server --certs-update-server-ca \
+    --certs-update-all
```
adamlazik1, line 47:
```suggestion
    --certs-update-server \
    --certs-update-server-ca \
    --certs-update-all
```

adamlazik1 (commented):
One additional thing that I would do is to have each option on a separate line to be consistent with the rest of the command. @ekohl what do you think?

ekohl (commented):
I'm going to make this complicated, so bear with me.

The header above it is literally:
> Example output of `katello-certs-check`

Now the question is: what does it output? I'll save to the digging and tell you it's in foreman-installer. The exact code is [here](https://github.com/theforeman/foreman-installer/blob/a28dabbb46746308b6f8fafc70bc6b971947d49b/bin/katello-certs-check#L274-L292). That doesn't mention those options. In theory this could be patched in downstream, but I checked and it isn't (I can't link to that).

However, the question is now what should be done. On the one hand, the options are needed. On the other hand, then it will no longer match the output. Worse, if the user actually runs the command that's shown on the terminal they end up with incorrect instructions.

Ideal case is that it's actually patched in downstream and this would match the output. Then the user (provided they're up to date) can't end up seeing invalid instructions.

@ehelms any thoughts on how best to proceed from here?

ekohl:
Would it then make sense to immediately remove the literal command output and list them as explicit instructions like we do with all our other instructions too? Once the tool is fixed, modify the instructions to "follow what's on screen".

mjivraja:
Hi @ekohl 

How can we proceed further with this...

ekohl:
I'd move forward with removing the invalid "Example output" and then rewriting it like we write all our procedures: give the user the explicit commands to run. Also mention that the output of check command contains a problem and that they should follow the documented steps, not the command output.

mjivraja:
> I'd move forward with removing the invalid "Example output" and then rewriting it like we write all our procedures: give the user the explicit commands to run. Also mention that the output of check command contains a problem and that they should follow the documented steps, not the command output.

Addressed :+1:

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
```diff
@@ -35,45 +35,42 @@ If you set a wildcard value `*` for the certificate's Common Name `CN =` in the
...
     --certs-tar  "~/$CAPSULE-certs.tar" \
     --server-cert "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem_" \
     --server-key "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert_key.pem_" \
     --server-ca-cert "_/root/{smart-proxy-context}_cert/ca_cert_bundle.pem_" \
-
-To use them inside an EXISTING $CAPSULE, run this command INSTEAD:
-
+----
++
+.To use them inside an EXISTING $CAPSULE, run this command INSTEAD:
```
ekohl, line 48: The only difference is the additional `--certs-update-*` arguments. I'm not sure how much that is actually still needed, since we've fixed various idempotency bugs in that area.

I'd prefer to align on a single set of instructions. AFAIK the update instructions do not hurt on a new installation, so perhaps we should use that for all.

@ehelms you worked in this area. Thoughts?

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
```diff
@@ -35,45 +35,42 @@ If you set a wildcard value `*` for the certificate's Common Name `CN =` in the
...
 +
-.Example output of `katello-certs-check`
+.To use them inside a NEW $FOREMAN_PROXY, run this command:
 [options="nowrap", subs="+quotes,attributes"]
 ----
-Validation succeeded.
-
-To use them inside a NEW $FOREMAN_PROXY, run this command:
   foreman-proxy-certs-generate --foreman-proxy-fqdn "$FOREMAN_PROXY" \
     --certs-tar  "~$FOREMAN_PROXY-certs.tar" \
```
ekohl, line 66: There's a typo in the (existing) instructions:
```suggestion
    --certs-tar  "~/$FOREMAN_PROXY-certs.tar" \
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
```diff
@@ -35,45 +35,42 @@ If you set a wildcard value `*` for the certificate's Common Name `CN =` in the
 If the command is successful, it returns two `{certs-generate}` commands, one of which you must use to generate the certificate archive file for your {SmartProxyServer}.
 ifdef::satellite[]
 +
-.Example output of `katello-certs-check`
+.To use them inside a NEW $CAPSULE, run this command:
```
ekohl, line 38: In our documentation we have the convention of using hostnames. Perhaps we should use `_new-{smartproxy-example-com}_` instead of `$CAPSULE`.

Alternatively you can define an attribute for it to remove the duplication. Like:
```asciidoc
ifdef::satellite[]
:smartproxy-fqdn-variable: $CAPSULE
endif::[]
ifndef::satellite[]
:smartproxy-fqdn-variable: $FOREMAN_PROXY
endif::[]
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
```diff
@@ -83,7 +80,9 @@ To use them inside an EXISTING $FOREMAN_PROXY, run this command INSTEAD:
 ----
 endif::[]
 
-. On {ProjectServer}, from the output of the `katello-certs-check` command, depending on your requirements, enter the `{certs-generate}` command that generates a certificate for a new or existing {SmartProxy}.
+. Depending on your requirements, enter the `{certs-generate}` command on the {ProjectServer} that generates a certificate for a new or existing {SmartProxy}.
+The output of the `katello-certs-check` command may not be accurate in some cases.
+Therefore, you must follow the steps mentioned above instead of command outputs.
```
ekohl, line 85: It'd be great to have a bug reference. For example:
```suggestion
// https://bugzilla.redhat.com/show_bug.cgi?id=XYZ
Therefore, you must follow the steps mentioned above instead of command outputs.
```

That way you easily find out in the future if it's still a problem.

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
maximiliankolb: I recommend going with host names such as https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-base.adoc#L107. We expect users to adjust those values if they're in italic font.

mjivraja: According to your comment, I preferred Ewoud's suggestion above. :+1:

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
mjivraja: Thanks for this suggestion. Wasn't really aware if we could add it for internal reference.

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
ehelms:
```suggestion
    --certs-update-all
```

This has been around for a while, but I believe has been incorrect. If `--certs-update-all` is present the other parameters are redundant. The code link below shows this:

https://github.com/theforeman/foreman-installer/blob/develop/hooks/pre/20-certs_update.rb#L28-L36

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
ehelms: See https://github.com/theforeman/foreman-documentation/pull/1678/files#r1071405936

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
ehelms: I think to be on the safe side I'd need to just test if we have changed things enough to not need the parameters. They come from the installer itself:

https://github.com/theforeman/foreman-installer/blob/develop/hooks/pre/20-certs_update.rb

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
ehelms: This can be simplified to:

```
    --certs-update-all
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
ehelms: This can be simplified to:

```
    --certs-update-all
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
mjivraja: Addressed :+1:

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
mjivraja: Addressed :+1:

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
```diff
@@ -35,45 +35,41 @@ If you set a wildcard value `*` for the certificate's Common Name `CN =` in the
...
     --certs-tar  "~/$CAPSULE-certs.tar" \
     --server-cert "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem_" \
     --server-key "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert_key.pem_" \
     --server-ca-cert "_/root/{smart-proxy-context}_cert/ca_cert_bundle.pem_" \
-
-To use them inside an EXISTING $CAPSULE, run this command INSTEAD:
-
+----
++
+.To use them inside an _existing-{smartproxy-example-com}_, run this command INSTEAD:
```
ekohl, line 48: I know it was there before, but I wonder why `INSTEAD` is in caps, but I'm not sure it makes that much sense. Before it was there to highlight the difference to a new, but you removed those. I'm worried the hostname may be too subtle for users to distinguish.

ekohl (commented):
Command wise I think this is :+1: but perhaps a writer can review the textual changes.

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
```diff
@@ -35,45 +35,41 @@ If you set a wildcard value `*` for the certificate's Common Name `CN =` in the
 If the command is successful, it returns two `{certs-generate}` commands, one of which you must use to generate the certificate archive file for your {SmartProxyServer}.
 ifdef::satellite[]
 +
-.Example output of `katello-certs-check`
+.To use them inside a _new-{smartproxy-example-com}_, run this command:
```
adamlazik1, line 38: What is `them` in this context? If it is certificates then `To use the certificates inside...` would be better. Also applies to other similar subtitles in this document.

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
```diff
@@ -83,7 +79,10 @@ To use them inside an EXISTING $FOREMAN_PROXY, run this command INSTEAD:
 ----
 endif::[]
 
-. On {ProjectServer}, from the output of the `katello-certs-check` command, depending on your requirements, enter the `{certs-generate}` command that generates a certificate for a new or existing {SmartProxy}.
+. Depending on your requirements, enter the `{certs-generate}` command on the {ProjectServer} that generates a certificate for a new or existing {SmartProxy}.
+The output of the `katello-certs-check` command may not be accurate in some cases.
+// https://bugzilla.redhat.com/show_bug.cgi?id=2093128
+Therefore, you must follow the steps mentioned above instead of command outputs.
```
adamlazik1, line 85:
```suggestion
Therefore, you must follow the steps mentioned above instead of the command outputs.
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
```diff
@@ -25,43 +25,37 @@ Note that for the `katello-certs-check` command to work correctly, Common Name (
...
   foreman-installer --scenario katello \\
     --certs-server-cert "_/root/satellite_cert/satellite_cert.pem_" \
     --certs-server-key "_/root/satellite_cert/satellite_cert_key.pem_" \
     --certs-server-ca-cert "_/root/satellite_cert/ca_cert_bundle.pem_"
+----
 
-To update the certificates on a currently running Katello installation, run:
-
+.To update the certificates on a currently running Katello installation, run:
+----
```
adamlazik1, line 58:
```suggestion
[options="nowrap"]
----
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
```diff
@@ -70,7 +64,10 @@ To update the certificates on a currently running Katello installation, run:
 ----
 endif::[]
 
-. From the output of the `katello-certs-check` command, depending on your requirements, enter the `{foreman-installer}` command that installs a new {Project} with custom SSL certificates or updates certificates on a currently running {Project}.
+. Depending on your requirements, enter the `{foreman-installer}` command
+that installs a new {Project} with custom SSL certificates or updates certificates on a currently running {Project}.
```
adamlazik1, line 68:
```suggestion
. Depending on your requirements, enter the `{foreman-installer}` command that installs a new {Project} with custom SSL certificates or updates certificates on a currently running {Project}.
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
```diff
@@ -70,7 +64,10 @@ To update the certificates on a currently running Katello installation, run:
...
+The output of the `katello-certs-check` command may not be accurate in some cases.
+Therefore, you must follow the steps mentioned above instead of command outputs.
```
adamlazik1, line 70:
```suggestion
Therefore, you must follow the steps mentioned above instead of the command outputs.
```

adamlazik1 (commented):
Pointed out some things to consider.

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
mjivraja: I understand the inconsistencies in this version of the guide. I have found a few more things and modified them.

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
mjivraja: Addressed :+1:

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
```diff
@@ -35,45 +35,41 @@ If you set a wildcard value `*` for the certificate's Common Name `CN =` in the
 If the command is successful, it returns two `{certs-generate}` commands, one of which you must use to generate the certificate archive file for your {SmartProxyServer}.
 ifdef::satellite[]
 +
-.Example output of `katello-certs-check`
+.To use the certificate inside a _new-{smartproxy-example-com}_, run this command:
```
adamlazik1, line 38: Since there was `them` previously I assume there is multiple? Although I could be wrong.
```suggestion
.To use the certificates inside a _new-{smartproxy-example-com}_, run this command:
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
```diff
@@ -25,43 +25,38 @@ Note that for the `katello-certs-check` command to work correctly, Common Name (
...
   {foreman-installer} --scenario satellite \
     --certs-server-cert "_/root/satellite_cert/satellite_cert.pem_" \
     --certs-server-key "_/root/satellite_cert/satellite_cert_key.pem_" \
     --certs-server-ca-cert "_/root/satellite_cert/ca_cert_bundle.pem_"
-
-To update the certificates on a currently running Red Hat Satellite installation, run:
-
+----
++
+.To update the certificates on a currently running {ProjectName} installation, run:
```
adamlazik1, line 37:
```suggestion
.To update the certificates on a currently running {Project} installation, run:
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
```diff
@@ -70,7 +65,9 @@ To update the certificates on a currently running Katello installation, run:
 ----
 endif::[]
 
-. From the output of the `katello-certs-check` command, depending on your requirements, enter the `{foreman-installer}` command that installs a new {Project} with custom SSL certificates or updates certificates on a currently running {Project}.
+. Depending on your requirements, enter the `{foreman-installer}` command that installs a new {Project} with custom SSL certificates or updates certificates on a currently running {Project}.
```
adamlazik1, line 68:
```suggestion
. Depending on your requirements, enter the `{foreman-installer}` command that installs a new {ProjectServer} with custom SSL certificates or updates certificates on a currently running {ProjectServer}.
```

adamlazik1 (commented):
Two little things I haven't caught before + one query, other than that this is good to go :+1:

mjivraja:
> Two little things I haven't caught before + one query, other than that this is good to go +1

Sure and Thank you :+1:

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
```diff
@@ -35,45 +35,41 @@ If you set a wildcard value `*` for the certificate's Common Name `CN =` in the
...
-    --certs-tar  "~$FOREMAN_PROXY-certs.tar" \
+    --certs-tar  "~/$FOREMAN_PROXY-certs.tar" \
     --server-cert "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem_" \
     --server-key "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert_key.pem_" \
     --server-ca-cert "_/root/{smart-proxy-context}_cert/ca_cert_bundle.pem_" \
-
-To use them inside an EXISTING $FOREMAN_PROXY, run this command INSTEAD:
+----
++
+.To use the certificate inside an existing $FOREMAN_PROXY, run this command instead:
```
adamlazik1, line 71:
```suggestion
.To use the certificates inside an existing $FOREMAN_PROXY, run this command instead:
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
```diff
@@ -35,45 +35,41 @@ If you set a wildcard value `*` for the certificate's Common Name `CN =` in the
...
     --server-ca-cert "_/root/{smart-proxy-context}_cert/ca_cert_bundle.pem_" \
-    --certs-update-server
+    --certs-update-all
 ----
 endif::[]
 
 ifndef::satellite[]
 +
-.Example output of `katello-certs-check`
+.To use the certificate inside a new $FOREMAN_PROXY, run this command:
```
adamlazik1, line 61:
```suggestion
.To use the certificates inside a new $FOREMAN_PROXY, run this command:
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
```diff
@@ -35,45 +35,41 @@ If you set a wildcard value `*` for the certificate's Common Name `CN =` in the
...
     --certs-tar  "~/$CAPSULE-certs.tar" \
     --server-cert "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem_" \
     --server-key "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert_key.pem_" \
     --server-ca-cert "_/root/{smart-proxy-context}_cert/ca_cert_bundle.pem_" \
-
-To use them inside an EXISTING $CAPSULE, run this command INSTEAD:
-
+----
++
+.To use the certificate inside an _existing-{smartproxy-example-com}_, run this command instead:
```
adamlazik1, line 48:
```suggestion
.To use the certificates inside an _existing-{smartproxy-example-com}_, run this command instead:
```

adamlazik1 (commented):
So since you accepted my [previous suggestion](https://github.com/theforeman/foreman-documentation/pull/1678#discussion_r1108541512), I assume other instances of this should be corrected as well.

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-capsule-server.adoc
mjivraja: Ahh yes! I missed it.

adamlazik1 (approved):
LGTM

adamlazik1:
Merged to 2.5.

PR #1608, opened 2022-09-07
Add EL8 to External Database guide alongside EL7

mjivraja:
Adding the reference of EL8 alongside EL7 for the upstream v3.1. It is required to keep the reference of both as they are supported by Foreman 3.1/Katello 4.3

Need to update guide to have instructions to use EL8 based external DB

https://bugzilla.redhat.com/show_bug.cgi?id=2097380

* [X] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [ ] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3
* For Foreman 3.0 or older, please create a separate PR.
* We do not accept PRs for Foreman 2.3 or older.

PR #1529, opened 2022-08-04
Add procedure to import kickstart repos for EL9

maximiliankolb:
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.3/Katello 4.5

PR #1709, opened 2022-10-17
Fix EL8 to External Database guide alongside EL7

mjivraja:
Adding the reference of EL8 alongside EL7 for the upstream v3.1. It is required to keep the reference of both as they are supported by Foreman 3.1/Katello 4.3

Need to update guide to have instructions to use EL8 based external DB

https://bugzilla.redhat.com/show_bug.cgi?id=2097380

* [ ] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/proc_installing-postgresql.adoc
+++ b/guides/common/modules/proc_installing-postgresql.adoc
@@ -2,12 +2,115 @@
 = Installing PostgreSQL
 
 You can install only the same version of PostgreSQL that is installed with the `{foreman-installer}` tool during an internal database installation.
-You can install PostgreSQL using {RHELServer} 7 repositories or from an external source, as long as the version is supported.
-{Project} supports PostgreSQL version 12.1.
+You can install PostgreSQL using {EL} 8 or {RHELServer} 7 repositories.
+{Project} supports PostgreSQL version 12.
 
 include::snip_firewalld.adoc[]
 
-.Procedure
+* xref:#pgsql-el-8[{EL} 8]
+* xref:#pgsql-el-7[{EL} 7]
+
+== [[pgsql-el-8]]{EL} 8
+
+. To install PostgreSQL, enter the following command:
++
+----
+ifdef::katello,satellite,orcharhino[]
+# dnf install postgresql-server postgresql-evr
+endif::[]
+ifndef::katello,satellite,orcharhino[]
+# dnf install postgresql-server
+endif::[]
+----
+
+. To initialize PostgreSQL, enter the following command:
++
+[options="nowrap" subs="verbatim,quotes"]
+----
+# postgresql-setup initdb
+----
++
+. Edit the `/var/lib/pgsql/data/postgresql.conf` file:
++
+[options="nowrap" subs="verbatim,quotes"]
+----
+# vi /var/lib/pgsql/data/postgresql.conf
+----
++
+. Remove the `#` and edit to listen to inbound connections:
++
+[options="nowrap" subs="verbatim,quotes"]
+----
+listen_addresses = '*'
+----
++
+. Edit the `/var/lib/pgsql/data/pg_hba.conf` file:
++
+[options="nowrap" subs="verbatim,quotes"]
+-----
+# vi /var/lib/pgsql/data/pg_hba.conf
+-----
++
+. Add the following line to the file:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+  host  all   all   _{Project}_ip_/24   md5
+----
+
+. To start, and enable PostgreSQL service, enter the following commands:
++
+[options="nowrap" subs="verbatim,quotes"]
+----
+# systemctl start postgresql
+# systemctl enable postgresql
+----
+
+. Open the *postgresql* port on the external PostgreSQL server:
++
+[options="nowrap" subs="verbatim,quotes"]
+----
+# firewall-cmd --add-service=postgresql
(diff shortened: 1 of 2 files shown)


mjivraja:
This PR is referring to #1378 and #1608

guides/common/modules/proc_installing-postgresql.adoc
```diff
@@ -2,12 +2,115 @@
...
+You can install PostgreSQL using {EL} 8 or {RHELServer} 7 repositories.
+{Project} supports PostgreSQL version 12.
 
 include::snip_firewalld.adoc[]
 
-.Procedure
+* xref:#pgsql-el-8[{EL} 8]
+* xref:#pgsql-el-7[{EL} 7]
+
+== [[pgsql-el-8]]{EL} 8
```
adamlazik1, line 13:
```suggestion
== [[pgsql-el-8]]Installing PostgreSQL on {EL} 8

.Procedure
```

adamlazik1 (commented):
Made a set of suggestions. Please also reflect the changes in EL 7 section. I can't seem to suggest anything there because the content wasn't changed in this PR.

evgeni:
> Made a set of suggestions. Please also reflect the changes in EL 7 section. I can't seem to suggest anything there because the content wasn't changed in this PR.

Those are all fine suggestions, but this PR is a "we need to special-case the 3.1 branch that supports woth EL7 and EL8.
I think we should focus on that and port your suggestions into a clean PR against `master` to be used in the EL8-only branches?

evgeni:
@maximiliankolb is it OK for Orcharino to also show *both* EL7 and EL8 instructions for 3.1? Or should this be "if not orcharino" (Foreman and Satellite are fine, but I am never sure about Orcha)

adamlazik1:
@evgeni 
> Those are all fine suggestions, but this PR is a "we need to special-case the 3.1 branch that supports woth EL7 and EL8. I think we should focus on that and port your suggestions into a clean PR against `master` to be used in the EL8-only branches?

Understood. So we should only focus on the technical content in this PR?

evgeni:
> Understood. So we should only focus on the technical content in this PR?

That's what I'd do. (But then again, I am just a dev :))

adamlazik1:
> Given the only differences in the procedure are the package names and the paths to the configuration files, I wonder if we only should differentiate in these steps (vs the whole procedure)? Are there docs best practices on such issues?

This could be done by extracting the introductory part of the module into new concept module and then including the procedural module twice, each time with redefined variable for EL version. Then all you'd have to do is to wrap the different passages in `ifeval::[]` blocks and you would be good with only one procedure. For reference, see https://github.com/theforeman/foreman-documentation/pull/1529/files. You can notice that the same module was used for EL 9, 8 and 7. 

> > Understood. So we should only focus on the technical content in this PR?
> 
> That's what I'd do. (But then again, I am just a dev :))

It depends on whether this documentation change has to be merged ASAP or whether there is time to give it a facelift.

mjivraja:
> Given the only differences in the procedure are the package names and the paths to the configuration files, I wonder if we only should differentiate in these steps (vs the whole procedure)? Are there docs best practices on such issues?

I went through that approach first. But then, it didn't convince me from the end-users perspective. So with `xref` I tried to navigate them effortlessly to avoid confusion even for a second. However, I am open to your idea as well. :+1:

evgeni:
> > Given the only differences in the procedure are the package names and the paths to the configuration files, I wonder if we only should differentiate in these steps (vs the whole procedure)? Are there docs best practices on such issues?
> 
> I went through that approach first. But then, it didn't convince me from the end-users perspective. So with `xref` I tried to navigate them effortlessly to avoid confusion even for a second. However, I am open to your idea as well. +1

:+1: I'm cool either way, just wanted to be sure we thought about the alternatives.

> > > Understood. So we should only focus on the technical content in this PR?
> > 
> > 
> > That's what I'd do. (But then again, I am just a dev :))
> 
> It depends on whether this documentation change has to be merged ASAP or whether there is time to give it a facelift.

Given this is updating docs for an existing release, I'd rather see it done ASAP.

evgeni:
FTR, the repository steps are not correct in the 3.1 branch, but also not touched in this PR, so I've opened https://github.com/theforeman/foreman-documentation/pull/1713 to fix that.

adamlazik1:
@evgeni If the procedure looks correct to you and you do not require any further changes, feel free to merge.

mjivraja:
> FTR, the repository steps are not correct in the 3.1 branch, but also not touched in this PR, so I've opened #1713 to fix that.

That's really great and I see the PR also got merged. Let me know if I need to modify anything in this PR.

evgeni:
Nah, I think this PR is fine as is. Just waiting on @maximiliankolb to tell me whether the Orcharino changes are fine, or whether w need to specialcase it somehow

maximiliankolb:
I will have a look at this tomorrow. Thanks for pinging me again.

maximiliankolb (commented):
No additional comments from me; just what has been said before.

evgeni (approved):

evgeni:
Thanks!

I've applied @adamlazik1 suggestions for the headings, as these are specific to this branch, for the others, please open a PR against `master` and we backport from there.

PR #1713, opened 2022-10-19
properly enable the needed repos on el8 too

evgeni:
Fixes: 35285d7b17647440e22f54d99cd5a93084728f75

* [ ] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/proc_preparing-a-host-for-external-databases.adoc
+++ b/guides/common/modules/proc_preparing-a-host-for-external-databases.adoc
@@ -41,11 +41,44 @@ ifdef::satellite,katello[]
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
+# subscription-manager repos --disable '*'
+# subscription-manager repos \
+ifdef::satellite[]
+--enable={RepoRHEL8ServerSatelliteServerProductVersion} \
+endif::[]
+--enable={RepoRHEL8BaseOS} \
+--enable={RepoRHEL8AppStream}
+----
+ifdef::katello[]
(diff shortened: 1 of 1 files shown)


guides/common/modules/proc_preparing-a-host-for-external-databases.adoc
```diff
@@ -41,11 +41,44 @@ ifdef::satellite,katello[]
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
+# subscription-manager repos --disable '*'
+# subscription-manager repos \
+ifdef::satellite[]
+--enable={RepoRHEL8ServerSatelliteServerProductVersion} \
+endif::[]
+--enable={RepoRHEL8BaseOS} \
```
ekohl, line 49: I think the heading on line 40 should be `{RHEL}` instead of `{EL}` given these are the RHEL repo names. And the same on line 23.

ekohl (approved):
Minor note, but what you add looks correct.

guides/common/modules/proc_preparing-a-host-for-external-databases.adoc
evgeni: Technically, correct. But this is the 3.1 branch, which is not supported by Foreman anyways anymore, and it renders correctly for Satellite, so I went with the smaller diff ;)

PR #1719, opened 2022-10-20
Remove reference for bug closed as WONTFIX

bangelic:
Bug #1649764 is closed as WONTFIX so the step referencing the bug
needing to be resolved has been updated to reflect that. The change is
required because the bug has been closed as WONTFIX.
Related to Jira ticket:
https://issues.redhat.com/browse/SATDOC-1216

* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/doc-Upgrading_and_Updating/topics/upgrading_content_hosts.adoc
+++ b/guides/doc-Upgrading_and_Updating/topics/upgrading_content_hosts.adoc
@@ -41,10 +41,6 @@ ifdef::satellite[]
 +
 * If your deployment uses `katello-agent` and *goferd*, enter `katello-agent`.
 * If your deployment does not use `katello-agent` and *goferd*, enter `katello-host-tools`.
-+
-. Select *via remote execution* from the *Update* list.
-This is required because if you update the package using the Katello agent, the package update disrupts the communication between the client and {Project} or {SmartProxyServer}, which causes the update to fail.
-For more information, see {ManagingHostsDocURL}Configuring_and_Setting_Up_Remote_Jobs_managing-hosts[Configuring and Setting Up Remote Jobs] in _{ManagingHostsDocTitle}_.
 endif::[]
 
 [[cli-upgrading_content_hosts]]
--- a/guides/doc-Upgrading_and_Updating/topics/upgrading_content_hosts.adoc
+++ b/guides/doc-Upgrading_and_Updating/topics/upgrading_content_hosts.adoc
@@ -43,6 +43,9 @@ ifdef::satellite[]
 * If your deployment does not use `katello-agent` and *goferd*, enter `katello-host-tools`.
 +
 . https://bugzilla.redhat.com/show_bug.cgi?id=1649764[BZ#1649764] is closed as WONTFIX.
+To workaround the problem, from the *Update* list, you must select *via remote execution*.
+This is required because if you update the package using the Katello agent, the package update disrupts the communication between the client and {Project} or {SmartProxyServer}, which causes the update to fail.
+For more information, see {ManagingHostsDocURL}Configuring_and_Setting_Up_Remote_Jobs_managing-hosts[Configuring and Setting Up Remote Jobs] in _{ManagingHostsDocTitle}_.
 endif::[]
 
 [[cli-upgrading_content_hosts]]
--- a/guides/doc-Upgrading_and_Updating/topics/upgrading_content_hosts.adoc
+++ b/guides/doc-Upgrading_and_Updating/topics/upgrading_content_hosts.adoc
@@ -42,9 +42,7 @@ ifdef::satellite[]
 * If your deployment uses `katello-agent` and *goferd*, enter `katello-agent`.
 * If your deployment does not use `katello-agent` and *goferd*, enter `katello-host-tools`.
 +
-. Until https://bugzilla.redhat.com/show_bug.cgi?id=1649764[BZ#1649764] is resolved, from the *Update* list, you must select *via remote execution*.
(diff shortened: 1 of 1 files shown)


guides/doc-Upgrading_and_Updating/topics/upgrading_content_hosts.adoc
```diff
@@ -41,9 +41,7 @@ ifdef::satellite[]
 * If your deployment uses `katello-agent` and *goferd*, enter `katello-agent`.
 * If your deployment does not use `katello-agent` and *goferd*, enter `katello-host-tools`.
 +
-. Until https://bugzilla.redhat.com/show_bug.cgi?id=1649764[BZ#1649764] is resolved, from the *Update* list, you must select *via remote execution*.
-This is required because if you update the package using the Katello agent, the package update disrupts the communication between the client and {Project} or {SmartProxyServer}, which causes the update to fail.
-For more information, see {ManagingHostsDocURL}Configuring_and_Setting_Up_Remote_Jobs_managing-hosts[Configuring and Setting Up Remote Jobs] in _{ManagingHostsDocTitle}_.
+. https://bugzilla.redhat.com/show_bug.cgi?id=1649764[BZ#1649764] is closed as WONTFIX.
```
adamlazik1, line 44: Isn't the workaround still required if the bug is closed as won't fix?

guides/doc-Upgrading_and_Updating/topics/upgrading_content_hosts.adoc
```diff
@@ -41,7 +41,8 @@ ifdef::satellite[]
 * If your deployment uses `katello-agent` and *goferd*, enter `katello-agent`.
 * If your deployment does not use `katello-agent` and *goferd*, enter `katello-host-tools`.
 +
-. Until https://bugzilla.redhat.com/show_bug.cgi?id=1649764[BZ#1649764] is resolved, from the *Update* list, you must select *via remote execution*.
+. https://bugzilla.redhat.com/show_bug.cgi?id=1649764[BZ#1649764] is closed as WONTFIX.
+To workaround the problem, from the *Update* list, you must select *via remote execution*.
```
adamlazik1, line 45:
```suggestion
. Select *via remote execution* from the *Update* list.
```
IMHO you can drop the mention to BZ completely. Also please drop "To work around the problem" since no problem was introduced prior to this sentence.

adamlazik1:
Looking at the BZ again, it is not clear whether the workaround does or doesn't need to be included. I think this needs an SME review.

bangelic:
@evgeni Can you take a quick look at this? We need a SME to clarify or suggest changes.

evgeni (approved):
The change here is correct.

However, we could go further and remove the whole section (at least for Satellite), as since 6.11 we ship the client bits in a separate, unversioned repository, which is compatible across different Satellite versions, so users don't have to update their clients when they update their Satellite.

bangelic:
@evgeni Please check to see if that was the correct removal. You said section but I didn't think you wanted me to remove the entire Procedure so I removed that last step.

guides/doc-Upgrading_and_Updating/topics/upgrading_content_hosts.adoc
```diff
@@ -41,10 +41,6 @@ ifdef::satellite[]
 +
 * If your deployment uses `katello-agent` and *goferd*, enter `katello-agent`.
 * If your deployment does not use `katello-agent` and *goferd*, enter `katello-host-tools`.
-+
```
lpramuk, line 44: If your deployment does not use `katello-agent` and *goferd*, you are given two options in 6.12:
- *via remote execution*
- *via remote execution - customize first*

Removing reference to closed WONTFIX BZ is OK but you are also removing reference to `via remote execution` w/o any replacement

lpramuk (changes requested):
do not remove reference to `via remote execution`

evgeni:
> @evgeni Please check to see if that was the correct removal. You said section but I didn't think you wanted me to remove the entire Procedure so I removed that last step.

No, I really meant it. Updating any of the `katello-*` packages on clients during a Satellite upgrade is not required anymore in 6.12+ as we introduced a unified client repository in 6.11 which has content that is supported *regardless* of the Satellite version (given it is 6.11+).

guides/doc-Upgrading_and_Updating/topics/upgrading_content_hosts.adoc
lpramuk: Update:  on 6.12 there can still be **via Katello Agent** if KA is enabled 

so it is the same for both 6.11 and 6.12

bangelic:
> > @evgeni Please check to see if that was the correct removal. You said section but I didn't think you wanted me to remove the entire Procedure so I removed that last step.
> 
> No, I really meant it. Updating any of the `katello-*` packages on clients during a Satellite upgrade is not required anymore in 6.12+ as we introduced a unified client repository in 6.11 which has content that is supported _regardless_ of the Satellite version (given it is 6.11+).

@evgeni my latest commit to #1724 has removed the whole section (Procedure). If that gets merged, can we just close this PR?

evgeni:
> @evgeni my latest commit to #1724 has removed the whole section (Procedure). If that gets merged, can we just close this PR?

I believe removing the section will fix the issue here too, yeah. It's probably safe to close.

bangelic:
Closing this PR because #1724 already removes what needs to be removed in this PR.

PR #1728, opened 2022-10-25
Label double xref manually 3.3

Lennonka:
Due to a bug in our internal build system, I'm adding this workaround for two xrefs on a single line.
The bug causes wrong automatic selection of link labels -- the resulting links are rendered as a single link, which results in bad UX with the docs.
I'm not adding this to _master_, because we expect this bug to be fixed before the next release.

Cherry-pick to 3.1 and 2.5.

### DIFF
--- a/guides/common/modules/con_transport-modes-for-remote-execution.adoc
+++ b/guides/common/modules/con_transport-modes-for-remote-execution.adoc
@@ -22,7 +22,7 @@ ifdef::katello,orcharhino,satellite[]
 endif::[]
 ifdef::managing-hosts[]
 * To migrate a host from Katello Agent, see xref:Migrating_From_Katello_Agent_to_Remote_Execution_{context}[].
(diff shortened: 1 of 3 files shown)


Lennonka:
All I need is an ack, I'll do the cherry-picking. :)

ekohl (approved):
> Due to a bug in our internal build system, I'm adding this workaround for two xrefs on a single line.
> The bug causes wrong automatic selection of link labels -- the resulting links are rendered as a single link, which results in bad UX with the docs.
> I'm not adding this to master, because we expect this bug to be fixed before the next release.

Feel free to add this to the git commit itself. It helps when you use `git log` or `git blame`.

Lennonka:
Cherry-picked:
- 4789c395d..0c5fae3ca  3.1 -> 3.1

Not applicable to 2.5.

PR #1729, opened 2022-10-26
add dependabot config to update our actions

evgeni:

* [ ] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


ekohl (approved):
I'm always a bit annoyed when it tries to pin to minor or even patch versions instead of just the major version but let's see how this goes.

evgeni:
> I'm always a bit annoyed when it tries to pin to minor or even patch versions instead of just the major version but let's see how this goes.

given https://github.com/dependabot/dependabot-core/issues/2304 it should not happen anymore :)

ekohl:
I missed that one, but glad to see it's resolved.

PR #1762, opened 2022-11-08
Set 'dnf' as package manager for Satellite on 3.3

Lennonka:
Because Sat 6.12 runs on RHEL8 only, we need to use `dnf` to install, update and remove packages.


Lennonka:
@ekohl Ack?

ekohl (approved):

PR #1763, opened 2022-11-08
Refactor Managing File Type Content 3.3

Lennonka:
Clarification of terminology and workflow.

I'll add a PR for master as well, there would be a conflict.


guides/common/modules/proc_creating-a-local-source-for-a-custom-file-type-repository.adoc
```diff
@@ -0,0 +1,71 @@
+[id="Creating_a_Local_Source_for_a_Custom_File_Type_Repository_{context}"]
+= Creating a Local Source for a {customfiletypetitle} Repository
+
+You can create a {customfiletype} repository source, from a directory of files, on the base system where {Project} is installed using *Pulp Manifest*.
+You can then synchronize the files into {ProjectServer}.
+When you add files to a file type repository, you can work with the files as with any other repository.
```
maximiliankolb, line 6: This is unclear to me. Do you mean "you can manage content of type *file* as any other content, such as yum and deb"?

guides/common/modules/proc_creating-a-local-source-for-a-custom-file-type-repository.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+Use this procedure to configure a repository in a directory on the base system where {Project} is installed.
+To create a file type repository in a directory on a remote server, see xref:Creating_a_Remote_Source_for_a_Custom_File_Type_Repository_{context}[].
+
+.Procedure
+ifdef::satellite[]
+. Ensure the Utils repository is enabled.
```
maximiliankolb, line 13:
```suggestion
. Ensure the Utils repository is enabled:
```

guides/common/modules/proc_creating-a-local-source-for-a-custom-file-type-repository.adoc
```diff
@@ -0,0 +1,71 @@
...
+Alternatively, to prevent downtime caused by stopping the service, you can use the following:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foreman-maintain} packages unlock
+# {package-install} python3-pulp_manifest
+# {foreman-maintain} packages lock
+----
+endif::[]
+. Create a directory that you want to use as the file type repository in the HTTP server's public folder:
```
maximiliankolb, line 40: Does that imply that you'd need a specific path in line 44?

IMHO This is true because of `pulpcore_additional_import_paths` which means Foreman can only import (by default) from  the `/var/lib/pulp/imports/` directory if I'm not mistaken.

You could add your path, e.g. `/var/my_repo` to  `/etc/foreman-installer/scenarios.d/katello-answers.yaml` and then run `foreman-installer`.

I am sure there is a more elegant solution to this.

guides/common/modules/proc_creating-a-remote-source-for-a-custom-file-type-repository.adoc
```diff
@@ -0,0 +1,84 @@
+[id="Creating_a_Remote_Source_for_a_Custom_File_Type_Repository_{context}"]
+= Creating a Remote Source for a {customfiletypetitle} Repository
+
+You can create a {customfiletype} repository source from a directory of files that is external to {ProjectServer} using *Pulp Manifest*.
+You can then synchronize the files into {ProjectServer} over HTTP or HTTPS.
+When you add files to a file type repository, you can work with the files as with any other repository.
```
maximiliankolb, line 6: same as above.

guides/common/modules/proc_creating-a-remote-source-for-a-custom-file-type-repository.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+Use this procedure to configure a repository in a directory on a remote server.
+To create a file type repository in a directory on the base system where {ProjectServer} is installed, see xref:Creating_a_Local_Source_for_a_Custom_File_Type_Repository_{context}[].
+
+.Prerequisites
+* You have a server running {EL} 7 or 8 registered to your {Project} or the Red{nbsp}Hat CDN.
```
maximiliankolb, line 12: IMHO the "or to RH CDN" is also satellite-only.

guides/common/modules/proc_creating-a-remote-source-for-a-custom-file-type-repository.adoc
```diff
@@ -0,0 +1,84 @@
...
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# dnf install python3-pulp_manifest
+----
+** On {EL} 7:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# yum install tfm-pulpcore-python3-pulp_manifest
```
maximiliankolb, line 51: IMHO this package is not yet available: http://yum.theforeman.org/pulpcore/3.18/el7/source/ Therefore I am unsure about the package name.

I've opened an issue against pulpcore-packaging: https://github.com/theforeman/pulpcore-packaging/issues/598

guides/common/modules/proc_creating-a-local-source-for-a-custom-file-type-repository.adoc
Lennonka: How about:
`You can then synchronize the files into a repository on {ProjectServer} and manage the {customfiletype} content like any other content type.`
Would that make it sound clearer?

guides/common/modules/proc_creating-a-local-source-for-a-custom-file-type-repository.adoc
Lennonka: Actually, I think that the path will be the same here as with the "remote source" option.
In both cases, the repository is synced over HTTP/HTTPS because I "import" these repositories via an Upstream URL.
So any other paths shouldn't be necessary.
Does that make sense?

Lennonka: Okay, you're right! I got this
> If you do see some error saying not in allowed import paths, you might have to add it in /etc/pulp/settings.py in the ALLOWED_IMPORT_PATHS list..

guides/common/modules/proc_creating-a-local-source-for-a-custom-file-type-repository.adoc
maximiliankolb: Yes, this is much clearer to me.

guides/common/modules/proc_creating-a-remote-source-for-a-custom-file-type-repository.adoc
Lennonka: According to theforeman/pulpcore-packaging/pull/596, these package names should be okay now.

maximiliankolb (approved):
LGTM

Issue #1766, opened 2022-11-08
Update proc_installing-the-discovery-service.adoc


ekohl:
For Satellite/Orcharhino it tells uses to install `tfm-rubygem-smart_proxy_discovery` but that should be `rubygem-smart_proxy_discovery` on EL8. That also needs to be backported to stable releases.

Then for upstream it downloads http://downloads.theforeman.org/discovery/releases/3.5/ but that's not the latest release. In fact, there is a `/latest` there and that's the default so I wonder if we should drop the parameter altogether.

Additionally, the service restart and proxy refreshing is only needed when the installer isn't used, AKA the Satellite/Orcharhino instructions. Ideally a feature is implemented in the installer so users have a smooth installation, but for now this is where we are.

evgeni:
@ekohl I think that can now be closed?

PR #1774, opened 2022-11-09
Add note to remove tfm as package prefix for EL8

mjivraja:
The EL8 does not support tfm as a prefix for packages but the EL7 does. Therefore, we added a note to alert users to remove the prefix if they are carrying out the process for EL8. The tfm prefix remains unchanged for EL7. This is applicable to Project versions 3.2 and 3.1.

https://bugzilla.redhat.com/show_bug.cgi?id=2138391

* [ ] Foreman 3.5/Katello 4.7
* [ ] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5
* [X] Foreman 3.2/Katello 4.4
* [X] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/proc_configuring-the-discovery-service.adoc
+++ b/guides/common/modules/proc_configuring-the-discovery-service.adoc
@@ -41,12 +41,6 @@ endif::[]
 hosts or virtual machines can be created with incorrect network configurations. On Dell servers, the new naming scheme can be disabled via `biosdevname=1` kernel command line option, for other hardware or virtual machines, the new naming scheme can be completely turned off via `net.ifnames=1`.
 endif::[]
 
-Configure the Discovery Service on the applicable {EL} versions below.
-
-* xref:#configuring-discovery-el-8[{EL} 8]
-* xref:#configuring-discovery-el-7[{EL} 7]
-
-.[[configuring-discovery-el-8]]{EL} 8
 . Install Discovery on {ProjectServer}:
 ifdef::satellite,orcharhino[]
 +
@@ -77,24 +71,6 @@ ifndef::satellite,orcharhino[]
 --enable-foreman-proxy-plugin-discovery \
 --foreman-proxy-plugin-discovery-install-images=true
 ----
-endif::[]
-
-.[[configuring-discovery-el-7]]{EL} 7
-
-To use {ProjectServer} to provide the Discovery image, install the following RPM packages:
-
-* `tfm-rubygem-foreman_discovery`
-ifdef::satellite[]
-* `foreman-discovery-image`
-endif::[]
-* `tfm-rubygem-smart_proxy_discovery`
-
-The `tfm-rubygem-foreman_discovery` package contains the {Project} plug-in to handle discovered nodes, connections, and necessary database structures, and API.
-
-The `tfm-rubygem-smart_proxy_discovery` package configures {SmartProxyServer}, such as the integrated {SmartProxy} of {ProjectServer}, to act as a proxy for the Discovery service.
-
-ifndef::satellite[]
-You can use the `{foreman-installer}` tool to install the packages and the Discovery image.
 endif::[]
 
 When the installation completes, you can view the new menu option by navigating to *Hosts* > *Discovered Hosts*.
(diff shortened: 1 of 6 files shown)


guides/common/modules/con_configuring-the-discovery-service.adoc
```diff
@@ -49,6 +49,8 @@ ifdef::satellite[]
 endif::[]
 * `tfm-rubygem-smart_proxy_discovery`
 
+NOTE: For {EL} 8, remove `tfm` from the prefix of packages.
```
ekohl, line 52:
```suggestion
NOTE: For {EL} 8, remove `tfm-` from the prefix of packages.
```

ekohl (changes requested):
Note we should also update this in master. Generally speaking I always prefer to start in master and then work back to older versions. I also mentioned the `tfm-` prefix in https://github.com/theforeman/foreman-documentation/issues/1766.

mjivraja:
> I also mentioned the `tfm-` prefix in #1766.

Right, so I guess I have opened PR at right time. Ack your suggestion and apply soon. :+1:

guides/common/modules/proc_configuring-the-discovery-service.adoc
```diff
@@ -41,6 +41,46 @@ endif::[]
 hosts or virtual machines can be created with incorrect network configurations. On Dell servers, the new naming scheme can be disabled via `biosdevname=1` kernel command line option, for other hardware or virtual machines, the new naming scheme can be completely turned off via `net.ifnames=1`.
 endif::[]
 
+Configure the Discovery Service on the applicable {EL} versions below.
+
+* xref:#configuring-discovery-el-8[{EL} 8]
+* xref:#configuring-discovery-el-7[{EL} 7]
```
ekohl, line 47: There is no difference in instructions between EL 7 & 8: the installer takes care of all of that. Please align on only using the installer for both EL 7 & 8.

guides/common/modules/proc_installing-the-discovery-service.adoc
```diff
@@ -1,18 +1,50 @@
 [id="Installing_the_Discovery_Service_{context}"]
 = Installing the Discovery Service
 
-Complete the following procedure to enable the Discovery service on {SmartProxyServer}.
+On the applicable operating system, complete the following procedure to enable the Discovery service on {SmartProxyServer}.
 
-.Procedure
+* xref:#installing-discovery-el-8[{EL} 8]
+* xref:#installing-discovery-el-7[{EL} 7]
```
ekohl, line 7: There is no difference in instructions between EL 7 & 8: the installer takes care of all of that. Please align on only using the installer for both EL 7 & 8.

guides/common/modules/proc_installing-webhooks-plugin.adoc
```diff
@@ -12,8 +12,17 @@ After installing webhooks, you can configure {ProjectServer} to send webhook req
 # {foreman-installer} --enable-foreman-plugin-webhooks
 ----
 * Optionally, you can install the CLI plugin using the following command:
+
+** For {EL} 8:
```
ekohl, line 16: There is no difference in instructions between EL 7 & 8: the installer takes care of all of that. Please align on only using the installer for both EL 7 & 8.

ekohl (changes requested):

guides/common/modules/proc_configuring-the-discovery-service.adoc
mjivraja: Addressed :+1:

guides/common/modules/proc_installing-the-discovery-service.adoc
mjivraja: Addressed :+1:

guides/common/modules/proc_configuring-the-discovery-service.adoc
mjivraja: Addressed :+1: 
Removing the separate section for EL7 as the commands from EL8 are applicable to both, as per my conversation with SME.

guides/common/modules/proc_installing-the-discovery-service.adoc
```diff
@@ -3,39 +3,35 @@
...
 Complete the following procedure to enable the Discovery service on {SmartProxyServer}.
 
-.Procedure
 . Enter the following commands on {SmartProxyServer}:
 +
 ifdef::satellite,orcharhino[]
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {package-install-project} foreman-discovery-image tfm-rubygem-smart_proxy_discovery
+# # {foreman-installer} \
```
ekohl, line 11:
```suggestion
# {foreman-installer} \
```

ekohl (commented):
I think the inline comment is the last one.

mjivraja:
Hi @ekohl 

If this looks good to you, can we move further?

ekohl:
5a9bdd277..6228d1d0c  3.1 -> 3.1

PR #1785, opened 2022-11-11
don't enable the repos on rhel twice

evgeni:
Fixes: b5fdcc05a78142212ff85da84f85de33e8e02cb4

* [x] Foreman 3.5/Katello 4.7
* [x] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/proc_configuring-repositories.adoc
+++ b/guides/common/modules/proc_configuring-repositories.adoc
@@ -103,13 +103,7 @@ For more information about modules and lifecycle streams on {RHEL} 8, see https:
 ====
 endif::[]
 endif::[]
-+
 
-ifdef::foreman-el,katello[]
-include::proc_configuring-repositories-el.adoc[]
-endif::[]
-
-+
 [NOTE]
 ====
 If you are installing {ProductName} as a virtual machine hosted on {oVirt}, you must also enable the *Red{nbsp}Hat Common* repository, and install {oVirt} guest agents and drivers.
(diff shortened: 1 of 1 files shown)


maximiliankolb (approved):
Thanks! It's a duplicate of line 83-85 if I read the source correctly.

Also, this should have never happened, because we should only include snippets in modules, not other modules.

evgeni:
> Thanks! It's a duplicate of line 83-85 if I read the source correctly.

Yeah, it's an oversight from the "drop EL7" PR

> Also, this should have never happened, because we should only include snippets in modules, not other modules.

I have no idea what that means, but also don't see how a different "type" of include would have fixed that -- the "double" include was there for a reason: once for el7 once for el8, it was just not correctly removed when el7 was dropped.

ekohl:
> I have no idea what that means, but also don't see how a different "type" of include would have fixed that -- the "double" include was there for a reason: once for el7 once for el8, it was just not correctly removed when el7 was dropped.

There are different includes that we have. You may recognize `proc` (procedure) and `snip` (snippet). I think the rule is that procedures may only include snippets, not other procedures.

evgeni:
> > I have no idea what that means, but also don't see how a different "type" of include would have fixed that -- the "double" include was there for a reason: once for el7 once for el8, it was just not correctly removed when el7 was dropped.
> 
> There are different includes that we have. You may recognize `proc` (procedure) and `snip` (snippet). I think the rule is that procedures may only include snippets, not other procedures.

Right, but renaming `proc_configuring-repositories-el.adoc` to `snip_configuring-repositories-el.adoc` wouldn't have prevented the issue at hand, no?

ekohl:
Simply renaming probably not, no.

evgeni:
   76f852c4a..86d7b50b2  3.4 -> 3.4
   3a66ef42a..94c9e23b9  3.5 -> 3.5

PR #1804, opened 2022-11-15
Use Python 3 in CI workflows

ekohl:
Python 2 is EOL and this allows moving forward.

It looks like it's only used for linkchecker so cherry picking to all releases looks safe.

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.5/Katello 4.7
* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3
* For Foreman 3.0 or older, please create a separate PR.
* We do not accept PRs for Foreman 2.3 or older.

PR #1805, opened 2022-11-15
Enable linkchecking again

ekohl:
Cherry pick of https://github.com/theforeman/foreman-documentation/pull/1804.

PR #1806, opened 2022-11-15
Enable linkchecking again

ekohl:
This is a cherry pick of https://github.com/theforeman/foreman-documentation/pull/1804

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -18,7 +18,7 @@ env:
 
 jobs:
   build-ccutil:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     container:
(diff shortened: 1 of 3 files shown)


ekohl:
@adamlazik1 @Lennonka these look like legitimate Satellite link issues. In https://github.com/theforeman/foreman-documentation/pull/1805 I simply removed the link to the Puppet guide (which appears to be gone now), but in 3.2 it's safe because it's not a downstream version. That makes me unsure what the right solution here is.

adamlazik1:
Uh-oh! Something you did uncovered some previously undiscovered invalid links. I will make a fix PR within an hour.

adamlazik1:
#1809 I should catch all the links outputted by linkchecker hopefully.

PR #1799, opened 2022-11-15
Step update in Chainbooting virtual machines

dubewarsagar:
A step in the procedure Chainbooting virtual machines was incorrect. 
The same has been reported in this BZ-2134961.
This PR contains fix for reported issue for foreman 2.5.

https://bugzilla.redhat.com/show_bug.cgi?id=2134961

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [ ] Foreman 3.5/Katello 4.7
* [ ] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3
* For Foreman 3.0 or older, please create a separate PR.
* We do not accept PRs for Foreman 2.3 or older.

PR #1810, opened 2022-11-15
Revert "Step update in Chainbooting virtual machines" on 2.5

Lennonka:
Reverts theforeman/foreman-documentation#1799

@dubewarsagar I'm truly sorry, Sagar, my mistake. The original procedure in 2.5 looks actually correct. Hosts > [Provisioning Setup] Operating Systems is a GUI path that does exist. A host group does not have a "Templates" tab that is mentioned in the next step.


Lennonka:
The change should be made between the lines 157-159.

PR #1812, opened 2022-11-15
Update Simple Content Access wording

jeremylenz:
As of Katello 4.6, Simple Content Access status is no longer tied to the upstream manifest, so we need to update some wording to ensure it continues to be accurate.

* [x] Foreman 3.5/Katello 4.7
* [x] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/con_managing-red-hat-subscriptions.adoc
+++ b/guides/common/modules/con_managing-red-hat-subscriptions.adoc
@@ -9,7 +9,7 @@ All subscription information is available in your Red Hat Customer Portal accoun
 Before you can complete the tasks in this chapter, you must create a Red{nbsp}Hat Subscription Manifest in the Customer Portal.
 
 ifdef::satellite[]
-Note that the subscription model is deprecated and will be removed in a future release.
+Note that legacy subscription management, including the need to attach subscriptions to hosts and activation keys, is deprecated and will be removed in a future release.
 {Team} recommends that you use https://access.redhat.com/articles/simple-content-access[Simple Content Access] as a substitute.
 endif::[]
 
--- a/guides/common/modules/proc_disabling-subscription-connection.adoc
+++ b/guides/common/modules/proc_disabling-subscription-connection.adoc
@@ -2,7 +2,7 @@
 = Disabling Subscription Connection
 
 Disable subscription connection on disconnected {ProjectServer} to avoid connecting to the Red{nbsp}Hat Portal.
-This will also prevent you from refreshing the manifest, updating upstream entitlements, and changing the status of Simple Content Access.
+This will also prevent you from refreshing the manifest and updating upstream entitlements.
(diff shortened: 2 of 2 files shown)


guides/common/modules/con_managing-red-hat-subscriptions.adoc
```diff
@@ -9,7 +9,7 @@ All subscription information is available in your Red Hat Customer Portal accoun
 Before you can complete the tasks in this chapter, you must create a Red{nbsp}Hat Subscription Manifest in the Customer Portal.
 
 ifdef::satellite[]
-Note that the subscription model is deprecated and will be removed in a future release.
+Note that legacy subscription management, including the need to attach subscriptions to hosts and activation keys, is deprecated and will be removed in a future release.
```
jeremylenz, line 12: I added this change as well because subscriptions themselves are not deprecated. Manifests still contain subscriptions, and subscriptions are still required in order to meet contractual obligations.

adamlazik1: Thanks for PR!

Question about this line: Are there right now two subscription management models - new and legacy - or is there only one subscription management model which will be changed in the future release, thus making the current workflow unsupported?

jeremylenz: Right now both SCA and non-SCA subscription models are available. Non-SCA is deprecated but will still be available for several releases. SCA is now the default and will eventually be the only option.

adamlazik1, line 12: Thanks, change looks good me!
Just a tiny detail.
```suggestion
Note that the legacy subscription management, including the need to attach subscriptions to hosts and activation keys, is deprecated and will be removed in a future release.
```

adamlazik1 (approved):

guides/common/modules/con_managing-red-hat-subscriptions.adoc
jeremylenz: No need for a definite article there. "Legacy" is an adjective modifying "subscription management."

I would also be fine with "entitlement-based subscription management" if you like that better; I think I've seen that term elsewhere.

adamlazik1: Understood. Leaving the current form is fine.

adamlazik1:
Html check fail is unrelated. I am merging.

adamlazik1:
Merged and cherry-picked:
   1cf3679ab..ea722cac6  3.4 -> 3.4
   31fb39ded..e0d6be60e  3.5 -> 3.5

PR #1817, opened 2022-11-17
Drop SatelliteAnsibleVersion attribute

ekohl:
This also removes a section where the attribute was not used, but should have. Since it's EL7 specific and doesn't apply to EL8, it should be removed.

Fixes: eb0b825d2a29c3bf0d195f8a854da375037073be

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.5/Katello 4.7
* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3
* For Foreman 3.0 or older, please create a separate PR.
* We do not accept PRs for Foreman 2.3 or older.

PR #1819, opened 2022-11-17
Unify ansible module installation

ekohl:
This uses Package(s) instead of RPM since Debian doesn't use RPMs, but does use the same package name.

Technically the file and id should also be changed, but I'm submitting this as a draft for discussion. https://github.com/theforeman/foreman-documentation/pull/1817 will introduce some conflicts as well.

Is this a good idea or not?

* [ ] Foreman 3.5/Katello 4.7
* [ ] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3


adamlazik1:
I like the idea :+1:

ekohl:
Looks like @evgeni already did this properly in 797df1e484233ab3466a2aebd7371943916c67b9.

PR #1820, opened 2022-11-17
Make attribute conditionals more consistent

ekohl:
This uses the pattern that if there's an ifdef::X[] for one sentence and then another for the others, it uses ifndef::X[] for the other bits.  Sort of like an if/else condition. This makes it easier to see what's related and you don't miss a sentence on another build.

This should have no content change: the diff is expected to be empty.

I'm picking 3.5 since I worry about complex cherry picks. On the other hand, it may also complicate future cherry picks. Up for debate what we prefer.

* [x] Foreman 3.5/Katello 4.7
* [ ] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3


ekohl:
@maximiliankolb I think the diff is actually correct and it was wrong before. Mind having a look?

guides/common/modules/proc_configuring-satellite-for-outgoing-emails.adoc
```diff
@@ -58,7 +58,7 @@ The `SMTP username` and `SMTP password` specify the login credentials for the SM
 +
 For security reasons, both Sendmail location and Sendmail argument settings are read-only and can be only set in `/etc/foreman/settings.yaml`.
 Both settings currently cannot be set via `{foreman-installer}`.
-ifdef::foreman,katello,orcharhino[]
+ifndef::satellite[]
```
maximiliankolb, line 61:
```suggestion
ifndef::satellite,orcharhino[]
```

maximiliankolb (commented):
One small request, rest LGTM. Esp. the ifdefs for images are easier to read. Thanks!

ekohl:
159601dfc..6447fcec4  3.5 -> 3.5

PR #1824, opened 2022-11-17
Remove outdated Software Collections reference

ekohl:
With EL8 there are no software collections so these directories no longer exist.

Also removes redundant ifdef statements which are all nested within the same ifdef and can be removed.

Since it's scoped to Satellite and Satellite has dropped EL7, this should be picked to 3.3.

* [x] Foreman 3.5/Katello 4.7
* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/ref_storage-guidelines.adoc
+++ b/guides/common/modules/ref_storage-guidelines.adoc
@@ -69,16 +69,12 @@ Software collections are installed in the `/opt/rh/` and `/opt/theforeman/` dire
 Write and execute permissions by the root user are required for installation to the `/opt` directory.
 endif::[]
 
-ifdef::katello,satellite[]
 .Symbolic links
 
 You cannot use symbolic links for `/var/lib/pulp/`.
-endif::[]
 
-ifdef::katello,satellite[]
 ifeval::["{mode}" == "connected"]
 .Synchronized RHEL ISO
 If you plan to synchronize RHEL content ISOs to {Project}, note that all minor versions of {RHEL} also synchronize.
 You must plan to have adequate storage on your {Project} to manage this.
-endif::[]
 endif::[]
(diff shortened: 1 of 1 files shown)


guides/common/modules/ref_storage-guidelines.adoc
```diff
@@ -61,24 +61,12 @@ The bulk of storage resides in the `/var/lib/pulp/` directory.
 These end points are not manually configurable.
 Ensure that storage is available on the `/var` file system to prevent storage problems.
 
-ifdef::satellite[]
-.Software Collections
-
-Software collections are installed in the `/opt/rh/` and `/opt/theforeman/` directories.
-
-Write and execute permissions by the root user are required for installation to the `/opt` directory.
```
evgeni, line 69: Isn't that still true, given Puppet is still installed in `/opt`?

But then again, it's the wrong "section" anyways :woman_shrugging:

evgeni (approved):

guides/common/modules/ref_storage-guidelines.adoc
ekohl: Hmm, fair point but now I wonder what to do. Should we drop it or add a new section?

evgeni: I'd drop it like a potato and see who's up to finding it again ;)

ekohl:
231244fe6..dcc3b7a11  3.3 -> 3.3
   f12a58976..e9c3c0ab9  3.4 -> 3.4
   a95134d5a..59477f9fb  3.5 -> 3.5

PR #1827, opened 2022-11-17
Use systemctl enable --now

ekohl:
Using systemctl enable --now is equivalent to enable and start. It is already used in some places in the documentation, but this makes it consistent. The benefit for users is that they have to copy fewer commands.

It's still a draft since there are some oddities which I'll highlight.

* [x] Foreman 3.5/Katello 4.7
* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/proc_configuring-a-host-for-registration.adoc
+++ b/guides/common/modules/proc_configuring-a-host-for-registration.adoc
@@ -18,7 +18,7 @@ You must update each virtual machine configuration so that they receive updates 
 ** For {EL} 7:
 +
 ----
-# systemctl enable chronyd; systemctl start chronyd
+# systemctl enable --now chronyd
 ----
 * Ensure that the daemon `rhsmcertd` is enabled and running on the hosts.
 ** For {EL} 6:
@@ -29,7 +29,7 @@ You must update each virtual machine configuration so that they receive updates 
 ** For {EL} 7:
 +
 ----
-# systemctl start rhsmcertd
+# systemctl enable --now rhsmcertd
 ----
 
 .To Configure a Host for Registration:
--- a/guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
+++ b/guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
@@ -119,20 +119,11 @@ _990_
 . Export the DHCP configuration and lease files using NFS:
 +
 [options="nowrap" subs="+quotes,attributes"]
-ifndef::foreman-deb[]
 ----
-# {package-install} nfs-utils
-# systemctl enable rpcbind nfs-server
-# systemctl start rpcbind nfs-server nfs-lock nfs-idmapd
+# {package-install} {nfs-server-package}
+# systemctl enable nfs-server
+# systemctl start nfs-server
 ----
-endif::[]
-ifdef::foreman-deb[]
-----
-# {package-install} nfs-kernel-server nfs-common
-# systemctl enable rpcbind nfs-server
-# systemctl start rpcbind nfs-server nfs-idmapd
-----
-endif::[]
 
 . Create directories for the DHCP configuration and lease files that you want to export using NFS:
 +
--- a/guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
+++ b/guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
@@ -121,8 +121,7 @@ _990_
 [options="nowrap" subs="+quotes,attributes"]
 ----
 # {package-install} {nfs-server-package}
-# systemctl enable nfs-server
-# systemctl start nfs-server
+# systemctl enable --now nfs-server
 ----
 
 . Create directories for the DHCP configuration and lease files that you want to export using NFS:
--- a/guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
+++ b/guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
@@ -109,11 +109,11 @@ _990_
 # chattr +i /etc/dhcp/ /etc/dhcp/dhcpd.conf
 ----
 
-. Start the DHCP service:
+. Enable and start the DHCP service:
 +
 [options="nowrap"]
 ----
-# systemctl start dhcpd
+# systemctl enable --now dhcpd
 ----
 
 . Export the DHCP configuration and lease files using NFS:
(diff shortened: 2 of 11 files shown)


guides/common/modules/proc_configuring-a-host-for-registration.adoc
```diff
@@ -29,7 +29,7 @@ You must update each virtual machine configuration so that they receive updates
 ** For {EL} 7:
 +
 ----
-# systemctl start rhsmcertd
+# systemctl enable --now rhsmcertd
```
ekohl, line 32: Previously this was only start, but not enable while for EL 6 it did both enable and start. I think this is fixing a bug.

guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
```diff
@@ -122,15 +122,15 @@ _990_
...
+# systemctl start nfs-lock nfs-idmapd
 ----
 endif::[]
 ifdef::foreman-deb[]
 ----
 # {package-install} nfs-kernel-server nfs-common
-# systemctl enable rpcbind nfs-server
-# systemctl start rpcbind nfs-server nfs-idmapd
+# systemctl enable --now rpcbind nfs-server
+# systemctl start nfs-idmapd
```
ekohl, line 133: It's weird that some services are started but not enabled, which means that it's broken after a reboot? Not sure what would be best.

At least on Fedora it looks like `nfs-server` only has `Wants` on `nfs-idmapd` but no `Requires`. Possible bug?

guides/common/modules/proc_configuring-direct-ad-integration-with-gss-proxy.adoc
```diff
@@ -70,7 +70,7 @@ euid = __ID_of_Apache_User__
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
 # systemctl restart gssproxy.service
```
ekohl, line 72: The restart here is weird. Is there a specific reason for this?

evgeni: It was added in the "original" version of that doc (in https://bugzilla.redhat.com/show_bug.cgi?id=1206138) but I don't see any real reason for that, no.

However, looking at a system that I have here, which *does not have IPA enabled at all*, the `gssproxy` service is running, even if it's marked as not enabled:

```console
# systemctl status gssproxy
● gssproxy.service - GSSAPI Proxy Daemon
   Loaded: loaded (/usr/lib/systemd/system/gssproxy.service; disabled; vendor preset: disabled)
   Active: active (running) since Fri 2022-11-25 07:07:56 UTC; 1h 47min ago
 Main PID: 745 (gssproxy)
    Tasks: 6 (limit: 50696)
   Memory: 1.6M
   CGroup: /system.slice/gssproxy.service
           └─745 /usr/sbin/gssproxy -D
```

Some steps above we do edit the config, so if the service is already running, a restart is really required.

ekohl: I'm guessing something activates it. Either via a socket or something else via a requirement. Let's be safe and keep restart in then. It doesn't hurt in this order (first `enable --now` and then `restart` could cause a double service start, which is slower).

evgeni: I could not find *what* activated it, but yeah there must be something.

And agreed, it doesn't hurt that way, so let's keep it

guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
evgeni: This is Debian, which enables services by default. But I do agree, just put it to the `enable --now` line.

guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
```diff
@@ -122,15 +122,15 @@ _990_
 ifndef::foreman-deb[]
 ----
 # {package-install} nfs-utils
-# systemctl enable rpcbind nfs-server
-# systemctl start rpcbind nfs-server nfs-lock nfs-idmapd
+# systemctl enable --now rpcbind nfs-server
+# systemctl start nfs-lock nfs-idmapd
```
evgeni, line 126: what's `nfs-lock`, neither my EL8 nor my F36 systems know such a service?!

ah, looking at an EL7 system:
```console
# systemctl status nfs-lock
● rpc-statd.service - NFS status monitor for NFSv2/3 locking.
   Loaded: loaded (/usr/lib/systemd/system/rpc-statd.service; static; vendor preset: disabled)
   Active: inactive (dead)
# ls -alh /usr/lib/systemd/system/nfs-lock.service 
lrwxrwxrwx. 1 root root 17 Dec 21  2020 /usr/lib/systemd/system/nfs-lock.service -> rpc-statd.service
```

So I guess this should be swapped out, and also be moved to `enable --now`?

Can we then just fold the two code sections into one, only differentiating on the package line?

guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
```diff
@@ -122,15 +122,15 @@ _990_
...
 ----
 endif::[]
 ifdef::foreman-deb[]
 ----
 # {package-install} nfs-kernel-server nfs-common
```
evgeni, line 131: `nfs-kernel-server` depends on `nfs-common`, so I see no good reason to explicitly call that out? (scope creep, sorry)

guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
ekohl: It looks like `nfs-server` has wants on `rpcbind` and `nfs-idmapd` so they will be started. Should we reduce the list of services to just `nfs-server`? If you only want an NFSv4 server that's really all you need according to https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_file_systems/configuring-an-nfsv4-only-server_managing-file-systems

guides/common/modules/proc_configuring-satellite-deployment-with-an-external-dhcp-server.adoc
```diff
@@ -8,19 +8,12 @@ You can configure {ProductName} with an external DHCP server.
 For more information, see xref:configuring-an-external-dhcp-server_{context}[].
 
 .Procedure
-. Install the `nfs-utils` utility:
+. Install the `{nfs-client-package}` utility:
```
evgeni, line 11:
```suggestion
. Install the `{nfs-client-package}` package:
```

It's hardly *a* utility, more a collection thereof.

evgeni (approved):
One wording nitpick, but not blocking, ack

evgeni:
   fb6a16ce5..9ef18fd4d  3.3 -> 3.3
   8631e690f..7dbde1198  3.4 -> 3.4
   6e68f4be6..49ed39e61  3.5 -> 3.5

PR #1836, opened 2022-11-21
Re-introduce linkchecker-tryer

### LINKED ISSUE #1831: Valid URLs failing html checks
I have noticed in the past couple of days that valid URLs fail the html checks. Not any specific URLs also, but each time they seem to be random. 
As an example, in the last check in #1830 it was these URLs:
http://mirror.stream.centos.org/9-stream/BaseOS/x86_64/
https://www.postgresql.org/docs/current/runtime-config-logging.html
Both of them are valid.

Do we have any idea what is causing this issue?

ekohl:
The official [linkchecker-tryer project](https://github.com/terryyin/linkchecker-tryer) has been unmaintained for the past 7 years, but clearly it was serving a purpose. Many links sometimes give some network error and aren't really a 404 which was breaking CI needlessly.

This takes the code from the original project and makes it run on Python 3. It removes reduce and filter since they're fairly hard to understand.  The current code should be easier to follow.

I also thought about submitting the script elsewhere so we don't have to maintain it in multiple places, but then we have to deal with installing it.

@evgeni I'd appreciate a look at this.

Fixes #1831

* [x] Foreman 3.5/Katello 4.7
* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/Makefile
+++ b/guides/Makefile
@@ -1,7 +1,7 @@
 SHELL := /bin/bash
 SUBDIRS = $(shell ls -d doc-*)
 
-.PHONY: all clean html pdf linkchecker serve subdirs $(SUBDIRS)
+.PHONY: all clean html pdf linkchecker linkchecker-tryer serve subdirs $(SUBDIRS)
 
 all: html
 
@@ -21,3 +21,6 @@ clean:
 
 linkchecker:
 	find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern
+
+linkchecker-tryer:
+	find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern | ./scripts/linkchecker-tryer
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -82,7 +82,7 @@ jobs:
 
       - name: Check HTML links of all builds
         run: |
-          make linkchecker
+          make linkchecker-tryer
 
   build-html:
     runs-on: ubuntu-22.04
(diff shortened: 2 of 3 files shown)


guides/Makefile
```diff
@@ -21,3 +21,6 @@ clean:
 
 linkchecker:
 	find build -type f -name index\*html | xargs linkchecker -r1 -f common/linkchecker.ini --check-extern
+
+linkchecker:
```
evgeni, line 25: This target name seems wrong?

ekohl: Copy paste gone wrong :(

Having said that: how is this passing:
```
Run make linkchecker-tryer
Makefile:26: warning: overriding recipe for target 'linkchecker'
Makefile:23: warning: ignoring old recipe for target 'linkchecker'
make: Nothing to be done for 'linkchecker-tryer'.
```

Is that because it's a phony target?

evgeni: Probably?

evgeni (approved):

adamlazik1 (approved):

maximiliankolb:
   107ee959a..ff1efe508  3.5 -> 3.5
   65dc2e422..772c6a6f7  3.4 -> 3.4
   30e0a0ab9..6d21f486c  3.3 -> 3.3
   dd3ed43b5..4b818b362  3.2 -> 3.2
   e0c5e4db0..92224a37a  3.1 -> 3.1

PR #1838, opened 2022-11-24
Show create OS for both Debian and Ubuntu

maximiliankolb:
This is due to user feedback: we should ensure users don't set the minor version for Ubuntu 22.04 when trying to provision hosts using Ubuntu Autoinstall. cc @bastian-src

* [x] Foreman 3.5/Katello 4.7
* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/assembly_configuring-provisioning-resources.adoc
+++ b/guides/common/assembly_configuring-provisioning-resources.adoc
@@ -7,6 +7,16 @@ include::modules/proc_setting-the-provisioning-context.adoc[leveloffset=+1]
 include::modules/proc_creating-operating-systems.adoc[leveloffset=+1]
 
 ifndef::satellite[]
+:os_codename: bullseye
+:os_major: 11
+:os_minor: 5
+:os_name: Debian
+include::modules/proc_creating-an-operating-system-for-debian.adoc[leveloffset=+1]
+
+:os_codename: focal
+:os_major: 20.04
+:os_minor: 5
+:os_name: Ubuntu
 include::modules/proc_creating-an-operating-system-for-debian.adoc[leveloffset=+1]
 endif::[]
 
--- a/guides/common/modules/proc_creating-an-operating-system-for-debian.adoc
+++ b/guides/common/modules/proc_creating-an-operating-system-for-debian.adoc
@@ -1,8 +1,8 @@
-[id="Creating_an_Operating_System_for_Debian_{context}"]
-= Creating an Operating System for Debian
+[id="Creating_an_Operating_System_for_{os_name}_{context}"]
+= Creating an Operating System for {os_name}
 
-Create an operating system in {Project} to provision hosts running Debian or Ubuntu.
-This examples uses Debian 11 "Bullseye".
+Create an operating system in {Project} to provision hosts running {os_name}.
+This example creates an operating system entry for {os_name} {os_major}.{os_minor}.
 
 ifdef::orcharhino[]
 [TIP]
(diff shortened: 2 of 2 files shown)


guides/common/modules/proc_creating-an-operating-system-for-debian.adoc
```diff
@@ -14,15 +13,23 @@ endif::[]
...
 Choose a name as reported by Ansible, Puppet, or Salt as fact.
-. Set the *Major Version* to `11` and omit the *Minor Version*.
-. Set the *Release Name* to `bullseye` for Debian 11 or `focal` for Ubuntu 20.04.
+. Set the *Major Version* to `{os_major}` and omit the *Minor Version*.
+ifeval::["{os_name}" == "Ubuntu"]
++
+[CAUTION]
+====
+If you set a minor version, you cannot provision hosts running Ubuntu 22.04 using Autoinstall.
+{Project} uses the major version to assemble the path for the ISO image, for example `/var/www/html/pub/installation_media/{os_name}/{os_major}-x86_64/`.
```
ekohl, line 25: `/var/www/html/pub` is Katello-only.

guides/common/modules/proc_creating-an-operating-system-for-debian.adoc
```diff
@@ -14,15 +13,23 @@ endif::[]
...
+. Enter `{os_name} {os_major}` as *Name* for the operating system.
 Choose a name as reported by Ansible, Puppet, or Salt as fact.
-. Set the *Major Version* to `11` and omit the *Minor Version*.
-. Set the *Release Name* to `bullseye` for Debian 11 or `focal` for Ubuntu 20.04.
+. Set the *Major Version* to `{os_major}` and omit the *Minor Version*.
+ifeval::["{os_name}" == "Ubuntu"]
++
+[CAUTION]
+====
+If you set a minor version, you cannot provision hosts running Ubuntu 22.04 using Autoinstall.
```
un-kasimir, line 24: You actually can (and should) set a minor version, but it must be set correctly.
For example with `Ubuntu 20.04`, some users would set `20` as a major and `04` as a minor. But it must be `20.04` as a major and then the corresponding sub-releases as a minor (e.g. `1`, `2`, `3`..). Starting with `Ubuntu 20.04.3`, the new Autoinstall mechanism must be used and we evaluate the major/minor field in order to identify this.
So, I would suggest something like the following:

```suggestion
If you create an operating system entry for Ubuntu in {Project}, it is important to pay attention to the versioning. For example in Ubuntu 20.04, the major field must be set to 20.04. Only in case of a follow-up release like 20.04.3, the major version is kept as 20.04 and the minor version is set to 3.
```

guides/common/modules/proc_creating-an-operating-system-for-debian.adoc
un-kasimir: I would even suggest removing this line. Otherwise, if you want to keep an argument for the statement above, use the following:

```suggestion
Depending on the major/minor version, {Project} searches at a different path for the boot files in the Installation Media. Especially for Ubuntu/Debian, the path is different according to the operating system version. 
```

guides/common/modules/proc_creating-an-operating-system-for-debian.adoc
bastian-src, line 24: You actually can (and should) set a minor version, but it must be set correctly.
For example with `Ubuntu 20.04`, some users would set `20` as a major and `04` as a minor. But it must be `20.04` as a major and then the corresponding sub-releases as a minor (e.g. `1`, `2`, `3`..). Starting with `Ubuntu 20.04.3`, the new Autoinstall mechanism must be used and we evaluate the major/minor field in order to identify this.
So, I would suggest something like the following:

```suggestion
If you create an operating system entry for Ubuntu in {Project}, it is important to pay attention to the versioning. For example in Ubuntu 20.04, the major field must be set to 20.04. Only in case of a follow-up release like 20.04.3, the major version is kept as 20.04 and the minor version is set to 3.
```

guides/common/modules/proc_creating-an-operating-system-for-debian.adoc
bastian-src: I would even suggest removing this line. Otherwise, if you want to keep an argument for the statement above, I'd suggest the following:

```suggestion
Depending on the major/minor version, {Project} searches at a different path for the boot files in the Installation Media. Especially for Ubuntu/Debian, the path is different according to the operating system version. 
```

maximiliankolb: Reworded.

maximiliankolb (commented):
Applied feedback and rebased to master. Please re-review @bastian-src

bastian-src:
LGTM! :1st_place_medal: 
I like the rewording!

guides/common/assembly_configuring-provisioning-resources.adoc
```diff
@@ -7,6 +7,16 @@ include::modules/proc_setting-the-provisioning-context.adoc[leveloffset=+1]
 include::modules/proc_creating-operating-systems.adoc[leveloffset=+1]
 
 ifndef::satellite[]
+:os_codename: Bullseye
```
ekohl, line 10: At least facter reports it in lower case so I'd expect:
```suggestion
:os_codename: bullseye
```

maximiliankolb: Applied.

ekohl (approved):

PR #1840, opened 2022-11-25
Update steps in upgrading Proxy

mjivraja:
The present upgrade command where maintain package was not able to process the value correctly. As a result, the version of project upgraded is incorrect. The new maintain package and enablement of maintenance repos will be able to resolve it. Also, the rubygem maintain package installed from older repo didn't had the bug fixed. Therefore, the new commands will help end-users to upgrade to correct project version.

https://bugzilla.redhat.com/show_bug.cgi?id=2148435

* [ ] Foreman 3.5/Katello 4.7
* [ ] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5
* [X] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
+++ b/guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
@@ -110,7 +110,7 @@ Use `grep tar_file /etc/foreman-installer/scenarios.d/capsule-answers.yaml` on y
 # yum clean metadata
 ----
 +
-. The `rubygem-foreman_maintain` is installed from the {Project} Maintenance repository or upgraded from the {Project} Maintenance repository if currently installed.
+. Ensure {SmartProxy} has access to `{RepoRHEL7ServerSatelliteMaintenanceProductVersion}` and update {foreman-maintain}.
 +
 [options="nowrap" subs="attributes"]
 ----
@@ -134,7 +134,7 @@ Use `grep tar_file /etc/foreman-installer/scenarios.d/capsule-answers.yaml` on y
 . Because of the lengthy upgrade time, use a utility such as `tmux` to suspend and reattach a communication session.
 You can then check the upgrade progress without staying connected to the command shell continuously.
 +
-If you lose connection to the command shell where the upgrade command is running you can see the logged messages in the `{installer-smartproxy-log-file}` file to check if the process completed successfully.
+If you lose connection to the command shell where the upgrade command is running, you can see the logged messages in the `{installer-smartproxy-log-file}` file to check if the process completed successfully.
 
 . Use the health check option to determine if the system is ready for upgrade:
 +
--- a/guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
+++ b/guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
@@ -112,11 +112,10 @@ Use `grep tar_file /etc/foreman-installer/scenarios.d/capsule-answers.yaml` on y
 +
 . The `rubygem-foreman_maintain` is installed from the {Project} Maintenance repository or upgraded from the {Project} Maintenance repository if currently installed.
 +
-Ensure {SmartProxy} has access to `{RepoRHEL7ServerSatelliteMaintenanceProductVersion}` and execute:
-+
 [options="nowrap" subs="attributes"]
 ----
-# {foreman-maintain} self-upgrade
+# subscription-manager repos --enable {RepoRHEL7ServerSatelliteMaintenanceProductVersion}
+# yum --disableplugin=foreman-protector update rubygem-foreman_maintain {foreman-maintain}
 ----
 
 . On {SmartProxyServer}, verify that the `foreman_url` setting points to the {Project} FQDN:


mjivraja:
I am looking to have a sequence for commands. Obviously, the upper one goes first. But still, I am wondering if we can number both code-blocks in point 5.

guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
```diff
@@ -116,7 +116,8 @@ Ensure {SmartProxy} has access to `{RepoRHEL7ServerSatelliteMaintenanceProductVe
 +
 [options="nowrap" subs="attributes"]
 ----
-# {foreman-maintain} self-upgrade
+# subscription-manager repos --enable rhel-7-server-satellite-maintenance-6.11-rpms
```
ekohl, line 119: If you do this, please use an attribute:

```suggestion
# subscription-manager repos --enable {RepoRHEL7ServerSatelliteMaintenanceProductVersion}
```

Having said that: I thought Capsules get their content from Satellite. In step 3.2 the repo should already be synced so I wonder why this is needed in the first place. However, I'm not that up to date on capsules and their content so perhaps I'm wrong.

ekohl (commented):
I'm

guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
mjivraja: >Having said that: I thought Capsules get their content from Satellite. In step 3.2 the repo should already be synced so I wonder why this is needed in the first place. However, I'm not that up to date on capsules and their content so perhaps I'm wrong.

It's been confirmed by relevant teams because this change was in priority.

guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
```diff
@@ -116,7 +116,8 @@ Ensure {SmartProxy} has access to `{RepoRHEL7ServerSatelliteMaintenanceProductVe
 +
```
ekohl, line 116: The line above this looks redundant now: that tells the user to ensure access and then run a command to ensure access.

guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
```diff
@@ -116,7 +116,8 @@ Ensure {SmartProxy} has access to `{RepoRHEL7ServerSatelliteMaintenanceProductVe
...
 [options="nowrap" subs="attributes"]
 ----
-# {foreman-maintain} self-upgrade
+# subscription-manager repos --enable {RepoRHEL7ServerSatelliteMaintenanceProductVersion}
+# yum --disableplugin=foreman-protector update -y rubygem-foreman_maintain {foreman-maintain}
```
ekohl, line 120: In general we have the convention to avoid `-y` for interactive commands:
```suggestion
# yum --disableplugin=foreman-protector update rubygem-foreman_maintain {foreman-maintain}
```

mjivraja: Addressed :+1:

guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
mjivraja: Addressed :+1:

guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
```diff
@@ -112,11 +112,10 @@ Use `grep tar_file /etc/foreman-installer/scenarios.d/capsule-answers.yaml` on y
 +
 . The `rubygem-foreman_maintain` is installed from the {Project} Maintenance repository or upgraded from the {Project} Maintenance repository if currently installed.
```
ekohl, line 113: I think this sentence was never correct - it always upgrades.

guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
```diff
@@ -112,11 +112,10 @@ Use `grep tar_file /etc/foreman-installer/scenarios.d/capsule-answers.yaml` on y
...
 +
-Ensure {SmartProxy} has access to `{RepoRHEL7ServerSatelliteMaintenanceProductVersion}` and execute:
-+
 [options="nowrap" subs="attributes"]
 ----
-# {foreman-maintain} self-upgrade
+# subscription-manager repos --enable {RepoRHEL7ServerSatelliteMaintenanceProductVersion}
+# yum --disableplugin=foreman-protector update rubygem-foreman_maintain {foreman-maintain}
```
ekohl, line 118: We have a convention to only have one command per code block.

ekohl (changes requested):

guides/doc-Upgrading_and_Updating/topics/upgrading_capsule_server.adoc
mjivraja: Does the second command have a different job? The reason is I have come across multi-line commands per block.

ekohl: They do have different intentions. If you look at the old version there is the line "Ensure {SmartProxy} has access to `{RepoRHEL7ServerSatelliteMaintenanceProductVersion}`". The subscription-manager command does exactly that. The other command updates the foreman-/satellite-maintain application.

So you can keep it in a single command. Then you should keep the old description and change `execture` to `update {foreman-maintain}`.

The alternative is to describe each command separately.

Either way, I'd drop the sentence on line 113.

ekohl:
The macro usage looks correct.

@evgeni @ehelms you're more familiar with the repos. Would you mind taking a look?

mjivraja:
> The macro usage looks correct.
> 
> @evgeni @ehelms you're more familiar with the repos. Would you mind taking a look?

Hi @ekohl 
Can we proceed further if it looks good to you?

PR #1841, opened 2022-11-25
Add link to blog article about using git

maximiliankolb:
Hey! As already revealed Monday to some of you, I wrote a blog article on using git as a technical writer. I hope this makes contributing ever easier.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,3 +58,4 @@ Ideally, each individual change correlates to an individual commit.
 
 * [Contributing Guidelines for Github documentation](https://github.com/github/docs/blob/main/CONTRIBUTING.md)
 * [theforeman.org/contribute](https://theforeman.org/contribute.html)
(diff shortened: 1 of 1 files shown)


ekohl:
I wasn't aware of this, but may I recommend using `git clone -o upstream` instead of clone + rename?

As for "I have accidentally committed to master", you can also `git reset --soft upstream/master` instead of `HEAD~1`. And you probably already know that `HEAD^` is the same as `HEAD~1`. Perhaps also interesting to refer to `man 7 gitrevisions` (or git help revisions`).

"I want to restructure multiple changes on a single branch"

Using `git fetch` and `git pull` is redundant. pull is roughly equivalent to fetch + merge.

Also, big +1 for using a tool to visualize history. I'm a big fan of https://jonas.github.io/tig/ for that.

Suggestion for the future: explain how

adamlazik1 (approved):
LGTM! Thanks a bunch for the article :confetti_ball:

PR #1846, opened 2022-11-28
Add EL 9 to supported host OS list

adamlazik1:
Also inverted the order so the newest EL is on top.

[Related BZ](https://bugzilla.redhat.com/show_bug.cgi?id=2144382)

* [x] Foreman 3.5/Katello 4.7
* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3

### DIFF
--- a/guides/common/modules/con_registering_hosts.adoc
+++ b/guides/common/modules/con_registering_hosts.adoc
@@ -35,9 +35,10 @@ ifdef::satellite[]
 
 Hosts must use one of the following {RHEL} versions:
 
-* 6.1 or later*
-* 7.0 or later
+* 9.0 or later
 * 8.0 or later
+* 7.0 or later
+* 6.1 or later*
 
 NOTE: Red{nbsp}Hat Enterprise{nbsp}Linux versions 6.1, 6.2, and 6.3 require `subscription-manager` and related packages to be updated manually.
 For more information, see https://access.redhat.com/solutions/1256763[].


evgeni (approved):
Side note: the list of architectures in the next section ("Supported Architectures") is incomplete/wrong (depending on which RHEL version we're talking about), can you track this somewhere that it needs updates?

adamlazik1:
> Side note: the list of architectures in the next section ("Supported Architectures") is incomplete/wrong (depending on which RHEL version we're talking about), can you track this somewhere that it needs updates?

@evgeni  I think it could be a part of this PR if you wish. Would you be able to provide me with information which architectures are supported on which RHEL? Another question, do they differ between Satellite 6.11 and 6.12?

evgeni:
> 



> > Side note: the list of architectures in the next section ("Supported Architectures") is incomplete/wrong (depending on which RHEL version we're talking about), can you track this somewhere that it needs updates?
> 
> @evgeni I think it could be a part of this PR if you wish. Would you be able to provide me with information which architectures are supported on which RHEL? Another question, do they differ between Satellite 6.11 and 6.12?

I don't have that data at hand, so would prefer we get this one in as it does fix actual inconsistencies in the docs, and then find the right data for the other section.

And no, Satellite 6.11 added support for RHEL9 and I am not aware of any changes in the list of supported OSes in 6.12

evgeni:
   0376dd7d4..9d99f52d8  3.1 -> 3.1
   149f1ecb0..e0808ecb3  3.2 -> 3.2
   7159f2565..fb6a16ce5  3.3 -> 3.3
   b12f588f4..8631e690f  3.4 -> 3.4
   dec840fd6..6e68f4be6  3.5 -> 3.5

Issue #1858, opened 2022-12-02
RFE: dnf needs restarting -r for el8 in guides/doc-Upgrading_and_Updating/topics/snip_needs_reboot.adoc


pcfe:
Since Foreman on EL is EL8 these days (no more EL7), it would be nice if guides/doc-Upgrading_and_Updating/topics/snip_needs_reboot.adoc recommended `dnf needs-restarting -r` instead of `rpm --query --last kernel | head -n 1 && uname -r`

something along the lines of:

```
[...]
ifndef::foreman-deb[]
.. Check if a reboot is needed:
+
[options="nowrap"]
----
# dnf needs-restarting --reboothint
----
. Optional: If the previous command told you to reboot, then reboot the system:
endif::[]
+
[options="nowrap"]
----
# reboot
----
```

#### example of no reboot needed:

```
[root@foreman ~]# dnf needs-restarting --reboothint
No core libraries or services have been updated since boot-up.
Reboot should not be necessary.
```

#### example of reboot needed:

```
[root@foreman ~]# dnf needs-restarting --reboothint
Core libraries or services have been updated since boot-up:
  * kernel

Reboot is required to fully utilize these updates.
More information: https://access.redhat.com/solutions/27943
```

#### availability of the command

with (AFAIR the recommended install set)
```
%packages
@Base
```

this command is available.

ekohl:
And today I learned something new. Thanks, this is much easier. Could you come up with a pull request for this?

PR #1859, opened 2022-12-02
dnf needs-rebooting -r instead of 2 steps

pcfe:
Resolves: #1858

* [x] Foreman 3.4/Katello 4.6
* [ ] Foreman 3.3/Katello 4.5 (Satellite 6.12)
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3 (Satellite 6.11, orcharhino 6.1+)

Since `dnf needs -restarting` is available on el8, I tried to select the version that for running on RHEL/CentOS Stream require 8.

The plugin used is part of `dnf-plugins-core` which is installed on a system with `@Base` (and if I read comps.xml correctly actually comes with @core).

plugin presence has been successfully verified on both RHEL8 and CentOS Stream 8 @Base installs

### DIFF
--- a/guides/doc-Upgrading_and_Updating/topics/snip_needs_reboot.adoc
+++ b/guides/doc-Upgrading_and_Updating/topics/snip_needs_reboot.adoc
@@ -8,19 +8,12 @@ $ ls /run/reboot-required
 . Optional: If the `/run/reboot-required` file exists, reboot the system:
 endif::[]
 ifndef::foreman-deb[]
-.. Check the version of newest installed kernel:
 +
 [options="nowrap"]
 ----
-# rpm --query --last kernel | head -n 1
+# dnf needs-restarting --reboothint
 ----
-.. Compare this to the version of currently running kernel:
-+
-[options="nowrap"]
-----
-# uname --kernel-release
-----
-. Optional: If the newest kernel differs from the currently running kernel, reboot the system:
+. Optional: If the previous command told you to reboot, then reboot the system:
 endif::[]
 +
 [options="nowrap"]


ekohl (approved):
LGTM from a technical perspective. This certainly makes it a lot easier for users.

pcfe:
[…]
> The following output files are affected by this PR:
> 
>     * [Upgrading_and_Updating/index-foreman-el.html](https://theforeman-foreman-documentation-preview-pr-1859.surge.sh/nightly/Upgrading_and_Updating/index-foreman-el.html)
This rendered as I intended with 78715f96668eb7c6c24c235d98856c9b3164840e 
 
>     * [Upgrading_and_Updating/index-katello.html](https://theforeman-foreman-documentation-preview-pr-1859.surge.sh/nightly/Upgrading_and_Updating/index-katello.html)
This also rendered as I intended. 

>     * [Upgrading_and_Updating/index-orcharhino.html](https://theforeman-foreman-documentation-preview-pr-1859.surge.sh/nightly/Upgrading_and_Updating/index-orcharhino.html)
I do not know what base operating system orcharino supports. if it's one with `dnf` available then the ifdef is fine. OTOH if it still supports e.g. el7 as base OS, then this MR needs _extending the ifdef::[] or ifndef::[] directive_, I might need help with that.

>     * [Upgrading_and_Updating/index-satellite.html](https://theforeman-foreman-documentation-preview-pr-1859.surge.sh/nightly/Upgrading_and_Updating/index-satellite.html)
This also rendered as I intended.

maximiliankolb (commented):
Thanks @pcfe for your PR! change looks good to me if it only runs on Foreman on EL8 due to dnf.

ekohl:
Thanks for this great change. I wasn't aware of the command, but it makes the instructions so much easier. Cherry picks:
   a728539bd..4d24c287f  3.4 -> 3.4
   f712e21a9..c890768b1  3.5 -> 3.5

maximiliankolb:
> IIRC in Foreman 3.4's packaging we dropped compatibility with EL7 so I think it's extremely unlikely that orcharhino will run on it.

Yes, I was missing this piece of information. Thanks for clarifying, I will create a PR to add this to the PR template.

PR #1865, opened 2022-12-05
Add support EL versions to PR template

maximiliankolb:
based on https://github.com/theforeman/foreman-documentation/milestone/2

cc @ekohl

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,9 +4,9 @@
 Please cherry-pick my commits into:
 
 * [ ] Foreman 3.5/Katello 4.7 (planned Satellite 6.13)
-* [ ] Foreman 3.4/Katello 4.6
-* [ ] Foreman 3.3/Katello 4.5 (Satellite 6.12)
-* [ ] Foreman 3.2/Katello 4.4
-* [ ] Foreman 3.1/Katello 4.3 (Satellite 6.11, orcharhino 6.1+)
+* [ ] Foreman 3.4/Katello 4.6 (EL8 only)
+* [ ] Foreman 3.3/Katello 4.5 on EL7 & EL8 (Satellite 6.12 on EL8 only)
+* [ ] Foreman 3.2/Katello 4.4 on EL7 & EL8
+* [ ] Foreman 3.1/Katello 4.3 on EL7 & EL8 (Satellite 6.11 EL7/8, orcharhino 6.1 on EL7, orcharhino 6.2 on EL7/8)
 * For Foreman 3.0 or older, please create a separate PR.
 * We do not accept PRs for Foreman 2.3 or older.


maximiliankolb:
For me, the milestone is rather hard to find; at least I knew we had this already somewhere...

I could also create a PR to add this to the README. I usually pay attention if something affects EL versions, so I need to know which branch for which build target supports which EL versions.

regarding orcharhino: Yes, we generally have multiple releases based on a single Foreman/Katello/Pulpcore (major.minor) version. orcharhino 6.1 is based on F 3.1 on EL7 only; whereas orcharhino 6.2 is also based on F 3.1, but for EL7 and EL8. This does NOT make things easy for me...

ekohl:
> I could also create a PR to add this to the README. I usually pay attention if something affects EL versions, so I need to know which branch for which build target supports which EL versions.

Is it time for a `CHERRYPICKING.md`?

mjivraja:
This will be helpful :smiley: :heart: 

Thanks @maximiliankolb @ekohl

guides/README.md
```diff
@@ -112,6 +112,17 @@ You can disable the linkcheck job for specific URL pattern, for example for unre
...
+## Recovering from a broken build due to the comparision with the previous commit
+
+If you open a PR on Github, your feature branch gets built and compared to the target branch into which you want to merge.
+Therefore, the Github Action builds the target branch.
+This implies that even if your change works on its own, it depends on a working state in the target branch.
+
+If the target branch does not build, your PR fails due to the comparision.
+To fix this, you have to manually set the reference from "merge base of the target branch and your feature branch" to a commit ID of a working state, for example the parent commit of the faulty yet merged commit on the target branch.
+
+Example: [REVERT ME: Use last working state as base](https://github.com/theforeman/foreman-documentation/commit/0ea9e707b6850487e60753a40fc2b2791528d175)
```
ekohl, line 124: I'm not a fan of this. Instead, I'd say that we should merge it with a failing CI. The actual publish job doesn't care about a diff. I'd rather "pollute" the git log with potentially incomplete roll forward commits than rollback + roll forward.

maximiliankolb: I had trouble getting the GHA to run & therefore could not start the "publish" job because previous jobs failed. Therefore, changes were never  deployed to docs.theforeman.org.

Would I get a different GHA when running on "master" or any protected branch that is not a PR due to `if: github.event_name == 'pull_request'` in `.github/workflows/deploy.yml`?

If this is the case, then I'll revert the second commit.

guides/README.md
```diff
@@ -112,6 +112,17 @@ You can disable the linkcheck job for specific URL pattern, for example for unre
 Append your pattern to `guides/common/linkchecker.ini`.
 Example: [64d825cc9](https://github.com/theforeman/foreman-documentation/commit/64d825cc9da3992879dfbfc088988197edc9f33b)
 
+## Recovering from a broken build due to the comparision with the previous commit
```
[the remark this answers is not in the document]
maximiliankolb: Applied and rebased to master.

guides/README.md
ekohl: I've been thinking about this, but have been hesitant to move forward.

In my mind you should have a branch `master` for how to assemble all guides and then have release branches (`nightly`, `3.5`, etc). So essentially make `web` a top level directory in `master` and `guides` the top level.

This is also more what Antora has, so I've also thought about moving https://github.com/theforeman/foreman-documentation/issues/873 forward instead of building ourselves. My biggest problem is that I have no time to drive such an effort.

maximiliankolb: OK, I'll drop it for now and merge only the first commit.

I'll add the second commit to the issue you've linked.

PR #1875, opened 2022-12-08
Fix broken GHA pipeline for 3.4 branch

maximiliankolb:


maximiliankolb:
I have reverted the last eight commits in the proper order on my branch, because the last successful GHA was 0660bc948cb66ee5d923a8da9e90b113eca2c4f8. See https://github.com/theforeman/foreman-documentation/commits/3.4

Any ideas?

.github/workflows/deploy.yml
```diff
@@ -161,7 +161,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
         with:
-          ref: ${{ github.base_ref }}
+          ref: 0660bc948cb66ee5d923a8da9e90b113eca2c4f8
```
adamlazik1, line 164: Changes look good to me. However I do not know what this does so it should be approved by someone else. The ifdefs seem to be correct, though.

maximiliankolb: I assume this is the git reference to the commit to compare the PR to. If the build for this commit is broken, you cannot successfully build and therefore compare your changes to this. I did not look up what "base_ref" stands for, but I assume it's the merge base commit. As my commit indicates, this should be reverted after merging. See my second PR.

ekohl: Oh, I think I misread what this did. I somehow thought it was using a certain ref for `checkout`, not the target to be checked out.

I suspect that that if we merge this, the base will be working again. So if you revert the last commit and we then merge it, it probably will be green again.

PR #1888, opened 2022-12-13
Show Using KernelCare for orcharhino only

maximiliankolb:


ekohl (commented):
I didn't check, but does orcharhino install via the installer or via package directly now?

maximiliankolb:
> I didn't check, but does orcharhino install via the installer or via package directly now?

Via package, so I've cherry-picked the commit to 3.1 as well.

PR #1895, opened 2022-12-15
Add akamaiedge.net Host name in adding HTTP proxy

mjivraja:
The *.akamaiedge.net Host hame should also be part of Adding an HTTP Proxy while Importing Content. The reason is it is a part of Configuring the HTTP Proxy in the Installation Guide.

https://bugzilla.redhat.com/show_bug.cgi?id=2152785

* [X] Foreman 3.4/Katello 4.6
* [X] Foreman 3.3/Katello 4.5 (Satellite 6.12)
* [X] Foreman 3.2/Katello 4.4
* [X] Foreman 3.1/Katello 4.3 (Satellite 6.11, orcharhino 6.1+)

### DIFF
--- a/guides/common/modules/proc_adding-an-http-proxy.adoc
+++ b/guides/common/modules/proc_adding-an-http-proxy.adoc
@@ -4,7 +4,11 @@
 Use this procedure to add HTTP proxies to {Project}.
 You can then specify which HTTP proxy to use for Products, repositories, and supported compute resources.
 
-If {ProjectServer} uses a proxy to communicate with subscription.rhsm.redhat.com or subscription.rhn.redhat.com, and cdn.redhat.com then the proxy must not perform SSL inspection on these communications.
+.Prerequisites
+Your HTTP proxy must allow access to the following hosts:
+include::snip_host-names-for-http-proxy.adoc[]
+
+If {ProjectServer} uses a proxy to communicate with subscription.rhsm.redhat.com and cdn.redhat.com then the proxy must not perform SSL inspection on these communications.
 
 To use the CLI instead of the {ProjectWebUI}, see the xref:cli-adding-an-http-proxy[].
 
@@ -12,21 +16,6 @@ To use the CLI instead of the {ProjectWebUI}, see the xref:cli-adding-an-http-pr
 . In the {ProjectWebUI}, navigate to *Infrastructure* > *HTTP Proxies* and select *New HTTP Proxy*.
 . In the *Name* field, enter a name for the HTTP proxy.
 . In the *URL* field, enter the URL for the HTTP proxy, including the port number.
-The following host names are available:
-+
-[cols="2,1,1",options="header"]
-|====
-| Host name | Port | Protocol
-| subscription.rhsm.redhat.com | 443 | HTTPS
-| subscription.rhn.redhat.com | 443 | HTTPS
-| cdn.redhat.com | 443 | HTTPS
-| *.akamaiedge.net |  443 | HTTPS
-ifdef::satellite[]
-| api.access.redhat.com (if using Red{nbsp}Hat Insights) | 443 | HTTPS
-| cert-api.access.redhat.com (if using Red{nbsp}Hat Insights) | 443 | HTTPS
-endif::[]
-|====
-+
 . If your HTTP proxy requires authentication, enter a *Username* and *Password*.
 . Optional: In the *Test URL* field, enter the HTTP proxy URL, then click *Test Connection* to ensure that you can connect to the HTTP proxy from {Project}.
 . Click the *Locations* tab and add a location.
--- a/guides/common/modules/proc_adding-an-http-proxy.adoc
+++ b/guides/common/modules/proc_adding-an-http-proxy.adoc
@@ -20,6 +20,7 @@ The following host names are available:
 | subscription.rhsm.redhat.com | 443 | HTTPS
 | subscription.rhn.redhat.com | 443 | HTTPS
 | cdn.redhat.com | 443 | HTTPS
+| *.akamaiedge.net |  443 | HTTPS
 ifdef::satellite[]
 | api.access.redhat.com (if using Red{nbsp}Hat Insights) | 443 | HTTPS
 | cert-api.access.redhat.com (if using Red{nbsp}Hat Insights) | 443 | HTTPS
--- a/guides/common/modules/proc_configuring-http-proxy-to-connect-to-cdn.adoc
+++ b/guides/common/modules/proc_configuring-http-proxy-to-connect-to-cdn.adoc
@@ -1,29 +1,19 @@
 [id="configuring-http-proxy-to-connect-to-cdn_{context}"]
 = Configuring the HTTP Proxy to Connect to Red Hat CDN
 
-To configure subscription-manager with an HTTP proxy, follow the procedure below.
+.Prerequisites
+Your network gateway and the HTTP proxy must allow access to the following hosts:
+include::snip_host-names-for-http-proxy.adoc[]
 
-.Procedure
-
-. On the network gateway and the HTTP Proxy, enable TCP for the following host names:
-+
-[cols="2,1,1",options="header"]
-|====
-| Host name | Port | Protocol
-| subscription.rhsm.redhat.com | 443 | HTTPS
-| cdn.redhat.com |  443 | HTTPS
-| *.akamaiedge.net |  443 | HTTPS
-| cert.console.redhat.com (if using Red{nbsp}Hat Insights) | 443 | HTTPS
-| cert-api.access.redhat.com (if using Red{nbsp}Hat Insights) |  443 | HTTPS
-| api.access.redhat.com (if using Red{nbsp}Hat Insights) |  443 | HTTPS
-|====
-+
 {ProjectServer} uses SSL to communicate with the Red{nbsp}Hat CDN securely.
 Use of an SSL interception proxy interferes with this communication.
 These hosts must be whitelisted on the proxy.
-+
+
 For a list of IP addresses used by the Red{nbsp}Hat CDN (cdn.redhat.com), see the Knowledgebase article https://access.redhat.com/articles/1525183[Public CIDR Lists for Red{nbsp}Hat] on the Red{nbsp}Hat Customer Portal.
-+
+
+To configure the subscription-manager with the HTTP proxy, follow the procedure below.
+
+.Procedure
 . On {ProjectServer}, complete the following details in the `/etc/rhsm/rhsm.conf` file:
 +
 [options="nowrap" subs="+quotes"]
--- a/guides/common/modules/snip_host-names-for-http-proxy.adoc
+++ b/guides/common/modules/snip_host-names-for-http-proxy.adoc
@@ -1,0 +1,14 @@
+[id="host-names-for-http-proxy_{context}"]
+
+[cols="2,1,1",options="header"]
+|====
+| Host name | Port | Protocol
+| subscription.rhsm.redhat.com | 443 | HTTPS
+| cdn.redhat.com | 443 | HTTPS
+| *.akamaiedge.net | 443 | HTTPS
+ifdef::satellite[]
+| cert.console.redhat.com (if using Red{nbsp}Hat Insights) | 443 | HTTPS
+| api.access.redhat.com (if using Red{nbsp}Hat Insights) | 443 | HTTPS
+| cert-api.access.redhat.com (if using Red{nbsp}Hat Insights) | 443 | HTTPS
+endif::[]
+|====


adamlazik1:
Is this confirmend also for 3.1/6.11?

mjivraja:
> Is this confirmend also for 3.1/6.11?

Not confirmed but logically it should IMHO.

adamlazik1:
In that case I have to leave the review to someone who understands the technical side of this.

guides/common/modules/proc_adding-an-http-proxy.adoc
```diff
@@ -20,6 +20,7 @@ The following host names are available:
 | subscription.rhsm.redhat.com | 443 | HTTPS
 | subscription.rhn.redhat.com | 443 | HTTPS
 | cdn.redhat.com | 443 | HTTPS
+| *.akamaiedge.net |  443 | HTTPS
```
maximiliankolb, line 23:
```suggestion
| *.akamaiedge.net | 443 | HTTPS
```

maximiliankolb (commented):
Change looks reasonable & it resolves the BZ you've linked. I haven't seen any other `*.` in the list of domains & haven't tested this myself. Maybe someone else can ACK this PR.

mjivraja:
> Change looks reasonable & it resolves the BZ you've linked. I haven't seen any other `*.` in the list of domains & haven't tested this myself. Maybe someone else can ACK this PR.

Thanks @maximiliankolb

evgeni (changes requested):
While I agree that the list here and in `guides/common/modules/proc_configuring-http-proxy-to-connect-to-cdn.adoc` should match, I disagree with the actual change and the placement of this table overall.

First of all, to avoid similar bugs in the future, the table should be made into a snippet and included in both modules, instead of keeping two identical copies.

But second, and IMHO more important, the current placement of the table doesn't make any sense (it is not fault of this PR! I just think it's the perfect time to go and fix it).
image
What host names is this table refer to? Certainly not the ones for the URL field, as that is the URL of the proxy.

If we look at the original bug (https://bugzilla.redhat.com/show_bug.cgi?id=1921928) and commit (10f125daafc618898d3e8d2cddd95cce6a96d009), I think the intent was to list all hosts that we need to be able to communicate with, which should place the table somewhere in the introduction (or in a currently non-existing "prerequisites" section).

Malhar, do you prefer to keep the scope of this PR narrow and just do the snippet conversion here, or should we also try to tackle the placement? I'm fine with either, we just really have to do both ASAP :)

mjivraja:
> While I agree that the list here and in `guides/common/modules/proc_configuring-http-proxy-to-connect-to-cdn.adoc` should match, I disagree with the actual change and the placement of this table overall.
> 
> First of all, to avoid similar bugs in the future, the table should be made into a snippet and included in both modules, instead of keeping two identical copies.
> 
> But second, and IMHO more important, the current placement of the table doesn't make any sense (it is not fault of this PR! I just think it's the perfect time to go and fix it). image What host names is this table refer to? Certainly not the ones for the URL field, as that is the URL of the proxy.
> 
> If we look at the original bug (https://bugzilla.redhat.com/show_bug.cgi?id=1921928) and commit ([10f125d](https://github.com/theforeman/foreman-documentation/commit/10f125daafc618898d3e8d2cddd95cce6a96d009)), I think the intent was to list all hosts that we need to be able to communicate with, which should place the table somewhere in the introduction (or in a currently non-existing "prerequisites" section).
> 
> Malhar, do you prefer to keep the scope of this PR narrow and just do the snippet conversion here, or should we also try to tackle the placement? I'm fine with either, we just really have to do both ASAP :)

In my opinion, if the placement is totally incorrect, we can fix it. I can make the `snip_*****.adoc` for this table and place it wherever it fits more accurately.

evgeni:
Yeah, I think it's totally off.

mjivraja:
> proc_configuring-http-proxy-to-connect-to-cdn.adoc

I have created `snip_host-names-for-http-proxy.adoc` that contains the above table. What can be its placement now?

evgeni:
Tough question. The other place (install guide: https://docs.theforeman.org/nightly/Installing_Server/index-katello.html#configuring-http-proxy-to-connect-to-cdn_foreman) has it as a separate step in the procedure in the form of "configure your proxy to allow that", followed by the "now configure foreman to use the proxy".

I *personally* don't like that, as it tells the user to do something, without giving clear instructions (which we even can't, as we don't know what proxy that is, etc), so I would probably do a "prerequisites" section saying "your proxy needs to allow access to the following hosts" `<table>`

mjivraja:
> I _personally_ don't like that, as it tells the user to do something, without giving clear instructions (which we even can't, as we don't know what proxy that is, etc), so I would probably do a "prerequisites" section saying "your proxy needs to allow access to the following hosts" `<table>`

And does this apply to `proc_adding-an-http-proxy.adoc` as well?

evgeni:
Are you asking whether you should rewrite the other place too? I wouldn't in this PR, not too explode the scope too much.
Just make it use the new snippet.

mjivraja:
> Are you asking whether you should rewrite the other place too? I wouldn't in this PR, not too explode the scope too much. Just make it use the new snippet.

I mean, I would like to include the `snip_***.adoc` file to `proc_adding-an-http-proxy.adoc` alongside `configuring-http-proxy-to-connect-to-cdn.adoc` (that you mentioned). Because both have the same table for host names.

evgeni:
Ah, yes, I'd replace the table in `proc_adding-an-http-proxy.adoc` with an include, but that's it.

guides/common/modules/proc_configuring-http-proxy-to-connect-to-cdn.adoc
```diff
@@ -1,29 +1,21 @@
...
+.Prerequisites
+
+Your network gateway and the HTTP proxy must allow access to the following hosts:
+include::snip_host-names-for-http-proxy.adoc[]
+
+To configure the subscription-manager with the HTTP proxy, follow the procedure below.
 
 .Procedure
 
-. On the network gateway and the HTTP Proxy, enable TCP for the following host names:
```
mjivraja, line 8: I have merged line `8` (old) with the line `6` (new).  Need the technical validation.

guides/common/modules/snip_host-names-for-http-proxy.adoc
```diff
@@ -0,0 +1,14 @@
...
+
+[cols="2,1,1",options="header"]
+|====
+| Host name | Port | Protocol
+| subscription.rhsm.redhat.com | 443 | HTTPS
+| subscription.rhn.redhat.com | 443 | HTTPS
+| cdn.redhat.com | 443 | HTTPS
+| *.akamaiedge.net | 443 | HTTPS
+ifdef::satellite[]
+| api.access.redhat.com (if using Red{nbsp}Hat Insights) | 443 | HTTPS
```
evgeni, line 11: I think you accidentally lost the `cert.console.redhat.com` line

guides/common/modules/snip_host-names-for-http-proxy.adoc
```diff
@@ -0,0 +1,14 @@
+[id="host-names-for-http-proxy_{context}"]
+
+[cols="2,1,1",options="header"]
+|====
+| Host name | Port | Protocol
+| subscription.rhsm.redhat.com | 443 | HTTPS
+| subscription.rhn.redhat.com | 443 | HTTPS
```
evgeni, line 7: I think noone uses the `rhn` name anymore (and it wasn't present in one of the tables anyway), so let's drop it.

guides/common/modules/proc_configuring-http-proxy-to-connect-to-cdn.adoc
```diff
@@ -1,29 +1,21 @@
...
-| cert-api.access.redhat.com (if using Red{nbsp}Hat Insights) |  443 | HTTPS
-| api.access.redhat.com (if using Red{nbsp}Hat Insights) |  443 | HTTPS
-|====
-+
 {ProjectServer} uses SSL to communicate with the Red{nbsp}Hat CDN securely.
 Use of an SSL interception proxy interferes with this communication.
 These hosts must be whitelisted on the proxy.
-+
+
 For a list of IP addresses used by the Red{nbsp}Hat CDN (cdn.redhat.com), see the Knowledgebase article https://access.redhat.com/articles/1525183[Public CIDR Lists for Red{nbsp}Hat] on the Red{nbsp}Hat Customer Portal.
```
evgeni, line 17: If we're already moving things around, I'd move this text together with the table to the Prerequisites.

guides/common/modules/proc_adding-an-http-proxy.adoc
```diff
@@ -14,17 +14,7 @@ To use the CLI instead of the {ProjectWebUI}, see the xref:cli-adding-an-http-pr
...
-| Host name | Port | Protocol
-| subscription.rhsm.redhat.com | 443 | HTTPS
-| subscription.rhn.redhat.com | 443 | HTTPS
-| cdn.redhat.com | 443 | HTTPS
-ifdef::satellite[]
-| api.access.redhat.com (if using Red{nbsp}Hat Insights) | 443 | HTTPS
-| cert-api.access.redhat.com (if using Red{nbsp}Hat Insights) | 443 | HTTPS
-endif::[]
-|====
+include::snip_host-names-for-http-proxy.adoc[]
```
evgeni, line 17: The table placement is still wrong in this file.

mjivraja: From the procedure, it still looks like the table is placed properly.
Also, I speculated there is no change in placement from your last comment. :sweat_smile: 

Can you guide me to the desired placement?

evgeni: Hah. I think I managed to confuse myself and then you. Sorry! Let me see if I can get the things straight.

The table of "hosts the proxy needs to allow" currently exists in two files:
* `guides/common/modules/proc_adding-an-http-proxy.adoc`
* `guides/common/modules/proc_configuring-http-proxy-to-connect-to-cdn.adoc`

The original issue was: "`proc_adding-an-http-proxy.adoc` is missing the `akamaiedge.net` entry" and while reviewing your change to that, I noticed the wrong placement and asked you to move the table to a snippet and include that snippet in both files, fixing the placement.

And then, you asked whether you should add it to both files, I accidentally said "I'd replace the table in `proc_adding-an-http-proxy.adoc` with an include, but that's it." which was wrong, I meant the "that's it part" to `proc_configuring-http-proxy-to-connect-to-cdn.adoc` as that's the file that we were originally not touching in this PR.

So in regard with `proc_adding-an-http-proxy.adoc` I'd add a "Prerequisites" section, reading something like this:
```
Your HTTP proxy must allow access to the following hosts:
include::snip_host-names-for-http-proxy.adoc[]
```

mjivraja: Thank you for the detailed clarification @evgeni 
It's perfectly okay for the confusion (usually you are always clear). 😀

I hope, we inched closer now.

guides/common/modules/snip_host-names-for-http-proxy.adoc
mjivraja: Apologies for this. I assumed there is only `ifdef` difference in the table of both files. But, a few host differences were also there. Addressed :+1:

guides/common/modules/proc_configuring-http-proxy-to-connect-to-cdn.adoc
```diff
@@ -1,29 +1,21 @@
...
-| cdn.redhat.com |  443 | HTTPS
-| *.akamaiedge.net |  443 | HTTPS
-| cert.console.redhat.com (if using Red{nbsp}Hat Insights) | 443 | HTTPS
-| cert-api.access.redhat.com (if using Red{nbsp}Hat Insights) |  443 | HTTPS
-| api.access.redhat.com (if using Red{nbsp}Hat Insights) |  443 | HTTPS
-|====
-+
 {ProjectServer} uses SSL to communicate with the Red{nbsp}Hat CDN securely.
 Use of an SSL interception proxy interferes with this communication.
 These hosts must be whitelisted on the proxy.
```
evgeni, line 17: this paragraph also belongs to the prerequisites

evgeni:
huh, wiki.debian.org is indeed down right now, which breaks the link checker

guides/common/modules/proc_configuring-http-proxy-to-connect-to-cdn.adoc
mjivraja: Addressed :+1:

evgeni (approved):
needs squashing, but otherwise LGTM!

mjivraja:
> needs squashing, but otherwise LGTM!

Thanks and addressed :+1:

evgeni:
   7f6e9b2ab..a47f4b15e  3.1 -> 3.1
   1218f82c1..5b4ad2b79  3.3 -> 3.3
   9e325d4f2..49bcef728  3.4 -> 3.4
   2b8286902..84ca1be79  3.5 -> 3.5

3.2 did not pick cleanly, but given the release is EOL upstream and there are no downstream releases based on it, I don't care too much and would leave it as is.

PR #1905, opened 2023-01-02
Mention supported Ansible versions

Lennonka:
We wanted to mention supported Ansible versions, however, they actually depend on the user's configuration of the base OS. Perhaps it's still worth mentioning.

https://bugzilla.redhat.com/show_bug.cgi?id=2139700

* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5 (Satellite 6.12)
* [x] Foreman 3.2/Katello 4.4
* [x] Foreman 3.1/Katello 4.3 (Satellite 6.11, orcharhino 6.1+)

### DIFF
--- a/guides/common/modules/con_supported-ansible-versions.adoc
+++ b/guides/common/modules/con_supported-ansible-versions.adoc
@@ -1,0 +1,5 @@
+[id="supported-ansible-versions_{context}"]
+= Supported Ansible Versions
+
+{Project} uses Ansible as provided by the base operating system of {ProjectServer} or any {SmartProxies} for remote execution.
+Therefore, the supported version of Ansible depends on your base OS configuration.
--- a/guides/common/assembly_getting-started-with-ansible.adoc
+++ b/guides/common/assembly_getting-started-with-ansible.adoc
@@ -6,6 +6,7 @@ ifdef::context[:parent-context: {context}]
 
 Use this guide to configure {Project} to use Ansible, and then information about how to use Ansible for remote execution.
 
+include::modules/con_supported-ansible-versions.adoc[leveloffset=+1]
 include::modules/proc_configuring-your-deployment-to-run-ansible-roles.adoc[leveloffset=+1]
 include::modules/proc_importing-ansible-roles.adoc[leveloffset=+1]
 include::modules/proc_overriding-ansible-variables.adoc[leveloffset=+1]


ekohl (commented):
Makes sense to me to at least mention that "it depends". Arguably even more in upstream where Debian, Ubuntu and EL could ship very different versions of Ansible.

guides/common/modules/con_supported-ansible-versions.adoc
```diff
@@ -0,0 +1,5 @@
+[id="supported-ansible-versions_{context}"]
+= Supported Ansible Versions
+
+{Project} uses Ansible as provided by the base operating system of {ProjectServer} and any external {SmartProxies} for remote execution.
```
ekohl, line 4: I'm debating the "external" bit here, so I'll write down my thoughts.
* For vanilla Foreman (foreman-deb, foreman-el) there is no concept of "internal" or "external" Smart Proxy. You just have hosts which run a Smart Proxy and they're all the same. This maps poorly to this repo which is still very Katello-centric
* For Katello people do talk about an "internal" Smart Proxy, mostly because it's the single Smart Proxy where Pulp runs in non-mirror mode while all other Content Proxies run Pulp in mirror mode.
* For Satellite this resolves in "external Capsules" but I'd argue that a Capsule is always external

I'm not sure what's best here. One possibility that avoids it is:
```suggestion
{Project} uses Ansible as provided by the base operating system of {ProjectServer} or any {SmartProxies} for remote execution.
```

ekohl (commented):
I should have written it down in my previous review.

guides/common/modules/con_supported-ansible-versions.adoc
Lennonka: Thanks, I wasn't sure if it was generally appropriate.

ekohl (approved):
@evgeni mind double checking my suggestions?

evgeni (approved):

Lennonka:
Cherry-picked:
- bad888402..8e0795433  3.5 -> 3.5
- 2b349bd93..888f34f4e  3.4 -> 3.4
- 93af0231e..1a4b62842  3.3 -> 3.3
- 80e2260ed..8196fb8c7  3.2 -> 3.2
- 6228d1d0c..80bae277b  3.1 -> 3.1

PR #1825, opened 2022-11-17
Introduce attributes for PostgreSQL paths & update storage tables

ekohl:
See individual commits for details. This should be picked back to 3.3.

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.5/Katello 4.7
* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3
* For Foreman 3.0 or older, please create a separate PR.
* We do not accept PRs for Foreman 2.3 or older.

PR #1914, opened 2023-01-04
Cherry pick PostgreSQL attributes PR to 3.3

ekohl:
3.3 had conflicts so a manual pick of https://github.com/theforeman/foreman-documentation/pull/1825 was needed.

### DIFF
--- a/guides/common/attributes-base.adoc
+++ b/guides/common/attributes-base.adoc
@@ -73,6 +73,10 @@
 :package-remove: yum remove
 :package-update-project: yum update
 :package-update: yum update
+:postgresql-lib-dir: /var/lib/pgsql
+:postgresql-data-dir: {postgresql-lib-dir}/data
+:postgresql-conf-dir: {postgresql-data-dir}
+:postgresql-log-dir: {postgresql-data-dir}/log
 :PIV: PIV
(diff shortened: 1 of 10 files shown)


maximiliankolb (approved):

guides/common/attributes-base.adoc
```diff
@@ -73,6 +73,10 @@
 :package-remove: yum remove
 :package-update-project: yum update
 :package-update: yum update
+:postgresql-lib-dir: /var/lib/pgsql
+:postgresql-data-dir: {postgresql-lib-dir}/data
+:postgresql-conf-dir: {postgresql-data-dir}
+:postgresql-log-dir: {postgresql-data-dir}/log
```
ekohl, line 79: These are the EL8 values. For Satellite 6.12 this should be safe since it only supports RHEL 8, which technically means it's also a bugfix in various places.

For upstream it's a bit unclear since both EL7 & EL8 should be supported, but that's now EOL anyway so I'm not too worried.

For orcharhino I don't know what it means.

maximiliankolb: Since there is no orcharhino using Foreman 3.3 yet, I think this is fine! I will create a follow-up PR in case orcharhino differs.

PR #1920, opened 2023-01-06
correct module-enable-note for capsules and non-satellite builds

evgeni:

* [x] Foreman 3.4/Katello 4.6
* [x] Foreman 3.3/Katello 4.5 (Satellite 6.12)
* [ ] Foreman 3.2/Katello 4.4
* [ ] Foreman 3.1/Katello 4.3 (Satellite 6.11, orcharhino 6.1+)

### DIFF
--- a/guides/common/attributes-foreman-el.adoc
+++ b/guides/common/attributes-foreman-el.adoc
@@ -1,2 +1,3 @@
 // Overrides for foreman-el build
-:dnf-modules: foreman:el8
+:dnf-module: foreman:el8
+:dnf-modules: {dnf-module}
--- a/guides/common/attributes-katello.adoc
+++ b/guides/common/attributes-katello.adoc
@@ -4,4 +4,5 @@
 :installer-scenario: foreman-installer --scenario katello
 :installer-log-file: /var/log/foreman-installer/katello.log
 :smartproxy_port: 9090
-:dnf-modules: katello:el8 pulpcore:el8
+:dnf-module: katello:el8
+:dnf-modules: {dnf-module} pulpcore:el8
--- a/guides/common/modules/proc_configuring-repositories-proxy.adoc
+++ b/guides/common/modules/proc_configuring-repositories-proxy.adoc
@@ -1,6 +1,7 @@
 [id="configuring-repositories-proxy_{context}"]
 
 = Configuring Repositories
+:dnf-module: satellite-capsule:el8
 
 Use this procedure to enable the repositories that are required to install {ProductName}.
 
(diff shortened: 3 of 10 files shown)


guides/common/modules/snip_module-enable-note.adoc
```diff
@@ -0,0 +1,6 @@
+Enablement of the module `{dnf-module}` warns about a conflict with `postgresql:10` and `ruby:2.5` as these modules are set to the default module versions on {EL} 8.
```
ekohl, line 1: Should this file be named `dnf-module-enable`? Module is used so often that I think it should be qualified.

guides/common/modules/proc_configuring-repositories-proxy.adoc
```diff
@@ -31,7 +31,8 @@ Use this procedure to enable the repositories that are required to install {Prod
 +
 [NOTE]
 ====
-include::snip_satellite-el8-enable-note.adoc[]
+:dnf-module: satellite-capsule:el8
```
ekohl, line 34: Perhaps this should be set a bit higher and also used on line 29 so you know they're connected?

ekohl (commented):
I'm sure you also debated this, but the `dnf-module` vs `dnf-modules` is a bit vague. Yet I don't know what a better solution would be.

guides/common/modules/snip_module-enable-note.adoc
evgeni: yes

evgeni:
> I'm sure you also debated this, but the `dnf-module` vs `dnf-modules` is a bit vague. Yet I don't know what a better solution would be.

Yeah, I didn't see a good way out of this

ekohl (approved):
Perhaps @maximiliankolb has a good idea here, but I think this is overall an improvement.

guides/common/attributes-foreman-el.adoc
```diff
@@ -1,2 +1,3 @@
 // Overrides for foreman-el build
-:dnf-modules: foreman:el8
+:dnf-module: foreman:el8
```
maximiliankolb, line 2: I thought it'd be possible to use `{project-context}` here, but it's missing for katello: https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-katello.adoc

I am unsure if this is an issue to begin with.

guides/common/attributes-katello.adoc
```diff
@@ -4,4 +4,5 @@
 :installer-scenario: foreman-installer --scenario katello
 :installer-log-file: /var/log/foreman-installer/katello.log
 :smartproxy_port: 9090
-:dnf-modules: katello:el8 pulpcore:el8
+:dnf-module: katello:el8
```
maximiliankolb, line 7: Naming suggestion: `main-dnf-module`? I am unsure if this makes things easier. Downside: sort in alphabetical order and the attributes are no longer side by side.

IMHO it's fine as is.

maximiliankolb (approved):
One small comment/question but it's outside of your PR.

LGTM!

evgeni:
Thanks, I'll keep it as is.

PR #1943, opened 2023-01-20
Simplify the firewall command for ports

mjivraja:
While enabling connections from a Client to Project Server, we open the ports by mentioning all the required ports in the command. However, it can be further simplified by adding just a single command that includes all those ports and practically does the same task as an older one. Therefore, we replaced it with a simple command. Although, we can keep both or either.

https://bugzilla.redhat.com/show_bug.cgi?id=2161296

* [X] Foreman 3.1/Katello 4.3 on EL7 & EL8 (Satellite 6.11 EL7/8, orcharhino 6.1 on EL7, orcharhino 6.2 on EL7/8)

### DIFF
--- a/guides/common/attributes-base.adoc
+++ b/guides/common/attributes-base.adoc
@@ -47,6 +47,8 @@
 :customrpmtitle: RPM
 :customssl: SSL
 :customssltitle: SSL
+:firewalld-profile: RH-Satellite-6
+:firewalld-profile-proxy: foreman-proxy
 :foreman-example-com: foreman.example.com
 :foreman-installer: foreman-installer
 :foreman-installer-package: foreman-installer
--- a/guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
+++ b/guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
@@ -11,21 +11,9 @@ include::snip_firewalld.adoc[]
 .Procedure
 . To open the ports for client to {Project} communication, enter the following command on the base operating system that you want to install {Project} on:
 +
-[options="nowrap"]
+[options="nowrap", subs="+quotes,attributes"]
 ----
-# firewall-cmd \
---add-port="53/udp" --add-port="53/tcp" \
---add-port="67/udp" \
---add-port="69/udp" \
---add-port="80/tcp" --add-port="443/tcp" \
-ifdef::katello,satellite,orcharhino[]
---add-port="5647/tcp" \
---add-port="8000/tcp" --add-port="9090/tcp" \
-endif::[]
-ifndef::katello,satellite,orcharhino[]
---add-port="8443/tcp" \
-endif::[]
---add-port="8140/tcp"
+# firewall-cmd --add-service={firewalld-profile}
 ----
 . Make the changes persistent:
 +
--- a/guides/common/attributes-katello.adoc
+++ b/guides/common/attributes-katello.adoc
@@ -6,3 +6,4 @@
 :smartproxy_port: 9090
 :dnf-module: katello:el8
 :dnf-modules: {dnf-module} pulpcore:el8
+:firewalld-profile-proxy: RH-Satellite-6-capsule
--- a/guides/common/attributes-orcharhino.adoc
+++ b/guides/common/attributes-orcharhino.adoc
@@ -36,6 +36,8 @@
 :customssl: custom SSL
 :customssltitle: Custom SSL
 :foreman-example-com: orcharhino.example.com
+:firewalld-profile: orcharhino
+:firewalld-profile-proxy: orcharhino-proxy
 :installer-scenario-smartproxy: orcharhino-installer --no-enable-foreman
 :installer-scenario: foreman-installer --scenario katello
 :installer-log-file: /var/log/foreman-installer/katello.log
--- a/guides/common/attributes-satellite.adoc
+++ b/guides/common/attributes-satellite.adoc
@@ -54,6 +54,7 @@
 :customssltitle: Custom SSL
 :DocState: satellite
 :EL: {RHEL}
+:firewalld-profile-proxy: RH-Satellite-6-capsule
 :foreman-example-com: satellite.example.com
 :foreman-installer-package: satellite-installer
 :foreman-installer: satellite-installer
--- a/guides/common/modules/proc_enabling-connections-from-capsule-to-satellite.adoc
+++ b/guides/common/modules/proc_enabling-connections-from-capsule-to-satellite.adoc
@@ -8,16 +8,3 @@ On {ProjectServer}, you must enable the incoming connection from {SmartProxyServ
 For more information, see {InstallingServerDocURL}Enabling_Connections_from_a_Client_to_Server_{project-context}[Enabling Connections from a Client to {ProjectServer}] in _{InstallingServerDocTitle}_.
 
 include::snip_firewalld.adoc[]
-
-.Procedure
-. On {ProjectServer}, enter the following command to open the port for {SmartProxy} to {Project} communication:
-+
-[options="nowrap"]
-----
-# firewall-cmd --add-port="5646/tcp"
-----
-. Make the changes persistent:
-+
-----
-# firewall-cmd --runtime-to-permanent
-----
--- a/guides/common/modules/proc_enabling-connections-to-capsule.adoc
+++ b/guides/common/modules/proc_enabling-connections-to-capsule.adoc
@@ -10,23 +10,10 @@ include::snip_firewalld.adoc[]
 
 . On the base operating system on which you want to install {SmartProxy}, enter the following command to open the ports for {ProjectServer} and clients communication to {SmartProxyServer}:
 +
-[options="nowrap"]
+[options="nowrap", subs="+quotes,attributes"]
 ----
-# firewall-cmd \
---add-port="53/udp" --add-port="53/tcp" \
---add-port="67/udp" \
---add-port="69/udp" \
-ifdef::katello,satellite,orcharhino[]
---add-port="80/tcp" --add-port="443/tcp" \
---add-port="5647/tcp" \
-endif::[]
---add-port="8140/tcp" \
---add-port="8443/tcp" \
-ifdef::katello,satellite,orcharhino[]
---add-port="8000/tcp" --add-port="9090/tcp"
-endif::[]
+# firewall-cmd --add-service={firewalld-profile-proxy}
 ----
-
 . Make the changes persistent:
 +
 ----


guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
```diff
@@ -13,19 +13,7 @@ include::snip_firewalld.adoc[]
...
---add-port="80/tcp" --add-port="443/tcp" \
-ifdef::katello,satellite,orcharhino[]
---add-port="5647/tcp" \
---add-port="8000/tcp" --add-port="9090/tcp" \
-endif::[]
-ifndef::katello,satellite,orcharhino[]
---add-port="8443/tcp" \
-endif::[]
---add-port="8140/tcp"
+# firewall-cmd --add-server=RH-Satellite-6
```
maximiliankolb, line 16: This is Satellite-only. For orcharhino, it's "orcharhino".

maximiliankolb (commented):
I am unsure about Foreman.

guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
mjivraja: Although the new command is just a replacement of the old one and does the same thing, I think we should keep the old commands unmodified (and close this PR) or keep the new one Satellite specific. 

What's your take?

maximiliankolb: I have not tested it completely, but your change in general looks like a great simplification/improvement. However, we should show different commands for different build targets.

For built target *satellite*, we should the command as above; for *orcharhino*, the service is called `orcharhino`; for Foreman/Katello, I don't know.

ekohl: There is no upstream version, but since it's part of firewalld itself users can use `RH-Satellite-6`. I'd recommend introducing some attribute for the firewalld profile name.

maximiliankolb: Sounds great!

mjivraja: > There is no upstream version, but since it's part of firewalld itself users can use `RH-Satellite-6`. I'd recommend introducing some attribute for the firewalld profile name.

So I should consider only `Orcharhino` (along with Satellite) at this level, right?

maximiliankolb: You can set the attribute for all build targets to `RH-Satellite-6` and only overwrite it for orcharhino. Example: https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-base.adoc#L36 & https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-satellite.adoc#L43

mjivraja: Figuring out why attribute value is not getting rendered.

mjivraja: Addressed.

maximiliankolb (approved):
LGTM!

maximiliankolb:
Does anyone else want to ack this? I am unsure who can answer this because it came from DDF, not a developer.

I'm good with this chance; any more comments?

ekohl (commented):
I think we should review the implications. For example, are things we add in other paragraphs now redundant?

One thing to consider is that looking in `/usr/lib/firewalld/services` (on my Fedora machine) I see both `RH-Satellite-6.xml` and `RH-Satellite-6-capsule.xml`. Then `RH-Satellite-6.xml` includes `foreman` as a service, so I stand corrected in that there's no upstream version. It means we should use that as a base attribute while setting `RH-Satellite-6` in Katello and Satellite.

Then the Satellite profile also includes port 5646 so `guides/common/modules/proc_enabling-connections-from-capsule-to-satellite.adoc` has a redundant instruction.

The same may be true for `guides/common/modules/proc_provisioning-ubuntu-autoinstall-through-smart-proxies.adoc` since port 8000 is part of `RH-Satellite-6`, though it's not for `foreman-proxy` so in upstream it may still be valid.

In `guides/common/modules/proc_enabling-connections-to-capsule.adoc` we see a similar listing, which could then be replaced by using `foreman-proxy` (base), `RH-Satellite-6-capsule` (katello, satellite) and something on Orcharhino.

Another thing is the commit message. I would be descriptive. "Simplify" is always tempting, but if you say "use RH-Satellite-6 firewalld service` it becomes easier to see in the log.

mjivraja:
> I think we should review the implications. For example, are things we add in other paragraphs now redundant?
> 
> Then the Satellite profile also includes port 5646 so `guides/common/modules/proc_enabling-connections-from-capsule-to-satellite.adoc` has a redundant instruction.

Noted. I can remove it from above mentioned file.
> 
> The same may be true for `guides/common/modules/proc_provisioning-ubuntu-autoinstall-through-smart-proxies.adoc` since port 8000 is part of `RH-Satellite-6`, though it's not for `foreman-proxy` so in upstream it may still be valid.

Will not modify this file.
> 
> In `guides/common/modules/proc_enabling-connections-to-capsule.adoc` we see a similar listing, which could then be replaced by using `foreman-proxy` (base), `RH-Satellite-6-capsule` (katello, satellite) and something on Orcharhino.

I will add another attribute for both of the above as `firewalld-profile-proxy`. @maximiliankolb do you have the same for Orcharhino?
> 
> Another thing is the commit message. I would be descriptive. "Simplify" is always tempting, but if you say "use RH-Satellite-6 firewalld service` it becomes easier to see in the log.

Noted. Will amend/squash it in the next push. :+1:

maximiliankolb:
Adding a second attribute works for me; please call the file "orcharhino-proxy.xml".

ekohl:
@mjivraja I think it may also be fine to work on the proxy/capsule side in a follow up PR if you want to keep the scope smaller.

mjivraja:
> @mjivraja I think it may also be fine to work on the proxy/capsule side in a follow up PR if you want to keep the scope smaller.

I think we extend it here itself. The reason is - this PR is not a bug, it's an improvement. So, there is no blocker from end-users.

guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
```diff
@@ -11,21 +11,9 @@ include::snip_firewalld.adoc[]
...
---add-port="80/tcp" --add-port="443/tcp" \
-ifdef::katello,satellite,orcharhino[]
---add-port="5647/tcp" \
---add-port="8000/tcp" --add-port="9090/tcp" \
-endif::[]
-ifndef::katello,satellite,orcharhino[]
---add-port="8443/tcp" \
-endif::[]
---add-port="8140/tcp"
+# firewall-cmd --add-server={firewalld-profile}
```
evgeni, line 16:
```suggestion
# firewall-cmd --add-service={firewalld-profile}
```

guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
```diff
@@ -11,21 +11,9 @@ include::snip_firewalld.adoc[]
 .Procedure
 . To open the ports for client to {Project} communication, enter the following command on the base operating system that you want to install {Project} on:
 +
-[options="nowrap"]
+[options="nowrap", subs="+quotes,attributes"]
 ----
-# firewall-cmd \
---add-port="53/udp" --add-port="53/tcp" \
```
evgeni, line 17: this is not part of `RH-Satellite-6`

guides/common/attributes-base.adoc
```diff
@@ -47,6 +47,7 @@
 :customrpmtitle: RPM
 :customssl: SSL
 :customssltitle: SSL
+:firewalld-profile: RH-Satellite-6
```
evgeni, line 50: there is a `foreman` profile that can be used for non-Katello/non-Satellite installs:

```
  <include service="dns"/>
  <include service="http"/>
  <include service="https"/>
  <include service="dhcp"/>
  <include service="tftp"/>
  <port protocol="udp" port="68"/>
  <port protocol="tcp" port="8140"/>
```

ekohl: Yesterday I looked at this too, but didn't want to dive too deep into it (rabbit holes ahead).

> <port protocol="udp" port="68"/>

I really do wonder why this was specified. It includes the dhcp service and that's why we need to open it, right?

And it would be nice if there was a firewalld service for Puppet instead of listing the port.

evgeni: Mysteries, all over the place.

https://github.com/firewalld/firewalld/pull/645 points at https://bugzilla.redhat.com/show_bug.cgi?id=1839781 which never was touched by anyone from the Foreman/Satellite team :(

ehelms:
The use of this profile will also open *all* potential ports which I also think is not what we want users to do? For example, the qpid ports are only necessary when using the katello-agent feature, which is disabled by default. We should not encourage users to enable a firewall profile that leaves unused ports open. As nice as the profiles are, given they rely on an external package to control them, I lean towards shying away from it's use and just being explicit about what ports to open, and when documenting an optional feature that can be enabled also noting the needed ports to open there.

ekohl:
I agree that the current profile isn't quite right. I really wonder how it was added to firewalld and if any Satellite engineer was involved in the process.

@ehelms should we reach out to the firewalld team to introduce corrected profiles?

Something we can do is consistently use use services. Today we list ports (like `1883/tcp`) but there is a MQTT service. One thing I noticed is that the tftp service loads the TFTP helper, which is the `ip_conntrack_tftp` kernel module. Perhaps that's no longer needed with nftables, but pushing that responsibility to firewalld is a good thing.

I've opened https://github.com/firewalld/firewalld/pull/1093 to start the discussion upstream & https://github.com/theforeman/foreman-documentation/pull/2051 here in this repo.

With that I think we can close this PR now.

mjivraja:
> I agree that the current profile isn't quite right. I really wonder how it was added to firewalld and if any Satellite engineer was involved in the process.
> 
> @ehelms should we reach out to the firewalld team to introduce corrected profiles?
> 
> Something we can do is consistently use use services. Today we list ports (like `1883/tcp`) but there is a MQTT service. One thing I noticed is that the tftp service loads the TFTP helper, which is the `ip_conntrack_tftp` kernel module. Perhaps that's no longer needed with nftables, but pushing that responsibility to firewalld is a good thing.
> 
> I've opened [firewalld/firewalld#1093](https://github.com/firewalld/firewalld/pull/1093) to start the discussion upstream & #2051 here in this repo.
> 
> With that I think we can close this PR now.

Ack! Ewoud :+1:

PR #1974, opened 2023-02-07
Add redis in the list of users to avoid conflicts

mjivraja:
We list some user accounts that are created by satellite that should not already exist or fetched from external entities. So here, the Redis was missing. It's applicable to even the Proxy Installation part. The same can be verified by the Redis Database server: /var/lib/redis:/sbin/nologin

https://bugzilla.redhat.com/show_bug.cgi?id=2165956

* [X] Foreman 3.1/Katello 4.3 on EL7 & EL8 (Satellite 6.11 EL7/8, orcharhino 6.1 on EL7, orcharhino 6.2 on EL7/8)


mjivraja:
For `2.5` : #1975

adamlazik1:
Wow, those ifdefs there are a mess. I realize this is out of the scope of this PR, but while you're at it, do you think you could simplify those ifdefs? There are a lot of single bullet points encapsulated in ifdef followed by exactly the same ifdef.

Also, could you please confirm with the reporter that this applies all the way back to 2.5? Thanks!

maximiliankolb:
Can someone verify that this user comes from `puppet-redis` which is mentioned in the Puppetfile[.lock] in foreman-installer? I have checked only develop and 3.1-stable; so I assume this PR should get cherry-picked to all versions listed.

Change looks good to me!

ekohl (changes requested):
This comes from installing the `redis` package. On EL8:
```console
# rpm -q --scripts redis
preinstall scriptlet (using /bin/sh):
getent group redis &> /dev/null || \
groupadd -r redis &> /dev/null
getent passwd redis &> /dev/null || \
useradd -r -g redis -d /var/lib/redis -s /sbin/nologin \
-c 'Redis Database Server' redis &> /dev/null
exit 0
...
```

We install Redis using `puppet-redis`, but not everywhere. It is pulled in by Foreman and Pulp. This means Redis is only present on Server and Content Proxies, but not regular proxies.

And I'll agree with @maximiliankolb that the ifdefs are very complex and could be simplified.

mjivraja:
> And I'll agree with @maximiliankolb that the ifdefs are very complex and could be simplified.

After going through that part, I understand that to keep the name of accounts in alphabetical order, the earlier contributors have played `ifdefs` more frequently. If alphabetical order is not necessary, we can definitely declutter.

maximiliankolb (approved):
LGTM! I honestly like the alphabetical order. Yes, the list with all its "ifdefs" is messy, but adding one more line does not significantly make it worse.

I however propose the following idea: create `snip_users-{project-context}.adoc` (or similar) and then simply include this one file with the attribute as file name.

guides/common/modules/snip_users-to-avoid-in-project-server.adoc
```diff
@@ -0,0 +1,30 @@
...
+* foreman-proxy
+* postgres
+ifdef::katello,satellite,orcharhino[]
+* pulp
+endif::[]
+* puppet
+ifdef::katello,satellite,orcharhino[]
+* qdrouterd
+endif::[]
+ifdef::katello,satellite,orcharhino[]
```
ekohl, line 20: This part is redundant.
```suggestion
```

ekohl (changes requested):
I always try to avoid moving things around **and** making a change in the same commit. Getting into the habit of creating one commit to move things and then another to add the new part (in this case add Redis) makes the life of a reviewer much easier. Now I need to do a lot more work to see what exactly changed.

maximiliankolb:
@mjivraja I meant something like this:

```
snip_users_foreman-deb.adoc:
  redis
  user 1
  user 2
  ...
```

```
snip_users_katello.adoc:
  katello-only-user
  user2
  ...
```

```
snip_users_satellite.adoc:
  satellite-one
  user-2
  ...
```

...

and then you'd include `snip_users_{project-context}.adoc` without any ifdefs etc, because asciidoc would only include the correct file depending on the build context.

maximiliankolb:
Maybe "project-context" is wrong, but this should work if files for all build targets exist.

Example: `guides/common/attributes.adoc:include::attributes-{build}.adoc[]`

adamlazik1:
> > And I'll agree with @maximiliankolb that the ifdefs are very complex and could be simplified.
> 
> After going through that part, I understand that to keep the name of accounts in alphabetical order, the earlier contributors have played `ifdefs` more frequently. If alphabetical order is not necessary, we can definitely declutter.
> 
> CC: @adamlazik1

It was not my intention to reorder the list. By redundant ifdefs I meant sequences like this:
```
ifdef::satellite,katello,orcharhino[]
* Item 1
endif:[]
ifdef::satellite,katello,orcharhino[]
* Item 2
endif:[]

**Comment by mjivraja** [comment]
> 

Right! I mis-calculated. 

Added an individual files for users' list in all builds. I have also added the `project-context` attribute for few build and I hope it will not have any major impact in the guides.

**Comment by mjivraja** [comment]
> Maybe "project-context" is wrong, but this should work if files for all build targets exist.
> 
> Example: `guides/common/attributes.adoc:include::attributes-{build}.adoc[]`

Right! I mis-calculated.

Added an individual files for users' list in all builds. I have also added the `project-context` attribute for few build and I hope it will not have any major impact in the guides.

File: guides/common/modules/ref_system-requirements.adoc [hunk 2]
```diff
@@ -19,13 +19,13 @@ endif::[]
 ifdef::katello,satellite[]
 ifeval::["{context}" == "{project-context}"]
 * A minimum of 20 GB RAM is required for {ProjectServer} to function.
-In addition, a minimum of 4 GB RAM of swap space is also recommended.
+In addition, a minimum 4 GB RAM of swap space is also recommended.
```
maximiliankolb, line 22: Was this change intentional? If not, please revert or move to a separate commit. As goes for below.

guides/common/modules/snip_users-to-avoid-in-foreman-deb.adoc
```diff
@@ -0,0 +1,9 @@
+[id="users-to-avoid-in-foreman-deb_{context}"]
+* www-data
+ifeval::["{context}" == "{project-context}"]
+* foreman
+endif::[]
```
maximiliankolb, line 5:
```suggestion
* foreman
```

I assume this can be shortened, because my understanding is that the build target for Foreman on Debian is "foreman-deb", same as the project-context attribute you've defined.

maximiliankolb (commented):
Adding the attribute should not be necessary:

```
guides/common/attributes-base.adoc::project-context: foreman
guides/common/attributes-orcharhino.adoc::project-context: orcharhino
guides/common/attributes-satellite.adoc::project-context: satellite
```

On a general note: I want to investigate if it's possible to omit the "project-context" because it should be identical to `{build}` as defined in guides/common/attributes.adoc. However, this should not be part of your PR.


For now, I think we can safely remove the "ifeval" in each snippet because we're already including the correct snippet per build target.

guides/common/modules/ref_system-requirements.adoc
mjivraja: Yes. This change is intentional from my side. Corrected from grammatical POV.

ekohl: Is it a correction? To me it reads worse.

And if it's a correction, it deserves its own PR. Adding multiple fixes in a single PR (especially one that's complicated) is a recipe for never extending the time to review by a lot.

mjivraja: Alright. I will revert it. :+1:  (Thought it's small and we often correct it so I considered it here itself).

mjivraja: > On a general note: I want to investigate if it's possible to omit the "project-context" because it should be identical to {build} as defined in guides/common/attributes.adoc. However, this should not be part of your PR.

For `katello` it loads fewer users (which is incorrect according to snip files). It is calling the file `snip_***foreman.adoc` instead `snip_***katello.adoc`

mjivraja: > For now, I think we can safely remove the "ifeval" in each snippet because we're already including the correct snippet per build target.

Noted :+1:

guides/common/modules/snip_users-to-avoid-in-satellite.adoc
```diff
@@ -0,0 +1,17 @@
+[id="users-to-avoid-in-satellite_{context}"]
+* apache
+ifeval::["{context}" == "{project-context}"]
+* foreman
+endif::[]
```
maximiliankolb, line 5:
```suggestion
* foreman
```

guides/common/modules/snip_users-to-avoid-in-satellite.adoc
```diff
@@ -0,0 +1,17 @@
...
+* foreman-proxy
+* postgres
+* pulp
+* puppet
+* qdrouterd
+ifeval::["{context}" == "{project-context}"]
+* qpidd
+endif::[]
```
maximiliankolb, line 13:
```suggestion
* qpidd
```

guides/common/modules/snip_users-to-avoid-in-satellite.adoc
```diff
@@ -0,0 +1,17 @@
...
+* redis
+ifeval::["{context}" == "{project-context}"]
+* tomcat
+endif::[]
```
maximiliankolb, line 17:
```suggestion
* tomcat
```

maximiliankolb (commented):
I still believe all ifdefs in the target-specific snippets are unnecessary because they'd always evaluate to true. Can you please test this? I am comments to one file as an example.

adamlazik1:
> I still believe all ifdefs in the target-specific snippets are unnecessary because they'd always evaluate to true. Can you please test this? I am comments to one file as an example.

Not necessarily true. In Satellite, we have the module included in two guides. One of the two happens to have the context set to project-context, while the other has a different context. This way the content is differentiated in the two guides.

guides/common/modules/ref_system-requirements.adoc
mjivraja: > For `katello` it loads fewer users (which is incorrect according to snip files). It is calling the file `snip_***foreman.adoc` instead `snip_***katello.adoc`

I am also concerned about this.

maximiliankolb:
> > I still believe all ifdefs in the target-specific snippets are unnecessary because they'd always evaluate to true. Can you please test this? I am comments to one file as an example.
> 
> Not necessarily true. In Satellite, we have the module included in two guides. One of the two happens to have the context set to project-context, while the other has a different context. This way the content is differentiated in the two guides.

If that's the case, what's our gain from moving it to several snippets? Does it make sense to keep the PR as is or revert everything?

maximiliankolb:
It was my idea to use a snippet which turned out to be a bad idea. :see_no_evil: 

I agree with Adam to please revert adding snippets and then remove unnecessary ifdef blocks by hand. Thanks for the patience Malhar.

maximiliankolb:
It was my idea to use a snippet which turned out to be a bad idea. :see_no_evil: 

I agree with Adam to please revert adding snippets and then remove unnecessary ifdef blocks by hand. Thanks for the patience Malhar.

mjivraja:
> It was my idea to use a snippet which turned out to be a bad idea. see_no_evil
> 
> I agree with Adam to please revert adding snippets and then remove unnecessary ifdef blocks by hand. Thanks for the patience Malhar.
> 
> My previous comment was only aiming towards "I came up with such an elegant solution using snippets which turns out to not work at all due to the attribute usage", nothing else.

No worries at all @maximiliankolb @adamlazik1. I am learning here :grinning: 
Reverted the changes :+1:

mjivraja:
> Search for ifdef blocks with the same condition in immediate succession and remove them. 

@adamlazik1 If we keep the accounts alphabetically, no immediate conditions are found that are the same. Maybe I am missing somewhere?

adamlazik1:
After a closer inspection it turns out there is only one instance. I am a bit embarrassed to cause such a fuss over nothing :sweat_smile: 

Anyway leaving it here as a patch since I cannot comment in the code directly
```diff
diff --git a/guides/common/modules/ref_system-requirements.adoc b/guides/common/modules/ref_system-requirements.adoc
index c0e76dc18..90fda556b 100644
--- a/guides/common/modules/ref_system-requirements.adoc
+++ b/guides/common/modules/ref_system-requirements.adoc
@@ -70,8 +70,6 @@ endif::[]
 * puppet
 ifdef::katello,satellite,orcharhino[]
 * qdrouterd
-endif::[]
-ifdef::katello,satellite,orcharhino[]
 ifeval::["{context}" == "{project-context}"]
 * qpidd
 endif::[]
```

mjivraja:
> After a closer inspection it turns out there is only one instance. I am a bit embarrassed to cause such a fuss over nothing sweat_smile
> 
> Anyway leaving it here as a patch since I cannot comment in the code directly
> 
> ```diff
> diff --git a/guides/common/modules/ref_system-requirements.adoc b/guides/common/modules/ref_system-requirements.adoc
> index c0e76dc18..90fda556b 100644
> --- a/guides/common/modules/ref_system-requirements.adoc
> +++ b/guides/common/modules/ref_system-requirements.adoc
> @@ -70,8 +70,6 @@ endif::[]
>  * puppet
>  ifdef::katello,satellite,orcharhino[]
>  * qdrouterd
> -endif::[]
> -ifdef::katello,satellite,orcharhino[]
>  ifeval::["{context}" == "{project-context}"]
>  * qpidd
>  endif::[]
> ```

Agreed :+1: 
I ignored `ifeval` initially :sweat_smile:

adamlazik1:
Merged and cherry-picked:
   eb0197267..0b7d7bdac  3.1 -> 3.1
   8d81e306f..171759078  3.2 -> 3.2
   6eb6f4624..608408fe3  3.3 -> 3.3
   5c3e7a5ab..d64e7438a  3.4 -> 3.4
   6c3b556a2..30ca6a3eb  3.5 -> 3.5
   6eaddb7b8..5557b197f  3.6 -> 3.6

PR #1978, opened 2023-02-08
Update chapter about Registration in Deploying on AWS

Lennonka:
The chapter was referring to the bootstrap script, but it should refer to the non-deprecated global registration method.

https://bugzilla.redhat.com/show_bug.cgi?id=2167767

* [x] Foreman 3.1/Katello 4.3 on EL7 & EL8 (Satellite 6.11 EL7/8, orcharhino 6.1 on EL7, orcharhino 6.2 on EL7/8)


adamlazik1:
Shouldn't we just include *Registering hosts* module instead? Than you can add a custom note there using `ifeval::[{context} == ...]`

adamlazik1:
Another solution that came to my mind is to include that note at the end of the respecting installing chapters.

guides/doc-Deploying_Project_on_AWS/topics/AWS-registering-hosts.adoc
```diff
@@ -0,0 +1,10 @@
+[[aws-registering-hosts]]
+= Registering Hosts to {Project}
+
+ifdef::foreman-el,katello[]
+This procedure applies only to users of the Katello plug-in.
+endif::[]
```
maximiliankolb, line 6: I think we should drop this and instead only show the file for Katello, orcharhino, and Satellite.

guides/doc-Deploying_Project_on_AWS/topics/AWS-registering-hosts.adoc
```diff
@@ -0,0 +1,10 @@
...
+
+When you install {ProjectServer} and {SmartProxyServer}, you must then register the content hosts on EC2 instances to {Project}.
+
```
maximiliankolb, line 9:
```suggestion
```

guides/doc-Deploying_Project_on_AWS/topics/AWS-registering-hosts.adoc
Lennonka: I'm confused.
Registering Hosts is enabled for foreman-el and foreman-deb as well. Isn't it still possible to register hosts even without Katello, only without using activation keys?

ekohl: I think we should hide the chapter altogether if it doesn't apply, rather than showing content that users shouldn't read. The "this only applies to Katello users" part is something we can deal with in asciidoc.

Lennonka: Actually, if hosts can be registered regardless of Katello (with subtle differences), then I think we can do the contrary and include this chapter for all builds, because the linked chapter Registering Hosts in Managing Hosts covers registration for all builds on a subtler level. And remove the sentence in the condition as well.

guides/doc-Deploying_Project_on_AWS/topics/AWS-registering-hosts.adoc
```diff
@@ -0,0 +1,10 @@
+[[aws-registering-hosts]]
+= Registering Hosts to {Project}
+
+ifdef::foreman-el,katello[]
+This procedure applies only to users of the Katello plug-in.
+endif::[]
+
+When you install {ProjectServer} and {SmartProxyServer}, you must then register the content hosts on EC2 instances to {Project}.
```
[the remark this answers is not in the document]
Lennonka: @ekohl Should we consider the navigation path *Hosts* > *Content Hosts* discouraged at this point and always navigate through *Hosts* > *All Hosts*?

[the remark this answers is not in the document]
ianballou: There are a couple reasons to still use the content hosts pages. Firstly, our "bulk manage packages/errata" workflows aren't moved over to the main Hosts page yet. Also, if you aren't using Simple Content Access, subscription management still needs to happen on a content host page.

So, overall, we are hoping that most users will not need to look at the old (legacy) content host pages. I suppose we could go as far to say that Hosts > Content Hosts is discouraged.

@jeremylenz do you agree?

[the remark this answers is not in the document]
jeremylenz: Yes, Hosts > Content Hosts at this point is only used for bulk actions.

Even for subscription actions, if you try to use Hosts > Content Hosts and click on a host, it will take you to the new host details page for that host (unless you've got that setting turned off), and then you'll have to use the "Legacy Content host UI" link to get to the old page that has the Subscriptions tab. So, may as well just use Hosts > All Hosts for everything at this point, except content host bulk actions.

Lennonka:
@adamlazik1 I'd prefer to keep the link because there are other sections in the same chapter that might interest the users and including the whole assembly would make the chapter unnecessarily long.

ekohl (approved):
I agree that we shouldn't include it. After installing it on AWS I'd consider it just any installation and then you proceed with the managing hosts guide.

maximiliankolb (approved):
LGTM.

Lennonka:
Cherry-picked:
- 4ee15691d..da6514c7a  3.5 -> 3.5
- 45e1a5a90..2406b198d  3.4 -> 3.4
- c0083d482..449c54e25  3.3 -> 3.3

Conflicts on 3.2 and 3.1, I'll file separate PRs.

PR #2012, opened 2023-02-21
Add insecure flag in API and CLI procedure

mjivraja:
In installing the Capsule Server, the operating system will not be able to connect to the Project with HTTPS. The `inecure` option is required for the command to get executed successfully. This is mainly because of the self-signed SSL certificate. Therefore, it is affected to hammer and curl commands both while registering the Project server with Proxy.

https://bugzilla.redhat.com/show_bug.cgi?id=2170829

* [X] Foreman 3.1/Katello 4.3 on EL7 & EL8 (Satellite 6.11 EL7/8, orcharhino 6.2 on EL7/8)


ekohl (changes requested):
We have been very strict in avoiding making it default, precisely because it's insecure. Documenting it this way makes it seem normal, which it shouldn't be.

I think this needs a separate example explaining that it may be needed. This is needed when the CA on Foreman is not trusted by clients. That's usually true for the default Ca, but users who have custom SSL certificates signed by a trusted CA won't have the issue. For them it's better to verify the certificates.

mjivraja:
> We have been very strict in avoiding making it default, precisely because it's insecure. Documenting it this way makes it seem normal, which it shouldn't be.
> 
> I think this needs a separate example explaining that it may be needed. This is needed when the CA on Foreman is not trusted by clients. That's usually true for the default Ca, but users who have custom SSL certificates signed by a trusted CA won't have the issue. For them it's better to verify the certificates.

So, in this case, can we just keep the `Note` instead of modifying the command?

ekohl:
As a reader an example is useful, but style wise I'm not sure what's best. It may end up too long.

mjivraja:
> As a reader an example is useful, but style wise I'm not sure what's best. It may end up too long.

Added a note and reverted the modifications in commands.

guides/common/modules/snip_host-registration-steps.adoc
```diff
@@ -59,6 +59,8 @@ ifdef::katello,satellite,orcharhino[]
 -d '{ "registration_command": { "activation_keys": ["_My_Activation_Key_1_, _My_Activation_Key_2_"] }}'
 ----
 +
+NOTE: If the resulting host registration command fails, you may require to add `--insecure` flag. It will then allow hosts to get registered with {Project} via {SmartProxy}.
```
maximiliankolb, line 62:
```suggestion
[NOTE]
====
If your hosts do not trust the SSL certificate of {ProjectServer}, add the `--insecure` flag to the registration command.
====
```

mjivraja: Addressed :+1:

maximiliankolb (approved):
LGTM! Any other comments/input?

adamlazik1:
LGTM also but maybe @ekohl wants to have the last look.

guides/common/modules/snip_host-registration-steps.adoc
```diff
@@ -59,6 +59,11 @@ ifdef::katello,satellite,orcharhino[]
 -d '{ "registration_command": { "activation_keys": ["_My_Activation_Key_1_, _My_Activation_Key_2_"] }}'
 ----
 +
+[NOTE]
+====
+If your hosts do not trust the SSL certificate of {ProjectServer}, add the `--insecure` flag to the registration command.
```
ekohl, line 64: I think this only applied to the Hammer command. For the API request it needs to be inside the JSON data.

The whole idea is that you have:
1. Client -> Foreman: request command
2. Foreman -> Client: return command
3. Client -> Host: run command
4. Host -> Foreman: register itself

If you add `--insecure` to `curl`, that's step one, but you want it to be done in step 4, so you must request Foreman in step one to render the correct flag in step 2.

The original PR was the correct way to do that, so the Bugzilla report has the correct instructions.

ekohl (changes requested):

guides/common/modules/snip_host-registration-steps.adoc
mjivraja: Means, we should go with this commit https://github.com/theforeman/foreman-documentation/commit/3a20d632ca9b1f18e44427a3958ee2bb46f20d0d ?

ekohl: That commit unconditionally specifies the parameter. I was suggesting to:
* For the CLI procedure: by default run without `--insecure 1`, but then document that it is an option. So move the note you have here to the CLI procedure.
* For the API procedure: provide examples where `insecure` is specified.

Katello (note compared to https://github.com/theforeman/foreman-documentation/commit/3a20d632ca9b1f18e44427a3958ee2bb46f20d0d it doesn't quote `true`):
```json
{ "registration_command": { "activation_keys": ["_My_Activation_Key_1_, _My_Activation_Key_2_"], "insecure": true }}
```

Non-Katello:
```json
{ "registration_command": { "insecure": true }}
```

ekohl: I can now see that `For the API procedure: provide examples` was ambiguous. What I meant was to first document the examples as they are today and then document you can make it insecure and then provide additional examples where it is insecure.

I know this makes it longer, but I think providing instructions on how to change the JSON is non-trivial and providing a secure example and an insecure example is better.

guides/common/modules/snip_host-registration-steps.adoc
```diff
@@ -32,13 +32,30 @@ ifndef::katello,satellite,orcharhino[]
 ----
 # hammer host-registration generate-command
 ----
++
+If your hosts do not trust the SSL certificate of {ProjectServer}, add the `--insecure` flag to the registration command.
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# hammer host-registration generate-command
```
evgeni, line 40:
```suggestion
# hammer host-registration generate-command \
```

guides/common/modules/snip_host-registration-steps.adoc
```diff
@@ -32,13 +32,30 @@ ifndef::katello,satellite,orcharhino[]
...
+--insecure 1
```
evgeni, line 41: should this be `--insecure true`? I guess both work as far as the API is concerned, but using a true (lol) boolean seems more correct here?

guides/common/modules/snip_host-registration-steps.adoc
```diff
@@ -32,13 +32,30 @@ ifndef::katello,satellite,orcharhino[]
...
 # hammer host-registration generate-command \
 --activation-keys "_My_Activation_Key_"
 ----
++
+If your hosts do not trust the SSL certificate of {ProjectServer}, add the `--insecure` flag to the registration command.
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# hammer host-registration generate-command \
+--activation-keys "_My_Activation_Key_"
```
evgeni, line 56:
```suggestion
--activation-keys "_My_Activation_Key_" \
```

guides/common/modules/snip_host-registration-steps.adoc
```diff
@@ -73,6 +90,17 @@ ifndef::katello,satellite,orcharhino[]
...
 ----
++
+If the above host registration command fails, you may require to add `--insecure` flag.
+It will then allow hosts to get registered with {Project} via {SmartProxy}.
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# curl -X POST https://{foreman-example-com}/api/registration_commands \
+--user "_My_User_Name_" \
+-H 'Content-Type: application/json'
```
evgeni, line 101:
```suggestion
-H 'Content-Type: application/json' \
```

guides/common/modules/snip_host-registration-steps.adoc
```diff
@@ -73,6 +90,17 @@ ifndef::katello,satellite,orcharhino[]
...
+-d '{ "registration_command": { "insecure": true }}
```
evgeni, line 102:
```suggestion
-d '{ "registration_command": { "insecure": true }}'
```

guides/common/modules/snip_host-registration-steps.adoc
```diff
@@ -32,13 +32,30 @@ ifndef::katello,satellite,orcharhino[]
 ----
 # hammer host-registration generate-command
 ----
++
+If your hosts do not trust the SSL certificate of {ProjectServer}, add the `--insecure` flag to the registration command.
```
ekohl, line 36: I'd also say what it does:
```suggestion
If your hosts do not trust the SSL certificate of {ProjectServer}, SSL validation can be disabled by adding the `--insecure` flag to the registration command.
````

File: guides/common/modules/snip_host-registration-steps.adoc [hunk 9]
```diff
@@ -73,6 +90,17 @@ ifndef::katello,satellite,orcharhino[]
 --user "_My_User_Name_" \
 -H 'Content-Type: application/json'
 ----
++
+If the above host registration command fails, you may require to add `--insecure` flag.
+It will then allow hosts to get registered with {Project} via {SmartProxy}.
```
ekohl, line 95: Can we use the same instruction as for the command line, with the exact instruction replaced?
```suggestion
If your hosts do not trust the SSL certificate of {ProjectServer}, SSL validation can be disabled by adding the `insecure` flag to the registration command.
```

guides/common/modules/snip_host-registration-steps.adoc
```diff
@@ -83,6 +111,17 @@ ifdef::katello,satellite,orcharhino[]
 -d '{ "registration_command": { "activation_keys": ["_My_Activation_Key_1_, _My_Activation_Key_2_"] }}'
 ----
 +
+If the above host registration command fails, you may require to add `--insecure` flag.
+It will then allow hosts to get registered with {Project} via {SmartProxy}.
```
ekohl, line 115: And here
```suggestion
If your hosts do not trust the SSL certificate of {ProjectServer}, SSL validation can be disabled by adding the `insecure` flag to the registration command.
```

guides/common/modules/snip_host-registration-steps.adoc
mjivraja: Agree and Addressed :+1:

guides/common/modules/snip_host-registration-steps.adoc
mjivraja: Added to CLI method as well :+1:

guides/common/modules/snip_host-registration-steps.adoc
```diff
@@ -94,7 +131,7 @@ You can use the ID of any {SmartProxyServer} that you configured for host regist
 +
 Include `{ "force": true }` to register a host that has been previously registered to a {SmartProxyServer}.
 endif::[]
-+
+
```
ekohl, line 134: I don't think this is correct since the diff of the rendered result now shows `+` before `Include` is rendered literally.

mjivraja: Fixed. Aligned those paras to 1st point.

guides/common/modules/snip_host-registration-steps.adoc
```diff
@@ -99,6 +136,7 @@ To enter a password as a command line argument, use `username:password` syntax.
 Keep in mind this can save the password in the shell history.
 Alternatively, you can use a temporary personal access token instead of a password.
 To generate a token in the {ProjectWebUI}, navigate to *My Account* > *Personal Access Tokens*.
+
```
maximiliankolb, line 139:
```suggestion
```

guides/common/modules/snip_host-registration-steps.adoc
```diff
@@ -83,11 +110,21 @@ ifdef::katello,satellite,orcharhino[]
...
+-H 'Content-Type: application/json' \
+-d '{ "registration_command": { "activation_keys": ["_My_Activation_Key_1_, _My_Activation_Key_2_"], "insecure": true }}'
+----
++
 Use an activation key to simplify specifying the environments.
 For more information, see {ContentManagementDocURL}Managing_Activation_Keys_content-management[Managing Activation Keys] in _{ContentManagementDocTitle}_.
 endif::[]
-ifeval::["{context}" == "load-balancing"]
 +
+ifeval::["{context}" == "load-balancing"]
```
maximiliankolb, line 127:
```suggestion
ifeval::["{context}" == "load-balancing"]
+
```

maximiliankolb (commented):
two small changes look unrelated to the "--insecure" change.

ekohl (approved):
I think this is correct. One CI check failed on a HTTP link check, so I restarted that. Once it completes the preview should be up to date.

ekohl:
Of course it now failed on another URL. Restarted again.

ekohl (approved):
Preview looks correct. @maximiliankolb any last comments?

guides/common/modules/snip_host-registration-steps.adoc
```diff
@@ -32,13 +32,30 @@ ifndef::katello,satellite,orcharhino[]
 ----
 # hammer host-registration generate-command
 ----
++
+If your hosts do not trust the SSL certificate of {ProjectServer}, SSL validation can be disabled by adding the `--insecure` flag to the registration command.
```
maximiliankolb, line 36: "can be disabled" > "you can disable"

guides/common/modules/snip_host-registration-steps.adoc
```diff
@@ -32,13 +32,30 @@ ifndef::katello,satellite,orcharhino[]
...
+----
 endif::[]
 ifdef::katello,satellite,orcharhino[]
 [options="nowrap" subs="+quotes,attributes"]
 ----
 # hammer host-registration generate-command \
 --activation-keys "_My_Activation_Key_"
 ----
++
+If your hosts do not trust the SSL certificate of {ProjectServer}, SSL validation can be disabled by adding the `insecure` flag to the registration command.
```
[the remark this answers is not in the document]
mjivraja: Apology. Missed it.

maximiliankolb (approved):
LGTM

maximiliankolb:
   0d488a969..8c313cfa1  3.6 -> 3.6
   e27cdb4fe..702da2527  3.5 -> 3.5
   3f6687b42..ed041ab1a  3.4 -> 3.4
   8b254fc64..2d6b8f971  3.3 -> 3.3
   793610e6c..f10e89b2b  3.2 -> 3.2
   241bf3460..7a46dbdb3  3.1 -> 3.1

PR #2014, opened 2023-02-22
Rename Satellite Client repo

Lennonka:
I suggest adding "Red Hat" to "Satellite Client 6" to make it clearer that it's a Red Hat repository.

### DIFF
--- a/guides/common/attributes-satellite.adoc
+++ b/guides/common/attributes-satellite.adoc
@@ -130,7 +130,7 @@
 // Used in downstream guides
 :RepoRHEL7ServerSoftwareCollections: rhel-server-rhscl-7-rpms
 
-:project-client-name: Satellite Client 6
+:project-client-name: {ProjectName} Client{nbsp}6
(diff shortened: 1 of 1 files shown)


Lennonka:
I'm not sure if this is a good idea and it might require more work to remove "Red Hat" on some occurences of the attribute.

adamlazik1:
I my opinion, I would leave the repository name as-is, reasoning being it is (from my understanding) the exact name of the repository. I could be wrong, though.

Lennonka:
@adamlazik1 My reasoning is the same, because the exact name of the repo does contain the "Red Hat" prefix.

adamlazik1:
If that's true then it would be of course better to change the name. But as you're saying, it won't be a trivial task because the Red Hat prefix is there at some places. Could probably be solved with some regexes, though.

Lennonka:
I've decided not to proceed with this change, because the opposite approach might actually work better.

PR #2015, opened 2023-02-22
Rewrite Installing Puppet agent during Provisioning

Lennonka:
Changes:

1. Rewrite of the workflow for downstream products that should use only puppet-agent shipped with the product.
2. Katello users should enable the puppetlabs*) repo using a CV/activation key as well.
3. puppetlabs*) enablement without activation keys is kept only for vanilla Foreman.
4. puppetlabs*) enablement without activation keys uses the non-versioned parameter `enable-puppetlabs-repo`, which enables the latest Puppet (that is version 7 currently)

*) puppetlabs are the official Puppet repositories at puppet.com

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
+++ b/guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
@@ -1,30 +1,59 @@
 [id="Installing_Puppet_Agent_during_Host_Provisioning_{context}"]
 = Installing Puppet Agent during Host Provisioning
 
-Install the Puppet agent during the host provisioning process.
+Use this procedure to install the Puppet agent during the host provisioning process.
+
+ifdef::satellite[]
+.Prerequisites
+* You enabled and synchronized the *{project-client-name}* repository to {Project}.
+For more information, see {ContentManagementDocURL}Importing_Content_content-management[Importing Content] in _{ContentManagementDocTitle}_.
+* You created an activation key that enables the *{project-client-name}* repository for hosts.
+For more information, see {ContentManagementDocURL}Managing_Activation_Keys_content-management[Managing Activation Keys] in _{ContentManagementDocTitle}_.
+endif::[]
+ifdef::katello,orcharhino[]
+.Prerequisites
+* You created a Product and repository for the upstream Puppet agent, such as `\https://yum.puppet.com` or `\https://apt.puppet.com`, and synchronized the repository to {Project}.
+For more information, see {ContentManagementDocURL}Importing_Content_content-management[Importing Content] in _{ContentManagementDocTitle}_.
+* You created an activation key that enables the Puppet agent repository for hosts.
+For more information, see {ContentManagementDocURL}Managing_Activation_Keys_content-management[Managing Activation Keys] in _{ContentManagementDocTitle}_.
+endif::[]
 
 .Procedure
-. Navigate to *Hosts > Provisioning Templates*.
+. Navigate to *Hosts* > *Provisioning Templates*.
 . Select a provisioning template depending on your host provisioning method.
 For more information, see {ProvisioningDocURL}types-of-provisioning-templates_provisioning[Types of Provisioning Templates] in _{ProvisioningDocTitle}_.
+ifndef::katello,orcharhino,satellite[]
+. Ensure the `puppetlabs_repo` snippet is included as follows:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+<%= snippet 'puppetlabs_repo' %>
+----
++
+Note that this snippet is already included in the templates shipped with {Project}, such as `Kickstart default` or `Preseed default`.
+endif::[]
 . Ensure the `puppet_setup` snippet is included as follows:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
 <%= snippet 'puppet_setup' %>
 ----
-. Enable Puppet using a host parameter for a single host or host group:
 +
-Add a host parameter named `enable-puppet7` for Puppet 7
-ifndef::satellite[]
-or `enable-puppet6` for Puppet 6
+Note that this snippet is already included in the templates shipped with {Project}, such as `Kickstart default` or `Preseed default`.
+ifndef::katello,orcharhino,satellite[]
+. Enable the Puppet agent from the official Puppet repository using one of the following host parameters in global parameters, a host group, or for a single host:
+
+* Add a host parameter named `enable-puppetlabs-repo` for the latest stable Puppet release in the unversioned repository.
+* Add a host parameter named `enable-official-puppet7-repo` for the Puppet 7 release repository.
+
++
+Select the *boolean* type and set the value to `true`.
 endif::[]
-as type boolean set to `true`.
-. Optional: To install the Puppet agent directly from https://yum.puppet.com/[yum.puppet.com], add a host parameter named `enable-puppetlabs-puppet7-repo` for Puppet 7
-ifndef::satellite[]
-or `enable-puppetlabs-puppet6-repo` for Puppet 6
+ifdef::katello,orcharhino,satellite[]
+. Enable the Puppet agent using a host parameter in global parameters, a host group, or for a single host.
+Add a host parameter named `enable-puppet7`, select the *boolean* type, and set the value to `true`.
 endif::[]
-as type boolean set to `true`.
-ifndef::katello[]
-Only use this if you don't provide Puppet agent to the host using its activation key.
+ifdef::katello,orcharhino,satellite[]
+. Ensure your host has access to the Puppet agent packages from {ProjectServer} by using an appropriate activation key.
 endif::[]
+. Ensure that you assign a Puppet environment, Puppet server and Puppet CA server to your host.


Lennonka:
@maximiliankolb @ekohl By the way, I noticed that [Puppet 6 is projected as EOL in February 2023](https://www.puppet.com/docs/puppet/7/platform_lifecycle.html#platform-releases). I wonder if we need to update the supported versions for orcharhino (only v6) and upstream (v6 and v7). I don't how the packaging of Puppet works in either of them :)

guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
```diff
@@ -1,30 +1,56 @@
 [id="Installing_Puppet_Agent_during_Host_Provisioning_{context}"]
 = Installing Puppet Agent during Host Provisioning
 
-Install the Puppet agent during the host provisioning process.
+Use this procedure to install the Puppet agent during the host provisioning process.
+
+ifdef::orcharhino,satellite[]
```
ekohl, line 6: @maximiliankolb can you verify if orcharhino ships a vendored Puppet in their client repositories? Or it is like Katello and relies on upstream repositories?

guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
```diff
@@ -1,30 +1,56 @@
...
 <%= snippet 'puppet_setup' %>
 ----
-. Enable Puppet using a host parameter for a single host or host group:
+ifndef::katello,orcharhino,satellite[]
+. Enable the official Puppet repository using a host parameter in global parameters, a host group, or for a single host.
 +
-Add a host parameter named `enable-puppet7` for Puppet 7
-ifndef::satellite[]
-or `enable-puppet6` for Puppet 6
+Add a host parameter named `enable-puppetlabs-repo` for the latest Puppet 7 or `enable-puppetlabs-puppet6-repo` for Puppet 6, select the `boolean` type, and set the value to `true`.
```
ekohl, line 44: We have 4 versions now:
* `enable-puppetlabs-repo`: unversioned repository, latest stable release (so will be updated to Puppet 8 when it comes out later this year)
* `enable-official-puppet7-repo`: Puppet 7, latest current stable now
* `enable-puppetlabs-puppet6-repo`: Puppet 6, going EOL soon
* `enable-puppetlabs-puppet5-repo`: Puppet 5, already EOL

Probably shouldn't document the EOL ones, but giving the full overview.

guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
```diff
@@ -1,30 +1,56 @@
...
 endif::[]
-as type boolean set to `true`.
-. Optional: To install the Puppet agent directly from https://yum.puppet.com/[yum.puppet.com], add a host parameter named `enable-puppetlabs-puppet7-repo` for Puppet 7
+. Enable Puppet agent using a host parameter in global parameters, a host group, or for a single host.
++
 ifndef::satellite[]
-or `enable-puppetlabs-puppet6-repo` for Puppet 6
+Add a host parameter named `enable-puppet7` for Puppet 7 or `enable-puppet6` for Puppet 6, select the `boolean` type and set the value to `true`.
```
ekohl, line 49: This is redundant if you already set the repo parameter, so I'd recommend to leave it out.

guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
```diff
@@ -39,16 +39,27 @@ endif::[]
...
 endif::[]
+ifdef::katello[]
 . Enable Puppet agent using a host parameter in global parameters, a host group, or for a single host.
 +
-ifndef::satellite[]
 Add a host parameter named `enable-puppet7` for Puppet 7 or `enable-puppet6` for Puppet 6, select the `boolean` type and set the value to `true`.
+ifdef::orcharhino[]
+. Enable Puppet agent using a host parameter in global parameters, a host group, or for a single host.
++
+Add a host parameter named `enable-puppet6` for Puppet 6, select the `boolean` type and set the value to `true`.
```
ekohl, line 59: Any reason to use Puppet 6 here? @maximiliankolb?

I'll stress that Puppet 6 is about to be end of life so I'd urge you to get to Puppet 7.

guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
```diff
@@ -1,30 +1,68 @@
...
-as type boolean set to `true`.
-ifndef::katello[]
-Only use this if you don't provide Puppet agent to the host using its activation key.
+ifdef::orcharhino[]
+. Enable Puppet agent using a host parameter in global parameters, a host group, or for a single host.
++
+Add a host parameter named `enable-puppet6` for Puppet 6, select the `boolean` type and set the value to `true`.
+endif::[]
+ifdef::satellite[]
+. Enable Puppet agent using a host parameter in global parameters, a host group, or for a single host.
```
ekohl, line 62: This paragraph looks the same as Katelloand orcharhino. Can we group them?

guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
Lennonka: Waiting with another push, until we clarify this.

maximiliankolb:
orcharhino relies on Puppet Clients from puppet.com and supports Puppet 7.

Lennonka:
@maximiliankolb In that case we had a mistake in the docs previously. I'm suggesting a fix to the supported versions as well.
@ekohl Is it reasonable to drop Puppet 6 already? And we should also cherry-pick this change to older releases as well, perhaps?

ekohl:
>  @ekohl Is it reasonable to drop Puppet 6 altogether already? And we should also cherry-pick this change to older releases as well, perhaps?

I'm not sure, but I just asked some Puppet folks when the real EOL is. He thought April, but will check to know for sure.

In upstream we've historically been more flexible in the supported range so I'm hesitant to become too aggressive in it. Removing support in an existing release is IMHO something we should avoid. Especially since there are no differences in the two from a Foreman perspective, other than a different repo.

guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
```diff
@@ -1,30 +1,63 @@
...
+.Prerequisites
+* You enabled and synchronized the *{project-client-name}* repository to {Project}.
+For more information, see {ContentManagementDocURL}Importing_Content_content-management[Importing Content] in _{ContentManagementDocTitle}_.
+* You created an activation key that enables the *{project-client-name}* repository for hosts.
+For more information, see {ContentManagementDocURL}Managing_Activation_Keys_content-management[Managing Activation Keys] in _{ContentManagementDocTitle}_.
+* You assigned the activation key to a host group.
+endif::[]
+ifdef::katello,orcharhino[]
+.Prerequisites
+* You created a Product and repository for the upstream Puppet platform, such as `\https://yum.puppet.com` or `\https://apt.puppet.com`, and synchronized the repository to {Project}.
```
maximiliankolb, line 16:
```suggestion
* You created a Product and repository for the upstream Puppet agent, such as `\https://yum.puppet.com` or `\https://apt.puppet.com`, and synchronized the repository to {Project}.
```

guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
```diff
@@ -1,30 +1,63 @@
...
+For more information, see {ContentManagementDocURL}Importing_Content_content-management[Importing Content] in _{ContentManagementDocTitle}_.
+* You created an activation key that enables the repository for hosts.
```
maximiliankolb, line 18:
```suggestion
* You created an activation key that enables the Puppet agent repository for hosts.
```

ekohl:
> I'm not sure, but I just asked some Puppet folks when the real EOL is. He thought April, but will check to know for sure.

It is indeed February, or in other words, after tomorrow.

The reason I was confused is that earlier the document stated July, but the document has been corrected.

guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
```diff
@@ -1,30 +1,63 @@
...
-Add a host parameter named `enable-puppet7` for Puppet 7
-ifndef::satellite[]
-or `enable-puppet6` for Puppet 6
 endif::[]
-as type boolean set to `true`.
-. Optional: To install the Puppet agent directly from https://yum.puppet.com/[yum.puppet.com], add a host parameter named `enable-puppetlabs-puppet7-repo` for Puppet 7
-ifndef::satellite[]
-or `enable-puppetlabs-puppet6-repo` for Puppet 6
+ifdef::katello,orcharhino[]
+Add a host parameter named `enable-puppet7` for Puppet 7 or `enable-puppet6` for Puppet 6, select the `boolean` type and set the value to `true`.
```
maximiliankolb, line 56:
```suggestion
Add a host parameter named `enable-puppet7` for Puppet 7 or `enable-puppet6` for Puppet 6, select the `boolean` type, and set the value to `true`.
```

guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
```diff
@@ -1,30 +1,63 @@
...
 endif::[]
-as type boolean set to `true`.
-ifndef::katello[]
-Only use this if you don't provide Puppet agent to the host using its activation key.
+ifdef::satellite[]
+Add a host parameter named `enable-puppet7` for Puppet 7, select the `boolean` type and set the value to `true`.
```
maximiliankolb, line 59:
```suggestion
Add a host parameter named `enable-puppet7` for Puppet 7, select the `boolean` type, and set the value to `true`.
```

guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
```diff
@@ -1,30 +1,63 @@
...
 endif::[]
+ifdef::katello,orcharhino,satellite[]
+. When you create the host, assign the host group that contains the appropriate activation key.
```
maximiliankolb, line 62:
```suggestion
. Ensure your host has access to the Puppet agent from {ProjectServer} by using an appropriate activation key.
```

maximiliankolb (commented):
A few small comments, rest LGTM

guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
Lennonka: During provisioning (Create Host), the user cannot assign an activation key directly, but only indirectly using a host group instead.

maximiliankolb: If I go to hosts > create hosts; create a host; select a host group, then I can still overwrite the AK as host parameter or add one.

Lennonka:
I've realized I'm missing some important steps. I'll have to give it another go. Sorry about that!

guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
Lennonka: If you mean the `kt_activation_keys` parameter, it seems that it's only interpreted like this in the Global Registration template and only when defined on the hostgroup level.
Or what parameters exactly do you mean?

Lennonka: Oh, it's in a snippet. Never mind!

Lennonka:
I've added one more step at the end.

Lennonka:
@ekohl I have removed Puppet 6 from here as well. Can you please take another look?

guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
```diff
@@ -1,30 +1,55 @@
...
-or `enable-puppetlabs-puppet6-repo` for Puppet 6
+ifdef::katello,orcharhino,satellite[]
+. Enable the Puppet agent using a host parameter in global parameters, a host group, or for a single host.
+Add a host parameter named `enable-puppet7`, select the *boolean* type, and set the value to `true`.
 endif::[]
-as type boolean set to `true`.
-ifndef::katello[]
-Only use this if you don't provide Puppet agent to the host using its activation key.
+ifdef::katello,orcharhino,satellite[]
+. Ensure your host has access to the Puppet agent from {ProjectServer} by using an appropriate activation key.
```
ekohl, line 53: Perhaps be explicit in what? My mind first went to firewalls when you said access
```suggestion
. Ensure your host has access to the Puppet agent packages from {ProjectServer} by using an appropriate activation key.
```

ekohl (approved):
Minor comment, but otherwise :+1: for the general content. Preview failed to render because of some timeout in the linkchecker, so I've just restarted that. It does mean I haven't read through the rendered versions yet and was limited to mental parsing.

Lennonka:
@ekohl Would you like to give it one last check or can I merge it?

guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
```diff
@@ -1,30 +1,55 @@
...
+For more information, see {ContentManagementDocURL}Managing_Activation_Keys_content-management[Managing Activation Keys] in _{ContentManagementDocTitle}_.
+endif::[]
 
 .Procedure
-. Navigate to *Hosts > Provisioning Templates*.
+. Navigate to *Hosts* > *Provisioning Templates*.
 . Select a provisioning template depending on your host provisioning method.
 For more information, see {ProvisioningDocURL}types-of-provisioning-templates_provisioning[Types of Provisioning Templates] in _{ProvisioningDocTitle}_.
+ifndef::katello,orcharhino,satellite[]
+. Ensure the `puppetlabs_repo` snippet is included as follows:
```
ekohl, line 26: Open question: I see it's copying what was there before with `puppet_setup`, but I wonder if it's clear to users that the default provisioning templates we ship (kickstart, preseed, etc) already do this. That is actually the reason we tell users to set the host parameter `enable-puppetlabs-repo`: it ensures the snippet gets loaded. So it feels like a redundant instruction.

ekohl (approved):
I don't mind merging this now and solving the open question later.

guides/common/modules/proc_installing-puppet-agent-during-host-provisioning.adoc
Lennonka: `enable-puppetlabs-repo` ensures both the snippet gets loaded in the default templates AND is also used inside the snippet.
But if we assume that the user uses the default provisioning templates, then both snippet instructions are redundant. Perhaps we should remove the whole part about editing a template?

Ron-Lavi: Do we want to remove the section or should we merge this and continue if needed in a follow-up? @Lennonka @ekohl

Lennonka: I was also thinking about adding a note to each snippet step that it's already included in the shipped templates. I'll do that instead removing the steps.

Lennonka:
Mentioned that the snippets are included in shipped templates and keeping the snippet steps.

Lennonka:
Cherry-picked:
- 9bb63c883..ecc720d3a  3.6 -> 3.6
- 8e2cda771..ad1608972  3.5 -> 3.5
- 9ac0a0c08..14a976958  3.4 -> 3.4

PR #2026, opened 2023-02-27
Highlight difference in shellhook dir vs URL

adamlazik1:

BZ link: https://bugzilla.redhat.com/show_bug.cgi?id=2173600

* [x] Foreman 3.6/Katello 4.8


adamlazik1:
Could someone review this small change?

maximiliankolb:
Is this a case where users read "shellhooks" instead of "shellhook" in the URL?

I am not really a huge fan of adding admonitions everywhere, but I am OK with adding this if you say it's important.

adamlazik1:
Yes. According to our folks from customer support, customers get frustrated because they do not notice the difference in singular vs plural.

maximiliankolb:
> Yes. According to our folks from customer support, customers get frustrated because they do not notice the difference in singular vs plural, which leads to invalid URL.

Then feel free to move ahead.

PR #2039, opened 2023-03-03
Introduce "client-content-X" attributes

maximiliankolb:

This is the start to allow content specifically aimed at users managing Debian or SLES hosts. My PR does not change the output yet, because for all build targets, `:dnf:` is set and `:apt:` remains unset.
In downstream, I build all three big guides (managing content, provisioning hosts, configuring hosts) for multiple OS which set or unset specific attributes.

For the future, I could see a way of adding specific procedures in upstream docs twice: once for rpm and once for deb.

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/attributes-base.adoc
+++ b/guides/common/attributes-base.adoc
@@ -117,3 +117,4 @@
 :SmartProxyServersTitle: {SmartProxyServerTitle}s
 :Team: {Project} developers
 :sectanchors:
+:client-content-dnf:
--- a/guides/common/modules/con_content-filter-overview.adoc
+++ b/guides/common/modules/con_content-filter-overview.adoc
@@ -43,7 +43,9 @@ The list of package groups is based on the repositories added to the Content Vie
 | *Erratum (by ID)* | Select which specific errata to add to the filter.
 The list of Errata is based on the repositories added to the Content View.
 | *Erratum (by Date and Type)* | Select a issued or updated date range and errata type (Bugfix, Enhancement, or Security) to add to the filter.
+ifdef::client-content-dnf[]
 | *Module Streams* | Select whether to include or exclude specific module streams.
 The *Module Streams* option filters modular RPMs and errata, but does not filter non-modular content that is associated with the selected module stream.
+endif::[]
 | *Container Image Tag* | Select whether to include or exclude specific container image tags.
 |===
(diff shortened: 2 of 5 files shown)


adamlazik1 (approved):
LGTM.
Just a side note, subject line of the fourth commits should probably be:
Show **module** streams only for rpm

ekohl (changes requested):
My first suspicion here was that `rpm` was for RPM-based platforms, so another way to capture EL vs Debian/Ubuntu. This all looks to be for content, so can we prefix them with `content-`? I can imagine in the future we may have more, like `content-python` or `content-container`.

maximiliankolb:
> My first suspicion here was that `rpm` was for RPM-based platforms, so another way to capture EL vs Debian/Ubuntu. This all looks to be for content, so can we prefix them with `content-`? I can imagine in the future we may have more, like `content-python` or `content-container`.

I have changed the attribute to "dnf" to allow for "apt" and "zypper" as additions to differentiate between managed hosts running EL, SLES, and Debian/Ubuntu.

If there is the need for python or container content, I'd be happy to extend this.

ekohl:
I still think this must have a prefix since `apt` and `dnf` are too generic.

maximiliankolb:
Is this OK for you @ekohl ?

maximiliankolb:
Are there any other comments or suggestions? If not, I'd like to merge this by the end of tomorrow.

Lennonka:
LGTM, I think you can go ahead with merging.

Lennonka (approved):
Approving.

ekohl (commented):
I don't really understand your objections to qualify it.

My reasoning is that we have various content types:
* rpm
* deb

So from there I'd expect `content-rpm` and `content-deb`.

Then we have the actual client systems, which you're using here. Those are the actual client systems. We have yum/dnf and apt. I'd consider using `content-client-yum` and `content-client-apt` instead. Or even `content-rpm-client-yum` and `content-deb-client-apt`.

maximiliankolb:
My reasoning was to keep it as short as possible because I'll have to type it a lot, both upstream and downstream. However, I see your point to make it more human readable and now went with `client-content-X`, X being "dnf" for EL, "apt" for Debian/Ubuntu, and "zypper" for SLES.

PR #2043, opened 2023-03-06
Correct content on Managing_Content for foreman-{el,deb}

ekohl:
The goal of this is to at least show some text making it actionable for users. I'm not sure what a good text is, but this is a pattern I'd like to apply more.

Fow now we don't display the Managing_Content guide at all for stable releases, so no cherry picks are needed though they should be safe.

* [ ] Foreman 3.6/Katello 4.8
* [ ] Foreman 3.5/Katello 4.7 (planned Satellite 6.13)

### DIFF
--- a/guides/doc-Managing_Content/master.adoc
+++ b/guides/doc-Managing_Content/master.adoc
@@ -8,7 +8,13 @@ ifdef::satellite[]
 include::common/modules/proc_providing-feedback-on-red-hat-documentation.adoc[leveloffset=+1]
 endif::[]
 
-// Render only for relevant and finished contexts
+ifndef::katello,satellite,orcharhino[]
+Mangaging content requires the Katello plugin.
+Migrating from Foreman without Katello to Foreman with Katello is unsupported.
+// TODO: This needs to link to index-katello.html, not index-foreman-el.html.
+// can we somehow get the attributes from katello without breaking the rest of the document?
+Please have a look at {InstallingServerDocURL}[{InstallingServerDocTitle}] and migrate your hosts.
+endif::[]
 ifdef::katello,satellite,orcharhino[]
 
 include::common/modules/con_introduction-to-content-management.adoc[leveloffset=+1]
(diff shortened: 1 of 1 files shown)


guides/doc-Managing_Content/master.adoc
```diff
@@ -8,14 +8,20 @@ ifdef::satellite[]
 include::common/modules/proc_providing-feedback-on-red-hat-documentation.adoc[leveloffset=+1]
 endif::[]
 
-// Render only for relevant and finished contexts
+ifndef::katello,satellite,orcharhino[]
+Mangaging content requires the Katello plugin.
+Migrating from Foreman without Katello to Foreman with Katello is unsupported.
+// TODO: This needs to link to index-katello.html, not index-foreman-el.html.
+// can we somehow get the attributes from katello without breaking the rest of the document?
+Please have a look at {InstallingServerDocURL}[{InstallingServerDocTitle}] and migrate your hosts.
```
adamlazik1, line 16: I believe you can paste the link directly since this is a single occurence.

ekohl: Order was a bit difficult, but I want to make this a more common pattern and used this as an example in https://community.theforeman.org/t/making-docs-theforeman-org-the-primary-documentation-source/32657. So we may run it into it more often. I'll turn it into a draft for now since it's not ready to merge.

adamlazik1: You can also create new attributes in attributes-base, something like `KatelloInstallingServerDocURL`. Of course it is not an ideal solution because it duplicates the attributes but nonetheless it is a way to access the attribute outside of a specific build.

ekohl: I had thought about including the Katello attributes, but then it should be the very last statement. Any thoughts on that?

adamlazik1: Yes, that works as well in this case. The question is if this procedure can be replicated to future places where build-specific attributes are needed outside the build. If you need to input such attribute somewhere in the middle of docs, you will have to find another way.  That being said, if this scenario is the only type of use-case for this than it works just fine.

One additional note:
It would probably be good to keep this text in a module to keep master file as clean as possible.

ekohl:
Already done in 724dc9cb039d8012b1aa2970756f8e4a68dc7d62

PR #2051, opened 2023-03-10
Use FirewallD services where possible

ekohl:
Using services is easier to read for users and a smaller chance of typos in port numbers. It also can enable helpers if needed.

It doesn't use the foreman-proxy for content proxies, since it runs on a non-standard port.

Currently untested, but here to start the discussion if this is something we want. From there on I also want to investigate what we open by default. For example, opening the DNS/DHCP/TFTP ports if we don't enable those services by default is redundant. Should we open them only if we tell users to enable the service, like we do with MQTT?

* [ ] Foreman 3.6/Katello 4.8
* [ ] Foreman 3.5/Katello 4.7 (planned Satellite 6.13)

### DIFF
--- a/guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
+++ b/guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
@@ -14,18 +14,20 @@ include::snip_firewalld.adoc[]
 [options="nowrap"]
 ----
 # firewall-cmd \
---add-port="53/udp" --add-port="53/tcp" \
---add-port="67/udp" \
---add-port="69/udp" \
---add-port="80/tcp" --add-port="443/tcp" \
+--add-service=dns \
+--add-service=dhcp \
+--add-service=tftp \
+--add-service=http \
+--add-service=https \
 ifdef::katello,satellite,orcharhino[]
 --add-port="5647/tcp" \
---add-port="8000/tcp" --add-port="9090/tcp" \
+--add-port="8000/tcp" \
+--add-port="9090/tcp" \
 endif::[]
 ifndef::katello,satellite,orcharhino[]
---add-port="8443/tcp" \
+--add-service=foreman-proxy \
 endif::[]
---add-port="8140/tcp"
+--add-service=puppetmaster
 ----
 . Make the changes persistent:
 +
--- a/guides/common/modules/proc_enabling-connections-to-capsule.adoc
+++ b/guides/common/modules/proc_enabling-connections-to-capsule.adoc
@@ -13,17 +13,21 @@ include::snip_firewalld.adoc[]
 [options="nowrap"]
 ----
 # firewall-cmd \
---add-port="53/udp" --add-port="53/tcp" \
---add-port="67/udp" \
---add-port="69/udp" \
+--add-service=dns \
+--add-service=dhcp \
+--add-service=tftp \
 ifdef::katello,satellite,orcharhino[]
---add-port="80/tcp" --add-port="443/tcp" \
+--add-service=http \
+--add-service=https \
 --add-port="5647/tcp" \
 endif::[]
---add-port="8140/tcp" \
---add-port="8443/tcp" \
+--add-service=puppetmaster \
+ifndef::katello,satellite,orcharhino[]
+--add-service=foreman-proxy
+endif::[]
 ifdef::katello,satellite,orcharhino[]
---add-port="8000/tcp" --add-port="9090/tcp"
+--add-port="8000/tcp" \
+--add-port="9090/tcp"
 endif::[]
 ----
 
--- a/guides/common/modules/proc_configuring-remote-execution-for-pull-client-smart-proxy.adoc
+++ b/guides/common/modules/proc_configuring-remote-execution-for-pull-client-smart-proxy.adoc
@@ -23,11 +23,11 @@ If you wish to avoid this scenario, configure all {SmartProxies} to use the same
  # {installer-scenario-smartproxy} \
 --foreman-proxy-plugin-remote-execution-script-mode pull-mqtt
 ----
-. Configure the firewall to allow MQTT service on port 1883:
+. Configure the firewall to allow MQTT service:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-# firewall-cmd --add-port="1883/tcp"
+# firewall-cmd --add-service=mqtt
 # firewall-cmd --runtime-to-permanent
 ----
 . In `pull-mqtt` mode, hosts subscribe for job notifications to the {SmartProxy} through which they are registered.


guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
```diff
@@ -14,18 +14,18 @@ include::snip_firewalld.adoc[]
...
 ----
 # firewall-cmd \
---add-port="53/udp" --add-port="53/tcp" \
---add-port="67/udp" \
---add-port="69/udp" \
---add-port="80/tcp" --add-port="443/tcp" \
+--add-service=dns \
+--add-service=dhcp \
+--add-service=tftp \
+--add-service=http --add-service=https \
```
adamlazik1, line 20:
```suggestion
--add-service=http \
--add-service=https \
```

guides/common/modules/proc_enabling-connections-to-capsule.adoc
```diff
@@ -13,15 +13,17 @@ include::snip_firewalld.adoc[]
...
 --add-port="5647/tcp" \
 endif::[]
---add-port="8140/tcp" \
---add-port="8443/tcp" \
+--add-service=puppetmaster \
+ifndef::katello,satellite,orcharhino[]
+--add-service=foreman-proxy
+endif::[]
 ifdef::katello,satellite,orcharhino[]
 --add-port="8000/tcp" --add-port="9090/tcp"
```
adamlazik1, line 28:
```suggestion
--add-port="8000/tcp" \ 
--add-port="9090/tcp"
```

adamlazik1 (commented):
Looks good from writing standpoint.

I also like the idea of adding services that have default ports rather than adding ports and protocols. Definitely more simple for the user and less prone to error.

guides/common/modules/proc_enabling-connections-to-capsule.adoc
```diff
@@ -13,15 +13,17 @@ include::snip_firewalld.adoc[]
...
 # firewall-cmd \
---add-port="53/udp" --add-port="53/tcp" \
---add-port="67/udp" \
---add-port="69/udp" \
+--add-service=dns \
+--add-service=dhcp \
+--add-service=tftp \
 ifdef::katello,satellite,orcharhino[]
---add-port="80/tcp" --add-port="443/tcp" \
+--add-service=http --add-service=https \
```
maximiliankolb, line 20:
```suggestion
--add-service=http \
--add-service=https \
```

guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
```diff
@@ -14,18 +14,20 @@ include::snip_firewalld.adoc[]
 [options="nowrap"]
 ----
 # firewall-cmd \
---add-port="53/udp" --add-port="53/tcp" \
---add-port="67/udp" \
---add-port="69/udp" \
---add-port="80/tcp" --add-port="443/tcp" \
+--add-service=dns \
+--add-service=dhcp \
+--add-service=tftp \
```
ekohl, line 19: Some discussion: we don't enable dns, dhcp and tftp by default, so why do we open the ports by default? Shouldn't it be part of the chapters where we do enable the services?

guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
```diff
@@ -14,18 +14,20 @@ include::snip_firewalld.adoc[]
...
+--add-service=http \
+--add-service=https \
 ifdef::katello,satellite,orcharhino[]
 --add-port="5647/tcp" \
---add-port="8000/tcp" --add-port="9090/tcp" \
+--add-port="8000/tcp" \
```
ekohl, line 24: Same here: port 8000 is the HTTP port of foreman-proxy, but that's not enabled by default in katello and satellite (not sure about orcharhino).

guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
```diff
@@ -14,18 +14,20 @@ include::snip_firewalld.adoc[]
...
---add-port="67/udp" \
---add-port="69/udp" \
---add-port="80/tcp" --add-port="443/tcp" \
+--add-service=dns \
+--add-service=dhcp \
+--add-service=tftp \
+--add-service=http \
+--add-service=https \
 ifdef::katello,satellite,orcharhino[]
 --add-port="5647/tcp" \
```
ekohl, line 23: Same here: katello-agent is not enabled by default so qpid is not there. Should this be open by default, or should it be part of the procedure to enable katello-agent?

guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
```diff
@@ -14,18 +14,20 @@ include::snip_firewalld.adoc[]
...
---add-port="8000/tcp" --add-port="9090/tcp" \
+--add-port="8000/tcp" \
+--add-port="9090/tcp" \
 endif::[]
 ifndef::katello,satellite,orcharhino[]
---add-port="8443/tcp" \
+--add-service=foreman-proxy \
 endif::[]
---add-port="8140/tcp"
+--add-service=puppetmaster
```
ekohl, line 30: Puppet is off by default in katello/satellite so I don't think we should open the port by default. Rather in the enabling Puppet procedure.

guides/common/modules/proc_enabling-connections-to-capsule.adoc
```diff
@@ -13,17 +13,21 @@ include::snip_firewalld.adoc[]
 [options="nowrap"]
 ----
 # firewall-cmd \
---add-port="53/udp" --add-port="53/tcp" \
---add-port="67/udp" \
---add-port="69/udp" \
+--add-service=dns \
+--add-service=dhcp \
+--add-service=tftp \
```
ekohl, line 18: Same as above: disabled by default

guides/common/modules/proc_enabling-connections-to-capsule.adoc
```diff
@@ -13,17 +13,21 @@ include::snip_firewalld.adoc[]
...
 ifdef::katello,satellite,orcharhino[]
---add-port="80/tcp" --add-port="443/tcp" \
+--add-service=http \
+--add-service=https \
 --add-port="5647/tcp" \
```
ekohl, line 22: Same as above: katello-agent is disabled by default

guides/common/modules/proc_enabling-connections-to-capsule.adoc
```diff
@@ -13,17 +13,21 @@ include::snip_firewalld.adoc[]
...
 endif::[]
---add-port="8140/tcp" \
---add-port="8443/tcp" \
+--add-service=puppetmaster \
```
ekohl, line 24: Same as above: puppet is disabled on katello/satellite by default.

guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
adamlazik1: I we are able to reliably track all the procedures which require enabling these services then adding them to firewall rules can certainly be moved there. That being said, I don't know how many of these procedures are there.

ekohl: DNS: `--foreman-proxy-dns true --foreman-proxy-dns-managed true` then we install a DNS server and DNS needs to be opened
DHCP: `--foreman-proxy-dhcp true --foreman-proxy-dhcp-managed true`
TFTP: `--foreman-proxy-tftp true --foreman-proxy-tftp-managed true`

Note we often `--foreman-proxy-dhcp=true` instead of `--foreman-proxy-dhcp true` in docs, so careful when grepping. Sometimes we omit the `-managed true` part since that's default, but technically we should include it when we rely on it to be sure.

guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
ekohl: Enabled via `--foreman-proxy-http true` (and on by default for the content proxies/capsules).

guides/common/modules/proc_enabling-connections-to-capsule.adoc
```diff
@@ -13,17 +13,21 @@ include::snip_firewalld.adoc[]
...
+ifndef::katello,satellite,orcharhino[]
+--add-service=foreman-proxy
+endif::[]
 ifdef::katello,satellite,orcharhino[]
---add-port="8000/tcp" --add-port="9090/tcp"
+--add-port="8000/tcp" \
+--add-port="9090/tcp"
```
adamlazik1, line 30: Can't we also add the `foreman-proxy` service here? Would that do something different for satellite than what it currently does?

ekohl: The firewalld service defines it as port 8443: https://github.com/firewalld/firewalld/blob/master/config/services/foreman-proxy.xml.

Today we discussed in our team, and there is a plan to move Katello's Smart Proxies to port 8443.

We used to run Candlepin (as a reverse proxy) on port 8443 on content proxies. Right now we're at the point where it has moved to port 443, but we maintain it on upgraded installations to allow users to migrate. We're close to the point where we want to drop that compatibility, which would free up the port. Then we can start to work on the Smart Proxy port migration as well. Once we've completed that, the ecosystem is standardized and we could use the port.

It is possible to modify the service in firewalld, but then the user no longer gets updates if firewalld changes the original service, so I opted to use the port number instead.

maximiliankolb (commented):
One small suggestion, rest LGTM!

guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
ekohl: I would like that as well, but I don't have time to work on it myself now and prefer to at least get this in. Then open an issue for the follow up work.

guides/common/modules/proc_enabling-connections-to-capsule.adoc
adamlazik1: Understood, thanks for explaining.

adamlazik1 (approved):
LGTM

ekohl:
Rebased. Can we merge this now?

adamlazik1:
Yes, go ahead.

PR #2057, opened 2023-03-13
Put unbreakable spaces around '>'

adamlazik1:
I wonder if this is desirable? When working on an unrelated task I realized that linebreak after or before **>** does not look good. I ran a sed command on it so if we decide we want to implement this change I would have to verify the changes manually.

```
sed -i 's/\*\([^ ]*\)\* > *\([^ ]*\)\*/\*\1\*\{nbsp\}\>\{nbsp\}\2\*/g' guides/common/modules/*.adoc
```

* [x] Foreman 3.6/Katello 4.8


maximiliankolb:
I'm OK with the change and will review this if you want to, but I'm not 100% sure if it's necessary or if there are any downsides. If there's more general approval, please ping me & I will review it.

adamlazik1:
Well, it's no huge improvement to the docs and it only takes effect on long enough lines, which there aren't that many of. But I guess it is not harmful either. Maybe I could create an attribute for it so it can be easier replicated in future, like `{NoBreakGT}` or `{NoBreakArrow}`? It also could be me just over-complicating things.

The downside I see could be potential merge conflicts in  future but this can be easily resolved by running the command on each branch. Then there is of course five times more reviewing. But I think I will undertake the task once I have free time.

I am open to more thoughts on this PR.

maximiliankolb:
Good idea. I'll review it next week if there are no objections. Feel free to ping people/request opinions.

adamlazik1:
@maximiliankolb Do you agree with creating an attribute for this?

maximiliankolb:
No, I'd rather not. Looking at it again, maybe it's completely unnecessary. I am currently leaning towards "no" as in "I am unsure where's the real benefit for a problem we've experienced".

adamlazik1:
Fair point. I might just close this, then.

PR #2064, opened 2023-03-16
3.4: Add missing prerequite file for load balancing

maximiliankolb:
I broke 3.4 when cherry-picking https://github.com/theforeman/foreman-documentation/pull/2061


adamlazik1 (approved):
Also applies to 3.3.

maximiliankolb:
> I see I approved before the checks finished. Looks like there is some other issue as well.

A failure to compare against 3.4 is expected because it cannot build 3.4. :upside_down_face: 

So I'll merge if/after "check-html" is green & then GHA on 3.4 itself should be competely green again.

PR #2089, opened 2023-03-23
Various small changes for  orcharhino build target

maximiliankolb:
Please review commit by commit. Note that "hammer proxy_content" must have been an error.

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/attributes-base.adoc
+++ b/guides/common/attributes-base.adoc
@@ -59,6 +59,7 @@
 :foreman-installer-package: foreman-installer
 :foreman-maintain: foreman-maintain
 :FreeIPA: FreeIPA
+:hammer-smart-proxy: hammer proxy
 :installer-log-file: /var/log/foreman-installer/foreman.log
 // this is foreman-installer or foreman-installer --scenario MyScenario if there are multiple
 :installer-scenario: foreman-installer
--- a/guides/common/modules/proc_registering-hosts.adoc
+++ b/guides/common/modules/proc_registering-hosts.adoc
@@ -98,11 +98,11 @@ It has no effect on other operating systems.
 +
 You must enable the following repositories on a registered machine:
 
-* RHEL 6: `rhel-6-server-rpms`
-* RHEL 7: `rhel-7-server-rpms`
-* RHEL 8: `rhel-8-for-x86_64-appstream-rpms`
+* {RHEL} 6: `rhel-6-server-rpms`
+* {RHEL} 7: `rhel-7-server-rpms`
+* {RHEL} 8: `rhel-8-for-x86_64-appstream-rpms`
 +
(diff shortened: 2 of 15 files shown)


guides/common/modules/proc_registering-hosts.adoc
```diff
@@ -90,6 +90,7 @@ If set to `Yes`, public SSH keys will be installed on the registered host.
 The inherited value is based on the `host_registration_remote_execution` parameter.
 It can be inherited, for example from a host group, an operating system, or an organization.
 When overridden, the selected value will be stored on host parameter level.
+ifdef::client-content-dnf[]
```
ekohl, line 93: I don't think this has to do with RPM content, but rather a RHEL client system. For me RPM content really is the Katello RPM content feature, but this part also applies to just Foreman. AFAIK all RHEL systems can be registered directly to https://cloud.redhat.com/ even without Katello and the rh-cloud plugin.

The source code says the same thing, because there's no `plugin_present?('katello')`: https://github.com/theforeman/foreman/blob/2bc0afba704a41e50a5ec30356f53ab97fbbf40b/app/views/unattended/provisioning_templates/host_init_config/host_init_config_default.erb#L68-L71

guides/common/attributes-base.adoc
```diff
@@ -54,6 +54,7 @@
 :customssl: SSL
 :customssltitle: SSL
 :Debian: Debian/Ubuntu
+:fdi-package-name: foreman-discovery-image
```
ekohl, line 57: There is no `foreman-discovery-image` package in upstream (nor any equivalent) and this should only be defined in `attributes-satellite.adoc`.

maximiliankolb: Moved & verified that it's not used for any other build target.

guides/common/modules/proc_registering-hosts.adoc
maximiliankolb: Yes, this is exactly what I want to bring upstream next: conditions for client operating systems so that we can differentiate between "client consumes Yum content" and "client runs RHEL".

For now, this is an improvement for downstream for me.

maximiliankolb (commented):
Applied one suggestion and rebased to master.

I would like to handle the second suggestion in a different PR if possible.

guides/common/modules/proc_registering-hosts.adoc
adamlazik1: I see no harm in this. From what I understand, this does not impact other than orcharhino documentation, so if it's an improvement, I say go for it.

maximiliankolb:
   d584c34e8..98337c0c9  3.6 -> 3.6
   ad70e2000..fd857b1e9  3.5 -> 3.5
   d7c1ce9f4..2f2ee0f15  3.4 -> 3.4
   c9739630d..87aefa7b4  3.3 -> 3.3

PR #2096, opened 2023-03-28
use installer scenario attribute for ssl cert deployment

evgeni:
this allows to drop the ifdef/ifndef as the rest of the section is identical

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
+++ b/guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
@@ -20,52 +20,27 @@ Note that for the `katello-certs-check` command to work correctly, Common Name (
 <3> Path to the Certificate Authority bundle.
 +
 If the command is successful, it returns two `{foreman-installer}` commands, one of which you must use to deploy a certificate to {ProjectServer}.
-ifdef::satellite[]
 +
 .Example output of `katello-certs-check`
 [options="nowrap", subs="+quotes,attributes"]
 ----
 Validation succeeded.
 
-To install the Red Hat Satellite Server with the custom certificates, run:
+To install the {ProjectName} server with the custom certificates, run:
 
-  {foreman-installer} --scenario satellite \
+  {installer-scenario} \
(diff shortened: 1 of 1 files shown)


evgeni:
meh, the diff was supposed to be empty…

evgeni:
In the docs context, `Project` renders as `Foreman`, `Red Hat Satellite` and `orcharino`.

In the `katello-certs-check` script, we default it it `Katello`: https://github.com/theforeman/foreman-installer/blob/416f45216493f9ad4d331a54d7cc92e42b8dbfd5/bin/katello-certs-check#L16 and later in Satellite override it to `Red Hat Satellite`:
```
# cat /usr/share/foreman-installer/katello-certs-check-branding 
PROJECT="Red Hat Satellite"
```

@maximiliankolb do you brand this output in Orcharino at all?

maximiliankolb:
> cat /usr/share/foreman-installer/katello-certs-check-branding

The latest orcharhino release does not use `/usr/share/foreman-installer/katello-certs-check-branding` to handle branding. From the "orcharhino build target" perspective, this looks good to me.

I also believe that it'd be OK to use `{ProjectServer}` despite the minor change in capitalization.

adamlazik1:
Diff looks good so far :+1: 

On a maybe related note, I always wondered whether we actually should be using "Satellite _server_" instead of "Satellite _Server_" downstream. @Lennonka any opinion?

Lennonka:
I think that {ProjectName} is correct in this case, because {ProjectServer} wouldn't render the required output for Satellite.

@adamlazik1 TBH, I don't have an opinion whether the capitalization should be "Satellite Server" or "Satellite server" :smile: I'm used to capitalize it, it's been established for quite some time, and I don't see a pressing reason to change it.

ekohl:
@evgeni this is still a draft. Do you intend to continue working on this?

evgeni:
I do, just not sure when

adamlazik1:
If https://github.com/theforeman/foreman-documentation/pull/2566 gets merged then this PR might be not necessary.

PR #2101, opened 2023-03-28
Improve Puppet-related text

Lennonka:
Various small improvements in Puppet-related modules. Mostly requested by RH QE, I added a couple more.

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_disabling-puppet.adoc
+++ b/guides/common/modules/proc_disabling-puppet.adoc
@@ -3,22 +3,22 @@
 
 To discontinue using Puppet in your {Project}, follow this procedure.
 
+Note that the command without the `--remove-all-data` argument removes all Puppet-related data in {Project} database.
+With the `--remove-all-data` argument, the command additionally removes Puppet server data files, including Puppet environments.
+
 .Prerequisite
 * Puppet is enabled on {Project}.
 
 .Procedure
-. If you used Puppet server on external {SmartProxies}, disable Puppet server on all external {SmartProxies}:
+. If you have used Puppet server on any {SmartProxies}, disable Puppet server on all {SmartProxies}:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {foreman-maintain} plugin purge-puppet
+# {foreman-maintain} plugin purge-puppet --remove-all-data
 ----
 . Disable Puppet server on {ProjectServer}:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {foreman-maintain} plugin purge-puppet [--remove-all-data]
+# {foreman-maintain} plugin purge-puppet --remove-all-data
 ----
-+
-The command without arguments will remove all Puppet-related data in {Project}.
-To remove Puppet server data files, including Puppet environments, add the `--remove-all-data` argument to the command.
--- a/guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
+++ b/guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
@@ -3,32 +3,26 @@
 
 Use this procedure to install and configure the Puppet agent on a host manually.
 
-ifndef::managing-configurations-puppet[]
-ifdef::satellite[]
-For more information about Puppet, see {ManagingConfigurationsPuppetDocURL}[Managing Configurations Using Puppet Integration in {Project}].
-endif::[]
-endif::[]
-
 .Prerequisites
 * The host must have a Puppet environment assigned to it.
 ifdef::satellite[]
-* The {ProjectName} Client repository must be enabled and synchronized to {ProjectServer}, and enabled on the host.
-The `puppet-agent` rpm is available within the following repositories:
-** {Team} {project-client-name} for RHEL 7 Server RPMs x86_64 {endash} `{RepoRHEL7ServerSatelliteToolsProductVersion}`
-** {Team} {project-client-name} for RHEL 8 Server RPMs <arch> {endash} `{RepoRHEL8ServerSatelliteToolsProductVersion}`
-** {Team} {project-client-name} for RHEL 9 Server RPMs <arch> {endash} `{RepoRHEL9ServerSatelliteToolsProductVersion}`
+* The *{project-client-name}* repository must be enabled and synchronized to {ProjectServer}, and enabled on the host.
+For more information, see {ContentManagementDocURL}Importing_Content_content-management[Importing Content] in _{ContentManagementDocTitle}_.
 endif::[]
-ifdef::orcharhino[]
-* The {Team} {project-client-name} repository must be enabled and synchronized to {ProjectServer}, and enabled on the host.
-endif::[]
-ifndef::satellite,orcharhino[]
+ifndef::satellite[]
 * Ensure a repository containing the Puppet agent is enabled on the host, for example https://apt.puppet.com/[apt.puppet.com] or https://yum.puppet.com/[yum.puppet.com].
 endif::[]
 
 .Procedure
 . Log in to the host as the `root` user.
 . Install the Puppet agent package.
-* On hosts running {EL}:
+* On hosts running {EL} 8 and above:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# dnf install puppet-agent
+----
+* On hosts running {EL} 7 and below:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
@@ -48,33 +42,27 @@ ifndef::satellite[]
 # zypper install puppet-agent
 ----
 endif::[]
-. Append the following server and environment settings to the `/etc/puppetlabs/puppet/puppet.conf` file.
+. Add the Puppet agent to `PATH` in your current shell using the following script:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+. /etc/profile.d/puppet-agent.sh
+----
+. Configure the Puppet agent.
 Set the `environment` parameter to the name of the Puppet environment to which the host belongs:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-environment = _My_Puppet_Environment_
-server = _{foreman-example-com}_
+# puppet config set server _{foreman-example-com}_ --section agent
+# puppet config set environment _My_Puppet_Environment_ --section agent
 ----
-. Configure the Puppet agent to start at boot:
-* On {EL} 6:
+. Start the Puppet agent service:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-# chkconfig puppet on
+# puppet resource service puppet ensure=running enable=true
 ----
-ifdef::satellite[]
-* On {EL} 7 and 8:
-endif::[]
-ifndef::satellite[]
-* On other operating systems:
-endif::[]
-+
-[options="nowrap", subs="+quotes,verbatim,attributes"]
-----
-# systemctl enable --now puppet
-----
-. Run the Puppet agent on the host:
+. Create a certificate for the host:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
@@ -83,9 +71,14 @@ endif::[]
 . In the {ProjectWebUI}, navigate to *Infrastructure* > *{SmartProxies}*.
 . From the list in the *Actions* column for the required {SmartProxyServer}, select *Certificates*.
 . Click *Sign* to the right of the required host to sign the SSL certificate for the Puppet agent.
+. On the host, run the Puppet agent again:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# puppet ssl bootstrap
+----
 
 ifndef::managing-configurations-puppet[]
-ifdef::satellite[]
-You can continue with {ManagingConfigurationsPuppetDocURL}[Managing Configurations Using Puppet Integration in {Project}].
-endif::[]
+.Additional Resources
+* For more information about Puppet, see {ManagingConfigurationsPuppetDocURL}[_{ManagingConfigurationsPuppetDocTitle}_].
 endif::[]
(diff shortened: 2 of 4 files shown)


guides/common/modules/proc_disabling-puppet.adoc
```diff
@@ -17,7 +17,7 @@ To discontinue using Puppet in your {Project}, follow this procedure.
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {foreman-maintain} plugin purge-puppet [--remove-all-data]
+# {foreman-maintain} plugin purge-puppet --remove-all-data
```
ekohl, line 20: This is now no longer in line with line 24.

guides/common/attributes-satellite.adoc
```diff
@@ -110,15 +110,15 @@
 :RepoSatelliteVersion: {ProductVersion}
 
 // RHEL 9 Satellite repos
-:RepoRHEL9ServerSatelliteToolsProductVersion: satellite-client-6-for-rhel-9-<arch>-rpms
```
ekohl, line 113: I think the idea of `<arch>` is that we support multiple arches (like `x86_64`), and this indicates that. I'm not sure which arches exactly, but perhaps we should list them somewhere to make it easier to consume.

guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
```diff
@@ -12,16 +12,13 @@ endif::[]
...
-* The {ProjectName} Client repository must be enabled and synchronized to {ProjectServer}, and enabled on the host.
-The `puppet-agent` rpm is available within the following repositories:
-** {Team} {project-client-name} for RHEL 7 Server RPMs x86_64 {endash} `{RepoRHEL7ServerSatelliteToolsProductVersion}`
-** {Team} {project-client-name} for RHEL 8 Server RPMs <arch> {endash} `{RepoRHEL8ServerSatelliteToolsProductVersion}`
-** {Team} {project-client-name} for RHEL 9 Server RPMs <arch> {endash} `{RepoRHEL9ServerSatelliteToolsProductVersion}`
+* The *{project-client-name}* repository must be enabled and synchronized to {ProjectServer}, and enabled on the host.
+The `puppet-agent` RPM is available within the following repositories:
+** {Team} {project-client-name} (for RHEL 7 Server) (RPMs) x86_64 {endash} `{RepoRHEL7ServerSatelliteToolsProductVersion}`
+** {Team} {project-client-name} for RHEL 8 x86_64 (RPMs) {endash} `{RepoRHEL8ServerSatelliteToolsProductVersion}`
+** {Team} {project-client-name} for RHEL 9 x86_64 (RPMs) {endash} `{RepoRHEL9ServerSatelliteToolsProductVersion}`
```
ekohl, line 19: Same here: AFAIk we have multiple arches and we should list them instead of implying we only have x86_64.

guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
```diff
@@ -64,7 +61,7 @@ server = _{foreman-example-com}_
 # chkconfig puppet on
 ----
 ifdef::satellite[]
-* On {EL} 7 and 8:
+* On {EL} 7, 8, and 9:
```
ekohl, line 64: Would it make sense to write 7 and beyond? I think it's fairly safe to assume EL 7+ will always use systemd.

guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
```diff
@@ -64,7 +61,7 @@ server = _{foreman-example-com}_
...
 endif::[]
 ifndef::satellite[]
 * On other operating systems:
```
ekohl, line 67: I can't reply inline, but line 74 states:

> Run the Puppet agent on the host:

But the command is `puppet ssl bootstrap`, which only ensures the certificates are set up. Running the agent is typically interpreted as also applying the catalog.

guides/common/attributes-satellite.adoc
adamlazik1: Also, if you decide to go through with this, you need to change line 27 in [proc_registering-clients-using-the-bootstrap-script.adoc](https://github.com/theforeman/foreman-documentation/blob/master/guides/common/modules/proc_registering-clients-using-the-bootstrap-script.adoc):
```
<1> Replace `<arch>` with the client architecture, for example `x86`.
```

guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
Lennonka: We only provide puppet-agent for x86_64 if you take a look into ohsnap.

guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
Lennonka: And which is needed here? Both?

guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
ekohl: We can ignore i686 for RHEL 6, but I also see `aarch64` for RHEL 8.

guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
ekohl: Looking closer: the order is surprising. Normally you:

* Run `puppet ssl bootstrap` to request a certificate
* Sign the request
* Enable the service

If you start the service, it will also submit a certificate request, but you have less control over it. So running `puppet ssl bootstrap` after enabling the service is redundant.

guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
Lennonka: I'll remove this from here altogether and fix [BZ2152621](https://bugzilla.redhat.com/show_bug.cgi?id=2152621) instead. I'll unify the prerequisites with other modules.

guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
Lennonka: I'll remove it.

guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
```diff
@@ -3,32 +3,25 @@
...
+ifdef::satellite[]
+. Install the Puppet agent package:
+endif::[]
+ifndef::satellite[]
 . Install the Puppet agent package.
 * On hosts running {EL}:
+endif::[]
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
```
adamlazik1, line 27: Should there be `{package-install}` used below?
Also, on 3.1 it will probably need to be split for EL 7 and 8.

guides/common/modules/proc_disabling-puppet.adoc
```diff
@@ -17,8 +17,8 @@ To discontinue using Puppet in your {Project}, follow this procedure.
...
 ----
 +
-The command without arguments will remove all Puppet-related data in {Project}.
-To remove Puppet server data files, including Puppet environments, add the `--remove-all-data` argument to the command.
+The command without arguments will remove all Puppet-related data in {Project} database.
```
adamlazik1, line 23: If all Puppet-related data does not include Puppet environments, shouldn't this be reworded?

Lennonka: That's why I added the "database" at the end. The argument will remove "files". (I assume.)

ekohl: https://github.com/theforeman/foreman_maintain/blob/master/definitions/procedures/puppet/remove_puppet_data.rb only gets executed if `--remove-all-data` is passed. So effectively `foreman-rake purge:puppet && rm -rf /etc/puppetlabs /opt/puppetlabs/server/data`. It always attempts to clean the DB by rolling back DB migrations from the plugin, but removing the files on disk only happens with `--remove-all-data`. The `foreman-rake purge:puppet` task can also remove the DB tables and columns without the plugin installed, so is there to really make sure it works.

guides/common/modules/proc_disabling-puppet.adoc
```diff
@@ -17,8 +17,8 @@ To discontinue using Puppet in your {Project}, follow this procedure.
...
+The `--remove-all-data` argument will remove Puppet server data files, including Puppet environments.
```
adamlazik1, line 24: Maybe like this (to highlight the difference)?

```suggestion
The command without arguments removes all Puppet-related data in {Project} database.
The command with the `--remove-all-data` argument removes all Puppet-related data in {Project} database and Puppet server data files, including Puppet environments.
```

guides/common/modules/proc_disabling-puppet.adoc
Lennonka: @ekohl So that means it's correct in the docs now, right?

guides/common/modules/proc_disabling-puppet.adoc
```diff
@@ -7,7 +7,7 @@ To discontinue using Puppet in your {Project}, follow this procedure.
 * Puppet is enabled on {Project}.
 
 .Procedure
-. If you used Puppet server on external {SmartProxies}, disable Puppet server on all external {SmartProxies}:
+. If you used Puppet server on any {SmartProxies}, disable Puppet server on all {SmartProxies}:
```
ekohl, line 10: Perhaps this is an influence of how you'd say it in Dutch, but I feel like something's missing here. Isn't it either "If you use Puppet server" or "If you've used Puppet server"?

guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
```diff
@@ -3,37 +3,31 @@
...
 # yum install puppet-agent
 ----
+* On hosts running {EL} 8 and above:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# dnf install puppet-agent
+----
```
adamlazik1, line 30:
```suggestion
* On hosts running {EL} 8 and above:
+
[options="nowrap", subs="+quotes,verbatim,attributes"]
----
# dnf install puppet-agent
----
* On hosts running {EL} 7 and below:
+
[options="nowrap", subs="+quotes,verbatim,attributes"]
----
# yum install puppet-agent
----
```

guides/common/modules/proc_disabling-puppet.adoc
ekohl: I'd say "without arguments" can be interpreted as simply running `{foreman-maintain}` since `plugin purge-puppet` are also arguments.

What's also incorrect is that the same command can be used (both with and without `--remove-all-data`) on both Server and Proxies. The current text implies `--remove-all-data` can only be used on Server.

It is required that you first run `plugin purge-puppet` (or `plugin purge-puppet --remove-all-data`) on all Proxies where Puppet was enabled. Once that's done, run it on Server. If there is still a Proxy with Puppet enabled, the purge command will fail. It's supposed to refuse to run, so no harm should be done, but we shouldn't set up users to run into that scenario.

Does that help on what should be achieved?

guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
ekohl, line 1: Note that https://www.puppet.com/docs/puppet/7/install_agents.html is the official documentation of how to set up an agent.

Some interesting bits:

* `sudo /opt/puppetlabs/bin/puppet resource service puppet ensure=running enable=true` is used as a platform independent way of starting and enabling Puppet
* `puppet config set server puppetserver.example.com --section main` as a command is used before suggesting to edit editing `/etc/puppetlabs/puppet/puppet.conf`
* `puppet ssl bootstrap` is used as a way to request the certificate (after setting the server)

guides/common/modules/proc_disabling-puppet.adoc
Lennonka: @ekohl Well, we're telling them to disable and remove Puppet on all SmartProxies in the previous step. Do you mean there also should be the `--remove-all-data` argument in the command in the previous step?

guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
```diff
@@ -48,44 +42,37 @@ ifndef::satellite[]
...
 endif::[]
-. Append the following server and environment settings to the `/etc/puppetlabs/puppet/puppet.conf` file.
+. Configure the Puppet agent.
 Set the `environment` parameter to the name of the Puppet environment to which the host belongs:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-environment = _My_Puppet_Environment_
-server = _{foreman-example-com}_
+# puppet config set server _{foreman-example-com}_ --section agent
```
ekohl, line 50: Not sure where to note this, but we should warn the user that it may be needed to load the settings.

The `puppet-agent` package installs `/etc/profile.d/puppet-agent.sh` which modifies `PATH` and `MANPATH` env vars. This is executed when your shell starts up, but you need to load it after running the `$x install puppet-agent` package if you're in the same shell.

The easiest way is:
```sh
. /etc/profile.d/puppet-agent.sh
```

Otherwise you end up with "command not found" errors.

This is a one time thing, but if a user just follows the procedure it's a common error. I'm unsure about the convention of documenting it, but please consider it.

ekohl (commented):
Thanks, I think this is getting much better.

guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
Lennonka: I simply added a step right after the package installation.

Lennonka:
Cherry-picked:
- 4ed29d2d5..2afa96cf3  3.6 -> 3.6
- acaedf0de..2b426fea5  3.5 -> 3.5
- 76008ac92..646f419cd  3.4 -> 3.4
- da8474809..057c61eb6  3.3 -> 3.3
- d259e76af..af293435c  3.2 -> 3.2
- f0db81747..c200d392e  3.1 -> 3.1

PR #2114, opened 2023-03-30
Remove redundant escaping

ekohl:
The URL is already part of a code block, so escaping is not needed. It now renders literally.

Introduced in 1ea4428abe25485eb2285b45a9ef1b7a246f84ff

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
+++ b/guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
@@ -74,5 +74,5 @@ If the file exists, run the second `{foreman-installer}` command that updates ce
 +
 IMPORTANT: Do not delete the certificate archive file after you deploy the certificate.
 It is required, for example, when upgrading {ProjectServer}.
-. On a computer with network access to {ProjectServer}, navigate to the following URL: `\https://{foreman-example-com}`.
+. On a computer with network access to {ProjectServer}, navigate to the following URL: `https://{foreman-example-com}`.
 . In your browser, view the certificate details to verify the deployed certificate.


ekohl:
Ok, so this is weird, because on the Red Hat site it is shown as escaped. I suppose that's a different rendering engine.

You can see it on https://access.redhat.com/documentation/en-us/red_hat_satellite/6.10/html/installing_satellite_server_from_a_connected_network/performing-additional-configuration#deploying-a-custom-ssl-certificate-to-satellite-server_satellite and I see https://access.redhat.com/documentation/en-us/red_hat_satellite/6.10/html/installing_satellite_server_from_a_connected_network/performing-additional-configuration#adding-a-default-http-proxy_satellite has another example.

cc @Lennonka @adamlazik1

maximiliankolb:
I'd appreciate if you had a look into making this work on Satellite docs/its framework. If it doesn't, then I'd also accept escaping the link for all build targets because it's really only a minor visual issue.

Lennonka:
@ekohl I think we use an older version of AsiiDoctor on our side. I would personally push for upgrading it, but first it has to be tested by all docs teams, which is a long shot.

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
```diff
@@ -74,5 +74,5 @@ If the file exists, run the second `{foreman-installer}` command that updates ce
 +
 IMPORTANT: Do not delete the certificate archive file after you deploy the certificate.
 It is required, for example, when upgrading {ProjectServer}.
-. On a computer with network access to {ProjectServer}, navigate to the following URL: `\https://{foreman-example-com}`.
+. On a computer with network access to {ProjectServer}, navigate to the following URL: `https://{foreman-example-com}`.
```
adamlazik1, line 77: Would something like this solve our issue?

```suggestion
. On a computer with network access to {ProjectServer}, navigate to the following URL:
ifdef::satellite[]
`https://{foreman-example-com}`.
endif::[]
ifndef::satellite[]
`\https://{foreman-example-com}`.
endif::[]
```

ekohl: Likely, but if you go that route I'd probably use an attribute.

ekohl:
@maximiliankolb can you check what Antora does in this case?

maximiliankolb:
## Antora

````
:foreman-example-com: foreman.example.com

Plain::

* no escape: {foreman-example-com}
* escape: \{foreman-example-com}

HTTPS::

* no escape: https://{foreman-example-com}
* escape: \https://{foreman-example-com}

Code::

----
see https://{foreman-example-com} and https://{foreman-example-com}

also escaped
see \{foreman-example-com} and \https://{foreman-example-com}
----
````

Result
image


## upstream
in upstream docs: https://docs.theforeman.org/nightly/Installing_Server/index-katello.html#Deploying_a_Custom_SSL_Certificate_to_Server_foreman

image

---

I could not verify the issue: https://access.redhat.com/documentation/en-us/red_hat_satellite/6.13/html-single/installing_satellite_server_in_a_connected_network_environment/index#Deploying_a_Custom_SSL_Certificate_to_Server_satellite

ekohl:
Thanks for checking. Perhaps the tech stack was updated in some version between Satellite 6.10 and 6.13, because it is in the 6.10 docs. I'm going to close this.

PR #2155, opened 2023-04-21
Mention parameter type for customizing template

mjivraja:
By default, the parameter value is set to false and vice versa for downstream, the mentioning the of parameter type was missing in the documentation. We added it for both, upstream and downstream along with its impact on insight clients and cloud reports at downstream.

https://bugzilla.redhat.com/show_bug.cgi?id=2167389

* [X] Foreman 3.6/Katello 4.8
* [X] Foreman 3.3/Katello 4.5 on EL7 & EL8 (Satellite 6.12 on EL8 only; planned orcharhino 6.4 on EL8 only)

### DIFF
--- a/guides/common/modules/ref_customizing-the-registration-templates.adoc
+++ b/guides/common/modules/ref_customizing-the-registration-templates.adoc
@@ -21,9 +21,12 @@ You can configure the following global parameters by navigating to *Configure* >
 
 ifndef::satellite[]
 * The `host_registration_insights` parameter is used in the `insights` snippet, the default value is `false`.
+The type must be set to `Boolean` to override the parameter value.
 endif::[]
 ifdef::satellite[]
 * The `host_registration_insights` parameter is used in the `insights` snippet, the default value is `true`.
+The type must be set to `Boolean` to override the parameter value.
+If the parameter is set to `false`, it prevents hosts from appearing in the `rh-cloud reports` and also restricts `insights-client` from uploading its reports.
 endif::[]
 * The `host_packages` parameter is for installing packages on the host.
 * The `remote_execution_ssh_keys`, `remote_execution_ssh_user`, `remote_execution_create_user` and `remote_execution_effective_user_method` parameters are used in the `remote_execution_ssh_keys`. For more details se the details of the snippet.
--- a/guides/common/modules/ref_customizing-the-registration-templates.adoc
+++ b/guides/common/modules/ref_customizing-the-registration-templates.adoc
@@ -21,13 +21,13 @@ You can configure the following global parameters by navigating to *Configure* >
 
 ifndef::satellite[]
 * The `host_registration_insights` parameter is used in the `insights` snippet, the default value is `false`.
-The type must be set to `Boolean` to override the parameter value.
+Set the parameter type to `boolean` to override the parameter value.
 endif::[]
 ifdef::satellite[]
 * The `host_registration_insights` parameter is used in the `insights` snippet, the default value is `true`.
-The type must be set to `Boolean` to override the parameter value.
-If the parameter is set to `false`, it prevents hosts from appearing in the `rh-cloud reports` and also restricts `insights-client` from uploading its reports.
+Set the parameter type to `boolean` to override the parameter value.
 endif::[]
+If the parameter is set to `false`, it prevents {Project} and the Insights client from uploading Inventory reports to your {RHCloud}.
 * The `host_packages` parameter is for installing packages on the host.
 * The `remote_execution_ssh_keys`, `remote_execution_ssh_user`, `remote_execution_create_user` and `remote_execution_effective_user_method` parameters are used in the `remote_execution_ssh_keys`. For more details se the details of the snippet.
 * The `encrypt_grub` parameter is to enable setting of an encrypted bootloader password for the host, the default value is `false`.
--- a/guides/common/attributes-base.adoc
+++ b/guides/common/attributes-base.adoc
@@ -106,6 +106,7 @@
 :ProductVersionPrevious: {ProjectVersionPrevious}
 :provision-script: OS installer recipe
 :EL: Enterprise Linux
+:RHCloud: Red{nbsp}Hat Hybrid Cloud Console
 :RHEL: Red{nbsp}Hat Enterprise Linux
 :RHELServer: {RHEL} Server
 :SLES: SUSE Linux Enterprise Server


guides/common/modules/ref_customizing-the-registration-templates.adoc
```diff
@@ -21,9 +21,12 @@ You can configure the following global parameters by navigating to *Configure* >
 
 ifndef::satellite[]
 * The `host_registration_insights` parameter is used in the `insights` snippet, the default value is `false`.
+The type must be set to `Boolean` to override the parameter value.
```
Lennonka, line 24:
```suggestion
Set the parameter type to `boolean` to override the parameter value.
```

guides/common/modules/ref_customizing-the-registration-templates.adoc
```diff
@@ -21,9 +21,12 @@ You can configure the following global parameters by navigating to *Configure* >
...
 endif::[]
 ifdef::satellite[]
 * The `host_registration_insights` parameter is used in the `insights` snippet, the default value is `true`.
+The type must be set to `Boolean` to override the parameter value.
```
Lennonka, line 28:
```suggestion
Set the parameter type to `boolean` to override the parameter value.
```

guides/common/modules/ref_customizing-the-registration-templates.adoc
```diff
@@ -21,9 +21,12 @@ You can configure the following global parameters by navigating to *Configure* >
...
+If the parameter is set to `false`, it prevents hosts from appearing in the `rh-cloud reports` and also restricts `insights-client` from uploading its reports.
```
Lennonka, line 29:
```suggestion
If the parameter is set to `false`, it prevents hosts from appearing in the Red Hat Cloud Console reports and also the `insights-client` from uploading reports to {Project}.
```
I'm not sure, about Insights client uploading reports to {Project}, verify with a SME.
Also, this sentence is quite important and I think it should be included for both non-Satellite and Satellite alike.

Lennonka (changes requested):
Left you a couple of suggestions.

guides/common/modules/ref_customizing-the-registration-templates.adoc
ShimShtein: I can confirm. The code looks for this parameter and will reject `insights-client` reports if the parameter is set to `false`.

ShimShtein: As for `Red Hat Cloud Console reports`: I think it's a bit vague. There are two types of reports: one generated by insights-client on the host itself, and the other generated by the rh_cloud plugin. Both are contributing to the Red Hat Inventory. While `insights-client` reports are used to generate recommendations, rh_cloud reports are used for inventory and RHEL usage reporting.

cmarinea: The word "restricted" is a bit confusing for few reason. Because the parameter influence the system during its deployment. A user can change the insights-client whenever they want after that. Setting the parameter to "false" just do not enable insights and therefore nothing is sent. But it does not restrict per say. 

insights-client do no send data to Satellite in a way to be consumed by it when the parameter is set to true, it uses Satellite as a proxy. 

This parameter is not something that influence Satellite itself, it influences what kickstart scripts are given to the client system getting deployed via Satellite.

ShimShtein: @cmarinea not only during the deployment. I have specifically added code that would prevent Foreman from forwarding reports from `insights-client` to the cloud: https://github.com/theforeman/foreman_rh_cloud/blob/foreman_3_7/app/controllers/insights_cloud/api/machine_telemetries_controller.rb#L76

Of course you are correct about its influence during the deployment, but it's additional functionality that we have added specifically to the rh_cloud plugin.

cmarinea: @ShimShtein So just to confirm the parameter intended to influence the kickstart template file "host_registration_insights" has signification elsewhere in the code?

If a user enable insights-client later on, the Satellite will not act as a proxy?

For the rh_cloud plugin, I thought it was another configuration completely.  I may be a little confused :)

ShimShtein: > So just to confirm the parameter intended to influence the kickstart template file "host_registration_insights" has signification elsewhere in the code?

The parameter intended to regulate a feature: insights-client reports. This means that it will influence everything insights-client related, be it provisioning or reporting. Imagine a user who is using RHV templates to provision new machines (no provisioning is done through the kickstart process) , if the user sets the parameter to false, it would be rational that insights-client will not try to report to the cloud, even if it's part of the template.

> If a user enable insights-client later on, the Satellite will not act as a proxy?

As far as I know insights-client has a way to discover it consumes content from Satellite, and if it discovers this fact, it will use Satellite as a proxy. I think there is a way to prevent this automatic discovery process and point it directly to the cloud, but it is definitely not the default configuration.

> For the rh_cloud plugin, I thought it was another configuration completely. I may be a little confused :)

I have tried to reduce the amount of knobs and levers in rh_cloud, and if I identified an already existing way of expressing user's idea, I used it in the rh_cloud plugin. There are a couple of configuration options specific to rh_cloud, but those are there only because rh_cloud introduced new features that required its own configuration.

mjivraja: Pushed. I hope we're closer if not good to go. :crossed_fingers:

guides/common/modules/ref_customizing-the-registration-templates.adoc
```diff
@@ -21,9 +21,12 @@ You can configure the following global parameters by navigating to *Configure* >
 
 ifndef::satellite[]
 * The `host_registration_insights` parameter is used in the `insights` snippet, the default value is `false`.
+Set the parameter type to `boolean` to override the parameter value.
 endif::[]
 ifdef::satellite[]
 * The `host_registration_insights` parameter is used in the `insights` snippet, the default value is `true`.
+Set the parameter type to `boolean` to override the parameter value.
+If the parameter is set to `false`, it prevents hosts from appearing in the {Team} Inventory reports and also the `insights-client` from uploading reports to `rh-cloud`.
```
Lennonka, line 29:
```suggestion
If the parameter is set to `false`, it prevents hosts from appearing in the {Team} Inventory reports and also the Insights client from uploading reports to RH Cloud.
```
I'm suggesting a couple of formatting improvements. "RH Cloud" is found in the WebUI, so I think it's vital to spell it like this.
Again, I suggest taking this line out of the condition, because it's relevant for all build targets.

Lennonka (changes requested):
A tiny suggestion.

guides/common/modules/ref_customizing-the-registration-templates.adoc
mjivraja: @Lennonka We are discussing [here](https://github.com/theforeman/foreman-documentation/pull/2180#discussion_r1187306218) the potential possibility of adding attributes for RH Cloud. Should we wait for it or we shall move forward here?

Lennonka: @mjivraja Perhaps we should introduce a new attribute for "Red Hat Cloud" as well in another PR so that we start using the product name consistently? #2180 is about something different if I understand correctly.

mjivraja: In this [PR](https://github.com/theforeman/foreman-documentation/pull/2180#pullrequestreview-1417989076) Maximilian shares sensible data to avoid introducing the new attribute for RH Cloud. 

However, I also agree with your purpose to introduce the attribute for "Red Hat Cloud" to start using the product name consistently.

mjivraja:
@Lennonka Does this look good to you to move further?

guides/common/modules/ref_customizing-the-registration-templates.adoc
Lennonka: @mjivraja Yes, but he's talking about an attribute for all builds and for a program name in the CLI. We can add an attribute for the product title only for the Satellite build.

guides/common/attributes-satellite.adoc
```diff
@@ -92,6 +92,7 @@
 :ProjectX: Satellite{nbsp}6
 :ProjectXY: Satellite{nbsp}{ProductVersionRepoTitle}
 :provision-script: kickstart
+:RHC: {Team} Cloud
```
Lennonka, line 95:
```suggestion
:RHCloud: {Team} Hybrid Cloud Console
```
I'd prefer to go with this if you agree. I'd definitely use the official product name.

Lennonka (changes requested):
I strongly suggest using the proper and official product name of the RH Cloud and a longer attribute name for better clarity.

guides/common/modules/ref_customizing-the-registration-templates.adoc
```diff
@@ -21,9 +21,12 @@ You can configure the following global parameters by navigating to *Configure* >
 
 ifndef::satellite[]
 * The `host_registration_insights` parameter is used in the `insights` snippet, the default value is `false`.
+Set the parameter type to `boolean` to override the parameter value.
 endif::[]
 ifdef::satellite[]
 * The `host_registration_insights` parameter is used in the `insights` snippet, the default value is `true`.
+Set the parameter type to `boolean` to override the parameter value.
+If the parameter is set to `false`, it prevents hosts from appearing in the {Team} Inventory reports and also the Insights client from uploading reports to the Inventory of your {RHCloud}.
```
Lennonka, line 29: This line applies to all builds.

Lennonka (changes requested):
One more detail.

guides/common/modules/ref_customizing-the-registration-templates.adoc
Lennonka: And I think this is the case where you want to use `Red Hat` instead of `{Team}` :)

mjivraja: Indeed :smiley:

maximiliankolb (commented):
one detail, rest LGTM

guides/common/modules/ref_customizing-the-registration-templates.adoc
```diff
@@ -21,10 +21,13 @@ You can configure the following global parameters by navigating to *Configure* >
 
 ifndef::satellite[]
 * The `host_registration_insights` parameter is used in the `insights` snippet, the default value is `false`.
+Set the parameter type to `boolean` to override the parameter value.
 endif::[]
 ifdef::satellite[]
 * The `host_registration_insights` parameter is used in the `insights` snippet, the default value is `true`.
+Set the parameter type to `boolean` to override the parameter value.
 endif::[]
+If the parameter is set to `false`, it prevents hosts from appearing in the Red{nbsp}Hat Inventory reports and also the Insights client from uploading reports to the Inventory of your {RHCloud}.
```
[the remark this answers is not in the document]
Lennonka: Should this say "Red Hat Insights" instead of "Red Hat Inventory reports"? @mjivraja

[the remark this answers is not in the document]
Lennonka: Because when I look into the WebUI, there are the following items under *Configure* > *RH Cloud*: *Inventory Upload* and *Insights*.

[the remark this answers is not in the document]
Lennonka:
```suggestion
If the parameter is set to `false`, it prevents hosts from appearing in the Red{nbsp}Hat Insights and also the Insights client from uploading Inventory reports to your {RHCloud}.
```
Maybe like this? Pls, check this with SMEs.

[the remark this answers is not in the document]
ekohl: @ShimShtein perhaps you can enlighten us on this.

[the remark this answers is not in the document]
mjivraja: I guess, hosts are still part of RH Insights despite the parameter being false. But definitely, let's wait for @ShimShtein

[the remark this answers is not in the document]
ShimShtein: The inventory on the cloud side accepts many types of reports. As long as at least one reporter continues to report about a host, it will appear in the cloud inventory.
Here we are talking about two reporters that are controlled by Satellite: the rh_cloud (Satellite) generated reports and `insights-client` reports. While those are the main contributors to the inventory, they are not the only ones that exist, and that's why I am a bit reluctant about documenting this parameter as something that will prevent the entry in the inventory for sure. We want to indicate that this parameter changes the behavior of those two reports.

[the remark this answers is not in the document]
ShimShtein: Maybe something along the lines of
```
It prevents Satellite and the Insights client from uploading Inventory reports to your {RHCloud}.
```
Does this make sense?

Lennonka (approved):
LGTM, merging

Issue #2154, opened 2023-04-20
Bring back sections about smart proxy


Imaanpreet:
- {SmartProxy} Configuration Tuning (https://github.com/theforeman/foreman-documentation/blob/3.2/guides/common/modules/con_smart-proxy-configuration-tuning.adoc)
- {SmartProxy} Performance Tests (https://github.com/theforeman/foreman-documentation/blob/3.2/guides/common/modules/con_smartproxy_performance_tests.adoc) 

Above mentioned sections are missing after 3.2 release. Is this intentional?  
For example, it is not present in 3.3 tuning [guide](https://docs.theforeman.org/3.3/Performance_Tuning_Guide/index-foreman-el.html#Smart_Proxy_Configuration_Tuning_performance-tuning). 

These sections should be present in 3.3, 3.4 and 3.5 onwards.

Imaanpreet:
Hello @maximiliankolb : can you please look into this? Please let me know what are you thoughts about this.

Imaanpreet:
Thank you @maximiliankolb for looking into this. Could you please help me to resolve this as soon as possible as I want this chapter or section 3.3, 3.4 and 3.5 onwards in this week. *(Satellite downstream documentation is linked with Red Hat Sat 6.13 -> foreman version 3.4 & 3.5)

maximiliankolb:
You can already use these files; they are already part of the source branches for 3.3 and 3.5.

Imaanpreet:
Yes, I can see. I am closing this issue. Thank you.

PR #2161, opened 2023-04-25
Remove unmaintained built docs

maximiliankolb:
We have renamed "Performance_Tuning_Guide" to "Tuning_Performance", but did not delete the old built docs. This causes confusion if users access the old docs, for example using bookmarks.

git rm -r 3.3/Performance_Tuning_Guide/
git rm -r 3.4/Performance_Tuning_Guide/

Solves https://github.com/theforeman/foreman-documentation/issues/2154


adamlazik1 (approved):
Thanks!

Issue #2172, opened 2023-04-27
Configuring DNS Zones in the IdM web UI is wrong


lessfoobar:
"Configuring DNS Zones in the IdM web UI" on the [katello docu](https://docs.theforeman.org/3.6/Installing_Server/index-katello.html#configuring-dynamic-dns-update-with-gss-tsig-authentication_foreman) is telling you to use for the foreward zone:
`grant smart-proxy/047foreman.example.com@EXAMPLE.COM wildcard * ANY;`
which is wrong also explained in the foreman docu why it is wrong. Also the reverse zone snippet is correct. They should be the same.
and the "FreeIPA configuration" on the [foreman manuals](https://theforeman.org/manuals/3.6/index.html#4.3.5DNS) is giving you the correct snippet:
`grant foremanproxy\047proxy.example.com@EXAMPLE.COM wildcard * ANY;`

PR #2175, opened 2023-05-02
Fix Smart Proxy DNS setting

maximiliankolb:
Do not use attributes such as "smart-proxy-context" because the name depends on the service, not a branded product.

Resolves https://github.com/theforeman/foreman-documentation/issues/2172
ping @lessfoobar

* [x] Foreman 3.6/Katello 4.8
* [x] Foreman 3.3/Katello 4.5 on EL7 & EL8 (Satellite 6.12 on EL8 only; planned orcharhino 6.4 on EL8 only)

### DIFF
--- a/guides/common/modules/proc_configuring-dynamic-dns-update-with-gss-tsig-authentication.adoc
+++ b/guides/common/modules/proc_configuring-dynamic-dns-update-with-gss-tsig-authentication.adoc
@@ -39,7 +39,7 @@ endif::[]
 ifeval::["{context}" == "{project-context}"]
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# ipa service-add _{smart-proxy-context}/{foreman-example-com}_
+# ipa service-add _{smart-proxy-principal}/{foreman-example-com}_
 ----
 endif::[]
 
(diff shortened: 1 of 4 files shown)


guides/common/modules/proc_configuring-dynamic-dns-update-with-gss-tsig-authentication.adoc
```diff
@@ -113,7 +113,7 @@ For example, `example.com`.
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-grant {smart-proxy-context}/047__{foreman-example-com}@EXAMPLE.COM__ wildcard * ANY;
+grant foremanproxy\047__{foreman-example-com}@EXAMPLE.COM__ wildcard * ANY;
```
ekohl, line 116: I don't think this is correct in itself. On line 101/102 the same principal is used to initialize it and it's used elsewhere too. I do wonder why there's 047 in here but not in the others.

ekohl (changes requested):
I haven't used GSS, but this looks wrong to me. The principal is created on line 36 and then used. Perhaps it makes more sense to declare an attribute and then consistently use that.

maximiliankolb:
I just looked into attributes but did not find anything that resembles "foremanproxy". Using `{smart-proxy-context}` looks reasonable to me.

How to move this forward?

cc @lessfoobar

ekohl:
> I just looked into attributes but did not find anything that resembles "foremanproxy". Using `{smart-proxy-context}` looks reasonable to me.

The only thing I can think of is that `-` is an invalid character in the principal.

ekohl:
f814915146e6..24098cac540b  3.3 -> 3.3
   266cd1b2eb0f..0adc01249591  3.4 -> 3.4
   9027afaec52a..9c7f26b220c2  3.5 -> 3.5
   3cb4187483e0..f14f14adb955  3.6 -> 3.6
   2caadfa6ac35..23ea2d434f7c  3.7 -> 3.7
   e04297d91af3..686557fc8b0e  3.8 -> 3.8

PR #2210, opened 2023-05-30
Mark optional steps in Host Registration

Lennonka:
All registration fields except for **Activation Key** are optional.
Adding "Optional" for steps that were previously mentioned as mandatory.

https://bugzilla.redhat.com/show_bug.cgi?id=1992283

* [ ] Foreman 3.6/Katello 4.8
* [ ] Foreman 3.3/Katello 4.5 on EL7 & EL8 (Satellite 6.12 on EL8 only; planned orcharhino 6.4 on EL8 only)

### DIFF
--- a/guides/common/modules/proc_registering-hosts.adoc
+++ b/guides/common/modules/proc_registering-hosts.adoc
@@ -85,14 +85,14 @@ The following is an example of the `curl` command with the `--insecure` option:
 # curl -sS --insecure https://{foreman-example-com}/register ...
 ----
 . Select the *Advanced* tab.
-. From the *Setup REX* list, select whether you want to deploy {Project} SSH keys to hosts or not.
+. Optional: From the *Setup REX* list, select whether you want to deploy {Project} SSH keys to hosts or not.
 +
 If set to `Yes`, public SSH keys will be installed on the registered host.
 The inherited value is based on the `host_registration_remote_execution` parameter.
 It can be inherited, for example from a host group, an operating system, or an organization.
 When overridden, the selected value will be stored on host parameter level.
 ifdef::client-content-dnf[]
-. From the *Setup Insights* list, select whether you want to install `insights-client` and register the hosts to Insights.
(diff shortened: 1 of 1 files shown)


guides/common/modules/proc_registering-hosts.adoc
```diff
@@ -85,14 +85,14 @@ The following is an example of the `curl` command with the `--insecure` option:
 # curl -sS --insecure https://{foreman-example-com}/register ...
 ----
 . Select the *Advanced* tab.
-. From the *Setup REX* list, select whether you want to deploy {Project} SSH keys to hosts or not.
+. Optional: From the *Setup REX* list, select whether you want to deploy {Project} SSH keys to hosts or not.
```
maximiliankolb, line 88:
```suggestion
. If you want to deploy {Project} SSH keys to your hosts, select the *Setup REX* checkbox.
```

I don't believe this step is optional. I'd rather say it's conditional depening on your use case.

IMHO only descriptions etc. are truly optional because they don't really affect the outcome.

Lennonka: By optional it's meant that the curl command can be generated regardless of filling in this field. Which it can.

maximiliankolb: But that's an active descision to say "I don't want XY". Does IBM Style Guide say anything about that?

Lennonka: It isn't the same case as with the SmartProxy prerequisite.
"Optional" says: Foreman doesn't require the value for this function to work, it's up to you if you set it or not, and it will affect the function's behavior in this or that way. If the user will leave it as it is, the function (registration in this case) will still work.
In case of the prerequisite, the user had to fulfill it in a certain scenario, otherwise the following procedure wouldn't work.
This is why I would stick with my original proposal.

Lennonka: REX setup is an addition to registration. Regardless of the REX setup, the user will be able to register the host, that's why it should be optional.
The whole registration needs rewriting, I would keep it simple for now and get to rewriting it later.

PR #2215, opened 2023-06-05
don't depend on compass and zurb, we don't use it

evgeni:
but it wants a super old sass which doesn't work on modern Ruby

pull in sass directly instead

Fixes: ac63a51ce4d5e9d46112639a5e2f459ae2fdd43d

* [ ] Foreman 3.6/Katello 4.8
* [ ] Foreman 3.3/Katello 4.5 on EL7 & EL8 (Satellite 6.12 on EL8 only; planned orcharhino 6.4 on EL8 only)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,6 @@ RUN dnf upgrade -y && \
                    rubygem-asciidoctor-pdf && \
     dnf groupinstall -y development-tools
 
-RUN gem install compass:0.12.7 \
-                zurb-foundation:4.3.2
+RUN gem install sass
 
 WORKDIR /foreman-documentation/guides
--- a/Gemfile
+++ b/Gemfile
@@ -5,5 +5,4 @@ gem 'asciidoctor-pdf'
 # missing dependency for asciidoctor-pdf
 gem 'matrix'
 
-gem 'compass', '0.12.7'
-gem 'zurb-foundation', '4.3.2'
+gem 'sass'
--- a/guides/common/Makefile
+++ b/guides/common/Makefile
@@ -45,7 +45,7 @@ endif
 all: html
 
 prepare:
-	@mkdir -p $(BUILD_DIR) $(DEST_DIR) $(IMAGES_DIR) $(DEST_DIR)/sass-cache
+	@mkdir -p $(BUILD_DIR) $(DEST_DIR) $(IMAGES_DIR)
 
 html: prepare $(IMAGES_TS) $(DEST_HTML)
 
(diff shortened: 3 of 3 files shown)


maximiliankolb:
Building the container image works for me. Maybe I should update it to a more recent version of Fedora and start using it again ...

However, I wonder about `Installing_Server/sass-cache/eabc6fbbf6bfab1575a0fda2b03180bdcc22267f/_asciidoc.scssc` etc. It's new, isn't it? Do we have to deliver this to docs.theforeman.org? Or add a step to delete it?

evgeni:
heh, so the `sass-cache` folder is also present in our current deployments: https://github.com/theforeman/foreman-documentation/tree/gh-pages/nightly/Administering_Project/sass-cache/eabc6fbbf6bfab1575a0fda2b03180bdcc22267f

it just never turned up in the diffs as we never changed the `sass` version or the CSS, so the cache remained stable

I guess we should not deploy the sass cache, but it's not a regression per se.

now the question is, do we need the cache at all? poor-evgeni-benchmark on his laptop says (only one run, not statistically significant):

| cache | 3.2.19 | 3.7.4 |
|---------|----------|--------|
| empty | 60.57s user 8.38s system 98% cpu 1:09.95 total |  63.01s user 8.54s system 99% cpu 1:11.89 total | 
| full, only html result deleted - this should not invoke sass at all! | 37.66s user 5.08s system 99% cpu 42.867 total |  37.84s user 5.09s system 99% cpu 43.074 total | 
| full, html and css deleted | 59.00s user 8.49s system 99% cpu 1:07.76 total |  59.87s user 8.49s system 99% cpu 1:08.68 total

not invoking sass at all clearly has a benefit, but the cache doesn't add *a lot* speedup, so we could also just disable it globally and delete the existing files from the deployment? that's easier than having different flags depending on local/remote deployment.

evgeni:
@maximiliankolb I have updated the PR to drop the cache, mind ACKing?

maximiliankolb (approved):
LGTM.

evgeni:
   c20deadb4..bf103ed8e  3.1 -> 3.1
   c3ca80102..4299c815c  3.2 -> 3.2
   e522a3985..85e3b7186  3.3 -> 3.3
   d5d8e4e58..ad2077860  3.4 -> 3.4
   25a4b34b2..dcf0342e8  3.5 -> 3.5
   b8cf81035..c24db18a1  3.6 -> 3.6
   d9cb2f6e9..e4bb1e365  3.7 -> 3.7

PR #2240, opened 2023-06-15
Update dependency solving guidelines

ianballou:

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/con_resolving-package-dependencies.adoc
+++ b/guides/common/modules/con_resolving-package-dependencies.adoc
@@ -1,18 +1,56 @@
 [id="Resolving_Package_Dependencies_{context}"]
 = Resolving Package Dependencies
 
-In {Project}, you can use the package dependency resolution feature to ensure that any dependencies that packages have within a Content View are added to the dependent repository as part of the Content View publication process.
+In {Project}, you can use the package dependency resolution feature, also known as dependency solving, to ensure that any dependencies that packages have within a Content View are added to the dependent repository as part of the Content View publication process.
 
-You can select to resolve package dependencies for any Content View that you want, or you can change the default setting to enable or disable resolving package dependencies for all new Content Views.
+Dependency solving is relevant only in limited contexts.
+In the most common use cases, such as adding a security errata to a Content View with incremental update, dependency solving is likely unnecessary and can cause significant delays to Content View publication.
 
-Note that resolving package dependencies can cause significant delays to Content View promotion.
-The package dependency resolution feature does not consider packages that are installed on your system independently of the Content View or solve dependencies across repositories.
+The package dependency resolution feature does not take into account which packages are installed on your clients.
+Dependency solving considers only packages within the repositories of the Content View.
+ifeval::["{client-os-family}" == "Red Hat"]
+For example, if a Content View has only AppStream in it, dependency solving will not know to pull dependent BaseOS content in at publish time.
+endif::[]
+For more information, see {ContentManagementDocURL}Limitations_to_Repository_Dependency_Resolution_content-management[Limitations to Repository Dependency Resolution] in _{ContentManagementDocTitle}_.
 
 .Resolving Package Dependencies and Filters
-
-Filters do not resolve any dependencies of the packages listed in the filters.
+Filters, by default, do not resolve any dependencies of the packages listed in the filters.
 This might require some level of testing to determine what dependencies are required.
 
 If you add a filter that excludes some packages that are required and the Content View has dependency resolution enabled, {Project} ignores the rules you create in your filter in favor of resolving the package dependency.
 
-If you create a content filter for security purposes, to resolve a package dependency, {Project} can add packages that you might consider insecure.
+If you create a content filter for security purposes but choose to resolve dependencies, {Project} can add packages that you might consider insecure.
+
+.Filtering and Dependency Solving Recommendations
+Be mindful when combining exclusion filters with dependency solving.
+ifeval::["{client-os-family}" == "Red Hat"]
+Consider this example: You recreate {RHEL} 8.3 via Content View filters and include select errata from a later {RHEL} 8 y-release.
+To achieve this, you exclude all errata after the {RHEL} 8.3 release date minus some select few that are needed, and enable dependency solving.
+Dependency solving may pull in more packages than expected, which would cause the client to stray further from being a {RHEL} 8.3 machine.
+endif::[]
+If dependency solving pulls in too many unwanted packages, consider identifying by hand the core basic dependencies that the extra packages and errata need.
+ifeval::["{client-os-family}" == "Red Hat"]
+If, in the above case, the user does not need the extra errata and packages, then Content View filtering is not needed at all.
+You can enable and use the {RHEL} 8.3 repository on the *Content* > *Red Hat Repositories* page in the {ProjectWebUI}.
+endif::[]
+
+ifeval::["{client-os-family}" == "Red Hat"]
+In some cases, excluding packages may make dependency solving impossible for DNF.
+For example, if you make a {RHEL} 8.3 repository with a few excluded packages, `dnf update` could fail.
+Instead of turning on dependency solving to resolve the issue, investigate the error from DNF and adjust the filters to stop excluding the missing dependency.
+If dependency solving is enabled, the repository may stray from {RHEL} 8.3 as discussed above.
+endif::[]
+
+If a Content View version is out of date, packages that are part of a newer erratum may depend on packages that are not within the Content View version.
+Before incrementally adding the erratum with dependency solving, consider publishing a new Content View version instead.
+Incrementally adding the erratum with dependency solving might pull in many new packages, which could alter the intended scope of the Content View version.
+
+If a single package is incrementally added to a Content View version, dependency solving may be needed to install that package.
+Keep in mind that the dependency solving recommendations above also apply to incremental updates.
+
+If dependency solving must be used, consider using more Content Views with fewer repositories and combining them together using composite Content Views.
+Dependency solving time increases significantly with more repositories because every repository in a Content View is examined for dependencies.
+
+.Enabling Dependency Solving
+You can select to resolve package dependencies per Content View or have it enabled by default in *Settings > Content > "content view Dependency Solving Default"*.
+{Team} does not recommend enabling Content View dependency solving by default unless the pitfalls above are well-understood.
--- a/guides/common/modules/proc_adding-errata-to-an-incremental-content-view.adoc
+++ b/guides/common/modules/proc_adding-errata-to-an-incremental-content-view.adoc
@@ -6,6 +6,9 @@ For example, if the Content View is version 1.0, it becomes Content View version
 
 To use the CLI instead of the {ProjectWebUI}, see the xref:cli-adding-errata-to-an-incremental-content-view[].
 
+Be careful when incrementally adding enhancement errata.
+Enhancements are typically scoped to the most current software in a repository, so you may find incompatibilities if your content view version is old.
+
 .Procedure
 . In the {ProjectWebUI}, navigate to *Content* > *Errata*.
 . From the *Errata* list, click the name of the errata that you want to apply.


guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,42 @@
...
-You can select to resolve package dependencies for any Content View that you want, or you can change the default setting to enable or disable resolving package dependencies for all new Content Views.
+You can select to resolve package dependencies per Content View or have it enabled by default via Settings -> Content -> "content view Dependency Solving Default".
+Enabling content view dependency solving by default is unrecommended unless the pitfalls (discussed later) are well-understood.
 
 Note that resolving package dependencies can cause significant delays to Content View promotion.
-The package dependency resolution feature does not consider packages that are installed on your system independently of the Content View or solve dependencies across repositories.
+The more content and repositories in the content view, the greater the delay.
+
+The package dependency resolution feature does not consider packages that are installed on your system independently of the Content View.
+See {ContentManagementDocURL}Limitations_to_Repository_Dependency_Resolution[Limitations to Repository Dependency Resolution] for more information about when dependency resolution may fail.
```
ianballou, line 13: I think my linking here isn't quite right.

ianballou:
@ggainey @dralley here are the CV filtering suggestions that I could think of adding so far. Do you see improvements to what I have?

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,42 @@
...
 
-You can select to resolve package dependencies for any Content View that you want, or you can change the default setting to enable or disable resolving package dependencies for all new Content Views.
+You can select to resolve package dependencies per Content View or have it enabled by default via Settings -> Content -> "content view Dependency Solving Default".
+Enabling content view dependency solving by default is unrecommended unless the pitfalls (discussed later) are well-understood.
 
 Note that resolving package dependencies can cause significant delays to Content View promotion.
-The package dependency resolution feature does not consider packages that are installed on your system independently of the Content View or solve dependencies across repositories.
+The more content and repositories in the content view, the greater the delay.
+
+The package dependency resolution feature does not consider packages that are installed on your system independently of the Content View.
```
jeremylenz, line 12:
```suggestion
When pulling in dependencies, the package dependency resolution feature does not consider packages from repositories outside the content view.
It also does not consider packages installed on your {ProjectServer} that are not managed by Satellite.
```

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,42 @@
...
 This might require some level of testing to determine what dependencies are required.
 
 If you add a filter that excludes some packages that are required and the Content View has dependency resolution enabled, {Project} ignores the rules you create in your filter in favor of resolving the package dependency.
 
-If you create a content filter for security purposes, to resolve a package dependency, {Project} can add packages that you might consider insecure.
+If you create a content filter for security purposes but choose to resolve dependencies, {Project} can add packages that you might consider insecure.
+
+.Filtering and Dependency Solving Recommendations
+
+Be mindful when combining exclusion filters with dependency solving.
```
jeremylenz, line 26:
```suggestion
Be mindful when combining filters with dependency solving.
```

Isn't an include filter really an "exclude everything except" filter?

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,44 @@
 
 In {Project}, you can use the package dependency resolution feature to ensure that any dependencies that packages have within a Content View are added to the dependent repository as part of the Content View publication process.
 
-You can select to resolve package dependencies for any Content View that you want, or you can change the default setting to enable or disable resolving package dependencies for all new Content Views.
+You can select to resolve package dependencies per Content View or have it enabled by default via Settings -> Content -> "content view Dependency Solving Default".
+Enabling content view dependency solving by default is unrecommended unless the pitfalls (discussed later) are well-understood.
```
dralley, line 7: unrecommended -> not recommended?

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,44 @@
...
 
 Note that resolving package dependencies can cause significant delays to Content View promotion.
-The package dependency resolution feature does not consider packages that are installed on your system independently of the Content View or solve dependencies across repositories.
+The more content and repositories in the content view, the greater the delay.
```
dralley, line 10: It's a bit weird to say pitfalls will be discussed later and then mention a pitfall in the very next sentence

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,44 @@
...
+Dependency solving may pull in more RPMs than expected, which would cause the system to stray further from being a RHEL 8.3 machine.
+If dependency solving pulls in too many unwanted RPMs, consider identifying by hand the core basic dependencies that the extra packages and errata need.
+
+If, in the above case, the user does not need the extra errata and RPMs, then content view filtering is not needed at all.
+The RHEL 8.3 repository could be enabled and used directly from the Red Hat Repositories page.
+
+In some cases, excluding RPMs may make dependency solving impossible.
+For example, if a user wants to make a RHEL 8.3 repo with a few RPMs missing, dependency solving should not be used because the missing RPMs could be dependencies for another RPMs in the repository.
+
+Dependency solving is most useful for cases like incremental update, where a critical security patch needs to be added to a content view version with all of its dependent packages.
```
dralley, line 40: I'm not sure I agree with this statement.  Dependency solving is most useful for papering over problems caused by incorrect security metadata.  In nearly all cases, a critical security patch isn't going to require any dependency changes.

guides/common/modules/con_resolving-package-dependencies.adoc
maximiliankolb: `{Team} does not recommend to enable ...`

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,44 @@
...
+Enabling content view dependency solving by default is unrecommended unless the pitfalls (discussed later) are well-understood.
 
 Note that resolving package dependencies can cause significant delays to Content View promotion.
-The package dependency resolution feature does not consider packages that are installed on your system independently of the Content View or solve dependencies across repositories.
+The more content and repositories in the content view, the greater the delay.
+
+The package dependency resolution feature does not take which packages are installed on your systems into account.
+Only packages within the content view's repositories are considered.
+For example, if a content view has only AppStream in it, dependency solving will not know to pull dependent BaseOS content in at publish time.
+See {ContentManagementDocURL}limitations-to-repository-dependency-resolution[Limitations to Repository Dependency Resolution] for more information about when dependency resolution may fail.
```
maximiliankolb, line 15:
```suggestion
For more information, see {ContentManagementDocURL}Limitations_to_Repository_Dependency_Resolution_content-management[Limitations to Repository Dependency Resolution] in _{ContentManagementDocTitle}_.
```

guides/common/modules/con_resolving-package-dependencies.adoc
ianballou: I was thinking more conceptually. Dependency solving + exclusion filters can mean the exclusions will not be respected.

guides/common/modules/con_resolving-package-dependencies.adoc
ianballou: @dralley if a user has an out-of-date RHEL 8 content view version and needs to add a new erratum, it's not possible that the erratum would require package updates outside of its package list?

ianballou: On the line below I have "While an erratum should include all related dependencies needed to apply a critical patch, there may be missing dependencies if content view filters are applied."

Do you feel that this isn't correct?

ianballou: I suppose if an erratum is an update for `packageA` and that package requires `packageB` and `packageC`, and it's applicable to a host, `packageA` must be in the host's content view version with its B and C dependencies already.

As in, there is no situation where content view filtering would not include B and C but include A and have A be installable.  And if A needs an update via an erratum, B and C will be included if they were updates as well.

ianballou: I think I may just take lines 40 and 41 out. I'll keep line 43 because that's package incremental update, which dependency solving could actually be useful for.

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,48 @@
...
+Dependency solving may pull in more RPMs than expected, which would cause the system to stray further from being a RHEL 8.3 machine.
+If dependency solving pulls in too many unwanted RPMs, consider identifying by hand the core basic dependencies that the extra packages and errata need.
+
+If, in the above case, the user does not need the extra errata and RPMs, then content view filtering is not needed at all.
+The RHEL 8.3 repository could be enabled and used directly from the Red Hat Repositories page.
+
+In some cases, excluding RPMs may make dependency solving impossible.
+For example, if a user wants to make a RHEL 8.3 repo with a few RPMs missing, dependency solving should not be used because the missing RPMs could be dependencies for another RPMs in the repository.
+
+If an erratum is being incrementally added to a recently published content view version, dependency solving is likely unneeded.
```
ianballou, line 40: @dralley I reworked the incremental update suggestions here.

guides/common/modules/con_resolving-package-dependencies.adoc
dralley: >@dralley if a user has an out-of-date RHEL 8 content view version and needs to add a new erratum, it's not possible that the erratum would require package updates outside of its package list? And I'm asking aside from incorrect security metadata.

I'm saying it is possible, not even because the errata metadata is wrong but because the RPM metadata is wrong, and doesn't tell the dependency solver that it needs some version X.  If there was a mistake there (which would be really easy to make), nobody running up-to-date RHEL would necessarily ever notice, because getting the latest would paper over it.

I think in practice this probably wouldn't happen often unless you start cherrypicking feature errata.

>On the line below I have "While an erratum should include all related dependencies needed to apply a critical patch, there may be missing dependencies if content view filters are applied."

>Do you feel that this isn't correct?

It's partially correct.  The erratum only lists the packages that fix, or are built alongside the package that fixes the errata. The "dependencies" otherwise are just normal RPM dependencies, they aren't listed in the erratum.

>Edit: on second thought, this is assuming the content view version isn't super old -- what if the system is behind a few y-versions and a package gained some new dependencies? I suppose dependency solving might pull so many packages in from the Library repository that the user should likely just publish a new content view version.

Yeah, this goes back to cherrypicking feature errata being a bad idea.  I don't think security errata are likely to ever pull in entire new dependencies, but cherrypicking feature errata is tempting fate.  At that point you should really just update the CV.

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,48 @@
 
 In {Project}, you can use the package dependency resolution feature to ensure that any dependencies that packages have within a Content View are added to the dependent repository as part of the Content View publication process.
 
-You can select to resolve package dependencies for any Content View that you want, or you can change the default setting to enable or disable resolving package dependencies for all new Content Views.
+You can select to resolve package dependencies per Content View or have it enabled by default via Settings -> Content -> "content view Dependency Solving Default".
```
mjivraja, line 6:
```suggestion
You can select to resolve package dependencies per Content View or have it enabled by default via *Settings > Content > "content view Dependency Solving Default" *.
```

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,48 @@
...
+For example, take a scenario where a user wants to recreate RHEL 8.3 via content view filters and include select packages and errata from a later RHEL 8 y-release.
+The user excludes all errata after the RHEL 8.3 release date minus some select few that are needed and enables dependency solving.
+Dependency solving may pull in more RPMs than expected, which would cause the system to stray further from being a RHEL 8.3 machine.
+If dependency solving pulls in too many unwanted RPMs, consider identifying by hand the core basic dependencies that the extra packages and errata need.
+
+If, in the above case, the user does not need the extra errata and RPMs, then content view filtering is not needed at all.
+The RHEL 8.3 repository could be enabled and used directly from the Red Hat Repositories page.
+
+In some cases, excluding RPMs may make dependency solving impossible.
+For example, if a user wants to make a RHEL 8.3 repo with a few RPMs missing, dependency solving should not be used because the missing RPMs could be dependencies for another RPMs in the repository.
```
mjivraja, line 38:
```suggestion
For example, if a user wants to make a RHEL 8.3 repo with a few RPMs missing, dependency solving should not be used because the missing RPMs could be dependencies for other RPMs in the repository.
```

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,48 @@
...
+
+If an erratum is being incrementally added to a recently published content view version, dependency solving is likely unneeded.
+
+If the content view version is out-of-date, a newer erratum's packages may rely on packages not within the content view version.
```
mjivraja, line 42:
```suggestion
If the content view version is out-of-date, newer erratum's packages may rely on packages that are not within the content view version.
```

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,48 @@
...
+Before incrementally adding the erratum with dependency solving, consider publishing a new content view version instead.
+Incrementally adding the erratum with dependency solving may pull in more packages than desired.
```
mjivraja, line 44:
```suggestion
Incrementally adding the erratum with dependency solving may pull in more packages than required.
```
Or something similar to that. We do not use `desired`.

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,48 @@
...
+
+If a single RPM is incrementally added to a content view version, dependency solving may be needed to install that RPM.
+Keep in mind that the dependency solving recommendations above also apply to incremental update.
```
mjivraja, line 47:
```suggestion
Keep in mind that the dependency solving recommendations above also apply to incremental updates.
```

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,48 @@
...
+
+If dependency solving must be used, consider using more content views with fewer repositories and combining them together using composite content views.
+Dependency solving time increases significantly with more repositories because every repository in a content view is cross-checked for dependencies.
```
mjivraja, line 50:
```suggestion
Dependency solving time increases significantly with more repositories because every repository in a content view is verified for dependencies.
```

guides/common/modules/con_resolving-package-dependencies.adoc
maximiliankolb: Please also replace "via" with something like "in".

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,48 @@
...
+You can select to resolve package dependencies per Content View or have it enabled by default via Settings -> Content -> "content view Dependency Solving Default".
+{Team} does not recommend enabling content view dependency solving by default unless the pitfalls are well-understood.
 
-Note that resolving package dependencies can cause significant delays to Content View promotion.
-The package dependency resolution feature does not consider packages that are installed on your system independently of the Content View or solve dependencies across repositories.
+Note that resolving package dependencies can cause significant delays to Content View publication.
+The more content and repositories in the content view, the greater the delay.
+
+The package dependency resolution feature does not take which packages are installed on your systems into account.
+Only packages within the content view's repositories are considered.
```
maximiliankolb, line 13:
```suggestion
Dependency resolution considers only packages within the repositories of the content view.
```

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,48 @@
...
+For example, if a content view has only AppStream in it, dependency solving will not know to pull dependent BaseOS content in at publish time.
```
maximiliankolb, line 14:
```suggestion
ifeval::["{client-os-family}" == "Red Hat"]
For example, if a content view has only AppStream in it, dependency solving will not know to pull dependent BaseOS content in at publish time.
endif::[]
```

This example depends on both the content type (Yum) and OS of the managed host (EL). See https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-base.adoc#L134

Its set for all build targets but allows ATIX in downstream to overwrite it for OS-specific guides; i.e. SLES and Debian/Ubuntu would not should this line in downstream.

guides/common/modules/con_resolving-package-dependencies.adoc
ianballou: Required isn't right here because the packages are indeed required.

guides/common/modules/con_resolving-package-dependencies.adoc
ianballou: I have some very specific RHEL references later on, should those also be locked into the client os family of "Red Hat" ?

ianballou:
I think I've made all the necessary changes now.

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,50 @@
...
-You can select to resolve package dependencies for any Content View that you want, or you can change the default setting to enable or disable resolving package dependencies for all new Content Views.
+You can select to resolve package dependencies per Content View or have it enabled by default in *Settings > Content > "content view Dependency Solving Default"*.
+{Team} does not recommend enabling content view dependency solving by default unless the pitfalls are well-understood.
 
-Note that resolving package dependencies can cause significant delays to Content View promotion.
-The package dependency resolution feature does not consider packages that are installed on your system independently of the Content View or solve dependencies across repositories.
+Note that resolving package dependencies can cause significant delays to Content View publication.
+The more content and repositories in the content view, the greater the delay.
+
+The package dependency resolution feature does not take which packages are installed on your systems into account.
```
jeremylenz, line 12:
```suggestion
The package dependency resolution feature does not take into account which packages are installed on your systems.
```

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,50 @@
...
+
+If, in the above case, the user does not need the extra errata and RPMs, then content view filtering is not needed at all.
+The RHEL 8.3 repository could be enabled and used directly from the Red Hat Repositories page.
+
+In some cases, excluding RPMs may make dependency solving impossible.
+For example, if a user wants to make a RHEL 8.3 repo with a few RPMs missing, dependency solving should not be used because the missing RPMs could be dependencies for other RPMs in the repository.
+
+If an erratum is being incrementally added to a recently published content view version, dependency solving is likely unneeded.
+
+If the content view version is out-of-date, newer erratas' packages may rely on packages that are not within the content view version.
```
jeremylenz, line 44: `erratas'` is a double plural. Let's just rephrase it.

```suggestion
If the content view version is out of date, packages that are part of a newer erratum may depend on packages that are not within the content view version.
```

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,50 @@
...
+For example, take a scenario where a user wants to recreate RHEL 8.3 via content view filters and include select packages and errata from a later RHEL 8 y-release.
+The user excludes all errata after the RHEL 8.3 release date minus some select few that are needed and enables dependency solving.
+Dependency solving may pull in more RPMs than expected, which would cause the system to stray further from being a RHEL 8.3 machine.
+If dependency solving pulls in too many unwanted RPMs, consider identifying by hand the core basic dependencies that the extra packages and errata need.
+
+If, in the above case, the user does not need the extra errata and RPMs, then content view filtering is not needed at all.
+The RHEL 8.3 repository could be enabled and used directly from the Red Hat Repositories page.
+
+In some cases, excluding RPMs may make dependency solving impossible.
+For example, if a user wants to make a RHEL 8.3 repo with a few RPMs missing, dependency solving should not be used because the missing RPMs could be dependencies for other RPMs in the repository.
```
jeremylenz, line 40: Is there a way this could be written without being RHEL-specific? if not, should add the ifdef:: or whatever :)

guides/common/modules/con_resolving-package-dependencies.adoc
jeremylenz: `intended`?

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,50 @@
...
+Be mindful when combining exclusion filters with dependency solving.
+For example, take a scenario where a user wants to recreate RHEL 8.3 via content view filters and include select packages and errata from a later RHEL 8 y-release.
+The user excludes all errata after the RHEL 8.3 release date minus some select few that are needed and enables dependency solving.
+Dependency solving may pull in more RPMs than expected, which would cause the system to stray further from being a RHEL 8.3 machine.
+If dependency solving pulls in too many unwanted RPMs, consider identifying by hand the core basic dependencies that the extra packages and errata need.
+
+If, in the above case, the user does not need the extra errata and RPMs, then content view filtering is not needed at all.
+The RHEL 8.3 repository could be enabled and used directly from the Red Hat Repositories page.
+
+In some cases, excluding RPMs may make dependency solving impossible.
```
dralley, line 39: If dependency solving ignores the exclude filter, what makes this true?  Or do you mean, it makes dependency solving impossible *for DNF*.

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,50 @@
 
 In {Project}, you can use the package dependency resolution feature to ensure that any dependencies that packages have within a Content View are added to the dependent repository as part of the Content View publication process.
 
-You can select to resolve package dependencies for any Content View that you want, or you can change the default setting to enable or disable resolving package dependencies for all new Content Views.
+You can select to resolve package dependencies per Content View or have it enabled by default in *Settings > Content > "content view Dependency Solving Default"*.
```
dralley, line 6: I would actually move line 42 up here.  Something along the lines of 

> This is relevant only in limited contexts.  In the most common use cases, such as adding a security errata to a content view with incremental update, dependency solving is likely unnecessary and can cause significant delays to Content View publication.

I think the steps to enable or disable it might make sense to be the very last thing discussed, only after the description of what it does, what is the purpose and what are the downsides.

I think line 10 and line 52 are also slightly duplicated.

guides/common/modules/con_resolving-package-dependencies.adoc
maximiliankolb: Yes, makes sense to me.

guides/common/modules/con_resolving-package-dependencies.adoc
ianballou: Yeah that's what I was going for, the messaging got a bit blurred here.

ianballou:
Appreciate the feedback, I think it's reading a lot better now. I moved the "how to enable" part to the end to encourage the reader to scan through the warnings above. Also got rid of some duplicate messaging.

dralley:
Should we give an explicit warning against trying to cherrypick feature errata?

ianballou:
> Should we give an explicit warning against trying to cherrypick feature errata?

@dralley is the idea that feature errata may have more updated dependencies that would likely be incompatible with older repository versions?

ggainey:
> > Should we give an explicit warning against trying to cherrypick feature errata?
> 
> @dralley is the idea that feature errata may have more updated dependencies that would likely be incompatible with older repository versions?

One thought is that features are built on top of "everything in the current release", and can "assume" behavior of things they're not explicitly dependent on. I know I need "foo" - what I didn't realize is that there's a bug in foo-1.2 that would prevent me from working, because I did my building/testing on foo-1.3 and it never came up, so my feature-requires says "foo" and not "foo-1.3". Pull in the feature-advisory into an "old" repo, and the dep-check says "we already have 'foo" we're good". And then the feature doesn't work, and everyone Has A Sad.

security-errata fixes tend to be much more tightly-focused. Note that *this can still happen* with them - it's just "less likely".

(The net of *all* of this continues to be "single-fix is not supported by modern software development methods, no matter how much we wish it were". Best-effort is the best we can do.)

dralley:
>@dralley is the idea that feature errata may have more updated dependencies that would likely be incompatible with older repository versions?

Kinda, but more along the lines of just nudging users into saner workflows in general.  That's one specific example of what is more likely to go wrong but there could be others.

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -1,18 +1,63 @@
 [id="Resolving_Package_Dependencies_{context}"]
 = Resolving Package Dependencies
 
+.Introduction
+
```
maximiliankolb, line 5:
```suggestion
```

IMHO not necessary because a) all concepts should start with an introduction and b) there's no text between this heading and the level 1 heading above.

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -1,18 +1,63 @@
...
 In {Project}, you can use the package dependency resolution feature to ensure that any dependencies that packages have within a Content View are added to the dependent repository as part of the Content View publication process.
 
-You can select to resolve package dependencies for any Content View that you want, or you can change the default setting to enable or disable resolving package dependencies for all new Content Views.
+Dependency solving is relevant only in limited contexts.
+In the most common use cases, such as adding a security errata to a content view with incremental update, dependency solving is likely unnecessary and can cause significant delays to Content View publication.
 
-Note that resolving package dependencies can cause significant delays to Content View promotion.
-The package dependency resolution feature does not consider packages that are installed on your system independently of the Content View or solve dependencies across repositories.
+The package dependency resolution feature does not take into account which packages are installed on your systems.
```
maximiliankolb, line 11:
```suggestion
The package dependency resolution feature does not take into account which packages are installed on your clients.
```

Or maybe "content hosts"?

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -1,18 +1,63 @@
...
+Filters, by default, do not resolve any dependencies of the packages listed in the filters.
 This might require some level of testing to determine what dependencies are required.
 
 If you add a filter that excludes some packages that are required and the Content View has dependency resolution enabled, {Project} ignores the rules you create in your filter in favor of resolving the package dependency.
 
-If you create a content filter for security purposes, to resolve a package dependency, {Project} can add packages that you might consider insecure.
+If you create a content filter for security purposes but choose to resolve dependencies, {Project} can add packages that you might consider insecure.
+
+.Filtering and Dependency Solving Recommendations
+
```
maximiliankolb, line 28:
```suggestion
```

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -1,18 +1,63 @@
...
+Be mindful when combining exclusion filters with dependency solving.
+ifeval::["{client-os-family}" == "Red Hat"]
+For example, take a scenario where a user wants to recreate RHEL 8.3 via content view filters and include select packages and errata from a later RHEL 8 y-release.
+The user excludes all errata after the RHEL 8.3 release date minus some select few that are needed and enables dependency solving.
+Dependency solving may pull in more RPMs than expected, which would cause the system to stray further from being a RHEL 8.3 machine.
```
maximiliankolb, line 33:
```suggestion
Dependency solving may pull in more packages than expected, which would cause the system to stray further from being a RHEL 8.3 machine.
```

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -1,18 +1,63 @@
...
+endif::[]
+If dependency solving pulls in too many unwanted RPMs, consider identifying by hand the core basic dependencies that the extra packages and errata need.
+
+
```
maximiliankolb, line 37:
```suggestion
```

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -1,18 +1,63 @@
...
+.Filtering and Dependency Solving Recommendations
+
+Be mindful when combining exclusion filters with dependency solving.
+ifeval::["{client-os-family}" == "Red Hat"]
+For example, take a scenario where a user wants to recreate RHEL 8.3 via content view filters and include select packages and errata from a later RHEL 8 y-release.
+The user excludes all errata after the RHEL 8.3 release date minus some select few that are needed and enables dependency solving.
+Dependency solving may pull in more RPMs than expected, which would cause the system to stray further from being a RHEL 8.3 machine.
+endif::[]
+If dependency solving pulls in too many unwanted RPMs, consider identifying by hand the core basic dependencies that the extra packages and errata need.
```
maximiliankolb, line 35:
```suggestion
If dependency solving pulls in too many unwanted packages, consider identifying by hand the core basic dependencies that the extra packages and errata need.
```

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -1,18 +1,63 @@
...
+
+
+ifeval::["{client-os-family}" == "Red Hat"]
+If, in the above case, the user does not need the extra errata and RPMs, then content view filtering is not needed at all.
+The RHEL 8.3 repository could be enabled and used directly from the Red Hat Repositories page.
```
maximiliankolb, line 40:
```suggestion
Instead, you can enable and use the RHEL 8.3 repository on *Content* > *Red Hat Repositories* in the {ProjectWebUI}.
```

Maybe drop the "instead"?

What do you think @adamlazik1 ?

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -1,18 +1,63 @@
...
+endif::[]
+
+In some cases, excluding RPMs may make dependency solving impossible for DNF.
+ifeval::["{client-os-family}" == "Red Hat"]
```
maximiliankolb, line 44:
```suggestion
ifeval::["{client-os-family}" == "Red Hat"]
In some cases, excluding RPMs may make dependency solving impossible for DNF.
```

The next best thing would be `client-pkg-ext` (https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-base.adoc#L137) for this line, but just switching the order is IMHO good enough for now.

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -1,18 +1,63 @@
...
+For example, if a user wants to make a RHEL 8.3 repository with a few excluded RPMs, `dnf update` could fail.
```
maximiliankolb, line 45: Please reword from "if a user wants to ..." to "You can/If you want to ...".

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -1,18 +1,63 @@
...
+Incrementally adding the erratum with dependency solving might pull in many new packages, which could alter the intended scope of the content view version.
+
+If a single RPM is incrementally added to a content view version, dependency solving may be needed to install that RPM.
+Keep in mind that the dependency solving recommendations above also apply to incremental updates.
+
+If dependency solving must be used, consider using more content views with fewer repositories and combining them together using composite content views.
+Dependency solving time increases significantly with more repositories because every repository in a content view is examined for dependencies.
+
+.Enabling Dependency Solving
+
```
maximiliankolb, line 61:
```suggestion
```

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -1,18 +1,63 @@
...
+For example, if a user wants to make a RHEL 8.3 repository with a few excluded RPMs, `dnf update` could fail.
+Instead of turning on dependency solving to resolve the issue, investigate the error from DNF and adjust the filters to stop excluding the missing dependency.
+If dependency solving is enabled, the repository may stray from RHEL 8.3 as discussed above.
+endif::[]
+
+If a content view version is out of date, packages that are part of a newer erratum may depend on packages that are not within the content view version.
+Before incrementally adding the erratum with dependency solving, consider publishing a new content view version instead.
+Incrementally adding the erratum with dependency solving might pull in many new packages, which could alter the intended scope of the content view version.
+
+If a single RPM is incrementally added to a content view version, dependency solving may be needed to install that RPM.
```
maximiliankolb, line 54:
```suggestion
If a single package is incrementally added to a content view version, dependency solving may be needed to install that package.
```

maximiliankolb (commented):
Made some more suggestions, rest LGTM.

guides/common/modules/con_resolving-package-dependencies.adoc
adamlazik1: I'd probably word it like this:

```suggestion
You can enable and use the RHEL 8.3 repository on the *Content* > *Red Hat Repositories* page in the {ProjectWebUI}.
```

However, does upstream/orcharhino have this page too?

maximiliankolb: Yes, this applies to all build targets (with Katello).

guides/common/modules/con_resolving-package-dependencies.adoc
ianballou: Just here? Or below every heading?

guides/common/modules/con_resolving-package-dependencies.adoc
ianballou: I wanted to avoid using "you" since I mentioned "the user" above, so I'll simplify it a little further.

guides/common/modules/con_resolving-package-dependencies.adoc
jeremylenz: I feel like we've been moving away from the term "content hosts" lately in favor of just "hosts."

ianballou:
I've fixed the latest requests. Also, @dralley + @ggainey I added a warning about enhancement errata.

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,55 @@
 
 In {Project}, you can use the package dependency resolution feature to ensure that any dependencies that packages have within a Content View are added to the dependent repository as part of the Content View publication process.
```
jeremylenz, line 4:
```suggestion
In {Project}, you can use the package dependency resolution feature, also known as dependency solving, to ensure that any dependencies that packages have within a Content View are added to the dependent repository as part of the Content View publication process.
```

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,55 @@
...
 
-You can select to resolve package dependencies for any Content View that you want, or you can change the default setting to enable or disable resolving package dependencies for all new Content Views.
+Dependency solving is relevant only in limited contexts.
+In the most common use cases, such as adding a security errata to a content view with incremental update, dependency solving is likely unnecessary and can cause significant delays to Content View publication.
 
-Note that resolving package dependencies can cause significant delays to Content View promotion.
-The package dependency resolution feature does not consider packages that are installed on your system independently of the Content View or solve dependencies across repositories.
+The package dependency resolution feature does not take into account which packages are installed on your clients.
+Dependency resolution considers only packages within the repositories of the content view.
```
jeremylenz, line 10: Let's keep it to two names for the same thing - "the package dependency resolution feature" or "dependency solving" for short.

```suggestion
Dependency solving considers only packages within the repositories of the content view.
```

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -3,16 +3,55 @@
...
 If you add a filter that excludes some packages that are required and the Content View has dependency resolution enabled, {Project} ignores the rules you create in your filter in favor of resolving the package dependency.
 
-If you create a content filter for security purposes, to resolve a package dependency, {Project} can add packages that you might consider insecure.
+If you create a content filter for security purposes but choose to resolve dependencies, {Project} can add packages that you might consider insecure.
+
+.Filtering and Dependency Solving Recommendations
+Be mindful when combining exclusion filters with dependency solving.
+ifeval::["{client-os-family}" == "Red Hat"]
+For example, take a scenario where a user wants to recreate RHEL 8.3 via content view filters and include select packages and errata from a later RHEL 8 y-release.
+The user excludes all errata after the RHEL 8.3 release date minus some select few that are needed and enables dependency solving.
```
jeremylenz, line 28:
```suggestion
The user excludes all errata after the RHEL 8.3 release date minus some select few that are needed, and enables dependency solving.
```

mjivraja:
@ianballou A minor view: You can keep the commit message title under 50 character. Provided, you can also add the further description to commit message for explanation (72 character per line).

bangelic:
@maximiliankolb Does this look good to be merged?

asteflova:
I'm totally new to this PR (hello everyone!), but after taking a brief look, I think the section could use some editing (to conform with style guides) and modularizing. I could do it if nobody objects.

What would be the right way to do it? Use the code suggestion feature within this PR? Or merge this PR and then create another one to do the editing?

mjivraja:
> I'm totally new to this PR (hello everyone!), but after taking a brief look, I think the section could use some editing (to conform with style guides) and modularizing. I could do it if nobody objects.
> 
> What would be the right way to do it? Use the code suggestion feature within this PR? Or merge this PR and then create another one to do the editing?

If it's not major, you can do it here itself.

guides/common/modules/con_resolving-package-dependencies.adoc
```diff
@@ -1,18 +1,56 @@
...
 
 If you add a filter that excludes some packages that are required and the Content View has dependency resolution enabled, {Project} ignores the rules you create in your filter in favor of resolving the package dependency.
 
-If you create a content filter for security purposes, to resolve a package dependency, {Project} can add packages that you might consider insecure.
+If you create a content filter for security purposes but choose to resolve dependencies, {Project} can add packages that you might consider insecure.
+
+.Filtering and Dependency Solving Recommendations
+Be mindful when combining exclusion filters with dependency solving.
+ifeval::["{client-os-family}" == "Red Hat"]
+For example, take a scenario where a user wants to recreate RHEL 8.3 via Content View filters and include select packages and errata from a later RHEL 8 y-release.
```
maximiliankolb, line 27: This is passive voice. Maybe "If you want to recreate ..."?

Also, please use "{RHEL}" instead of "RHEL".

guides/common/modules/proc_adding-errata-to-an-incremental-content-view.adoc
```diff
@@ -6,6 +6,10 @@ For example, if the Content View is version 1.0, it becomes Content View version
 
 To use the CLI instead of the {ProjectWebUI}, see the xref:cli-adding-errata-to-an-incremental-content-view[].
 
+Be careful when incrementally adding enhancement errata.
+Enhancements are typically scoped to the most current software in a repository, so you may find incompatibilities if your content view version is old.
+
```
maximiliankolb, line 11:
```suggestion
```

maximiliankolb (commented):
Hi @ianballou, I agree with @asteflova that this PR needs some minor language/styling rework. I am _totally OK_ with merging this now (because the change itself is correct) with the committment from Anet to work on a follow-up PR.

Does that work for everyone?

asteflova:
@maximiliankolb @ianballou I'll prioritize it to get it done this week.

I assume it will involve creating new files (modules) in addition to editing, so I'd prefer to work in my own environment. So merging this with me following up with my own PR as soon as possible works for me.

But in any case, this is really great work @ianballou The changes will be just to make the addition consistent with our style guides.

ianballou:
@maximiliankolb I made some changes to address your last two comments.

ianballou:
Can I merge this? Didn't wanna assume to go ahead since there were some last changes I made.

maximiliankolb:
   3192ba640..285aa32a1  3.7 -> 3.7
   cb341b241..fce9a3c00  3.6 -> 3.6
   944275781..696522b2d  3.5 -> 3.5

PR #2233, opened 2023-06-14
Split Updating Guide

mjivraja:
We need to split the updating guide from the Upgrading and Updating Guide. I have kept this limited to Downstream only at this level but definitely we can raise the scope to other builds. This may contain technical modifications, general modifications, and at a modular level so, please don't hesitate to provide even minor feedback.

Please NOTE that this PR only focuses on Updating Guide and not the Upgrading Guide. It will have another PR from @AkshayGadhaveRH. So, please be cautious if you are merging this PR first.

* [X] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [X] Foreman 3.7/Katello 4.9 (planned Satellite 6.14)
* [ ] Foreman 3.6/Katello 4.8
* [ ] Foreman 3.5/Katello 4.7 (Satellite 6.13)
* [ ] Foreman 3.4/Katello 4.6 (EL8 only)
* [ ] Foreman 3.3/Katello 4.5 on EL7 & EL8 (Satellite 6.12 on EL8 only; orcharhino 6.4 on EL8 only)
* [ ] Foreman 3.2/Katello 4.4 on EL7 & EL8
* [ ] Foreman 3.1/Katello 4.3 on EL7 & EL8 (Satellite 6.11 EL7/8; orcharhino 6.3 on EL7/8)
* For Foreman 3.0 or older, please create a separate PR.
* We do not accept PRs for Foreman 2.4 or older.

PR #2244, opened 2023-06-19
Additional arguments in foreman-maintain

mjivraja:
The additional arguments that can be used for foreman-maintain can be listed while updating server to next minor version.

https://bugzilla.redhat.com/show_bug.cgi?id=2189967

* [X] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
+++ b/guides/doc-Upgrading_and_Updating/topics/upgrading_a_connected_satellite_server.adoc
@@ -61,9 +61,9 @@ ifdef::satellite[]
 When prompted, enter the hammer admin user credentials to configure `{foreman-maintain}` with hammer credentials.
 These changes are applied to the `/etc/foreman-maintain/foreman-maintain-hammer.yml` file.
 +
-[options="nowrap" subs="attributes"]
+[options="nowrap" subs="+quotes,attributes"]
 ----
-# {foreman-maintain} upgrade check --target-version {ProductVersion}
+# {foreman-maintain} upgrade check --target-version {ProductVersion} --whitelist _whitelist_arguments_
 ----
 +
 Review the results and address any highlighted error conditions before performing the upgrade.


mjivraja:
The merged changes of this PR shall be applicable to this PR #2233 as well.

adamlazik1:
Should it be explained what whitelist does and what whitelist arguments can be?

mjivraja:
> Should it be explained what whitelist does and what whitelist arguments can be?

IMO, users may be aware of. But, let's wait for other reviews.

maximiliankolb (commented):
I am against recommending users to bypass certain upgrade checks; it is potentially dangerous if users just reword "check" to "run" to perform the upgrade.

But we're currently not using the procedure in downstream. If someone ACKs this, then it's good to go.

ekohl:
Like the others I doubt this is a good direction. A few reasons:

* whitelist is poorly named: it's really a blacklist where it doesn't run checks that were listed (https://github.com/theforeman/foreman_maintain/pull/723 renames it to `--skip`)
* We don't want users to skip checks.
* We don't provide any help on how to find out which checks the user could list

So overall I'd say this is not a direction we should head into. At best you could add a section on how to skip certain checks but it should not be the default user instruction.

mjivraja:
> Like the others I doubt this is a good direction. A few reasons:
> 
> * whitelist is poorly named: it's really a blacklist where it doesn't run checks that were listed ([Change --whitelist to --skip foreman_maintain#723](https://github.com/theforeman/foreman_maintain/pull/723) renames it to `--skip`)
> * We don't want users to skip checks.
> * We don't provide any help on how to find out which checks the user could list
> 
> So overall I'd say this is not a direction we should head into. At best you could add a section on how to skip certain checks but it should not be the default user instruction.

Agree with this. Even prior collaborations with SMEs had the same point to direct the users in other ways, instead of whitelisting the parameters.

As it is not recommended to add the argument to this, I am closing this PR. Adding a note on my side of having some resources for users on using arguments.

PR #2245, opened 2023-06-20
Drop umask system requirement

mjivraja:
https://projects.theforeman.org/issues/35138

* [X] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/ref_system-requirements.adoc
+++ b/guides/common/modules/ref_system-requirements.adoc
@@ -38,7 +38,6 @@ ifdef::satellite[]
 * A current {ProjectName} subscription
 endif::[]
 * Administrative user (root) access
-* A system umask of 0022
 * Full forward and reverse DNS resolution using a fully-qualified domain name
 
 {Project} only supports `UTF-8` encoding.


ekohl (changes requested):
While the change is good, it's better to use `git revert 1427fd2e39d6f6a025c02c1b56bb53fd16e11601` as the command. That nicely ties it back to the commit that introduced the regression.

adamlazik1 (approved):
+1 to Ewoud's suggestion, the change looks good. Once the revert commit is applied this is good to go.

mjivraja:
> While the change is good, it's better to use `git revert 1427fd2e39d6f6a025c02c1b56bb53fd16e11601` as the command. That nicely ties it back to the commit that introduced the regression.

I am unable to revert this commit 😐 Don't know the reason as there is no error or info displayed. However, I am able to revert other commits.

Can we try any other way?

adamlazik1:
I went ahead and reverted this on master and 3.7 and 3.6 branches. Are those the only branches where this needs to be reverted @ekohl?

mjivraja:
> @mjivraja did you reset your commit? It's likely it can't revert because your commit already does the same. Without the command output I can only guess.
> 
> @adamlazik1 the Redmine says it was fixed in 3.4.0 I'd say the 3.4 & 3.5 branches also need the revert.

Ahh yes, I didn't reset the commit assuming the new revert commit will be on top. Thanks @ekohl :rocket:

adamlazik1:
I also reverted on 3.5 and 3.4 seems to not have the change in the first place. All changes are upstream so this PR can be safely closed.

mjivraja:
> I also reverted on 3.5 and 3.4 seems to not have the change in the first place. All changes are upstream so this PR can be safely closed.

Owwk. Thanks @adamlazik1 @ekohl 
Closing this PR.

PR #2249, opened 2023-06-21
Remove creating custom SSL certificate section

mjivraja:
Creating a custom SSL certificate for Project Server often created issues for end-users. Hence, we are expecting from end-users to get the custom SSL certificate from the third-party provider within their organization and this section should be removed to avoid confusion.

https://bugzilla.redhat.com/show_bug.cgi?id=2211341

* [X] Foreman 3.6/Katello 4.8


mjivraja:
> I would not remove this section. Instead, we should clarify which values need to be adjusted/replaced by users.

@ehelms Can we consider keeping this section and making the procedure accurate?

ehelms:
@maximiliankolb Can you expand on why you think we should keep this as part of our documentation? I don't believe we should be in the business of telling users how to generate certificates. We should support them providing custom certificates, but generating or acquiring custom certificates should be up to the user.

maximiliankolb:
At first I believed this is somewhat Foreman-specific. If it's not, then I think a sufficient alternative would be to a) specify what a user has to create/provide, and b), if possible, add a link on further documentation.

maximiliankolb (approved):
OK, then I consider this ready. Any other thoughts?

ehelms (approved):

ekohl:
@ehelms when merging in this repo you should also look at the requested cherry picks. In this case all the way back to 3.1.

ehelms:
> @ehelms when merging in this repo you should also look at the requested cherry picks. In this case all the way back to 3.1.

Do we have this documented somewhere that the merging party is responsible for the cherry picks?
Is Patchback able to help us with this?

maximiliankolb:
@ehelms There's only the "Please cherry-pick my commits into:" in the PR template. So far, we always do this manually. Do you have an example of "patchback" used in Ansible/Pulp?

ehelms:
Here is a Pulp example (note the labels):

 * https://github.com/pulp/pulp_rpm/pull/3200

Lennonka:
@maximiliankolb @ehelms Please, open a new issue for this topic. Discussing topics in merged PRs is a perfect way to make them lost. :)

PR #2259, opened 2023-06-26
Tweak install attributes to differentiate between project and client

asteflova:
First attempt at tweaking the install attributes. The goal is to be able to differentiate between commands run on the server (like {project-package-install}) and the client (like {client-package-install}).

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/attributes-base.adoc
+++ b/guides/common/attributes-base.adoc
@@ -77,7 +77,7 @@
 :oVirtEngine: {oVirt} Engine
 :oVirtShort: {oVirt}
 :package-clean: dnf clean
-:package-install-project: dnf install
+:project-package-install: dnf install
 :package-install: dnf install
 :package-remove-project: dnf remove
 :package-remove: dnf remove
--- a/guides/common/attributes-base.adoc
+++ b/guides/common/attributes-base.adoc
@@ -78,7 +78,7 @@
 :oVirtShort: {oVirt}
 :package-clean: dnf clean
 :project-package-install: dnf install
-:package-install: dnf install
+:client-package-install: dnf install
 :package-remove-project: dnf remove
 :package-remove: dnf remove
 :package-update-project: dnf update
--- a/guides/common/modules/con_troubleshooting-postgresql.adoc
+++ b/guides/common/modules/con_troubleshooting-postgresql.adoc
@@ -30,7 +30,7 @@ If a database was previously created using PostgreSQL 10, perform an upgrade:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {client-package-install} postgresql-upgrade
+# {project-package-install} postgresql-upgrade
 ----
 . Perform the upgrade:
 +
--- a/guides/common/modules/con_troubleshooting-postgresql.adoc
+++ b/guides/common/modules/con_troubleshooting-postgresql.adoc
@@ -30,7 +30,7 @@ If a database was previously created using PostgreSQL 10, perform an upgrade:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {package-install} postgresql-upgrade
+# {client-package-install} postgresql-upgrade
 ----
 . Perform the upgrade:
 +
--- a/guides/common/modules/proc_adding-rhel-system-roles.adoc
+++ b/guides/common/modules/proc_adding-rhel-system-roles.adoc
@@ -33,7 +33,7 @@ endif::[]
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {package-install-project} rhel-system-roles
+# {project-package-install} rhel-system-roles
 ----
 +
 The `rhel-system-roles` package downloads to `/usr/share/ansible/roles/`.
--- a/guides/common/modules/proc_benchmarking-raw-db-performance.adoc
+++ b/guides/common/modules/proc_benchmarking-raw-db-performance.adoc
@@ -6,7 +6,7 @@ To get a list of the top table sizes in disk space for both Candlepin, Foreman, 
 endif::[]
 
 PGbench utility (note you may need to resize PostgreSQL data directory `{postgresql-lib-dir}` directory to 100 GiB or what does benchmark take to run) might be used to measure PostgreSQL performance on your system.
-Use `{package-install} postgresql-contrib` to install it.
+Use `{client-package-install} postgresql-contrib` to install it.
 ifndef::orcharhino[]
 For more information, see https://github.com/RedHatSatellite/satellite-support[github.com/RedHatSatellite/satellite-support].
 endif::[]
--- a/guides/common/modules/proc_cloning_satellite_server.adoc
+++ b/guides/common/modules/proc_cloning_satellite_server.adoc
@@ -177,7 +177,7 @@ You can either mount the shared storage or copy the backup files to the `/backup
 +
 [options="nowrap" subs="attributes"]
 ----
-# {package-install} satellite-clone
+# {client-package-install} satellite-clone
 ----
 +
 After you install the `satellite-clone` tool, you can adjust any configuration to suit your own deployment in the `/etc/satellite-clone/satellite-clone-vars.yml` file.
--- a/guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
+++ b/guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
@@ -18,7 +18,7 @@ include::snip_firewalld.adoc[]
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {package-install} dhcp bind
+# {client-package-install} dhcp bind
 ----
 
 . Generate a security token:
@@ -120,7 +120,7 @@ _990_
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {package-install} {nfs-server-package}
+# {client-package-install} {nfs-server-package}
 # systemctl enable --now nfs-server
 ----
 
--- a/guides/common/modules/proc_configuring-ipxe-to-reduce-provisioning-times.adoc
+++ b/guides/common/modules/proc_configuring-ipxe-to-reduce-provisioning-times.adoc
@@ -59,7 +59,7 @@ ifdef::foreman-el,katello,satellite,orcharhino[]
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {client-package-install} ipxe-bootimgs
+# {project-package-install} ipxe-bootimgs
 ----
 . Correct the SELinux file contexts:
 +
@@ -72,7 +72,7 @@ ifdef::foreman-deb[]
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {client-package-install} ipxe
+# {project-package-install} ipxe
 ----
 endif::[]
 
--- a/guides/common/modules/proc_configuring-ipxe-to-reduce-provisioning-times.adoc
+++ b/guides/common/modules/proc_configuring-ipxe-to-reduce-provisioning-times.adoc
@@ -59,7 +59,7 @@ ifdef::foreman-el,katello,satellite,orcharhino[]
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {package-install} ipxe-bootimgs
+# {client-package-install} ipxe-bootimgs
 ----
 . Correct the SELinux file contexts:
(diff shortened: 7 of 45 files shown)


guides/common/modules/con_troubleshooting-postgresql.adoc
```diff
@@ -30,7 +30,7 @@ If a database was previously created using PostgreSQL 10, perform an upgrade:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {package-install} postgresql-upgrade
+# {client-package-install} postgresql-upgrade
```
ekohl, line 33: This is on the server, so I think `project-package-install` is more appropriate:
```suggestion
# {project-package-install} postgresql-upgrade
```

guides/common/modules/proc_adding-rhel-system-roles.adoc
```diff
@@ -33,7 +33,7 @@ endif::[]
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {package-install-project} rhel-system-roles
+# {project-package-install} rhel-system-roles
```
ekohl, line 36: I'm reading the guide and everything suggests this should be on the server (so `project`) is correct, but that means we only need to care about RHEL 8, since we no longer run on RHEL 7.

Could you submit that as a separate PR so we can easily cherry pick the fix to the appropriate branches?

guides/common/attributes-base.adoc
```diff
@@ -77,8 +77,8 @@
 :oVirtEngine: {oVirt} Engine
 :oVirtShort: {oVirt}
 :package-clean: dnf clean
-:package-install-project: dnf install
-:package-install: dnf install
+:project-package-install: dnf install
+:client-package-install: dnf install
```
ekohl, line 81: While I think this is a better name than we had (since it's qualified at least a bit), I do wonder about what the client is. I worry that we won't have "a" client, but will always need to list a few ways for the various clients.

guides/common/modules/proc_cloning_satellite_server.adoc
```diff
@@ -177,7 +177,7 @@ You can either mount the shared storage or copy the backup files to the `/backup
 +
 [options="nowrap" subs="attributes"]
 ----
-# {package-install} satellite-clone
+# {client-package-install} satellite-clone
```
ekohl, line 180: This bit is going to be confusing: cloning happens on a server, but at this point you don't have `{foreman-maintain}` installed. I'll leave a longer explanation in the whole review.

guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
```diff
@@ -18,7 +18,7 @@ include::snip_firewalld.adoc[]
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {package-install} dhcp bind
+# {client-package-install} dhcp bind
```
ekohl, line 21: This isn't a client, but it's a server that's not using foreman-maintain.

guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
```diff
@@ -120,7 +120,7 @@ _990_
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {package-install} {nfs-server-package}
+# {client-package-install} {nfs-server-package}
```
ekohl, line 123: Same here

guides/common/modules/proc_configuring-ipxe-to-reduce-provisioning-times.adoc
```diff
@@ -59,7 +59,7 @@ ifdef::foreman-el,katello,satellite,orcharhino[]
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {package-install} ipxe-bootimgs
+# {client-package-install} ipxe-bootimgs
```
ekohl, line 62: This is on a server, so technically this should be 
```suggestion
# {project-package-install} ipxe-bootimgs
```

Because they are the same for all non-satellite, it doesn't matter but it happens on the server.

guides/common/modules/proc_configuring-ipxe-to-reduce-provisioning-times.adoc
```diff
@@ -72,7 +72,7 @@ ifdef::foreman-deb[]
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {package-install} ipxe
+# {client-package-install} ipxe
```
ekohl, line 75: Same here
```suggestion
# {project-package-install} ipxe
```

ekohl (commented):
Today you use `{package-install}` while doing basic installations and `{package-install-project}` once `{foreman-maintain}` is installed. The explanation can be confusing, but on Satellite we use a foreman-maintain feature where we lock all package versions, to prevent accidental upgrades. This package locking feature is not used outside of Satellite, so `{package-install}` and `{package-install-project}` are the same.

You also use plain `{package-install}` to install foreman-maintain, indirectly via the `satellite` or `capsule` meta package.

I didn't go through all of the files, but I wanted to clarify that now.

asteflova:
@ekohl This is amazing! Thanks a lot for the review and the explanations.

@maximiliankolb After trying to make this work and reading Ewoud's comments, I'm starting to think it's not a good idea to introduce two different attributes: client-package-install and project-package-install. A lot of times, I just won't be able to tell which one to use, and I don't think it's just because I'm new to the product. Maybe the other writers on the team would be able to tell, but anyone who's new or just not that experienced would have trouble doing so. Which would be a burden on the more experienced writers and SMEs to correct the junior members. Not to mention SMEs would first need to be explained what the difference is and that they need to check the attrs. What do you think?

maximiliankolb:
I think esp. because it is hard, we should do it. I believe that renaming the attribute is the first logical step (a simple regex as per your commits) which results in no diff in the HTML output; see first comment.

This means that all of Ewouds comments are important and must be applied because they hint towards mistakes in the docs, esp. for foreman-deb.

I didn't have time this week yet, but your work on this is definitely appreciated. I will try to have a closer look at it next week maybe.

asteflova:
I'll apply Ewoud's suggestions next week and take more time to think on them. I will do my best to get better understanding of the difference, maybe things will get clearer to me after that.

In the meantime, thanks everyone, I really appreciate the input.

ekohl:
You could consider `{project-package-install-bootstrap}` and `{project-package-install}` to distinguish that one happens on the project while "bootstrapping" (making sure you can use the regular tooling) and one where you have the tooling available. Of course these names aren't perfect, so take them as a hint on the direction and not the exact naming.

asteflova:
Tagging the other doc team members.

This is a draft pull request to implement changes to how install attributes are used in the docs set. The goal is to be able to differentiate between commands run on the server ({project-package-install}) and the client ({client-package-install}).

To sumarize the current status of the PR:
* Ewoud posted helpful feedback, to be implemented
* I [expressed concern](https://github.com/theforeman/foreman-documentation/pull/2259#issuecomment-1614140983) about the practical part of being able to maintain this system
* Max believes this is [very much worth doing](https://github.com/theforeman/foreman-documentation/pull/2259#issuecomment-1614213409) to ensure the docs are accurate

@adamlazik1 @AkshayGadhaveRH @bangelic @Lennonka @mjivraja  If you have any ideas or comments, please share them.

guides/common/attributes-base.adoc
Lennonka: Client is supposed to be a host.
And I agree that for upstream we usually need to have a list of more than a single OS, so I'm not sure how this will help.

guides/common/modules/proc_benchmarking-raw-db-performance.adoc
```diff
@@ -6,7 +6,7 @@ To get a list of the top table sizes in disk space for both Candlepin, Foreman,
 endif::[]
 
 PGbench utility (note you may need to resize PostgreSQL data directory `{postgresql-lib-dir}` directory to 100 GiB or what does benchmark take to run) might be used to measure PostgreSQL performance on your system.
-Use `{package-install} postgresql-contrib` to install it.
+Use `{client-package-install} postgresql-contrib` to install it.
```
Lennonka, line 9: @ekohl This also looks like it should be `project-`. Can you confirm?

guides/common/attributes-base.adoc
Lennonka, line 81:
```suggestion
:client-package-install-el: dnf install
:client-package-install-deb: apt-get install
```
Perhaps this would be more helpful?

guides/common/modules/proc_benchmarking-raw-db-performance.adoc
ekohl: That's correct.

Now that I'm reading this I'm looking at what this chapter is supposed to achieve. It's quite. Katello-centric in assuming that Candlepin and Pulp are present.

Overall I wonder what the point of the linked sizes script is. https://stackoverflow.com/questions/18907047/postgres-db-size-command says you can use `\l+` the psql utility and that does work for me:
```console
$ sudo -u postgres psql
psql (12.12)
Type "help" for help.

postgres=# \l+
                                                                    List of databases
   Name    |  Owner   | Encoding |   Collate   |    Ctype    |   Access privileges   |  Size   | Tablespace |                Description                 
-----------+----------+----------+-------------+-------------+-----------------------+---------+------------+--------------------------------------------
 foreman   | foreman  | UTF8     | en_US.utf8  | en_US.utf8  | foreman=CTc/foreman  +| 38 MB   | pg_default | 
           |          |          |             |             | =T/foreman            |         |            | 
 postgres  | postgres | UTF8     | nl_NL.UTF-8 | nl_NL.UTF-8 |                       | 7337 kB | pg_default | default administrative connection database
 template0 | postgres | UTF8     | nl_NL.UTF-8 | nl_NL.UTF-8 | =c/postgres          +| 8089 kB | pg_default | unmodifiable empty database
           |          |          |             |             | postgres=CTc/postgres |         |            | 
 template1 | postgres | UTF8     | nl_NL.UTF-8 | nl_NL.UTF-8 | postgres=CTc/postgres+| 7177 kB | pg_default | default template for new databases
           |          |          |             |             | =c/postgres           |         |            | 
(4 rows)
```

Perhaps @evgeni knows better if there's a difference.

While on the topic: since Pulp 3.22 (included in Katello 4.9) we always require postgresql-contrib anyway, but it's still needed on vanilla Foreman (without Katello).

Then about the specific PGbench part: I'm not sure what's expected of the reader.

In conclusion: what's the point of this chapter?

evgeni: Well, the linked script shows table sizes, not database sizes, so the output is quite different.
Whether this amount of detail is required at that point or what the reader should do with it, I do not know.

(I also prefer not to link to random GitHub repos in docs, but that's another totally different topic).

guides/common/attributes-base.adoc
mjivraja: Can we set the value as `{package-manager} install` and have an applicable package manager attribute in `attribute-{build}`?

Lennonka: @maximiliankolb @ekohl Opinions?

guides/common/modules/proc_benchmarking-raw-db-performance.adoc
ekohl: > Well, the linked script shows table sizes, not database sizes (there is `\dt+` for table sizes, but that's still less detail than the script), so the output is quite different.

Ah, good point.

> Whether this amount of detail is required at that point or what the reader should do with it, I do not know.

I also wonder what the point of this chapter is. Yes, you know table sizes. But then what do you do with it?

> (I also prefer not to link to random GitHub repos in docs, but that's another totally different topic).

Also what triggered me and why I started to look at native PostgreSQL capabilities. So you can inline it.

guides/common/attributes-base.adoc
Lennonka: @maximiliankolb Yes, please, bring it to the meeting. I don't understand very well what you have in mind.

maximiliankolb:
I will re-use this _some day_ to differentiate between "OS of Foreman" and "OS of managed hosts".

maximiliankolb:
Closing this draft PR in favor of https://github.com/theforeman/foreman-documentation/pull/2625 cc @asteflova

PR #2267, opened 2023-06-28
Rerun buttons

asteflova:
This update aims to document the enhancement in [BZ#2203183](https://bugzilla.redhat.com/show_bug.cgi?id=2203183).

The buttons in the form for running jobs on hosts have been changed, as described [in the Foreman community](https://github.com/theforeman/foreman_remote_execution/pull/794).

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_executing-a-remote-job.adoc
+++ b/guides/common/modules/proc_executing-a-remote-job.adoc
@@ -46,7 +46,7 @@ There can only be one active job with a given purpose at a time.
 Click *Next* after you have entered the required information.
 . Review job details.
 You have the option to return to any part of the job wizard and edit the information.
-. Click *Run* to schedule the job for execution.
+. Click *Submit* to schedule the job for execution.
 
 [id="cli-executing-a-remote-job_{context}"]
 .CLI procedure


asteflova:
@adamruzicka Can you please review this?

I'm fairly certain about changing the label for the button to run the job (s/Run/Submit). I actually didn't find any *Run* button in the web UI in the form for running jobs.

About adding description for the two new buttons (*Run on selected hosts* and *Skip to review*): I'd actually prefer to leave this part out before finalizing this PR. What those two buttons do seems perfectly clear from the web UI, thanks in part to their hover description. That's why adding the same thing to the guide seems pointless to me. But feel free to disagree with me.

@adamlazik1 Can you please have a look as well?

guides/common/modules/proc_executing-a-remote-job.adoc
```diff
@@ -46,7 +46,12 @@ There can only be one active job with a given purpose at a time.
 Click *Next* after you have entered the required information.
 . Review job details.
 You have the option to return to any part of the job wizard and edit the information.
-. Click *Run* to schedule the job for execution.
+. Click *Submit* to execute the job.
+
+When rerunning an existing job, the following buttons are also available:
```
adamruzicka, line 51: The buttons are always there, it doesn't matter if you're rerunning a job or running a brand new one.

adamruzicka, line 51: It should probably also say that all required inputs need to be filled in for these buttons to become clickable

adamruzicka (commented):
> I'm fairly certain about changing the label for the button to run the job (s/Run/Submit).

You're right

> I'd actually prefer to leave this part out before finalizing this PR

Fine by me

guides/common/modules/proc_executing-a-remote-job.adoc
asteflova: There is actually hover text for this too - when the input is missing, the buttons are grey and unclickable, and a hint appears saying that you need to fill all the required fields. Good UI doesn't need docs :)

guides/common/modules/proc_executing-a-remote-job.adoc
```diff
@@ -46,7 +46,7 @@ There can only be one active job with a given purpose at a time.
 Click *Next* after you have entered the required information.
 . Review job details.
 You have the option to return to any part of the job wizard and edit the information.
-. Click *Run* to schedule the job for execution.
+. Click *Submit* to execute the job.
```
adamlazik1, line 49: I think you should keep it like this because the job doesn't need to be executed immediately based on the execution time that user selected.
```suggestion
. Click *Submit* to schedule the job for execution.
```

adamlazik1 (approved):
LGTM, what about @adamruzicka ?

adamruzicka (approved):

adamlazik1:
Merged and cherry-picked:
   7cc353b85..0762f8faa  3.7 -> 3.7

PR #2287, opened 2023-07-11
Update procedure to deploy custom SSL to hosts

mjivraja:
As the `katello-ca-consumer` rpm is deprecated, there are 2 ways end-users can deploy the custom SSL certificate to the hosts after deploying it to the Project. We have documented the recommended way to deploy it to the hosts. Also, we have added the insecure arguments will curl command so that it uses port 443 which is reliable at this point.

https://bugzilla.redhat.com/show_bug.cgi?id=2216692

* [X] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-hosts.adoc
+++ b/guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-hosts.adoc
@@ -2,13 +2,15 @@
 
 = Deploying a Custom SSL Certificate to Hosts
 
-After you configure {ProductName} to use a custom SSL certificate, you must also install the `katello-ca-consumer` package on every host that is registered to this {ProductName}.
+After you configure {ProductName} to use a {customssl} certificate, you must also deploy the new {customssl} certificate on every host that is registered to this {ProductName}.
 
 .Procedure
 
-* On each host, install the `katello-ca-consumer` package:
+* On each host, run the following commands:
 +
 [options="nowrap", subs="+quotes,attributes"]
 ----
-# {package-install} http://_{context}.example.com_/pub/katello-ca-consumer-latest.noarch.rpm
+# _SERVER=$(subscription-manager config | awk '/ hostname/{print $NF}'| tr -d '[]')
+# curl -k https://$_SERVER/pub/katello-rhsm-consumer | bash
+# subscription-manager refresh
 ----


ekohl (commented):
@ehelms I wonder if this really is the best replacement. Do we have any item like a remote execution job we can easily run that has this? Then we can recommend running it for all hosts with `content_source = hostname`. New hosts should use host registration anyway.

ehelms:
This is an incorrect replacement but I can understand why it's not obvious from our messaging. The `katello-ca-consumer` is deprecated and is essentially a wrapper around the steps added here: run bash script and sub-man refresh. When we finally remove the `katello-ca-consumer` this script will also go away. In that light we should have a better "this is how you refresh certificates" story. I think that part of things needs more flushing out and thus until we have that, I would close this for now and hold off on this change.

mjivraja:
Ack! As `katello-ca-consumer` is deprecated (and will be removed), the changes in this PR will become obsolete too. We will have a very different procedure for updating custom SSL on hosts.

However, we can keep this PR to a version specific instead of `master` if required as a workaround.

CC: @ekohl

ehelms:
My point is, this work around as written is not a change we should make. Using katello-ca-consumer or as you have it written is the same thing and thus changing this gives a false impression. 

What we really need to tackle is having steps for how to do with global registration... except that as @ekohl has pointed out Global Registration is about *registering* and not re-configuration. This shows up here as well -- https://github.com/theforeman/foreman-packaging/pull/9591

My recommendation is to close this PR, and not tackle that bug until we have an agreed upon solution for re-configuration in a Global Registration only world. I will follow up on this.

mjivraja:
> My point is, this work around as written is not a change we should make. Using katello-ca-consumer or as you have it written is the same thing and thus changing this gives a false impression.
> 
> What we really need to tackle is having steps for how to do with global registration... except that as @ekohl has pointed out Global Registration is about _registering_ and not re-configuration. This shows up here as well -- [theforeman/foreman-packaging#9591](https://github.com/theforeman/foreman-packaging/pull/9591)
> 
> My recommendation is to close this PR, and not tackle that bug until we have an agreed upon solution for re-configuration in a Global Registration only world. I will follow up on this.

Ack! @ehelms

mjivraja:
We are closing this PR as the changes do not technically address the solution. It can have an optimal future solution as mentioned in the above conversations.

PR #2319, opened 2023-07-31
Updates for Katello 4.8.1, 4.8.2, 4.8.3 and 4.8.4

lfu:

* [x] Foreman 3.6/Katello 4.8


guides/doc-Release_Notes/topics/katello-4.8.3.adoc
```diff
@@ -0,0 +1,3 @@
+= Katello 4.8.3
+
+A full list of changes is available on https://projects.theforeman.org/versions/1743[Redmine]
```
lfu, line 3: Link created by command `./guides/doc-Release_Notes/redmine_release_notes katello 4.8.3 > ./guides/doc-Release_Notes/topics/katello-4.8.3.adoc` did not work for Katello releases.

Foreman uses `fixed in releases` as the filter while Katello uses `target version`.

I used Katello release version link in the file.

ekohl (changes requested):
The whole idea is that the page provides a single overview a user can read through.

I think we should use `Fixed in releases` and correct them where needed. I already made sure the bot sets the field correctly on stable branches (https://github.com/theforeman/prprocessor/commit/4bf832a5a43b32b5cc9a598db048431e88633d5a). So I'd now suggest to either correct the fields in Redmine or (temporarily) modify the script to read the `target version` field. The latter can be done by changing this line:

https://github.com/theforeman/foreman-documentation/blob/30bae89368ee5d416e8e0ff3638d6e67b3d52898/guides/doc-Release_Notes/redmine_release_notes#L37

However, I now scripted it. You can easily get the issue numbers using `git log 4.8.0..4.8.1 --oneline | awk '/Fixes/ { print substr($3, 2) }'`. The Python bits were a bit more verbose.

I do see there's a problem in my script and that probably happened because the 4.8.2 was not part of the `KATELLO-4.8` branch. This is because you tagged the PR, not the final merged commit. So `git log 4.8.2..4.8.3` also lists all changes from 4.8.2.

At this point I would not tag it again, because that's asking for problems. However, it can point to a problem in the procedure.

mjivraja (approved):
All links tested, LGTM :+1:

ekohl:
> The Python bits were a bit more verbose.

I posted it @ https://gist.github.com/ekohl/56b1baddb7247c2fb0d00d8908f483a8

guides/doc-Release_Notes/topics/katello-4.8.4.adoc
```diff
@@ -0,0 +1,13 @@
+= Katello 4.8.4
+
+A full list of changes is available on https://projects.theforeman.org/issues?set_filter=1&sort=id%3Adesc&status_id=closed&f%5B%5D=cf_12&op%5Bcf_12%5D=%3D&v%5Bcf_12%5D%5B%5D=1749[Redmine]
```
lfu, line 3: This link does not work.

lfu: I fixed the remine to add 4.8.4 in `fixed in releases`. Then the link worked 👍🏻

guides/doc-Release_Notes/topics/katello-4.8.1.adoc
```diff
@@ -0,0 +1,24 @@
...
+A full list of changes is available on https://projects.theforeman.org/issues?set_filter=1&sort=id%3Adesc&status_id=closed&f%5B%5D=cf_12&op%5Bcf_12%5D=%3D&v%5Bcf_12%5D%5B%5D=1688[Redmine]
+
+== Katello
+
+* DB Migration fails because of missing Setting::Content - https://projects.theforeman.org/issues/36343[#36343]
+* Enable tracer on host page (debian os) is always on - https://projects.theforeman.org/issues/36297[#36297]
+
+=== Content Views
+
+* Getting "NoMethodError undefined method \`get_status\' for nil:NilClass" when publishing content view - https://projects.theforeman.org/issues/36303[#36303]
```
ekohl, line 12: I think this doesn't need escaping because it doesn't render well. I never understand escaping in asciidoc.
```suggestion
* Getting "NoMethodError undefined method `get_status' for nil:NilClass" when publishing content view - https://projects.theforeman.org/issues/36303[#36303]
```

Looks like it's repeated a few times.

ekohl (commented):
I just found https://docs.asciidoctor.org/asciidoc/latest/pass/pass-macro/ and I should experiment with that. For now, please make sure it renders well by removing some escaping where it breaks.

ekohl (approved):

PR #2321, opened 2023-08-01
Hide guides which aren't ready for consumption

ekohl:
Some guides aren't usable in some flavors. They aren't linked in the navigation, but this also makes sure the files don't have any content. That makes diffs smaller, which makes reviewing easier.

<del>Right now this is just a proof of concept and I haven't gone through all the guides.</del>

### DIFF
--- a/guides/doc-Administering_Project/master.adoc
+++ b/guides/doc-Administering_Project/master.adoc
@@ -4,6 +4,12 @@ include::common/header.adoc[]
 
 = {AdministeringDocTitle}
 
+// This guide is not ready for stable releases
+ifdef::HideDocumentOnStable[]
+include::common/modules/snip_guide-not-ready.adoc[]
+endif::[]
+ifndef::HideDocumentOnStable[]
+
 ifdef::satellite[]
 include::common/modules/proc_providing-feedback-on-red-hat-documentation.adoc[leveloffset=+1]
 endif::[]
@@ -60,3 +66,4 @@ include::common/assembly_searching-and-bookmarking.adoc[leveloffset=+1]
 
 [appendix]
 include::common/assembly_administer-settings-information.adoc[]
+endif::[]
--- a/guides/doc-Configuring_Load_Balancer/master.adoc
+++ b/guides/doc-Configuring_Load_Balancer/master.adoc
@@ -4,6 +4,12 @@ include::common/header.adoc[]
 
 = {ConfiguringLoadBalancerDocTitle}
 
+// This guide only works on content proxies, which is Katello only
+ifdef::foreman-deb,foreman-el[]
+include::common/modules/snip_guide-not-ready.adoc[]
+endif::[]
+ifndef::foreman-deb,foreman-el[]
+
 ifdef::satellite[]
 include::common/modules/proc_providing-feedback-on-red-hat-documentation.adoc[leveloffset=+1]
 endif::[]
@@ -62,3 +68,4 @@ include::common/modules/con_propagating-scap-content-through-the-load-balancer.a
 include::common/modules/proc_propagating-scap-content-using-ansible-deployment.adoc[leveloffset=+2]
 
 include::common/modules/proc_propagating-scap-content-using-puppet-deployment.adoc[leveloffset=+2]
+endif::[]
--- a/guides/doc-Installing_Server/master.adoc
+++ b/guides/doc-Installing_Server/master.adoc
@@ -7,6 +7,12 @@ include::common/header.adoc[]
 
 = {InstallingServerDocTitle}
 
+// This guide is not ready for stable releases
+ifdef::HideDocumentOnStable[]
+include::common/modules/snip_guide-not-ready.adoc[]
+endif::[]
+ifndef::HideDocumentOnStable[]
+
 ifdef::satellite[]
 include::common/modules/proc_providing-feedback-on-red-hat-documentation.adoc[leveloffset=+1]
 endif::[]
@@ -82,3 +88,5 @@ include::common/modules/con_applying-custom-configuration.adoc[leveloffset=+1]
 
 [appendix]
 include::common/modules/proc_restoring-manual-changes-overwritten-by-a-puppet-run.adoc[leveloffset=+1]
+
+endif::[]
--- a/guides/doc-Managing_Content/master.adoc
+++ b/guides/doc-Managing_Content/master.adoc
@@ -9,6 +9,12 @@ include::common/modules/proc_providing-feedback-on-red-hat-documentation.adoc[le
 endif::[]
 
 // Render only for relevant and finished contexts
+ifndef::katello,satellite,orcharhino[]
+Content management requires the Katello plugin.
+Conversion of an existing installation is not supported.
+// Can't use the regular macros because they are for the wrong flavor
+See the {BaseURL}Installing_Server/index-katello.html[Katello installation guide] on how to install Katello.
+endif::[]
 ifdef::katello,satellite,orcharhino[]
 
 include::common/modules/con_introduction-to-content-management.adoc[leveloffset=+1]
--- a/guides/doc-Managing_Hosts/master.adoc
+++ b/guides/doc-Managing_Hosts/master.adoc
@@ -3,10 +3,13 @@ include::common/header.adoc[]
 :context: managing-hosts
 :managing-hosts:
 
-// Render only for relevant and finished contexts
-ifdef::foreman-el,katello,satellite,orcharhino[]
-
 = {ManagingHostsDocTitle}
+
+// This guide is not ready for stable releases and not on Debian
+ifdef::HideDocumentOnStable,foreman-deb[]
+include::common/modules/snip_guide-not-ready.adoc[]
+endif::[]
+ifndef::HideDocumentOnStable,foreman-deb[]
 
 ifdef::satellite[]
 include::common/modules/proc_providing-feedback-on-red-hat-documentation.adoc[leveloffset=+1]
--- a/guides/doc-Managing_Organizations_and_Locations/master.adoc
+++ b/guides/doc-Managing_Organizations_and_Locations/master.adoc
@@ -4,7 +4,13 @@ include::common/header.adoc[]
 
 = {ManagingOrganizationsLocationsDocTitle}
 
-ifndef::satellite[]
+// This guide is not ready for stable releases
+// Satellite publishes this documentation elsewhere
+ifdef::HideDocumentOnStable,satellite[]
+include::common/modules/snip_guide-not-ready.adoc[]
+endif::[]
+ifndef::HideDocumentOnStable,satellite[]
+
 include::common/modules/con_introduction-to-organization-and-location-context.adoc[leveloffset=+1]
 
 include::common/assembly_managing-organizations.adoc[leveloffset=+1]
--- a/guides/doc-Managing_Security_Compliance/master.adoc
+++ b/guides/doc-Managing_Security_Compliance/master.adoc
@@ -11,6 +11,12 @@ This feature is not available for Debian/Ubuntu.
 endif::[]
 ifndef::foreman-deb[]
 
+// This guide is not ready for stable releases
+ifdef::HideDocumentOnStable[]
+include::common/modules/snip_guide-not-ready.adoc[]
+endif::[]
+ifndef::HideDocumentOnStable[]
+
 ifdef::satellite[]
 include::common/modules/proc_providing-feedback-on-red-hat-documentation.adoc[leveloffset=+1]
 endif::[]
@@ -87,4 +93,5 @@ include::common/modules/proc_deleting-a-compliance-report.adoc[leveloffset=+2]
 
 include::common/modules/proc_deleting-multiple-compliance-reports.adoc[leveloffset=+2]
 
+endif::[]
 endif::[]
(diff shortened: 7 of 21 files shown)


guides/doc-Configuring_Load_Balancer/master.adoc
```diff
@@ -4,6 +4,12 @@ include::common/header.adoc[]
 
 = {ConfiguringLoadBalancerDocTitle}
 
+// This guide only works on content proxies, which is Katello only
+ifdef::HideDocument,foreman-deb,foreman-el[]
```
Lennonka, line 8:
```suggestion
ifdef::HideDocument[]
```
Aren't those redundant when HideDocument is only defined in foreman-deb and foreman-el builds?

guides/doc-Installing_Server/master.adoc
```diff
@@ -7,6 +7,11 @@ include::common/header.adoc[]
 
 = {InstallingServerDocTitle}
 
+ifdef::HideDocument[]
+include::common/modules/snip_guide_not_ready.adoc
+endif::[]
+ifndef::foreman-deb,foreman-el[]
```
Lennonka, line 13:
```suggestion
ifndef::HideDocument[]
```
?

Lennonka (commented):
If the point is to hide stuff only for stable branches, the conditions can be fine-tuned as follows, I think.

Lennonka:
Otherwise a good idea. I'd wait for #2233 to add it to Updating guide as well.

guides/common/modules/snip_guide_not_ready.adoc
maximiliankolb, line 1: Please rename to `snip_guide-not-ready.adoc`. By convention, all files in "guides/common/modules" only have one "_".

guides/doc-Configuring_Load_Balancer/master.adoc
```diff
@@ -4,6 +4,12 @@ include::common/header.adoc[]
 
 = {ConfiguringLoadBalancerDocTitle}
 
+// This guide only works on content proxies, which is Katello only
```
maximiliankolb, line 7: :+1: 

A small note _why_ we choose not to show a guide is very helpful.

maximiliankolb (changes requested):
I like the approach with `HideDocument`. Do you think we can/should use this in other guides with a more granular approach too?

guides/doc-Configuring_Load_Balancer/master.adoc
ekohl: Kind of. Perhaps it's a premature optimization for when the guide would be ready. Now it's just not usable at all. I guess we should keep it simple and deal with that stuff when we do get to it.

guides/doc-Installing_Server/master.adoc
ekohl: Yes, left over from before I introduced HideDocument.

ekohl:
> I like the approach with `HideDocument`. Do you think we can/should use this in other guides with a more granular approach too?

What would be more granular?

guides/common/modules/snip_guide_not_ready.adoc
ekohl: Will do. Any feedback on the actual content? Should we make it more actionable and link to https://github.com/theforeman/foreman-documentation/issues to create an issue if readers want to see it?

guides/common/modules/snip_guide_not_ready.adoc
```diff
@@ -0,0 +1 @@
+This guide is not ready yet.
```
maximiliankolb, line 1:
```suggestion
This guide is not ready yet.

Foreman is a community project.
If you want to improve documentation, please go to https://github.com/theforeman/foreman-documentation/[github.com > theforeman > foreman-documentation/] to create issues and pull requests.
Thanks for your contribution.
```

maximiliankolb (approved):
I like this approach.

maximiliankolb:
> > I like the approach with `HideDocument`. Do you think we can/should use this in other guides with a more granular approach too?
> 
> What would be more granular?

Down the road, we could only hide invidual assemblies instead of everything in "master.adoc". But IMHO better showing a hint "please contribute and make it happen" than to never show docs for Foreman (and Foreman on deb) at all!

So :+1:

ekohl:
> Down the road, we could only hide invidual assemblies instead of everything in "master.adoc". But IMHO better showing a hint "please contribute and make it happen" than to never show docs for Foreman (and Foreman on deb) at all!

Possibly, though I wonder what's more work: fixing the assembly or hiding it. I guess it's one of those "it depends" kind of things.

ekohl:
While reviewing this I found https://github.com/theforeman/foreman-documentation/issues/2333 & https://github.com/theforeman/foreman-documentation/issues/2332.

I think this is now reasonable ready, but I would ask that we don't pick this as we usually do. Having the review with the diff is useful. I'm leaning to picking to 3.7 even before we merge just so we see the impact. There's https://github.com/theforeman/foreman-documentation/pull/2334 now.

guides/doc-Administering_Project/master.adoc
```diff
@@ -4,6 +4,12 @@ include::common/header.adoc[]
 
 = {AdministeringDocTitle}
 
+// This guide is not ready for stable releases
+ifdef::HideDocument[]
+include::common/modules/snip_guide_not_ready.adoc
```
ekohl, line 9: The 3.7 PR highlighted that I forgot `[]` here.
```suggestion
include::common/modules/snip_guide_not_ready.adoc[]
```

guides/common/modules/snip_guide_not_ready.adoc
ekohl: I went with a slightly different text. Please review.

guides/doc-Managing_Organizations_and_Locations/master.adoc
```diff
@@ -4,7 +4,12 @@ include::common/header.adoc[]
 
 = {ManagingOrganizationsLocationsDocTitle}
 
-ifndef::satellite[]
+// This guide is not ready for stable releases
+ifdef::HideDocument,satellite[]
+include::common/modules/snip_guide-not-ready.adoc[]
```
ekohl, line 9: Perhaps the text is not applicable for satellite since it's very upstream focused. Is that a problem?

guides/doc-Managing_Content/master.adoc
```diff
@@ -9,6 +9,11 @@ include::common/modules/proc_providing-feedback-on-red-hat-documentation.adoc[le
 endif::[]
 
 // Render only for relevant and finished contexts
+ifndef::katello,satellite,orcharhino[]
+Content management requires the Katello plugin.
+Conversion of an existing installation is not supported.
+See the {InstallingServerDocURL}[Katello installation guide] on how to install Katello.
```
maximiliankolb, line 15:
```suggestion
Content management requires the Katello plug-in.
You cannot install the Katello plug-in on your existing {ProjectServer}.
For more information, see {InstallingServerDocURL}{InstallingServerDocTitle}.
```

guides/doc-Managing_Hosts/master.adoc
```diff
@@ -3,11 +3,14 @@ include::common/header.adoc[]
 :context: managing-hosts
 :managing-hosts:
 
-// Render only for relevant and finished contexts
-ifdef::foreman-el,katello,satellite,orcharhino[]
-
 = {ManagingHostsDocTitle}
 
+// This guide is not ready for stable releases and not on Debian
```
maximiliankolb, line 8: What's the reasoning here? Is it about [Chapter 2.4+](https://docs.theforeman.org/nightly/Managing_Hosts/index-foreman-el.html#Changing_a_Module_Stream_for_a_Host_managing-hosts) + Chapter 6 + Chapter 14?

If so, then maybe we can only hide specific assemblies?

guides/doc-Managing_Organizations_and_Locations/master.adoc
maximiliankolb: I doubt that many people look at "satellite" docs on theforeman.org. So I am fine either way; not building it because Satellite provided the content at other places is probably a good idea though.

guides/doc-Managing_Security_Compliance/master.adoc
```diff
@@ -11,6 +11,12 @@ This feature is not available for Debian/Ubuntu.
 endif::[]
 ifndef::foreman-deb[]
 
+// This guide is not ready for stable releases
```
maximiliankolb, line 14: Maybe combine with line 9+10?

guides/doc-Planning_for_Project/master.adoc
```diff
@@ -4,6 +4,12 @@ include::common/header.adoc[]
 
 = {PlanningDocTitle}
 
+// This guide is Katello specific, in particular the diagrams
+ifdef::foreman-el,foreman-deb[]
```
maximiliankolb, line 8: So "HideDocument" here? Is there a diff between this and other guides that we're not even want to show it for nightly?

guides/doc-Provisioning_Hosts/master.adoc
```diff
@@ -4,6 +4,12 @@ include::common/header.adoc[]
 
 = {ProvisioningDocTitle}
 
+// This guide is not ready for stable releases
```
maximiliankolb, line 7: same as for "managing hosts" above: I believe we can/should hide smaller parts; more specifically those that mention Katello-only features.

maximiliankolb (commented):
At some places, I wonder if _always_ using "if not foreman-el and not foreman-deb" is more verbose/better than "hideDocument". I am unsure if it makes sense to hide some guides completely for releases but not nightly; and some guides completely for any revision for foreman-deb?

Nevertheless, apart from my comments/questions, I belive this adds value to our built docs.

guides/doc-Managing_Content/master.adoc
ekohl: I explicitly didn't use `{InstallingServerDocTitle}` because it's for the current flavor, not the Katello flavor.

guides/doc-Managing_Hosts/master.adoc
ekohl: It's completely hidden today (https://docs.theforeman.org/nightly/Managing_Hosts/index-foreman-deb.html), which is what I went with. It's a good idea to hide specific assemblies and simplify it to `ifdef::HideDocument`, but I consider that out of scope for this PR.

guides/doc-Managing_Organizations_and_Locations/master.adoc
ekohl: That was also my thought process, but I didn't know it was hidden for Satellite since it's provided elsewhere. Actually a good comment to have.

guides/doc-Planning_for_Project/master.adoc
ekohl: No, HideDocument only hides it for stable branches but this whole document is not ready. Not even in nightly. Should the `HideDocument` attribute reflect that?

guides/doc-Provisioning_Hosts/master.adoc
ekohl: My idea is that we should certainly do that, but this provides a good base for it.

For example, https://docs.theforeman.org/nightly/Provisioning_Hosts/index-foreman-deb.html#using-a-synced-kickstart-repository_provisioning is a section we should hide. https://docs.theforeman.org/nightly/Provisioning_Hosts/index-foreman-deb.html#Creating_an_Installation_Medium_for_Debian-Ubuntu_provisioning contains an unreplaced macro.

Again, doing everything in a single PR is massive. What I'd like to do is gather a list of guides that are ready and start exposing those in the navigation. Then one by one we can review guides and expose them.

guides/doc-Planning_for_Project/master.adoc
maximiliankolb: Thinking about it: Maybe we should use "if not foreman-deb ..." for **all** instances where we want to hide a guide for a specific build target, irrespective of the version & rename "HideDocument" to something like "hide on stable branches", "not ready for production" or "nightly only". I like the latter.

guides/doc-Managing_Content/master.adoc
maximiliankolb: But if the link resolves for "foreman-el", users will end up at "Installing Foreman on EL", not katello.

ekohl: Oh yes, you're right.

ekohl: Now you've distracted me with unrelated work: https://github.com/theforeman/foreman-documentation/pull/2341

guides/doc-Planning_for_Project/master.adoc
ekohl: I've chosen to name it `HideDocumentOnStable`. The downside of nightly only is that the logic gets a bit complicated. Katello guides don't use it and always show. A good example where naming is hard.

guides/doc-Managing_Content/master.adoc
ekohl: Updated now with an explicit link & comment.

maximiliankolb (approved):
LGTM!

ekohl:
I've just picked it to https://github.com/theforeman/foreman-documentation/pull/2334 to see the impact on a stable branch.

Lennonka (approved):
LGTM!

PR #2334, opened 2023-08-03
[3.7] Hide guides which aren't ready for consumption

ekohl:
Some guides aren't usable in some flavors. They aren't linked in the navigation, but this also makes sure the files don't have any content.  That makes diffs smaller, which makes reviewing easier.

(cherry picked from commit f5790e05e0b8fa50513efb048ab68f00057ec32c)

This is just as a PR to see the diff of https://github.com/theforeman/foreman-documentation/pull/2321 for now.


ekohl:
I've just picked the merged commit from master and this should be the final version of this PR.

PR #2346, opened 2023-08-08
Edit overview of permissions granted by default roles

asteflova:
[BZ#2182751](https://bugzilla.redhat.com/show_bug.cgi?id=2182751) requests more information about permissions provided by the Organization admin default role. This PR attempts to fix that by proposing the following edits to the overview table:

* Splitting the descriptions of the roles into a column with _permissions_ and a column with _additional information_.
* Making sure the Organization admin role is described properly.

In the interest of avoiding duplication and further improving readability of the table, I've also cross-checked the role descriptions in this module with the descriptions in the web UI. Then, I removed descriptions from this guide that did not provide additional value.

* [ ] Foreman 3.6/Katello 4.8


asteflova:
@adamruzicka I added descriptions for a few default roles, based on the feedback received in [BZ#2182751](https://bugzilla.redhat.com/show_bug.cgi?id=2182751). Can you please take a look?

I think you can limit your review to this commit: https://github.com/theforeman/foreman-documentation/pull/2346/commits/2edd8606269307e79a7a42049db83835cc06593c The rest of this PR is just editorial changes, i.e. no significant changes to the content that was already included.

guides/common/modules/ref_predefined-roles.adoc
```diff
@@ -7,48 +7,140 @@ To view the exact set of permissions a predefined role grants, display the role
...
+|Compliance viewer
+|View compliance reports.
+|
+
+|Create ARF report
+|Create compliance reports.
+|
+
+|Default role
+|View tasks and jobs invocations.
```
adamruzicka, line 57: While this is true in Satellite scenario (and maybe Orcharhino as well, I'm not sure what plugins are enabled there by default), in pure Foreman deployments one may not have remote execution or tasks plugins enabled and then they will not have those permissions. In vanilla Foreman without any plugins, the role shouldn't give any permissions, it is mostly a placeholder.

guides/common/modules/ref_predefined-roles.adoc
```diff
@@ -7,48 +7,140 @@ To view the exact set of permissions a predefined role grants, display the role
...
+|Site manager
+|View hosts, users, logs, and other items.
+
+Manage hosts, users, installation media, subnets, and other items.
+
+Update Puppet parametrs.
+|A restrained version of the Manager role.
+
+|System admin
+a|Edit global settings in *Administer* > *Settings*.
```
adamruzicka, line 110:
```suggestion
|Edit global settings in *Administer* > *Settings*.
```

I assume the leading `a` shouldn't be there?

guides/common/modules/ref_predefined-roles.adoc
```diff
@@ -7,48 +7,140 @@ To view the exact set of permissions a predefined role grants, display the role
...
+|Remote Execution User
+|Run remote execution jobs against hosts.
+|
+
+|Site manager
+|View hosts, users, logs, and other items.
+
+Manage hosts, users, installation media, subnets, and other items.
+
+Update Puppet parametrs.
```
adamruzicka, line 106: What are puppet parameters?

guides/common/modules/ref_predefined-roles.adoc
```diff
@@ -7,48 +7,140 @@ To view the exact set of permissions a predefined role grants, display the role
...
+|
+
+|Remote Execution User
+|Run remote execution jobs against hosts.
+|
+
+|Site manager
+|View hosts, users, logs, and other items.
+
+Manage hosts, users, installation media, subnets, and other items.
```
adamruzicka, line 104: Users with this role should be able to also view all the things they can manage. It is a little bit odd that subnets are mentioned in the list of things that can be managed, but not in the list of things that can be viewed.

guides/common/modules/ref_predefined-roles.adoc
asteflova: @maximiliankolb Can you please take a look and let me know what permissions this role provides in Orcharhino?

@adamruzicka It seems I should hide this permission for Foreman builds, is that right? Or am I misunderstanding something here?

maximiliankolb: orcharhino _always_ contains the remote execution plug-in. The default role has "view_job_invocations" and "view_foreman_tasks" permissions.

ekohl: Katello depends on REX, so you can assume it's always present in `katello,orcharhino,satellite`. In `foreman-deb,foreman-el` it's optional.

adamruzicka: Doesn't it optionally depend on rex?

ekohl: It's in the gemspec: https://github.com/Katello/katello/blob/f9ce129c658d6405c5d3b47dcd5cb63b27417cff/katello.gemspec#L37

adamruzicka: Indeed, I stand corrected.

asteflova: Thanks all for the input.

- For foreman-deb and foreman-el, the description now states that the permissions provided by Default role depend on plug-ins.
- For other builds, no change -- so View tasks and jobs invocations.

Does this work, or have I got it wrong?

guides/common/modules/ref_predefined-roles.adoc
```diff
@@ -7,48 +7,145 @@ To view the exact set of permissions a predefined role grants, display the role
...
-| System admin a|
-* Edit global settings in *Administer* > *Settings*.
-* View, create, edit and destroy users, user groups, and roles.
-* View, create, edit, destroy, and assign organizations and locations but not view resources within them.
-
-Users with this role can create users and assign all roles to them.
+
+By cloning this role and assigning an organization, you can delegate administration of that organization to a user.
+
+|Red{nbsp}Hat Access Logs
```
maximiliankolb, line 96: I did not find this role in Foreman 3.7. Does it depend on RH Insights plug-in?

guides/common/modules/ref_predefined-roles.adoc
```diff
@@ -7,48 +7,145 @@ To view the exact set of permissions a predefined role grants, display the role
...
-| Viewer | A passive role that provides the ability to view the configuration of every element of the {Project} structure, logs, reports, and statistics.
-| View hosts | A role that can only view hosts.
-| Virt-who Manager | A role with full virt-who permissions.
-| Virt-who Reporter | Upload reports generated by virt-who to {Project}.
-It can be used if you configure virt-who manually and require a user role that has limited virt-who permissions.
-| Virt-who Viewer | View virt-who configurations.
-Users with this role can deploy virt-who instances using existing virt-who configurations.
+
+|Tasks manager
+|View and edit {Project} tasks.
```
maximiliankolb, line 125: I'd either go with "tasks in {Project}" or just drop "{Project}" because it's kinda obvious.

Same for (all) other instances.

guides/common/modules/ref_predefined-roles.adoc
```diff
@@ -7,48 +7,145 @@ To view the exact set of permissions a predefined role grants, display the role
...
+|Access Insights Admin
+|Add and edit Insights rules.
+|
+
+|Access Insights Viewer
+|View Insight reports.
+|
+
+|Ansible Roles Manager
+|Play roles on hosts and host groups.
```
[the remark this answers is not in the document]
asteflova: The permissions in the Ansible Roles Manager role include these two: _play_roles_on_hostgroup_ and _play_roles_on_host_. So I just want to check @maximiliankolb should I really replace _host groups_ with _host collections_? It won't match the name of the permission.

guides/common/modules/ref_predefined-roles.adoc
asteflova: I'll drop it, thanks.

guides/common/modules/ref_predefined-roles.adoc
asteflova: It looks like the role is called Auditor now. At least judging from what I can see in the web UI. I'll update the list.

asteflova:
Hi team @adamlazik1 @AkshayGadhaveRH @bangelic @Lennonka @maximiliankolb @mjivraja What are your general opinions on listing the default permissions in the guide?

Predefined Roles Available in Satellite: https://theforeman-foreman-documentation-preview-pr-2346.surge.sh/nightly/Administering_Project/index-satellite.html#Predefined_Roles_admin

The roles are described in pretty good detail in the product itself (web UI), and the vast majority of what is in the guide just duplicates the descriptions in the web UI. I see almost no added value, just a maintenance burden for us. 

I'm not saying I want to delete that table (I mean I'd love to... but I'm not ready to say it just yet :D). At this point, I'd like to understand the conventions of the docs team: Is it a standard procedure to include information like this? Do most team members find it useful? Does anyone share my sentiment?

adamlazik1:
Logically it would make sense to not include this in the docs since it's described in the UI, however; I am afraid that as soon as you do this we get a bug report asking for it to be added back :smile:. So I advise to keep it.

mjivraja:
I agree with @Lennonka's view and also with @adamlazik1. Until WebUI, CLI, and API descriptions do not get accurate & complete, we should keep it untouched.

guides/common/modules/ref_predefined-roles.adoc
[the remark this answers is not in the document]
maximiliankolb: You're right. Going to Configure > Host Groups; you can click "Run all Ansible roles" for a host group. Please dismiss my comment.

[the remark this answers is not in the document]
asteflova: Thanks for checking! I'll just do s/Play roles/Run Ansible roles/ then.

asteflova:
Thanks for chiming in, everyone. [@asteflova suppresses the urge to delete the table]

I have a different proposal then: What if I cross-checked the table with the list of permissions in Project, and delete those table rows than just duplicate the descriptions in Project? Consider this example:

* Documentation: Boot disk access -- Download the boot disk.
* Web UI: Boot disk access -- Role granting permissions to download bootdisks

If I delete duplicates like this, then the guide would only contain permissions whose descriptions provide extra value. Would you be okay with that? @Lennonka @mjivraja  @adamlazik1

Lennonka:
I'm sorry, but that would make the UX even worse IMHO.
Not just that the WebUI descriptions are incomplete and s*ck, and therefore the user has to read documentation, but then the reference in the docs is incomplete on top of that?

How about making the descriptions in the docs more complete, useful, comprehensive, and human readable (unlike viewing the list of permissions per role) instead?

The other option would be to stop working on the docs and improve the descriptions in the WebUI directly. It's still an option, but it is the longer route and it can be done only for +/- newer releases.

asteflova:
> I'm sorry, but that would make the UX even worse IMHO. Not just that the WebUI descriptions are incomplete and s*ck, and therefore the user has to read documentation, but then the reference in the docs is incomplete on top of that?

Well, to be honest, the documentation is incomplete as well. There are a few permissions missing in the guide already. And it has probably been incomplete for some time without anyone noticing -- because duplication like this is very hard (impossible?) to maintain.

> How about making the descriptions in the docs clearer, more complete, useful, comprehensive, and human readable (unlike viewing the list of permissions per role) instead?

At this point, that would mean copy-pasting content from the web UI into the guide. And keeping it up-to-date. From the current state of things, it has not been kept up-to-date. Which is why I expect that to be a problem in the future as well (it usually is...).

More importantly, the section in question currently admits that it isn't even designed to provide complete information. It directs users to the web UI for complete information. So in my opinion, the proposal to remove obvious duplicates is not even that radical. It's in line with the expectations that section sets anyway.

guides/common/modules/ref_predefined-roles-available-in-project.adoc
```diff
@@ -1,6 +1,5 @@
-[id="Predefined_Roles_{context}"]
+[id="Predefined_Roles__Available_in_Project_{context}"]
```
Lennonka, line 1:
```suggestion
[id="Predefined_Roles_Available_in_{project-context}_{context}"]
```
We actually have a variable you can use in IDs and links! :)

asteflova: Thanks! I was looking for {projectid}. I saw other variables designed for IDs to use that syntax. Will do my best to remember this one.

asteflova:
Hi all, any further comments?

Lennonka (approved):
LGTM

asteflova:
That's very cool @ekohl , to be able to propose changes to descriptions in the web UI like that.

I checked the web UI descriptions for predefined roles one more time and I have to say that the only thing I personally can see that could use fixing is that some descriptions end with a period and some don't :D And I don't think that's worth drafting a PR. Other than that, I don't feel like imposing our rather strict documentation style conventions to the web UI.

But if I spot a problem any time in the future, it's good to know that there may be a way to propose a fix right away.

To be honest, I'm very happy with what I've seen in the web UI so far -- that's why I'd like to keep linking to it for more details from the docs.

asteflova:
... and just to be sure, I checked the BZ report that spurred this PR. The BZ requested to clarify a description for a predefined role in the documentation; that same role is actually described pretty clearly in the web UI. So no problem there either, I think.

adamlazik1:
Hi @asteflova, is this ready to be merged?

asteflova:
I think so @adamlazik1. There is no formal SME ack on the final version, but considering that all content that remains in this section has been there before or has been written based on the descriptions in the web UI, I don't think we need to wait anymore.

If you're okay with that ^, go ahead and merge.

ekohl:
> I checked the web UI descriptions for predefined roles one more time and I have to say that the only thing I personally can see that could use fixing is that some descriptions end with a period and some don't :D

Consistency. I love it, but sadly it's not as common as I'd like.

Perhaps this is something we should note in the developer docs. https://github.com/theforeman/foreman/blob/develop/developer_docs/how_to_create_a_plugin.asciidoc#adding-roles describes adding roles. It doesn't even mention adding a description, but if it did it'd good to mention that we should either always add a period or never.

> And I don't think that's worth drafting a PR. Other than that, I don't feel like imposing our rather strict documentation style conventions to the web UI.

I disagree: good style conventions are also good for a consistent UX. There's a reason you do it in docs. Small PRs like that are actually very nice as a maintainer and usually easy to merge.

adamlazik1:
Oof, I forgot to squash, my apologies

adamlazik1:
Merged and cherry-picked:
   8ab48f9ee..ba3bf823d  3.7 -> 3.7

asteflova:
> I disagree: good style conventions are also good for a consistent UX. There's a reason you do it in docs. Small PRs like that are actually very nice as a maintainer and usually easy to merge.

It's just that our style conventions are very robust (the IBM style guide alone has ~500 pages... and it's not the only one we use). So I wouldn't dare to ask any web UI content developer to adhere to them.

But it could actually be fun to try to submit PRs against the foreman project itself. So okay, I'll add fixing up role descriptions to my list of things to do on slow and lazy afternoons :)

ekohl:
> But it could actually be fun to try to submit PRs against the foreman project itself. So okay, I'll add fixing up role descriptions to my list of things to do on slow and lazy afternoons :)

Sounds great. Let me know if you need any help. To me the power of open source is still that you can fix what bothers you :)

guides/common/modules/ref_predefined-roles-available-in-project.adoc
```diff
@@ -0,0 +1,56 @@
+[id="Predefined_Roles_Available_in_{project-context}_{context}"]
+= Predefined Roles Available in {Project}
+
+The following table provides an overview of permissions that predefined roles in {Project} grant to a user.
+
+For a complete set of predefined roles and the permissions they grant, log in to {ProjectwebUI} as the privileged user and navigate to *Administer* > *Roles*.
```
maximiliankolb, line 6: Should be "ProjectWebUI"; but attributes are IMHO case insensitive: https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-base.adoc#L109

cc @asteflova

asteflova: Oops, thanks, I submitted https://github.com/theforeman/foreman-documentation/pull/2397

PR #2372, opened 2023-08-17
Document enabling REX ssh mode for upstream

adamlazik1:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2216976

* [ ] Foreman 3.6/Katello 4.8


adamlazik1:
@jeremylenz please verify the text and please confirm this only applies to foreman 3.6 and 3.7.

@maximiliankolb this instruction was only required to be added to upstream since it comes enabled by default on satellite. Does it come enabled by default on orcharhino too?

maximiliankolb (approved):
Works for me if you don't cherry-pick to 3.5 :+1:

guides/common/modules/con_transport-modes-for-remote-execution.adoc
```diff
@@ -9,6 +9,15 @@ The SSH service must be enabled and active on the target hosts.
 The remote execution {SmartProxy} must have access to the SSH port on the target hosts.
 Unless you have a different setting, the standard SSH port is 22.
 
+ifdef::foreman-el,foreman-deb[]
+To use the `ssh` mode, run the following command on your {ProjectServer}:
```
adamruzicka, line 13: Wouldn't `{SmartProxy}` be better here?

jeremylenz, line 13:
```suggestion
To enable remote execution, run the following command on your {ProjectServer}:
```

"To use the ssh mode" implies that this command only allows use of the ssh mode, when actually it's required for both pull and ssh.

jeremylenz (commented):
This belongs in the "Installing Smart Proxy with Content" guide, not here in Managing Hosts. It should be a section in the chapter "Performing Additional Configuration on Smart Proxy Server",  inserted before the section "Configuring Remote Execution for Pull Client."

I'd say it applies to both Foreman and Katello builds.

jeremylenz:
> confirm this only applies to foreman 3.6 and 3.7

Before `ssh` was renamed to `script` (I'm assuming that was in 3.6?), the command was 

```
--enable-foreman-proxy-plugin-remote-execution-ssh
```

With that older command, this would be relevant all the way back. That may be easier in a separate PR, or you may not want to bother; up to you of course!

ekohl (changes requested):
Is this really the best approach? Wouldn't it be more in line to have a procedure to enable REX with SSH mode and one with pull mode? On line 28 you link to "configuring in pull mode" so perhaps this chapter should be

* Explain SSH (push) mode
* Link to how to set up SSH mode
* Explain MQTT (pull) mode
* Link to how to set up MQTT mode

adamlazik1:
Thank you all for review.
Applied changes:
* Created new section in **Performing Additional Configuration on SmartProxyServer** just before the **Configuring Remote Execution for Pull Client** section.
* Made the section visible for Foreman and Katello.
* Linked the existing chapter about REX for users who wish to learn more.

@ekohl does this cover your suggestions?
CC @jeremylenz 

After the PR is finalized I will create a second PR for lower versions.

ekohl:
> @ekohl does this cover your suggestions?

I think it does, but right now the rendering fails so I can't review the final result

> After the PR is finalized I will create a second PR for lower versions.

Note that upstream 3.6 is the oldest supported release. Documenting older versions isn't needed. The only benefit you would have is smaller chance for cherry picks conflicts, but I think that's not worth it here.

adamlazik1:
It appears tht I used non-existent attribute which failed the build. That is now fixed and the preview is available. @ekohl feel free to review.
@jeremylenz does this work for you?

guides/common/modules/proc_enabling-remote-execution.adoc
```diff
@@ -0,0 +1,13 @@
...
+
+Use this procedure to enable remote execution on your {SmartProxyServer}.
+To learn more about remote execution, see {ManagingHostsDocURL}Configuring_and_Setting_Up_Remote_Jobs_managing-hosts[Configuring and Setting Up Remote Jobs] in _{ManagingHostsDocTitle}_.
+
+.Procedure
+* Enable remote execution with {foreman-installer}:
++
+[options="nowrap", subs="quotes,verbatim,attributes"]
+----
+{foreman-installer} --enable-foreman-proxy-plugin-remote-execution-script
```
ekohl, line 12: In all other commands we use the console style
```suggestion
# {foreman-installer} --enable-foreman-proxy-plugin-remote-execution-script
```

guides/common/modules/proc_enabling-remote-execution.adoc
```diff
@@ -0,0 +1,13 @@
+[id="enabling-remote-execution_{context}"]
+= Enabling Remote Execution
+
+Use this procedure to enable remote execution on your {SmartProxyServer}.
+To learn more about remote execution, see {ManagingHostsDocURL}Configuring_and_Setting_Up_Remote_Jobs_managing-hosts[Configuring and Setting Up Remote Jobs] in _{ManagingHostsDocTitle}_.
+
+.Procedure
```
ekohl, line 7: For `foreman-el` and `foreman-deb` think needs a prerequisite that REX is already enabled on the Server. Looking at the Installer Server guide I don't see a matching instruction you can link to. There it needs to be `{foreman-installer} --enable-foreman-plugin-remote-execution` or `{foreman-installer} --enable-foreman-plugin-remote-execution --enable-foreman-proxy-plugin-remote-execution-script` if you also want to run jobs from the Server itself.

For `katello` it's a dependency so already present. However, you may still need to enable the foreman-proxy-plugin on Server if you want to run jobs on it.

ekohl (changes requested):

adamlazik1:
Thank you for review. Added the console style commands and a prerequisite for foreman.

guides/common/modules/proc_enabling-remote-execution.adoc
```diff
@@ -0,0 +1,24 @@
+[id="enabling-remote-execution_{context}"]
+= Enabling Remote Execution
+
+Use this procedure to enable remote execution on your {SmartProxyServer}.
+To learn more about remote execution, see {ManagingHostsDocURL}Configuring_and_Setting_Up_Remote_Jobs_managing-hosts[Configuring and Setting Up Remote Jobs] in _{ManagingHostsDocTitle}_.
+
+ifdef::foreman-el,foreman-deb[]
+.Prerequisite
+* You have enabled the remote execution plug-in on your {ProjectServer}.
```
ekohl, line 9: Would it make sense to introduce a similar part in the Installing Server guide? Perhaps even link there instead of inlining?

ekohl (approved):
We can further improve later if you want, or do it now. But this can be merged now.

guides/common/modules/proc_enabling-remote-execution.adoc
adamlazik1: I suppose it would be good to add this section; however; the entire chapter dealing with remote execution is currently in [Managing Hosts](https://docs.theforeman.org/nightly/Managing_Hosts/index-foreman-el.html#Configuring_and_Setting_Up_Remote_Jobs_managing-hosts), so I think it would be the best to add it there. I am thinking to add it after the **Remote Execution Workflow** chapter. What do you think?

ekohl: A good question, and perhaps it comes down to preference. I think there is value in having it in the install guides (both server and proxy). Even if they're small, I really like cross references. But if you it for one plugin, should we do it for all? For example, you could make the same argument for OpenSCAP.

Perhaps @Lennonka or @asteflova can weigh in here since it's less about technical details (my specialty) and more about a bigger content strategy.

jeremylenz (approved):

guides/common/modules/proc_enabling-remote-execution.adoc
adamlazik1: I see this could turn into a work of potentially large scope. I will merger this PR and open an issue where we can discuss the next steps.

adamlazik1:
Merged and cherry-picked:
   a6c282d74..6f4f8ba64  3.6 -> 3.6
   b7cc28839..8aed38c0f  3.7 -> 3.7

PR #2374, opened 2023-08-18
Add Updating Guide to Upstream

mjivraja:

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/web/content/js/versions.js
+++ b/web/content/js/versions.js
@@ -24,6 +24,10 @@ const navVersions = [
             "path": "Installing_Proxy"
           },
           {
+            "title": "Upgrading Foreman to Nightly",
+            "path": "Upgrading_Project"
+          },
+          {
             "title": "Deploying Foreman on AWS",
             "path": "Deploying_Project_on_AWS"
           },
@@ -124,7 +128,7 @@ const navVersions = [
             "path": "Installing_Proxy"
           },
           {
-            "title": "Upgrading Katello",
+            "title": "Upgrading Katello to Nightly",
             "path": "Upgrading_Project"
           },
           {
@@ -240,7 +244,7 @@ const navVersions = [
             "path": "Installing_Proxy"
           },
           {
-            "title": "Upgrading Katello",
+            "title": "Upgrading Katello to 4.10",
             "path": "Upgrading_Project"
           },
           {
@@ -328,7 +332,7 @@ const navVersions = [
             "path": "Installing_Proxy"
           },
           {
-            "title": "Upgrading Foreman",
+            "title": "Upgrading Katello to 4.9",
             "path": "Upgrading_Project"
           },
           {


web/content/js/versions.js
```diff
@@ -266,6 +278,14 @@ const navVersions = [
           {
             "title": "Application Centric Deployment",
             "path": "Deploying_Hosts_AppCentric"
+          },
```
maximiliankolb, line 281: I think you should move "updating Foreman" as "Updating Katello" below line 229 and drop "Upgrading Foreman" because it's a duplicate.

Maybe also rename 227 to "Upgrading Katello"?

mjivraja: Addressed :+1:

Kept both guides at last.

web/content/js/versions.js
```diff
@@ -178,6 +182,14 @@ const navVersions = [
...
```
maximiliankolb, line 185: Change itself looks good IMHO. Could you please move both items to line 126?

mjivraja: Addressed :+1:
Out of my curiosity, I would like to learn how we design the sequence.

web/content/js/versions.js
```diff
@@ -203,6 +215,14 @@ const navVersions = [
             "title": "Quickstart Guide",
             "path": "Quickstart"
           },
+          {
+            "title": "Updating Katello",
```
ekohl, line 219: I'd say installing is more important than upgrading/updating, so the previous position was correct.

ekohl (changes requested):
The order is inconsistent. For Foreman you have it all the way at the end while for Katello you have it before installing.

I think the order should be:
* Quickstart
* Install
* Update
* Upgrade

And I'd also make it clear what update vs upgrade is, so say `Upgrade to Foreman X.Y`.

That also makes me wonder: why do we list the upgrade guide for nightly? You can't upgrade nightly to something newer.

web/content/js/versions.js
maximiliankolb: My point was: it does not make sense to delete "upgrading" in line 126 and then add updating + upgrading in line 218+.

mjivraja: Addressed. I hope I understood correctly.

mjivraja:
@maximiliankolb @ekohl is the sequence correct now?

maximiliankolb (approved):
Yes, LGTM! Handing over to Ewoud to ACK/merge.

web/content/js/versions.js
```diff
@@ -212,7 +224,11 @@ const navVersions = [
             "path": "Installing_Proxy"
           },
           {
-            "title": "Upgrading Foreman",
+            "title": "Updating Katello",
+            "path": "Updating_Project"
+          },
+          {
+            "title": "Upgrading Katello",
```
ekohl, line 231: Can we be explicit what Upgrading is?
```suggestion
            "title": "Upgrading Katello to nightly",
```

ekohl (changes requested):
I still think we shouldn't show `Upgrading_Project` on nightly because for the reader it's not going to be useful. You can't upgrade nightly to something else.

We also have pages like https://docs.theforeman.org/release/3.7/ so please also update this part:
https://github.com/theforeman/foreman-documentation/blob/35f4fb8d315ba36af4cb7839790f767ba75a2b86/web/content/release-3.7.adoc#L20-L23

mjivraja:
> We also have pages like https://docs.theforeman.org/release/3.7/ so please also update this part:

Addressed :+1:

web/content/release-3.7.adoc
```diff
@@ -20,4 +20,4 @@ Use the top menu bar for navigation for all guides.
 === Upgrading
 
 * link:/{FOREMAN_VER}/Release_Notes/index-katello.html[Release notes for Katello on RHEL/CentOS]
-* link:/{FOREMAN_VER}/Upgrading_and_Updating/index-katello.html[Upgrading and Updating Katello on RHEL/CentOS]
+* link:/{FOREMAN_VER}/Upgrading_Project/index-katello.html[Upgrading Katello on RHEL/CentOS]
```
ekohl, line 23: I had expected this to also match the navigation and mention an upgrade is to 3.7. Then also add a link to the `Updating_Project` guide. Perhaps also update the header.

web/content/js/versions.js
```diff
@@ -212,7 +220,11 @@ const navVersions = [
             "path": "Installing_Proxy"
           },
           {
-            "title": "Upgrading Foreman",
+            "title": "Updating Katello",
+            "path": "Updating_Project"
+          },
+          {
+            "title": "Upgrading Katello to nightly",
```
ekohl, line 227: And then this is
```suggestion
            "title": "Upgrading Katello to 3.7",
```

ekohl (changes requested):
I think I've made a mistake in my previous review and assumed the guide in x.y is to upgrade to x.(y+1) but it's the other way around. So please add the links back in nightly. Apologies.

web/content/js/versions.js
mjivraja: Are you sure version `3.7` is also applicable to Katello alongside `Foreman`? I guess it should be `4.9`

But, my query is why do we write `3.7` when we are already there? It should be `nightly` or higher version IMHO.

ekohl: You're right, 4.9 since it's Katello. But it's 4.9 since this is the Foreman 3.7 set of links.

mjivraja: I hope I understood correctly.

web/content/js/versions.js
```diff
@@ -23,6 +23,14 @@ const navVersions = [
             "title": "Installing Smart Proxy",
             "path": "Installing_Proxy"
           },
+          {
+            "title": "Updating Foreman",
+            "path": "Updating_Project"
+          },
+          {
+            "title": "Upgrading Foreman",
```
ekohl, line 31:
```suggestion
            "title": "Upgrading Foreman to nightly",
```

web/content/js/versions.js
```diff
@@ -123,6 +131,10 @@ const navVersions = [
             "title": "Installing Smart Proxy with Content",
             "path": "Installing_Proxy"
           },
+          {
+            "title": "Updating Katello",
```
ekohl, line 135:
```suggestion
            "title": "Updating Katello to nightly",
```

web/content/release-3.7.adoc
```diff
@@ -17,7 +17,14 @@ Use the top menu bar for navigation for all guides.
...
+
+* link:/{FOREMAN_VER}/Updating_Project/index-katello.html[Updating Katello on RHEL/CentOS]
+
 === Upgrading
 
+* link:/{FOREMAN_VER}/Upgrading_Project/index-katello.html[Upgrading Katello to 4.9 on RHEL/CentOS]
+
+=== Release Note
+
 * link:/{FOREMAN_VER}/Release_Notes/index-katello.html[Release notes for Katello on RHEL/CentOS]
```
ekohl, line 30: I'd keep this as one section
```suggestion
=== Updating & upgrading

* link:/{FOREMAN_VER}/Release_Notes/index-katello.html[Release notes for Katello on RHEL/CentOS]
* link:/{FOREMAN_VER}/Updating_Project/index-katello.html[Updating Katello on RHEL/CentOS]
* link:/{FOREMAN_VER}/Upgrading_Project/index-katello.html[Upgrading Katello to 4.9 on RHEL/CentOS]
```

ekohl (changes requested):
I wasn't explicit, but I meant that the various release-* files in [web/content](https://github.com/theforeman/foreman-documentation/tree/master/web/content) should be updated, just as you did with the main navigation. 3.7 was just one example.

web/content/js/versions.js
mjivraja: I wasn't sure to modify this until #2382 
Addressed :+1:

ekohl (changes requested):
Please also update https://github.com/theforeman/foreman-documentation/blob/master/web/content/release-nightly.adoc

mjivraja:
> Please also update https://github.com/theforeman/foreman-documentation/blob/master/web/content/release-nightly.adoc

Addressed :+1:

web/content/release-nightly.adoc
```diff
@@ -29,8 +29,5 @@ Use the top menu bar for navigation for all guides.
 * link:/{FOREMAN_VER}/Release_Notes/index-foreman-el.html[Release notes for Foreman on RHEL/CentOS]
 * link:/{FOREMAN_VER}/Release_Notes/index-foreman-deb.html[Release notes for Foreman on Debian/Ubuntu]
 * link:/{FOREMAN_VER}/Release_Notes/index-katello.html[Release notes for Katello on RHEL/CentOS]
-
-// Upgrading guides are not ready for non-Katello
```
ekohl, line 33: The reason I previously had this visual separation was indicate grouping as well as a TODO. It also made sure that the order was already set: you only needed to uncomment them once they were done.

I wonder what others think. Does it make sense to keep that in?

mjivraja: In `3.7` we don't have these. It should be identical to both places, no? IMO, guides are ready for `non-Katello`

ekohl: I now see https://github.com/theforeman/foreman-documentation/pull/2334 wasn't merged yet, but I consider that an oversight. After that the guides are properly hidden.

mjivraja:
> Taking a step back: we have 2 ways to navigate. You have per version pages (https://docs.theforeman.org/release/nightly & https://docs.theforeman.org/release/3.7) and the top bar. These should be consistent.
> 
> If you go to nightly, you will see we link to `Foreman on EL`, `Foreman on Debian` and `Katello on EL`. If you go to 3.7 we only link to `Katello on EL`. The version specific pages reflect this.
> 
> In this PR you are adding the updating and upgrading guides to the top bar navigation (for nightly) but not to the version specific page.
> 
> By now we've also branched Foreman 3.8 so this needs to be rebased to reflect that. I did notice that [ddba2d7](https://github.com/theforeman/foreman-documentation/commit/ddba2d718d91ade9bdfd3375f875cc883993449a) didn't do it right and included all links. This should be updated so it's more like 3.7 and #2410 does that.

The `version page` and `top bar` of `nightly`, `3.8`, and `3.7` should be the same?

I understand that the guides that are visible in `nightly` including `top-bar` and `version pages` should be consistent with `3.8` and `3.7` (which is not right now). If this is correct, the #2410 is confusing as many guides from the `top bar` are removed. I am considering that the `Updating` and `Upgrading` guides are ready to consume for all 3 - `foreman-el`, `foreman-deb`, and `katello`. 

If only `3.8` should be kept consistent with `3.7`, the `version page` and `top bar` of the same i.e. `3.7` requires some modification as I rebased this branch with master. At present, `3.8` indicates that the `Updating` and `Upgrading` guides are not ready to consume at all. 

Now, I am even clear about this. Needs your final call about the guides we want to keep right now.

ekohl:
> The `version page` and `top bar` of `nightly`, `3.8`, and `3.7` should be the same?

The slightly longer version: the version page is a subset of the top bar. We list the a few guides, but if those guides exist they should be present in both places. The top bar does list all other guides as well.

It may be a good idea to change this in the future, but this requires some thought what a good way of presenting it is.

> I am considering that the `Updating` and `Upgrading` guides are ready to consume for all 3 - `foreman-el`, `foreman-deb`, and `katello`.

I didn't check before, but `foreman-maintain` in upstream doesn't implement updating to .z versions so all the Updating guides are invalid. We should stop rendering those.

At least `foreman-deb` upgrading is not ready, and the guide clearly renders that: https://docs.theforeman.org/nightly/Upgrading_Project/index-foreman-deb.html & https://docs.theforeman.org/3.8/Upgrading_Project/index-foreman-deb.html aren't rendered. The reason is that it doesn't even tell users to update their repositories so the whole workflow is broken.

`foreman-el` upgrading may work, but it does use `foreman-maintain` and there is no instruction ever telling the user to install that. In downstream it's a dependency that gets installed automatically, but that's not the case for upstream Foreman. It is true for katello.

> Now, I am even clear about this. Needs your final call about the guides we want to keep right now.

We should make sure all the `Updating` guides (including in nightly) aren't rendered anymore and don't link to it. Simply because it relies on instructions that do not work in upstream. Then also cherry pick that to all branches where `Updating` guides were introduced.

Then for the `Upgrading` guides I'm unsure. Going guide by guide:

* `foreman-deb`: manual needs work, can't be shown anywhere
* [foreman-el](https://docs.theforeman.org/nightly/Upgrading_Project/index-foreman-el.html)
  * FIPS mode: AFAIK we don't support that at all, so the note could be removed. This is probably something for multiple guides so not blocking
  * Upgrade process itself: stop services, make backup and start services feels redundant. Isn't stop & start part of the backup procedure? Should we link to the backup manual as well? Probably also not blocking.
  * The Foreman upgrade is usually much quicker than 1 - 2 hours. In my experience (which is with smaller installations) I think 10 - 30 minutes might be closer. Again, not blocking.
  * The instructions about DNS and DHCP: I'd be in favor of dropping them. We shouldn't encourage unmanaged setups IMHO.
  * It assumes foreman-maintain is present, but we never told users about this
  * External DB: we never talk about it elsewhere. That's why I started https://github.com/theforeman/foreman-documentation/pull/2396
* [katello](https://docs.theforeman.org/nightly/Upgrading_Project/index-katello.html)
  * The user is instructed to let content hosts consume directly from the internet. Really they should sync in the content and distribute the packages via content management.
  * Same notes about backups and DNS & DHCP

So overall I'm not too too pleased with the the upgrading guides, but besides the Katello content hosts thing I don't think they really are wrong or missing steps.

mjivraja:
Completely removed `Updating Guide` from `top bar` and `version page`. `Upgrading Guide` is now limited to Katello based on your latest comment.

web/content/release-3.8.adoc
```diff
@@ -10,14 +10,23 @@ Use the top menu bar for navigation for all guides.
 
 === Quickstart
 
+* link:/{FOREMAN_VER}/Quickstart/index-foreman-el.html[Foreman on RHEL/CentOS Quickstart Guide]
+* link:/{FOREMAN_VER}/Quickstart/index-foreman-deb.html[Foreman on Debian/Ubuntu Quickstart Guide]
```
ekohl, line 14: These guides are not ready: https://docs.theforeman.org/3.8/Quickstart/index-foreman-el.html & https://docs.theforeman.org/3.8/Quickstart/index-foreman-deb.html say so.

web/content/release-3.8.adoc
```diff
@@ -10,14 +10,23 @@ Use the top menu bar for navigation for all guides.
...
 * link:/{FOREMAN_VER}/Quickstart/index-katello.html[Katello on RHEL/CentOS Quickstart Guide]
 
 === Installation
 
+* link:/{FOREMAN_VER}/Installing_Server/index-foreman-el.html[Installing Foreman on RHEL/CentOS]
+* link:/{FOREMAN_VER}/Installing_Server/index-foreman-deb.html[Installing Foreman on Debian/Ubuntu]
```
ekohl, line 20: Same here.

web/content/release-3.7.adoc
```diff
@@ -10,14 +10,23 @@ Use the top menu bar for navigation for all guides.
 
 === Quickstart
 
+* link:/{FOREMAN_VER}/Quickstart/index-foreman-el.html[Foreman on RHEL/CentOS Quickstart Guide]
+* link:/{FOREMAN_VER}/Quickstart/index-foreman-deb.html[Foreman on Debian/Ubuntu Quickstart Guide]
```
ekohl, line 14: These guides are not ready: https://docs.theforeman.org/3.7/Quickstart/index-foreman-el.html & https://docs.theforeman.org/3.7/Quickstart/index-foreman-deb.html

ekohl (changes requested):
For now I've opened https://github.com/theforeman/foreman-documentation/pull/2416 because I couldn't think of how to explain it better than I've already tried and we don't want a broken navigation for too long. I've also opened https://github.com/theforeman/foreman-documentation/pull/2417 to remove the old guide.

For the future: if you want to understand what will be rendered in the navigation, you can use the script (https://github.com/theforeman/foreman-documentation/blob/master/web/show_versions) or render the website (follow https://github.com/theforeman/foreman-documentation/tree/master/web#readme).

The guides now explicitly say themselves whether they are ready or not so please open the link when you think it should be included.

web/content/js/versions.js
```diff
@@ -124,7 +128,7 @@ const navVersions = [
             "path": "Installing_Proxy"
           },
           {
-            "title": "Upgrading Katello",
+            "title": "Upgrading Katello to Nightly",
```
maximiliankolb, line 131: either both "nightly" or "Nightly" please.

mjivraja:
@ekohl Would you mind cherry-picking?

Lennonka:
I'm cherry-picking it

ekohl:
This is only touching the navigation, so cherry-picking is not needed. For clarification: the actual updating guide is not usable in upstream (because the commands are not implemented for upstream) so 9065292a2175191decc96df29e9a43ddfe638171 hides it. I did cherry pick that.

ekohl:
Oh, on merge I changed the commit message to better cover the actual change: 7c5583c8436bf9282998e4ffc5578beab9e80aab is that commit. I've now also changed the PR title to match.

PR #2382, opened 2023-08-22
Reflect the upgrade target in upgrading title

ekohl:
This is useful if you have multiple tabs open or link to it from something like Discourse.

I'm only selecting the support stable releases, but I wonder if we should pick it to all branches.

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/attributes-titles.adoc
+++ b/guides/common/attributes-titles.adoc
@@ -24,7 +24,7 @@
 :ReleaseNotesDocTitle: Release Notes
 :TuningDocTitle: Tuning Performance of {ProjectName}
 :UpdatingDocTitle: Updating {ProjectName}
-:UpgradingDocTitle: Upgrading {ProjectName}
+:UpgradingDocTitle: Upgrading {ProjectName} to {ProjectVersion}
 
 // Not upstreamed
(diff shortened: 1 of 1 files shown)


maximiliankolb (approved):

Lennonka:
@AkshayGadhaveRH @mjivraja Opinions?

mjivraja:
> I'd like to hear opinions from @AkshayGadhaveRH and @mjivraja before I say anything.

My concern is we are reading `Upgrade Project to Nightly` even though the document is of Nightly. End-users will always be consuming n-1 document when planning to Upgrade, right?

mjivraja:
> Yes, but the document does describe that. Quoting the chapter `Upgrade Paths`:

Alright, let's move forward. (I learned that the new branch for the new foreman version will inherit this change and thus, it's good against making changes to the new version everytime. We have `Upgrade Paths` to avoid confusion.)

mjivraja (approved):

Lennonka (approved):
LGTM

Please, do not cherry-pick to older releases, because the change in the doc title would require an update of the title in Pantheon. CP to 6.14 is cool. cc @mjivraja @AkshayGadhaveRH

PR #2387, opened 2023-08-23
Remove content view Dependency Solving Default

bangelic:
The setting for **content view Dependency Solving
Default** is a dead
setting so the procedure was removed from the Content Management Guide.
The procedures for both the **Enabling Dependency
Solving Globally**
section and the **Enabling Dependency Solving Per
Content View** section
were in the incorrect modules so those were switched out before the
procedure for **Enabling Dependency Solving Globally** was removed from
the assembly.

* [x] Foreman 3.6/Katello 4.8


guides/common/modules/proc_enabling-dependency-solving-per-content-view.adoc
```diff
@@ -1,8 +1,8 @@
-[id="enabling-dependency-solving-globally_{context}"]
-= Enabling Dependency Solving Globally
+[id="enabling-dependency-solving-per-content-view_{context}"]
+= Enabling Dependency Solving Per Content View
```
jeremylenz, line 2: Now that the global setting isn't available, "per" doesn't make as much sense.
```suggestion
[id="enabling-dependency-solving-for-a-content-view_{context}"]
= Enabling Dependency Solving for a Content View
```

guides/common/modules/proc_enabling-dependency-solving-per-content-view.adoc
```diff
@@ -1,8 +1,8 @@
...
 
 You can enable dependency solving for a Content View or by default for your whole {Project} environment.
```
jeremylenz, line 4: This is no longer true. You can only enable it for individual content views.

jeremylenz (commented):
Is there also somewhere that covers enabling depsolving while creating a content view?

guides/common/modules/proc_enabling-dependency-solving-per-content-view.adoc
maximiliankolb: "no longer": Is this true for Foreman 3.7 all the way down to Foreman 3.1? @jeremylenz

jeremylenz: Good point! The setting removal isn't until Katello 4.10/Satellite 6.15. It would still be available in older versions that offer the old content view UI to create content views, but I don't remember when that was discontinued. 

Regardless, I think it's still coherent if we just change the sentence to apply to individual content views and don't mention the global setting.

bangelic:
> Is there also somewhere that covers enabling depsolving while creating a content view?

Not that I can see in the Content Management Guide.

guides/common/modules/proc_enabling-dependency-solving-for-a-content-view.adoc
```diff
@@ -1,7 +1,7 @@
-[id="enabling-dependency-solving-per-content-view_{context}"]
-= Enabling Dependency Solving Per Content View
+[id="enabling-dependency-solving-for-a-content-view_{context}"]
+= Enabling Dependency Solving for a Content View
 
-You can enable dependency solving for a Content View or by default for your whole {Project} environment.
+You can enable dependency solving for a Content View or by default for your {Project} environment.
```
jeremylenz, line 4:
```suggestion
You can enable dependency solving for a Content View.
```

I would leave off this part of the sentence. It's accurate, but only for the older versions where the setting was still there and did something. Leaving it out makes the sentence correct for all versions.

jeremylenz (approved):

mjivraja (approved):

asteflova:
Reviewed, looks good.

maximiliankolb:
I've just checked on Foreman 3.7 and 3.5; both still contain this setting:

foreman-3-7
foreman-3-5

Again: Does it make sense to cherry-pick this to released versions? Does it depend on using "new UI" How could I check this?

mjivraja:
image

Downstream (6.14)

maximiliankolb:
Yes, that's the procedure when creating a CV (on Foreman 3.7):

foreman-3-7-cv-create

@mjivraja Can you please check in Administer > Settings > Content?

mjivraja:
> @mjivraja Can you please check in Administer > Settings > Content?

It's the same as yours of upstream.

jeremylenz:
> Does it make sense to cherry-pick this to released versions? Does it depend on using "new UI" How could I check this?

Yes, the new CV UI ignores this setting. That's the reason we're removing it.

jeremylenz:
@sjha4 Do you remember what Katello version removed the old CV UI?

That's probably the version we should be cherry-picking back to. No sense documenting a setting that doesn't do anything.

sjha4:
6.11 is when we moved to the new CV UI..

jeremylenz:
> 6.11 is when we moved to the new CV UI..

Ok, so I recommend cherry-picking this back to Foreman 3.1 / Sat 6.11. The text will be accurate across all those versions.

maximiliankolb:
> 6.11 is when we moved to the new CV UI..

Do you know if this was default for upstream with Foreman 3.5+?

Does this also mean that it's currently a bug in Foreman 3.7/Katello 4.9 because the settings is still there but does not work/do anything?

jeremylenz:
> Does this also mean that it's currently a bug in Foreman 3.7/Katello 4.9 because the settings is still there but does not work/do anything?

Not a bug per se. The new Content View UI has been available for quite some time. The setting only affected the old Content View UI, and not the new. So it has been a "dead" setting since then.

(There was a bit of overlap where you could force it to use the old CV UI, in which case the setting remained relevant. But for the most part that setting has been dead since the new CV UI became the default.)

maximiliankolb (approved):
Thanks Brian, LGTM!

maximiliankolb:
   0861b65dc..33e182c11  3.8 -> 3.8
   40464c9a0..a54051409  3.7 -> 3.7
   8782698a0..c5e68b073  3.6 -> 3.6
   7cf011a04..7eb2f19c8  3.5 -> 3.5
   10dad1b5f..f9a4122d5  3.4 -> 3.4 (resolved a minor merge conflict here)
   ef79a4e0e..8f1f2993e  3.3 -> 3.3 (cherry-picked from 3.4)
   85a58d38b..7e30b2b46  3.2 -> 3.2 (cherry-picked from 3.4)
   dd2bda8f4..9b59c0e2c  3.1 -> 3.1 (cherry-picked from 3.4)

PR #2399, opened 2023-08-28
Add warning republishing metadata and mirroring

bangelic:
A warning has been added for republishing metadata for repositories
with Complete Mirroring because the risk is too great and causes issues.

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/con_mirroring-policies-overview.adoc
+++ b/guides/common/modules/con_mirroring-policies-overview.adoc
@@ -24,6 +24,6 @@ This mirroring policy is only available for Yum content.
 +
 [WARNING]
 ====
-{Team} does not recommend republishing metadata for repositories with Complete Mirror mirroring policy.
+Avoid republishing metadata for repositories with Complete Mirror mirroring policy.
 This also applies to Content Views containing repositories with the Complete Mirror mirroring policy.
 ====
--- a/guides/common/modules/con_mirroring-policies-overview.adoc
+++ b/guides/common/modules/con_mirroring-policies-overview.adoc
@@ -21,6 +21,9 @@ Complete Mirroring::
 Mirrors content as well as repodata.
 This is the fastest method.
 This mirroring policy is only available for Yum content.
-Republishing metadata in repositories with Complete Mirroring is high-risk.
-Hammer allows it using the `--force flag` command.
-This also applies to Content Views containing repositories with the Complete Mirroring policy.
++
+[WARNING]
+====
+{Team} does not recommend republishing metadata for repositories with Complete Mirror mirroring policy.
+This also applies to Content Views containing repositories with the Complete Mirror mirroring policy.
+====
--- a/guides/common/modules/con_mirroring-policies-overview.adoc
+++ b/guides/common/modules/con_mirroring-policies-overview.adoc
@@ -1,10 +1,10 @@
 [id="Mirroring_Policies_Overview_{context}"]
 = Mirroring Policies Overview
 
-Mirroring keeps the local repository exactly in synchronization with the upstream repository. 
+Mirroring keeps the local repository exactly in synchronization with the upstream repository.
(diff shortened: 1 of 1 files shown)


guides/common/modules/con_mirroring-policies-overview.adoc
```diff
@@ -21,3 +21,6 @@ Complete Mirroring::
 Mirrors content as well as repodata.
 This is the fastest method.
 This mirroring policy is only available for Yum content.
+Republishing metadata in repositories with Complete Mirroring is high-risk.
```
maximiliankolb, line 24: Does that even make sense? If you republish the metadata, it's hardly "complete mirroring" anymore. What happens if metadata is signed, for example with SUSE content?

_If_ we really want to document this, I'd _at least_ add a "WARNING" admonition or say that "{Team} does not recommend to republish metadata for repositories with 'complete mirroring'" or similar.

maximiliankolb (changes requested):

guides/common/modules/con_mirroring-policies-overview.adoc
sjha4: That is right. Once you republish metadata locally, that repo is no longer technically Mirror Complete till the next complete sync of the repo which will overwrite the locally generated metadata with the upstream metadata again. The warning is trying to communicate that with the 'high-risk' warning. We don't have the option to do this on the UI (to make this harder to do). Only hammer allows this with a --force option for some corner cases.

bangelic: I can add a "WARNING" to this. I initially thought about doing that but wanted to see what reviewers might think.

maximiliankolb:
Instead of documenting "You can use `--force` to republish metadata", I'd say "{Team} does not recommend to republish metadata for repositories with mirroring policy "complete mirror"." I would not tell users how to perform this using Hammer CLI.

guides/common/modules/con_mirroring-policies-overview.adoc
```diff
@@ -21,6 +21,9 @@ Complete Mirroring::
 Mirrors content as well as repodata.
 This is the fastest method.
 This mirroring policy is only available for Yum content.
-Republishing metadata in repositories with Complete Mirroring is high-risk.
-Hammer allows it using the `--force flag` command.
-This also applies to Content Views containing repositories with the Complete Mirroring policy.
++
+[WARNING]
+====
+{Team} does not recommend republishing metadata for repositories with Complete Mirror mirroring policy.
```
asteflova, line 27: Per the Red Hat style guide, I'd suggest avoiding the verb "recommend", especially in connection with {Team}: https://redhat-documentation.github.io/supplementary-style-guide/#_r

Possible alternatives include "Avoid republishing" or "Do not republish".

maximiliankolb (commented):
Thanks Brian, I believe this makes it much clearer that "complete mirroring" and "republishing metadata" are not a good match. However, this also completely changes the scope of your PR @bangelic 

Is that OK for you?

bangelic:
> Thanks Brian, I believe this makes it much clearer that "complete mirroring" and "republishing metadata" are not a good match. However, this also completely changes the scope of your PR @bangelic
> 
> Is that OK for you?

That works for me.

bangelic:
Reworded the scope of this PR to fit the change made.

bangelic:
@Lennonka Can you merge if you think this looks good to go?

bangelic:
@adamlazik1 Can you merge after taking a quick look?

adamlazik1 (approved):

adamlazik1:
Merged and cherry-picked:
   e2a380568..e9df8bf26  3.7 -> 3.7
   ecdd3528c..01148a93c  3.8 -> 3.8

PR #2408, opened 2023-08-31
Remove drafts

ekohl:
These have been added, but it doesn't appear to be picked up. Having it in the repository can give false positives when quickly searching for things.

* [ ] Foreman 3.6/Katello 4.8


maximiliankolb:
I am all +1 for this. Do you want to ping both Eric and Maxim who contributed those drafts?

ekohl:
Maxim left the company a while back, so I don't think he's going to finish it.

@ehelms thoughts about the architecture draft? It mostly feels like a copy of https://github.com/theforeman/puppet-pulpcore/#installation-layout. Perhaps we should file it as an issue with a reference to it.

ehelms:
> @ehelms thoughts about the architecture draft? It mostly feels like a copy of https://github.com/theforeman/puppet-pulpcore/#installation-layout. Perhaps we should file it as an issue with a reference to it.

Drop it - if we want to do an architecture document we should just visit it with fresh eyes.

ekohl:
> Assuming this is just "git rm -r drafts": LGTM

It is exactly that.

PR #2410, opened 2023-08-31
Correct Foreman 3.8 navigation

ekohl:
Various guides are hidden because they're not ready. These should not show up in the navigation.

Fixes: ddba2d718d91ade9bdfd3375f875cc883993449a ("Branch Foreman 3.8")

### DIFF
--- a/web/content/js/versions.js
+++ b/web/content/js/versions.js
@@ -196,52 +196,8 @@ const navVersions = [
             "path": "Release_Notes"
           },
           {
-            "title": "Quickstart Guide",
-            "path": "Quickstart"
-          },
-          {
-            "title": "Installing Foreman Server",
-            "path": "Installing_Server"
-          },
-          {
-            "title": "Installing Smart Proxy",
-            "path": "Installing_Proxy"
-          },
-          {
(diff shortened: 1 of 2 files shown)


maximiliankolb (commented):
Looks reasonable to me; I have no strong feelings about the navigation itself.

I am a bit unsure about ACD, managing hosts, provisioning hosts, and configuring hosts. Those guides should, generally speaking, be independent of the build target. Managing content obviously depends on Katello.

I am fine merging this as is & think of this as another reminder for myself to go through guides like ACD to ensure that they're either independent of the build target or create small PRs to add notes to specific targets for specific procedures that differ.

ekohl:
> I am a bit unsure about ACD, managing hosts, provisioning hosts, and configuring hosts. Those guides should, generally speaking, be independent of the build target. Managing content obviously depends on Katello.

If there are packages and they use the installer, then it should be easy.

> I am fine merging this as is & think of this as another reminder for myself to go through guides like ACD to ensure that they're either independent of the build target or create small PRs to add notes to specific targets for specific procedures that differ.

That is exactly my plan as well.

PR #2411, opened 2023-08-31
Run multiple builds concurrently

ekohl:
By declaring them as separate steps it can run concurrently when using make -j. This can speed things up considerably.

* [ ] Foreman 3.6/Katello 4.8


Makefile
```diff
@@ -24,9 +24,15 @@ clean:
...
+html: build-foreman-el build-foreman-deb build-katello
+
+build-foreman-el: prep
 	$(MAKE) -C guides/ html BUILD=foreman-el
+
+build-foreman-deb: prep
 	$(MAKE) -C guides/ html BUILD=foreman-deb
+
+build-katello: prep
 	$(MAKE) -C guides/ html BUILD=katello
```
evgeni, line 36:
```suggestion
build-%: prep
	$(MAKE) -C guides/ html BUILD=$*
```

Makes it easier to create new targets?

No idea how to make the same for `.PHONY`.

evgeni: Ah, you could make them all depend on a `FORCE` target and make that `PHONY`. works in https://github.com/theforeman/foreman-ansible-modules/blob/develop/Makefile

PR #2416, opened 2023-09-04
Hide updating guide upstream & correct links to upgrading guide

ekohl:
The updating procedure relies on code that's not implemented for upstream installations. By hiding it we can't confuse users with something that doesn't work.

It also reflects the correct links to the new upgrading guide.

This should replace https://github.com/theforeman/foreman-documentation/pull/2374.

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

This first commit should be picked into 3.8 & 3.7 because the changes were only made in those branches. The web changes in the second commit don't really matter.

* [x] Foreman 3.8/Katello 4.10
* [x] Foreman 3.7/Katello 4.9 (planned Satellite 6.14)
* [ ] Foreman 3.6/Katello 4.8
* [ ] Foreman 3.5/Katello 4.7 (Satellite 6.13)
* [ ] Foreman 3.4/Katello 4.6 (EL8 only)
* [ ] Foreman 3.3/Katello 4.5 on EL7 & EL8 (Satellite 6.12 on EL8 only; orcharhino 6.4/6.5 on EL8 only)
* [ ] Foreman 3.2/Katello 4.4 on EL7 & EL8
* [ ] Foreman 3.1/Katello 4.3 on EL7 & EL8 (Satellite 6.11 EL7/8; orcharhino 6.3 on EL7/8)
* We do not accept PRs for Foreman older than 3.1.

PR #2417, opened 2023-09-04
Remove Upgrading_and_Updating guides from 3.7 & nightly

ekohl:
Result of: git rm -r {3.7,nightly}/Upgrading_and_Updating

It's not in 3.8.


maximiliankolb (approved):
https://docs.theforeman.org/3.7/Upgrading_Project/index-katello.html# is part of the navigation. the key piece is "we have renamed the guide but the old one is still published". If so, please merge.

ekohl:
Correct. https://github.com/theforeman/foreman-documentation/pull/2416 does the navigation part.

PR #2431, opened 2023-09-14
Correct multi line escaping

ekohl:
Fixes: dd093af738ce ("Refs #35985 - Update Puppet enablement options")

This is only submitted against 3.7 because nightly and 3.8 need another fix.


ekohl:
> This is only submitted against 3.7 because nightly and 3.8 need another fix.

https://github.com/theforeman/foreman-documentation/pull/2432 is that fix.

maximiliankolb (approved):
LGTM!

PR #2427, opened 2023-09-12
Use inline passthrough macro to avoid escaping

ekohl:
Instead of trying to figure out which characters to escape, this uses the [inline passthrough macro](https://docs.asciidoctor.org/asciidoc/latest/pass/pass-macro/).

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.8/Katello 4.10
* [x] Foreman 3.7/Katello 4.9 (planned Satellite 6.14)
* [x] Foreman 3.6/Katello 4.8
* [ ] Foreman 3.5/Katello 4.7 (Satellite 6.13)
* [ ] Foreman 3.4/Katello 4.6 (EL8 only)
* [ ] Foreman 3.3/Katello 4.5 on EL7 & EL8 (Satellite 6.12 on EL8 only; orcharhino 6.4/6.5 on EL8 only)
* [ ] Foreman 3.2/Katello 4.4 on EL7 & EL8
* [ ] Foreman 3.1/Katello 4.3 on EL7 & EL8 (Satellite 6.11 EL7/8; orcharhino 6.3 on EL7/8)
* We do not accept PRs for Foreman older than 3.1.

PR #2437, opened 2023-09-18
Do not build PDF documents

maximiliankolb:
As discussed earlier: we do not publish PDF documents in upstream or downstream, so we can stop building them to speed up GHA and save resources.


ekohl:
I needed to pick this because of https://github.com/theforeman/foreman-documentation/pull/2427:
   4ba4e6d75..941a6bc0f  3.8 -> 3.8
   16fb0d854..43ebf6812  3.7 -> 3.7
   66d8f26d7..af1c60875  3.6 -> 3.6

PR #2441, opened 2023-09-19
Configure SubMan to connect via HTTP proxy for registration

Lennonka:
If Project is behind an HTTP proxy, configure the SubMan to connect through the proxy.

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_registering-a-host.adoc
+++ b/guides/common/modules/proc_registering-a-host.adoc
@@ -16,6 +16,12 @@ ifdef::satellite[]
 This is required to install the `insights-client` package on hosts.
 endif::[]
 include::snip_prerequisite-configured-smart-proxy-registration-provisioning.adoc[]
+ifdef::katello,orcharhino,satellite[]
+* If your {ProjectServer} or {SmartProxyServer} is behind an HTTP proxy, configure the Subscription Manager on your host to use the HTTP proxy for connection.
+endif::[]
+ifdef::satellite[]
+For more information, see https://access.redhat.com/solutions/65300[How to access Red Hat Subscription Manager (RHSM) through a firewall or proxy] in the _Red{nbsp}Hat Knowledgebase_.
+endif::[]
 
 .Procedure
 . In the {ProjectWebUI}, navigate to *Hosts* > *Register Host*.


guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -16,6 +16,10 @@ ifdef::satellite[]
 This is required to install the `insights-client` package on hosts.
 endif::[]
 include::snip_prerequisite-configured-smart-proxy-registration-provisioning.adoc[]
+* If your {ProjectServer} or {SmartProxyServer} is behind an HTTP proxy, configure the Subscription Manager on your host to use the HTTP proxy for connection.
```
adamlazik1, line 19: Subcription Manager is only on RHEL hosts, shoud that be accounted for somehow?

Lennonka: It seems that the Global registration template always uses subscription-manager, so this is probably not the issue.
Perhaps the question rather is whether the Global registration currently supports OSs without subscription-manager?

@maximiliankolb @ekohl Can you please clarify this?

maximiliankolb: @adamlazik1 orcharhino Clients for Amazon Linux, CentOS 7, Debian, Oracle Linux, SUSE Linux Enterprise Server, and Ubuntu also contain a packaged `subscription-manager` + dependencies. AlmaLinux, CentOS Stream, Red Hat Enterprise Linux, and Rocky Linux provide the package and its dependencies from the default OS repositories.

ekohl: As @maximiliankolb said, it is available on many platforms, but IMHO it is only relevant for the Katello use case.

I do wonder when it's needed. The way it's designed is that you have either:

Client: subscription-manager -> Foreman

Or:

Client: subscription-manager -> Content Proxy: Apache -> Foreman

How often is there a HTTP proxy in that first step? And how well does it work client certificates? From an architecture perspective I question if is a good idea. Deploying an additional content proxy where clients don't need HTTP proxies is IMHO better, but I also know that in practice some companies have crazy policies.

Lennonka: Oh, right, yes, it's a Katello use case. And yes, it looks strange to me to from the architecture perspective, but we have a customer case based on which it was requested, so... this is just in case the customer has a crazy policy in place. :)

adamlazik1 (approved):
Thanks, my concern is resolved.

Lennonka:
Cherry-picked:
- e2158bb38..0e36b370c  3.8 -> 3.8
- 2fe50792e..d408026da  3.7 -> 3.7
- 13d4b598b..f167eff59  3.6 -> 3.6
- ec8691d56..04d79f007  3.5 -> 3.5
- 333001986..6c05969c3  3.4 -> 3.4
- 86298187a..fd48ab34c  3.3 -> 3.3
- c155371ca..887b46ddc  3.2 -> 3.2
- fbc716108..a302c8fb3  3.1 -> 3.1

PR #2442, opened 2023-09-19
Fix warning in configuring AD integration

adamlazik1:
Currently, the warning is contradictory to the prerequisite. BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2232693

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_configuring-direct-ad-integration-with-gss-proxy.adoc
+++ b/guides/common/modules/proc_configuring-direct-ad-integration-with-gss-proxy.adoc
@@ -53,8 +53,6 @@ euid = __ID_of_Apache_User__
 ----
 . Create a keytab entry:
 +
-WARNING: The host must not be enrolled to a domain before creating a keytab entry.
-+
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
 # KRB5_KTNAME=FILE:/etc/httpd/conf/http.keytab net ads keytab add HTTP -U administrator -d3 -s /etc/net-keytab.conf


asteflova:
I'd actually suggest to remove that warning. If the requirement is already documented as a prerequisite, there is no need to duplicate it.

Lennonka:
I wonder if the change should be more complex than just removing the "not". What if the "not" was correct, but the rest of the sentence should get clarification?
What host? Does "the host" mean the Satellite Server?
What domain? Is the domain something else than the AD server?
Is there a possibility to enroll the SatServer to the AD server, but not to an AD domain? If that's the case, then perhaps it shoud be specified in the prerequisite and the warning can be removed.

maximiliankolb:
Without knowing anything about AD: maybe go with a second prerequisite "Ensure that your host is enrolled to a domain before creating a keytab entry."?

adamlazik1:
> I wonder if the change should be more complex than just removing the "not". What if the "not" was correct, but the rest of the sentence should get clarification?
> What host? Does "the host" mean the Satellite Server?
> What domain? Is the domain something else than the AD server?
> Is there a possibility to enroll the SatServer to the AD server, but not to an [AD domain](https://blog.netwrix.com/2017/01/31/active-directory-domain/)? If that's the case, then perhaps it shoud be specified in the prerequisite and the warning can be removed.
> Also, if that's the case and it must be enrolled without specifying the AD domain, then having it in this section is too late. :)

I will get more info on this before continuing.

mjivraja (approved):

mjivraja:
Oops, there seems to be ongoing discussion on this. Please disregard my approval.

Lennonka (approved):

Lennonka:
Cherry-picked:
- 4536eb909..cb5682dfb  3.8 -> 3.8
- bd487a90a..4d6d29ee5  3.7 -> 3.7
- a5d64c2cb..f42440409  3.6 -> 3.6
- 0c0d6be40..4e966492b  3.5 -> 3.5
- 84f684b9f..cd625ef24  3.4 -> 3.4
- db25343f2..acfc55769  3.3 -> 3.3
- c69ddbb77..74c2f0db9  3.2 -> 3.2
- f7489c9e3..dac55d237  3.1 -> 3.1

PR #1694, opened 2022-10-12
Add troubleshooting for ruby & pgsql

mjivraja:
Project needs ruby:2.7 and postgresql:12 modules to work\get installed. If end-users proceed with enabling present el8 module in installing project and proxy, it will pop-up a huge warning related to conflict with ruby:2.5 and postgresql:10 modules. In the end, the command does enable the ruby:2.7 and postgresql:12 but the list of warnings itself is not a healthy status of the output and this occurrence should be prevented. However, in the document, these warnings are mentioned as expected and can be ignored but it should not be the standard practice of user experience. Hence, this warning can be easily avoided by enabling the required dependent modules along with the actual module.

https://bugzilla.redhat.com/show_bug.cgi?id=2126884

* [X] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [X] Foreman 3.7/Katello 4.9
* [X] Foreman 3.6/Katello 4.8
* [X] Foreman 3.5/Katello 4.7
* [X] Foreman 3.4/Katello 4.6
* [X] Foreman 3.3/Katello 4.5
* [X] Foreman 3.2/Katello 4.4
* [X] Foreman 3.1/Katello 4.3
* For Foreman 3.0 or older, please create a separate PR.
* We do not accept PRs for Foreman 2.3 or older.

PR #2452, opened 2023-09-22
Add troubleshooting for ruby & pgsql

mjivraja:
Project needs ruby:2.7 and postgresql:12 modules to work\get installed. If end-users proceed with enabling present el8 module in installing project and proxy, it will pop-up a huge warning related to conflict with ruby:2.5 and postgresql:10 modules. At the end, the command does enable the ruby:2.7 and postgresql:12 but the list of warnings itself is not a healthy status of the output and this occurrence should be prevented. However, in the document, these warnings are mentioned as expected and can be ignored but it should not be the standard practise of user experience. Therefore, this warning can be easily avoided in documentation by adding the dedicated section in appendix.

https://bugzilla.redhat.com/show_bug.cgi?id=2126884


mjivraja:
This PR resolves conflicts from #1694

PR #2484, opened 2023-10-03
Add steps for adding repo manually when importing container images

asteflova:
[BZ#2022803](https://bugzilla.redhat.com/show_bug.cgi?id=2022803) requests adding steps on importing container images not using repo discovery, but manually for existing products. This PR updates the procedure with these steps, while also making other minor edits to accommodate the new steps.

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_importing-container-images.adoc
+++ b/guides/common/modules/proc_importing-container-images.adoc
@@ -30,6 +30,14 @@ Click to edit, and add any tags that you want to limit the content that synchron
 . In the {ProjectWebUI}, navigate to *Content* > *Products* and select the name of your product.
 . Select the new repositories and then click *Sync Now* to start the synchronization process.
 
+.Procedure with creating a repository manually
+. In the {ProjectWebUI}, navigate to *Content* > *Products*.
+Click the name of the required product.
+. Click *New repository*.
+. From the *Type* list, select *docker*.
+Enter the details for the repository, and click *Save*.
+. Select the new repository, and click *Sync Now*.
+
 [role="_additional-resources"]
 .Next steps
 * To view the progress of the synchronization, navigate to *Content* > *Sync Status* and expand the repository tree.
--- a/guides/common/modules/proc_importing-container-images.adoc
+++ b/guides/common/modules/proc_importing-container-images.adoc
@@ -8,12 +8,9 @@ ifdef::orcharhino[]
 You can import container image repositories from any container image registry.
 endif::[]
 
-This procedure uses repository discovery to find container images and import them as repositories.
-For more information about creating a product and repository manually, see xref:Importing_Content_{context}[].
-
 To use the CLI instead of the {ProjectWebUI}, see the xref:cli-importing-container-images[].
 
-.Procedure
+.Procedure with repository discovery
 . In the {ProjectWebUI}, navigate to *Content* > *Products* and click *Repo Discovery*.
 . From the *Repository Type* list, select *Container Images*.
 . In the *Registry to Discover* field, enter the URL of the registry to import images from.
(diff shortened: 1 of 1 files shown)


maximiliankolb:
I would recommend to split this procedure into two files/procedures.

ianballou (commented):
I was going back and forth about if we need to include documentation for the details that go into a repository (like "upstream repository name"), but I think once a user sees the page in the UI, all will become clear.

Anyway, I think this solves the issue where users think repo discovery is the only way to make container repositories.

asteflova:
> I was going back and forth about if we need to include documentation for the details that go into a repository (like "upstream repository name"), but I think once a user sees the page in the UI, all will become clear.

That's what I think too. This is described in the web UI in a very helpful way. And because users fill in the fields in the web UI, they will see those details right where they need them, when they need them.

> 
> Anyway, I think this solves the issue where users think repo discovery is the only way to make container repositories.

Thanks for reviewing!

asteflova:
> I would recommend to split this procedure into two files/procedures.

In my first draft, there were two modules. But I changed my mind because these are two alternatives for doing one thing, and the customer who opened the case tried the first alternative (repo discovery) but that didn't work, so he ended up using the other one (manual). So we want users who read one of the procedures to know that the other one exists too.

If I document them as separate modules, it would involve cross-linking and snippets for introduction and next steps. Overall, that would not create a smooth customer experience (clicking links and reading the same text twice).

On the other hand, documenting them as one module diverges slightly from the modular docs guidelines, but is better for the customer experience (it shows the procedures are tightly linked).

That said @maximiliankolb, I don't have a particularly strong opinion on the matter, so if you still think this would work better as two modules, I'll do the splitting.

guides/common/modules/proc_importing-container-images.adoc
```diff
@@ -33,9 +30,18 @@ Click to edit, and add any tags that you want to limit the content that synchron
 . In the {ProjectWebUI}, navigate to *Content* > *Products* and select the name of your product.
 . Select the new repositories and then click *Sync Now* to start the synchronization process.
 
-To view the progress of the synchronization, navigate to *Content* > *Sync Status* and expand the repository tree.
+.Procedure with creating a repository manually
+. In the {ProjectWebUI}, navigate to *Content* > *Products*.
+Click the name of the required product.
+. Click *New repository*.
+. From the *Type* list, select *docker*.
+Enter the details for the repository, and click *Save*.
```
maximiliankolb, line 38: IMHO needs a list of steps what do do/fill in for each field.

maximiliankolb (commented):
I am OK with this as is.

guides/common/modules/proc_importing-container-images.adoc
asteflova: I don't think so. The fields are described in the UI form itself, I would just be copy-pasting the descriptions in the docs.

new_repo_scrn

maximiliankolb: That's what we currently do, don't we?

* https://docs.theforeman.org/nightly/Managing_Content/index-katello.html#Creating_a_Custom_Product_content-management
* https://docs.theforeman.org/nightly/Managing_Content/index-katello.html#Creating_a_Sync_Plan_content-management
* https://docs.theforeman.org/nightly/Managing_Content/index-katello.html#Creating_an_Activation_Key_content-management
* https://docs.theforeman.org/nightly/Provisioning_Hosts/index-katello.html#creating-operating-systems_provisioning
* etc

asteflova: We do, and it's not something I support. I believe it creates an undesirable user experience and is also a maintenance burden for writers.

We can discuss this in more detail at the next meeting if you want to. Or should we move this conversation to an issue?

maximiliankolb: Yes, please open an issue for this.

maximiliankolb (approved):

guides/common/modules/proc_importing-container-images.adoc
asteflova: https://github.com/theforeman/foreman-documentation/issues/2490

I'm looking forward to the discussion!

maximiliankolb (approved):

maximiliankolb:
   d1e7b7aa5..538c863af  3.8 -> 3.8
   44f1feece..27ec52aa2  3.7 -> 3.7

PR #2488, opened 2023-10-04
Remove network interface from foreman-installer

maximiliankolb:

### DIFF
--- a/guides/common/modules/proc_configuring-network-services.adoc
+++ b/guides/common/modules/proc_configuring-network-services.adoc
@@ -22,10 +22,6 @@ While performing a Katello deployment, to configure these services on an externa
 ====
 endif::[]
 
-ifdef::satellite[]
-{ProjectServer} uses `eth0` for external communication, such as connecting to Red Hat's CDN.
-endif::[]
-
 .Procedure
 . Enter the `{foreman-installer}` command to configure the required network services:
 +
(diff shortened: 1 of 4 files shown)


maximiliankolb:
I am unsure if I should make any more changes:

```
$ rg eth1
guides/common/modules/ref_dns-options-for-network-configuration.adoc
24:Set this to `eth1`.

guides/common/modules/proc_completing-the-rhel6-image.adoc
31:If `/etc/udev/rules.d/70-persistent-net.rules` is created, networking might not function properly when booting from snapshots (the network interface is created as "eth1" rather than "eth0" and IP address is not assigned).

guides/common/modules/ref_dhcp-options-for-network-configuration.adoc
18:Set this to `eth1`.
22:Set this to the address for {ProjectServer} on `eth1`.

guides/common/modules/ref_host-specific-variables.adoc
30:You can pass an array of multiple identifiers as an input, for example @host.interfaces_with_identifier(['eth0', 'eth1']).

guides/common/modules/proc_adding-a-virtual-interface.adoc
22:If creating a VLAN, specify ID in the form of *eth1.10* in the *Device Identifier* field.
23:If creating an alias, use ID in the form of *eth1:10*.
31:* *Attached to*: Specify the identifier of the physical interface to which the virtual interface belongs, for example *eth1*.

guides/common/modules/proc_adding-a-bonded-interface.adoc
48:   --interface="identifier=_eth1_, \
64:               attached_devices=[_eth1,eth2_], \

$ rg eth2
guides/common/modules/proc_adding-a-bonded-interface.adoc
54:   --interface="identifier=_eth2_, \
64:               attached_devices=[_eth1,eth2_], \

```

Do we want to phase out "eth1" in favor of something more generic?

guides/common/modules/proc_configuring-network-services.adoc
ekohl, line 1: I'd remove line 26 where it says `eth0` is used.

ekohl (changes requested):
Minor note, but otherwise :+1:

ekohl:
> Do we want to phase out "eth1" in favor of something more generic?

I didn't see this until I submitted my review. When reviewing those files I think they're OK to keep.

guides/common/modules/proc_configuring-network-services.adoc
maximiliankolb: Applied.

maximiliankolb:
Applied your feedback & ready for re-review @ekohl 

Is it OK to squash the last four commits before merging?

adamlazik1 (approved):
Looks good from writer perspective.

maximiliankolb:
Rebased to master and squashed some commits as outlined. Merging after green GHA.

ekohl (approved):

ekohl:
I think ipxe.org times out very often. I restarted it again, but I'm wondering if we should stop checking the domain, given how often it gives false positives. We just link to 3 URLs and I'd feel fairly confident that they're stable.

maximiliankolb:
   0036b2c3a..faaa4c1cc  3.8 -> 3.8
   83b2be9fc..8c2e2722a  3.7 -> 3.7

PR #2497, opened 2023-10-06
Update code samples in VMware cloud-init

Lennonka:
cloud.cfg is missing the SSH module for VMWare cloud-init 20 and newer.
Also, we can tell the user to remove all SSH keys in a clearer way.

https://bugzilla.redhat.com/show_bug.cgi?id=2237783

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_using-vmware-cloud-init-and-userdata-templates-for-provisioning.adoc
+++ b/guides/common/modules/proc_using-vmware-cloud-init-and-userdata-templates-for-provisioning.adoc
@@ -98,6 +98,7 @@ If you intend to provision through {SmartProxyServer}, use the URL of your {Smar
 # cat << EOM > /etc/cloud/cloud.cfg
 cloud_init_modules:
  - bootcmd
+ - ssh
 
 cloud_config_modules:
  - runcmd
@@ -179,7 +180,7 @@ endif::[]
 +
 [options="nowrap" subs="+quotes"]
 ----
-# rm -f /etc/ssh/_SSH_keys_
+# rm -f /etc/ssh/ssh_host_*
 ----
 . Remove root user's SSH history:
 +


guides/common/modules/proc_using-vmware-cloud-init-and-userdata-templates-for-provisioning.adoc
```diff
@@ -179,7 +180,7 @@ endif::[]
 +
 [options="nowrap" subs="+quotes"]
 ----
-# rm -f /etc/ssh/_SSH_keys_
+# rm -f /etc/ssh/\*key*
```
maximiliankolb, line 183:
```suggestion
# rm -f /etc/ssh/ssh_host_*
```
See https://www.ssh.com/academy/ssh/host-key#host-keys-in-openssh

I believe having bold text in the code block goes against our current conventions. Maybe we can instead reword the list item or explain why this is important?

ekohl (commented):
Wouldn't it be cleaner to rely on `virt-sysprep` for this? For RHEL 7 it's documented: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/virtualization_deployment_and_administration_guide/sect-guest_virtual_machine_disk_access_with_offline_tools-using_virt_sysprep and RHEL 8 has https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_virtualization/cloning-virtual-machines_configuring-and-managing-virtualization#preparing-a-virtual-machine-template_assembly_creating-virtual-machine-templates.

guides/common/modules/proc_using-vmware-cloud-init-and-userdata-templates-for-provisioning.adoc
Lennonka: It isn't supposed to be bold text. It's supposed to be wildcard replacement in filename lookup. It's supposed to say: delete all files that contain the "key" substring.

Lennonka:
> Wouldn't it be cleaner to rely on `virt-sysprep` for this? For RHEL 7 it's documented: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/virtualization_deployment_and_administration_guide/sect-guest_virtual_machine_disk_access_with_offline_tools-using_virt_sysprep and RHEL 8 has https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_virtualization/cloning-virtual-machines_configuring-and-managing-virtualization#preparing-a-virtual-machine-template_assembly_creating-virtual-machine-templates.

@ekohl I have no idea.

But I've taken a look at the RHEL docs and that tools seems to clone the VM from "the outside", whereas this procedure counts on preparations of a VM on VMware and then using VMware to create the image from this prepared VM. So is the tool actually applicable in this use case? I have no idea how VMware works. If you think it's worth looking into it, I need more information and perhaps another BZ, because that's out of scope of this PR.

guides/common/modules/proc_using-vmware-cloud-init-and-userdata-templates-for-provisioning.adoc
Lennonka: That's why the first asterisk is escaped.

Lennonka: But your suggestion will probably work too.

ekohl:
> But I've taken a look at the RHEL docs and that tools seems to clone the VM "from the outside", whereas this procedure counts on preparations "inside a VM" on VMware and then using VMware to create the image from this prepared VM. So how is the tool applicable in this use case? I have no idea how VMware works. If you think it's worth looking into it, I need more information and perhaps another BZ, because that's out of scope of this PR.

I agree it's out of scope of this PR, but even though it's been a long time since I used VMware it feels like we're reinventing something RHEL already has figured out.

maximiliankolb (approved):
The change to remove SSH host keys LGTM.

Lennonka:
Cherry-picked:
- eaaab1330..4ba4e6d75  3.8 -> 3.8
- 78c073aed..16fb0d854  3.7 -> 3.7
- 16eeacb39..66d8f26d7  3.6 -> 3.6
- 9785fd822..5e045f809  3.5 -> 3.5
- dde27b7b3..588f0210b  3.4 -> 3.4
- caa21b288..da0b187d4  3.3 -> 3.3
- 6fe352a9d..a08e91fc0  3.2 -> 3.2
- e24234d3a..5a886bbf5  3.1 -> 3.1

PR #2501, opened 2023-10-09
Consolidate version attribute

maximiliankolb:
See my commit message.

### DIFF
--- a/guides/common/attributes-base.adoc
+++ b/guides/common/attributes-base.adoc
@@ -111,8 +111,6 @@
 :ProjectWebUI-context: {project-context}_web_UI
 :ProjectX: {Project}
 :ProjectXY: {Project}{nbsp}{ProjectVersion}
-:ProductVersion: {ProjectVersion}
-:ProductVersionPrevious: {ProjectVersionPrevious}
 :provision-script: OS installer recipe
 :EL: Enterprise Linux
 :RHCloud: Red{nbsp}Hat Hybrid Cloud Console
--- a/guides/common/attributes-orcharhino.adoc
+++ b/guides/common/attributes-orcharhino.adoc
@@ -1,8 +1,8 @@
 // Overrides for orcharhino build
 :install-on-os: Enterprise Linux
 :BaseURL: https://docs.orcharhino.com/
-:ProductVersion: 6.5
-:ProductVersionPrevious: 6.4
+:ProjectVersion: 6.5
+:ProjectVersionPrevious: 6.4
 :Project: orcharhino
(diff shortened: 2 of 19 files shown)


maximiliankolb:
When creating the PR, I thought that I'd use the attribute that aligns nicely with the list of other attributes like "ProjectName" etc: https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-base.adoc#L106

Not sure why I didn't ... :see_no_evil: Anyway, it's fixed now and the diff is still zero.

Lennonka:
> cc @Lennonka This would break our downstream builds, but I think the fix would be simple. We would, however; have to act quickly.

@adamlazik1 Yes, this will definitely break the 6.14 RN build. We have to update the attributes used in the `docinfo.xml` at least. I'm not aware if any of these attributes is used in the individual RNs. I don't know if these attributes are used in any non-upstreamized guides. But we can merge and then fix it. Will you be able to look into it?

Lennonka (approved):

Lennonka (commented):
Damn, it has conflicts with the base branch. Please, rebase.

maximiliankolb:
Rebased to master & resolved the merge conflict :v:

adamlazik1:
> > cc @Lennonka This would break our downstream builds, but I think the fix would be simple. We would, however; have to act quickly.
> 
> @adamlazik1 Yes, this will definitely break the 6.14 RN build. We have to update the attributes used in the `docinfo.xml` at least. I'm not aware if any of these attributes is used in the individual RNs. I don't know if these attributes are used in any non-upstreamized guides. But we can merge and then fix it. Will you be able to look into it?

I would be able to, but we have to agree on a good time to merge this. Ideally, Tuesday or Thursday at noon. I am not sure how long it's gonna take to sync to downstream, but iirc the script runs every hour? Also, it's not just release notes but potentially hammer, api, and, and other downstream guides. Simple sed command should hopefully take care of it.

Lennonka:
@adamlazik1 Yes, the sync runs every hour. I've already created a PR to fix the attribute usage for the RN repo. So I'm leaving it with you to fix the main downstream repo for master and 6.14. Max has the replacement script in the commit message, I think you can use that.

Lennonka (approved):
Let's do this! :)

Lennonka:
Cherry-picked:
- 31aad3da1..f23edf88a  3.8 -> 3.8
- 98edfe01e..bdc929b4d  3.7 -> 3.7

maximiliankolb:
Thanks Lena and Adam!

PR #2509, opened 2023-10-12
Replace outdated bind package for EL8

mjivraja:
On RHEL 8.x, 'dnssec-keygen' is provided by the 'bind-utils' package. Simply 'Bind' package does not fulfill the required packages. This change is applicable to only EL8.

https://bugzilla.redhat.com/show_bug.cgi?id=2243039

* [X] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
+++ b/guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
@@ -2,7 +2,7 @@
 
 = Configuring an External DHCP Server to Use with {ProductName}
 
-To configure an external DHCP server running {EL} to use with {ProductName}, you must install the ISC DHCP Service and Berkeley Internet Name Domain (BIND) packages.
+To configure an external DHCP server running {EL} to use with {ProductName}, you must install the ISC DHCP Service and Berkeley Internet Name Domain (BIND) utilities packages.
 You must also share the DHCP configuration and lease files with {ProductName}.
 The example in this procedure uses the distributed Network File System (NFS) protocol to share the DHCP configuration and lease files.
 
@@ -14,11 +14,11 @@ include::snip_firewalld.adoc[]
 
 .Procedure
 
-. On your {EL} host, install the ISC DHCP Service and Berkeley Internet Name Domain (BIND) packages:
+. On your {EL} host, install the ISC DHCP Service and Berkeley Internet Name Domain (BIND) utilities packages:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {package-install} dhcp bind
+# {package-install} dhcp-server bind-utils
 ----
 
 . Generate a security token:
@@ -34,10 +34,10 @@ As a result, a key pair that consists of two files is created in the current dir
 +
 [options="nowrap"]
 ----
-# cat Komapi_key.+*.private |grep ^Key|cut -d ' ' -f2
+# grep ^Key Komapi_key.+*.private  | cut -d ' ' -f2
 ----
 
-. Edit the `dhcpd` configuration file for all of the subnets and add the key.
+. Edit the `dhcpd` configuration file for all subnets and add the key.
 The following is an example:
 +
 [options="nowrap" subs="+quotes"]
@@ -166,12 +166,12 @@ Note that the IP address that you enter is the {Project} or {SmartProxy} IP addr
 # exportfs -rva
 ----
 
-. Configure the firewall for the DHCP omapi port 7911:
+. Configure the firewall for DHCP omapi port 7911:
 +
 [options="nowrap"]
 ----
-# firewall-cmd --add-port="7911/tcp" \
-&& firewall-cmd --runtime-to-permanent
+# firewall-cmd --add-port=7911/tcp
+# firewall-cmd --runtime-to-permanent
 ----
 
 . Optional: Configure the firewall for external access to NFS.


guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
```diff
@@ -18,7 +18,7 @@ include::snip_firewalld.adoc[]
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {package-install} dhcp bind
+# {package-install} dhcp bind-utils
```
maximiliankolb, line 21: I could reproduce `bind-utils`, but not `dhcp`. There is both `dhcp-client` and `dhcp-server` on EL8

maximiliankolb (changes requested):

guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
momran-rh: Good catch. I checked, and found out the same. `dhcp-server` is the one that we need since we are setting up a DHCP server as per the procedure. However, I could now notice something that we missed earlier. The procedure is for installing an external DHCP server to use with Satellite or Capsule server. It seems that majority of readers of the procedure (including me) got the impression that the packages are to be installed on the Satellite or Capsule server itself, when they are actually supposed to be installed on a separate RHEL system.

So, depending on the release of the RHEL system to be set up and configured as the external DHCP server, the packages that need to be installed will differ:
 - On a RHEL 7.x system, 'dhcp' and 'bind' packages are to be installed.
 - On RHEL 8.x and RHEL 9.x, the 'dhcp-server' and 'bind-utils' packages are to be installed.

Now, the question is: which of these shall we include in the procedure? RHEL 8.x and 9.x only, or RHEL 7.x as well?

Also, I would suggest to explicitly refer to the RHEL host on which the packages to be installed in step (1) as 'On your Red Hat Enterprise Linux host that is going to be configured as the external DHCP server' rather than just 'On your Red Hat Enterprise Linux host', to clear out any confusion.

mjivraja: So, I will be explicitly adding EL7 and EL8 both in downstream 6.11 (another PR). For this PR, I am only considering EL8 and hence, only working at package level.

mjivraja: Addressed :+1:

momran-rh:
> The description on line 17 is no longer correct. It doesn't install BIND anymore, just its utilities.

I believe also line 5 needs to be modified.

guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
```diff
@@ -34,10 +34,10 @@ As a result, a key pair that consists of two files is created in the current dir
 +
 [options="nowrap"]
 ----
-# cat Komapi_key.+*.private |grep ^Key|cut -d ' ' -f2
+# grep ^Key Komapi_key.+*.private  | cut -d ' ' -f 2
```
mjivraja, line 37: Why read the whole file when you can directly grep. This command looks better to me.
Just a suggestion.

guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
```diff
@@ -34,10 +34,10 @@ As a result, a key pair that consists of two files is created in the current dir
 +
 [options="nowrap"]
 ----
-# cat Komapi_key.+*.private |grep ^Key|cut -d ' ' -f2
+# grep ^Key Komapi_key.+*.private  | cut -d ' ' -f2
 ----
 
-. Edit the `dhcpd` configuration file for all of the subnets and add the key.
+. Edit the `dhcpd` configuration file for all the subnets and add the key.
```
maximiliankolb, line 40:
```suggestion
. Edit the `dhcpd` configuration file for all subnets and add the key.
```

guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
```diff
@@ -166,7 +166,7 @@ Note that the IP address that you enter is the {Project} or {SmartProxy} IP addr
 # exportfs -rva
 ----
 
-. Configure the firewall for the DHCP omapi port 7911:
+. Configure the firewall for DHCP omapi port 7911:
 +
 [options="nowrap"]
 ----
```
maximiliankolb, line 172: comment on line 173: it's probably a good idea to drop "\ &&" in favor of starting the second line with "# " so that users run two commands. That's AFAIK our general way.

maximiliankolb (commented):
Tested the package names on RHEL8.8: :heavy_check_mark:

guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
mjivraja: Addressed :+1: 
And also removed `double quotes` for adding ports in command. I guess, it was not necessary.

maximiliankolb (approved):
LGTM!

maximiliankolb:
Merged to master and cherry-picked :heavy_check_mark: 
   f23edf88a..0036b2c3a  3.8 -> 3.8
   bdc929b4d..83b2be9fc  3.7 -> 3.7
   2ed43ac8a..830d39b52  3.6 -> 3.6
   f1181469b..5077568a7  3.5 -> 3.5
   66378d7f7..058c6df91  3.4 -> 3.4
   bb5e50e9b..9b19658be  3.3 -> 3.3

PR #2518, opened 2023-10-16
Fix reference rendering in satellite

adamlazik1:
This change is required for Satellite, where we have `proc_configuring-repositories.adoc` also in the QuickStart guide and while we do not use the guide, it builds as a part of the downstream pipeline so this change is required at the very least downstream. The reason I am putting it here is to prevent desynchronization in the future.

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_configuring-repositories.adoc
+++ b/guides/common/modules/proc_configuring-repositories.adoc
@@ -100,7 +100,13 @@ endif::[]
 +
 [NOTE]
 ====
-If there is any warning about conflicts with Ruby or PostgreSQL while enabling `{dnf-modules}` module, see xref:troubleshooting-dnf-modules_{context}[Troubleshooting DNF modules].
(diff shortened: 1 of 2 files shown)


adamlazik1:
On a side note: In satellite the context is already set to quick-start. I guess the master probably doesn't sync since the guide is unused.

adamlazik1:
@ekohl @maximiliankolb is this ok for upstream and orcharhino?

maximiliankolb:
In downstream, we do not use the quickstart guide (yet). For me, it'd be OK to keep upstream close to what you do for Satellite, so there's fewer slightly different variants/guides/etc.

ekohl (approved):
Upstream doesn't really link that AFAIK so I doubt it breaks anything.

maximiliankolb:
   c6ffbe888..efa9c9421  3.8 -> 3.8
   a4adeeabd..9147cc034  3.7 -> 3.7

PR #2517, opened 2023-10-16
Use attribute to reduce duplicate content

maximiliankolb:
Output should not differ much. Using an attribute is IMO much nicer than duplicate content.

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.8/Katello 4.10
* [x] Foreman 3.7/Katello 4.9 (planned Satellite 6.14)

PR #2522, opened 2023-10-17
Use attribute as shell variable

maximiliankolb:
* Rename file to align with upstream terminology
* Revert 41a0b96 partially to ensure the docs match the stdout from "katello-certs-check" on Foreman + Satellite
* Use proper attribute for Smart Proxy FQDN vs Smart Proxy name

Follow-up PR to https://github.com/theforeman/foreman-documentation/pull/2517 
Sorry for the trouble @ekohl Please review to ensure both issues are addressed.

### DIFF
--- a/guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
+++ b/guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
@@ -37,16 +37,16 @@ If the command is successful, it returns two `{certs-generate}` commands, one of
 ----
 Validation succeeded.
 
-To use them inside a NEW {SmartProxyServer}, run this command:
-  {certs-generate} --foreman-proxy-fqdn "{SmartProxyServer}" \
-    --certs-tar "~/{smartproxy-example-com}-certs.tar" \
+To use them inside a NEW ${smart-proxy-capitalized}, run this command:
+  {certs-generate} --foreman-proxy-fqdn "${smart-proxy-capitalized}" \
+    --certs-tar "~/{smart-proxy-capitalized}-certs.tar" \
     --server-cert "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem_" \
     --server-key "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert_key.pem_" \
     --server-ca-cert "_/root/{smart-proxy-context}_cert/ca_cert_bundle.pem_" \
 
-To use them inside an EXISTING {SmartProxyServer}, run this command INSTEAD:
-  {certs-generate} --foreman-proxy-fqdn "\{SmartProxyServer}" \
-    --certs-tar "~/{smartproxy-example-com}-certs.tar" \
+To use them inside an EXISTING ${smart-proxy-capitalized}, run this command INSTEAD:
+  {certs-generate} --foreman-proxy-fqdn "${smart-proxy-capitalized}" \
+    --certs-tar "~/{smart-proxy-capitalized}-certs.tar" \
     --server-cert "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem_" \
     --server-key "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert_key.pem_" \
     --server-ca-cert "_/root/{smart-proxy-context}_cert/ca_cert_bundle.pem_" \
@@ -54,7 +54,7 @@ To use them inside an EXISTING {SmartProxyServer}, run this command INSTEAD:
 ----
 . On your {ProjectServer}, from the output of the `katello-certs-check` command, depending on your requirements, enter the `{certs-generate}` command that generates a certificate for a new or existing {SmartProxy}.
 +
-In this command, change `{SmartProxyServer}` to the FQDN of your {SmartProxyServer}.
+In this command, change `{smart-proxy-capitalized}` to the FQDN of your {SmartProxyServer}.
 +
 . Retain a copy of the `{foreman-installer}` command that the `{certs-generate}` command returns for deploying the certificate to your {SmartProxyServer}.
 +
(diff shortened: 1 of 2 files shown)


adamlazik1 (approved):

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
```diff
@@ -37,15 +37,15 @@ If the command is successful, it returns two `{certs-generate}` commands, one of
 ----
 Validation succeeded.
 
-To use them inside a NEW {SmartProxyServer}, run this command:
-  {certs-generate} --foreman-proxy-fqdn "{SmartProxyServer}" \
+To use them inside a NEW ${smart-proxy-capitalized}, run this command:
```
ekohl, line 40: For context: it comes from https://github.com/theforeman/foreman-installer/blob/ee24dcce8b709d1badf4137f25ebd1ab8be7c9a4/bin/katello-certs-check#L276

The value is set here: https://github.com/theforeman/foreman-installer/blob/ee24dcce8b709d1badf4137f25ebd1ab8be7c9a4/bin/katello-certs-check#L22

And branding is loaded if present: https://github.com/theforeman/foreman-installer/blob/ee24dcce8b709d1badf4137f25ebd1ab8be7c9a4/bin/katello-certs-check#L11-L15

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
```diff
@@ -37,15 +37,15 @@ If the command is successful, it returns two `{certs-generate}` commands, one of
...
+  {certs-generate} --foreman-proxy-fqdn "{smartproxy-example-com}" \
```
ekohl, line 41: Here I'm not sure: the command output is literally `$FOREMAN_PROXY` and not `{smartproxy-example-com}`. So the intended workflow is:
```bash
FOREMAN_PROXY={smartproxy-example-com}
copy_the_suggested_commands_here
```

I'm not sure if this is the best, but the current documentation flow is a bit confusing to me.

ekohl (changes requested):
Line 57 (`In this command, change `{SmartProxyServer}` to the FQDN of your {SmartProxyServer}.`) doesn't look correct after this.

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
maximiliankolb: I have replace this attribute. I would also be open to drop the examples completely because a user _will_ get these commands as stdout.

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
maximiliankolb: Thanks; i've also added those links to my commit message.

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
```diff
@@ -37,15 +37,15 @@ If the command is successful, it returns two `{certs-generate}` commands, one of
 ----
 Validation succeeded.
 
-To use them inside a NEW {SmartProxyServer}, run this command:
-  {certs-generate} --foreman-proxy-fqdn "{SmartProxyServer}" \
+To use them inside a NEW ${smart-proxy-capitalized}, run this command:
+  {certs-generate} --foreman-proxy-fqdn "${smart-proxy-capitalized}" \
     --certs-tar "~/{smartproxy-example-com}-certs.tar" \
```
ekohl, line 42: If you want to be consistent with the output:
```suggestion
    --certs-tar "~/${smart-proxy-capitalized}-certs.tar" \
```

And the same below

ekohl (commented):
I struggle with this workflow. It's confusing to say the least.

# Summarizing current workflow

Step 1 is to run `katello-certs-check`. This is good, but there's some sentences that don't make sense.

> If you set a wildcard value * for the certificate’s Common Name CN = in the /root/smart-proxy_cert/openssl.cnf configuration file, you must add the -t foreman-proxy option to the katello-certs-check command.

We always instruct users to pass `-t foreman-proxy` so the comment is not needed.

Step 2 is to essentially tell users "read what's on the screen and follow that. The example of that output was in step 1.

In your version there's still this line:

> In this command, change Smart Proxy server to the FQDN of your Smart Proxy server.

That should be `${smart-proxy-capitalized}`.

Then step 3 is again essentially "read what's on the screen", it looks like this.

Step 4 is to copy the bundle, which is correct

Step 5 is "run the command we told you to remember in step 3".

Here I'd argue that this bit:

> If network connections or ports to Foreman are not yet open, you can set the --foreman-proxy-register-in-foreman option to false to prevent Smart Proxy from attempting to connect to Foreman and reporting errors. Run the installer again with this option set to true when the network and firewalls are correctly configured.

It's something we shouldn't support IMHO. Especially in the Satellite case where the templates feature is enabled by default it indicates a problem. If the Smart Proxy Server can't talk to Foreman, you're gonna have a bad time.

# Where to go from here?

We should change the code so the first step can be simplified. `foreman-proxy-certs-generate` should run `katello-certs-check` internally (it has all the parameters) and halt execution if it detects a problem. Then we can simplify instructions. Captured as https://projects.theforeman.org/issues/36843.

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
maximiliankolb: Applied and rebased to master.

ekohl (approved):
Thanks! I've opened https://github.com/theforeman/foreman-installer/pull/895 as a start to just make the workflow easier so less documentation is needed. Docs are simply a good way to review end-to-end steps and realize there's something wrong.

maximiliankolb:
   722316daa..dcab07b9f  3.8 -> 3.8
   60bda02f3..b3c18c9a0  3.7 -> 3.7

PR #2547, opened 2023-10-23
Add conversions appliance guide info

Lennonka:
We're creating a new downstream Satellite guide. We need to add title and URL attributes for it.
The title was agreed upon in our team. The URL must match the title in snake_case.

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/attributes-satellite.adoc
+++ b/guides/common/attributes-satellite.adoc
@@ -11,6 +11,7 @@
 :AdministeringDocURL: {BaseURL}administering_red_hat_satellite/index#
 :ConfiguringLoadBalancerDocURL: {BaseURL}configuring_capsules_with_a_load_balancer/index#
 :ContentManagementDocURL: {BaseURL}managing_content/index#
+:ConversionsApplianceDocURL: {BaseURL}converting_hosts_to_rhel_by_using_conversions_appliance/index#
 :InstallingServerDisconnectedDocURL: {BaseURL}installing_satellite_server_in_a_disconnected_network_environment/index#
 :InstallingServerDocURL: {BaseURL}installing_satellite_server_in_a_connected_network_environment/index#
 :InstallingSmartProxyDocURL: {BaseURL}installing_capsule_server/index#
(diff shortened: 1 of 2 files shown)


guides/common/attributes-titles.adoc
```diff
@@ -31,6 +31,7 @@
 :APIDocTitle: API Guide
 :HammerDocTitle: Hammer CLI Guide
 :ConfiguringVMSubscriptionsDocTitle: Configuring Virtual Machine Subscriptions in {ProjectName}
+:ConversionsApplianceDocTitle: Converting hosts to RHEL by using Conversions Appliance
```
adamlazik1, line 34:
```suggestion
:ConversionsApplianceDocTitle: Converting Hosts to RHEL by Using Conversions Appliance
```

guides/common/attributes-satellite.adoc
```diff
@@ -11,6 +11,7 @@
 :AdministeringDocURL: {BaseURL}administering_red_hat_satellite/index#
 :ConfiguringLoadBalancerDocURL: {BaseURL}configuring_capsules_with_a_load_balancer/index#
 :ContentManagementDocURL: {BaseURL}managing_content/index#
+:ConversionsApplianceDocURL: {BaseURL}converting_hosts_to_rhel_by_using_conversions_appliance/index#
```
adamlazik1, line 14: I wonder if this should be moved to `attributes-base` if for any reason someone wants to link it outside the satellite build. On the other hand it is better categorized here. What do you think?

Lennonka: That's a good point. Would it be worth moving all guides that aren't upstreamized yet to `attributes-base`?
@adamlazik1 @maximiliankolb Opinions?

adamlazik1: For me it seems like a good change but it is also a non-blocking suggestion.

maximiliankolb: Yeah, works for me.

Lennonka:
Cherry-picked:
- 5cb306073..8ec41fc32  3.8 -> 3.8
- 6700128c7..bad336608  3.7 -> 3.7
- cc2b96236..9b6ae480f  3.6 -> 3.6
- 2a8c301ad..fa6856bf7  3.5 -> 3.5

guides/common/attributes-satellite.adoc
Lennonka: Unfortunately, there's no easy way of including RH links in `attributes-base`, because RH base URL includes Satellite version number and attributes-base doesn't define Satellite version. So I'm leaving it as it is for now.

PR #2481, opened 2023-09-29
Updates for Katello 4.10.0 - Release notes

damoore044:
* [✓] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:
* [x] Foreman 3.8/Katello 4.10

PR #2560, opened 2023-10-27
Readd deprecations heading for external links

maximiliankolb:
I suggest to keep the structure based on headings in all versions in case there are external links in the sense of "docs.theforeman.org/katello-release-notes/$Version#deprecations".

Follow-up PR to https://github.com/theforeman/foreman-documentation/pull/2481/files#diff-a3acd643a9303d35f8f0d8c49bd0099a29f8576e1fa1648475da1258c8fad7deL10


ekohl:
Interesting that I didn't think of this in the past. I think it makes sense. Perhaps we should make this the default in `master` so it's automatically there when we branch? And I'd say we should then apply the same to upgrade warnings. I'm doubtful about Headline features because I think every release should have at least something.

adamlazik1 (approved):

maximiliankolb:
follow-up PR to cherry-pick placerholder sentence to master: https://github.com/theforeman/foreman-documentation/pull/2574

PR #2595, opened 2023-11-16
Document keyboard shortcuts

adamlazik1:

### DIFF
--- a/guides/common/modules/con_using-keyboard-shortcuts.adoc
+++ b/guides/common/modules/con_using-keyboard-shortcuts.adoc
@@ -1,0 +1,7 @@
+[id="using-keyboard-shortcuts_{context}"]
+= Using Keyboard Shortcuts
+
+You can use keyboard shortcuts to quickly focus search bars.
+
+* To focus the vertical navigation search bar, press `Ctrl + Shift + F`.
+* To focus the page search bar, press `/`.
--- a/guides/common/assembly_searching-and-bookmarking.adoc
+++ b/guides/common/assembly_searching-and-bookmarking.adoc
@@ -15,3 +15,5 @@ include::modules/con_managing-bookmarks.adoc[leveloffset=+1]
 include::modules/proc_creating-bookmarks.adoc[leveloffset=+2]
 
 include::modules/proc_deleting-bookmarks.adoc[leveloffset=+2]
+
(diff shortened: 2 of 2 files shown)


MariaAga (approved):

guides/common/modules/con_using-keyboard-shortcuts.adoc
```diff
@@ -0,0 +1,8 @@
+[id="using-keyboard-shortcuts_{context}"]
+= Using Keyboard Shortcuts
+
+You can use keyboard shortcuts to quickly focus search bars.
+
+To focus the vertical navigation search bar, press `CTRL + SHIFT + F`.
+
+To focus the page search bar, press `/`.
```
maximiliankolb, line 8: Can you please try https://docs.asciidoctor.org/asciidoc/latest/macros/keyboard-macro/?

Also, maybe make line 6+8 an unordered list?

adamlazik1: Seems like the macro does not work:
image
so I am reverting that.

maximiliankolb: Ah, I remember that they need ":experimental:" set. Probably not something we want to run in prod right now. See https://docs.asciidoctor.org/asciidoc/latest/macros/keyboard-macro/

Thanks for testing anyway.

mjivraja (approved):

guides/common/modules/con_using-keyboard-shortcuts.adoc
adamlazik1: Hmm, I actually did not notice that. To be honest, I wouldn't mind having the macro in the module, it does say that the macro is stable. Shall I add it back?

mjivraja: There is a catch. They have also mentioned that they might change the name of attribute (:experimental:) after their internal decision. We can delay adding macro?

maximiliankolb: If it's not cherry-picked, then trying it out is really OK for me. Let's see what happens.

adamlazik1: That as fair enough. I will merge as-is after QE review.

adamlazik1:
Merged and cherry-picked:
   6ab251459..7e290701b  3.9 -> 3.9

PR #2598, opened 2023-11-20
Update module on Email notification types

asteflova:
Changes in this PR lay the groundwork for documenting new email notification types:

* I renamed the reference section "Notification Types" to "Email Notification Types" to make its purpose clearer.
* I renamed and removed some notification types in the outdated list.
* I added tips for where to find the complete list, to help users if/when the list changes with future releases.

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.8/Katello 4.10
* [x] Foreman 3.7/Katello 4.9 (Satellite 6.14)
* [x] Foreman 3.6/Katello 4.8
* [x] Foreman 3.5/Katello 4.7 (Satellite 6.13; orcharhino 6.6)
* [x] Foreman 3.4/Katello 4.6 (EL8 only)
* [x] Foreman 3.3/Katello 4.5 on EL7 & EL8 (Satellite 6.12 on EL8 only; orcharhino 6.4/6.5 on EL8 only)
* [ ] Foreman 3.2/Katello 4.4 on EL7 & EL8
* [ ] Foreman 3.1/Katello 4.3 on EL7 & EL8 (Satellite 6.11 EL7/8; orcharhino 6.3 on EL7/8)
* We do not accept PRs for Foreman older than 3.1.

PR #2600, opened 2023-11-21
New email notifications for failed bg actions

asteflova:
This PR adds four new types of email notifications (based on https://projects.theforeman.org/issues/17748). Additionally, I changed the markup used for the list of notifications and made minor tweaks to ensure the list is ordered alphabetically.

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/ref_email-notification-types.adoc
+++ b/guides/common/modules/ref_email-notification-types.adoc
@@ -12,8 +12,8 @@ Compliance policy summary:: A summary of OpenSCAP policy reports and their resul
 Promote errata:: A notification sent only after a Content View promotion.
 It contains a summary of errata applicable and installable to hosts registered to the promoted Content View.
 This allows a user to monitor what updates have been applied to which hosts.
-Repository sync failure:: A notification sent after repository synchronization fails.
 {SmartProxy} sync failure:: A notification sent after {SmartProxy} synchronization fails.
+Repository sync failure:: A notification sent after repository synchronization fails.
 Sync errata:: A notification sent only after synchronizing a repository.
 It contains a summary of new errata introduced by the synchronization.
 
--- a/guides/common/modules/ref_email-notification-types.adoc
+++ b/guides/common/modules/ref_email-notification-types.adoc
@@ -10,7 +10,7 @@ endif::[]
 Compliance policy summary:: A summary of OpenSCAP policy reports and their results.
 Content view promote failure:: A notification sent after content view promotion fails.
 Content view publish failure:: A notification sent after content view publication fails.
-Host built:: A notification sent when a host is built.
+Host built:: A notification sent after a host is built.
 Host errata advisory:: A summary of applicable and installable errata for hosts managed by the user.
 ifdef::orcharhino[]
 {SmartProxy} sync failure:: A notification sent after {SmartProxy} synchronization fails.
--- a/guides/common/modules/ref_email-notification-types.adoc
+++ b/guides/common/modules/ref_email-notification-types.adoc
@@ -4,16 +4,24 @@
 {Project} can create the following email notifications:
 
 Audit summary:: A summary of all activity audited by {ProjectServer}.
+ifdef::satellite[]
+{SmartProxy} sync failure:: A notification sent after {SmartProxy} synchronization fails.
+endif::[]
+Compliance policy summary:: A summary of OpenSCAP policy reports and their results.
 Content view promote failure:: A notification sent after content view promotion fails.
 Content view publish failure:: A notification sent after content view publication fails.
 Host built:: A notification sent when a host is built.
 Host errata advisory:: A summary of applicable and installable errata for hosts managed by the user.
-Compliance policy summary:: A summary of OpenSCAP policy reports and their results.
+ifdef::orcharhino[]
+{SmartProxy} sync failure:: A notification sent after {SmartProxy} synchronization fails.
+endif::[]
 Promote errata:: A notification sent only after a Content View promotion.
 It contains a summary of errata applicable and installable to hosts registered to the promoted Content View.
 This allows a user to monitor what updates have been applied to which hosts.
-{SmartProxy} sync failure:: A notification sent after {SmartProxy} synchronization fails.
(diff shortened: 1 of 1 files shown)


asteflova:
@adamruzicka Can you please review?

adamruzicka (commented):
Looks good, nothing to add here

guides/common/modules/ref_notification-types.adoc
```diff
@@ -12,5 +14,7 @@ It contains a summary of errata applicable and installable to hosts registered t
 This allows a user to monitor what updates have been applied to which hosts.
 * *Puppet error state*: A notification sent after a host reports an error related to Puppet.
 * *Puppet summary*: A summary of Puppet reports.
+* *Repository sync failure*: A notification sent after repository synchronization fails.
+* *{SmartProxy} sync failure*: A notification sent after {SmartProxy} synchronization fails.
```
adamlazik1, line 18: If we are ordering this list alphabetically, then you need to place to this at different places depending on the build:
```diff
bef58c6bccdb71933c94e2e932fbe44cc3087c0e Commit
diff --git a/guides/common/modules/ref_notification-types.adoc b/guides/common/modules/ref_notification-types.adoc
index e8c0366e8..9efa698c2 100644
--- a/guides/common/modules/ref_notification-types.adoc
+++ b/guides/common/modules/ref_notification-types.adoc
@@ -4,17 +4,25 @@
 The following are the notifications created by {Project}:
 
 * *Audit summary*: A summary of all activity audited by {ProjectServer}.
+ifdef::satellite[]
+* *{SmartProxy} sync failure*: A notification sent after {SmartProxy} synchronization fails.
+endif::[]
 * *Content view promote failure*: A notification sent after content view promotion fails.
 * *Content view publish failure*: A notification sent after content view publication fails.
 * *Host built*: A notification sent when a host is built.
 * *Host errata advisory*: A summary of applicable and installable errata for hosts managed by the user.
 * *OpenSCAP policy summary*: A summary of OpenSCAP policy reports and their results.
+ifdef::orcharhino[]
+* *{SmartProxy} sync failure*: A notification sent after {SmartProxy} synchronization fails.
+endif::[]
 * *Promote errata*: A notification sent only after a Content View promotion.
 It contains a summary of errata applicable and installable to hosts registered to the promoted Content View.
 This allows a user to monitor what updates have been applied to which hosts.
 * *Puppet error state*: A notification sent after a host reports an error related to Puppet.
 * *Puppet summary*: A summary of Puppet reports.
 * *Repository sync failure*: A notification sent after repository synchronization fails.
+ifdef::foreman-el,foreman-deb,katello[]
 * *{SmartProxy} sync failure*: A notification sent after {SmartProxy} synchronization fails.
+endif::[]
 * *Sync errata*: A notification sent only after synchronizing a repository.
 It contains a summary of new errata introduced by the synchronization.
```

Feel free to use this patch, although you will need to resolve merge conflict caused by #2598.

asteflova:
Rebased and fixed markup to use definition lists for the definitions of notification types.

guides/common/modules/ref_notification-types.adoc
asteflova: That's a very good point on how to order the list. Ordering it alphabetically would normally make sense, but I don't think it's worth the maintenance burden of three ifdefs (the same description being included three times, the risk of someone accidentally deleting an endif which asciidoctor doesn't report...).

Another option would be to look into how the notification types are ordered in the Web UI and mirror that.

I'll investigate.

asteflova: I decided to go with the order in Web UI.

asteflova:
I reordered the list item to match the order in the Web UI.

Tech review is done, so I believe that all that is missing for this to be merged is a style ack.

adamruzicka:
> I reordered the list item to match the order in the Web UI.

Are you sure the ordering in webui is stable?

asteflova:
> > I reordered the list item to match the order in the Web UI.
> 
> Are you sure the ordering in webui is stable?

What I did was check how the items are ordered in snap 38 and in the Foreman demo testing instance. The proxy-related item was named differently in each (Capsule vs Proxy), but it was located on the same line in the list. That led me to believe that the ordering is stable and I can use it in the docs too. Is there something I missed @adamruzicka?

adamruzicka:
At least on my machine:
By default it looks like this 
image

But if I subscribe to something (capsule sync failure in this case) it gets shown on top
image

asteflova:
Oh, I didn't realize the order changes like this. Thanks for pointing it out!

Let's circle back to the suggestion to use ifdefs then, because it does seem to fit our needs best after all (thanks @adamlazik1!). The latest version of this PR ensures that the list of email notification types is ordered alphabetically in any doc build.

guides/common/modules/ref_email-notification-types.adoc
```diff
@@ -3,14 +3,26 @@
...
-* *Compliance policy summary*: A summary of OpenSCAP policy reports and their results.
-* *Promote errata*: A notification sent only after a Content View promotion.
+Audit summary:: A summary of all activity audited by {ProjectServer}.
+ifdef::satellite[]
+{SmartProxy} sync failure:: A notification sent after {SmartProxy} synchronization fails.
+endif::[]
+Compliance policy summary:: A summary of OpenSCAP policy reports and their results.
+Content view promote failure:: A notification sent after content view promotion fails.
+Content view publish failure:: A notification sent after content view publication fails.
+Host built:: A notification sent when a host is built.
```
maximiliankolb, line 13: "is built" or "has been built"? Or maybe "after" instead of "when"?

maximiliankolb (commented):
One small suggestion, rest LGTM.

FYI: I could not reproduce this order in the GUI. Can someone else please ACK this? Tested on Foreman 3.8 with a Smart Proxy server.

guides/common/modules/ref_email-notification-types.adoc
asteflova: Let's go with "after". Addressed.

maximiliankolb (approved):
LGTM

PR #2604, opened 2023-11-24
Improve Convert2RHEL [priority]

Lennonka:
A couple of formal improvements.

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.8/Katello 4.10
* [x] Foreman 3.7/Katello 4.9 (Satellite 6.14)
* [x] Foreman 3.6/Katello 4.8
* [x] Foreman 3.5/Katello 4.7 (Satellite 6.13; orcharhino 6.6)
* [x] Foreman 3.4/Katello 4.6 (EL8 only)
* [x] Foreman 3.3/Katello 4.5 on EL7 & EL8 (Satellite 6.12 on EL8 only; orcharhino 6.4/6.5 on EL8 only)
* [x] Foreman 3.2/Katello 4.4 on EL7 & EL8
* [x] Foreman 3.1/Katello 4.3 on EL7 & EL8 (Satellite 6.11 EL7/8; orcharhino 6.3 on EL7/8)
* We do not accept PRs for Foreman older than 3.1.

PR #2605, opened 2023-11-24
Add pre-conversion check in Convert2RHEL

Lennonka:
Documents https://github.com/theforeman/foreman_remote_execution/pull/840, i.e. Convert2RHEL analyze template

* [x] Foreman 3.9 if available

### DIFF
--- a/guides/common/modules/con_converting-a-host-to-rhel.adoc
+++ b/guides/common/modules/con_converting-a-host-to-rhel.adoc
@@ -63,6 +63,23 @@ endif::[]
 ifndef::managing-hosts[]
 For more information, see {ManagingHostsDocURL}Registering_Hosts_by_Using_Global_Registration_managing-hosts[Registering Hosts by Using Global Registration] in _{ManagingHostsDocTitle}_.
 endif::[]
+. Run the pre-conversion analysis on the host group to verify if your hosts are ready for the conversion.
+Execute a remote job with the following settings:
+* **Job category**: `Convert 2 RHEL`
+* **Job template**: `Convert2RHEL analyze`
+
++
+ifdef::managing-hosts[]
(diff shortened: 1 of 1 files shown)


Lennonka:
@stejskalleos In which Foreman can this be expected? 3.9+?

Lennonka:
This should be merged after #2604

guides/common/modules/con_converting-a-host-to-rhel.adoc
```diff
@@ -62,6 +62,23 @@ endif::[]
...
+ifdef::managing-hosts[]
+For more information, see xref:executing-a-remote-job_{context}[].
+endif::[]
+ifndef::managing-hosts[]
+For more information, see {ManagingHostsDocURL}executing-a-remote-job_managing-hosts[Executing a Remote Job] in _{ManagingHostsDocTitle}_.
+endif::[]
+
++
+Review pre-conversion analysis reports and resolve any issues that prevent the conversion.
+Repeat this step until you resolve all issues.
```
stejskalleos, line 80: I would add `resolve all issues that are blocking the conversion`

Some of the issues are not always problems users have to solve.

stejskalleos:
In which Foreman can this be expected? 3.9+? Yes
Is this step optional? Well, yes and know. Yes if you are an adventurous person, otherwise no. You can start the conversion process without it, but that doesn't sound to me like a good idea.
Should this step replace any prerequisites? I don't think so.

adamlazik1 (approved):
Looks good stylewise.

Lennonka:
I've verified that this has to be cherry-picked:
- f8376970c..77edff987  3.8 -> 3.8
- bbcd4e289..4001a5405  3.7 -> 3.7

PR #2623, opened 2023-12-05
Add procedure to uninstall foreman_openscap

maximiliankolb:
procedure is based on `guides/common/modules/proc_uninstalling-amazon-ec2-plugin.adoc`

### DIFF
--- a/guides/common/modules/proc_uninstalling-openscap-plugin.adoc
+++ b/guides/common/modules/proc_uninstalling-openscap-plugin.adoc
@@ -1,0 +1,15 @@
+[id="Uninstalling_OpenSCAP_Plugin_{context}"]
+= Uninstalling OpenSCAP Plugin
+
+If you have previously installed the OpenSCAP plug-in but do not use it anymore to scan managed hosts, you can uninstall it from your {ProjectServer}.
+
+.Procedure
+. Uninstall the OpenSCAP plug-in from your {ProjectServer}:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# {foreman-installer} --no-enable-foreman-cli-openscap \
+--no-enable-foreman-plugin-openscap \
+--no-enable-foreman-proxy-plugin-openscap
+----
+. Optional: In the {ProjectWebUI}, navigate to *Administer* > *About* and select the *Plugins* tab to verify the removal of the OpenSCAP plug-in.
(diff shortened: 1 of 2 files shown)


guides/common/modules/proc_uninstalling-openscap-plugin.adoc
```diff
@@ -0,0 +1,15 @@
...
+If you have previously installed the OpenSCAP plug-in but do not use it anymore to scan managed hosts, you can uninstall it from your {ProjectServer}.
+
+.Procedure
+. Uninstall the OpenSCAP plug-in from your {ProjectServer}:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# {foreman-installer} --no-enable-foreman-cli-openscap \
+--no-enable-foreman-plugin-openscap \
+--no-enable-foreman-proxy-plugin-openscap
```
ekohl, line 13: Technically this disables management in the installer, but that simply means it's unmanaged. It remains installed and active.

ekohl (changes requested):
We don't really support this. There is no code to remove the database entries from Foreman and there is no code to actually uninstall packages.

Technically you can use `--foreman-proxy-plugin-openscap-enabled false` to disable the Foreman Proxy module.

maximiliankolb:
Can I salvage this PR by just removing the installed packages first similar to [Uninstalling Amazon EC2 Plugin](https://docs.theforeman.org/nightly/Provisioning_Hosts/index-katello.html#Uninstalling_Amazon_EC2_Plugin_ec2-provisioning)? Or close this for now with "it's unsupported"?

ekohl:
There are 3 components to it:
* CLI
* Foreman
* Foreman proxy

For both CLI & Foreman Proxy it would be sufficient to remove the packages, but for Foreman you also have the DB tables. We don't have a way to clean that up. For some we have a [CleanupHelper](https://github.com/theforeman/foreman/blob/develop/lib/cleanup_helper.rb), but removing Foreman plugins is generally speaking not supported.

For the EC2 compute provider it's a bit different, because the DB model is always the same, regardless of `foreman-ec2` being installed or not. Removal thus doesn't matter.

sbernhard (approved):
LGTM

maximiliankolb:
Closing PR because it's not officially supported to uninstall plug-ins. Eventually, all plug-ins should defined procedures on what happens when they're uninstalled.

PR #2630, opened 2023-12-07
Fix firewall commands for ports and services

mjivraja:
Earlier in the last release version, we had no services to add for the firewall. But with the latest version i.e. going to be released, we add services and ports in the firewall. But, the firewall command does not accept the services and ports together in a single instance and is therefore split here. Also, adding --permanent to the command in the first place replaces the requirement of a new step for making it persistent.

https://bugzilla.redhat.com/show_bug.cgi?id=2253246

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
+++ b/guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
@@ -14,7 +14,7 @@ ifdef::katello,satellite,orcharhino[]
 +
 [options="nowrap"]
 ----
-# firewall-cmd --permanent \
+# firewall-cmd \
 --add-port="5647/tcp" \
 --add-port="8000/tcp" \
 --add-port="9090/tcp"
@@ -24,7 +24,7 @@ endif::[]
 +
 [options="nowrap"]
 ----
-# firewall-cmd --permanent \
+# firewall-cmd \
 --add-service=dns \
 --add-service=dhcp \
 --add-service=tftp \
@@ -34,6 +34,12 @@ ifndef::katello,satellite,orcharhino[]
 --add-service=foreman-proxy \
 endif::[]
 --add-service=puppetmaster
+----
+. Make the changes persistent:
++
+[options="nowrap"]
+----
+# firewall-cmd --runtime-to-permanent
 ----
 
 include::snip_verify-firewall-settings.adoc[]
--- a/guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
+++ b/guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
@@ -9,31 +9,31 @@ For more information on the ports used, see {InstallingServerDocURL}satellite-po
 include::snip_firewalld.adoc[]
 
 .Procedure
-. To open the ports for client to {Project} communication, enter the following command on the base operating system that you want to install {Project} on:
+ifdef::katello,satellite,orcharhino[]
+. Open the ports for clients on {ProjectServer}:
 +
 [options="nowrap"]
 ----
-# firewall-cmd \
+# firewall-cmd --permanent \
+--add-port="5647/tcp" \
+--add-port="8000/tcp" \
+--add-port="9090/tcp"
+----
+endif::[]
+. Allow access to services on {ProjectServer}:
++
+[options="nowrap"]
+----
+# firewall-cmd --permanent \
 --add-service=dns \
 --add-service=dhcp \
 --add-service=tftp \
 --add-service=http \
 --add-service=https \
-ifdef::katello,satellite,orcharhino[]
---add-port="5647/tcp" \
---add-port="8000/tcp" \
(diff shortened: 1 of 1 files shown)


guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
```diff
@@ -9,31 +9,33 @@ For more information on the ports used, see {InstallingServerDocURL}satellite-po
 include::snip_firewalld.adoc[]
 
 .Procedure
-. To open the ports for client to {Project} communication, enter the following command on the base operating system that you want to install {Project} on:
+Enter the following commands on the base operating system that you want to install {Project} on.
+
```
Lennonka, line 13: This sentence here doesn't adhere to our formatting standards and it looks bad. Also, I don't think it's necessary, because the same is said right before the procedure. Can we remove it?

guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
```diff
@@ -9,31 +9,33 @@ For more information on the ports used, see {InstallingServerDocURL}satellite-po
...
+ifdef::katello,satellite,orcharhino[]
+. Open the ports for client:
```
Lennonka, line 15:
```suggestion
. Open the ports for clients on {ProjectServer}:
```
And we might this here instead.

guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
```diff
@@ -9,31 +9,33 @@ For more information on the ports used, see {InstallingServerDocURL}satellite-po
...
++
+[options="nowrap"]
+----
+# firewall-cmd --permanent \
+--add-port="5647/tcp" \
+--add-port="8000/tcp" \
+--add-port="9090/tcp"
+----
+endif::[]
+. Add the services:
```
Lennonka, line 25:
```suggestion
. Allow access to services on {ProjectServer}:
```
And here.

Lennonka (changes requested):
Suggestions for phrasing and style. I can't verify this for technical accuracy.

ekohl:
This was a regression introduced in 338f86588c0e6151003a9abc7c2e2f6841c65445.

guides/common/modules/proc_enabling-client-connections-to-satellite.adoc
```diff
@@ -9,31 +9,31 @@ For more information on the ports used, see {InstallingServerDocURL}satellite-po
...
 --add-service=foreman-proxy \
 endif::[]
 --add-service=puppetmaster
 ----
-. Make the changes persistent:
-+
-[options="nowrap"]
-----
-# firewall-cmd --runtime-to-permanent
-----
```
ekohl, line 37: This part is incorrect. You either make changes to the permanent config, or the runtime config. Your current approach only applies it to the permanent config and doesn't apply it to runtime.

mjivraja: I have added `firewall-cmd --permanent` instead, for ports and services both. Do you want me to restore these removed lines?

Lennonka: @ekohl ^^ What solution do you suggest instead?

ekohl: My understanding of firewalld is that it's not the same. See https://firewalld.org/documentation/configuration/runtime-versus-permanent.html for more information, but please restore it as it previously was.

Lennonka: @ekohl How about setting it as permanent and then reloading the service? That would work too, no?

ekohl: @Lennonka I don't know firewalld well enough to answer. To me it sounds like the right thing but it sounds like something you'd do across the entire documentation and not in this PR.

Perhaps there's some internal guidance within Red Hat / RHEL.

mjivraja: My understanding is that `firewall-cmd --runtime-to-permanent` does not require the `firewall-cmd --reload` but `firewall-cmd --permanent` does (Lennonka's suggestion).

After careful consideration, we should just split ports and services into individual sets here IMO and restore them in the previous state (Ewoud's suggestion). 

CC: @sayan3296

sayan3296: +1 
It makes sense to use `--permanent` while adding ports\services in firewalld, followed by `firewall-cmd --reload`. That's rather a widely acceptable and familiar approach for many system-admins.

Lennonka (approved):
LGTM

Lennonka:
Cherry-picked:
- 9334ae602..23a1de660  3.9 -> 3.9

PR #2608, opened 2023-11-24
Revise the type of HTTP Proxy mentioned

mjivraja:
Presently, in setting up the Project in the IPv6 network, we mention only the IPv4 type of HTTP Proxy server which is incorrect. Because, the HTTP Proxy server configured with IPv4 only cannot communicate with the Project configured in the IPv6 network, except CDN. Hence, it should support both, IPv4 and IPv6.

Please cherry-pick my commits into:

* [X] Foreman 3.8/Katello 4.10
* [X] Foreman 3.7/Katello 4.9 (Satellite 6.14)
* [X] Foreman 3.6/Katello 4.8
* [X] Foreman 3.5/Katello 4.7 (Satellite 6.13; orcharhino 6.6)
* [X] Foreman 3.4/Katello 4.6 (EL8 only)
* [X] Foreman 3.3/Katello 4.5 on EL7 & EL8 (Satellite 6.12 on EL8 only; orcharhino 6.4/6.5 on EL8 only)
* [X] Foreman 3.2/Katello 4.4 on EL7 & EL8
* [X] Foreman 3.1/Katello 4.3 on EL7 & EL8 (Satellite 6.11 EL7/8; orcharhino 6.3 on EL7/8)
* We do not accept PRs for Foreman older than 3.1.

PR #2651, opened 2024-01-02
Reword IPv4 and IPv6 HTTP Proxy requirement (#2608)

mjivraja:
Presently, in setting up the Project in the IPv6 network, we mention only IPv4 type of HTTP Proxy server which is incorrect. Because, HTTP Proxy server configured with IPv4 only cannot communicate with Project configure in IPv6 network, except CDN.
Hence, it should support both, IPv4 and IPv6.

(cherry picked from commit 28d8c905c5bc6a9aea3315d2e8ceb0fcb0ba9d2c)

### DIFF
--- a/guides/common/modules/con_requirements-for-installation-in-an-ipv6-network.adoc
+++ b/guides/common/modules/con_requirements-for-installation-in-an-ipv6-network.adoc
@@ -8,23 +8,18 @@ ifndef::satellite[]
 For other operating systems, copy the newest GRUB build to `/var/lib/tftpboot/grub2/grubx64.efi`.
 endif::[]
 
-* You must deploy an external DHCP IPv6 server as a separate unmanaged service to bootstrap clients into GRUB2, which then configures IPv6 networking either using DHCPv6 or or assigning static IPv6 address.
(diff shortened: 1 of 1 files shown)


guides/common/modules/con_requirements-for-installation-in-an-ipv6-network.adoc
```diff
@@ -3,28 +3,18 @@
 
 Before installing {Project} in an IPv6 network, ensure that you meet the following requirements:
 
-ifndef::satellite[]
-* If you plan to provision hosts from {Project} or {SmartProxy}, you must install {Project} and {SmartProxies} on a system with `grub2` version 2.05 or higher or system with fixes for HTTP Boot, for example CentOS version 7.9 or 8.3 or higher.
-For other operating systems, copy the newest GRUB build to `/var/lib/tftpboot/grub2/grubx64.efi`.
-endif::[]
```
maximiliankolb, line 9: This is different to the original PR. Is this by design?

mjivraja: Apologies, I assumed this to be part of `master` changes but it's not. Restoring back in all 3 PRs.

maximiliankolb (approved):

PR #2685, opened 2024-01-19
Add more details on host group settings and nested host groups

asteflova:
[BZ#2255404](https://bugzilla.redhat.com/show_bug.cgi?id=2255404) requests expanding the existing description of host group settings, nested host groups, and how the settings are inherited through the hierarchy of nested host groups. Because the conceptual information is now more extensive, I split it into a separate module, which then required moving all host group-related modules into a separate assembly.

* [ ] Foreman 3.6/Katello 4.8


asteflova:
Hi @jeremylenz This is my attempt to resolve [BZ#2255404](https://bugzilla.redhat.com/show_bug.cgi?id=2255404), can you please review? It turned out to be pretty tricky, so I'd appreciate it if you could read it carefully. For you tech review, reading only `con_host-group-settings-and-nested-host-groups.adoc` ([2.8.1. Host Group Settings and Nested Host Groups](https://theforeman-foreman-documentation-preview-pr-2685.surge.sh/nightly/Managing_Hosts/index-satellite.html#Host_Group_Settings_and_Nested_Host_Groups_managing-hosts)) should be enough.

guides/common/modules/con_host-group-settings-and-nested-host-groups.adoc
```diff
@@ -0,0 +1,111 @@
+[id="Host_Group_Settings_and_Nested_Host_Groups_{context}"]
+= Host Group Settings and Nested Host Groups
+
+A host group can define many settings for hosts, such as Lifecycle Environment, Content View, or Ansible roles that are available to the hosts.
+
+NOTE: When you change the properties of a host group, the new settings do not propagate to the hosts assigned to the host group.
```
jeremylenz, line 6: This is important enough behavior that I think it should be elevated beyond a note (I almost skipped over it).

jeremylenz (commented):
This is looking great! Everything seems accurate so far, besides the few comments below.

Also would be good to get input from @ShimShtein and/or @ares

guides/common/modules/con_host-group-settings-and-nested-host-groups.adoc
asteflova: Let's go with IMPORTANT, then. Let me know if this doesn't work for you.

asteflova:
> This is looking great! Everything seems accurate so far, besides the few comments below.
> 
> Also would be good to get input from @ShimShtein and/or @ares

Thanks!

Also, @ares, just in case -- I'm aware of [BZ#2253496](https://bugzilla.redhat.com/show_bug.cgi?id=2253496) which seems related, but I want to handle it separately.

jeremylenz:
And see also :) https://github.com/Katello/katello/pull/10841

guides/common/modules/con_host-group-settings-and-nested-host-groups.adoc
jeremylenz: Works for me :+1:

guides/common/modules/con_host-group-settings-and-nested-host-groups.adoc
```diff
@@ -0,0 +1,111 @@
...
+|
+|*custom.example.com*
+
+|===
+
+In this hierarchy:
+
+* `Base` defines the settings common to all hosts. All hosts use {EL}{nbsp}8.8 as their operating system.
+* `webserver1.example.com` and `webserver2.example.com` use the `linux-system-roles.selinux` Ansible role.
+`storage1.example.com` and `storage2.example.com` use the `linux-system-roles.postfix` Puppet class.
```
ShimShtein, line 49:
```suggestion
`storage1.example.com` and `storage2.example.com` use the `linux-system-roles.postfix` Ansible role.
```
To be consistent across the document

guides/common/modules/con_host-group-settings-and-nested-host-groups.adoc
```diff
@@ -0,0 +1,111 @@
...
+* `custom.example.com` inherits none of the Puppet classes defined by the host groups.
```
ShimShtein, line 50:
```suggestion
* `custom.example.com` inherits none of the Puppet classes or Ansible roles defined by the host groups.
```

ShimShtein (commented):
Worth also mentioning that the values are copied into the host object. This means that changes made later to the hostgroup are not propagated to hosts that are already provisioned.
In the UI there is an option to edit the host instance and force re-inheritance of the properties for the current state.

guides/common/assembly_working-with-host-groups.adoc
Lennonka, line 1: I believe you forgot to include this new assembly in `master.adoc` :)

Lennonka (changes requested):
Great job!
I'll review it more thoroughly once the new assembly is available in the preview.

guides/common/assembly_working-with-host-groups.adoc
asteflova: Thanks for taking a look! I included the assembly in assembly_administering-hosts.adoc (even though I hate nested assemblies, ugh!), but actually... yes, it might work better in master.adoc after all: https://github.com/theforeman/foreman-documentation/pull/2685/commits/75bb5a65d171715fce40c3cbf7b97ebdcfb9585d
If anyone objects, I can always revert.

asteflova:
> Great job! I'll review it more thoroughly once the new assembly is available in the preview.

@Lennonka Please keep in mind that most of the new assembly's content are actually old procedures that I just moved and otherwise left untouched :) Letting you know just to be sure you don't review docs that I consider out of scope of this PR.

guides/common/modules/con_working-with-host-groups.adoc
```diff
@@ -0,0 +1,6 @@
+[id="Working_with_Host_Groups_{context}"]
+= Working with Host Groups
+
+A host group is a template for common host settings.
+Adding host settings and attributes for every new host is time consuming.
+If you use host groups, you can apply common attributes to hosts that you create.
```
Lennonka, line 6: Perhaps this intro could be rephrased to make a bit clearer that it's "define settings once, apply them many times"?
I'm not sure using the term "template" is a good idea, because it's a technical term for stuff like job templates, provisioning templates, report templates etc. (Non-blocker, I don't have a better idea.)
Also, inconsistency in terminology -- attributes vs. properties. We should stick with a single term, ideally.

guides/common/modules/con_host-group-settings-and-nested-host-groups.adoc
```diff
@@ -0,0 +1,111 @@
...
+* `Base` defines the settings common to all hosts. All hosts use {EL}{nbsp}8.8 as their operating system.
+* `webserver1.example.com` and `webserver2.example.com` use the `linux-system-roles.selinux` Ansible role.
+`storage1.example.com` and `storage2.example.com` use the `linux-system-roles.postfix` Ansible role.
+* `custom.example.com` inherits none of the Ansible roles defined by the host groups.
+====
+
+{Project} applies host settings in the following order when nesting host groups:
+
+* Host settings take priority over host group settings.
+* Nested host group settings take priority over parent host group settings.
```
Lennonka, line 56: I'd move this above the first example.

guides/common/modules/con_host-group-settings-and-nested-host-groups.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+* *webserver1.example.com* and *webserver2.example.com* inherit these settings:
+** The `{EL}{nbsp}8.9` operating system from host group *Webservers*
+** The `linux-system-roles.selinux` Ansible role from host group *Base*
+* *storage1.example.com* and *storage2.example.com* inherit these settings:
+** The `{EL}{nbsp}8.8` operating system from host group *Base*
+** The `linux-system-roles.postfix` Ansible role from host group *Storage*
+* *custom.example.com* inherits this setting:
+** The `{EL}{nbsp}8.8` operating system from host group *Base*
+
```
Lennonka, line 110: Why not having these explanations in the last column inside the table? Similarly for the other example.

Lennonka (changes requested):
A couple of suggestions.

guides/common/modules/con_host-group-settings-and-nested-host-groups.adoc
jeremylenz: Speaking of attributes vs. properties, we're calling them a third thing here (settings) :slightly_smiling_face:

guides/common/modules/con_working-with-host-groups.adoc
jeremylenz: Re: 'templates' - We're consistently inconsistent. We do similar things all over the place: content sources vs. Alternate Content Sources; host collections vs. host groups; permissions filters vs. content view filters; etc.

Maybe "acts like a template" is a bit clearer? But I think using "template" is fine, given the above.

asteflova: I rewrote the intro in the spirit you suggested @Lennonka, let me know if it doesn't work for you like this.

I kept the word "template" here because it's actually a more accurate description than "[host] group". See also https://bugzilla.redhat.com/show_bug.cgi?id=2253496#c12 (private comment explaining that "template" describes the behavior of host settings inheritance better than "group"). I understand that the word is used for different purposes within the project, but I believe it's still worth it to use it here as well.

guides/common/modules/con_host-group-settings-and-nested-host-groups.adoc
asteflova: Moved. I also changed all occurrences of 'attributes' and 'properties' to 'settings'. (There are some left in the procedures that follow, but I'll try to take care of that in https://github.com/theforeman/foreman-documentation/pull/2689.)

guides/common/modules/con_host-group-settings-and-nested-host-groups.adoc
asteflova: I thought about this, and I must say I like the explanations where there are now. If I added them to the table, the table might get too information heavy. As always, let me know if you see it differently :)

Lennonka: I don't think it's necessarily a bad thing. When I was going through the table, I was expecting the "results of the maths" in the last column and I was surprised to find them under the tables.
Like this, the information is spread over too much area. Also, consider how we display tables on the CP: The user can enlarge the table and then they would to exit the table to read the rest of the information.
But I'm leaving the final decision to you.

guides/common/modules/con_host-group-settings-and-nested-host-groups.adoc
```diff
@@ -0,0 +1,111 @@
...
+Only Puppet class settings get updated on hosts after you change them in the host group.
+
+You can create a hierarchy of host groups.
+Aim to have one base level host group that represents all hosts in your organization and provides general settings, and then nested groups that provide specific settings.
+
+.Nested Host Group Hierarchy
+====
+You create a top-level host group named `Base` and two nested host groups named `Webserver` and `Storage`.
+The nested host groups are associated with multiple hosts.
+You also create host `custom.example.com` that is not associated with any host group.
```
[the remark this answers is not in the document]
asteflova: That's a very good catch! Fixed.

guides/common/modules/con_host-group-settings-and-nested-host-groups.adoc
asteflova: Thanks for explaining @Lennonka. I gave it a try (have I mentioned how I hate tables in asciidoc..?). Can you take another look?

asteflova:
Switching back to draft because I think I should rearrange a few bits around the example blocks. But feel free to keep reviewing in the meantime.

asteflova:
I moved a piece of general information about nested host groups out of the example boxes, and slightly tweaked the example boxes to match each other.

Lennonka (approved):
Supercool.

jeremylenz (approved):
Nice work! LGTM :+1:

guides/common/modules/con_working-with-host-groups.adoc
```diff
@@ -0,0 +1,6 @@
+[id="Working_with_Host_Groups_{context}"]
+= Working with Host Groups
+
+A host group acts as a template for common host settings.
+Instead of adding host settings individually for each host, use host groups to define common settings once and apply them across multiple hosts.
```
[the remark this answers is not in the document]
Lennonka: I'm not sure about the "new", though. If I reassign a registered host to another host group, the settings of the new host group will be applied and used e.g. during re-provisioning, right? @jeremylenz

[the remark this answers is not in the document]
jeremylenz: As I understand it, inherited attributes are applied on every host update, but _only_ for those attributes not already set on the host. To achieve this the host form sends all existing host attributes on any update, as well as any you're changing. Any attributes not set this way will then be inherited from the new host group. cc @ShimShtein to double check me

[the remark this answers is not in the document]
jeremylenz: > applied and used e.g. during re-provisioning

I'm not entirely sure what re-provisioning means here exactly?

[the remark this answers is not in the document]
asteflova: @Lennonka I agree with the tweaks for consistent terminology, so I applied that part of your suggestion: https://github.com/theforeman/foreman-documentation/pull/2685/commits/906b8533990c8de09c587591095f8c297834eec3
But I'd be super careful about using words like "new host" etc. at this point. Because the way I understand the behavior, the changes apply differently to newly created hosts and to existing hosts (it's part of why it's more of a "host template" than "host group").

My plan is to explain this in the procedures for adding a new host, changing an existing host's host group, adding an existing host to a host group etc. (WIP PR https://github.com/theforeman/foreman-documentation/pull/2689). Explaining all of this here in the conceptual parts of the assembly would just give users a headache.

Can we agree on this @Lennonka? Once the procedure modules are updated too, we can revisit this issue and see if it works like this or if some updates to the conceptual parts of the assembly are perhaps needed too.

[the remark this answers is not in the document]
Lennonka: Totally! Thank you

PR #2689, opened 2024-01-23
WIP Describe which settings get updated after host group change

asteflova:
Still work in progress, don't review yet.

Partially a follow-up for https://github.com/theforeman/foreman-documentation/pull/2685

Will include https://bugzilla.redhat.com/show_bug.cgi?id=2253496, https://github.com/Katello/katello/pull/10841, and https://github.com/theforeman/foreman-documentation/pull/2685#pullrequestreview-1835363249. Also, I want to make sure the procedures use consistent terminology for "host [group] settings" (not "attributes" or "properties").

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_adding-a-host-to-a-host-group.adoc
+++ b/guides/common/modules/proc_adding-a-host-to-a-host-group.adoc
@@ -12,3 +12,4 @@ You can add a host to a host group in the {ProjectWebUI}.
 
 .Verification
 * The *Details* card under the *Overview* tab now shows the host group your host belongs to.
+The host settings defined in the host group are now applied to the host.
--- a/guides/common/modules/proc_creating-a-host.adoc
+++ b/guides/common/modules/proc_creating-a-host.adoc
@@ -2,6 +2,7 @@
 = Creating a host in {ProjectName}
 
 Use this procedure to create a host in {ProjectName}.
+
 To use the CLI instead of the {ProjectWebUI}, see the xref:cli-creating-a-host_{context}[].
 
 .Procedure
--- a/guides/common/modules/proc_changing-the-host-group-of-a-host.adoc
+++ b/guides/common/modules/proc_changing-the-host-group-of-a-host.adoc
@@ -1,9 +1,12 @@
 [id="Changing_the_Host_Group_of_a_Host_{context}"]
 = Changing the host group of a host
 
-Use this procedure to change the Host Group of a host.
+Use this procedure to change the host group of a host.
 
+[NOTE]
+====
 If you reprovision a host after changing the host group, the fresh values that the host inherits from the host group will be applied.
+====
 
 .Procedure
 . In the {ProjectWebUI}, navigate to *Hosts* > *All Hosts*.
@@ -14,3 +17,4 @@ If you reprovision a host after changing the host group, the fresh values that t
 
 .Verification
 * The *Details* card under the *Overview* tab now shows the host group your host belongs to.
+The host settings now show the settings defined by the host group.


guides/common/modules/proc_adding-a-host-to-a-host-group.adoc
```diff
@@ -3,6 +3,8 @@
 
 You can add a host to a host group in the {ProjectWebUI}.
 
+NOTE: Adding a host to a host group updates the host settings to the values defined in the host group.
```
apinnick, line 6: Is there a reason for not using the Asciidoc note format?

maximiliankolb: I think so far we have somestimes used "NOTE: ..." if it was only a single line. I am leaning towards `==== > [NOTE] > .... > ====`, but am OK with both.

I would be happy to create a PR to fix this if we agree on it beforehand.

asteflova: Per [Asciidoctor Docs](https://docs.asciidoctor.org/asciidoc/latest/blocks/admonitions/), a single para note is to be introduced by `NOTE:`. Using `[NOTE]` with a block is to be used for complex content (more than a single paragraph, I assume).

Based on this, I'd prefer not to mandate block notes only. Or is there a reason to avoid simple single para notes that I'm not seeing? @apinnick

apinnick: The [RH Supplementary Style Guide](https://redhat-documentation.github.io/supplementary-style-guide/#admonitions) uses the block style for admonitions. [OpenShift docs](https://github.com/openshift/openshift-docs/blob/main/contributing_to_docs/doc_guidelines.adoc#admonitions) requires the block style. I do not know whether RHEL has a similar requirement.

asteflova: @apinnick I asked around and RHEL apparently has the same guideline: use code blocks rather than the "shorthand" version. I'll fix it here and let @maximiliankolb decide if he wants to fix it repo-wide :)

apinnick: Or I could do it with yet another huge 'sed' command since I have so much time on my hands. :-)

apinnick: Surprisingly, there are only 21 occurrences of `NOTE: text` in the repo! That's doable.

guides/common/modules/proc_creating-a-host.adoc
```diff
@@ -2,6 +2,14 @@
 = Creating a host in {ProjectName}
 
 Use this procedure to create a host in {ProjectName}.
+
+[NOTE]
+====
+If you specify a host group when creating a new host, all settings defined in the host group apply to the new host.
+
+Changing the host group of the host later will not update the following host settings: Lifecycle Environment and Content View. A host inherits the Content View and Lifecycle Environment settings from a host group at the host creation, not upon host group change.
```
bangelic, line 10:
```suggestion
Changing the host group of the host later will not update the following host settings: Lifecycle Environment and Content View. 
A host inherits the Content View and Lifecycle Environment settings from a host group at the host creation, not upon host group change.
```
One sentence per line rule.

guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -27,52 +27,64 @@ endif::[]
 . In the {ProjectWebUI}, navigate to *Hosts* > *Register Host*.
 . Optional: Select a different *Organization*.
```
apinnick, line 28: If you want to streamline this procedure, you could combine the optional steps: "Optional: You can select a different organization, location, host group, or operating system." 

Alternatively, you could omit the optional steps if their meaning is self-evident.

guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -27,52 +27,64 @@ endif::[]
...
-Copy the CA file to the `/etc/pki/ca-trust/source/anchors/` directory on your host and enter the following commands:
+Copy the CA file to the `/etc/pki/ca-trust/source/anchors/` directory on hosts and enter the following commands:
 +
 [options="nowrap", subs="+quotes,attributes"]
 ----
 # update-ca-trust enable
 # update-ca-trust
 ----
++
+Then register the hosts with a secure `curl` command, such as:
```
apinnick, line 66: ISG says that lead-in sentences must be a complete sentence. 

Suggestion: "Register the hosts with a secure curl command as in the following example:" with a callout for the "--insecure" option. Using a callout saves space as well.

guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -7,13 +7,13 @@ You can register a host by using registration templates and set up various integ
 * Your user account has a role assigned that has the `create_hosts` permission.
 * You must have root privileges on the host that you want to register.
 ifdef::satellite,orcharhino[]
-* {ProjectServer}, any {SmartProxyServers}, and your host must be synchronized with the same NTP server, and have a time synchronization tool enabled and running.
-* An activation key must be available for your host.
+* {ProjectServer}, any {SmartProxyServers}, and all hosts must be synchronized with the same NTP server, and have a time synchronization tool enabled and running.
```
apinnick, line 10: I recommend rewriting this prerequisite so that it is active voice. "Ensure that ...." or "Synchronize ... with the same NTP server." 

"time synchronization tool enabled and running" looks like a separate prerequisite.

guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -27,52 +27,64 @@ endif::[]
...
+If an attacker, located in the network between {Project} and a host, fetches the CA file from the first insecure call, the attacker will be able to access the content of the API calls to and from the registered host and the JSON Web Tokens (JWT).
+Therefore, if you have chosen to deploy SSH keys during registration, the attacker will be able to access the host using the SSH key.
 +
-Instead, you can manually copy and install the CA file on each host before registration.
+Instead, you can manually copy and install the CA file on each host before registering the host.
 +
 To do this, find where {Project} stores the CA file by navigating to *Administer* > *Settings* > *Authentication* and locating the value of the *SSL CA file* setting.
 +
-Copy the CA file to the `/etc/pki/ca-trust/source/anchors/` directory on your host and enter the following commands:
+Copy the CA file to the `/etc/pki/ca-trust/source/anchors/` directory on hosts and enter the following commands:
```
apinnick, line 58: You use the singular "host" above and plural "hosts" here. Perhaps change "hosts" to "each host". That makes the change in number less noticeable and still conveys that the CA certificate must be copied to each host.

guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -27,52 +27,64 @@ endif::[]
...
++
+The following is an example of the `curl` command with the `--insecure` option:
++
+[options="nowrap", subs="+quotes,attributes"]
+----
+# curl -sS --insecure https://{foreman-example-com}/register ...
+----
 . Select the *Advanced* tab.
-. Optional: From the *Setup REX* list, select whether you want to deploy {Project} SSH keys to your host or not.
+. Optional: From the *Setup REX* list, select whether you want to deploy {Project} SSH keys to hosts or not.
```
apinnick, line 80: You can remove "or not". Using "whether" implies that there are 2 options. "... select whether to deploy {Project} SSH keys to the hosts".

guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -27,52 +27,64 @@ endif::[]
...
 +
 If set to `Yes`, public SSH keys will be installed on the registered host.
 The inherited value is based on the `host_registration_remote_execution` parameter.
 It can be inherited, for example from a host group, an operating system, or an organization.
 When overridden, the selected value will be stored on host parameter level.
```
apinnick, line 85: "will be stored" > "is stored", since this is also a case where you can use the present tense.

asteflova:
@apinnick Thanks for the feedback! I must say, though, that the purpose of this PR is not to review these procedures, but to update them to address a specific aspect of product behavior. It's also still a WIP (the PR is in draft state, not ready for a review yet).

I'll keep your notes in mind and fix them if I'm able. But my priority will be to get the original request resolved as soon as possible.

asteflova:
Hi @jeremylenz I'm picking up work on host group procedures again.
* [This Katello PR](https://github.com/Katello/katello/pull/10841) suggests that CV and LCE settings now behave the same way as any other host settings when changing the host group of a host. My testing seems to confirm that, can you confirm as well? (I’m checking because it either contradicts https://bugzilla.redhat.com/show_bug.cgi?id=2253496#c13, or I’m misunderstanding something.)
* As of now, I don’t see anything in these procedures that would need to be changed. Do you? [4.2. Creating a host group](https://theforeman-foreman-documentation-preview-pr-2689.surge.sh/nightly/Managing_Hosts/index-satellite.html#Creating_a_Host_Group_managing-hosts), [4.4. Adding a host to a host group](https://theforeman-foreman-documentation-preview-pr-2689.surge.sh/nightly/Managing_Hosts/index-satellite.html#adding-a-host-to-a-host-group_managing-hosts), and [4.5. Changing the host group of a host](https://theforeman-foreman-documentation-preview-pr-2689.surge.sh/nightly/Managing_Hosts/index-satellite.html#Changing_the_Host_Group_of_a_Host_managing-hosts).

jeremylenz:
> * [This Katello PR](https://github.com/Katello/katello/pull/10841) suggests that CV and LCE settings now behave the same way as any other host settings when changing the host group of a host. My testing seems to confirm that, can you confirm as well? (I’m checking because it either contradicts https://bugzilla.redhat.com/show_bug.cgi?id=2253496#c13, or I’m misunderstanding something.)

Yes, this is the case. When a host's hostgroup changes, we do call `apply_inherited_attributes`. This means that any attributes that the host doesn't have a value for, receive their values from the (new) hostgroup. However, since the host already has a content view and lifecycle environment when the hostgroup changes, a hostgroup change won't change the content view or lifecycle environment of the host, even if the new hostgroup would normally assign a different one.

Previously (before the PR you mentioned) Katello was overriding the `apply_inherited_attributes` method incorrectly, and the effect was that they may be inherited when they shouldn't have been. But now, that's no longer the case.

cc @qcjames53

jeremylenz:
> * As of now, I don’t see anything in these procedures that would need to be changed. Do you? [4.2. Creating a host group](https://theforeman-foreman-documentation-preview-pr-2689.surge.sh/nightly/Managing_Hosts/index-satellite.html#Creating_a_Host_Group_managing-hosts), [4.4. Adding a host to a host group](https://theforeman-foreman-documentation-preview-pr-2689.surge.sh/nightly/Managing_Hosts/index-satellite.html#adding-a-host-to-a-host-group_managing-hosts), and [4.5. Changing the host group of a host](https://theforeman-foreman-documentation-preview-pr-2689.surge.sh/nightly/Managing_Hosts/index-satellite.html#Changing_the_Host_Group_of_a_Host_managing-hosts).

At the end of the "Adding" and "Changing" sections, you have

`The host settings defined in the host group are now applied to the host.`
`The host settings now show the settings defined by the host group.`

For both of these, it may be worth mentioning that attributes for which the host already has values are not changed, even if the new hostgroup has different values. Values will be inherited from the new hostgroup _only_ for attributes the host does not have an existing value for.

Also, my understanding is that Puppet classes are an exception to this rule, and do change things about the host when the hostgroup is changed. But I'm not a Puppet expert so I'd confirm this. :)

jeremylenz:
cc also @ShimShtein @ares in case you have anything to add

asteflova:
Thanks for the info, Jeremy.

Since the project behavior has changed as described above, and the points made also above have actually been handled as part of https://github.com/theforeman/foreman-documentation/pull/2685, I think it's safe to just close this PR. If anyone disagrees, please reopen.

PR #2697, opened 2024-01-25
Add message about cloning templates

AkshayGadhaveRH:
Custom templates created from default templates need to be cloned again after upgrade to ensure that any changes in the default templates apply to custom template as well. BZ link : https://bugzilla.redhat.com/show_bug.cgi?id=2257737

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/con_performing-post-upgrade-tasks.adoc
+++ b/guides/common/modules/con_performing-post-upgrade-tasks.adoc
@@ -4,3 +4,7 @@
 Some of the procedures in this section are optional.
 You can choose to perform only those procedures that are relevant to your installation.
 
+ifdef::Satellite[]
+If you use custom templates created from a cloned default template, create the custom template again after the upgrade.
+Sometimes the default templates are changed during the upgrade, this will ensure that the custom templates are compatible with the latest version.
+endif::[]
--- a/guides/common/modules/con_performing-post-upgrade-tasks.adoc
+++ b/guides/common/modules/con_performing-post-upgrade-tasks.adoc
@@ -4,7 +4,6 @@
 Some of the procedures in this section are optional.
 You can choose to perform only those procedures that are relevant to your installation.
 
-ifdef::Satellite[]
-If you use custom templates created from a cloned default template, create the custom template again after the upgrade.
-Sometimes the default templates are changed during the upgrade, this will ensure that the custom templates are compatible with the latest version.
-endif::[]
+If you use custom templates created from a cloned default template, you must re-create the custom template after the upgrade.
+Sometimes the default templates are changed during the upgrade.
+Re-create the custom templates so that they are compatible with the latest version.


Lennonka:
@AkshayGadhaveRH 
1. The BZ is reported for Sat 6.14 (Foreman 3.7), so should this PR be filed against _master_ and cherry-picked to all versions?
2. What was wrong with the original message suggested in the BZ that you decided not to use it?

guides/common/modules/con_performing-post-upgrade-tasks.adoc
```diff
@@ -4,3 +4,7 @@
 Some of the procedures in this section are optional.
 You can choose to perform only those procedures that are relevant to your installation.
 
+ifdef::Satellite[]
+If you use custom templates created from a cloned default template, create the custom template again after the upgrade.
```
apinnick, line 8: Is re-creating custom templates a requirement? If so, perhaps state it like this: "If you use a custom template cloned from a default template, you must re-create the custom template after the upgrade."

guides/common/modules/con_performing-post-upgrade-tasks.adoc
```diff
@@ -4,3 +4,7 @@
...
+Sometimes the default templates are changed during the upgrade, this will ensure that the custom templates are compatible with the latest version.
```
apinnick, line 9: This sentence has a comma splice. Either break the sentence into 2 sentences or replace the comma with a semi-colon.

ISG says to avoid the future tense and ambiguous pronouns. Suggestion: "Sometimes the default templates are changed during the upgrade. Re-creating custom templates ensures that your templates are compatible with the latest version."

guides/common/modules/con_performing-post-upgrade-tasks.adoc
```diff
@@ -4,3 +4,7 @@
 Some of the procedures in this section are optional.
 You can choose to perform only those procedures that are relevant to your installation.
 
+ifdef::Satellite[]
```
maximiliankolb, line 7:
```suggestion
ifdef::satellite,orcharhino[]
```

maximiliankolb (commented):
Maybe we should also reference the "custom pre/post" snippets that are included if present? Depending on the use case, this is a more convenient method to "customize" templates without actually unlocking them/cloning locked templates.

Examples: See https://github.com/theforeman/foreman/blob/3.9-stable/app/views/unattended/provisioning_templates/provision/kickstart_default.erb#L272 & https://github.com/theforeman/foreman/blob/3.9-stable/app/views/unattended/provisioning_templates/provision/kickstart_default.erb#L360

cc @sbernhard

AkshayGadhaveRH:
Hey @Lennonka the Bz is siupposed to be for all versions. Since the file structures for lower versions differs for lower versions, I pushed a PR for 3.1 separately.
The original text used words like "advisable" and was in passive voice, which is why I thought I'll reword it.

guides/common/modules/con_performing-post-upgrade-tasks.adoc
```diff
@@ -4,3 +4,7 @@
 Some of the procedures in this section are optional.
 You can choose to perform only those procedures that are relevant to your installation.
 
+ifdef::satellite,orcharhino[]
```
evgeni, line 7: I wouldn't limit this, it's applicable to all deployments.

evgeni:
I have no idea about the exact wording, but the overall statement is correct.

On upgrade (shipped) templates *can* (but don't have to) change, either because we improved them or because some underlying change required a change in the template too. The former doesn't require changing custom templates, while the later does. We have no way to adjust custom templates (regardless whether they were created as a clone or fully self-written!), so users are advised to check that themself.

Sadly, this advice is not actionable, as we do not offer a good way to see the changes (and their reason) :disappointed:

AkshayGadhaveRH:
https://github.com/theforeman/foreman-documentation/pull/2780 created for Master.

PR #2706, opened 2024-01-29
Use guide link instead of xref [fix]

Lennonka:
xref is not working, the target is in another guide.

### DIFF
--- a/guides/common/modules/ref_supported-operating-systems.adoc
+++ b/guides/common/modules/ref_supported-operating-systems.adoc
@@ -17,7 +17,7 @@ The following operating systems are supported by the installer, have packages, a
 |====
 | Operating System | Architecture | Notes
 ifdef::foreman-el,katello,orcharhino[]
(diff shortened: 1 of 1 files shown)


Lennonka:
However, the Planning guide is not displayed for some flavors at the moment, so the target will not show. Perhaps we should at least prioritize finishing at least the Glossary. cc @ekohl

maximiliankolb:
https://docs.theforeman.org/nightly/Planning_for_Project/index-foreman-el.html#varl-Glossary_of_Terms-Enterprise_Linux in HTML diff does not work.

ekohl (approved):
Perhaps we should hide all sections of the planning guide and just show the glossary on platforms where it's not ready?

I don't mind merging this now and doing the planning part later.

Lennonka:
> Perhaps we should hide all sections of the planning guide and just show the glossary on platforms where it's not ready?
> 
> I don't mind merging this now and doing the planning part later.

@ekohl I'm all for this! I'll do that in another PR.

PR #2684, opened 2024-01-19
Use attributes depending on Host OS

maximiliankolb:
Follow-up task to splitting Foreman/SmartProxy and Host attributes. Please review commit by commit.

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.9/Katello 4.11 (planned Satellite 6.15)

PR #2709, opened 2024-01-30
Clarify target of securing DHCP API

maximiliankolb:
This procedure is only part of one assembly: "assembly_managing-dhcp-on-smart-proxies.adoc", which itself is only part of "doc-Installing_Proxy".

follow-up PR based on https://github.com/theforeman/foreman-documentation/pull/2684#discussion_r1469487245

### DIFF
--- a/guides/common/modules/proc_securing-the-dhcp-api.adoc
+++ b/guides/common/modules/proc_securing-the-dhcp-api.adoc
@@ -1,12 +1,12 @@
 [id="Securing_the_dhcpd_API_{context}"]
 = Securing the dhcpd API
 
-{SmartProxy} interacts with DHCP daemon using the dhcpd API to manage DHCP.
+{SmartProxyServer} interacts with DHCP daemon using the dhcpd API to manage DHCP.
 By default, the dhcpd API listens to any host without access control.
 You can add an `omapi_key` to provide basic security.
 
 .Procedure
-. On your {SmartProxy}, install the required packages:
+. On your {SmartProxyServer}, install the required packages:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----


guides/common/modules/proc_securing-the-dhcp-api.adoc
```diff
@@ -1,12 +1,12 @@
 [id="Securing_the_dhcpd_API_{context}"]
 = Securing the dhcpd API
 
-{SmartProxy} interacts with DHCP daemon using the dhcpd API to manage DHCP.
+{SmartProxyServer} interacts with DHCP daemon using the dhcpd API to manage DHCP.
```
ekohl, line 4: This isn't true. It's the Smart Proxy process that communicates with the DHCP daemon. IMHO is SmartProxyServer the whole server.

guides/common/modules/proc_securing-the-dhcp-api.adoc
```diff
@@ -1,12 +1,12 @@
...
 By default, the dhcpd API listens to any host without access control.
 You can add an `omapi_key` to provide basic security.
 
 .Procedure
-. On your {SmartProxy}, install the required packages:
+. On your {SmartProxyServer}, install the required packages:
```
ekohl, line 9: This can also be done on your ProjectServer if you run DHCP there.

ekohl (changes requested):

maximiliankolb:
I have checked; we do not include this procedure in "Installing Foreman Server". That's why I went with the "SmartProxyServer" attribute. See https://docs.theforeman.org/nightly/Installing_Proxy/index-katello.html#Managing_DHCP_Using_Smart_Proxy_smart-proxy

Should I add this to "Installing Foreman/Katello Server"?

ekohl:
I started to automate this in https://github.com/theforeman/puppet-foreman_proxy/pull/827 but need to spend a bit more time finalizing it. Time I don't have right now.

> have checked; we do not include this procedure in "Installing Foreman Server"

I've talked to @Lennonka about this before, but I think we need a DNS integration guide about how to set up the various DNS integrations. That would all apply to both targets we have. And the exact same thing with DHCP integration.

I'll discuss it with @apinnick as well, because it's one of those overall content strategy things.

Short term I'm not sure what's best.

PR #2710, opened 2024-01-30
Make content type depend on Host OS

maximiliankolb:

### DIFF
--- a/guides/common/modules/con_restricting-hosts-access-to-content.adoc
+++ b/guides/common/modules/con_restricting-hosts-access-to-content.adoc
@@ -57,7 +57,7 @@ Release version::
 Certain Red{nbsp}Hat repositories, such as the {RHEL} dot release repositories, include a *Release version* in their repository metadata.
 The release version is then compared with the release version specified in the *System purpose* properties of the host.
 Access to content may be limited or restricted based on this comparison.
-For more information about setting system purpose attributes,see {ManagingHostsDocURL}Creating_a_Host_managing-hosts[Creating a Host in {ProjectName}] in _{ManagingHostsDocTitle}_.
+For more information about setting system purpose attributes, see {ManagingHostsDocURL}Creating_a_Host_managing-hosts[Creating a Host in {ProjectName}] in _{ManagingHostsDocTitle}_.
 endif::[]
 ifdef::orcharhino[]
 ifdef::red_hat_enterprise_linux[]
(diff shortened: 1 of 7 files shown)


guides/common/modules/con_restricting-hosts-access-to-content.adoc
```diff
@@ -30,18 +40,36 @@ You can use composite Content Views to combine and give hosts access to the cont
...
+ifdef::orcharhino[]
+ifdef::red_hat_enterprise_linux[]
+Red{nbsp}Hat repositories set architecture and OS version restrictions automatically.
+endif::[]
+endif::[]
+ifndef::orcharhino[]
 Red{nbsp}Hat repositories set architecture and OS version restrictions automatically.
+endif::[]
 
+ifndef::orcharhino[]
```
adamlazik1, line 57:
```suggestion

```

guides/common/modules/con_restricting-hosts-access-to-content.adoc
```diff
@@ -14,7 +14,17 @@ For more information about lifecycle environments, see xref:Managing_Application
 
 Content overrides::
 By default, content hosted by {Project} can be either enabled or disabled.
+ifdef::orcharhino[]
+ifdef::red_hat_enterprise_linux[]
```
adamlazik1, line 18: I am guessing this is an attribute that you define downstream. If it is possible to rename that attribute, you can simply rename it to orcharhino_red_hat_enterprise_linux, and then it is sufficient to just check for that attribute instead of checking for orcharhino as well.

adamlazik1 (approved):
LGTM, left some suggestions

guides/common/modules/con_restricting-hosts-access-to-content.adoc
maximiliankolb: yeah, unfortunately I cannot rename this easily. That's why I went with the double "ifdef" directives. I did not experimenting with "and" and "or" to combine ifdefs, but this would not work for all cases in a clean way in my local tests.

guides/common/modules/con_restricting-hosts-access-to-content.adoc
maximiliankolb: Applied with an empty line in between.

maximiliankolb (commented):
Rebased to master and applied all feedback.

Issue #2467, opened 2023-09-26
Replace apt-get with apt in Foreman-deb


mjivraja:
`apt` is a subset of `apt-get` and `apt-cache` that provides necessary commands for package management. Moreover, it is an enhanced approach to apt-get that eliminates the use of further options/arguments. So with apt, you get all the necessary tools in one place.

In my case, if I want to upgrade the Linux kernel with simply `apt-get upgrade` I can't. But with `apt upgrade` I can. Hence, I believe that as `apt` is the advanced and recommended approach for Debian, we can replace the package manager apt-get for `foreman-deb` with apt. I am not sure, if `apt-get upgrade` can upgrade or update the Foreman-deb without any extra options/arguments.

Note: apt-get is not deprecated at all and there is no news on it.

ekohl:
> In my case, if I want to upgrade the Linux kernel with simply `apt-get upgrade` I can't.

Could you elaborate on this? It's been a while since I used `apt-get`, but I thought there was no difference compared to `apt` here.

mjivraja:
In `apt-get --simulate upgrade`, I saw that the packages of the Linux kernel were kept back. I need to go through `apt-get dist-upgrade` path to update those.

ekohl:
Ah right. https://itsfoss.com/apt-get-upgrade-vs-dist-upgrade/ does clarify it well. :+1: for using `apt`.

mjivraja:
Closing this issue as PR for the same has been considered.

PR #2719, opened 2024-01-31
Replace apt-get with apt for Debian

mjivraja:
Debian introduced `apt` a long time ago. Its older package manager `apt-get` is still working but it does not manage packages as effectively as apt does. For example, `apt-get` requires more arguments to update system kernels but `apt` does not. I am not sure how it affects the package management in `Foreman-deb` but at present end-users are more inclined and familiar with `apt`. Hence, we should replace the older package manager even in our documentation.

Issue: https://github.com/theforeman/foreman-documentation/issues/2467

* [X] Foreman 3.7/Katello 4.9 (Satellite 6.14)
* [X] Foreman 3.6/Katello 4.8
* [X] Foreman 3.5/Katello 4.7 (Satellite 6.13; orcharhino 6.6)


mjivraja:
@maximiliankolb I found some places (visible here) where we can replace package manager with `project-package-*` and `client-package-*`.

guides/common/modules/proc_configuring-a-host-to-use-the-pull-client.adoc
```diff
@@ -30,7 +30,7 @@ ifdef::client-content-apt[]
 * On {Debian} hosts:
 +
 ----
-# apt-get install katello-pull-transport-migrate
+# apt install katello-pull-transport-migrate
```
maximiliankolb, line 33:
```suggestion
# {client-package-install-deb} install katello-pull-transport-migrate
```

Please also add `[options="nowrap", subs="+quotes,verbatim,attributes"]` above the code block.

guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
```diff
@@ -46,7 +46,7 @@ ifndef::satellite[]
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-# apt-get install puppet-agent
+# apt install puppet-agent
```
maximiliankolb, line 49:
```suggestion
# {client-package-install-deb} install puppet-agent
```

guides/common/modules/ref_custom-provisioning-snippet-example-for-debian.adoc
```diff
@@ -8,7 +8,7 @@ You can use `Custom Post` snippets to call external APIs from within the provisi
 ----
 echo "Calling API to report successful host deployment"
 
-apt-get install -y curl ca-certificates
+apt install -y curl ca-certificates
```
maximiliankolb, line 11:
```suggestion
{client-package-install-deb} install -y curl ca-certificates
```

as above: you have to allow attributes in the code block; you can just overwrite line 7; see comment above.

maximiliankolb (commented):
I am leaning against changing "apt-get" to "apt". `apt` still returns "WARNING: apt does not have a stable CLI interface. Use with caution in scripts." on various Debian/Ubuntu versions.

I believe there is a reason provisioning templates still use `apt-get`. However, I am OK either way because I can handle this in downstream by overwriting "client-package-*-deb".

Maybe @evgeni can weigh in?

evgeni:
`apt` is not intended for script use (today), that's true, but I'd argue docs aren't scripts (contrary to e.g. provisioning templates), and thus it's fine to use `apt` here (and it does indeed produce better results, just not identical ones to before)

maximiliankolb (approved):
LGTM

maximiliankolb:
Merged to master and cherry-picked to 3.9: :heavy_check_mark: 
   16758fca4..fe9f29454  3.9 -> 3.9

I did not cherry-pick to 3.8 or below because we have reworked the attributes but did not cherry-pick this any further. If you want, you could open a separate PR or just leave it as is which is perfectly fine for me! @mjivraja

mjivraja:
> Merged to master and cherry-picked to 3.9: ✔️ [16758fc](https://github.com/theforeman/foreman-documentation/commit/16758fca42bbcc411d939df4b578815af867b94c)..[fe9f294](https://github.com/theforeman/foreman-documentation/commit/fe9f29454d22579770092acd34a0fdd8005037f4) 3.9 -> 3.9
> 
> I did not cherry-pick to 3.8 or below because we have reworked the attributes but did not cherry-pick this any further. If you want, you could open a separate PR or just leave it as is which is perfectly fine for me! @mjivraja

Ah yes. I am OK to keep this up to `3.9` :+1:

PR #2737, opened 2024-02-01
Fix the GSS-Proxy AD integration procedure

adamlazik1:

* [x] Foreman 3.6/Katello 4.8
* [x] Foreman 3.5/Katello 4.7 (Satellite 6.13; orcharhino 6.6)

### DIFF
--- a/guides/common/modules/proc_configuring-direct-ad-integration-with-gss-proxy.adoc
+++ b/guides/common/modules/proc_configuring-direct-ad-integration-with-gss-proxy.adoc
@@ -47,17 +47,49 @@ Apache user must not have access to the keytab file.
 ----
 [service/HTTP]
 mechs = krb5
-cred_store = keytab:/etc/httpd/conf/http.keytab
+cred_store = keytab:/etc/gssproxy/http.keytab
 cred_store = ccache:/var/lib/gssproxy/clients/krb5cc_%U
 euid = __ID_of_Apache_User__
+----
+. Set the correct SELinux context for the future keytab entry:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# semanage fcontext -a -s system_u -t krb5_keytab_t '/etc/gssproxy/http\.keytab'
 ----
 . Create a keytab entry:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-# KRB5_KTNAME=FILE:/etc/httpd/conf/http.keytab net ads keytab add HTTP -U administrator -d3 -s /etc/net-keytab.conf
-# chown root.{apache-user} /etc/httpd/conf/http.keytab
-# chmod 640 /etc/httpd/conf/http.keytab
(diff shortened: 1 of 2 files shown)


adamlazik1:
According to @alsouza, setting the SPN might not be necessary at all. To be confirmed by retesting next week.

adamruzicka:
> To be confirmed by retesting next week.

Did this happen?

adamlazik1:
Not yet, but hopefully on the way. I'll ask if there has been any progress.

guides/common/modules/proc_configuring-direct-ad-integration-with-gss-proxy.adoc
```diff
@@ -47,17 +47,43 @@ Apache user must not have access to the keytab file.
...
+----
+// TO BE CONFIRMED: This step should not be needed
+. Add a *Service Principal Name (SPN)* for your AD server.
+You can use a tool such as `adcli`:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# adcli spn --add _My_Service_/{foreman-example-com} _My_Username_
+----
+. Make sure the following line is included in the `/etc/sssd/sssd.conf` file:
```
asteflova, line 76:
```suggestion
. Configure SSSD to map Group Policy Objects (GPOs) from AD to the `foreman` PAM service by ensuring the following line is included in the `/etc/sssd/sssd.conf` file:
```

I don't know if this wording is technically correct, but my point is to suggest to describe what adding the line does.

guides/common/modules/proc_configuring-direct-ad-integration-with-gss-proxy.adoc
```diff
@@ -47,17 +47,43 @@ Apache user must not have access to the keytab file.
...
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+ad_gpo_map_service = +foreman
+----
+. If you changed the `/etc/sssd/sssd.conf` configuration file, restart the sssd service:
```
asteflova, line 82:
```suggestion
. If you changed the `/etc/sssd/sssd.conf` configuration file, restart the `sssd` service:
```
It's either "the SSSD service" or "the `sssd` service". And other steps reference services in the format of "the `httpd` service" etc.

guides/common/modules/proc_configuring-direct-ad-integration-with-gss-proxy.adoc
asteflova: Also, the IMPORTANT box right under this procedure also tells users to add this config option to sssd.conf. That could be confusing. Do you have any ideas for how to deal with it? I think you could take the entire IMPORTANT box and merge it with this step... would that make sense?

ekohl:
Typo in the commit message: GGS -> GSS

asteflova:
@adamlazik1 and I agreed that I would take over his effort to update and verify the procedure. I filed https://github.com/theforeman/foreman-documentation/pull/3056 for that purpose.

PR #2746, opened 2024-02-06
Add capitalization to contributing guidelines

### LINKED ISSUE #2744: List of capitalized nouns for Contributing guidelines
Add a "Capitalization" section to [Contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) so that new contributors know what needs to be capitalized. To make it easy to maintain, I recommend creating a single list in alphabetical order. If it gets too long, we can consider using generic terms like "Software packages" if we end up with too many examples to list.

Suggestions:
Ansible, Salt, Chef, Python, Kickstart, Discovery (service), Tracer (utility), Insights (RH product), OpenScap, Hammer, Candlepin, Pulp, Redis, Puppet

maximiliankolb:
closes #2744

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,6 +42,8 @@ See [LICENSE](LICENSE).
 * [ ] If I make more than one change on my branch, I create more than one commit and rebase my branch to master.
 * [ ] My PR does not solely rely on internal resources to answer the _why_, but instead contains at least a basic description for the community.
 * [ ] When creating my PR, I select all branches I want my change to get cherry-picked to.
+* [ ] I am familiar to proper capitalization for project-specific terminology.
+See [Capitalization](#Capitalization).
 
 ## Crafting Great Pull Requests
 
@@ -53,6 +55,37 @@ Please explain why someone else should merge your proposed change:
 * Are there any regressions or downsides to my PR?
 * Does my PR make multiple changes and therefore require multiple commits?
 Ideally, each individual change correlates to an individual commit.
+
+## Capitalization
+
+All headings are sentence case.
+We capitalize the following terms:
+
+* Ansible
+* Candlepin
+* Chef
+* Content View
+* Discovery (service)
+* Hammer
+* Insights (Red Hat product)
+* Kickstart
+* Library (environment)
+* OpenSCAP
+* PostgreSQL
+* Pulp
+* Puppet
+* Python
+* Red Hat Network
+* Redis
+* Salt
+* Salt Minions
+* Tracer (utility)
+
+We do not capitalize the following terms:
+
+* Red Hat subscription manifest
+* manifest
+* subscription
 
 ## Further Information
 


CONTRIBUTING.md
```diff
@@ -54,6 +56,37 @@ Please explain why someone else should merge your proposed change:
...
+
+* Ansible
+* Candlepin
+* Chef
+* Content View
+* Discovery (service)
+* Hammer
+* Insights (Red Hat product)
+* Kickstart
+* Library
```
adamlazik1, line 72: This is strange at least. Library is very common word. Is this a product name @maximiliankolb?

adamlazik1 (commented):
Looks good for now, we can change capitalization of content view later when a PR to change it is ready.

CONTRIBUTING.md
maximiliankolb: I have no idea. This came up during review of "sentence case in headings" by @apinnick 

I can only assume this comes from Web UI and it means the default "Library" lifecycle environment?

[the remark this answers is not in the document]
apinnick: [reference](https://access.redhat.com/documentation/en-us/red_hat_satellite/6.14/html-single/managing_content/index#Creating_a_Lifecycle_Environment_Path_content-management)

CONTRIBUTING.md
```diff
@@ -54,6 +56,37 @@ Please explain why someone else should merge your proposed change:
...
 
+## Capitalization
+
+All headings are sentence case.
+We capitalize the following terms:
+
+* Ansible
+* Candlepin
+* Chef
+* Content View
```
apinnick, line 67: Interesting that "Content View" is consistently capitalized in the documentation but lower case in the [UI](https://access.redhat.com/documentation/en-us/red_hat_satellite/6.14/html-single/managing_content/index#Creating_a_Composite_Content_View_content-management).

CONTRIBUTING.md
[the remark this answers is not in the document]
maximiliankolb: Applied.

CONTRIBUTING.md
adamlazik1: Yeah, I am fairly certain that it was supposed to be lowercase to begin with. However, since currently all guides use the uppercase version, I propose to keep this here for now and then remove it in the PR changing CV to lowercase. Or should we say now that it shouldn't be uppercase to not confuse writers later? But then there might be confusion as to why is CV currently uppercase everywhere.

CONTRIBUTING.md
```diff
@@ -54,6 +56,37 @@ Please explain why someone else should merge your proposed change:
...
+* Python
+* Red Hat Network
+* Redis
+* Salt
+* Salt Minions
+* Tracer (utility)
+
+We do not capitalize the following terms:
+
+* Red Hat subscription manifests
```
adamlazik1, line 86:
```suggestion
* Red Hat subscription manifest
```
I think lowercasing the plural will be implied.

maximiliankolb: Fixed.

CONTRIBUTING.md
maximiliankolb: https://github.com/theforeman/foreman-documentation/issues/2747

adamlazik1 (approved):

PR #2780, opened 2024-02-12
Add message about cloning templates for master

AkshayGadhaveRH:
Same change as https://github.com/theforeman/foreman-documentation/pull/2697 for master. Custom templates created from default templates need to be cloned again after upgrade to ensure that any changes in the default templates apply to custom template as well. BZ link : https://bugzilla.redhat.com/show_bug.cgi?id=2257737

* [X] Foreman 3.7/Katello 4.9 (Satellite 6.14)
* [X] Foreman 3.6/Katello 4.8
* [X] Foreman 3.5/Katello 4.7 (Satellite 6.13; orcharhino 6.6/6.7)

### DIFF
--- a/guides/common/modules/con_performing-post-upgrade-tasks.adoc
+++ b/guides/common/modules/con_performing-post-upgrade-tasks.adoc
@@ -2,3 +2,5 @@
 = Performing post-upgrade tasks
 
 Optional: If the default provisioning templates have been changed during the upgrade, recreate any templates cloned from the default templates.
+If the custom code is executed before and/or after the provisioning process, use custom provisioning snippets to avoid recreating cloned templates.
+For more information about configuring custom provisioning snippets, see {ProvisioningDocURL}Creating_Custom_Provisioning_Snippets_provisioning[Creating Custom Provisioning Snippets] in _{ProvisioningDocTitle}_.
--- a/guides/common/modules/con_performing-post-upgrade-tasks.adoc
+++ b/guides/common/modules/con_performing-post-upgrade-tasks.adoc
@@ -1,23 +1,4 @@
 [id="performing-post-upgrade-tasks_{context}"]
 = Performing post-upgrade tasks
 
-Some of the procedures in this section are optional.
-You can choose to perform only those procedures that are relevant to your installation.
-
-If you use custom templates created from a cloned default template, create the custom template again after the upgrade.
-Sometimes the default templates are changed during the upgrade, this will ensure that the custom templates are compatible with the latest version.
-
-* xref:upgrading_discovery_parent[]
-ifdef::katello,satellite,orcharhino[]
-* xref:upgrading_virt_who[]
-endif::[]
-ifdef::satellite[]
-* xref:removing_satellite_tools_repository[]
-endif::[]
-ifndef::foreman-deb[]
-* xref:migrating-ansible-content_{context}[]
-endif::[]
-* xref:reclaiming-postgresql-space_{context}[]
-ifndef::foreman-el,foreman-deb[]
-* xref:tuning-with-predefined-profiles_{context}[]
-endif::[]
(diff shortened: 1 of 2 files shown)


maximiliankolb (changes requested):
This file is currently not included in any assembly/"master.adoc" file. 

Second, I still think we should hint users towards using "custom" snippets as outlined in https://github.com/theforeman/foreman-documentation/pull/2697#pullrequestreview-1849001533. This is IMHO helpful to users because they would not have to reapply their changes onto new templates with each upgrade.

guides/common/modules/con_performing-post-upgrade-tasks.adoc
```diff
@@ -4,6 +4,9 @@
 Some of the procedures in this section are optional.
 You can choose to perform only those procedures that are relevant to your installation.
 
+If you use custom templates created from a cloned default template, create the custom template again after the upgrade.
+Sometimes the default templates are changed during the upgrade, this will ensure that the custom templates are compatible with the latest version.
```
mjivraja, line 8:
```suggestion
The default templates are changed during the upgrade, this will ensure that the custom templates are compatible with the latest version.
```
If we have clear data, we should not use this at all. 
If this is less frequent, we can opt for `occasionally` instead of sometimes, according to SG.

bangelic, line 8:
```suggestion
The default templates are changed during the upgrade. 
This will ensure that the custom templates are compatible with the latest version.
```
Following on from Malhar's suggestion, I think this should be two sentences.

AkshayGadhaveRH:
@maximiliankolb I agree it would be less messier to use custom templates. Should we add this somewhere in the provisioning guide? I was thinking of adding the same warning about having to recreate the templates under the section that describes cloning templates, maybe that could be a good place to add this?

AkshayGadhaveRH:
@mjivraja @bangelic Rephrased. Does this look better?

maximiliankolb:
> @maximiliankolb I agree it would be less messier to use custom templates. Should we add this somewhere in the provisioning guide? I was thinking of adding the same warning about having to recreate the templates under the section that describes cloning templates, maybe that could be a good place to add this?

@AkshayGadhaveRH I am not talking about custom templates, but custom snippets that follow a specific naming convention. For example:

````
app/views/unattended/provisioning_templates/finish/kickstart_default_finish.erb
35:<%= snippet_if_exists(template_name + " custom pre") -%>
83:<%= snippet_if_exists(template_name + " custom post") -%>
````

This means that if a user creates a snippet called "Kickstart default finish custom pre", then it's automatically included in the default Kickstart finish template. Foreman/Satellite/orcharhino does not deliver those custom snippets AFAIK. This means, a user will never run into a conflict or miss updates. Instead, users can use the default templates as delivered by Foreman/Satellite/orcharhino _and_ still include their own custom sections.

You can grep

AkshayGadhaveRH:
@maximiliankolb oh my bad, I thought you were referring to the custom templates. But, there are users who edit the custom templates to include changes that don't necessarily take place before and/or after provisioning. This warning might be useful for them IMO.

maximiliankolb:
I think both is correct: a) tell users they need to "rebase" their templates if the clone them and upgrade to a newer version; and b) add a hint about possible "custom" snippets that are included in various templates automatically if present.

maximiliankolb (approved):
Yes, makes sense to me! Any other input/ACK?

adamlazik1 (approved):

adamlazik1:
Merged and cherry-picked:
   8ed2178e0e..f84b27533b  3.7 -> 3.7
   bb08d2aab6..a86277a422  3.8 -> 3.8
   2ccf4abed6..0336b4a9dd  3.9 -> 3.9

There were merge conflicts on lower branches. Please create separate PR.

PR #2787, opened 2024-02-13
Use attributes in docinfo.xml (Sat)

Lennonka:
This way, we only have to update the version in a single place - `attributes-satellite.adoc`
Yes, they work in downstream build. I can provide PoC upon request.


ekohl (approved):
This makes so many things better

adamlazik1 (approved):
LGTM, thanks a lot!

PR #2791, opened 2024-02-15
Add inclusive language module for Satellite

apinnick:
Added the downstream "inclusive language" module for Satellite docs only.

### DIFF
--- a/guides/common/modules/con_making-open-source-more-inclusive.adoc
+++ b/guides/common/modules/con_making-open-source-more-inclusive.adoc
@@ -1,0 +1,4 @@
+[id="making-open-source-more-inclusive_{context}"]
+= Making open source more inclusive
+
+Red Hat is committed to replacing problematic language in our code, documentation, and web properties. Because of the enormity of this endeavor, these changes are being updated gradually and where possible. For more details, see link:https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language[our CTO Chris Wright's message].
--- a/guides/common/modules/con_making-open-source-more-inclusive.adoc
+++ b/guides/common/modules/con_making-open-source-more-inclusive.adoc
@@ -1,4 +1,6 @@
 [id="making-open-source-more-inclusive_{context}"]
 = Making open source more inclusive
 
-Red Hat is committed to replacing problematic language in our code, documentation, and web properties. Because of the enormity of this endeavor, these changes are being updated gradually and where possible. For more details, see link:https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language[our CTO Chris Wright's message].
+Red Hat is committed to replacing problematic language in our code, documentation, and web properties.
+Because of the enormity of this endeavor, these changes are being updated gradually and where possible.
+For more details, see link:https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language[our CTO Chris Wright's message].
--- a/guides/doc-Administering_Project/master.adoc
+++ b/guides/doc-Administering_Project/master.adoc
@@ -11,6 +11,8 @@ endif::[]
 ifndef::HideDocumentOnStable[]
 
 ifdef::satellite[]
+include::common/modules/con_making-open-source-more-inclusive.adoc[]
+
 include::common/modules/proc_providing-feedback-on-red-hat-documentation.adoc[leveloffset=+1]
 endif::[]
 
--- a/guides/doc-Administering_Project/master.adoc
+++ b/guides/doc-Administering_Project/master.adoc
@@ -11,7 +11,7 @@ endif::[]
 ifndef::HideDocumentOnStable[]
 
 ifdef::satellite[]
-include::common/modules/con_making-open-source-more-inclusive.adoc[]
+include::common/modules/con_making-open-source-more-inclusive.adoc[leveloffset=+1]
 
 include::common/modules/proc_providing-feedback-on-red-hat-documentation.adoc[leveloffset=+1]
 endif::[]
--- a/guides/doc-Administering_with_Ansible/master.adoc
+++ b/guides/doc-Administering_with_Ansible/master.adoc
@@ -12,6 +12,8 @@ endif::[]
 ifndef::HideDocumentOnStable[]
 
 ifdef::satellite[]
+include::common/modules/con_making-open-source-more-inclusive.adoc[]
+
 include::common/modules/proc_providing-feedback-on-red-hat-documentation.adoc[leveloffset=+1]
 endif::[]
 
--- a/guides/doc-Administering_with_Ansible/master.adoc
+++ b/guides/doc-Administering_with_Ansible/master.adoc
@@ -12,7 +12,7 @@ endif::[]
 ifndef::HideDocumentOnStable[]
 
 ifdef::satellite[]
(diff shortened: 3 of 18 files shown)


guides/common/modules/con_making-open-source-more-inclusive.adoc
```diff
@@ -0,0 +1,4 @@
+[id="making-open-source-more-inclusive_{context}"]
+= Making open source more inclusive
+
+Red Hat is committed to replacing problematic language in our code, documentation, and web properties. Because of the enormity of this endeavor, these changes are being updated gradually and where possible. For more details, see link:https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language[our CTO Chris Wright's message].
```
maximiliankolb, line 4:
```suggestion
Red Hat is committed to replacing problematic language in our code, documentation, and web properties.
Because of the enormity of this endeavor, these changes are being updated gradually and where possible.
For more details, see https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language[our CTO Chris Wright's message].
```

maximiliankolb: We could also replace "Red Hat" with team, make the third line Satellite-only, and include this statement to all build targets.

ekohl: I'd be against including it in the Foreman project. We had a discussion in https://community.theforeman.org/t/rfc-replacing-blacklist-with-blocklist-in-our-templates/19276 and https://community.theforeman.org/t/terminology-updates-in-foreman/19477 in the past. Tomer made a few small changes, but the biggest effect is that a very valued contributor left the project. IMHO it's only virtue signaling. Adding this would only distract from the real documentation.

maximiliankolb: I (think I) know who left and thought about them when reading this too. But still, I strongly support making an effort to use inclusive language in the docs. I help maintain the docs and review/support this PR to reword ~blacklist~ to _blocklist_ etc.

I am OK either way, but it basically leads to the question _who_ can speak for the Foreman Community in terms of "who decides if we can write 'Foreman supports inclusive language'". It's not like every other (problematic) change is discussed before implementing it.

I am OK with creating a small RFC or keeping it as is.

ekohl: I think there's a difference in including the header and making changes. I'm firmly against including the header in upstream. The other individual changes can be judged separate.

guides/doc-Administering_Project/master.adoc
```diff
@@ -11,6 +11,8 @@ endif::[]
 ifndef::HideDocumentOnStable[]
 
 ifdef::satellite[]
+include::common/modules/con_making-open-source-more-inclusive.adoc[]
```
maximiliankolb, line 14: I also wonder if this needs a level offset like line 16. Your call.

maximiliankolb (commented):
Please convert to "one sentence per line". the rest should not block your PR.

guides/common/modules/con_making-open-source-more-inclusive.adoc
maximiliankolb: OK, then let's keep it as suggestion by @apinnick

apinnick: @maximiliankolb 
I do not recommend adding an inclusive language module while your default branch is still "master". :-)
That is why I marked this module for Satellite only.

Re: Not including it in the Foreman repo at all. I could write a script to inject this module into downstream adoc files. I added it up stream when I saw that the "Feedback" module is included.

Will fix the lines.

guides/doc-Administering_Project/master.adoc
apinnick: Yes, it works better with a level offset. We used to use no level offset and Asciidoctor generated a preface. Now, it generates a "Part 1", which is not the desired outcome. Will fix.

guides/common/modules/con_making-open-source-more-inclusive.adoc
```diff
@@ -0,0 +1,6 @@
+[id="making-open-source-more-inclusive_{context}"]
+= Making open source more inclusive
+
+Red Hat is committed to replacing problematic language in our code, documentation, and web properties.
+Because of the enormity of this endeavor, these changes are being updated gradually and where possible.
+For more details, see link:https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language[our CTO Chris Wright's message].
```
maximiliankolb, line 6:
```suggestion
For more details, see https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language[our CTO Chris Wright's message].
```

maximiliankolb (changes requested):

guides/common/modules/con_making-open-source-more-inclusive.adoc
ekohl: I meant to include it in the rendered upstream docs, not repo

apinnick: @ekohl Has anyone ever proposed renaming the default branches in Foreman and Foreman docs? Most of our  projects, like [openshift-docs](https://github.com/openshift/openshift-docs), use `main` instead of `master`.

guides/common/modules/con_making-open-source-more-inclusive.adoc
apinnick: I was going to ask about the link formatting at some point. The [RH Asciidoc quick ref](https://redhat-documentation.github.io/asciidoc-markup-conventions/), [RHSSG](https://redhat-documentation.github.io/supplementary-style-guide/#external-links), and [openshift-doc guidelines](https://github.com/openshift/openshift-docs/blob/main/contributing_to_docs/doc_guidelines.adoc#links-to-external-websites) require the link macro instead of the URL macro, even for ordinary URLs. I'm not sure why. We do not have many style guides that cover Asciidoc formatting. 

Is there a good reason not to use the link macro?

guides/common/modules/con_making-open-source-more-inclusive.adoc
ekohl: @apinnick foreman.git has used `develop` for the past 10+ years but I don't think anyone wants to put in the effort to rename other branches. It's utterly painful and I don't see a good reason to.

apinnick: @ekohl  
> I meant to include it in the rendered upstream docs, not  repo

I thought this comment meant that you were considering including the new module in rendered upstream docs. 
If that is the case, I was pointing out that it would look odd because the upstream default branch is 'master'. I agree that it is not a trivial matter to rename branches.

guides/common/modules/con_making-open-source-more-inclusive.adoc
apinnick: @maximiliankolb 
Is there a reason not to use the link macro?

maximiliankolb (approved):
LGTM

maximiliankolb:
Merged to "master" and cherry-picked to "3.9": :heavy_check_mark: 
   0642a5c357..b5adc37a9a  3.9 -> 3.9

PR #2801, opened 2024-02-20
Keep backup of orig cfg files

Lennonka:
User: "Is it a good idea to wipe the cfg files out to add/change these options?"
SME: "Not unless you have a backup."
https://bugzilla.redhat.com/show_bug.cgi?id=2239040

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_using-vmware-cloud-init-and-userdata-templates-for-provisioning.adoc
+++ b/guides/common/modules/proc_using-vmware-cloud-init-and-userdata-templates-for-provisioning.adoc
@@ -33,6 +33,10 @@ ifndef::orcharhino[]
 For more information, see {InstallingServerDocURL}Port_and_firewall_requirements_{project-context}[Port and firewall requirements] in _{InstallingServerDocTitle}_ and {InstallingSmartProxyDocURL}{smart-proxy-context}-port-and-firewall-requirements_{smart-proxy-context}[Port and firewall requirements] in _{InstallingSmartProxyDocTitle}_.
 endif::[]
 include::snip_prerequisite-configured-smart-proxy-registration-provisioning.adoc[]
+* Back up the following configuration files:
+** `/etc/cloud/cloud.cfg.d/01_network.cfg`
+** `/etc/cloud/cloud.cfg.d/10_datasource.cfg`
+** `/etc/cloud/cloud.cfg`
 
 .Associating the `Userdata` and `Cloud-init` templates with the operating system
 . In the {ProjectWebUI}, navigate to *Hosts* > *Templates* > *Provisioning Templates*.


Lennonka:
Or should I add a new step to backup all the cfg files before the steps that wipe them out?

ekohl (commented):
I want to have a debate with that SME, because I think it's nonsense. You prepare a VM specifically to provision with Foreman. Why would you care about the original contents?

guides/common/modules/proc_using-vmware-cloud-init-and-userdata-templates-for-provisioning.adoc
```diff
@@ -78,6 +78,8 @@ network:
   config: disabled
 EOM
 ----
++
+Back up the cfg file before you overwrite it.
```
apinnick, line 82: I recommend moving this step to "Prerequisites".

Example:

Back up the following files:
* `/etc/cloud/cloud.cfg.d/01_network.cfg`
* etc

Lennonka:
> I want to have a debate with that SME, because I think it's nonsense. You prepare a VM specifically to provision with Foreman. Why would you care about the original contents?

Well, would you suggest I closed the BZ as not a bug instead? It's an option. I also had doubts about it.

ekohl:
I'm looking at the whole procedure and now I'm wondering: why is this done? I'm not an expert, but AFAIK Foreman can provide a VMware with this user data via the API so you wouldn't need to use NoCloud: https://github.com/theforeman/foreman/blob/50a2b61a679ae952837ff1eb8675181bbccf7a9c/app/models/compute_resources/foreman/model/vmware.rb#L569-L572

Though reading https://developer.vmware.com/docs/14558/GUID-75E27FA9-2E40-4CBF-BF3D-22DCFC8F11F7.html we don't actually support this. Instead, we require these tedious manual steps to be taken by users.

In other words, we can't use https://cloudinit.readthedocs.io/en/latest/reference/datasources/vmware.html now.

Lennonka:
@ekohl So the procedure is alright after all? Then we can get back to the scope of this PR and you can perhaps answer my effin' question, please?

Lennonka:
Cherry-picked:
- c40110125a..fd09ac7691  3.9 -> 3.9
- cb663cbe8f..f3aef5f59a  3.8 -> 3.8
- 08a5c5f59e..9574fa43f4  3.7 -> 3.7
- 9676701da3..1d0b9c8fa8  3.6 -> 3.6
- e3481fadab..f60fb9fe1d  3.5 -> 3.5
- 9db96c07ad..d29dd2b709  3.4 -> 3.4
- 3d23a3af6d..811f8bf9a9  3.3 -> 3.3

Lennonka:
Cherry-picked:
- c14ff1f9c2..978b75a004  3.10 -> 3.10

PR #2802, opened 2024-02-20
Fix capitalization

maximiliankolb:


maximiliankolb:
@apinnick Can you please have a look at my PR?

guides/common/modules/proc_creating-a-custom-file-type-repository.adoc
```diff
@@ -17,7 +17,7 @@ To use the CLI instead of the {ProjectWebUI}, see the xref:cli-creating-a-custom
 . Optional: In the *Upstream URL* field, enter the URL of the upstream repository to use as a source.
 If you do not enter an upstream URL, you can manually upload packages.
 For more information, see xref:Uploading_Files_To_a_Custom_File_Type_Repository_{context}[].
-. Select the *Verify SSL* checkbox if you want to verify that the upstream repository's SSL certificates are signed by a trusted CA.
+. Select the *Verify SSL* checkbox if you want to verify that the SSL certificates of the repository are signed by a trusted CA.
```
apinnick, line 20: Small nit - Normally we do not specify the UI element if there is no chance of ambiguity. Also, I think you can remove "if you want to" because we assume that the user wants to do the action. 

I recommend you change this to `Select *Verify SSL* to verify that the SSL certificates of the repository are signed by a trusted CA.`

CONTRIBUTING.md
```diff
@@ -64,6 +64,7 @@ We capitalize the following terms:
 * Ansible
 * Candlepin
 * Chef
+* DEB
```
apinnick, line 67: Is there a source for this? I see "Deb" used frequently.

apinnick (commented):
Reviewed. Couple comments.

CONTRIBUTING.md
apinnick: I mean anywhere. For example, [debian.org](https://www.debian.org/). Does their documentation use "DEB"?

ekohl: https://www.debian.org/doc/manuals/debian-handbook/apt.en.html at least uses APT. https://en.wikipedia.org/wiki/Deb_(file_format) uses `deb`, all lower case.

maximiliankolb:
from Foreman 3.9: 

image

At first, I went with the smallest possible diff. Now I spoke with Quirin (who maintains pulp_deb) about it.

Thanks for all the input. I will have another look at it and currently lean towards "Deb" from our discussions. I will change this in a small follow-up PR.

apinnick:
LGTM

maximiliankolb:
Merged to "master" and cherry-picked to "3.9": :heavy_check_mark: 
   35005ad5d5..ea5d86db7b  3.9 -> 3.9

PR #2803, opened 2024-02-20
Update master to include Foreman 3.10

Griffin-Sullivan:

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/web/content/index.adoc
+++ b/web/content/index.adoc
@@ -29,7 +29,9 @@ The future version is built in nightly.
 
 * link:/release/nightly/[Nightly]
 
-// There is a release candidate available for testing.
+There is a release candidate available for testing.
+
+* link:/release/3.10/[Foreman 3.10 & Katello 4.12]
 
 These releases are unsupported and no longer receive updates. Users should update to a supported release.
 
--- a/web/content/js/versions.js
+++ b/web/content/js/versions.js
@@ -200,6 +200,138 @@ const navVersions = [
     ]
   },
   {
+    "title": "Foreman 3.10 - Katello 4.12 (RC)",
+    "path": "3.10",
+    "builds": [
+      {
+        "title": "Foreman on EL",
+        "filename": "index-foreman-el.html",
+        "guides": [
+          {
+            "title": "Release Notes",
+            "path": "Release_Notes"
+          },
+          {
+            "title": "Upgrading Foreman to 3.10",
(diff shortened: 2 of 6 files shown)


web/content/index.adoc
```diff
@@ -27,6 +27,8 @@ The following releases are supported, meaning they receive (security) updates.
 
 The future version is built in nightly.
 
+* link:/release/3.10/[Foreman 3.10 & Katello 4.12]
```
maximiliankolb, line 30: Please move this below line 32 and drop the `// ` so that there's a short introduction to the 3.10 RC.

web/content/js/versions.js
```diff
@@ -199,6 +199,206 @@ const navVersions = [
       }
     ]
   },
+  {
+    "title": "Foreman 3.10 - Katello 4.12 (unsupported)",
```
maximiliankolb, line 203:
```suggestion
    "title": "Foreman 3.10 - Katello 4.12 (RC)",
```

web/.ruby-version
maximiliankolb, line 1: Is this necessary? where does it come from?

maximiliankolb (commented):
one small question, rest LGTM.

web/.ruby-version
Griffin-Sullivan: Oops shouldn't have been added

web/content/js/versions.js
```diff
@@ -199,6 +199,206 @@ const navVersions = [
...
     ]
   },
+  {
+    "title": "Foreman 3.10 - Katello 4.12 (RC)",
+    "path": "3.10",
+    "builds": [
+      {
+        "title": "Foreman on EL",
+        "filename": "index-foreman-el.html",
+        "guides": [
```
ekohl, line 209: Many of these guides aren't ready and hidden on stable branches. I'd look at 3.9 and see which ones should be hidden. Same for Foreman on Debian.

ekohl (changes requested):

web/content/js/versions.js
ekohl: Well, the docs are "special". Many guides aren't really ready for non-Katello. Ideally speaking we'd review all of them and fix incorrect content, but until then we hide a log of guides for stable releases.

maximiliankolb (approved):
LGTM; any other comments?

adamlazik1 (approved):

PR #2806, opened 2024-02-21
Set of commits to use "client-*" attributes

maximiliankolb:

### DIFF
--- a/guides/common/attributes-base.adoc
+++ b/guides/common/attributes-base.adoc
@@ -59,8 +59,8 @@
 :customrpmtitle: RPM
 :customssl: SSL
 :customssltitle: SSL
-:Debian: Debian/Ubuntu
-:Debian-context: Debian-Ubuntu
+:DL: Debian/Ubuntu
+:DL-ID: Debian-Ubuntu
 :EL: Enterprise Linux
 :foreman-example-com: foreman.example.com
 :foreman-installer: foreman-installer
--- a/guides/common/attributes-base.adoc
+++ b/guides/common/attributes-base.adoc
@@ -133,14 +133,15 @@
 :Team: {Project} community
 :sectanchors:
 :client-content-dnf:
-:client-installation-medium: centos
 // Hosts
+:client-installation-medium: centos
 :client-content-type: yum
 :client-container-image-name: rhel7
 :client-container-product-name: Red Hat Container Catalog
 :client-container-repository-name: RHEL7
 :client-container-url: http://registry.access.redhat.com/
 :client-debian-installation-medium-path: http://deb.debian.org/debian/
+:client-installation-medium-path: download.example.com/{client-installation-medium}/$version/Server/$arch/os/
 :client-os-family-hammer: Redhat
 :client-os-family: Red Hat
 :client-os: {EL}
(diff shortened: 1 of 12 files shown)


guides/common/attributes-base.adoc
```diff
@@ -59,8 +59,8 @@
 :customrpmtitle: RPM
 :customssl: SSL
 :customssltitle: SSL
-:Debian: Debian/Ubuntu
-:Debian-context: Debian-Ubuntu
+:DL: Debian/Ubuntu
```
ekohl, line 62: I think DL is ambiguous and haven't seen that used elsewhere. Would `deb-os` be more descriptive?

maximiliankolb: I'd like to keep it similar to "EL" here. It's not yet used in upstream but in downstream docs. I've added it as "Debian Linux" and the colleague who reviewed it was like "ah, Debian-like as attribute for Debian and Ubuntu".

adamlazik1: For me, both `DL` and `deb-os` is fine. There is a parallel between `DL` and `EL`, so no issues from my side.

Lennonka: DL is fine!

guides/common/modules/proc_creating-an-operating-system-for-debian.adoc
```diff
@@ -31,6 +31,12 @@ endif::[]
 +
 For more information, see xref:creating-partition-tables_{context}[].
 . On the *Templates* tab, select the `Preseed Finish` template as *Finish template*, the `Preseed` template as *Provisioning Template*, and the `Preseed PXELinux` template as *PXELinux template*.
+ifdef::orcharhino[]
+. On the *Parameters* tab, add the `or_client_repo_url` parameter, select the *string* type, and enter the value `\http://_{foreman-example-com}_/pulp/repos/_My_Organization_/Library/custom/{client-os}_Client/{client-os}_Client_{client-os-major}/`.
++
+If you want to provision hosts through {SmartProxyServers}, you have to use the `or_client_repo_path` parameter containing the *Published At* path of your content on {ProjectServer}.
+{SmartProxyServer} will add the FQDN of your {SmartProxyServer} to generate the URL based on the content source setting of your host.
+endif::[]
```
adamlazik1, line 39: Optional: This can be a snippet, but it can also stay as it is. Your call. Also, the format of the URL differs in the two instances, is that intentional?

guides/common/modules/proc_creating-partition-tables.adoc
```diff
@@ -29,28 +31,25 @@ autopart
...
 {Project} adds the partition table to the current provisioning context.
 . Click *Submit* to save your partition table.
 
 [id="cli-creating-partition-tables_{context}"]
 .CLI procedure
-
 . Before you create a partition table with the CLI, create a plain text file that contains the partition layout.
-This example uses the `~/my-partition` file.
-
+This example uses the `~/My_Partition_Table_` file.
```
adamlazik1, line 43:
```suggestion
This example uses the `_~/My_Partition_Table_` file.
```

guides/common/modules/proc_creating-an-operating-system-for-debian.adoc
maximiliankolb: Thanks, fixed by adding a snippet and attribute (which I can overwrite in downstream for each OS differently)

maximiliankolb (commented):
Thanks, kindly asking for a re-review @adamlazik1

adamlazik1 (approved):
LGTM

maximiliankolb:
Merged to "master" and cherry-picked: :heavy_check_mark: 
   48fd23f888..317ff88048  3.10 -> 3.10
   7e07e03cfd..1713776a57  3.9 -> 3.9

PR #2817, opened 2024-02-22
Add procedure to add passphrase to SSH key used for REX

adamlazik1:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2170031

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_adding-a-passphrase-to-ssh-key-used-for-remote-execution.adoc
+++ b/guides/common/modules/proc_adding-a-passphrase-to-ssh-key-used-for-remote-execution.adoc
@@ -1,0 +1,16 @@
+[id="adding-a-passphrase-to-ssh-key-used-for-remote-execution_{context}"]
+= Adding a passphrase to SSH key used for remote execution
+
+By default, {SmartProxy} uses a non-passphrase protected SSH key to execute remote jobs on hosts.
+You can protect the SSH key with a passphrase by following this procedure.
+
+.Procedure
+* On your {ProjectServer} or {SmartProxyServer}, use `ssh-keygen` to add a passphrase to your SSH key:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# ssh-keygen -p -f ~foreman-proxy/.ssh/id_rsa_foreman_proxy
+----
+
+.Next steps
+* Users now must use a passphrase when running remote execution jobs on hosts.
--- a/guides/common/assembly_configuring-and-setting-up-remote-jobs.adoc
+++ b/guides/common/assembly_configuring-and-setting-up-remote-jobs.adoc
@@ -30,6 +30,8 @@ include::modules/proc_distributing-ssh-keys-for-remote-execution.adoc[leveloffse
 
 include::modules/proc_distributing-ssh-keys-for-remote-execution-manually.adoc[leveloffset=+1]
 
+include::modules/proc_adding-a-passphrase-to-ssh-key-used-for-remote-execution.adoc[leveloffset=+1]
+
 include::modules/proc_using-the-api-to-obtain-ssh-keys-for-remote-execution.adoc[leveloffset=+1]
 
 ifndef::satellite[]


guides/common/modules/proc_creating-new-ssh-keys-for-remote-execution.adoc
```diff
@@ -0,0 +1,19 @@
+[id="creating-new-ssh-keys-for-remote-execution_{context}"]
+= Creating new SSH keys for remote execution
+
+By default, {Project} uses a non-passphrase protected SSH key to execute remote jobs on hosts.
+You can create additional passphrase protected or non-passphrase protected SSH keys to authenticate for remote execution jobs on hosts.
```
adamruzicka, line 5: The word `additional` could be misleading in this context. It makes it read as if you could use multiple keys at once, while a smart proxy expects exactly one key.

guides/common/modules/proc_creating-new-ssh-keys-for-remote-execution.adoc
```diff
@@ -0,0 +1,19 @@
...
+
+.Procedure
+. On your {ProjectServer} or {SmartProxyServer}, create a new SSH key:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# ssh-keygen
```
adamruzicka, line 12: Just generating the key is not enough. The key in questio has to be placed into an expected location (and thus replace the old key) so something like `ssh-keygen -f ~foreman-proxy/.ssh/id_rsa_foreman_proxy` would be appropriate

Alternatively, you can change the expected location by passing a combaination of `--foreman-proxy-plugin-remote-execution-script-ssh-identity-dir=$DIR` and `--foreman-proxy-plugin-remote-execution-script-ssh-identity-file=$KEY` to the installer

adamruzicka (changes requested):

adamlazik1:
Since there can be only one key I opted for adding a passphrase to existing key instead. This way there is no additional action required for existing hosts.

guides/common/modules/proc_adding-a-passphrase-to-ssh-key-used-for-remote-execution.adoc
```diff
@@ -0,0 +1,16 @@
...
+.Procedure
+* On your {ProjectServer} or {SmartProxyServer}, use `ssh-keygen` to add a passphrase to your SSH key:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# ssh-keygen -p -f ~foreman-proxy/.ssh/id_rsa_foreman_proxy
+----
+
+.Next steps
+* Users now must authenticate with a passphrase when running remote execution jobs on hosts.
```
adamruzicka, line 16: Nit: You use the passphrase to unlock the key and then authenticate with the unlocked key. You don't really authenticate with the passphrase

guides/common/modules/proc_adding-a-passphrase-to-ssh-key-used-for-remote-execution.adoc
```diff
@@ -0,0 +1,16 @@
...
+
+By default, {SmartProxy} uses a non-passphrase protected SSH key to execute remote jobs on hosts.
+You can protect the SSH key with a passphrase by following this procedure.
+
+.Procedure
+* On your {ProjectServer} or {SmartProxyServer}, use `ssh-keygen` to add a passphrase to your SSH key:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# ssh-keygen -p -f ~foreman-proxy/.ssh/id_rsa_foreman_proxy
```
adamruzicka, line 12: Should this mention that this is just the default path?

adamlazik1: I was considering something like "if you changed the location of your ssh key, then use the new location" but just sounds too obvious, so I would probably opt for not mentioning that unless there are objections.

guides/common/modules/proc_adding-a-passphrase-to-ssh-key-used-for-remote-execution.adoc
adamlazik1: Changed wording.

adamruzicka (approved):
LGTM

adamlazik1:
Merged and cherry-picked:
   3cb5aecad3..a2c0d0b37d  3.10 -> 3.10
   3569e9d026..5f13a91957  3.3 -> 3.3
   d338e4af51..ce59624bd8  3.4 -> 3.4
   617e5f4327..9786906eae  3.5 -> 3.5
   1dd717a432..731bb132fe  3.6 -> 3.6
   f84b27533b..144f1eab68  3.7 -> 3.7
   a86277a422..33c2d9d1ce  3.8 -> 3.8
   dec4d737df..902ba48156  3.9 -> 3.9

PR #2828, opened 2024-02-26
Add information about Distinguished Names

AkshayGadhaveRH:
When generating a certificate signing request, adding DN information increases trust. Although this is optional, users can benefit from this information.

There are two ways of doing this:
- Add the configuration to the openssl.cnf file.
- Provide the details when prompted by the command.

However, in the docs, we specifically ask the user to provide "prompt = no", this will tell the openssl command to not prompt for these details. So, adding this to the `openssl.cnf` file under the `[ req_distinguished_name ]` section.

BZ link:https://bugzilla.redhat.com/show_bug.cgi?id=2256000

* [X] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [X] Foreman 3.9/Katello 4.11 (planned Satellite 6.15)
* [X] Foreman 3.8/Katello 4.10
* [X] Foreman 3.7/Katello 4.9 (Satellite 6.14)
* [X] Foreman 3.6/Katello 4.8
* [X] Foreman 3.5/Katello 4.7 (Satellite 6.13; orcharhino 6.6/6.7)
* [X] Foreman 3.4/Katello 4.6 (EL8 only)
* [X] Foreman 3.3/Katello 4.5 on EL7 & EL8 (Satellite 6.12 on EL8 only; orcharhino 6.4/6.5 on EL8 only)
* We do not accept PRs for Foreman older than 3.3.

PR #2832, opened 2024-02-27
Rectify	differences in the custom SSL CSR for LB and Smart Proxy

AkshayGadhaveRH:
There are differences in the `openssl.cnf` files created for cert signing requests for LB and for Smart Proxy:
- guides/common/modules/proc_creating-custom-ssl-certificates-for-smart-proxy-server.adoc > LB custom SSL CSR file
- guides/common/modules/proc_creating-a-custom-ssl-certificate.adoc > Smart Proxy custom SSL CSR file

Trying to make these two files similar as far as possible.

This PR will also include changes to `guides/common/modules/proc_creating-a-custom-ssl-certificate.adoc`. However, there is another PR in progress for changes in this file: https://github.com/theforeman/foreman-documentation/pull/2828

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2256021

* [x] Foreman 3.6/Katello 4.8


guides/common/modules/proc_creating-custom-ssl-certificates-for-smart-proxy-server.adoc
```diff
@@ -31,33 +31,43 @@ If you already have a private key for this {SmartProxyServer}, skip this step.
...
 DNS.1 = _{loadbalancer-example-com}_
 DNS.2 = _{smartproxy-example-com}_
 ----
 <1> The certificate's common name must match the FQDN of {SmartProxyServer}.
 Ensure to change this when running the command on each {SmartProxyServer} that you configure for load balancing.
 You can also set a wildcard value `*`.
 If you set a wildcard value, you must add the `-t {certs-proxy-context}` option when you use the `katello-certs-check` command.
-<2> Under `[alt_names]`, include the FQDN of the load balancer as `DNS.1` and the FQDN of {SmartProxyServer} as `DNS.2`.
-
+<2> Two letter code.
```
mjivraja, line 66:
```suggestion
<2> Two-letter code of a country. For example, US.
```

guides/common/modules/proc_creating-custom-ssl-certificates-for-smart-proxy-server.adoc
```diff
@@ -31,33 +31,43 @@ If you already have a private key for this {SmartProxyServer}, skip this step.
...
+<3> Full name.
```
mjivraja, line 67:
```suggestion
<3> Full name of a state or province.
```

guides/common/modules/proc_creating-custom-ssl-certificates-for-smart-proxy-server.adoc
```diff
@@ -31,33 +31,43 @@ If you already have a private key for this {SmartProxyServer}, skip this step.
...
+<4> For example, a city name.
```
mjivraja, line 68:
```suggestion
<4> Name of a locality. For example, a city name.
```

mjivraja (commented):
How about this?

guides/common/modules/proc_creating-custom-ssl-certificates-for-smart-proxy-server.adoc
```diff
@@ -31,33 +31,43 @@ If you already have a private key for this {SmartProxyServer}, skip this step.
 default_bits       = 4096
 distinguished_name = req_distinguished_name
 req_extensions     = req_ext
+x509_extensions = usr_cert
```
evgeni, line 34: I don't think we need those (in both places!), as `x509_extensions` apply to the certificate, and that's what the CA produces based on our input, not what we produce using the commands in these procedures.

evgeni:
IMHO the only difference in those two sections should be the number of `DNS.X` entries in the `alt_names` section.

AkshayGadhaveRH:
@evgeni does this look correct?

guides/common/modules/proc_creating-custom-ssl-certificates-for-smart-proxy-server.adoc
```diff
@@ -57,7 +51,22 @@ Ensure to change this when running the command on each {SmartProxyServer} that y
 You can also set a wildcard value `*`.
 If you set a wildcard value, you must add the `-t {certs-proxy-context}` option when you use the `katello-certs-check` command.
 <2> Under `[alt_names]`, include the FQDN of the load balancer as `DNS.1` and the FQDN of {SmartProxyServer} as `DNS.2`.
-
+Optional: If you want to add Distinguished Name (DN) details to the CSR, add the following information to the `[ req_distinguished_name ]` section:
```
evgeni, line 54: we don't have this section in https://docs.theforeman.org/nightly/Installing_Server/index-katello.html#Configuring_Server_with_a_Custom_SSL_Certificate_foreman, should we add it there too?

AkshayGadhaveRH: I have filed a PR for this https://github.com/theforeman/foreman-documentation/pull/2828/files
I'll make changes in this file (including dropping the x509 sections) once this PR is merged.

AkshayGadhaveRH: @evgeni done!

guides/common/modules/proc_creating-a-custom-ssl-certificate.adoc
```diff
@@ -36,29 +36,19 @@ If you already have a private key for this {ProductName}, skip this step.
 [ req ]
 req_extensions = v3_req
 distinguished_name = req_distinguished_name
-x509_extensions = usr_cert
 prompt = no
 
 [ req_distinguished_name ]
-CN = _{context}.example.com_
+commonName = _{smartproxy-example-com}_
```
Lennonka, line 42: If this file is re-used for both Foreman Server and Foreman SmartProxy, then `{context}.example.com` is correct.

Lennonka (changes requested):
Is it necessary to have both files?
Especially when one of them is used for both Foreman Server and SmartProxy.

Lennonka:
May need rebase after #2872 is merged.

AkshayGadhaveRH:
@Lennonka let me check if the same file can be used in both places. There is a slight difference in the two files in the alt names in the code blocks and references to the LB in one file, but let me check if otherwise they are the same.

AkshayGadhaveRH:
@Lennonka tried merging the two modules together, this might need a few tweaks though. Could you take a look?

Lennonka:
@AkshayGadhaveRH I can't see the changes. Forgot to push?

AkshayGadhaveRH:
@Lennonka whoops yep :')

guides/doc-Configuring_Load_Balancer/master.adoc
```diff
@@ -34,7 +34,7 @@ include::common/modules/proc_configuring-remaining-smart-proxy-servers-with-defa
 
 include::common/modules/con_configuring-smart-proxy-server-with-custom-ssl-certificates-for-load-balancing-without-puppet.adoc[leveloffset=+2]
 
-include::common/modules/proc_creating-custom-ssl-certificates-for-smart-proxy-server.adoc[leveloffset=+3]
+include::common/modules/proc_creating-a-custom-ssl-certificate.adoc[leveloffset=+3]
```
Lennonka, line 37:
```suggestion
:ProductName: {SmartProxyServer}
include::common/modules/proc_creating-a-custom-ssl-certificate.adoc[leveloffset=+3]
```
because
```
asciidoctor: WARNING: skipping reference to missing attribute: productname
```

Lennonka:
@AkshayGadhaveRH TBH, I wonder if I lead you astray when I suggested using proc_creating-a-custom-ssl-certificate.adoc for LB. LB uses "load-balancing" as the context, whereas other includes use either "satellite" or "capsule" as the context. Therefore, I'm not sure if it's rendering appropriately when used in LB. What do you think? Are you sure this is the right approach?

AkshayGadhaveRH:
@Lennonka I went through the original guides and the builds for this PR, the content seems to be rendering correctly for both guides.

guides/common/modules/proc_creating-a-custom-ssl-certificate.adoc
Lennonka: This doesn't render correctly for https://theforeman-foreman-documentation-preview-pr-2832.surge.sh/nightly/Installing_Server/index-satellite.html#creating-a-custom-ssl-certificate_satellite

maximiliankolb: I think we should not overwrite the context but instead "wrap" the module in the assembly with a new attributes such as "target" or "ssl-common-name" or similar. Then we have full control and only change what we actually want to change without affecting other instances of "{context}".

AkshayGadhaveRH: @maximiliankolb sounds good to me! @Lennonka should I do this instead?

Lennonka: Yes, please.

AkshayGadhaveRH: @Lennonka @maximiliankolb done. Does this look okay now?

guides/common/modules/proc_creating-a-custom-ssl-certificate.adoc
```diff
@@ -29,43 +34,67 @@ If you already have a private key for this {ProductName}, skip this step.
...
+If you set a wildcard value, you must add the `-t {certs-proxy-context}` option when you use the `katello-certs-check` command.
+<2> Under `[alt_names]`, include the FQDN of the load balancer as `DNS.1` and the FQDN of {SmartProxyServer} as `DNS.2`.
+endif::[]
+. Optional: If you want to add Distinguished Name (DN) details to the CSR, add the following information to the `[ req_distinguished_name ]` section:
 +
 [options="nowrap", subs="+quotes,attributes"]
 ----
 [req_distinguished_name]
-CN = _{context}.example.com_
+CN = _{smartproxy-example-com}_
```
Lennonka, line 97: Missed this?

guides/common/modules/proc_creating-custom-ssl-certificates-for-smart-proxy-server.adoc
Lennonka, line 1: When this file is no longer used, we can probably delete it.

guides/common/assembly_configuring-capsule-custom-server-certificate.adoc
```diff
@@ -11,6 +11,7 @@ To configure your {SmartProxyServer} with a custom certificate, complete the fol
 . xref:deploying-a-custom-ssl-certificate-to-{smart-proxy-context}-server_{smart-proxy-context}[]
 . xref:deploying-a-custom-ssl-certificate-to-hosts_{smart-proxy-context}[]
 
+:ssl-common-name: {smartproxy-example-com}
 //Creating a Custom SSL Certificate for {SmartProxyServer}
 include::modules/proc_creating-a-custom-ssl-certificate.adoc[leveloffset=+1]
```
maximiliankolb, line 17:
```suggestion
:!ssl-common-name:

```

Please also add this to the other file.

guides/doc-Configuring_Load_Balancer/master.adoc
```diff
@@ -34,16 +34,16 @@ include::common/modules/proc_configuring-remaining-smart-proxy-servers-with-defa
 
 include::common/modules/con_configuring-smart-proxy-server-with-custom-ssl-certificates-for-load-balancing-without-puppet.adoc[leveloffset=+2]
 
-include::common/modules/proc_creating-custom-ssl-certificates-for-smart-proxy-server.adoc[leveloffset=+3]
+:ssl-common-name: {smartproxy-example-com}
+:ProductName: {SmartProxyServer}
+include::common/modules/proc_creating-a-custom-ssl-certificate.adoc[leveloffset=+3]
```
maximiliankolb, line 39:
```suggestion
include::common/modules/proc_creating-a-custom-ssl-certificate.adoc[leveloffset=+3]
:!ProductName:
```

Unless "ProductName" was set before, then please use the following:

````
:parent-ProductName: {ProductName}
my include ...
:ProductName: {parent-ProductName}
:!parent-ProductName:
````

**Review by maximiliankolb — COMMENTED** [review]
Yes, makes sense to me!

File: guides/doc-Configuring_Load_Balancer/master.adoc (hunk 10 above)
Comment (AkshayGadhaveRH) [reply guides/doc-Configuring_Load_Balancer/master.adoc:39]: @maximiliankolb if I set these values on line 37, the value it acquires is `Foreman` and not `Smart Proxy`, looks like I have to set it to `{SmartProxyServer}` manually first.

Comment (maximiliankolb) [reply guides/doc-Configuring_Load_Balancer/master.adoc:39]: No, you should not have to set "SmartProxyServer" anywhere because it's part of guides/common/attributes*.adoc".

Maybe you can do some local testing and see what's the value of "ProductName" is certain points in the docs.

Comment (AkshayGadhaveRH) [reply guides/doc-Configuring_Load_Balancer/master.adoc:39]: So the value is specifically set to `{SmartProxyServer}` in `doc-Installing_Proxy/master.adoc` otherwise it is set to `{ProjectServer}`. The `doc-doc-Configuring_Load_Balancer/master.adoc` file picks the universal value which is `{ProjectServer}`.

I have tweaked the `parent-ProductName` local variable usage and used it a little more flexibly. Does that work?

Comment (AkshayGadhaveRH) [reply guides/doc-Configuring_Load_Balancer/master.adoc:39]: okay looks like it doesn't :(

**Review by maximiliankolb — CHANGES_REQUESTED** [review]
I think there's a misunderstanding.

The idea was to "save" the value of the "ProductName" attribute as "parent-ProductName", then overwrite it, and keep using "ProductName" in the procedure.

**Comment by AkshayGadhaveRH** [comment]
@maximiliankolb in that case, does it make sense to have a local variable instead?

**Comment by maximiliankolb** [comment]
@AkshayGadhaveRH Yes, I think using a variable similar to you use "ssl-common-name" is also a nice solution.

**Comment by AkshayGadhaveRH** [comment]
@maximiliankolb @Lennonka added `parent-ProductName` as a "local" variable. Looks like it works fine.

File: guides/common/assembly_configuring-capsule-custom-server-certificate.adoc [hunk 11]
```diff
@@ -11,8 +11,12 @@ To configure your {SmartProxyServer} with a custom certificate, complete the fol
 . xref:deploying-a-custom-ssl-certificate-to-{smart-proxy-context}-server_{smart-proxy-context}[]
 . xref:deploying-a-custom-ssl-certificate-to-hosts_{smart-proxy-context}[]
 
+:parent-ProductName: {SmartProxyServer}
```
maximiliankolb, line 14:
```suggestion
:parent-ProductName: {ProductName}
:ProductName: {SmartProxyServer}
```

With this, you can use the "{ProductName}" attribute in your module as expected

guides/common/assembly_configuring-capsule-custom-server-certificate.adoc
```diff
@@ -11,8 +11,12 @@ To configure your {SmartProxyServer} with a custom certificate, complete the fol
...
+:ssl-common-name: {smartproxy-example-com}
 //Creating a Custom SSL Certificate for {SmartProxyServer}
 include::modules/proc_creating-a-custom-ssl-certificate.adoc[leveloffset=+1]
+:!ssl-common-name:
+:!parent-ProductName:
```
maximiliankolb, line 19:
```suggestion
:ProductName: {parent-ProductName}
:!parent-ProductName:
```

maximiliankolb (changes requested):
Made two suggestions to simplifiy attribute usage. In the module itself, you should use "ProductName".

"parent-ProductName" is just to "save" the value temporarily. I hope this explains things. If not, please feel free to reach out to me on Matrix or Github!

guides/common/assembly_configuring-capsule-custom-server-certificate.adoc
AkshayGadhaveRH: Oh yes! That's a great idea!

guides/common/assembly_configuring-capsule-custom-server-certificate.adoc
AkshayGadhaveRH: Getting `unidentified attribute productname` in `deploy / build-html`. Other build seems to be happening fine, works locally for me too.

AkshayGadhaveRH: Local build fails for `make browser BUILD=katello` with the same error for `productname`, but if I follow this up with a ` make browser BUILD=satellite`, that succeeds. Now if I try to build for Katello again, it succeeds.

guides/doc-Configuring_Load_Balancer/master.adoc
```diff
@@ -34,16 +34,26 @@ include::common/modules/proc_configuring-remaining-smart-proxy-servers-with-defa
 
 include::common/modules/con_configuring-smart-proxy-server-with-custom-ssl-certificates-for-load-balancing-without-puppet.adoc[leveloffset=+2]
 
-include::common/modules/proc_creating-custom-ssl-certificates-for-smart-proxy-server.adoc[leveloffset=+3]
+:parent-ProductName: {ProductName}
```
maximiliankolb, line 37: I looks like "ProductName" is not set in https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes.adoc?plain=1 and https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-base.adoc which means that you cannot/do not have to safe the original value in "parent-productname".

So maybe you can try to just set and unset it without using the "parent-X" and "X: parent" lines.

AkshayGadhaveRH: Yep! That did the trick!

maximiliankolb (commented):
Thanks Akshay, structure LGTM. Can someone else please do a technical review? I don't have a testing scenario for this right now.

Lennonka:
@AkshayGadhaveRH You can probably ask a QE from your team to test the new procedures.

guides/common/modules/proc_creating-a-custom-ssl-certificate.adoc
```diff
@@ -11,6 +11,15 @@ On {ProjectServer}, create a custom certificate for your {ProductName}.
 If you already have a custom SSL certificate for {ProductName}, skip this procedure.
 endif::[]
 
+ifeval::["{context}" == "load-balancing-with-custom-ssl-certificates-with-puppet"]
+:context: load-balancing
+endif::[]
```
Lennonka, line 16: I'm not sure if changing the context mid-air is a good idea.
Perhaps you can use `ifdef:load-balancing[]` instead of ifeval.
This attribute is set for the LB guide.

guides/common/modules/proc_creating-a-custom-ssl-certificate.adoc
```diff
@@ -76,6 +109,7 @@ organizationalUnitName = _My_Organizational_Unit_Name_ <4>
 <2> Full name
 <3> Full name (example: New York)
 <4> Division responsible for the certificate (example: IT department)
++
```
Lennonka, line 112:
```suggestion
```
Not needed before a new list item.

guides/common/modules/proc_creating-a-custom-ssl-certificate.adoc
```diff
@@ -88,10 +122,29 @@ organizationalUnitName = _My_Organizational_Unit_Name_ <4>
 <1> Path to the private key
 <2> Path to the configuration file
 <3> Path to the CSR to generate
-
++
```
Lennonka, line 125:
```suggestion
```
Not needed before a new list item.

Lennonka (commented):
A couple more suggestions.

guides/common/modules/proc_creating-a-custom-ssl-certificate.adoc
AkshayGadhaveRH: So I have to manipulate the context because there are two instances of `common/modules/proc_creating-a-custom-ssl-certificate.adoc` in `doc-Configuring_Load_Balancer/master.adoc`, this fails because there are duplicate IDs in the same doc.

Lennonka: So perhaps instead of using context to populate the values in the modules you can use a new attribute and set it on the `master.adoc` level? Then you don't have to change the context.

AkshayGadhaveRH: Introduced a new attribute `context_title`.

Lennonka:
@AkshayGadhaveRH I'm sorry, I didn't explain clearly what I meant.
Try applying the patch I've sent you (git apply --check file.patch) and if the result is acceptable, please use it.

Lennonka (changes requested):
Please, patch.

maximiliankolb (commented):
I cannot comment much on the technical details, but the diff, esp. attribute usage looks good to me!

Lennonka (approved):
Looks good and clean content-wise.

guides/doc-Configuring_Load_Balancer/master.adoc
Lennonka, line 1: Oops. We need to add the `:load-balancing:` attribute definition at the beginning of this file. Otherwise, the modules will not produce the correct output.

Lennonka (changes requested):
Please, fix that ^^ and then double-check if the outputs render as expected.
I'd consider a final tech review before merging as well.

Lennonka (approved):
LGTM

AkshayGadhaveRH:
@Lennonka looks good to me

Lennonka:
Cherry-picked:
- d6d7283e1e..b366b61e3f  3.10 -> 3.10
- c14a19c6f7..c535864ba5  3.9 -> 3.9

Merge conflicts on 3.8. Please file another PR @AkshayGadhaveRH

PR #2839, opened 2024-02-28
Add warning to overriding Ansible variables

adamlazik1:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2215453

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_overriding-ansible-variables-in-project.adoc
+++ b/guides/common/modules/proc_overriding-ansible-variables-in-project.adoc
@@ -13,6 +13,7 @@ To ensure the variable that you override follows the correct order of precedence
 .Prerequisites
 * You must have Ansible variables in {Project}.
 For more information, see xref:Importing_Ansible_Roles_and_Variables_{context}[]
+* To use overridden Ansible variables, a user must have a role that allows them to see those variables.
 
 .Procedure
 . In the {ProjectWebUI}, navigate to *Configure* > *Ansible* > *Variables*.


guides/common/modules/proc_overriding-ansible-variables-in-project.adoc
```diff
@@ -6,6 +6,11 @@ If you run Ansible roles in {Project}, you can use {Project} to override Ansible
 The following procedure refers to hosts and host groups.
 For more information, see {ManagingHostsDocURL}[_{ManagingHostsDocTitle}_].
 
+[WARNING]
+====
+To be able to use overriden Ansible variables, a user must have a role that allows them to see the those variables.
```
mjivraja, line 11:
```suggestion
To be able to use overridden Ansible variables, a user must have a role that allows them to see those variables.
```
Some options I prepared for you if you like:
1) Users must have roles allowing them to see overridden Ansible variables in order to use them.
2) The user must have the right role to see overridden Ansible variables if they want to use them.

mjivraja (changes requested):
Some options.

maximiliankolb:
Does this make sense to you @OttaviaB ?

guides/common/modules/proc_overriding-ansible-variables-in-project.adoc
adamlazik1: Thank you for the fix, applied.

guides/common/modules/proc_overriding-ansible-variables-in-project.adoc
```diff
@@ -6,6 +6,11 @@ If you run Ansible roles in {Project}, you can use {Project} to override Ansible
 The following procedure refers to hosts and host groups.
 For more information, see {ManagingHostsDocURL}[_{ManagingHostsDocTitle}_].
 
+[WARNING]
+====
+To use overridden Ansible variables, a user must have a role that allows them to see those variables.
```
asteflova, line 11: Can you please take a look at line 53? https://github.com/theforeman/foreman-documentation/blob/6243070f7c89f52704201590ebf008fc7ad4fa79/guides/common/modules/proc_overriding-ansible-variables-in-project.adoc?plain=1#L53

Both your addition, and line 53, talk about how to use the overridden variables, right? I think you should document both together.

adamlazik1: I am not sure about that. That looks like next step after the procedure, while this warning is more of a prerequisite. Which reminds me, I should probably just put it into prerequisites instead of a warning. Would that be better?

adamlazik1: @asteflova

asteflova: Oh, thanks for explaining! Yes, adding it to `.Prerequisites` is a great idea then.

adamlazik1: Alright, it is now in prerequisites.

adamruzicka (approved):
Nothing to add, +1

adamlazik1:
Waiting until tomorrow and merging unless there are objections.

maximiliankolb:
I spoke with Ottavia about this PR and we cannot yet reproduce this on Foreman 3.5 and 3.9.

adamruzicka:
Should be like that everywhere, as long as you have ansible there

adamlazik1:
@maximiliankolb this behavior was confirmed by Adam and also in the linked BZ, I believe that should be enough SME verification.

PR #2840, opened 2024-02-28
Store the menu navigation in a file per release

ekohl:
Previously there was one huge file for all the versions. This was difficult to maintain. With this a single file per release is introduced. That allows easy copying when a release is branched and provides an easy overview which builds belong to which release.

To reduce some duplication, it automatically inserts the path attribute.

There is still some room for improvement on the title. This could be composed from values in `guides/common/attributes.adoc`. In a further theoretical future every release could contain a `menu.json` file, allowing the maintenance to happen inside the individual branches.

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [ ] Foreman 3.9/Katello 4.11 (planned Satellite 6.15)
* [ ] Foreman 3.8/Katello 4.10
* [ ] Foreman 3.7/Katello 4.9 (Satellite 6.14)
* [ ] Foreman 3.6/Katello 4.8
* [ ] Foreman 3.5/Katello 4.7 (Satellite 6.13; orcharhino 6.6/6.7)
* [ ] Foreman 3.4/Katello 4.6 (EL8 only)
* [ ] Foreman 3.3/Katello 4.5 on EL7 & EL8 (Satellite 6.12 on EL8 only; orcharhino 6.4/6.5 on EL8 only)
* We do not accept PRs for Foreman older than 3.3.

PR #2845, opened 2024-02-28
Implement the releases as a data source

ekohl:
Previously there was one huge file for all the versions in the navigation. In addition to that, there were individual release pages. This was difficult to maintain.

The new way uses [Nanoc data sources](https://nanoc.app/doc/data-sources/) to essentially act as a database. Based on that, the various release pages are generated from a single source. For that, guides are identified by tags so they can be renamed without affecting the website. The releases on the index page are also generated.

That allows easy copying when a release is branched and provides an easy overview which builds belong to which release.

There is still some room for improvement. A lot of content in the release file overlaps with the content in guides. If it could be extracted from guides, that would reduce maintenance even further.

This is a more native way of implementing https://github.com/theforeman/foreman-documentation/pull/2840. The primary benefit is that it's aware of the release files and pages are live rerendered when a change is made.

This is a draft because not all release have implemented the tags.

* [ ] Foreman 3.6/Katello 4.8


ekohl:
> This is a draft because not all release have implemented the tags.

And in addition to that: the release pages also contain links to guides with similar names. The links on overview are also broken. But I'm calling it a day.

ekohl:
I have dropped the dynamic per-release page and moved to static pages. fb3e7f904c87cce7878b2f63aea69eabbf1b49cf is a commit I wrote that does it mostly, but I'm not happy with it yet and I don't want to block the whole effort on it.

Please take a look and let me know what you think.

Lennonka (approved):
Brilliant, I love it!
Not sure I'm qualified to comment on the code.
Have you tested it locally?

web/nanoc.yaml
```diff
@@ -7,6 +7,8 @@ prune:
 data_sources:
   - type: filesystem
     encoding: utf-8
+  - type: releases
+    items_root: /release
```
Lennonka, line 11:
```suggestion
    items_root: /releases
```
Typo?

web/releases/3.8.json
```diff
@@ -0,0 +1,116 @@
+{
+	"katello": "4.10",
```
maximiliankolb, line 2: identation looks weird.

web/releases/3.7.json
```diff
@@ -0,0 +1,87 @@
+{
+  "katello": "4.9",
```
maximiliankolb, line 2: no state == unsupported?

maximiliankolb (commented):
Thanks Ewoud!

I had a look at the adoc and json files; they look sane to me. If you feel confident, please merge and let's try it out!

web/nanoc.yaml
ekohl: Looking at https://docs.theforeman.org/release/3.10/ I think it should be `/release`.

evgeni (approved):
needs a rebase after https://github.com/theforeman/foreman-documentation/commit/967b8c8b453f97122b4864049d9a7b247af1b5f7 but otherwise :+1:

README.md
```diff
@@ -65,8 +65,11 @@ When a commit is pushed into `X.Y`:
...
 * Update master
   * Update `ProjectVersionPrevious` to `X.Y` in `guides/common/attributes.adoc`
-  * Create a copy of [/web/content/release-nightly.md](https://github.com/theforeman/foreman-documentation/tree/master/web/content/release-nightly.md) as `release-X.Y.md` and edit it accordingly.
-  * Add the new release versions to the site's index page and navigation bar by editing [/web/content/index.adoc](https://github.com/theforeman/foreman-documentation/blob/master/web/content/index.adoc) and [/web/content/js/versions.js](https://github.com/theforeman/foreman-documentation/blob/master/web/content/js/versions.js).
+  * Create a copy of [/web/releases/nightly.json](https://github.com/theforeman/foreman-documentation/tree/master/web/releases/nightly.json) as `X.Y.json` and edit it accordingly. Remove guides which aren't ready for stable branches.
+  * Create a copy of [/web/releases/nightly.adoc](https://github.com/theforeman/foreman-documentation/tree/master/web/releases/nightly.adoc) as `X.Y.adoc` and edit it accordingly.
+    * Set the `state` to `RC`
+    * Add `katello` with the right version
+    * Remove guides which aren't ready for stable branches.
```
evgeni, line 72: I think .adoc needs other changes than listed here?

ekohl: Right, these apply to the `.json`.

ekohl:
Updated the README to change the `nightly.json` and `nightly.adoc` files. Also added `katello` to `nightly.json` so on branching you only need to update the value instead of adding a line.

PR #2875, opened 2024-03-12
Fix capitalization of attributes

apinnick:
- Change capitalization of title attributes from title case to sentence case.
- Remove title attribute if not required.
- Add :Kubernetesfirst: attribute to comply with Red Hat branding requirements.

* [ ] Foreman 3.6/Katello 4.8


maximiliankolb (commented):
Full ACK on "Cockpit" and the attributes for various content types. However, the rest such as "customfiletypestart" and "Kubernetesfirst" is pretty confusing to me.

apinnick:
> Full ACK on "Cockpit" and the attributes for various content types. However, the rest such as "customfiletypestart" and "Kubernetesfirst" is pretty confusing to me.

@maximiliankolb  "customfiletypetitle" was unnecessary but I did find one occasion where the {customfiletype} was at the beginning of a sentence. I created {customfiletypestart} so that the first word would be capitalized (start = beginning of a sentence).
"Kubernetesfirst = Red Hat OpenShift Container Platform" for Satellite. In d/s docs, we are supposed to use the full name of the product the first time it is mentioned and the short form ("OpenShift Container Platform") after that. I did not bother using "Kubernetesfirst" in a procedure module where it refers to a cluster. However, the short form looked strange in the first sentence of the OpenShift Virtualization module. That is the only place where the new attribute is used.

adamlazik1:
I'd maybe suggest renaming `AttributeStart` to `AttributeCapitalized` for more clarity.

apinnick:
> I'd maybe suggest renaming `AttributeStart` to `AttributeCapitalized` for more clarity.

@adamlazik1 I'm not sure that would help because `AttributeCapitalized` could be interpreted to mean each word capitalized instead of first word capitalized. I will try to come up with a better name.

adamlazik1:
Another thing that comes to mind is `AttributeSentenceCase`, but that is probably too long.

apinnick:
What about `AttributeFirstCap`?

adamlazik1:
I would be ok with `AttributeFirstCap`, @maximiliankolb ?

maximiliankolb:
Works for me! I still think it's not perfect but I don't want to block your PR either if I cannot provide a better solution myself.

apinnick:
OK, let me just update the PR with a new commit

guides/common/modules/con_products-and-repositories.adoc
```diff
@@ -27,7 +27,7 @@ ifdef::orcharhino[]
 You can organize content in Products.
 Products bundle an arbitrary number of repositories.
 
-{customproducttitle}s require a subscription for hosts to access.
+{customproduct}s require a subscription for hosts to access.
```
maximiliankolb, line 30:
```suggestion
{customproductFirstCap}s require a subscription for hosts to access.
```

guides/common/attributes-satellite.adoc
```diff
@@ -48,20 +48,16 @@
...
 :Cockpit: RHEL web console
-:CockpitTitle: RHEL Web Console
 :customcontent: custom content
 :customfiletype: custom file type
-:customfiletypetitle: Custom File Type
-:customgpgtitle: Custom GPG
-:customproduct: custom Product
-:customproducttitle: Custom Product
+:customfiletypeFirstCap: Custom file type
+:customgpgtitle: custom GPG
```
maximiliankolb, line 54: Can we rename "customgpgtitle" to "customgpg"?

adamlazik1:
Merged and cherry-picked:
   835314c5f8..37d1af97db  3.10 -> 3.10
   6fc65b1a48..8fac59f5ab  3.9 -> 3.9

PR #2876, opened 2024-03-12
Implement ccutil as a make rule

ekohl:
This removes a lot of logic from the GitHub Action and reuses infrastructure we already have in place. It also allows running in parallel.

It does give up the checks for images and images/common, but those are rather uncommon given how uncommon it is to introduce new guides.

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -25,20 +25,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Compile html-single
-        run: |
-          for DIR in guides/doc-*; do
-            pushd $DIR &>/dev/null
-            [[ -d images/ ]] || (echo "Required images/ directory missing in $DIR"; exit 1)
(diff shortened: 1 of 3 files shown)


maximiliankolb (approved):
Looks reasonable to me! I cannot commit much on ccutil though.

ekohl:
I'll admit I didn't look at ccutil myself either. @adamlazik1 could you take a look?

adamlazik1:
Personally, I'd probably not remove the image checks. Even though images are rarely changed, I can see this potentially causing us headache in the future.

maximiliankolb:
Why do we run `ccutils` in the first place? What would we lose if we drop this?

ekohl:
> Why do we run `ccutils` in the first place? What would we lose if we drop this?

It's a tool that's used in Red Hat downstream to build, so it verifies that.

PR #2877, opened 2024-03-12
Run a single make command to build all guides

ekohl:
Since 4ca7a1ff4bbfefe7f02ba2ffe14671d39c4ad617 it's possible to run `make html` to build all guides in parallel but that only built upstream. Now there are two separate targets. The GitHub Action now uses this to simplify logic and increase parallelization.

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -69,11 +69,7 @@ jobs:
       - name: Build HTML with real links
         run: |
           make clean
-          make -j ${{ env.MAKE_J }} html BUILD=foreman-el
-          make -j ${{ env.MAKE_J }} html BUILD=foreman-deb
-          make -j ${{ env.MAKE_J }} html BUILD=katello
-          make -j ${{ env.MAKE_J }} html BUILD=satellite
-          make -j ${{ env.MAKE_J }} html BUILD=orcharhino
+          make -j ${{ env.MAKE_J }} html
 
       - name: Upload HTML
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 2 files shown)


maximiliankolb (commented):
Diff looks reasonable to me, but I could not reproduce this locally. Do you mind also implementing this for "guides/Makefile"? Or am I just calling the wrong Makefile?

I have tried this locally:

````
$ git fetch upstream pull/2877/head:pr_2877
$ podman build --no-cache --tag tmp_docs --file Dockerfile .
$ rm -fr guides/build && podman run --rm -v $(pwd):/foreman-documentation:Z tmp_docs make html-downstream
make: *** No rule to make target 'html-downstream'.  Stop.

# my default way to build locally:
$ rm -fr guides/build && podman run --rm -v $(pwd):/foreman-documentation:Z foreman_documentation make html BUILD=katello
````

ekohl:
> Diff looks reasonable to me, but I could not reproduce this locally. Do you mind also implementing this for "guides/Makefile"? Or am I just calling the wrong Makefile?

I fear it would be more complicated since today `guides/Makefile` doesn't touch any env vars. My immediate need is to simplify the GitHub Action and this achieves it.

> $ git fetch upstream pull/2877/head:pr_2877
> $ podman build --no-cache --tag tmp_docs --file Dockerfile .

I think you missed a `git checkout` there.

> $ rm -fr guides/build && podman run --rm -v $(pwd):/foreman-documentation:Z foreman_documentation make html BUILD=katello

I'd assume `make clean` would achieve the same as `rm -rf guides/build`, but the implementation tells me it doesn't guarantee the same.

maximiliankolb (changes requested):
This breaks my workflow, so I am not comfortable merging this as is.

maximiliankolb:
Discussed in Triage: closing this PR as it breaks the workflow of maintainers. We can discuss how to resolve this/come to a good solution.

PR #2880, opened 2024-03-12
Clarify remote SCAP resources

Lennonka:
The usage scenario for remote SCAP resources is a bit different that previously documented.
The ProjectServer does not have to be disconnected, what matters is whether the hosts have internet access or not.
I've tried to clarify that.

* [ ] Foreman 3.6/Katello 4.8


Lennonka:
I'm wondering if we should enhance it with an admonition. Opinions?

guides/common/modules/proc_applying-remote-scap-resources-in-a-disconnected-environment.adoc
```diff
@@ -2,10 +2,10 @@
 = Applying remote SCAP resources in a disconnected environment
 
 SCAP data streams can contain remote resources, such as OVAL files, that the SCAP client can fetch over the internet when it runs on hosts.
-If your host does not have internet access, you must download remote SCAP resources and distribute them to your host as local files.
+If your hosts do not have internet access, you must download remote SCAP resources and distribute them from {ProjectServer} to your hosts as local files.
```
maximiliankolb, line 5: I assume you can also distribute it via Smart Proxy Servers; just as any other synced content.

guides/common/modules/ref_inclusion-of-remote-scap-resources.adoc
```diff
@@ -16,7 +16,11 @@ which is referenced from datastream
 
 By default, the SCAP client is configured to ignore the remote resources and skip the XCCDF rules that rely on the resources.
 The skipped rules then result in the `notchecked` status.
-In a connected network environment, you can enable the download of remote resources on hosts in {Project}.
+
+For hosts with internet access, you can enable the download of remote resources on hosts in {Project}.
```
maximiliankolb, line 20: How would you do that/is this documented or maybe just the default for Ansible/Puppet? Or maybe reword to "...., your chosen OpenSCAP deployment method will automatically download SCAP content from the internet"?

maximiliankolb (commented):
two small suggestions/questions. I didn't immediately see the need for an admonition, so I lean towards "no, not needed".

guides/common/modules/proc_applying-remote-scap-resources-in-a-disconnected-environment.adoc
Lennonka: I'm not sure what you mean. You still have to add the content on ProjectServer first?

guides/common/modules/ref_inclusion-of-remote-scap-resources.adoc
Lennonka: That isn't accurate. It doesn't depend on the deployment method. The SCAP scanner (oscap) on the host downloads the remote SCAP content.

guides/common/modules/proc_applying-remote-scap-resources-in-a-disconnected-environment.adoc
maximiliankolb: Yes, you'd have to add content to {Project} first. Then, irrespective of whether hosts can access the internet, you distribute OpenSCAP content to your hosts similar to other content.

guides/common/modules/ref_inclusion-of-remote-scap-resources.adoc
maximiliankolb: OK, thanks for clarifying!

guides/common/modules/proc_applying-remote-scap-resources-in-a-disconnected-environment.adoc
Lennonka: Or does it? :D Ohhh, I think I'm getting an idea what you mean...

Lennonka: If I wanted to distribute it via SmartProxy, it would have to be added to a content view etc., right?

Lennonka: The latter is what is documented.

Lennonka: Actually, neither. When the host cannot access the internet, the file has to be distributed to the host BEFORE the SCAP client runs, because the SCAP client reads local files.

maximiliankolb: OK. Then I'm thinking: In the context of Foreman and centralized infrastructure management: wouldn't you _always_ want to distribute those files via Foreman Server/Smart Proxy Servers as you distribute other content too?

Lennonka: Yeah, probably. Fair point.

Lennonka:
Converted to draft, because it will probably be reworked.

The only concern I have about the new solution is that I don't know if I can get the file from a CV/AK to a particular directory. But maybe I don't know how at this point. I will explore.

Lennonka:
@maximiliankolb I've explored the option and I haven't found a way how to do it through CVs/AKs. If you can find a way, I don't mind changing the steps, but I don't think that there's functionality for it.

maximiliankolb (approved):
One last non-blocking suggestion: in "guides/common/modules/proc_applying-remote-scap-resources-in-a-disconnected-environment.adoc", link to https://docs.theforeman.org/nightly/Managing_Content/index-katello.html#Downloading_Files_to_a_Host_from_a_Custom_File_Type_Repository_content-management and/or https://docs.theforeman.org/nightly/Managing_Content/index-katello.html#Managing_Custom_File_Type_Content_content-management to guide users/explain by what you mean with "distribute files from Foreman to your hosts.

Lennonka:
@maximiliankolb Thank you. Isn't that clear enough from the procedure steps, which do contain the link to Managing Custom File Type Content? I think it's enough.

PR #2892, opened 2024-03-15
New procedure for Ansible job templates

bangelic:
A new procedure was created for using Ansible job
templates for package
and errata actions.

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_using-ansible-job-templates-for-package-and-errata-actions.adoc
+++ b/guides/common/modules/proc_using-ansible-job-templates-for-package-and-errata-actions.adoc
@@ -1,0 +1,17 @@
+[id="Using_Ansible_Job_Templates_for_Package_and_Errata_Actions{context}"]
+= Using Ansible job templates for package and errata actions
+
+By default, {Project} is configured to use the Script provider templates for remote execution jobs.
+If you prefer using the Ansible provider templates for your remote jobs, you can configure {Project} to use the Ansible provider templates for remote execution features that take a search query input.
+
+.Prerequisites
+Use remote execution in `ssh` mode.
+Pull mode does not work with Ansible.
+
+.Procedure
+. Navigate to *Administer > Remote Execution Features*.
+. Find each feature whose name contains `by_search`.
+. Change the job template for these features from `Katello Script Default` to `Katello Ansible Default`.
+. Click *Submit*.
+
+Ansible job templates are now used for remote execution when using the web UI to perform package and errata actions, and when using hammer job-invocation created with remote execution features that take a search query input.
--- a/guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
+++ b/guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
@@ -4,8 +4,10 @@
 By default, {Project} is configured to use the Script provider templates for remote execution jobs.
 If you prefer using Ansible job templates for your remote jobs, you can configure {Project} to use them by default for remote execution features associated with them.
 
-.Prerequisites
-* Use remote execution in `ssh` mode.
+[NOTE]
+====
+Remember that Ansible job templates only work when remote execution is configured for `ssh` mode.
+====
 
 .Procedure
 . In the {ProjectWebUI}, navigate to *Administer* > *Remote Execution Features*.
@@ -13,5 +15,5 @@ If you prefer using Ansible job templates for your remote jobs, you can configur
 . Change the job template for these features from `Katello Script Default` to `Katello Ansible Default`.
 . Click *Submit*.
 
-Ansible job templates are now used for remote execution when using the {ProjectWebUI} to perform package and errata actions.
-They are also used when using `hammer job-invocation create` with the same remote execution features used in the procedure above.
+{Project} now uses Ansible provider templates for remote execution jobs by which you can perform package and errata actions.
+This applies to job invocations from the {ProjectWebUI} as well as by using `hammer job-invocation create` with the same remote execution features that you have changed.
--- a/guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
+++ b/guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
@@ -1,15 +1,14 @@
-[id="Using_Ansible_Job_Templates_for_Package_and_Errata_Actions{context}"]
-= Using Ansible job templates for package and errata actions
+[id="Using_Ansible_Provider_for_Package_and_Errata_Actions_{context}"]
+= Using Ansible provider for package and errata actions
 
 By default, {Project} is configured to use the Script provider templates for remote execution jobs.
 If you prefer using the Ansible provider templates for your remote jobs, you can configure {Project} to use the Ansible provider templates for remote execution features that take a search query input.
 
 .Prerequisites
 Use remote execution in `ssh` mode.
-Pull mode does not work with Ansible.
 
 .Procedure
-. Navigate to *Administer > Remote Execution Features*.
+. In the {ProjectWebUI}, navigate to *Administer* > *Remote Execution Features*.
 . Find each feature whose name contains `by_search`.
 . Change the job template for these features from `Katello Script Default` to `Katello Ansible Default`.
 . Click *Submit*.
--- a/guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
+++ b/guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
@@ -13,5 +13,5 @@ If you prefer using Ansible job templates for your remote jobs, you can configur
 . Change the job template for these features from `Katello Script Default` to `Katello Ansible Default`.
 . Click *Submit*.
 
-Ansible job templates are now used for remote execution when using the web UI to perform package and errata actions.
+Ansible job templates are now used for remote execution when using the {ProjectWebUI} to perform package and errata actions.
 They are also used when using `hammer job-invocation create` with the same remote execution features used in the procedure above.
--- a/guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
+++ b/guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
@@ -2,10 +2,10 @@
 = Using Ansible provider for package and errata actions
 
 By default, {Project} is configured to use the Script provider templates for remote execution jobs.
-If you prefer using the Ansible provider templates for your remote jobs, you can configure {Project} to use the Ansible provider templates for remote execution features that take a search query input.
+If you prefer using Ansible job templates for your remote jobs, you can configure {Project} to use them by default for remote execution features associated with them.
 
 .Prerequisites
-Use remote execution in `ssh` mode.
+* Use remote execution in `ssh` mode.
 
 .Procedure
 . In the {ProjectWebUI}, navigate to *Administer* > *Remote Execution Features*.
@@ -13,4 +13,5 @@ Use remote execution in `ssh` mode.
 . Change the job template for these features from `Katello Script Default` to `Katello Ansible Default`.
 . Click *Submit*.
 
-Ansible job templates are now used for remote execution when using the web UI to perform package and errata actions, and when using hammer job-invocation created with remote execution features that take a search query input.
+Ansible job templates are now used for remote execution when using the web UI to perform package and errata actions.
+They are also used when using `hammer job-invocation create` with the same remote execution features used in the procedure above.
--- a/guides/common/assembly_configuring-and-setting-up-remote-jobs.adoc
+++ b/guides/common/assembly_configuring-and-setting-up-remote-jobs.adoc
@@ -75,4 +75,6 @@ include::modules/proc_scheduling-a-recurring-ansible-job-for-a-host-group.adoc[l
 
 include::modules/proc_monitoring-remote-jobs.adoc[leveloffset=+1]
 
+include::modules/proc_using-ansible-job-templates-for-package-and-errata-actions.adoc[leveloffset=+1]
+
 include::modules/proc_setting-the-job-rate-limit-on-smartproxy.adoc[leveloffset=+1]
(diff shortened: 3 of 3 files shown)


bangelic:
@Lennonka Still want to potentially change the title of this procedure.

guides/common/modules/proc_using-ansible-job-templates-for-package-and-errata-actions.adoc
```diff
@@ -0,0 +1,17 @@
+[id="Using_Ansible_Job_Templates_for_Package_and_Errata_Actions{context}"]
+= Using Ansible job templates for package and errata actions
```
Lennonka, line 2: Switching provider for REX jobs?

I'm not sure the actions are limited to packages and errata.
Can you try to find an umbrella term for the actions that you can change the provider for?

guides/common/modules/proc_using-ansible-job-templates-for-package-and-errata-actions.adoc
```diff
@@ -0,0 +1,17 @@
...
+
+By default, {Project} is configured to use the Script provider templates for remote execution jobs.
+If you prefer using the Ansible provider templates for your remote jobs, you can configure {Project} to use the Ansible provider templates for remote execution features that take a search query input.
+
+.Prerequisites
+Use remote execution in `ssh` mode.
+Pull mode does not work with Ansible.
```
Lennonka, line 9: List.

guides/common/modules/proc_using-ansible-job-templates-for-package-and-errata-actions.adoc
Lennonka, line 2: Perhaps something like "Switching provider for REX jobs"

I'm not sure the actions are limited to packages and errata.
Can you try to find an umbrella term for the actions that you can change the provider for?

guides/common/modules/proc_using-ansible-job-templates-for-package-and-errata-actions.adoc
Lennonka, line 9:
```suggestion
* Use remote execution in `ssh` mode.
Pull mode does not work with Ansible.
```
List :)

Lennonka (changes requested):

guides/common/modules/proc_using-ansible-job-templates-for-package-and-errata-actions.adoc
```diff
@@ -0,0 +1,17 @@
...
+
+.Procedure
+. Navigate to *Administer > Remote Execution Features*.
```
maximiliankolb, line 12:
```suggestion
. In the {ProjectWebUI}, navigate to *Administer* > *Remote Execution Features*.
```

guides/common/modules/proc_using-ansible-job-templates-for-package-and-errata-actions.adoc
```diff
@@ -0,0 +1,17 @@
+[id="Using_Ansible_Job_Templates_for_Package_and_Errata_Actions{context}"]
```
maximiliankolb, line 1:
```suggestion
[id="Using_Ansible_Job_Templates_for_Package_and_Errata_Actions"]
```

or

```suggestion
[id="Using_Ansible_Job_Templates_for_Package_and_Errata_Actions_{context}"]
```

AFAIK it would be valid to _no_ use the context attribute in the anchor if it's not necessary aka. if you don't intend to include this procudure twice in the same guide. Depends on whether this is part of "Managing Hosts" twice: once for REX, once for Ansible?

guides/common/modules/proc_using-ansible-job-templates-for-package-and-errata-actions.adoc
jeremylenz: cc @adamruzicka, make sure we use the "provider" term correctly if we go with that :)

adamruzicka: > something like "Switching provider for REX jobs"

It sounds a bit overly general. We have job templates and each job template is bound to a single provider. When you run a job you just select a single template, you don't really fiddle with providers.

Remote execution provides `Remote Execution Features` which are just shortcuts tying together a thing a user might want to do with a particular job template. What is being described here is how to attach a different job template to a particular rex feature. We also have much fewer rex features than job templates and I'm not sure if we can come up with a definition that would cover all rex features we have.

I would be fine with the current wording. If you really want to squeeze provider in there, then something like "Using ansible provider for package and errata actions" would work for me.

guides/common/modules/proc_using-ansible-job-templates-for-package-and-errata-actions.adoc
```diff
@@ -0,0 +1,17 @@
...
+= Using Ansible job templates for package and errata actions
+
+By default, {Project} is configured to use the Script provider templates for remote execution jobs.
+If you prefer using the Ansible provider templates for your remote jobs, you can configure {Project} to use the Ansible provider templates for remote execution features that take a search query input.
```
adamruzicka, line 5: This reads to me a little bit as if "{Project} will always use Script for *ALL* rex jobs. If I follow these steps it will use ansible for *ALL* jobs".

adamruzicka, line 5: With my pedantic hat on:
This reads as if this applies to "remote execution features that take a search query input", but features don't really take inputs, job templates do.

adamruzicka, line 5: I can create my own job template that will take a search query as an input and then this will no longer hold.

adamruzicka (commented):
On a more general note, don't we already have docs for "how to change a template attached to a rex feature" somewhere?

adamruzicka:
Ah, right, not in a generic way. This is the thing i remember seeing https://github.com/theforeman/foreman-documentation/blob/master/guides/common/modules/proc_upgrading-hosts-to-next-major-release.adoc?plain=1#L35

guides/common/modules/proc_using-ansible-job-templates-for-package-and-errata-actions.adoc
bangelic: Looks like it's just for Ansible. I will remove the context attribute.

guides/common/modules/proc_using-ansible-job-templates-for-package-and-errata-actions.adoc
bangelic: @jeremylenz Thoughts on how this should be written?

bangelic: @jeremylenz Thoughts on how this should be written?

guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
```diff
@@ -0,0 +1,16 @@
+[id="Using_Ansible_Provider_for_Package_and_Errata_Actions_{context}"]
+= Using Ansible provider for package and errata actions
+
+By default, {Project} is configured to use the Script provider templates for remote execution jobs.
+If you prefer using the Ansible provider templates for your remote jobs, you can configure {Project} to use the Ansible provider templates for remote execution features that take a search query input.
```
jeremylenz, line 5: Okaaay how about this?

```suggestion
If you prefer using Ansible job templates for your remote jobs, you can configure {Project} to use them by default for remote execution features associated with them.
```

I originally had "features that support them," but I think it's more accurate to say that a template "supports" a feature, rather than the other way around, since you specify the `feature: xxxx` in the template's YAML header.

guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
```diff
@@ -0,0 +1,16 @@
...
+.Prerequisites
+Use remote execution in `ssh` mode.
+
+.Procedure
+. In the {ProjectWebUI}, navigate to *Administer* > *Remote Execution Features*.
+. Find each feature whose name contains `by_search`.
+. Change the job template for these features from `Katello Script Default` to `Katello Ansible Default`.
+. Click *Submit*.
+
+Ansible job templates are now used for remote execution when using the web UI to perform package and errata actions, and when using hammer job-invocation created with remote execution features that take a search query input.
```
jeremylenz, line 16:
```suggestion
Ansible job templates are now used for remote execution when using the web UI to perform package and errata actions. They are also used when using "hammer job-invocation create" with the same remote execution features which you altered above.
```

jeremylenz: `hammer job-invocation create` is a Hammer command.

guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
```diff
@@ -0,0 +1,17 @@
...
+.Prerequisites
+* Use remote execution in `ssh` mode.
+
+.Procedure
+. In the {ProjectWebUI}, navigate to *Administer* > *Remote Execution Features*.
+. Find each feature whose name contains `by_search`.
+. Change the job template for these features from `Katello Script Default` to `Katello Ansible Default`.
+. Click *Submit*.
+
+Ansible job templates are now used for remote execution when using the web UI to perform package and errata actions.
```
maximiliankolb, line 16:
```suggestion
Ansible job templates are now used for remote execution when using the {ProjectWebUI} to perform package and errata actions.
```

maximiliankolb (approved):
Tested on Foreman 3.10 RC: LGTM.

I am unsure about the last two sentences. Maybe someone else can either ACK or make a suggestion.

guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
```diff
@@ -0,0 +1,17 @@
...
+* Use remote execution in `ssh` mode.
+
+.Procedure
+. In the {ProjectWebUI}, navigate to *Administer* > *Remote Execution Features*.
+. Find each feature whose name contains `by_search`.
+. Change the job template for these features from `Katello Script Default` to `Katello Ansible Default`.
+. Click *Submit*.
+
+Ansible job templates are now used for remote execution when using the {ProjectWebUI} to perform package and errata actions.
+They are also used when using `hammer job-invocation create` with the same remote execution features used in the procedure above.
```
Lennonka, line 17:
```suggestion
{Project} now uses Ansible provider templates for remote execution jobs by which you can perform package and errata actions.
This applies to job invocations from the {ProjectWebUI} as well as by using Hammer CLI, with the same remote execution features that you have changed.
```
Perhaps something like this?

guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
```diff
@@ -0,0 +1,17 @@
+[id="Using_Ansible_Provider_for_Package_and_Errata_Actions_{context}"]
+= Using Ansible provider for package and errata actions
+
+By default, {Project} is configured to use the Script provider templates for remote execution jobs.
+If you prefer using Ansible job templates for your remote jobs, you can configure {Project} to use them by default for remote execution features associated with them.
+
+.Prerequisites
+* Use remote execution in `ssh` mode.
```
Lennonka, line 8:
```suggestion
By default, {Project} is configured to use the Script provider templates for remote execution jobs.
If you prefer using Ansible job templates for your remote jobs, you can configure {Project} to use them by default for remote execution features associated with them.
This change can only apply to remote execution in `ssh` mode.
```
I don't think this is an actual prerequisite. But I'm not sure about the wording here.

Lennonka (changes requested):
A couple more suggestions.

guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
jeremylenz: Why avoid naming `hammer job-invocation create`? I guess it's fine but I feel like the original was easier to read. (I am biased, though.. ;)

guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
jeremylenz: How is prerequisite defined? If you haven't configured your smart proxy and hosts for `ssh` mode, that means your hosts will not have the SSH keys installed and will never receive the REX job.

Lennonka: Well, sure, but you don't need it in order to perform the following procedure, do you?

guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
Lennonka: I think that level of detail is not necessary in this place, but if @bangelic thinks it's better to include it, I'm not against it.

guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
jeremylenz: I suppose you could still set the default templates for those features, yes. But then things would stop working.

jeremylenz: How about just a Note, then?

```
Remember that Ansible job templates only work when remote execution is configured for `ssh` mode.`
```

Lennonka: That was the idea.

guides/common/modules/proc_using-ansible-provider-for-package-and-errata-actions.adoc
bangelic: I'll go ahead and keep `hammer job-invocation create` in there.

Lennonka (approved):
Looks good

Lennonka:
Cherry-picked:
- 5c72fce0b3..e1c487a782  3.10 -> 3.10
- 898693c62e..698bfed41b  3.9 -> 3.9

Issue #885, opened 2021-11-25
Switch to CC BY-SA 4.0


lzap:
Hello, before I pushed the content into github, I checked with Red Hat legal. Together with confirmation, I got a comment from Richard Fontana that we should consider upgrading the license. I totally forgot about this, so at least creating a ticket in case someone wants to jump into this.

Quote from Richard:

> 1.  I would recommend using this opportunity to modernize the
> documentation license (which should be the same for upstream and
> redhat.com). CC BY-SA 3.0 is kind of an outdated license at this point
> and there's an odd addendum in our legal notice that would not be
> needed if we switched to CC BY-SA 4.0 -- very similar to CC BY-SA 3.0
> but somewhat improved.

ekohl:
@apinnick has started to look at the licensing situation and will consider this.

apinnick:
I got the same message from Richard (I asked before I knew about this issue). I agree that we should upgrade. Oddly enough, the d/s RH docs still use CC BY-SA 3.0 and refer to the Shadowman logo. I wonder whether RH Legal plans to update the notice. I will create a PR to update the site license.

PR #2915, opened 2024-03-26
Update license

apinnick:
- Updated the site license to CC BY-SA 4.0.
- Updated only the year on the Red Hat license because d/s docs still refers to CC BY-SA 3.0. If this changes, I will update the RH license file.

Closes https://github.com/theforeman/foreman-documentation/issues/885

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,359 +1,428 @@
-Creative Commons Legal Code
+Attribution-ShareAlike 4.0 International
 
-Attribution-ShareAlike 3.0 Unported
+=======================================================================
 
-    CREATIVE COMMONS CORPORATION IS NOT A LAW FIRM AND DOES NOT PROVIDE
-    LEGAL SERVICES. DISTRIBUTION OF THIS LICENSE DOES NOT CREATE AN
-    ATTORNEY-CLIENT RELATIONSHIP. CREATIVE COMMONS PROVIDES THIS
-    INFORMATION ON AN "AS-IS" BASIS. CREATIVE COMMONS MAKES NO WARRANTIES
-    REGARDING THE INFORMATION PROVIDED, AND DISCLAIMS LIABILITY FOR
-    DAMAGES RESULTING FROM ITS USE.
+Creative Commons Corporation ("Creative Commons") is not a law firm and
+does not provide legal services or legal advice. Distribution of
+Creative Commons public licenses does not create a lawyer-client or
+other relationship. Creative Commons makes its licenses and related
+information available on an "as-is" basis. Creative Commons gives no
+warranties regarding its licenses, any material licensed under their
+terms and conditions, or any related information. Creative Commons
+disclaims all liability for damages resulting from their use to the
+fullest extent possible.
 
+Using Creative Commons Public Licenses
+
+Creative Commons public licenses provide a standard set of terms and
+conditions that creators and other rights holders may use to share
+original works of authorship and other material subject to copyright
+and certain other rights specified in the public license below. The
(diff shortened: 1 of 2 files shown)


ekohl (commented):
I had a look at the authors. First of all, I've taken the time to update the mailmap: https://github.com/theforeman/foreman-documentation/pull/2921

With that you can easily retrieve the various authors: `git log  --pretty='format:%aN <%aE>' web | sort -u`. That identifies b397c04886bcf1f417eaed3d217fa23d93019e18 as a commit by @lumarel while all others are done by Red Hatters or @maximiliankolb.

I think technically we should get approval from @lumarel to change the license.

As for the guides themselves, the list is a bit longer. Due to privacy I'll work with @apinnick offline to come up with a plan instead copy-pasting all names here.

lumarel (approved):
@ekohl it speaks very vividly for the project to even ask about such in my eyes small license changes!
Thank you for asking, and yes of course I approve 👍🏻

guides/LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright © 2019 Red Hat, Inc.
+Copyright © 2024 Red Hat, Inc.
 
 The text of and illustrations in this document are licensed by Red Hat under a Creative Commons Attribution–Share Alike 3.0 Unported license ("CC-BY-SA"). An explanation of CC-BY-SA is available at http://creativecommons.org/licenses/by-sa/3.0/. In accordance with CC-BY-SA, if you distribute this document or an adaptation of it, you must provide the URL for the original version.
```
maximiliankolb, line 3: This does not match the LICENSE file anymore. Is this intentional? Also, lot's of content is probably no longer copyright by Red Hat anymore in "guides/common/modules/" etc.

LICENSE
```diff
@@ -1,359 +1,428 @@
...
-    the terms of the License. This Section 4(a) applies to the Work as
-    incorporated in a Collection, but this does not require the Collection
-    apart from the Work itself to be made subject to the terms of this
-    License. If You create a Collection, upon notice from any Licensor You
-    must, to the extent practicable, remove from the Collection any credit
-    as required by Section 4(c), as requested. If You create an
-    Adaptation, upon notice from any Licensor You must, to the extent
-    practicable, remove from the Adaptation any credit as required by
-    Section 4(c), as requested.
- b. You may Distribute or Publicly Perform an Adaptation only under the
```
maximiliankolb, line 187: I think this is the crucial part: **I am not a lawyer**, but I interpret this as in "we can choose to update the version of the license from "CC BY-SA 3.0" to "CC BY-SA 4.0".

maximiliankolb (commented):
I am OK with changing the license from "CC BY-SA 3.0" to "CC BY-SA 4.0" for my contributions.

I would like to see more people review/ACK this change.

LICENSE
Lennonka: I'm not a lawyer either, but I'd interpret it the same way.

guides/LICENSE
apinnick: The current text reflects the license in our d/s docs. That's why I only changed the date. I have no idea whether it needs to be the same as the license text for the repo. :-(

LICENSE
apinnick: Yes, that aligns with what I was told by Legal.

Lennonka (approved):
LGTM

ekohl:
As discussed, we don't cherry pick this to stable branches.

PR #2918, opened 2024-03-28
Install python3-lxml for keycloak-httpd-client-install

adamlazik1:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2271398

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_registering-project-as-a-keycloak-client.adoc
+++ b/guides/common/modules/proc_registering-project-as-a-keycloak-client.adoc
@@ -14,11 +14,12 @@ You can also register two different {Project} clients to {Keycloak} if you want 
 
 .Procedure
 
-. On the {Project} server, install the following packages:
+// python3-lxml is only needed on EL8 because of https://issues.redhat.com/browse/RHEL-31496
+. On {ProjectServer}, install the following packages:
 +
 [options="nowrap", subs="verbatim,quotes,attributes"]
 ----
-# {project-package-install} mod_auth_openidc keycloak-httpd-client-install
+# {project-package-install} mod_auth_openidc keycloak-httpd-client-install python3-lxml
 ----
 
 . Register {Project} to {Keycloak} as a client.


maximiliankolb (approved):
LGTM; verified locally:

````
$ podman run --rm -it almalinux:8.9

# dnf makecache && dnf install python3-lxml
...
python3-lxml                                            x86_64                                    4.2.3-4.el8                                             appstream                                    1.5 M
````

**Review by Lennonka — APPROVED** [review]
I think you can merge it, Adam :)

**Review by adamruzicka — APPROVED** [review]

**Review by ekohl — COMMENTED** [review]
This is weird, because https://github.com/theforeman/foreman-packaging/blob/d9f2c1bedad625bfe8a7839b205a21de9918f150/packages/foreman/keycloak-httpd-client-install/keycloak-httpd-client-install.spec#L72 is there and has been since 2020.

**Review by ekohl — CHANGES_REQUESTED** [review]
In our package (version 1.2.2) it is required. The problem is that in RHEL base there's the same package (version 1.0) with the dependency missing.

It may be that the DNF module wasn't enabled, but that version may not be right anyway.

IMHO :-1: to documenting this and we need to figure out why the wrong version is being installed.

**Comment by ekohl** [comment]
After looking into the BZ I feel this is a Satellite-only bug because our specific package isn't used. I'd like to look into it a bit further, but perhaps we need to use ifdefs for Satellite.

**Comment by ekohl** [comment]
I did some more digging and summarized it here: https://bugzilla.redhat.com/show_bug.cgi?id=2271398#c6

Short summary: it's only an issue on RHEL 8 with Satellite, but documenting the workaround is the quickest fix. I would like to open a BZ against RHEL 8 so we can at least reference it in our docs. Essentially the RHEL equivalent of https://bugzilla.redhat.com/show_bug.cgi?id=2115262.

**Comment by ekohl** [comment]
Now filed as https://issues.redhat.com/browse/RHEL-31496

File: guides/common/modules/proc_registering-project-as-a-keycloak-client.adoc [hunk 1]
```diff
@@ -14,11 +14,12 @@ You can also register two different {Project} clients to {Keycloak} if you want
 
 .Procedure
 
+// python3-lxml is only needed because of https://issues.redhat.com/browse/RHEL-31496
```
ekohl, line 17: Let's make it clear that EL 9 is unaffected
```suggestion
// python3-lxml is only needed on EL8 because of https://issues.redhat.com/browse/RHEL-31496
```

guides/common/modules/proc_registering-project-as-a-keycloak-client.adoc
```diff
@@ -14,11 +14,12 @@ You can also register two different {Project} clients to {Keycloak} if you want
...
 . On the {Project} server, install the following packages:
```
ekohl, line 18: Existing issue, but we generally follow this style now:
```suggestion
. On {ProjectServer}, install the following packages:
```

adamlazik1:
Thank you, updated the PR.

ekohl (approved):

adamlazik1:
Merged and cherry-picked:

   dc76aed2de..0b42dac942  3.10 -> 3.10
   67dc3a0dda..7c95172008  3.3 -> 3.3
   e47bee6c90..001f032925  3.4 -> 3.4
   2725e1b1ed..f557396240  3.5 -> 3.5
   5f0104803b..993c21885c  3.6 -> 3.6
   901cef8dba..7d9cf7b1f2  3.7 -> 3.7
   3174cfe316..a7d0341b8d  3.8 -> 3.8
   5d8db84bc2..6818dd756a  3.9 -> 3.9

PR #2931, opened 2024-04-03
Hide fapolicyd sections on all Debian guides

ekohl:
In f187b7dfa0bb399ffd6d59c644b0dd48f769eb80 it was already hidden for the installing server guide, but this also hides it for other guides where it's included.

This came up while reviewing the Installing Foreman Proxy on Debian/Ubuntu with @apinnick.

I think it raises the question if this is the best way of hiding things. Would it make sense to have something like an assembly for this?

* [ ] Foreman 3.6/Katello 4.8


maximiliankolb (approved):
Thanks Ewoud, LGTM

ekohl:
8ccc7f2363f3..83c8eb89600f  3.10 -> 3.10
   bf21d97cf5b0..2a6f891e1a13  3.9 -> 3.9

PR #2936, opened 2024-04-03
Correct Debian package installation instructions

ekohl:
In fd8ed92fd762376cf0ae11e16b5e91b8280ec787 the package installation instructions were rewritten to use dnf, which isn't available. It now never uses the `client-package-*` attributes on servers, but instead relies on the `package-manager` attribute.

It also introduces an attribute for the NFS server package and uses the attribute for the bind utils package.

Fixes: fd8ed92fd762 ("Use attributes to manage pkgs on Hosts")

* [ ] Foreman 3.6/Katello 4.8


guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
```diff
@@ -22,11 +15,11 @@ If the `dhcp-no-override` setting is disabled, hosts fetch the bootloader and it
 include::snip_firewalld.adoc[]
 
 .Procedure
-. On your {EL} host, install the ISC DHCP Service and Berkeley Internet Name Domain (BIND) utilities packages:
+. On your external DHCP host, install the ISC DHCP Service and Berkeley Internet Name Domain (BIND) utilities packages:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
-# {client-package-install-el8} dhcp-server bind-utils
+# {package-manager} install {dhcp-server-package} {bind-package}
```
maximiliankolb, line 22: Where does "package-manager" come from? Do we only document/allow running external DHCP on the same base OS as Foreman Server?

I can create a follow-up PR to unset "package-manager" after using it in various procedures. If possible, it is defined within "attributes-*.adoc".

ekohl: > Do we only document/allow running external DHCP on the same base OS as Foreman Server?

No, but it's reasonable to expect that users use the same OS for their servers. Most users try to limit the different number of operating systems and if you install Foreman on Debian, you're most likely a Debian shop. Expecting them to install an external server on EL 8 is then unlikely.

> I think the issues comes from "we cannot use project-package-install because Satellite uses "satellite-maintain ...", and that's obviously not present on hosts that are not Satellite Server/Capsule Servers yet.

Yes, that's correct.

ekohl: I'm looking more at https://github.com/theforeman/foreman-documentation/commit/fd8ed92fd762376cf0ae11e16b5e91b8280ec787 and I'm sorry I didn't get to review it then, but I think it made things worse. IMHO we need attributes for the server package manager. I'll work on reintroducing those.

maximiliankolb: There are only a few select instances that actually require a third way to install packages next to client-pkg-install and project-pkg-install. We should use a meaningful third attribute for this where required for Satellite. All other build targets can probably go with "third-attribute: {project-pkg-install}".

ekohl:
This escalated a bit in size and it now removes more lines than it adds. Individual commits explain what's going on.

guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
ekohl: We also have `{server-package-manager} upgrade` in my current patch - would you prefer separate attributes for install/upgrade (we can get away with hardcoding `apt update` since it'll always be Debian-only because yum/dnf has no equivalent) or `{project-pkg-manager}` / `{project-package-manager}`?

maximiliankolb:
Frist thoughts:

* I think it's generally bad practice to set an attribute somewhere throughout the docs and never unset it. If it's relevant for all guides > guides/common/attributes*. If it's only relevant for a particular module, use "set, module, unset". (context: it was probably me who introduced this issue!)

* I propose the following patch to fix this:

````
diff --git a/guides/common/modules/proc_configuring-repositories.adoc b/guides/common/modules/proc_configuring-repositories.adoc
index f8cbdb460d..0552d1e8fb 100644
--- a/guides/common/modules/proc_configuring-repositories.adoc
+++ b/guides/common/modules/proc_configuring-repositories.adoc
@@ -51,6 +51,7 @@ endif::[]
 ifdef::foreman-el,katello[]
 include::proc_configuring-repositories-el.adoc[]
 endif::[]
+:!package-manager:
 
 include::snip_step-verify-enabled-repolist.adoc[]
 
@@ -85,6 +86,7 @@ ifdef::foreman-el,katello[]
 include::proc_configuring-repositories-el.adoc[]
 endif::[]
 
+:!package-manager:
 include::snip_step-verify-enabled-repolist.adoc[]
 
 ifdef::foreman-el,katello,satellite[]

````

* Alternatively, you could also replace "dnf" with "{package-manager}" in `guides/common/modules/snip_step-verify-enabled-repolist.adoc` and move unsetting the attribute two lines lower.
* If unsetting attributes scattered throughout any modules breaks the build, I consider this a good thing. Then we're forced to rethink using this specific attribute some more!

* `guides/common/modules/proc_installing-server-packages-el.adoc` appears to be unused. Is the PR missing an include? If not, please drop this file.

````
$ rg "proc_installing-server-packages"
$ rg "proc_installing-*-packages"
guides/common/assembly_managing-packages.adoc
8:include::modules/proc_installing-packages-on-a-host.adoc[leveloffset=+1]
````

* There's a duplicate attribute in "guides/common/attributes-satellite.adoc".

File: guides/common/attributes-satellite.adoc [hunk 2]
```diff
@@ -111,9 +111,11 @@
 :smart-proxy-context: capsule
 :smart-proxy-context-titlecase: Capsule
 :smart-proxy-principal: {smart-proxy-context}
+:smartproxy-installer-package: satellite-capsule
 :SmartProxies: Capsules
 :smartproxy_port: 9090
 :smartproxy-example-com: capsule.example.com
+:smartproxy-installer-package: satellite-capsule
```
maximiliankolb, line 118:
```suggestion
```

guides/common/attributes-satellite.adoc
```diff
@@ -93,6 +92,7 @@
 :PIV: CAC
 :project-context: satellite
 :project-change-hostname: satellite-change-hostname
+:project-installer-package: satellite
```
maximiliankolb, line 95:
```suggestion
:project-installer-package: {project-context}
```

guides/common/modules/proc_applying-errata-to-hosts-el.adoc
```diff
@@ -22,5 +22,5 @@ include::snip_applying-errata.adoc[]
 +
 [options="nowrap" subs="+quotes"]
 ----
-# dnf update _Module_Stream_Name_
+# dnf upgrade _Module_Stream_Name_
```
maximiliankolb, line 25:
```suggestion
# dnf upgrade _My_Module_Stream_
```

guides/common/modules/proc_configuring-repositories-proxy.adoc
```diff
@@ -1,7 +1,6 @@
 [id="configuring-repositories-proxy_{context}"]
 = Configuring repositories
 :dnf-module: satellite-capsule:el8
-:package-manager: dnf
```
maximiliankolb, line 4: I propose the following patch:

````
diff --git a/guides/common/assembly_installing-capsule-server.adoc b/guides/common/assembly_installing-capsule-server.adoc
index 575cfda8b2..d1c8685bdc 100644
--- a/guides/common/assembly_installing-capsule-server.adoc
+++ b/guides/common/assembly_installing-capsule-server.adoc
@@ -18,7 +18,11 @@ endif::[]
 
 // Configuring Repositories
 ifdef::satellite[]
+:parent-dnf-module: {dnf-module}
+:dnf-module: satellite-capsule:el8
 include::modules/proc_configuring-repositories-proxy.adoc[leveloffset=+1]
+:dnf-module: {parent-dnf-module}
+:!parent-dnf-module:
 endif::[]
 include::modules/proc_configuring-repositories.adoc[leveloffset=+1]
diff --git a/guides/common/modules/proc_configuring-repositories-proxy.adoc b/guides/common/modules/proc_configuring-repositories-proxy.adoc
index 6f00b45578..26c88b95bc 100644
--- a/guides/common/modules/proc_configuring-repositories-proxy.adoc
+++ b/guides/common/modules/proc_configuring-repositories-proxy.adoc
@@ -1,6 +1,5 @@
 [id="configuring-repositories-proxy_{context}"]
 = Configuring repositories
-:dnf-module: satellite-capsule:el8
 
 [NOTE]
 ====
````

File: guides/common/modules/proc_installing-proxy-packages.adoc [hunk 6]
```diff
@@ -1,37 +1,14 @@
...
-# dnf update
-endif::[]
-ifndef::satellite[]
-# {project-package-update}
-endif::[]
+# {server-package-manager} upgrade
 ----
-ifdef::satellite[]
-. Install the {ProjectServer} packages:
+. Install the packages:
```
maximiliankolb, line 9:
```suggestion
. Install the {SmartProxyServer} packages:
```

guides/common/modules/proc_installing-proxy-packages.adoc
maximiliankolb, line 1: This procedure should either be a snippet if used multiple times (it's not), or simply moved to a different file. See my patch:

````
diff --git a/guides/common/modules/proc_installing-capsule-server-packages.adoc b/guides/common/modules/proc_installing-capsule-server-packages.adoc
index 56585d94a4..4dc2c3425b 100644
--- a/guides/common/modules/proc_installing-capsule-server-packages.adoc
+++ b/guides/common/modules/proc_installing-capsule-server-packages.adoc
@@ -1,9 +1,18 @@
-[id="Installing_Proxy_Packages_{context}"]
+[id="installing-{smart-proxy-context}-packages_{context}"]
 = Installing {SmartProxyServer} packages
 
-Before installing {SmartProxyServer} packages, you must update all packages that are installed on the base operating system.
+Perform the following steps to install {SmartProxyServer}.
 
 .Procedure
-To install {SmartProxyServer}, complete the following steps:
-
-include::proc_installing-proxy-packages.adoc[]
+. Update all packages:
++
+----
+# {server-package-manager} upgrade
+----
+. Install the {SmartProxyServer} packages:
++
+----
+# {server-package-manager} install {smartproxy-installer-package}
+----
````

Then, run `$ git rm guides/common/modules/proc_installing-proxy-packages.adoc`.

File: guides/common/modules/proc_installing-server-packages-el.adoc
Comment (maximiliankolb) [inline guides/common/modules/proc_installing-server-packages-el.adoc:1]: This is also missing a heading or converted into a snippet.

**Review by maximiliankolb — CHANGES_REQUESTED** [review]
I had a look at the diff. I think not using "client" attributes on the "to be" Foreman Server host is generally a good idea. Thanks for implementing this.

File: guides/common/modules/proc_configuring-an-external-dhcp-server.adoc (hunk 1 above)
Comment (maximiliankolb) [reply guides/common/modules/proc_configuring-an-external-dhcp-server.adoc]: I don't yet have a better alternative to "server-package-manager", so I think it's fine.

**Comment by ekohl** [comment]
> * I think it's generally bad practice to set an attribute somewhere throughout the docs and never unset it. If it's relevant for all guides > guides/common/attributes*. If it's only relevant for a particular module, use "set, module, unset". (context: it was probably me who introduced this issue!)

Agreed, and I've tried to reduce that use here.

> * `guides/common/modules/proc_installing-server-packages-el.adoc` appears to be unused. Is the PR missing an include? If not, please drop this file.

I was sure I had a `git rm` somewhere along the way, but clearly it didn't end up in the PR.

File: guides/common/attributes-satellite.adoc (hunk 2 above)
Comment (ekohl) [reply guides/common/attributes-satellite.adoc:118]: I noticed this too when I showed it during the meeting.

File: guides/common/attributes-satellite.adoc (hunk 3 above)
Comment (ekohl) [reply guides/common/attributes-satellite.adoc:95]: I disagree here. It's very specific that package name and it's not like something else inherits from the satellite attributes.

I'll also note we (installer team) are rethinking the packaging so that the end result is that you only have a single scenario present (think `foreman-installer-$scenario` packages). That way you never have to specify `--scenario` to the installer (just like in vanilla Foreman). These attributes will make the transition easier in the future

File: guides/common/modules/proc_installing-proxy-packages.adoc
Comment (ekohl) [reply guides/common/modules/proc_installing-proxy-packages.adoc:1]: I had something like that initially, but wasn't quite sure if this was the right approach.

File: guides/common/assembly_installing-capsule-server.adoc [hunk 7]
```diff
@@ -32,7 +32,7 @@ include::modules/proc_installing-fapolicyd-on-server.adoc[leveloffset=+2]
 endif::[]
 
 // Installing {SmartProxyServer} Packages
-include::modules/proc_installing-capsule-server-packages.adoc[leveloffset=+1]
+include::modules/proc_installing-proxy-server-packages.adoc[leveloffset=+1]
```
maximiliankolb, line 35:
```suggestion
include::modules/proc_installing-smart-proxy-server-packages.adoc[leveloffset=+1]
```

guides/common/modules/proc_installing-proxy-server-packages.adoc
maximiliankolb, line 1: please rename as commented above to "guides/common/modules/proc_installing-smart-proxy-server-packages.adoc"

guides/common/modules/proc_installing-proxy-server-packages.adoc
```diff
@@ -0,0 +1,19 @@
+[id="Installing_Proxy_Packages_{context}"]
```
maximiliankolb, line 1:
```suggestion
[id="installing-{smart-proxy-context}-packages_{context}"]
```

guides/common/modules/proc_installing-proxy-server-packages.adoc
```diff
@@ -0,0 +1,19 @@
...
+= Installing {SmartProxyServer} packages
+
+Perform the following steps to install {SmartProxyServer}.
+
+.Procedure
+
```
maximiliankolb, line 7:
```suggestion
```

guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
maximiliankolb: From triage: let's consider the name of the attribute non-blocking for your PR.

ekohl:
Some work has already been merged. I've split the remaining work into separate PRs that can be individually reviewed:
* https://github.com/theforeman/foreman-documentation/pull/4002
* https://github.com/theforeman/foreman-documentation/pull/4003
* https://github.com/theforeman/foreman-documentation/pull/4004

PR #2947, opened 2024-04-08
Readd assembly that got dropped

maximiliankolb:
Fixes #2945


asteflova (approved):
Ack from me, but let's wait for @bangelic. Brian, can you confirm this assembly got removed by accident along with the SCA references?

bangelic:
Yes it was removed accidentally because it was failing one of my builds. I swear I didn't remove it. I wasn't sure why it was gone.

PR #2950, opened 2024-04-08
Rework glossary

maximiliankolb:
Please review commit by commit and focus on the diff. I know that there are more questionable entries. I will create a follow-up PR to fix some more.

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.10/Katello 4.12
* [x] Foreman 3.9/Katello 4.11 (planned Satellite 6.15)

PR #2959, opened 2024-04-11
Remove deprecated plugin foreman_hooks

maximiliankolb:
foreman_hooks is deprecated and will be removed with Foreman 3.11. See https://github.com/theforeman/foreman_hooks

This PR is a response to https://github.com/theforeman/foreman-documentation/pull/2950#discussion_r1560682095

### DIFF
--- a/guides/common/modules/proc_upgrading-a-disconnected-project-server.adoc
+++ b/guides/common/modules/proc_upgrading-a-disconnected-project-server.adoc
@@ -27,8 +27,6 @@ endif::[]
 * Review and update your firewall configuration before upgrading your {ProjectServer}.
 For more information, see {InstallingServerDisconnectedDocURL}Port_and_firewall_requirements_satellite[Port and firewall requirements] in _{InstallingServerDisconnectedDocTitle}_.
 * Ensure that you do not delete the manifest from the Customer Portal or in the {ProjectWebUI} because this removes all the entitlements of your content hosts.
-* Back up and remove all Foreman hooks before upgrading.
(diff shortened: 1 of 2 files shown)


maximiliankolb:
Hey @adamlazik1 Do you know more about the deprecation of the foreman_hooks plugin with Foreman 3.11? Is this official yet? Do you have any BZ? If we decide to drop the docs, we should get at least one ACK from a Foreman developer. Do you know _who_ to ping?

adamlazik1:
All of the webhook and shellhook functionality should remain unchanged as foreman_hooks is being replaced with foreman_webhooks. We actually discussed this today with @adamruzicka so I believe it is safe to keep the documentation.

adamlazik1:
Maybe we could drop the migration procedure, I will confirm that.

adamlazik1:
Okay so we should also keep the migration chapter for people who upgrade from 3.10 and wonder why it suddenly stopped working. We can remove the migration chapter in 3.12. Only things that should be removed in this PR is the mention from glossary and the sentence about backing up foreman hooks.

ekohl:
The removal also deserves a release note. @adamruzicka

adamlazik1:
@ekohl if you are talking about a downstream release note, we have that completed.

ekohl:
@adamlazik1 I was thinking about upstream.

adamlazik1:
@ekohl in that case, should I create a PR to add this RN into upstream manually?

adamruzicka (approved):
This looks like the right amount of changes

Lennonka (approved):
LGTM

PR #2965, opened 2024-04-15
Add missing procedure to assembly

maximiliankolb:
Previously, the procedure was only part of the "Administering Hosts" assembly.

the workaround for the context is necessary because the guide contains boths assemblies but the anchors cannot both use the same context.

* [x] Foreman 3.9/Katello 4.11 (planned Satellite 6.15; orcharhino 6.8)

### DIFF
--- a/guides/common/assembly_administering-hosts.adoc
+++ b/guides/common/assembly_administering-hosts.adoc
@@ -17,8 +17,6 @@ include::modules/proc_changing-the-module-stream-for-a-host.adoc[leveloffset=+1]
 include::modules/proc_enabling-custom-repositories-on-content-hosts.adoc[leveloffset=+1]
 
 ifdef::katello,satellite,orcharhino[]
-include::modules/proc_adding-a-host-to-a-host-collection.adoc[leveloffset=+1]
-
 include::modules/proc_using-the-host-details-tab.adoc[leveloffset=+1]
 
 include::modules/proc_changing-the-content-source-of-a-host.adoc[leveloffset=+1]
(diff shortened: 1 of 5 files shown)


maximiliankolb:
Rebased to "master" and simplified. Do you mind having a look @asteflova ?

asteflova:
Everything looks ok, but as an alternative, how about _moving_ the procedure from assembly_administering-hosts.adoc to assembly_configuring-host-collections.adoc?

assembly_administering-hosts.adoc is really long. Too long for an assembly, IMO. A while ago, I moved all host group-related modules to a separate assembly in https://github.com/theforeman/foreman-documentation/pull/2685 in an effort to make assembly_administering-hosts.adoc shorter and easier to take in. Moving proc_adding-a-host-to-a-host-collection.adoc could help too. But it's your call.

guides/common/assembly_administering-hosts.adoc
```diff
@@ -17,8 +17,6 @@ include::modules/proc_changing-the-module-stream-for-a-host.adoc[leveloffset=+1]
 include::modules/proc_enabling-custom-repositories-on-content-hosts.adoc[leveloffset=+1]
 
 ifdef::katello,satellite,orcharhino[]
```
asteflova, line 19: @maximiliankolb Do you know why the procedure was included within an ifdef/endif here? That's the only concern I have a this point -- you're not using a conditional in guides/common/assembly_configuring-host-collections.adoc and I don't see it in doc-Managing_Hosts/master.adoc either. I don't think there is a reason to use it but I still want to check that it's okay.

maximiliankolb: Yes, looks like it's indeed a Katello feature.

maximiliankolb:
   b366b61e3f..573bbca336  3.10 -> 3.10
   c535864ba5..f81c02e834  3.9 -> 3.9

PR #2985, opened 2024-04-29
Add a warning about deadlocks in sync

AkshayGadhaveRH:
For Satellite 6.12, users face deadlock errors with Smart Proxy sync post upgrade. This is due to a bug which has been resolved in v6.13. Adding a warning for Satellite users who upgrade to v6.12.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2272673

* [ ] Foreman 3.6/Katello 4.8


guides/doc-Upgrading_and_Updating/master.adoc
```diff
@@ -106,6 +106,11 @@ ifdef::katello,orcharhino,satellite[]
 include::common/modules/proc_tuning-with-predefined-profiles.adoc[leveloffset=+3]
 endif::[]
 
+ifdef::satellite[]
+//SmartProxy Sync Deadlock Warning
+include::common/modules/snip_deadlock_warning.adoc[leveloffset=+3]
+endif::[]
+
```
asteflova, line 113: You're including this as a snippet in master.adoc. Does this mean it's not part of proc_tuning-with-predefined-profiles.adoc? Because right now, it looks like users are supposed to run the commands as the last step of the "Tuning Satellite Server with Predefined Profiles" procedure. But that seems unlikely to me.

When you write "To avoid this, ensure that no Capsule synchronization tasks are in progress", when exactly should users ensure this? Is there an existing procedure into which you could add this as a regular procedure step? For information like this, placement is crucial -- we should make sure users read this advice exactly at the time when they need it.

AkshayGadhaveRH: So the procedure needs to be performed post-upgrade, that is the only restriction. Since, the subsections in post-upgrade are optional depending on the setup the user has, I have moved the warning to the top. Does this work?

asteflova: Yes, this works. (Normally, I would consider adding this as [a troubleshooting section](https://redhat-documentation.github.io/modular-docs/#troubleshooting) to an existing procedure. But I can see that this problem exists only in the earliest supported version... so let's skip the trouble of major-ish restructuring :)) Thanks!

asteflova:
Can you please include the Warning directly into con_performing-post-upgrade-tasks.adoc, Akshay? I don't think there is a reason to create a separate snippet file if you aren't reusing the warning elsewhere.

guides/common/modules/snip_deadlock_warning.adoc
```diff
@@ -0,0 +1,11 @@
+[WARNING]
+====
+{SmartProxy} synchronization might fail with deadlock after upgrading the {Project}.
+To avoid this, ensure that no {SmartProxy} synchronization tasks are in progress and then stop all workers on the {SmartProxy} except one:
```
asteflova, line 4:
```suggestion
To avoid this, ensure that no {SmartProxy} synchronization tasks are in progress and then stop all workers on the {SmartProxy} except one after the upgrade:
```
or probably even better:
```suggestion
After the upgrade, ensure that no {SmartProxy} synchronization tasks are in progress and then stop all workers on the {SmartProxy} except one:
```
I just think it would be useful to really drive home the point that users must do this after the upgrade. (I know that the whole section is called 'post-upgrade tasks', but still :shrug:)

AkshayGadhaveRH:
Thanks @asteflova committed the changes you suggested.

asteflova (approved):

PR #2970, opened 2024-04-17
Refs #37334 - Add keycloak-httpd-client-install EL9 removal warning

ekohl:
This package has been dropped from Foreman's packaging and users need to take manual action.

I think a manual pick to 3.10 is best. There we should mention that it will be dropped with Foreman 3.10.1.

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.10/Katello 4.12
* [ ] Foreman 3.9/Katello 4.11 (planned Satellite 6.15; orcharhino 6.8)
* [ ] Foreman 3.8/Katello 4.10
* [ ] Foreman 3.7/Katello 4.9 (Satellite 6.14)
* [ ] Foreman 3.6/Katello 4.8
* [ ] Foreman 3.5/Katello 4.7 (Satellite 6.13; orcharhino 6.6/6.7)
* [ ] Foreman 3.4/Katello 4.6 (EL8 only)
* [ ] Foreman 3.3/Katello 4.5 on EL7 & EL8 (Satellite 6.12 on EL8 only; orcharhino 6.4/6.5 on EL8 only)
* We do not accept PRs for Foreman older than 3.3.

PR #2992, opened 2024-05-02
Revert "Refs #37334 - Add keycloak-httpd-client-install EL9 removal warning (#2970)"

maximiliankolb:
This reverts commit 5ca070d0f723f3070e062366dca9d91fb762874e.

According to #37334, this has been fixed for Foreman 3.11. Because we will branch docs for Foreman 3.11 based on "master", we have to revert this.

**I have merged this by mistake to "master".**

### DIFF
--- a/guides/doc-Release_Notes/topics/foreman.adoc
+++ b/guides/doc-Release_Notes/topics/foreman.adoc
@@ -19,17 +19,6 @@ Users are required to change the DNF module, but the actual upgrade, including d
 During the upgrade a backup of the data is created in `/var/lib/pgsql/data-old`.
 This backup can be removed once the upgrade is completed.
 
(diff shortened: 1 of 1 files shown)


Lennonka (approved):

ekohl:
It was picked to 3.10 as well iirc. And master is 3.11 now

ekohl:
So to be clear, it was intended to be merged to master and also picked to 3.10

maximiliankolb:
Hey @ekohl , was reverting the commit on "master" also a mistake? I thought it is only required for "3.10" because "master" aka "3.11 to be" has a fix according to the linked ticket?

current state: it's only on 3.10; not on "master". Please LMK if I need to fix things again.

ekohl:
I'd like to have it in both. Technically it will be changed in 3.10.1 and 3.11.0

PR #3000, opened 2024-05-03
Hide Provisioning sections unsupported in Satellite

Lennonka:
Hide sections in Provisioning that are not supported in Satellite.

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/doc-Provisioning_Hosts/master.adoc
+++ b/guides/doc-Provisioning_Hosts/master.adoc
@@ -26,7 +26,9 @@ include::common/assembly_using-infoblox-as-dhcp-and-dns-providers.adoc[leveloffs
 
 include::common/assembly_using-pxe-to-provision-hosts.adoc[leveloffset=+1]
 
+ifndef::satellite[]
 include::common/assembly_using-ipxe-to-reduce-provisioning-times.adoc[leveloffset=+1]
+endif::[]
 
 include::common/assembly_configuring-the-discovery-service.adoc[leveloffset=+1]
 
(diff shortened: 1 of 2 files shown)


maximiliankolb (approved):
diff LGTM

guides/doc-Provisioning_Hosts/master.adoc
```diff
@@ -26,7 +26,9 @@ include::common/assembly_using-infoblox-as-dhcp-and-dns-providers.adoc[leveloffs
 
 include::common/assembly_using-pxe-to-provision-hosts.adoc[leveloffset=+1]
 
+ifndef::satellite[]
```
apinnick, line 29: I'm not sure that this is correct. Is there a connection between booting with installation media and iPXE boot?

ekohl:
Note I found the installation media because the application linked to it.  It may be good to keep the chapter in, but mention that installation media aren't used when Katello is present. That would also apply to upstream.

And I'd like @jeremylenz and/or @ianballou to have a look.

guides/doc-Provisioning_Hosts/master.adoc
Lennonka: The connection was supportability, but let's discuss on Slack

Lennonka:
Then I can close this PR, because the note is already present:

```
ifdef::foreman-el,katello[]
When you install the Katello plugin, you can download packages from a Pulp mirror.
In this case, installation media are ignored.
endif::[]
ifdef::satellite[]
You can use this parameter to install third-party content.
Red Hat content is delivered through repository syncing instead.
endif::[]
```

cc @ekohl

Lennonka:
Re-open if needed.

PR #3013, opened 2024-05-10
Modularize assembly

maximiliankolb:


guides/common/modules/con_performing-additional-configuration-on-smart-proxy-server.adoc
```diff
@@ -0,0 +1,4 @@
+[id="performing-additional-configuration-on-{smart-proxy-context}-server_{context}"]
+= Performing additional configuration on {SmartProxyServer}
+
+Use this chapter to configure additional settings on your {SmartProxyServer}.
```
asteflova, line 4: Referring to parts of documentation as chapter/section/procedure is tricky because users might not understand these terms in the same way. Also, the words "Use this chapter" don't really provide any value.

What [we're recommended to do](https://redhat-documentation.github.io/supplementary-style-guide/#shortdesc) is to avoid talking about docs ("This chapter covers/Use this chapter/etc.") and instead go straight to explaining the what and the why with a focus on the user.

An easy fix would be:

```suggestion
You can configure additional settings on your {SmartProxyServer}.
```

but that just copies what's already in the chapter title. So how about something that provides a bit of context ("After installing") and sums up what can be configured and/or why:

```suggestion
After installing your {SmartProxyServer}, you can configure additional settings, such as [...].
```
or
```suggestion
After installing your {SmartProxyServer}, you can configure additional settings in order to [...].
```

asteflova (approved):
I'd like to see the chapter introduction improved, but I also realize that you're just using the wording that's already in. So feel free to skip my feedback if you consider it out of scope of your PR.

maximiliankolb:
   4e7ac30894..2891a0cab5  3.10 -> 3.10
   e8eae33870..efb270f9d6  3.9 -> 3.9

PR #3019, opened 2024-05-14
Unify URL for pulp content on Foreman/Katello

maximiliankolb:
Pulp serves content under "FQDN/pulp/content". Historically, it was (also) "FQDN/pulp/repos". Looking at "git grep", this seems to be the only outlier.


quba42:
I don't know the full background as to why content is shown under multiple different base paths. AFAIK all the various base paths serve the exact same content so it does not matter which of them one actually uses. The `/pulp/content/` base path is used for the GUI URL's on the repository page (and is also the Pulp default), so that is the path users are most likely to know.

Long story short: I support the change proposed in this PR.

PR #3036, opened 2024-05-18
Optimize the diff text when there's no change

ekohl:
If a PR doens't generate a diff (like this change) then the message shouldn't link to any diffs. By making this message explicit it makes reviewing a bit easier.

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -68,8 +68,10 @@ jobs:
       - name: Create diff to old base
         run: |
           diff -Nrwu old/ preview/${{ fromJSON(env.PR_DATA).target_name }}/ | cat > preview/diff.patch
-          pygmentize -o preview/diff.html -l diff -f html -O full preview/diff.patch
-          diffstat -l -p2 preview/diff.patch > diff.txt
+          if [[ -s preview/diff.patch ]] ; then
+            pygmentize -o preview/diff.html -l diff -f html -O full preview/diff.patch
+            diffstat -l -p2 preview/diff.patch > diff.txt
+          fi
 
       - name: Deploy to surge.sh
         run: ./node_modules/.bin/surge ./preview $PREVIEW_DOMAIN --token ${{ secrets.SURGE_TOKEN }}
@@ -78,12 +80,23 @@ jobs:
         run: |
           echo "The PR preview for ${{ fromJSON(env.PR_DATA).head_sha }} is available at [${{ env.PREVIEW_DOMAIN }}](https://${{ env.PREVIEW_DOMAIN }})" >> pr.md
           echo "" >> pr.md
-          echo "The following output files are affected by this PR:" >> pr.md
-          sed -E "s#(.*)#- [\1](https://${{ env.PREVIEW_DOMAIN }}/${{ fromJSON(env.PR_DATA).target_name }}/\1)#" diff.txt >> pr.md
-          echo "" >> pr.md
-          echo '[show diff](https://${{ env.PREVIEW_DOMAIN }}/diff.patch)' >> pr.md
-          echo "" >> pr.md
-          echo '[show diff as HTML](https://${{ env.PREVIEW_DOMAIN }}/diff.html)' >> pr.md
+
+          if [[ -f diff.txt ]] ; then
+            echo "The following output files are affected by this PR:" >> pr.md
+            sed -E "s#(.*)#- [\1](https://${{ env.PREVIEW_DOMAIN }}/${{ fromJSON(env.PR_DATA).target_name }}/\1)#" diff.txt >> pr.md
+          else
+            echo "No diff compared to the current base" >> pr.md
+          fi
+
+          if [[ -f preview/diff.patch ]] ; then
+            echo "" >> pr.md
+            echo '[show diff](https://${{ env.PREVIEW_DOMAIN }}/diff.patch)' >> pr.md
+          fi
+
+          if [[ -f preview/diff.html ]] ; then
+            echo "" >> pr.md
+            echo '[show diff as HTML](https://${{ env.PREVIEW_DOMAIN }}/diff.html)' >> pr.md
+          fi
 
       - name: Comment on PR
         uses: marocchino/sticky-pull-request-comment@v2


ekohl:
> The following output files are affected by this PR:
> 
> [show diff](https://theforeman-foreman-documentation-preview-pr-3036.surge.sh/diff.patch)
> 
> [show diff as HTML](https://theforeman-foreman-documentation-preview-pr-3036.surge.sh/diff.html)

This suggests it's not working as intended, but the workflow runs after `deploy` finished and uses the defintiion from core rather than the PR.

.github/workflows/preview.yml
```diff
@@ -96,12 +98,23 @@ jobs:
...
-          echo '[show diff as HTML](https://${{ env.PREVIEW_DOMAIN }}/diff.html)' >> pr.md
+
+          if [[ -f diff.txt ]] ; then
+            echo "The following output files are affected by this PR:" >> pr.md
+            sed -E "s#(.*)#- [\1](https://${{ env.PREVIEW_DOMAIN }}/${{ fromJSON(env.PR_DATA).target_name }}/\1)#" diff.txt >> pr.md
+          else
+            echo "No diff compared to the current base" >> pr.md
+          fi
+
+          if [[ -f diff.patch ]] ; theni
```
Lennonka, line 109:
```suggestion
          if [[ -f diff.patch ]] ; then
```

.github/workflows/preview.yml
```diff
@@ -96,12 +98,23 @@ jobs:
...
+            echo "" >> pr.md
+            echo '[show diff](https://${{ env.PREVIEW_DOMAIN }}/diff.patch)' >> pr.md
+          fi
+
+          if [[ -f diff.html ]] ; theni
```
Lennonka, line 114:
```suggestion
          if [[ -f diff.html ]] ; then
```

.github/workflows/preview.yml
```diff
@@ -96,12 +98,23 @@ jobs:
...
           echo "The PR preview for ${{ fromJSON(env.PR_DATA).head_sha }} is available at [${{ env.PREVIEW_DOMAIN }}](https://${{ env.PREVIEW_DOMAIN }})" >> pr.md
           echo "" >> pr.md
-          echo "The following output files are affected by this PR:" >> pr.md
-          sed -E "s#(.*)#- [\1](https://${{ env.PREVIEW_DOMAIN }}/${{ fromJSON(env.PR_DATA).target_name }}/\1)#" diff.txt >> pr.md
-          echo "" >> pr.md
-          echo '[show diff](https://${{ env.PREVIEW_DOMAIN }}/diff.patch)' >> pr.md
-          echo "" >> pr.md
-          echo '[show diff as HTML](https://${{ env.PREVIEW_DOMAIN }}/diff.html)' >> pr.md
+
+          if [[ -f diff.txt ]] ; then
```
Lennonka, line 102:
```suggestion
          if [ -f diff.txt ] ; then
```
AFAIK, `[[ ]]` is for conditional command execution, whereas `[ ]` is `test`, which is probably what you're trying to do.
If you're using `[[ ]]` with an intent, can you please explain why is it better here than `[ ]` (`test`)?

Lennonka (changes requested):
Lovely idea. Not sure about the implementation, though :)

ekohl:
When testing the implementation appeared to have an issue, so I need to dive into it a bit more. Though perhaps that was in the other repo I was playing with

maximiliankolb:
triage: @ekohl friendly reminder

.github/workflows/preview.yml
ekohl: `[[` is a bash version of `[`. https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html explains some of the differences. In this particular case it probably doesn't matter but it absolutely can:
```sh
#!/bin/bash

if [ -n $DOESNOTEXIST ] ; then
	echo "plain: The length of '$DOESNOTEXIST' is nonzero"
else
	echo "plain: The length of '$DOESNOTEXIST' is zero"
fi

if [[ -n $DOESNOTEXIST ]] ; then
	echo "bash: The length of '$DOESNOTEXIST' is nonzero"
else
	echo "bash: The length of '$DOESNOTEXIST' is zero"
fi

VALUEWITHSPACES="this has spaces"
if [ -n $VALUEWITHSPACES ] ; then
	echo "plain: The length of '$VALUEWITHSPACES' is nonzero"
else
	echo "plain: The length of '$VALUEWITHSPACES' is zero"
fi

if [[ -n $VALUEWITHSPACES ]] ; then
	echo "bash: The length of '$VALUEWITHSPACES' is nonzero"
else
	echo "bash: The length of '$VALUEWITHSPACES' is zero"
fi
```

When you run it:
```console
$ bash example.sh
plain: The length of '' is nonzero
bash: The length of '' is zero
example.sh: line 16: [: too many arguments
plain: The length of 'this has spaces' is zero
bash: The length of 'this has spaces' is nonzero
```

You can see the plain check does not what you expect it to do and actually crashes at some point. Now there are POSIX shell compliant options to make it work and [shellcheck](https://www.shellcheck.net/) will loudly yell at you about these, but in general if you can assume bash it's easier to use `[[`.

ekohl:
I've fixed the `theni` and also modified some paths, but as we've just seen: this thing is hard to test.

maximiliankolb (approved):
LGTM

.github/workflows/preview.yml
Lennonka: Thank you.

Lennonka (approved):

.github/workflows/preview.yml
```diff
@@ -68,8 +68,10 @@ jobs:
       - name: Create diff to old base
         run: |
           diff -Nrwu old/ preview/${{ fromJSON(env.PR_DATA).target_name }}/ | cat > preview/diff.patch
-          pygmentize -o preview/diff.html -l diff -f html -O full preview/diff.patch
-          diffstat -l -p2 preview/diff.patch > diff.txt
+          if [[ -s preview/diff.patch ]] ; then
```
ekohl, line 71: Potential performance improvement: only install `diffstat python3-pygments` if there is a diff.

maximiliankolb:
triage: Kindly asking @evgeni for a tech ACK.

evgeni (approved):

PR #3044, opened 2024-05-23
Introduce different attributes for RHSSO and RHBK

asteflova:
Wildfly-based Keycloak is Red Hat Single Sign-on. Quarkus-based Keycloak is Red Hat Build of Keycloak. This PR proposes to reflect that by making the following changes:

* Introducing two new Keycloak-related attributes: `keycloak-wildfly` and `keycloak-quarkus`.
* Removing the `keycloak-short` attribute and replacing it with `keycloak-example-com` and `keycloak-realm` attributes to be used as example values in commands.
* Changing the `keycloak` attribute to resolve to "Keycloak" in non-Satellite builds and to "RHBK or RHSSO" in Satellite builds. 

UPDATE: I edited the description to reflect the change of scope of this PR based on feedback in the comments.
UPDATE 2: Another edit to the description to reflect a change in how the attributes are defined.

* [x] Foreman 3.11
* [ ] Foreman 3.6/Katello 4.8


ekohl (commented):
Wouldn't this be redundant? If Quarkus based is always RHBK. We also tend to prefer upstream names. So can we make it:

| attribute | base | satellite |
| --- | --- | --- |
| `keycloak-wildfly` | Wildfly-based Keycloak | Red{nbsp}Hat Single Sign-On |
| `keycloak-quarkus` | Quarkus-based Keycloak | Red{nbsp}Hat build of Keycloak |

I think this will simplify headers, especially for Satellite builds. Now you have "Quarkus based Red{nbsp}Hat build of Keycloak" in titles, which is information Red Hat customers shouldn't care about.

Lennonka (changes requested):
I would consider Ewoud's suggestion ^^

asteflova:
> Wouldn't this be redundant? If Quarkus based is always RHBK. We also tend to prefer upstream names. So can we make it:
> attribute 	base 	satellite
> `keycloak-wildfly` 	Wildfly-based Keycloak 	Red{nbsp}Hat Single Sign-On
> `keycloak-quarkus` 	Quarkus-based Keycloak 	Red{nbsp}Hat build of Keycloak
> 

Nice! Thank you, applied.

I also took this PR one step further and introduced `keycloak-example-com` and `keycloak-realm` attributes to replace `keycloak-short`. And I did away with the last remaining general `keycloak` attribute instances. (This makes referencing keycloak in proc_disabling-keycloak-authentication.adoc slightly awkward, but I think it's still better than to keep `keycloak` as a standalone attribute.)

guides/common/modules/proc_configuring-project-settings-for-quarkus-keycloak-authentication-using-the-cli.adoc
```diff
@@ -57,18 +57,18 @@ For example:
 [options="nowrap", subs="+quotes,attributes"]
 ----
 # hammer settings set --name oidc_issuer \
---value "_{Keycloak-short}.example.com_/auth/realms/_{Keycloak-short}_Realm_"
+--value "_{Keycloak-example-com}_/auth/realms/_{keycloak-realm}_"
```
ekohl, line 60:
```suggestion
--value "_{keycloak-example-com}_/auth/realms/_{keycloak-realm}_"
```

guides/common/modules/proc_configuring-group-mapping-for-keycloak-authentication.adoc
ekohl, line 1: If this is now solely for keycloak-wildfly, should the filename reflect that? There are more files for this this would be true.

asteflova: No, it looks like I got overenthusiastic along the way and used the quarkus/wildfly-specific attributes also in modules that are reused in both quarkus and wildfly scenarios. Thanks for the catch.

I think I'll need to take a closer look at how to reuse the modules in a way that makes them fit into both scenarios, perhaps by introducing a parent-context attribute.

asteflova: I decided to address this by https://github.com/theforeman/foreman-documentation/pull/3044/commits/d99f4ab869c26750115dcad8c91337de6fbd5bb2: I'm re-introducing a general, platform-agnostic Keycloak attribute (`keycloak`). For non-Satellite builds, this attribute translates simply to `Keycloak`. For Satellite builds, it translates to `RHBK or RHSSO`.

I considered using context-based conditionals to ensure that only `RHBK` or `RHSSO` is displayed in respective assemblies. In `assembly_configuring-external-authentication.adoc`, contexts are already defined for each keycloak assembly. But I believe this would be unnecessarily complex and thus much more difficult to maintain in the long run.

Does that make sense? Does anyone see a better way to handle this?

guides/common/modules/proc_registering-project-as-a-keycloak-client.adoc
```diff
@@ -1,16 +1,16 @@
 [id="registering-{project-context}-as-a-keycloak-client_{context}"]
-= Registering {Project} as a {Keycloak} client
-Use this procedure to register {Project} to {Keycloak} as a client and configure {Project} to use {Keycloak} as an authentication source.
+= Registering {Project} as a {keycloak-wildfly} client
```
maximiliankolb, line 2: would be great if the file name matched the title more closely to differentiate between wildfly and quarkus when editing files.

asteflova: This is actually yet another module used in both wildfly and keycloak scenarios. So my bad, this should say {keycloak} rather than {keycloak-wildfly}. Updated here and in a few other places.

guides/common/modules/proc_registering-project-as-a-client-of-keycloak.adoc
```diff
@@ -22,16 +22,16 @@ You can also register two different {Project} clients to {Keycloak} if you want
...
 +
 * If you want you users to authenticate to {Project} using the web UI, create a client as follows:
 +
 [options="nowrap", subs="verbatim,quotes,attributes"]
 ----
 # keycloak-httpd-client-install --app-name foreman-openidc \
---keycloak-server-url "_https://{Keycloak-short}.example.com_" \
+--keycloak-server-url "_https://{keycloak-example-com}_" \
 --keycloak-admin-username "_admin_" \
 --keycloak-realm "_{Project}_Realm_" \
```
Lennonka, line 36:
```suggestion
--keycloak-realm "_{keycloak-realm}_" \
```
I wonder if this is supposed to be the same value. Also other occurences.

Lennonka (commented):
Generally, LGTM.

guides/common/modules/proc_registering-project-as-a-client-of-keycloak.adoc
Lennonka: Also, it would be nice to clarify allowed characters for this field. If it's going to be a part of the URL, there must be limitations.

asteflova: Addressing this would be a much bigger scope creep than I'm comfortable with in this PR. In this PR, I'm only making changes necessary to replace the currently-used attributes `keycloak-short` and `Keycloak` -- because these attrs have actually stopped working for our d/s now that we're dealing with Red Hat build of Keycloak in addition to Red Hat Single Sign On.

So, no change for this comment thread @Lennonka because if there is indeed a need to create a new attribute for a Project/Keycloak realm, I'll do that in a subsequent PR where I edit and test the procedures.

Lennonka (approved):
In that case, prolly good to go :+1:

asteflova:

d4784d2815..6ff2a4c42d  3.11 -> 3.11

PR #3053, opened 2024-05-31
Use ERB variables for easier branching

ekohl:
While I tried to somehow find a way to set asciidoctor attributes, I couldn't find a way. The data source already rendered it as .adoc.erb which means it first applies ERB and then asciidoctor. This allows us to use the item's data source and removes duplication, making branching easier.

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,9 +70,7 @@ When a commit is pushed into `X.Y`:
     * Change `katello` to the right version
     * Change `Nightly` in titles to the appropriate version
     * Remove guides which aren't ready for stable branches.
-  * Create a copy of [/web/releases/nightly.adoc](https://github.com/theforeman/foreman-documentation/tree/master/web/releases/nightly.adoc) as `X.Y.adoc` and edit it accordingly.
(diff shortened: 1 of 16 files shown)


ekohl:
How this internally works is by using [data-sources](https://nanoc.app/doc/data-sources/). So we have this code that determines the item.
https://github.com/theforeman/foreman-documentation/blob/d5fd29de1eae46fbe2ff4736a0c67a0e931b3b95/web/lib/releases.rb#L19-L40

Note a context is passed in to `new_item`. This is available as `@item`.

Another thing to note is `"/#{version}/index.adoc.erb"` so it already is rendered as if the filename was `index.adoc.erb`. I must admit I partly forgot this. We could rename `$version.adoc` to `$version.adoc.erb`, but I think I kept it as `.adoc` for syntax highlighting.

evgeni:
oookay, that explains why it works while my brain was saying "but it's not processed by erb!"

evgeni (approved):

PR #3058, opened 2024-06-04
Remove 'Monitoring jobs' to reflect the redesign of the job invocation details page in web UI

asteflova:
The job invocation details page [is being redesigned](https://community.theforeman.org/t/rfc-job-invocation-detail-page-redesign/33839). I want to make sure that the docs cover the new design appropriately. And I actually believe the most appropriate way to do that is to delete the module that describes the old design.

Looking at the contents of the existing `proc_monitoring-remote-jobs.adoc` module:

* The part with the web UI procedure: This describes things that users can plainly see in the web UI without documentation support. It's unlikely that users investigating the job details page to monitor jobs would need to look up documentation for further details; or at least the new design is clear enough in this regard.
* The part with the CLI procedure: This part describes using "hammer job-invocation" commands. While it's useful to mention these commands, section `proc_executing-a-remote-job.adoc` already directs users to `hammer job-invocation --help`, which I believe is sufficient.
* A note about canceling Ansible jobs: This looks like a result of fixing a doc bug in the past. We should preserve this, but I propose to move it to `proc_executing-a-remote-job.adoc`.

* [ ] Foreman 3.6/Katello 4.8


guides/common/modules/proc_executing-a-remote-job.adoc
```diff
@@ -3,6 +3,12 @@
 
 You can execute a job that is based on a job template against one or more hosts.
 
+[NOTE]
+====
+Ansible jobs run on batches of 100 hosts, so you cannot cancel a job running on a specific host.
```
Lennonka, line 8:
```suggestion
Ansible jobs run in batches on multiple hosts, so you cannot cancel a job running on a specific host.
```
No reason for specific numbers :)

Lennonka (changes requested):
I'm not sure if it's a good idea to remove the cancelling jobs CLI command entirely. Would you consider a "Canceling a remote job" module instead?

guides/common/modules/proc_executing-a-remote-job.adoc
Lennonka: I'm also wondering if the user can cancel the whole batch in mid flight.

asteflova: The best I'm willing to do about this is to create an issue for your question :) Would that work @Lennonka? (I'm just moving the note from the module I propose to delete so I'm not familiar with the background; I can only assume it's the result of a past doc bug fix.)

asteflova:
> I'm not sure if it's a good idea to remove the cancelling jobs CLI command entirely. Would you consider a "Canceling a remote job" module instead?

I think the guide is filled with small modules like that that mirror individual Hammer commands or web UI buttons/forms. It doesn't seem like a particularly helpful way to design documentation: If you look at the assembly/chapter I'm working in with this PR, it contains 28 modules/sections. I don't believe adding a module that covers a single command would enhance the doc UX.

Individual commands are described in the CLI as Hammer microcopy content. I could investigate a way to reference canceling jobs specifically in an existing module, perhaps through `.Additional resources` that would highlight that particular command... I'll experiment a little and see what works.

asteflova:
@Lennonka What do you think about https://github.com/theforeman/foreman-documentation/pull/3058/commits/f7a3d4fe963397fa1abffe492de93eb7889ce2f3? Using `.Additional resources` like this, I can reference the CLI option to cancel a running job (thus ensuring that users can find it with a full text search for example), but I don't need to create a whole module around it. Does this address your concern?

guides/common/modules/proc_executing-a-remote-job.adoc
```diff
@@ -96,4 +96,7 @@ You have the option to return to any part of the job wizard and edit the informa
 ----
 +
 Replace `_My_Search_Query_` with the filter expression that defines hosts, for example `"name ~ _My_Pattern_"`.
-For more information about executing remote commands with hammer, enter `hammer job-template --help` and `hammer job-invocation --help`.
+
+.Additional resources
+
+* For more information about creating, monitoring, or canceling remote jobs with hammer, enter `hammer job-template --help` and `hammer job-invocation --help`.
```
maximiliankolb, line 102:
```suggestion
* For more information about creating, monitoring, or canceling remote jobs with Hammer CLI, enter `hammer job-template --help` and `hammer job-invocation --help`.
```

We capitalize "Hammer": https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md#capitalization

CLI is probably optional.

guides/common/modules/proc_executing-a-remote-job.adoc
Lennonka: Probably not necessary until a customer asks for it :) Thanks

Lennonka (approved):
LGTM!

asteflova:
For the record: I have discussed these changes internally with the appropriate engineering team, which covers tech review.

PR #3059, opened 2024-06-04
Fix building docs using container image

maximiliankolb:

I assume building the docs locally by using the container image broke with a change to the Gemfile, but this is only a guess.

tested locally:

````
$ podman build --tag tmp:2 --file Dockerfile .
$ rm -fr guides/build && podman run --rm -v $(pwd):/foreman-documentation:Z tmp:2 make html BUILD=katello
````

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,19 @@
-FROM fedora:38
+FROM fedora:40
 
-RUN dnf upgrade -y && \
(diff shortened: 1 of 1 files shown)


Lennonka (approved):
LGTM.
If you've tested it and it works, I'm okay with it.

PR #2890, opened 2024-03-15
Improve Discovery in provisioning

Lennonka:
Goals
* Restructuring of original content
* Re-modularization
* Terminology improvements
* Beginner-friendliness
* Reduction of bloat (hopefully) and "de-spaghettization" :)
* Renamed images and image sources

For review, you can either go commit by commit or treat it as an entirely new chapter (latter recommended).

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into: **for Foreman 3.12**

* [ ] Foreman 3.11
* [ ] Foreman 3.10/Katello 4.12
* [ ] Foreman 3.9/Katello 4.11 (planned Satellite 6.15)
* [ ] Foreman 3.8/Katello 4.10
* [ ] Foreman 3.7/Katello 4.9 (Satellite 6.14)
* [ ] Foreman 3.6/Katello 4.8
* [ ] Foreman 3.5/Katello 4.7 (Satellite 6.13; orcharhino 6.6/6.7)
* [ ] Foreman 3.4/Katello 4.6 (EL8 only)
* [ ] Foreman 3.3/Katello 4.5 on EL7 & EL8 (Satellite 6.12 on EL8 only; orcharhino 6.4/6.5 on EL8 only)
* We do not accept PRs for Foreman older than 3.3.

PR #555, opened 2021-06-10
Add adoc version of original 10 Steps SOE document

wzzrd:
As discussed, a PR to merge the original version of the 10 steps into the documentation repo in order to facilitate working on this with Red Hatters, customers and Foreman community members alike.

PR #3069, opened 2024-06-06
Add diagrams of provisioning methods

Lennonka:
New diagrams to describe provisioning methods in PlantUML (SVG export available).

The diagrams will **not be included** in the docs for now, but I will include them once I'll document the provisioning methods properly. This classification of provisioning methods is incompatible with the current docs.

### Classification of provisioning methods

* Installer-based provisioning with network boot
  - PXE boot
  - HTTP boot

* Installer-based provisioning with boot disk
  - Full host image
  - Subnet image

* Image-based provisioning
  - with cloud-init script
  - with Finish script

In some cases I had to guess a lot, so don't take my word as gospel and raise your concerns if you think it's wrong.

Please cherry-pick my commits into: N/A (hoping to finish it for 3.12)


Lennonka:
Installer-based provisioning with PXE and HTTP boot are both ready for comments.

ekohl (commented):
Instead of adding the images (which end up in every single guide), can't we use https://docs.asciidoctor.org/diagram-extension/latest/diagram_types/plantuml/ to render them when needed? Especially since we won't use them downstream I'd much prefer that.

Lennonka:
@ekohl I can't include them at the moment (see PR Description), but noted. For now, I'd like to focus on getting feedback on the diagrams and use this PR for this purpose. We don't have to merge it.

Lennonka:
I don't really want to make things more complicated. I'll add the source files in the existing image-sources folder, but I will merge the PR without the generated images.

Lennonka:
I've hidden the Puppet stuff from the diagrams to simplify them. (This can be enabled in `foreman.pstyle`.)

guides/common/images/provisioning-installer-bootdisk-fullhost.png
knoppi, line 1: Two comments here:

- We use bootdisk based provisioning when there is no DHCP. IP configuration then gets written onto the bootdisk. (No steps 3 and 16). I've never really tried it with a DHCP server and probably it is simply correct as you've drawn it here. But since without a DHCP server bootdisk is THE preferred way it would be cool to have it reflected here. Like have step 3 optional and in step 16 alternatively "Use IP configuration from bootdisk".
- Step 8 does not need to target Pulp (Katello). Depending on the setup and also the Linux distribution you want to provision, the installation medium is a valid target, too.

guides/image-sources/provisioning-installer-bootdisk-fullhost.plantuml
```diff
@@ -0,0 +1,61 @@
...
+Foreman -> Proxy : requests free IP address
+Proxy -> DHCP : reads DHCP cofiguration
+note over Foreman : pings IP address
+User -> Foreman : submits the **Create host** form
+group Templates [*PXE* installer boot]
+    Foreman -> Proxy : renders bootloader configuration files
+end
+Proxy -> Proxy : deploys bootloader configuration files
+Foreman -> Proxy : commands to download OS installer kernel\nand initial RAM disk
+Proxy -> Pulp : downloads OS installer kernel\nand initial RAM disk
```
stejskalleos, line 33: Isn't this happening  **only** if Katello is installed and only for RHEL operating systems? Like getting the data from the Kickstart repository?

Otherwise, they are downloaded from the media URL, which could be somewhere on the internet.

guides/image-sources/provisioning-installer-bootdisk-fullhost.plantuml
```diff
@@ -0,0 +1,61 @@
...
+note over User : writes the bootdisk ISO\nto a USB/CD/DVD drive
+
+== Boot into OS installer ==
+
+User -> Host : configures the machine to boot\nfrom the USB/CD/DVD drive
+User -> Host : powers on the machine
+Host -> DHCP : requests the reserved IP
+note over Host : boots from the USB/CD/DVD drive
+note over Host : OS installer loads
+User -> Host : eliminates the USB/CD/DVD drive\n(too soon?)
```
stejskalleos, line 50: If we don't know the exact step where it happens, do we have to mention it?

guides/image-sources/prov-os-content.iuml
```diff
@@ -0,0 +1,8 @@
+loop
+   Host -> Foreman : requests OS content
+   Foreman -> Proxy : requests OS content
+   Proxy -> Pulp : requests OS content
+   Pulp -> Proxy : provides OS content
+   Proxy -> Foreman : provides OS content
+   Foreman -> Host : provides OS content
+end
```
stejskalleos, line 8: Nit: Empty line at the end

guides/image-sources/prov-initial-configuration.iuml
```diff
@@ -0,0 +1,7 @@
+!if ($puppet)
+opt when Puppet enabled
+    Host -> Puppet : signs Puppet client certificate (Puppet agent)
+end
+!else
+note over Host : initial host configuration\n- remote execution,\n- signs Puppet client cert,\n- runs Ansible roles,\n- etc.
+!endif
```
stejskalleos, line 7: Nit: Empty line at the end

guides/image-sources/prov-first-local-boot-pxe.iuml
```diff
@@ -0,0 +1,15 @@
+Host -> DHCP : requests the reserved IP and DHCP options
+Host -> TFTP : downloads bootloader
+Host -> TFTP : gets  MAC-based bootloader config
+Host -> Host : config tells bootloader to boot from HDD
+Host -> Host : boots from HDD
+!if ($puppet)
```
stejskalleos, line 6: I am just curious: Where does the `$puppet` come from? Could we also have `$katello` to skip/add some steps in the diagrams? Or would that be too complicated?

guides/common/images/provisioning-installer-bootdisk-fullhost.png
stejskalleos: @Lennonka No DHCP setup could be a great improvement in follow-up changes; there are quite a few users who have no DHCP setups.

stejskalleos (changes requested):
Impressive work, this will greatly benefit everyone using Foreman.

@Lennonka I would suggest sharing it on community.theforeman.org, the content is huge, and more eyes see more things. I don't know each workflow in detail myself, would feel more confident to get more opinions from the community.

guides/image-sources/provisioning-installer-bootdisk-fullhost.plantuml
Lennonka: Well, they have to do it at some point, otherwise they would be stuck in a loop :D
So we might as well suggest when is the right time.

As it is a necessary step in the process, I would prefer to keep it.

guides/image-sources/prov-first-local-boot-pxe.iuml
Lennonka: The original diagrams by lzap were very Puppet centric. I wasn't sure if we want to remove it altogether.
It is possible to only include katello stuff conditionally, that should be easy.

guides/common/images/provisioning-image-finish.png
shubhamsg199, line 1: Step 8 Create DNS Records is repeated here

Lennonka: Nice catch, thanks!

guides/common/images/provisioning-installer-bootdisk-fullhost.png
Lennonka: @knoppi Does it mean that steps 9 and 10 are also irrelevant?

guides/image-sources/prov-first-local-boot-pxe.iuml
Lennonka: I've edited the diagrams to be neutral towards the source of installation media.

guides/common/images/provisioning-installer-bootdisk-fullhost.png
knoppi: No, it's not. Good catch. Basically you can drop the whole lane DHCP.

apinnick:
@Lennonka Will your diagrams eventually appear in d/s docs? If so, you will need to submit a 'diagram request' form so that they comply with branding guidelines.

Lennonka:
> Will your diagrams eventually appear in d/s docs? If so, you will need to submit a 'diagram request' form to ensure that the diagrams comply with branding guidelines.

@apinnick Yes, they will, and yes, I am aware. TY!

guides/common/images/provisioning-installer-bootdisk-fullhost.png
Lennonka: Updated.

guides/image-sources/provisioning-image-cloudinit.plantuml
```diff
@@ -0,0 +1,55 @@
...
+
+== Create host in Foreman ==
+
+User -> Foreman : clicks **Create Host**
+User -> Foreman : selects cloud resource,\nthe image and submits
+group Template [cloud-init]
+   Foreman -> Proxy : renders the cloud-init script
+end
+Foreman -> Cloud : creates new instance\nwith cloud-init info
+Cloud -> Foreman : reports IP address
```
ekohl, line 27: Technically it can report IPv4 and IPv6. If both are reported, that also means the DNS record will get both `A` and `AAAA` records:
https://github.com/theforeman/foreman/blob/e80d5558767bb713c13264e821122baea577af76/app/models/concerns/orchestration/compute.rb#L69-L72

After this will store the instance details that the compute resource provided:
https://github.com/theforeman/foreman/blob/e80d5558767bb713c13264e821122baea577af76/app/models/concerns/orchestration/compute.rb#L73-L74

Not all compute resources provide IPs and some provide MAC addresses. In that case it will also try to reserve an IP:
https://github.com/theforeman/foreman/blob/e80d5558767bb713c13264e821122baea577af76/app/models/concerns/orchestration/compute.rb#L75-L78

guides/image-sources/provisioning-image-cloudinit.plantuml
```diff
@@ -0,0 +1,55 @@
...
+note over Foreman : has an image\nwith cloud-init flag
+
+== Create host in Foreman ==
+
+User -> Foreman : clicks **Create Host**
+User -> Foreman : selects cloud resource,\nthe image and submits
+group Template [cloud-init]
+   Foreman -> Proxy : renders the cloud-init script
+end
+Foreman -> Cloud : creates new instance\nwith cloud-init info
```
ekohl, line 26: This is correct: https://github.com/theforeman/foreman/blob/e80d5558767bb713c13264e821122baea577af76/app/models/concerns/orchestration/compute.rb#L67-L68

guides/image-sources/provisioning-image-cloudinit.plantuml
```diff
@@ -0,0 +1,55 @@
...
+Cloud -> Foreman : reports IP address
+Foreman -> Proxy : creates DNS records
+Proxy -> DNS : forwards DNS records
```
ekohl, line 29: This is not entirely correct. First the code reference: https://github.com/theforeman/foreman/blob/e80d5558767bb713c13264e821122baea577af76/app/models/concerns/orchestration/dns.rb#L63-L71

It will create all feasible DNS record types. That can mean `A`, `AAAA` and `PTR` (for both v4 and v6), depending on the known IP addresses. So the first part is simplified, but correct.

The second part is IMHO confusing. The Proxy tells the DNS server to create the DNS records. A Foreman Proxy, especially the DNS module, implements the [facade pattern](https://en.wikipedia.org/wiki/Facade_pattern). It abstracts away the actual implementation away from Foreman. IMHO it is forwarding the DNS record creation, not the records themselves.

guides/image-sources/provisioning-image-cloudinit.plantuml
```diff
@@ -0,0 +1,55 @@
...
+!if ($puppet)
+Foreman -> Proxy : creates Puppet sign request (autosign enabled)
+Proxy -> Puppet : forwards Puppet sign request
+!endif
+Foreman -> Cloud : starts the instance
```
ekohl, line 34: Referencing the code: https://github.com/theforeman/foreman/blob/e80d5558767bb713c13264e821122baea577af76/app/models/concerns/orchestration/compute.rb#L79-L82

guides/image-sources/provisioning-image-cloudinit.plantuml
```diff
@@ -0,0 +1,55 @@
...
+note over Host : boots up
+
+== Cloud-init script ==
+
+!include prov-initial-configuration.iuml
+Host -> Foreman : calls home\n(disables build mode)
```
ekohl, line 40: This is the part that **can** go through the Proxy's template feature. The use of the Foreman Proxy templates module is optional. If that isn't used then cloud-init requests it directly from Foreman.

guides/image-sources/provisioning-image-cloudinit.plantuml
```diff
@@ -0,0 +1,55 @@
...
+
+!if ($puppet)
+== First Puppet run ==
+
+Host -> Puppet : sends initial facts
+Puppet -> Foreman : forwards initial facts
+Host -> Puppet : compiles catalog
+note over Host : host is configured
```
ekohl, line 48: I'd say this is different, depending on how much detail you want to include.

If the host bootstrapped, we can assume a certificate. Our default templates do not include this, so it first starts with the SSL bootstrap:
* Host -> Puppet : send CSR
* Puppet -> Host : send certificate (if allowed to)

Then a regular Puppet run continues:

* Host -> Puppet : send facts
* Host -> Puppet : request catalog
* Puppet -> Foreman : forward facts
* Puppet -> Foreman : request ENC
* Puppet -> Host : respond with requested catalog
* Host runs catalog
* Host -> Puppet : send report
* Puppet -> Foreman : forward report

ekohl (changes requested):
What you're describing is Foreman's orchestration, which is complex. I only looked at one workflow, but I think they're not really correct. For example, I get the impression you think templates are pre-rendered. This is not the case and they're actually all rendered live as requested.

Lennonka:
> What you're describing is Foreman's orchestration, which is complex. I only looked at one workflow, but I think they're not really correct. For example, I get the impression you think templates are pre-rendered. This is not the case and they're actually all rendered live as requested.

@ekohl Regarding template rendering - that depends on the template. The *PXE* templates are rendered after the user creates a host (PXE and HTTP boot workflows). I have verified this in the Foreman log. Prove me otherwise. What you mention definitely applies to the provision kind of templates, e.g. `Kickstart default`. For userdata and Finish I'm not sure when exactly it happens in the sequence.

By the way, saying that "it's wrong" without providing correct information is not helpful.

guides/image-sources/provisioning-image-cloudinit.plantuml
Lennonka: Isn't the SSL bootstrap part of initial configuration, i.e. `prov-initial-configuration.iuml`?

guides/image-sources/provisioning-image-cloudinit.plantuml
Lennonka: > Technically it can report IPv4 and IPv6. 

That is not the level of detail we want here.

> After this will store the instance details that the compute resource provided
>Not all compute resources provide IPs and some provide MAC addresses. In that case it will also try to reserve an IP 

I don't understand what it means for the diagram.

guides/image-sources/provisioning-image-cloudinit.plantuml
```diff
@@ -0,0 +1,55 @@
...
+
+note over Foreman : has an image\nwith cloud-init flag
+
+== Create host in Foreman ==
+
+User -> Foreman : clicks **Create Host**
+User -> Foreman : selects cloud resource,\nthe image and submits
+group Template [cloud-init]
+   Foreman -> Proxy : renders the cloud-init script
+end
```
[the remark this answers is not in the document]
Lennonka: So this is correct then.

guides/image-sources/provisioning-image-cloudinit.plantuml
Lennonka: So this?

```suggestion
Foreman -> Cloud : sets the compute instance\nwith cloud-init info
```

guides/image-sources/provisioning-image-cloudinit.plantuml
Lennonka: So what wording do you suggest precisely?

guides/image-sources/provisioning-image-cloudinit.plantuml
Lennonka: I'm assuming that means that it's correct.

ekohl:
> I am aware of that and different template kinds are requested at different times, which I'm trying to convey in the diagrams:
> 
>     * The `*PXE*` templates are rendered after the user creates a host (PXE and HTTP boot workflows). I have verified this in the Foreman log.
> 
>     * The `provision` kind of templates, e.g. `Kickstart default`, is rendered when requested by the Anaconda installer.
> 
>     * For `userdata` and `Finish` I'm not sure when exactly it happens in the sequence.

Ah, now I understand the confusion. They are templates and they are rendered, but I was looking at it from a Smart Proxy modules perspective.

We have one module called `templates`, which you can also see in the UI if you go to Infrastructure -> Smart Proxies -> choose the proxy. However, this module's purpose is to serve as a proxy between the host and Foreman to render templates. Yes, we overuse proxy a lot. In the case of isolated hosts, this is used to retrieve rendered templates like `provision` and `userdata`. These are all rendered on the fly.

Then we also have PXE templates. These are managed by the `tftp` feature (again, same thing with visibility in the UI). These are indeed rendered before a host is actually started. That happens here: https://github.com/theforeman/foreman/blob/e80d5558767bb713c13264e821122baea577af76/app/models/concerns/orchestration/tftp.rb#L148-L154

To clarify how the correct proxy is related: it's via the subnet on the provisioning interface. On the subnet you can select a proxy to be used for templates (if not, it's retrieved directly from Foreman). You can select a different proxy to be used for TFTP.

In practice I'd say it's uncommon for them to point to different proxies. A more common scenario is that you use one feature but not the other and this is the simplest UI to achieve that.

Finish templates are yet another thing. That is the SSH provisioning method, which I believe is typically used on image based provisioning without cloud-init. The code for this is: https://github.com/theforeman/foreman/blob/e80d5558767bb713c13264e821122baea577af76/app/models/concerns/orchestration/ssh_provision.rb#L20-L28 (note the priority is 2000, so it happens way after a host is actually created).

I very much think this is the effort I wanted to get started when we had the provisioning discussion in Brno back in May. Digging into this and describing it was what I tried to steer towards. That is why I refer to the code a lot: I'm also looking things up myself.

Note that for now I've only looked at `foreman.git` but every plugin can add orchestration actions. A few more plugins also add orchestration steps. For provisioning, foreman_bookdisk is the most relevant but that's just looking at filenames.

If you looked at the code, you may also have noticed I only linked to `_create` methods and that there are also `_update` and `_destroy`. That's for another PR.

I'll look at your other comments tomorrow, but wanted to get this out of my head before I went to sleep.

Lennonka:
@ekohl Please, remember that UML is about abstraction. I want to make these diagrams as simplified as possible to make them easy to understand for first-time users. Savvy?

apinnick:
Very cool tool!

ekohl:
> @ekohl Please, remember that UML is about abstraction. I want to make these diagrams as simplified as possible to make them easy to understand for first-time users. Savvy?

For now I've only had time to do research and haven't had time to distill it into user-friendly diagrams. I work bottom-up, so I start with all the details.

As a next step I'd like to build a table with all the possible orchestration actions and put them in a table with:

* priority
* name
* description
* code reference

That will help me verify the diagrams for factual correctness.

Right now we're in the middle of branching preparations and due to breakages in nightly I'm behind, so I won't have time to do so this week.

Lennonka:
@ekohl You're coming very late and I need to finish this before branching. You go ahead and build that table. I'm going ahead and merge this as it is so that I have something to show for the work I've done on this.

Lennonka:
@ekohl This [has been available for comments since July 20th](https://github.com/theforeman/foreman-documentation/pull/3069#issuecomment-2241296998), [some pieces even before that](https://github.com/theforeman/foreman-documentation/pull/3069#issuecomment-2209609038). And now at the last minute when I need to merge it, you are telling me that you have to do reverse engineering, but you don't really have time to do it before the dealine?
No, sir.
I'm going to merge it regardless of your comments and YOU will file another PR for improvements whenever you want.

ekohl:
I started a review because you explicitly asked for it in https://community.theforeman.org/t/rfc-diagrams-of-provisioning-workflows/39045, even though I was very busy with unblocking the Foreman 3.12 branching that happens this week. Then I had questions about half the lines in the first diagram I looked at so I took the time to investigate what they should be. I did that because you didn't provide any sources or references how you built these diagrams.

After reviewing one diagram I came to the conclusion that I don't trust your research and factual correctness. Just because you are impatient doesn't mean we should just merge it. I don't consider this time sensitive because it's all well established functionality and we haven't had these diagrams for years. I have a really hard time accepting that we need to merge this when there are known concerns.

guides/image-sources/provisioning-image-cloudinit.plantuml
ekohl: > That is not the level of detail we want here.

It's about singular vs plural.

```suggestion
Cloud -> Foreman : reports IP addresses
```

> I don't understand what it means for the diagram.

Probably not for this specific diagram.

Lennonka:
Well, it's true that we have time before it gets actually published. I'll explain it to my managers.

ekohl:
Sharing my research here. First I did some small investigation where the classes are used. At least for core. For now this is just a list of tasks. I'd like this to be converted into a complete UML diagram, similar to what we have for Kafo (https://github.com/theforeman/kafo/blob/master/doc/kafo_run.uml & https://github.com/theforeman/kafo/blob/master/doc/kafo_run.png). The intended audience for that is developers, but it can be a great source for user documentation.

<details>
<summary>Small rake task I wrote that generate everything below</summary>

This is ugly and a hack. Ideally this is extended a bit with additional plugins, such as Katello. A more dynamic way to find the classes (perhaps using `Orchestration.constants`?) would be great. Also parsing the source with regular expressions is bad form.

```ruby
task methods: :environment do
  classes = [
    Orchestration::Compute,
    Orchestration::Puppetca,
    Orchestration::SSHProvision,
    Orchestration::Realm,
    Orchestration::DHCP,
    Orchestration::DNS,
    Orchestration::TFTP,
    Orchestration::ExternalIPAM,
  ]

  Row = Struct.new('Row', :priority, :action, :name, :code)

  table = []

  puts <<~HEADER
    # Models with orchestration

    ## Host::Managed

    * `Orchestration::Compute`
    * `Orchestration::Puppetca`
    * `Orchestration::SSHProvision`
    * `Orchestration::Realm`

    ## Nic::Managed

    * `Orchestration::DHCP`
    * `Orchestration::DNS`
    * `Orchestration::TFTP`
    * `Orchestration::ExternalIPAM`

    # Overview

    <details>
    <summary>Extracted code</summary>
  HEADER

  classes.each do |cls|
    methods = cls.instance_methods + cls.private_instance_methods
    methods.select { |method| method.end_with?('_create') }.each do |method|
      code = "#{cls.name}##{method.name}"

      puts "## #{code}"
      puts
      puts "```ruby"
      puts cls.instance_method(method).source
      puts "```"
      puts

      cls.instance_method(method).source.scan(/queue\.create\(.+?\)$/m) do |parameters|
        name = parameters.match(/(:name\s*=>|name:)\s*_\(["'](?<name>.+)["']\)/)[:name]
        priority = parameters.match(/(:priority\s*=>|priority:)\s*(?<priority>[\d_]+)/)[:priority]
        action = parameters.match(/(:action\s*=>|action:)\s*\[self, :(?<action>[A-Za-z_]+)/)[:action]

        table << Row.new(priority, action, name, code)
      end
    end
  end
  puts "</details>"
  puts

  puts "| priority | action | name | code |"
  puts "|---|---|---|---|"
  table.sort_by { |row| row.priority.to_i }.each do |row|
    puts "| #{row.priority} | `#{row.action}` | #{row.name} | `#{row.code}` |"
  end
end
```
</details>

# Models with orchestration

## Host::Managed

* `Orchestration::Compute`
* `Orchestration::Puppetca`
* `Orchestration::SSHProvision`
* `Orchestration::Realm`

## Nic::Managed

* `Orchestration::DHCP`
* `Orchestration::DNS`
* `Orchestration::TFTP`
* `Orchestration::ExternalIPAM`

# Overview

<details>
<summary>Extracted code</summary>
## Orchestration::Compute#queue_compute_create

```ruby
  def queue_compute_create
    if find_image.try(:user_data)
      queue.create(:name   => _("Render user data template for %s") % self, :priority => 2,
                   :action => [self, :setUserData])
    end
    queue.create(:name   => _("Set up compute instance %s") % self, :priority => 3,
                 :action => [self, :setCompute])
    if compute_provides?(:ip) || compute_provides?(:ip6)
      queue.create(:name   => _("Acquire IP addresses for %s") % self, :priority => 4,
                   :action => [self, :setComputeIP])
    end
    queue.create(:name   => _("Query instance details for %s") % self, :priority => 5,
                 :action => [self, :setComputeDetails])
    if compute_provides?(:mac) && (mac_based_ipam?(:subnet) || mac_based_ipam?(:subnet6))
      queue.create(:name   => _("Set IP addresses for %s") % self, :priority => 6,
                   :action => [self, :setComputeIPAM])
    end
    if compute_attributes && compute_attributes[:start] == '1'
      queue.create(:name   => _("Power up compute instance %s") % self, :priority => 1000,
                   :action => [self, :setComputePowerUp])
    end
  end
```

## Orchestration::Puppetca#queue_puppetca_create

```ruby
  def queue_puppetca_create
    post_queue.create(:name => _("Cleanup PuppetCA certificates for %s") % self, :priority => 51,
                      :action => [self, :delCertificate])
    post_queue.create(:name => _("Enable PuppetCA autosigning for %s") % self, :priority => 55,
                      :action => [self, :setAutosign])
  end
```

## Orchestration::SSHProvision#queue_ssh_provision_create

```ruby
  def queue_ssh_provision_create
    post_queue.create(:name   => _("Prepare post installation script for %s") % self, :priority => 2000,
                 :action => [self, :setSSHProvisionScript])
    post_queue.create(:name   => _("Wait for %s to come online") % self, :priority => 2001,
                 :action => [self, :setSSHWaitForResponse])
    post_queue.create(:name   => _("Configure instance %s via SSH") % self, :priority => 2003,
                 :action => [self, :setSSHProvision])
  end
```

## Orchestration::Realm#queue_realm_create

```ruby
  def queue_realm_create
    queue.create(:name => _("Create realm entry for %s") % self, :priority => 1,
                 :action => [self, :set_realm])
  end
```

## Orchestration::DHCP#queue_dhcp_create

```ruby
  def queue_dhcp_create
    logger.debug "Scheduling new DHCP reservations for #{self}"
    queue.create(id: generate_dhcp_task_id("create"), name: _("Create DHCP Settings for %s") % self, priority: 10, action: [self, :set_dhcp]) if dhcp?
  end
```

## Orchestration::DNS#queue_dns_create

```ruby
  def queue_dns_create
    logger.debug "Scheduling new DNS entries"
    DnsInterface::RECORD_TYPES.each do |record_type|
      if dns_feasible?(record_type)
        queue.create(:name   => _("Create %{type} for %{host}") % {:host => self, :type => dns_class(record_type).human}, :priority => 10,
                     :action => [self, :set_dns_record, record_type])
      end
    end
  end
```

## Orchestration::TFTP#queue_tftp_create

```ruby
  def queue_tftp_create
    host.operatingsystem.template_kinds.each do |kind|
      queue.create(:name => _("Deploy TFTP %{kind} config for %{host}") % {:kind => kind, :host => self}, :priority => 20, :action => [self, :setTFTP, kind])
    end
    return unless build
    queue.create(:name => _("Fetch TFTP boot files for %s") % self, :priority => 25, :action => [self, :setTFTPBootFiles])
  end
```

## Orchestration::ExternalIPAM#queue_external_ipam_create

```ruby
  def queue_external_ipam_create
    return unless (external_ipam?(subnet) || external_ipam?(subnet6)) && errors.empty?
    logger.debug "Scheduling new IP reservation(s) in external IPAM for #{self}"
    queue.create(id: generate_external_ipam_task_id("create", subnet.network_type), name: _("Creating IPv4 in External IPAM for %s") % self, priority: 10, action: [self, :set_add_external_ip, {:ip => ip, :subnet => subnet}]) if ip.present? && subnet.present? && external_ipam?(subnet)
    queue.create(id: generate_external_ipam_task_id("create", subnet6.network_type), name: _("Creating IPv6 in External IPAM for %s") % self, priority: 10, action: [self, :set_add_external_ip, {:ip => ip6, :subnet => subnet6}]) if ip6.present? && subnet6.present? && external_ipam?(subnet6)
    true
  end
```

</details>

| priority | action | name | code |
|---|---|---|---|
| 1 | `set_realm` | Create realm entry for %s | `Orchestration::Realm#queue_realm_create` |
| 2 | `setUserData` | Render user data template for %s | `Orchestration::Compute#queue_compute_create` |
| 3 | `setCompute` | Set up compute instance %s | `Orchestration::Compute#queue_compute_create` |
| 4 | `setComputeIP` | Acquire IP addresses for %s | `Orchestration::Compute#queue_compute_create` |
| 5 | `setComputeDetails` | Query instance details for %s | `Orchestration::Compute#queue_compute_create` |
| 6 | `setComputeIPAM` | Set IP addresses for %s | `Orchestration::Compute#queue_compute_create` |
| 10 | `set_dns_record` | Create %{type} for %{host} | `Orchestration::DNS#queue_dns_create` |
| 10 | `set_add_external_ip` | Creating IPv4 in External IPAM for %s | `Orchestration::ExternalIPAM#queue_external_ipam_create` |
| 10 | `set_add_external_ip` | Creating IPv6 in External IPAM for %s | `Orchestration::ExternalIPAM#queue_external_ipam_create` |
| 20 | `setTFTP` | Deploy TFTP %{kind} config for %{host} | `Orchestration::TFTP#queue_tftp_create` |
| 25 | `setTFTPBootFiles` | Fetch TFTP boot files for %s | `Orchestration::TFTP#queue_tftp_create` |
| 51 | `delCertificate` | Cleanup PuppetCA certificates for %s | `Orchestration::Puppetca#queue_puppetca_create` |
| 55 | `setAutosign` | Enable PuppetCA autosigning for %s | `Orchestration::Puppetca#queue_puppetca_create` |
| 1000 | `setComputePowerUp` | Power up compute instance %s | `Orchestration::Compute#queue_compute_create` |
| 2000 | `setSSHProvisionScript` | Prepare post installation script for %s | `Orchestration::SSHProvision#queue_ssh_provision_create` |
| 2001 | `setSSHWaitForResponse` | Wait for %s to come online | `Orchestration::SSHProvision#queue_ssh_provision_create` |
| 2003 | `setSSHProvision` | Configure instance %s via SSH | `Orchestration::SSHProvision#queue_ssh_provision_create` |

Lennonka:
@ekohl I don't see how I'm supposed to use that table. It seems to me that the table introduces too much detail for what I'm trying to achieve and it would make the diagrams even more complicated than they already are.

Lennonka:
Feel free to suggest further modifications in a subsequent PR. Thank you for cooperation!

asteflova:
I have serious concerns over the fact that this PR seems to have been merged without an approval and with an unresolved request for changes. (For the record, I also searched through Jira and Slack to see if I could find an ack of some sorts in downstream-only resources but didn’t find anything convincing.) I’m worried that merging PRs like this one would set a precedent that I would be highly uncomfortable with.

From my point of view, the following conventions were not adhered to before this PR got merged:

- Requesting to close unresolved threads before merging: https://github.com/theforeman/foreman-documentation/pull/3070#pullrequestreview-2152185523
- Announcing a time for last-call comments before self-merge, especially if there is room for doubt: https://github.com/theforeman/foreman-documentation/pull/2890#issuecomment-2163252014 
- Opening an issue or a subsequent PR right away to ensure we have a tracker to address any additional changes that can't be resolved right away (but this should never be intended to delay addressing technical concerns over any new content that is being added).

I am aware of one precedent when this team merged an update without following the conventions and without finalizing a technical review, but that merge was at least preceded by a team conversation about whether everyone was okay with this approach (https://github.com/theforeman/foreman-documentation/pull/555#issuecomment-863972124) . And as a matter of fact, that strategy did not work out so well: because the content was not intended for publication, it wasn't anyone's priority, no one took the time to work on it, and then it was deleted a couple of years later: https://github.com/theforeman/foreman-documentation/commit/5e75a54e136c1fd1402eda74bd53454f84f75eba 

Because of all of the above, I would like to propose reverting these changes until @Lennonka can secure a transparent approval and address the lingering concerns.

Additionally, to prevent a similar situation in the future, we could consider updating our merge process to, for example, prevent anyone from merging a PR that has not received proper technical and peer approvals (by requesting at least one formal ack on a PR). I’d like to know what others would think about this, please do let me know (here or elsewhere, publicly or privately).

One final note: The fact that I'm sharing my concerns and proposing the above steps to mitigate the situation does not mean I don't appreciate @Lennonka's effort. I do appreciate it. I think this is a significant and impressive initiative that deserves proper attention and due diligence.

PR #3072, opened 2024-06-11
Add section on updating operating system

ehelms:

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_updating-server.adoc
+++ b/guides/common/modules/proc_updating-server.adoc
@@ -3,6 +3,7 @@
 
 ifdef::satellite[]
 Update your connected {ProjectServer} to the next minor version.
+You can follow this process to update the underlying operating system between minor release versions.
 For information to update a disconnected {Project} setup, see xref:Updating-Disconnected-{project-context}_{context}[].
 endif::[]
 ifndef::satellite[]
--- a/guides/common/modules/con_updating-project-to-next-minor-version.adoc
+++ b/guides/common/modules/con_updating-project-to-next-minor-version.adoc
@@ -4,4 +4,7 @@
 You can update your {ProjectServer} and {SmartProxyServer} to a new minor release version, such as from {ProjectVersion}.0 to {ProjectVersion}.1, using the {Project} maintain tool.
 The minor releases are non-disruptive to your operating environment and often fast.
 
+You can update the underlying operating system.
+If there are pending {ProjectServer} updates, updating the operating system will update both.
+
 {Team} recommends performing updates regularly, because the minor releases patch security vulnerabilities and minor issues discovered after code is released.
--- a/guides/common/modules/proc_updating-disconnected-server.adoc
+++ b/guides/common/modules/proc_updating-disconnected-server.adoc
@@ -2,6 +2,7 @@
 = Updating a disconnected {ProjectServer}
 
 Update your air-gapped {Project} setup where the connected {ProjectServer}, which synchronizes content from CDN, is air gapped from a disconnected {ProjectServer}, to the next minor version.
+You can follow this process to update the underlying operating system between minor release versions.
 
 .Prerequisites
 * Back up your {ProjectServer}.


guides/common/modules/proc_updating-operating-system.adoc
```diff
@@ -0,0 +1,36 @@
+[id="updating-operating-system_{context}"]
+= Updating the operating system
+
+{Team} recommends performing operating system updates regularly, because the minor releases patch security vulnerabilities and minor issues discovered after code is released.
+
+.Prerequisites
+* Back up your {ProjectServer}.
+For more information, see {AdministeringDocURL}backing-up-{project-context}-server-and-{smart-proxy-context}_admin[Backing Up {ProjectServer} and {SmartProxyServer}] in _{AdministeringDocTitle}_.
+
+include::snip_warning-maintain-config-noop.adoc[]
```
evgeni, line 10: This warning reads:

> If you customize configuration files, manually or using a tool such as Hiera, these changes are overwritten when the maintenance script runs during upgrading or updating. You can use the --noop option with the satellite-installer to test for changes. For more information, see the Red Hat Knowledgebase solution [How to use the noop option to check for changes in Satellite config files during an upgrade.](https://access.redhat.com/solutions/3351311)

The thing is, it's sorta-useless.
- The linked KCS talks about EL6 and EL7, and mentions MongoDB.
- While this would work for OS-only updates, when there are Satellite updates `--noop` would run with the wrong installer version (yet we recommend it in the "update satellite" case too)

guides/common/modules/proc_updating-operating-system.adoc
```diff
@@ -0,0 +1,36 @@
...
+Review the results and address any highlighted error conditions before performing the update.
+. Because of the lengthy update time, use a utility such as `tmux` to suspend and reattach a communication session.
+You can then check the update progress without staying connected to the command shell continuously.
++
+If you lose connection to the command shell where the update command is running, you can see the logged messages in the `{installer-log-file}` file to check if the process completed successfully.
+. Perform the update:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foreman-maintain} update run
```
evgeni, line 30: What happens if there are not only OS but also Satellite updates outstanding?
Will this ensure Satellite is (not?) updated? (I hope it won't, as we often ship fixes that are required for "latest BaseOS", but the wording in this chapter implies otherwise).

ehelms: It will include them - I added a note.

guides/common/modules/proc_updating-operating-system.adoc
```diff
@@ -0,0 +1,36 @@
+[id="updating-operating-system_{context}"]
+= Updating the operating system
+
+{Team} recommends performing operating system updates regularly, because the minor releases patch security vulnerabilities and minor issues discovered after code is released.
+
+If there are any released {ProjectServer} updates, those will be included when performing this update.
```
ehelms, line 6: @AkshayGadhaveRH What is the proper way / convention to emphasize this point to users?

guides/common/modules/proc_updating-operating-system.adoc
ehelms: It's not really - but users don't understand that it can be used for either, and I figured a top level section dedicated to it increases its visibility and ability for users to understand it.

guides/common/modules/proc_updating-operating-system.adoc
evgeni, line 6: Should this also point out that updating OS-only (w/o Satellite) is a source for trouble and people shall not do that?

ehelms: Added some language let me know.

guides/common/modules/proc_updating-operating-system.adoc
```diff
@@ -0,0 +1,38 @@
+[id="updating-operating-system_{context}"]
+= Updating the operating system
```
lpramuk, line 2: title should read `Updating Satellite`   or  `Updating Satellite and operating system`
It is always update of `Satellite` + `OS`

guides/common/modules/proc_updating-operating-system.adoc
```diff
@@ -0,0 +1,38 @@
...
+
+{Team} recommends performing operating system updates regularly, because the minor releases patch security vulnerabilities and minor issues discovered after code is released.
+
+If there are any released {ProjectServer} updates, those will be included when performing this update.
+We do not recommend updating the operating system independently of {ProjectServer}.
```
evgeni, line 7: Title: Updating the operating system 
Text: We do not recommend updating the operating system 

Yes, that's not what you mean, but it can be read that way.

We should somehow emphasize that you should not update OS alone IF there are project updates. But not sure how to w/o making the sentence super long :/

guides/common/modules/proc_updating-operating-system.adoc
Griffin-Sullivan: The sections above have `Updating Satellite Server` already. I think it makes sense for a separate one to be `Updating the Operating System`, but can we get a capitalization for Operating System so it looks consistent.

guides/common/modules/proc_updating-operating-system.adoc
```diff
@@ -0,0 +1,38 @@
...
+Disconnected users should ensure that their installation is pointing at up to date {ProjectServer} repositories.
```
Griffin-Sullivan, line 8:
```suggestion
Disconnected users should ensure that their installation is pointing at up-to-date {ProjectServer} repositories.
```

guides/common/modules/proc_updating-operating-system.adoc
```diff
@@ -0,0 +1,38 @@
...
+For more information, see {AdministeringDocURL}backing-up-{project-context}-server-and-{smart-proxy-context}_admin[Backing Up {ProjectServer} and {SmartProxyServer}] in _{AdministeringDocTitle}_.
+
+.Procedure
+. Use the health check option to determine if the system is ready for update.
+On first use of this command, `{foreman-maintain}` prompts you to enter the hammer admin user credentials and saves them in the `/etc/foreman-maintain/foreman-maintain-hammer.yml` file.
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foreman-maintain} update check
+----
```
Griffin-Sullivan, line 21: Why are we mentioning health check but showing update check?

guides/common/modules/proc_updating-operating-system.adoc
ehelms: I want to emphasize rather that you *cannot* update just the OS if there are pending project updates

guides/doc-Updating_Project/master.adoc
```diff
@@ -28,3 +28,7 @@ endif::[]
 
 include::common/modules/proc_updating-smart-proxy.adoc[leveloffset=+1]
 endif::[]
+
+ifdef::satellite[]
```
evgeni, line 32: This should also work for Katello? (No idea about Foreman, especially on Debian).

guides/common/modules/proc_updating-operating-system.adoc
evgeni: Because `update check` is essentially a health check. Just not *the* health check `f-m health check`.

And probably because it's copied from the "updating project" chapter.

guides/common/modules/proc_updating-operating-system.adoc
evgeni, line 8: This should be under `Prerequisites`, IMHO.

guides/common/modules/proc_updating-operating-system.adoc
```diff
@@ -0,0 +1,38 @@
...
++
+Review the results and address any highlighted error conditions before performing the update.
+. Because of the lengthy update time, use a utility such as `tmux` to suspend and reattach a communication session.
+You can then check the update progress without staying connected to the command shell continuously.
++
+If you lose connection to the command shell where the update command is running, you can see the logged messages in the `{installer-log-file}` file to check if the process completed successfully.
```
evgeni, line 27: Technically people also should look at the maintain logs (and can re-attach to the tmux session).

But this is also a "bug" in the other chapters, why I was suggesting unification.

guides/common/modules/proc_updating-operating-system.adoc
asteflova: > Title: Updating the operating system Text: We do not recommend updating the operating system
> 
> Yes, that's not what you mean, but it can be read that way.

Yep, that's how I read it.

> We should somehow emphasize that you should not update OS alone IF there are project updates. But not sure how to w/o making the sentence super long :/

There might be a prerequisite in here somewhere: "Check that there are no project updates. <instructions on how to check> If there are,  do not update OS. <do XYZ instead>"

asteflova, line 7: Oh, this is tricky. I gave it a try based on your conversation here but, well, I don't know:

```suggestion
// Starting with an introduction that focuses on what the user action is and what benefits it brings:
Update your {ProjectServer} and its underlying operating system regularly.
Minor releases patch security vulnerabilities and issues discovered since the previous versions.

// Clarifying the additional information and what to avoid.
[IMPORTANT]
====
Always update the operating system together with updating your {ProjectServer}.
Updating the operating system includes all released {ProjectServer} updates.
====
```

Among other things, I suggest to switch from "do not update independently" to "always update together with" (if that's what you meant). A positive recommendation takes less effort on the user's side to understand than a negative one.

asteflova (commented):
After reading the conversation in this PR, I left a few suggestions that I think might help or that might be totally irrelevant :shrug: 

Either way, the majority of the steps you're adding seem to duplicate `proc_updating-server.adoc`. I'm not convinced that's the best way to go. Have you considered reusing the existing procedure (by applying ifdefs to distinguish between the server and OS bits) instead of duplicating it?

ehelms:
Based on the feedback, I've opted not to introduce a new guide section. Instead I have incorporated notes into the update guides related to the operating system update.

guides/common/modules/con_updating-project.adoc
Lennonka, line 1: This should probably be a procedure (`proc_*` file) and I don't see it included anywhere?

guides/common/modules/con_updating-project.adoc
```diff
@@ -0,0 +1,30 @@
+[id="updating-project_{context}"]
```
Lennonka, line 1: Please, add a module title and an introductory sentence or paragraph.

Lennonka (changes requested):
Suggestions.

guides/common/modules/proc_updating-server.adoc
```diff
@@ -3,6 +3,7 @@
 
 ifdef::satellite[]
 Update your connected {ProjectServer} to the next minor version.
+You can follow this process to update the underlying operating system between minor version releases.
```
Lennonka, line 6:
```suggestion
You can follow this process to update the underlying operating system between minor release versions.
```

guides/common/modules/proc_updating-server.adoc
Lennonka, line 1: I'm wondering if we have a pre-existing problem in this Procedure. In the disconnected update proc., we tell the user to sync 4 repos. Here, we tell them only about one and we don't even tell them to sync it. Is this alright?

Lennonka (changes requested):
Does it apply to disconnected Satellite as well? If so, you might want to mention it in `guides/common/modules/proc_updating-disconnected-server.adoc` as well.

Lennonka:
FYI, I've also filed #3121 to improve the guide.

guides/common/modules/proc_updating-server.adoc
ehelms: I think we are fine. I think we could drop from this procedure the mention of the maintenance repository because its redundant now a days.

Lennonka (approved):
LGTM

maximiliankolb (commented):
diff LGTM

Lennonka:
@ehelms Was this always true? Ie. should we cherry pick it to older versions?

PR #3077, opened 2024-06-17
Include latest two Debian and Ubuntu versions

maximiliankolb:
* second commit is to unify/simplify attributes to make it easier for future updates
* major/minor versions: start container, update packages, look into "/etc/os-release"

### DIFF
--- a/guides/common/assembly_configuring-provisioning-resources.adoc
+++ b/guides/common/assembly_configuring-provisioning-resources.adoc
@@ -7,33 +7,42 @@ include::modules/proc_setting-the-provisioning-context.adoc[leveloffset=+1]
 include::modules/proc_creating-operating-systems.adoc[leveloffset=+1]
 
 ifndef::satellite[]
-:os_codename: bookworm
-:os_major: 12
-:os_major_context: {os_major}
-:os_minor: 5
-:os_name: Debian
+:parent-DL: {DL}
+
+:DL: Debian
+:DL-codename: bookworm
+:DL-major: 12
+:DL-major-context: {DL-major}
+:DL-minor: 5
 include::modules/proc_creating-an-operating-system-for-debian.adoc[leveloffset=+1]
 
-:os_codename: bullseye
-:os_major: 11
-:os_major_context: {os_major}
-:os_minor: 9
-:os_name: Debian
+:DL: Debian
+:DL-codename: bullseye
+:DL-major: 11
+:DL-major-context: {DL-major}
+:DL-minor: 9
 include::modules/proc_creating-an-operating-system-for-debian.adoc[leveloffset=+1]
 
-:os_codename: noble
-:os_major: 24.04
-:os_major_context: 24-04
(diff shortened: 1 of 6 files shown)


ekohl (commented):
I recently learned Debian 11 standard security support ends in just over a month (2024-07-31). The LTS is still until 2026-06-30, but that's not support for all. As https://endoflife.date/debian states it:

> [Debian Long Term Support (LTS)](https://wiki.debian.org/LTS) is a project to extend the lifetime of all Debian stable releases to (at least) 5 years on [a limited set of architectures](https://lts-team.pages.debian.net/wiki/FAQ.html#what-architectures-are-supported). Debian LTS will not be handled by the Debian security team, but by a separate group of volunteers and companies. Not all packages of the Debian archive are supported by LTS, the [debian-security-support](https://wiki.debian.org/LTS/Using#Check_for_unsupported_packages) package can check for unsupported packages.

So do we really want to include Debian 11? I'd also think that users can simply follow the instructions of Debian 12 and modify where needed by themselves.

maximiliankolb:
@ekohl I would like to include it: considering Debian releases every ~2 years and they support a release for five years with LTS, always including the latest two versions seems like a sensible choice to me. The work to get the codename/versions/URLs is already done.

ekohl:
> * major/minor versions: start container, update packages, look into "/etc/os-release"

Or look at https://wiki.debian.org/DebianReleases#Production_Releases and https://wiki.ubuntu.com/Releases though I can understand that `VERSION_CODENAME` is already the short code. I think the easiest way is
```console
$ podman run --rm debian:12 /bin/sh -c '. /etc/os-release && echo $VERSION_CODENAME'
bookworm
```

> @ekohl I would like to include it: considering Debian releases every ~2 years and they support a release for five years with LTS, always including the latest two versions seems like a sensible choice to me. The work to get the codename/versions/URLs is already done.

I still think the massive list in importing content is really not the best way to maintain it and would think that an appendix with the parameters would be more useful (like we have appendix B to import kickstart repos). Also, we have foreman-ansible-modules which can set this up. Isn't this a great example to suggest that? You could even provide a reference playbook that creates all the content sources.

But I'll let others weigh in on this.

Lennonka (approved):
However, I don't have a problem with merging this. LGTM

ekohl:
To be clear, I don't object to merging this now since it's consistent. But I do want to raise the topic that we may want to revisit it.

maximiliankolb:
   2521ac33e0..7902238063  3.11 -> 3.11
   bf3f47de55..a0bccb06ec  3.10 -> 3.10
   063d35c393..218eebbe87  3.9 -> 3.9

orcharhino generally supports all supported Debian and Ubuntu versions; oss.atix.de provides a packaged and maintained `subscription-manager` for the last two releases. If there are any updates necessary, I will open a follow-up PR. It's definitely an area that we want to maintain & keep up to date. :rocket:

PR #3081, opened 2024-06-19
Add guidance for backups to NFS share

asteflova:
https://issues.redhat.com/browse/SAT-21300 requests adding "lightweight guidance" on backing up Project on an NFS share. This PR adds a short section to the procedure module for a full backup that covers the most important points of an NFS share backup.

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_performing-a-full-backup.adoc
+++ b/guides/common/modules/proc_performing-a-full-backup.adoc
@@ -43,6 +43,21 @@ You can use the `{foreman-maintain} backup` command to back up remote databases.
 
 You can use both online and offline methods to back up remote databases, but if you use offline method, the `{foreman-maintain} backup` command performs a database dump.
 
+.Backing up to a remote NFS share
+
+You can back up your {ProjectServer} or {SmartProxyServer} in a remote NFS share.
+
+* Make sure you export the share in your NFS server with the `no_root_squash` option to disable root squashing.
+For example:
+
+[options="nowrap", subs="+quotes,attributes"]
+----
+/backup/{project-context}/backup {foreman-example-com}(rw,sync,no_root_squash)
+----
+
+* Make sure the NFS share has the appropriate permissions set on the NFS server.
+Otherwise, {Project} might fail to save the backup to the NFS share.
+
 .Prerequisites
 * Ensure that your backup location has sufficient available disk space to store the backup.
 For more information, see xref:Estimating_the_Size_of_a_Backup_{context}[].
--- a/guides/common/modules/proc_performing-a-full-backup.adoc
+++ b/guides/common/modules/proc_performing-a-full-backup.adoc
@@ -45,7 +45,7 @@ You can use both online and offline methods to back up remote databases, but if 
 
 .Backing up to a remote NFS share
 
-To enable {Project} to save the backup to the NFS share, make sure that the `root` user of your {ProjectServer} or {SmartProxyServer} can write to the NFS share.
+To enable {Project} to save the backup to an NFS share, make sure that the `root` user of your {ProjectServer} or {SmartProxyServer} can write to the NFS share.
 NFS export options such as `root_squash` and `all_squash` are known to prevent this.
 
 For more information, see link:{RHELDocsBaseURL}9/html/configuring_and_using_network_file_services/deploying-an-nfs-server_configuring-and-using-network-file-services[{RHEL} _Configuring and using network files services_] and link:{RHELDocsBaseURL}9/html/securing_networks/securing-network-services_securing-networks#export-options-for-securing-an-nfs-server_securing-the-nfs-service[{RHEL} _Securing network services_].
--- a/guides/common/modules/proc_performing-a-full-backup.adoc
+++ b/guides/common/modules/proc_performing-a-full-backup.adoc
@@ -45,13 +45,10 @@ You can use both online and offline methods to back up remote databases, but if 
 
 .Backing up to a remote NFS share
 
-You can back up your {ProjectServer} or {SmartProxyServer} in a remote NFS share.
-To enable {Project} to save the backup to the NFS share, make sure you meet the following conditions:
-
-* The `root` user of your {ProjectServer} or {SmartProxyServer} can write to the NFS share.
+To enable {Project} to save the backup to the NFS share, make sure that the `root` user of your {ProjectServer} or {SmartProxyServer} can write to the NFS share.
 NFS export options such as `root_squash` and `all_squash` are known to prevent this.
+
 For more information, see link:{RHELDocsBaseURL}9/html/configuring_and_using_network_file_services/deploying-an-nfs-server_configuring-and-using-network-file-services[{RHEL} _Configuring and using network files services_] and link:{RHELDocsBaseURL}9/html/securing_networks/securing-network-services_securing-networks#export-options-for-securing-an-nfs-server_securing-the-nfs-service[{RHEL} _Securing network services_].
-* The NFS share has the appropriate permissions set on the NFS server.
 
 .Prerequisites
 * Ensure that your backup location has sufficient available disk space to store the backup.
--- a/guides/common/modules/proc_performing-a-full-backup.adoc
+++ b/guides/common/modules/proc_performing-a-full-backup.adoc
@@ -48,7 +48,9 @@ You can use both online and offline methods to back up remote databases, but if 
 To enable {Project} to save the backup to an NFS share, make sure that the `root` user of your {ProjectServer} or {SmartProxyServer} can write to the NFS share.
 NFS export options such as `root_squash` and `all_squash` are known to prevent this.
 
+ifndef::orcharhino[]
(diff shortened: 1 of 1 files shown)


asteflova:
Hi @ehelms @sayan3296 Can you please take a look and let me know if this PR covers https://issues.redhat.com/browse/SAT-21300? Would you change anything?

sayan3296:
> Hi @ehelms @sayan3296 Can you please take a look and let me know if this PR covers https://issues.redhat.com/browse/SAT-21300? Would you change anything?

While the no_root_squash part looks good, I am a bit concerned about [line 60 here](https://github.com/theforeman/foreman-documentation/pull/3081/files#diff-b7eeab4240730d5b3b826f51339be5c862e4afd491ff142dc37af06e77cf607bR60) i.e. 
```
the root user on the NFS client has root privileges on the NFS server
```

It probably should not be `NFS server` but we perhaps want it to be `NFS share` .

asteflova:
> ```
> the root user on the NFS client has root privileges on the NFS server
> ```
> 
> It probably should not be `NFS server` but we perhaps want it to be `NFS share` .

Thanks, addressed.

guides/common/modules/proc_performing-a-full-backup.adoc
```diff
@@ -43,6 +43,27 @@ You can use the `{foreman-maintain} backup` command to back up remote databases.
 
 You can use both online and offline methods to back up remote databases, but if you use offline method, the `{foreman-maintain} backup` command performs a database dump.
 
+.Backing up to a remote NFS share
+
+You can back up your {ProjectServer} or {SmartProxyServer} in a remote NFS share.
+
+* Make sure you export the share in your NFS server with the `no_root_squash` option to disable root squashing.
```
evgeni, line 50: I think that is bad advise, even if we add that warning at the end. How about:
```suggestion
* Make sure the `root` user of your {ProjectServer} or {SmartProxyServer} can write to the NFS share.
NFS export options like `root_squash` and `all_squash` are known to prevent this.
For more information, see https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/configuring_and_using_network_file_services/deploying-an-nfs-server_configuring-and-using-network-file-services
```

Especially, you can backup to `root_squash`ed NFS share just fine if the directory you write to is writable by `nobody` on the NFS server (and you use `--preserve-directory` to foreman-maintain, but that part is a bug in f-m that I am going to fix).

evgeni (changes requested):

guides/common/modules/proc_performing-a-full-backup.adoc
evgeni: https://github.com/theforeman/foreman_maintain/pull/888 for the code change

sayan3296: @asteflova I would say, 
* Have QE do some testing with some recommended options
* The test should be done for both backup as well as restore
* Once done and confirmed to be successfully happening, Document those options as primary recommendations

An end-user can export the NFS share with different combinations of options. So the best we can do is to provide:
* Minimum recommended options that would work
* Some NFS export options that may affect the functionality and should not be used

evgeni: Well, to be honest, my whole point is that we should NOT document ANY NFS options as we're not managing the NFS server. Especially as those options vary in meaning depending on the NFS implementation.

E.g. I have a NetApp NFS here, which I assume has `root_squash` on, as any file I create as `root` gets `nobody` as owner instead, but I can backup there just fine as `chown .postgres` works. When I replicate the setup with an EL9 NFS server, I get "Operation not permitted" for the `chown`.

sayan3296: Without any recommendations at all, The end-users run into permission errors and come back to RH for further support on resolving those matters. So if we mention something like NFS share can be used, we also should mention how they can confirm that the NFS share is usable.

So if not recommendations about exporting NFS share, We can document some points about how to check\test whether the exported share is usable for a backup\restore operation or not.

evgeni: > So if we mention something like NFS share can be used, we also should mention how they can confirm that the NFS share is usable.

That's why I wrote "Make sure the `root` user of your can write to the NFS share." as that's exactly the test for usability of the share. Which btw also applies to any other storage, not only NFS (it's just for local permission checks `root` usually has `DAC_OVERRIDE` capability and can ignore any constraints).

asteflova: Applied.

asteflova:
I did my best to capture the conversation in this PR. I'm keeping the details to the bare minimum, only mentioning the options that blocked the backup for the customer and linking to NFS docs for more information.

I'm not entirely convinced that this will help avoid customer cases in the future. It would be really nice if we could give users a complete procedure -- even if it would be just an example based on a specific NFS implementation. Perhaps we can expand on this PR later when/if another customer case arrives.

guides/common/modules/proc_performing-a-full-backup.adoc
```diff
@@ -45,24 +45,10 @@ You can use both online and offline methods to back up remote databases, but if
...
-
-[WARNING]
-====
-With `no_root_squash`, the root user on the NFS client has root privileges on the NFS share.
-Only use `no_root_squash` in a trusted environment.
-====
-
-* Make sure the NFS share has the appropriate permissions set on the NFS server.
-Otherwise, {Project} might fail to save the backup to the NFS share.
+For more information, see link:{RHELDocsBaseURL}9/html/configuring_and_using_network_file_services/deploying-an-nfs-server_configuring-and-using-network-file-services[{RHEL} _Configuring and using network files services_] and link:{RHELDocsBaseURL}9/html/securing_networks/securing-network-services_securing-networks#export-options-for-securing-an-nfs-server_securing-the-nfs-service[{RHEL} _Securing network services_].
```
maximiliankolb, line 51:
```suggestion
ifndef::orcharhino[]
For more information, see link:{RHELDocsBaseURL}9/html/configuring_and_using_network_file_services/deploying-an-nfs-server_configuring-and-using-network-file-services[{RHEL} _Configuring and using network files services_] and link:{RHELDocsBaseURL}9/html/securing_networks/securing-network-services_securing-networks#export-options-for-securing-an-nfs-server_securing-the-nfs-service[{RHEL} _Securing network services_].
endif::[]
```

guides/common/modules/proc_performing-a-full-backup.adoc
```diff
@@ -43,6 +43,13 @@ You can use the `{foreman-maintain} backup` command to back up remote databases.
 
 You can use both online and offline methods to back up remote databases, but if you use offline method, the `{foreman-maintain} backup` command performs a database dump.
 
+.Backing up to a remote NFS share
+
+To enable {Project} to save the backup to the NFS share, make sure that the `root` user of your {ProjectServer} or {SmartProxyServer} can write to the NFS share.
```
evgeni, line 48:
```suggestion
To enable {Project} to save the backup to an NFS share, make sure that the `root` user of your {ProjectServer} or {SmartProxyServer} can write to the NFS share.
```

Not sure about `a`/`an`, but there is no specific NFS share at that point, so `the` seems wrong :)

evgeni (approved):

guides/common/modules/proc_performing-a-full-backup.adoc
```diff
@@ -43,6 +43,15 @@ You can use the `{foreman-maintain} backup` command to back up remote databases.
 
 You can use both online and offline methods to back up remote databases, but if you use offline method, the `{foreman-maintain} backup` command performs a database dump.
 
+.Backing up to a remote NFS share
+
```
maximiliankolb, line 47:
```suggestion
```

guides/common/modules/proc_performing-a-full-backup.adoc
```diff
@@ -43,6 +43,15 @@ You can use the `{foreman-maintain} backup` command to back up remote databases.
...
+To enable {Project} to save the backup to an NFS share, make sure that the `root` user of your {ProjectServer} or {SmartProxyServer} can write to the NFS share.
```
maximiliankolb, line 48:
```suggestion
To enable {Project} to save the backup to an NFS share, ensure that the `root` user of your {ProjectServer} or {SmartProxyServer} can write to the NFS share.
```

maximiliankolb (approved):
two minor suggestions; rest LGTM!

maximiliankolb (approved):

asteflova:

79d9c470c8..0ddd814b10  3.11 -> 3.11
2a9cf5f5b6..1be025b261  3.10 -> 3.10
cea0ee3e4b..f7481de0ee  3.9 -> 3.9
0eb873b78f..03f0cc8c18  3.8 -> 3.8
8b44b4fdff..6549eb96ea  3.7 -> 3.7
fd2900225b..fad69f4f12  3.6 -> 3.6

I'm not cherry-picking to 3.5 due to a build error.

Issue #884, opened 2021-11-25
Proposal: Do not (by default) squash commits when merging


maximiliankolb:
Hi all,
after digging around in git log to find where, how, why, and who changed things, I -again- tripped over the default setting to squash commits when merging using the "Squash and merge" button in Github.

Some thoughts:
* One change should be one commit: for example, changing the links to a procedure in multiple files is one commit. Fixing a typo in the same file should be a separate commit.
* (Probably for CLI users only) You can use `!fixup` commits to indicate squashing when merging, for example using `git commit --fixup=head` to create a fixup commit for the last commit; or `git commit --fixup=MY_SHA1` to create a fixup commit for a specific commit. Note that you must only do fixup commits for commits that have not been merged yet, i.e. only for commits that are exclusively on your current branch. Run `git rebase -i master` to rebase your branch to master before pushing; git will automatically correct the order of "fixup commits" and squash (technically: fixup^^) them).
* If you are using software (read: not `git` in your console) that uses the same commit message as your first commit on your branch, then yes, it absolutely makes sense to squash those commits before merging; we do not want multiple commits like "Fix broken xref" merged to master.

I see a lot of advantages when using `git diff`, `git log`, and `git blame`. This also makes cherry-picking and reverting much easier and safer.

For me, it is no accident when I create a PR with more than one commit.

I would also like to enable "merge commits", but I fear this might be pushing it^^.

What are your thoughts? Does anyone else experience the same issues?

melcorr:
@lzap what do you think?

ekohl:
In my experience merging things (and caring about a good git history) the answer is (as always) "it depends".

When there is a single commit, I always use Rebase and merge if the commit message is alright. What's right, again, depends. In this project I'm less strict in what others do but I like to hold myself to high standards. That means describing what you did and, more importantly, why.

Sometimes the commit message is not great. If it's a single commit, I sometimes fix it up using Squash and merge. Which brings me to the next topic. I dislike the "Squash and merge" message that GH generates. I never like it as-is and if I do use the functionality, I always edit it to make sense. This also means stripping the PR number from the commit message. I also think the GH dialog doesn't write hard line endings which is also against the convention so I often write it in vim, format it and copy it back (back in the day I had a Firefox extension for this).

I also use Squash and merge if there's obvious fixups needed in the commits.

> For me, it is no accident when I create a PR with more than one commit.

[...]

PR #3082, opened 2024-06-20
Use RHDocsBaseURL where possible

ekohl:
This is the result of runnning:

```
sed -i 's|https://access.redhat.com/documentation/en-us/|{RHDocsBaseURL}|' $(rg https://access.redhat.com/documentation/ guides/common/modules/ -l)
```

This should make it easier to switch to https://docs.redhat.com/

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_using-insights-with-satellite-server.adoc
+++ b/guides/common/modules/proc_using-insights-with-satellite-server.adoc
@@ -14,7 +14,7 @@ To maintain your {ProjectServer}, and improve your ability to monitor and diagno
 .Scheduling insights-client
 
 Note that you can change the default schedule for running `insights-client` by configuring `insights-client.timer` on {Project}.
-For more information, see https://docs.redhat.com/en/documentation/red_hat_insights/1-latest/html/client_configuration_guide_for_red_hat_insights/assembly-client-changing-schedule[Changing the insights-client schedule] in the _Client Configuration Guide for Red Hat Insights_.
+For more information, see {RHDocsBaseURL}red_hat_insights/1-latest/html/client_configuration_guide_for_red_hat_insights/assembly-client-changing-schedule[Changing the insights-client schedule] in the _Client Configuration Guide for Red Hat Insights_.
 
 .Procedure
 
--- a/guides/common/modules/proc_using-insights-with-satellite-server.adoc
+++ b/guides/common/modules/proc_using-insights-with-satellite-server.adoc
@@ -14,7 +14,7 @@ To maintain your {ProjectServer}, and improve your ability to monitor and diagno
 .Scheduling insights-client
 
 Note that you can change the default schedule for running `insights-client` by configuring `insights-client.timer` on {Project}.
-For more information, see https://access.redhat.com/documentation/en-us/red_hat_insights/2020-10/html/client_configuration_guide_for_red_hat_insights/changing-the-client-schedule[Changing the insights-client schedule] in the _Client Configuration Guide for Red Hat Insights_.
+For more information, see https://docs.redhat.com/en/documentation/red_hat_insights/1-latest/html/client_configuration_guide_for_red_hat_insights/assembly-client-changing-schedule[Changing the insights-client schedule] in the _Client Configuration Guide for Red Hat Insights_.
 
 .Procedure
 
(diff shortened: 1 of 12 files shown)


asteflova:
It looks like you ended up with a double `en-US` in the new URLs. Can you try this instead?

```
sed -i 's|https://access.redhat.com/documentation/en-US/|{RHDocsBaseURL}|' $(rg https://access.redhat.com/documentation/ guides/common/modules/ -l)
```

ekohl:
Oops, updated.

guides/common/modules/proc_using-insights-with-satellite-server.adoc
```diff
@@ -14,7 +14,7 @@ To maintain your {ProjectServer}, and improve your ability to monitor and diagno
 .Scheduling insights-client
 
 Note that you can change the default schedule for running `insights-client` by configuring `insights-client.timer` on {Project}.
-For more information, see https://access.redhat.com/documentation/en-us/red_hat_insights/2020-10/html/client_configuration_guide_for_red_hat_insights/changing-the-client-schedule[Changing the insights-client schedule] in the _Client Configuration Guide for Red Hat Insights_.
+For more information, see {RHDocsBaseURL}red_hat_insights/2020-10/html/client_configuration_guide_for_red_hat_insights/changing-the-client-schedule[Changing the insights-client schedule] in the _Client Configuration Guide for Red Hat Insights_.
```
asteflova, line 17: This link gives a 404. I'm not familiar with Insights docs but I think this is the up-to-date one:
```suggestion
For more information, see {RHDocsBaseURL}red_hat_insights/1-latest/html/client_configuration_guide_for_red_hat_insights/assembly-client-changing-schedule[Changing the insights-client schedule] in the _Client Configuration Guide for Red Hat Insights_.
```

ekohl: The exact file rename looks like a regression while the version makes sense. Is that something we should let the insights team know?

asteflova: Ok, I pinged the Insights docs team in an internal chat and shared a link to this comment thread. I don't think it should block this PR; if there is something to change on the Insights docs' side, they know where to start investigating.

ekohl:
I noticed all pages redirect away from `en-us` so I added a commit that drops it. Waiting on the linkchecker to tell if that's correct.

guides/common/modules/proc_using-insights-with-satellite-server.adoc
asteflova: The Insights docs team has confirmed that the suggested link is correct and follows their file naming guidelines: https://docs.redhat.com/en/documentation/red_hat_insights/1-latest/html-single/client_configuration_guide_for_red_hat_insights/index#assembly-client-changing-schedule

ekohl:
If only Red Hat maintained stable documentation URLs ... but I think I have it now.

asteflova (approved):

Lennonka:
Cherry-picked:
- 7902238063..eadc12316a  3.11 -> 3.11

It would be useful to do that in all mainted versions.

ekohl:
The attribute commit itself wasn't picked so that's why I only checked 3.11, but I agree it makes sense

ekohl:
@Lennonka I'd have preferred the commits weren't squashed. They were explicitly separated.

Lennonka:
@ekohl Please, let us know next time in the PR description.

ekohl:
Sorry, that's just how I always work so I didn't think about it

Lennonka:
There's no need to apologize, but perhaps we should consider unifying the process, ie. clarifying when to squash and when not, how to label it, perhaps?

ekohl:
I thought we settled that in https://github.com/theforeman/foreman-documentation/issues/884.

Lennonka:
I wasn't even here when that discussion happened. How was I supposed to know? I still think we should revisit it. I'll put it on agenda for the next meeting.

asteflova:
How about highlighting this in the PR template? https://github.com/theforeman/foreman-documentation/pull/3096

PR #3083, opened 2024-06-20
Correct Puppetserver description in glossary

ekohl:
The Puppetserver service compiles manifests into a catalog. The agent retries this catalog and applies it. Manifests don't leave the Puppetserver process.

Puppetserver does a lot more, so this is still an inaccurate description but less than it used to be. For example, if the Certificate Authority component is enabled it also deals with certificates and it's a fileserver used by the Puppet Agent. The Puppet Agent also sends its facts and reports to Puppetserver.

We could also copy https://www.puppet.com/docs/puppet/8/server/about_server.html

> Puppet is configured in an agent-server architecture, in which a primary server node manages the configuration information for a fleet of agent nodes. Puppet Server acts as the primary server node. Puppet Server is a Ruby and Clojure application that runs on the Java Virtual Machine (JVM). Puppet Server runs Ruby code for compiling Puppet catalogs and for serving files in several JRuby interpreters. It also provides a certificate authority through Clojure.

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/ref_glossary-of-terms-used-in-project.adoc
+++ b/guides/common/modules/ref_glossary-of-terms-used-in-project.adoc
@@ -70,7 +70,7 @@ endif::[]
 [[Catalog]]
 Catalog:: A document that describes the desired system state for one specific host managed by Puppet.
 It lists all of the resources that need to be managed, as well as any dependencies between those resources.
-Catalogs are compiled by a Puppet server from Puppet Manifests and data from Puppet Agents.
+Catalogs are compiled by a Puppet server from Puppet Manifests and data from Puppet agents.
 
 ifdef::katello,orcharhino,satellite[]
 [[Candlepin]]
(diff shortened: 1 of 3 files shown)


maximiliankolb (approved):
LGTM based on the linked source :+1:

guides/common/modules/ref_glossary-of-terms-used-in-project.adoc
```diff
@@ -422,7 +422,7 @@ The files contain code to define a set of necessary resources, such as packages,
 Do not confuse with xref:Manifest[Manifest (Red{nbsp}Hat subscription manifest)].
 
 [[Puppet_server]]
-Puppet server:: A {SmartProxyServer} component that provides Puppet Manifests to hosts for execution by the Puppet Agent.
+Puppet server:: A {SmartProxyServer} component that provides a Puppet catalog to hosts for execution by the Puppet Agent.
```
Lennonka, line 425:
```suggestion
Puppet server:: A {SmartProxyServer} component that provides a Puppet catalog to hosts for execution by the Puppet agent.
```
I don't think this has to be capitalized.

apinnick: @ekohl If you make this small change, I think it should be fine for merging.

ekohl: Then I think it should also be done in other places, so I added a commit that does that first.

maximiliankolb (approved):
Thanks for fixing

apinnick (approved):

apinnick:
Changes merged to 'master' and cherry-picked:

06c3175d7c..5a2f7e86f0  3.11 -> 3.11
9ae458544a..f2609aa91b  3.10 -> 3.10

PR #3088, opened 2024-06-24
Change 4.12 (RC) to 4.13 (RC)

ianballou:
Attempts to fix the following:
image

Foreman 3.11 should pair with Katello 4.13, not 4.12.


ianballou:
It doesn't seem to fix the issue in the generated changes, which I'm confused about.

ekohl (approved):
I vaguely recall that we don't generate the web part in the preview.

PR #2866, opened 2024-03-06
Introduce SOE content

asteflova:
_I'm trying to find a way to introduce the [SOE content](https://github.com/theforeman/foreman-documentation/commit/4d99fe84535ab38bf04dc823b33bdaede4fc9b12) to the upstream docs set in a way that's useful, maintainable, and doesn't duplicate existing content. This PR is a WIP effort to 'kind of' finalize the first of the 10 steps of the SOE guide and to figure out a way to configure a preview._

The SOE guide is a few years old and heavily outdated. The goal is to update it, modularize it, and publish upstream. But there are caveats:

* “Just” copy-pasting the contents of the SOE guide and updating it with the latest functionality would introduce a lot of duplication into the docs set. A lot of the topics in the SOE guide are already covered by existing modules.
* Our current docs set provides a rich repository of modules on various useful topics; _but finding the right modules at the right times might sometimes be difficult_ – and this is something a new guide based on SOE could solve. For example:
  * SOE Step One covers setting up DNS, logging in to web UI, using Hammer CLI, an overview of organizations + how to create a custom organization – all in one chapter, in sequence.
  * In the existing docs set, these topics are covered over 3 different guides and the respective modules are included among a huge number of other modules so it’s not easy for users to find them.

The benefit of bringing SOE into the fold could be to provide a more focused document that would be designed to guide users through the process of deploying Foreman after the initial installation, but would not cover all the features of a given topic. A guide that is:
* Very user-focused
* Based on scenarios
* Limited to the most common workflows (linking to the less common workflows when needed)
* Explaining the when and the why

Contents of this PR:

* master.adoc for a new guide (proposals for a good title would be appreciated)
* A list of assemblies (created specifically for SOE) that include several existing modules (pre-edited in previous PRs to fit into SOE)
* Overall: The first step of a curated content path that users can follow to set up their environment without having to sift through large guides.

Target audience: Users who have recently installed a Foreman server and are looking for further steps on how to further develop their infrastructure.
* What they want: Easy-to-follow steps and examples (“So I’ve just installed a Foreman server, and now what?” or “So I can see that Foreman supports creating organizations, but how can I make the most out of them?”) with links to additional deep-dive docs that they can follow if they choose to.
* What they don’t want: To be overwhelmed with too much information at this point (either in the form of heavy-to-read conceptual modules or a vast repertoire of all the configuration options Foreman offers).

[...]

PR #3090, opened 2024-06-25
Move ifdef to hide non-katello content to module

asteflova:
This came up in [another PR](https://github.com/theforeman/foreman-documentation/pull/2866#discussion_r1535410858). When an ifdef is included in the assembly file, we need to duplicate the ifdef every time the module in question gets reused. But if the ifdef was included in the module file itself, we would not need to duplicate it.

Any objections? With this PR, I propose to change the location of one particular ifdef that I need for another PR. If anyone feels like we should do it for the whole repo, that would need to be handled in a separate PR.

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/assembly_accessing-server.adoc
+++ b/guides/common/assembly_accessing-server.adoc
@@ -2,9 +2,7 @@ include::modules/con_accessing-server.adoc[]
 
 include::modules/con_web-ui-overview.adoc[leveloffset=+1]
 
-ifdef::katello,orcharhino,satellite[]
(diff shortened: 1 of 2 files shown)


maximiliankolb:
I would like to keep it as is and actually have a commit downstream to go the other way: ensure that the first line is the anchor, no `ifdef` macro.

In my mind, it totally makes sense to shift the responsibility to the guide and/or assembly. If someone wants to include a particular file, they can. They'd have to verify that all content they include works for all build targets.

asteflova:
Alright, thanks for your input everybody!

PR #3109, opened 2024-07-03
Various small extensions for orcharhino

maximiliankolb:

### DIFF
--- a/guides/common/modules/proc_registering-a-host.adoc
+++ b/guides/common/modules/proc_registering-a-host.adoc
@@ -113,16 +113,25 @@ endif::[]
 This can be set by the `host_packages` parameter.
 . Optional: Select the *Update packages* option to update all packages on the host upon registration.
 This can be set by the `host_update_packages` parameter.
-. Optional: In the *Repository* field, enter a repository to be added before the registration is performed.
+. In the *Repositories* field, click *Add repositories for registration*.
++
+On the *Repository list* window, add content that is required before performing the registration.
 For example, it can be useful to make the `subscription-manager` package available for the purpose of the registration.
++
+* In the *Repository* field, enter a repository to be added before the registration is performed.
 ifdef::client-content-dnf[]
 For Red Hat family distributions, enter the URL of the repository, for example `\http://rpm.example.com/`.
 endif::[]
 ifdef::client-content-apt[]
 For Debian OS families, enter the whole line of list file content, for example `deb \http://deb.example.com/ bookworm 1.0`.
 endif::[]
-. Optional: In the *Repository GPG key URL* field, specify the public key to verify the signatures of GPG-signed packages.
+* Optional: In the *Repository GPG key URL* field, specify the public key to verify the signatures of GPG-signed packages.
 It needs to be specified in the ASCII form with the GPG public key header.
+ifdef::katello,orcharhino,satellite[]
+
++
+You do not have to specify repositories if you provide them in an activation key.
+endif::[]
 . Optional: In the *Token lifetime (hours)* field, change the validity duration of the JSON Web Token (JWT) that {Project} uses for authentication.
 The duration of this token defines how long the generated `curl` command works.
 You can set the duration to 0{range}999 999 hours or unlimited.
--- a/guides/common/modules/snip_host-registration-steps.adoc
+++ b/guides/common/modules/snip_host-registration-steps.adoc
@@ -8,8 +8,8 @@ For more information, see xref:Setting_the_Load_Balancer_for_Host_Registration_{
 endif::[]
 
 .Procedure
-. In the {ProjectWebUI}, navigate to *Hosts > Register Host*.
-ifeval::["{context}" == "load-balancing"]
+. In the {ProjectWebUI}, navigate to *Hosts* > *Register Host*.
+ifdef::load-balancing[]
 . From the *{SmartProxy}* dropdown list, select the load balancer.
 . Select *Force* to register a host that has been previously registered to a {SmartProxyServer}.
 endif::[]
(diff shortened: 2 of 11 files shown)


guides/common/modules/proc_creating-an-installation-medium-for-debian.adoc
```diff
@@ -11,7 +11,7 @@ endif::[]
 . In the {ProjectWebUI}, navigate to *Hosts* > *Provisioning Setup* > *Installation Media*.
 . Click *Create Medium* to create an installation medium entry.
 . Enter a *Name* for the installation medium.
-. Set the *Path* to the upstream URL `{client-debian-installation-medium-path}`.
```
Lennonka, line 14: I think that paths should be represented as literals. What is the reason not to?

guides/common/modules/snip_host-registration-steps.adoc
```diff
@@ -29,6 +29,11 @@ endif::[]
 ifdef::client-content-apt[]
 * On Debian: Check the `/etc/apt/sources.list` file and ensure that the appropriate repositories have been enabled.
 endif::[]
+ifeval::["{context}" == "load-balancing"]
```
Lennonka, line 32: I suggest `ifdef::load-balancing[]` for guide-specific conditions.
However, I see it's a pre-existing problem for other occurences, so consider this non-blocking.

guides/common/modules/proc_registering-a-host.adoc
Lennonka, line 1: Unfortunately, I don't have a working Foreman testing instance to verify it this week. But style-wise it looks good.

Lennonka (changes requested):
Generally LGTM

Lennonka:
Please, rebase.

guides/common/modules/proc_registering-a-host.adoc
maximiliankolb: Asking a colleague for confirmation.

maximiliankolb (commented):
Rebased to "master" and applied all feedback.

guides/common/modules/proc_registering-a-host.adoc
nadjaheitmann: Content-wise it also looks good. Here is how it looks like in the UI:

image

image

guides/common/modules/snip_host-registration-steps.adoc
Lennonka: That would be nice to have, I agree. If you have time.

guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -113,15 +113,19 @@ endif::[]
 This can be set by the `host_packages` parameter.
 . Optional: Select the *Update packages* option to update all packages on the host upon registration.
 This can be set by the `host_update_packages` parameter.
-. Optional: In the *Repository* field, enter a repository to be added before the registration is performed.
+. In the *Repositories* field, click the *+* symbol.
```
Lennonka, line 116:
```suggestion
. In the *Repositories* field, click *+ Add repositories for registration*.
```
Or even without the "+", as you prefer. I'm assuming that the text is clickable.

guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -113,15 +113,19 @@ endif::[]
...
+* In the *Repository* field, enter a repository to be added before the registration is performed.
 ifdef::client-content-dnf[]
 For Red Hat family distributions, enter the URL of the repository, for example `\http://rpm.example.com/`.
 endif::[]
 ifdef::client-content-apt[]
 For Debian OS families, enter the whole line of list file content, for example `deb \http://deb.example.com/ bookworm 1.0`.
 endif::[]
-. Optional: In the *Repository GPG key URL* field, specify the public key to verify the signatures of GPG-signed packages.
+* Optional: In the *Repository GPG key URL* field, specify the public key to verify the signatures of GPG-signed packages.
 It needs to be specified in the ASCII form with the GPG public key header.
```
Lennonka, line 129:
```suggestion
It needs to be specified in the ASCII form with the GPG public key header.

+
You do not have to specify repositories if you provide them in an activation key.
```
I know this is out of PR scope and I'm not sure if my wording is true/correct, but I think that there are cases when the user doesn't have to specify any repos. It would be good to mention it.

Lennonka (changes requested):
A couple of suggestions.

guides/common/modules/snip_host-registration-steps.adoc
Lennonka: One more thing - I'm not sure since which version the attribute is present. So it's a good idea to check that when cherry-picking.

guides/common/modules/proc_registering-a-host.adoc
maximiliankolb: Applied without `+`. Looks cleaner & more readable this way. Text is clickable.

guides/common/modules/proc_registering-a-host.adoc
maximiliankolb: Applied.

maximiliankolb (commented):
Rebased to "master" and applied all feedback. Ready for re-review @Lennonka

guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -113,16 +113,23 @@ endif::[]
...
+
++
+You do not have to specify repositories if you provide them in an activation key.
```
Lennonka, line 132: And this is probably Katello only, so it should be enveloped by a condition>
```suggestion
ifdef::katello,orcharhino,satellite[]

+
You do not have to specify repositories if you provide them in an activation key.
endif::[]
```

Lennonka (approved):
Otherwise looks good! Just that condition perhaps ^^

guides/common/modules/proc_registering-a-host.adoc
maximiliankolb: Applied.

maximiliankolb (commented):
Applied the last suggestion; will merge with green GHA.

Thanks for the reviews Lena.

maximiliankolb:
   b5e67c98a4..5a18158143  3.11 -> 3.11
   313a774c9e..0f7088b781  3.10 -> 3.10
   41f5585184..13603cd39e  3.9 -> 3.9

PR #3119, opened 2024-07-11
Move TemplateSync and hide enablement

Lennonka:
The chapter doesn't belong into _Managing hosts_, but into _Administering Project_.
And we can hide plug-in enablement for Sat, because it's enabled by default.

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/assembly_synchronizing-template-repositories.adoc
+++ b/guides/common/assembly_synchronizing-template-repositories.adoc
@@ -1,6 +1,8 @@
 include::modules/con_synchronizing-template-repositories.adoc[]
 
+ifndef::satellite[]
 include::modules/proc_enabling-the-templatesync-plugin.adoc[leveloffset=+1]
+endif::[]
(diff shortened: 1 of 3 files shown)


maximiliankolb (approved):
diff LGTM but I'd also like to hear more input about moving the assembly.

My concern is that "Administering Foreman" is already too big even before this PR.

Lennonka:
> diff LGTM but I'd also like to hear more input about moving the assembly.

More input as in reasoning behind moving to Administering?

Well, the procedures performed in this assembly do not relate to hosts, but to all kinds of templates. The procedures solve how to manage versions of templates and this is the most likely guide where it could have a place.

> My concern is that "Administering Foreman" is already too big even before this PR.

I hear you, but we can split this guide at a later time, once we have the capacity to do that. At this time, we focus on improving Provisioning hosts and Managing hosts.

PR #3123, opened 2024-07-11
Split proc_configuring-repositories.adoc into several files

ekohl:
The procedure to enable repositories is very specific to every build flavor. Trying to squeeze this into a single file makes it very complicated. This makes maintenance easier and doesn't really duplicate that much.

This PR is split off from https://github.com/theforeman/foreman-documentation/pull/3104 to verify it's just an internal refactor and doesn't have any impact on the resulting guides. It may also be easier to review this standalone, but I'll leave that up to others.

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/assembly_configuring-repositories.adoc
+++ b/guides/common/assembly_configuring-repositories.adoc
@@ -1,0 +1,3 @@
+include::modules/con_configuring-repositories.adoc[]
+
+include::modules/proc_configuring-repositories-{build}.adoc[]
--- a/guides/common/modules/proc_configuring-repositories-foreman-deb.adoc
+++ b/guides/common/modules/proc_configuring-repositories-foreman-deb.adoc
@@ -1,0 +1,24 @@
+.Procedure
+Select the operating system and version you are installing on:
+
+* xref:#repositories-debian-11[Debian 11 (Bullseye)]
+* xref:#repositories-ubuntu-2004[Ubuntu 20.04 (Focal)]
+* xref:#repositories-ubuntu-2204[Ubuntu 22.04 (Jammy)]
+
+[id="repositories-debian-11"]
+== Debian 11 (Bullseye)
+
+:distribution-codename: bullseye
+include::proc_configuring-repositories-deb.adoc[]
+
+[id="repositories-ubuntu-2004"]
+== Ubuntu 20.04 (Focal)
+
+:distribution-codename: focal
+include::proc_configuring-repositories-deb.adoc[]
+
+[id="repositories-ubuntu-2204"]
+== Ubuntu 22.04 (Jammy)
+
+:distribution-codename: jammy
+include::proc_configuring-repositories-deb.adoc[]
--- a/guides/common/assembly_installing-capsule-server.adoc
+++ b/guides/common/assembly_installing-capsule-server.adoc
@@ -21,7 +21,7 @@ ifdef::satellite[]
 include::modules/proc_configuring-repositories-proxy.adoc[leveloffset=+1]
 endif::[]
 ifndef::satellite[]
-include::modules/proc_configuring-repositories.adoc[leveloffset=+1]
+include::assembly_configuring-repositories.adoc[leveloffset=+1]
 endif::[]
 
 // we do not package "fapolicy" for Foreman on Debian/Ubuntu
--- a/guides/common/assembly_installing-server-connected.adoc
+++ b/guides/common/assembly_installing-server-connected.adoc
@@ -12,7 +12,7 @@ include::modules/proc_registering-to-red-hat-subscription-management.adoc[levelo
 include::modules/proc_attaching-satellite-infrastructure-subscription.adoc[leveloffset=+1]
 endif::[]
 
-include::modules/proc_configuring-repositories.adoc[leveloffset=+1]
+include::assembly_configuring-repositories.adoc[leveloffset=+1]
 
 // we do not package "fapolicy" for Foreman on Debian/Ubuntu
 ifndef::foreman-deb[]
--- a/guides/common/assembly_preparing-smartproxyservers-for-load-balancing.adoc
+++ b/guides/common/assembly_preparing-smartproxyservers-for-load-balancing.adoc
@@ -10,7 +10,7 @@ ifdef::satellite[]
 include::modules/proc_configuring-repositories-proxy.adoc[leveloffset=+1]
 endif::[]
 ifndef::satellite[]
-include::modules/proc_configuring-repositories.adoc[leveloffset=+1]
+include::assembly_configuring-repositories.adoc[leveloffset=+1]
 endif::[]
 
 // Installing {SmartProxyServer} Packages
--- a/guides/common/modules/con_configuring-repositories.adoc
+++ b/guides/common/modules/con_configuring-repositories.adoc
@@ -1,0 +1,2 @@
+[id="configuring-repositories_{context}"]
+= Configuring repositories
--- a/guides/common/modules/proc_configuring-repositories-foreman-el.adoc
+++ b/guides/common/modules/proc_configuring-repositories-foreman-el.adoc
@@ -1,0 +1,46 @@
+.Procedure
+Select the operating system and version you are installing on:
+
+* xref:#repositories-el-9[{EL} 9]
+* xref:#repositories-el-8[{EL} 8]
+
+[id="repositories-el-9"]
+== {EL} 9
+
+:distribution: el
+:distribution-major-version: 9
+:package-manager: dnf
+
+include::proc_configuring-repositories-el.adoc[]
+
+include::snip_verification-enabled-repolist.adoc[]
+
+[id="repositories-el-8"]
+== {EL} 8
+
+:distribution: el
+:distribution-major-version: 8
+:package-manager: dnf
+
+include::proc_configuring-repositories-el.adoc[]
+
+. Enable the DNF modules:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+# dnf module enable {dnf-module}
+----
++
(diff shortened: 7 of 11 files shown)


ekohl:
> This PR is split off from #3104 to verify it's just an internal refactor and doesn't have any impact on the resulting guides. It may also be easier to review this standalone, but I'll leave that up to others.

After fixing one mistake, you can now see that the diff is empty. Just as intended.

guides/common/modules/proc_configuring-repositories.adoc
```diff
@@ -4,127 +4,12 @@
 ifdef::orcharhino[]
 Ensure the repositories required to install {ProductName} are enabled on your {EL} host.
 endif::[]
```
maximiliankolb, line 6: Please also move this to a separate file.

maximiliankolb (commented):
Please rename the file to a) use "snip_" as prefix because the files don't contain any heading; and b) maybe use the build target in the file name: example: `snip_configuring-repositories-katello.adoc`. Then, you can always, without any `ifdef` macro, include `snip_configuring-repositories-{build}.adoc` similar to https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes.adoc?plain=1#L51

guides/common/modules/proc_configuring-repositories.adoc
ekohl: I had that at first, but since it's just a single line I questioned if it was useful. Do you have plans to expand this or do you prefer it for consistency?

guides/common/modules/proc_configuring-upstream-rpm-repositories.adoc
ekohl, line 1: I debated `upstream-rpm` and `upstream-el`. Thoughts?

guides/common/modules/proc_configuring-satellite-repositories.adoc
```diff
@@ -0,0 +1,70 @@
...
+====
+If there is any warning about conflicts with Ruby or PostgreSQL while enabling `{dnf-module}` module, see
+ifeval::["{context}" == "{project-context}"]
+xref:troubleshooting-dnf-modules_{context}[].
+endif::[]
+ifeval::["{context}" != "{project-context}"]
+{InstallingServerDocURL}troubleshooting-dnf-modules_{project-context}[Troubleshooting DNF modules] in _{InstallingServerDocTitle}_.
+endif::[]
+For more information about modules and lifecycle streams on {RHEL} 8, see https://access.redhat.com/support/policy/updates/rhel-app-streams-life-cycle[{RHEL} Application Streams Lifecycle].
+====
```
ekohl, line 68: This note is duplicated for `upstream-rpm`. I think that's acceptable and Foreman 3.13 will drop EL8 support anyway. For now we don't have modules on EL 9.

asteflova:
I'm not a huge fan of the idea to split off intros from procedure modules. As far as contributor experience goes, that's a change for the worse. Then again, I'm not a fan of splitting off assembly introductions into separate files either, and I still found a way to live with that, so... :)

The idea behind modularity is to enable reuse -- so you're supposed to split off "reusable" pieces of content, pieces that can stand on their own and that make sense on their own[1]. Because you can't reuse `.Procedure` without its intro and vice versa, then as far as the modular documentation framework goes, this would mean bending the rules quite a bit. Whether that's a problem or not depends on the value that it would bring to implementing tabs.

[1] "A module must make sense and provide value on its own, even when read separately from the other modules. The templates included in this manual help ensure this."
https://redhat-documentation.github.io/modular-docs/#what-modular-documentation-is-not

ekohl:
I agree `guides/common/modules/proc_configuring-repositories.adoc` has become more of an assembly. Should we convert it to an assembly and then include that assembly in multiple places?

> The idea behind modularity is to enable reuse -- so you're supposed to split off "reusable" pieces of content, pieces that can stand on their own and that make sense on their own[1]. Because you can't reuse `.Procedure` without its intro and vice versa, then as far as the modular documentation framework goes, this would mean bending the rules quite a bit. Whether that's a problem or not depends on the value that it would bring to implementing tabs.

While it's true, there is also a point of maintainability. In effect the modules really aren't reusable between build flavors (except `foreman-el` and `katello`).

Another way we could structure it is to create `modules/proc_configuring-repositories-{{ build }}.adoc` and then symlink `modules/proc_configuring-repositories-katello.adoc` to `modules/proc_configuring-repositories-foreman-el.adoc`.

Then `guides/common/assembly_configuring-repositories.adoc` becomes:
```asciidoc
[id="configuring-repositories_{context}"]
= Configuring repositories

include::modules/proc_configuring-repositories-{build}.adoc[]
```

ekohl:
Updated as discussed.

ekohl:
> For now, we don't nest assemblies AFAIK and I'd like to keep it like this

Yesterday during the meeting it was mentioned that was can be allowed. I interpreted it as something we can do. But I'm fine with refactoring.

asteflova:
I might not know what others mean by nested assemblies, but I can see plenty of them. One example: [13.3. Using repository sources](https://docs.theforeman.org/nightly/Administering_Project/index-katello.html#using-repository-sources_admin). So basically anything like [ASSEMBLY 3](https://redhat-documentation.github.io/modular-docs/#what-modular-documentation-is).

I'm not a fan, but I haven't been able to provide reasons for avoiding them that people who _are_ fans would consider valid. (Yet. I'm not giving up.)

maximiliankolb:
> I might not know what others mean by nested assemblies, but I can see plenty of them. One example: [13.3. Using repository sources](https://docs.theforeman.org/nightly/Administering_Project/index-katello.html#using-repository-sources_admin). So basically anything like [ASSEMBLY 3](https://redhat-documentation.github.io/modular-docs/#what-modular-documentation-is).
> 
> I'm not a fan, but I haven't been able to provide reasons for avoiding them that people who _are_ fans would consider valid. (Yet. I'm not giving up.)

I was wrong: `cat guides/common/assembly_* | grep assembly` returns plenty of results; I thought there would be none. Therefore, this approach has been used before and we should allow PRs that use it. level offset is something different than nesting to me and independent, isn't it?

guides/common/modules/proc_configuring-repositories-foreman-deb.adoc
```diff
@@ -0,0 +1,24 @@
+.Procedure
+Select the operating system and version you are installing on:
+
+* xref:#repositories-debian-11[Debian 11 (Bullseye)]
+* xref:#repositories-ubuntu-2004[Ubuntu 20.04 (Focal)]
+* xref:#repositories-ubuntu-2204[Ubuntu 22.04 (Jammy)]
```
asteflova, line 6: This should still give no diff, right? If so, it's another way to make maintenance easier.

```suggestion
* xref:#repositories-debian-11[]
* xref:#repositories-ubuntu-2004[]
* xref:#repositories-ubuntu-2204[]
```

asteflova, line 6: These custom tables of contents (ToCs) are kind of frowned upon. It's a maintenance burden (What if someone adds another OS? Will they remember to update the list? My experience says that they often don't.) and it duplicates the regular ToC.

However, if the goal is to enable the implementation of tabs, I assume you wouldn't even need to use a list like this. So if this is just one of the steps on the path to https://github.com/theforeman/foreman-documentation/pull/3104, feel free to ignore this comment.

guides/common/assembly_configuring-repositories.adoc
```diff
@@ -0,0 +1,4 @@
+[id="configuring-repositories_{context}"]
+= Configuring repositories
```
asteflova, line 2: Assemblies are currently only containers for includes, no content gets included (not even a heading). The first include in an assembly is usually a con_*.adoc that includes a heading and an assembly introduction.

guides/common/modules/proc_configuring-repositories-foreman-deb.adoc
ekohl: > However, if the goal is to enable the implementation of tabs, I assume you wouldn't even need to use a list like this. So if this is just one of the steps on the path to #3104, feel free to ignore this comment.

That is exactly my goal: I want to get rid of them and my tabs PR intends to do that. This PR is just about moving things so it's obvious we're not touching downstream when we do that.

maximiliankolb:
GHA fail: asciidoctor: ERROR: common/assembly_configuring-repositories.adoc: line 4: include file not found: /home/runner/work/foreman-documentation/foreman-documentation/guides/doc-Configuring_Load_Balancer/common/modules/proc_configuring-repositories-katello.adoc

guides/common/assembly_configuring-repositories.adoc
ekohl: I introduced `guides/common/modules/con_configuring-repositories.adoc`. Is that what you intended?

ekohl:
> GHA fail: asciidoctor: ERROR: common/assembly_configuring-repositories.adoc: line 4: include file not found: /home/runner/work/foreman-documentation/foreman-documentation/guides/doc-Configuring_Load_Balancer/common/modules/proc_configuring-repositories-katello.adoc

Solved with a symlink

asteflova (approved):

guides/common/assembly_configuring-repositories.adoc
```diff
@@ -0,0 +1,2 @@
+include::modules/con_configuring-repositories.adoc[]
```
maximiliankolb, line 1:
```suggestion
include::modules/con_configuring-repositories.adoc[]

```

maximiliankolb (approved):
one tiny suggestion. No diff in rendered HTML is exactly what we expect. So LGTM :+1:

PR #3135, opened 2024-07-16
Use a direct command to open a database shell

ekohl:
Rather than using various small steps, this uses a single command to run psql as the postgres user. You can also pass a database to open.

I'd like to note that this procedure in itself is problematic. This really isn't a normal procedure that any user can (or should) execute. Direct database access is frowned upon. Also, we should be encrypting the data at rest so that this is possible is IMHO a bug in itself.

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_connecting-to-an-amazon-ec2-instance-using-ssh.adoc
+++ b/guides/common/modules/proc_connecting-to-an-amazon-ec2-instance-using-ssh.adoc
@@ -10,23 +10,11 @@ However, to connect to any Amazon Web Services EC2 instance that you provision t
 ----
 # hammer compute-resource list
 ----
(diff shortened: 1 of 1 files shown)


maximiliankolb (approved):
Tested on Foreman Server 3.11 and a Smart Proxy Server. Command works; though I'm no expert with postgres.

ekohl:
To be clear: I'd also question if we need to keep this in at all. Direct database access is not something we want. Like, at all.

maximiliankolb:
> To be clear: I'd also question if we need to keep this in at all. Direct database access is not something we want. Like, at all.

Yeah, true. Sounds like something users/customers should avoid. But still: I consider your PR an improvement; so feel free to merge as is.

PR #3143, opened 2024-07-17
Unify the Client repo prerequisites

Lennonka:
These two prerequisites are very frequent and they deserve snippets so that the wording is consistent and we can edit it in a single place.

One is for registered hosts, the other for AKs.

* [x] Foreman 3.9/Katello 4.11 (Satellite 6.15; orcharhino 6.8/6.9)
* [x] Foreman 3.6/Katello 4.8


maximiliankolb (approved):
LGTM

guides/common/attributes-base.adoc
```diff
@@ -106,9 +106,10 @@
 :postgresql-conf-dir: {postgresql-data-dir}
 :postgresql-log-dir: {postgresql-data-dir}/log
 :PIV: PIV
-:project-client-name: https://yum.theforeman.org/client/{ProjectVersion}/
-:project-client-RHEL7-url: https://yum.theforeman.org/client/{ProjectVersion}/el7/x86_64/foreman-client-release.rpm
-:project-client-RHEL8-url: https://yum.theforeman.org/client/{ProjectVersion}/el8/x86_64/foreman-client-release.rpm
+:project-client-url: https://yum.theforeman.org/client/{ProjectVersion}
+:project-client-name: {project-client-url}[Foreman Client]
```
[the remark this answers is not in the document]
Lennonka: I don't think so. For both vanilla Foreman and Foreman+Katello, it's the same Foreman Client repository, isn't it?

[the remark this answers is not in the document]
maximiliankolb: Ah yes, you're correct. :+1:

guides/common/attributes-base.adoc
```diff
@@ -106,9 +106,10 @@
...
+:project-client-RHEL7-url: {project-client-url}/el7/x86_64/foreman-client-release.rpm
+:project-client-RHEL8-url: {project-client-url}/el8/x86_64/foreman-client-release.rpm
```
ekohl, line 112: Upstream we have EL repos so I'd prefer to rename them to `project-client-url-el$version`. That way you can simply grep for `project-client-url` and find all usages without having to resort to some regex.

guides/common/attributes-base.adoc
ekohl, line 110: I always thought it was confusing that `project-client-name` holds the URL in upstream where in Satellite it's actually a name. Adding a label looks like a decent workaround.

guides/common/modules/proc_enabling-the-project-client-name-repository.adoc
```diff
@@ -1,7 +1,7 @@
 [id="enabling-the-project-client-name-repository_{context}"]
 = Enabling the {project-client-name} repository
 
-Enable the {Team} {project-client-name} repository for every major version of {RHEL} that you intend to run on your hosts.
+Enable the {project-client-name} repository for every major version of {RHEL} that you intend to run on your hosts.
```
ekohl, line 4: Should we generalize this while we're at it? Shouldn't make a difference for Satellite, but makes things clearer for upstream.
```suggestion
Enable the {project-client-name} repository for every major version of {EL} that you intend to run on your hosts.
```

Lennonka:
When checking the outputs, I've realized that it's not that simple. There are actually two cases - one with AKs before registration or provisioning, and the second one for registered/provisioned hosts. So I'm going to remake it...

guides/common/attributes-satellite.adoc
```diff
@@ -155,7 +155,7 @@
 // Used in downstream guides
 :RepoRHEL7ServerSoftwareCollections: rhel-server-rhscl-7-rpms
 
-:project-client-name: Satellite Client 6
+:project-client-name: Red Hat Satellite Client 6
 :project-client-RHEL7-url: {RepoRHEL7ServerSatelliteToolsProjectVersion}
 :project-client-RHEL8-url: {RepoRHEL8ServerSatelliteToolsProjectVersion}
```
ekohl, line 160: To make it even more confusing, this attribute is the reverse of `project-client-name`: in upstream it's the URL where in downstream it's the name. I'm starting to get a feeling that we shouldn't attempt to reuse the attributes between upstream and downstream since they're fundamentally different and used in different contexts.

What if we undefine `project-client-RHEL7-url` in this guide (you can do that with asciidoc, right?) to make sure it errors out if it's used. Then simply use `RepoRHEL7ServerSatelliteToolsProjectVersion` where it is needed.

ekohl (commented):
Feel free to ignore my comments, but these are some long standing annoyances I have with the current attributes.

guides/common/attributes-satellite.adoc
Lennonka: I would actually prefer to remove the `*rhel*-url`s altogether and replace them with the Sat attributes. They're only used in 3 places in total.

guides/common/modules/snip_prerequisite-project-client-repository.adoc
```diff
@@ -0,0 +1,2 @@
+* {project-client-name} repository for the operating system version of the host is synchronized on {ProjectServer} and enabled in the activation key you use.
```
Lennonka, line 1: @ekohl Help me out here, please, if you can. Will this wording work also for hosts that are already registered? Assuming the host has an AK assigned, can I tell the user to enable the repo in the AK and it will available on the host without further steps?

I might be overthinking this, but I got stuck on it :laughing:

guides/common/attributes-base.adoc
Lennonka: As I say below, I would just do a quick process and annihilate these.

guides/common/attributes-satellite.adoc
ekohl: Well, that's sort of what I was suggesting. Except I had still assumed we needed them upstream. But you're right that the `rhel` part doesn't make sense since in upstream we prefer `el`.

guides/common/modules/proc_deploying-a-policy-on-a-host-using-ansible.adoc
```diff
@@ -15,7 +15,8 @@ ifndef::satellite[]
 For more information, see xref:Installing_the_OpenSCAP_plugin_{context}[].
 endif::[]
 include::snip_prerequisite-rhel-repositories.adoc[]
```
ekohl, line 17: This is really incorrect for upstream. In non-Katello you can't even sync content but in upstream you can't assume they run RHEL.

guides/common/modules/snip_prerequisite-project-client-repository.adoc
ekohl: I'm not an expert on content and normally look at @jeremylenz or @ianballou.

Normally I'd say something like this is at least needed.
```suggestion
ifdef::foreman-el[]
* {project-client-name} repository is available on the host.
endif::[]
ifndef::foreman-el[]
* {project-client-name} repository for the operating system version of the host is synchronized on {ProjectServer} and enabled in the activation key you use.
endif::[]
```

And the link below is also invalid for `foreman-el`.

guides/common/modules/proc_deploying-a-policy-on-a-host-using-ansible.adoc
Lennonka: I'm planning to work on this in another PR. Not the scope of this PR.

guides/common/modules/snip_prerequisite-project-client-repository.adoc
jeremylenz: @Lennonka No, if a host is already registered, changes to its activation key will have no effect on the host. I'd suggest maybe just "enabled for the host". I'm on the fence as to whether more elaboration is needed here about content overrides.

Lennonka: @jeremylenz Thank you! The content overrides would be [this one](https://docs.theforeman.org/nightly/Managing_Content/index-katello.html#Changing_the_Repository_Sets_Status_for-a-Host-in_foreman_content-management), right?

jeremylenz: Yep, that's where you'd set content overrides for a host. (They can also be set on activation keys, but note my original comment above.)

guides/common/modules/proc_deploying-a-policy-on-a-host-using-ansible.adoc
Lennonka: -> https://github.com/theforeman/foreman-documentation/pull/3159

guides/common/attributes-satellite.adoc
Lennonka: -> #3160

Lennonka:
Mind taking one more look?

maximiliankolb (approved):
ACK; handing over to Ewoud

Lennonka:
Rebased.

Since there were no further comments over the last two weeks, I'm merging it.

Lennonka:
Cherry-picked:
- 8df3d18869..170b9dab2b  3.11 -> 3.11
- 7800f5c976..ee3a20f55a  3.10 -> 3.10
- 4b62fe4c08..d5667abd72  3.9 -> 3.9
- afedcd515f..0a773fa7e2  3.8 -> 3.8
- 385d32af61..fe014ad22e  3.7 -> 3.7

I've decided not to cherrypick further because there are too many conflicts.

PR #3144, opened 2024-07-17
Change "product" to "Product"

apinnick:
Changed "product" to "Product" to align with RHSSG. Not sure whether we do the same for SUSE.

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_adding-an-scc-account.adoc
+++ b/guides/common/modules/proc_adding-an-scc-account.adoc
@@ -20,23 +20,23 @@ Add the GPG public key for SLES 15 SP3 from https://www.suse.com/support/securit
 By default, it is set to `\https://scc.suse.com`.
 . Optional: Set a *Sync interval* to periodically update the SCC authentication tokens.
 Note that this does not refer to synchronizing content to {Project}.
-. Optional: In the *Use GPG key for SUSE products* field, you can select the previously created content credential to automatically add a GPG public key to your SUSE products.
+. Optional: In the *Use GPG key for SUSE products* field, you can select the previously created content credential to automatically add a GPG public key to your SUSE Products.
 `zypper` automatically verifies the signatures of each software package to ensure their authenticity.
 +
 [NOTE]
 ====
 You can also set the GPG public key for repositories from SUSE at a later stage.
-However, changing it does not affect already synchronized products.
-If you already have synchronized products in {Project}, navigate to *Content > Products* and replace the GPG key in each respective product.
+However, changing it does not affect already synchronized Products.
+If you already have synchronized Products in {Project}, navigate to *Content > Products* and replace the GPG key in each respective Product.
 ====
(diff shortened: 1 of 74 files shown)


ekohl (commented):
As I mentioned in private: I think we should treat `product` the same as `content view` and `lifecycle environment`. Perhaps this is my engineer's view, but all of them are models in Katello. Since it was decided that `content view` and `lifecycle environment` should be lower cased, I think `product` should also be lower case.

guides/common/modules/proc_adding-an-scc-account.adoc
```diff
@@ -20,23 +20,23 @@ Add the GPG public key for SLES 15 SP3 from https://www.suse.com/support/securit
 By default, it is set to `\https://scc.suse.com`.
 . Optional: Set a *Sync interval* to periodically update the SCC authentication tokens.
 Note that this does not refer to synchronizing content to {Project}.
-. Optional: In the *Use GPG key for SUSE products* field, you can select the previously created content credential to automatically add a GPG public key to your SUSE products.
+. Optional: In the *Use GPG key for SUSE products* field, you can select the previously created content credential to automatically add a GPG public key to your SUSE Products.
```
asteflova, line 23: Stating the obvious: product/Product is capitalized inconsistently on this line. I don't have a VM with the right plug-in available so I can't check the context but it does look like "Product" is the right choice here too. Perhaps we could change the label of that field in the web UI? I suppose the PR should be simple, except that I can't find the right place in the code.

maximiliankolb (approved):
diff LGTM

maximiliankolb:
> As I mentioned in private: I think we should treat `product` the same as `content view` and `lifecycle environment`. Perhaps this is my engineer's view, but all of them are models in Katello. Since it was decided that `content view` and `lifecycle environment` should be lower cased, I think `product` should also be lower case.

While reviewing this PR, I had the same feeling (the other way around: capitalize repository, CV, and LE etc). Some sentences that contain "Product" and any other entity such as repository feel weird to read and make you wonder if there's a mistake in capitalization.

However, I am still OK with this change.

asteflova:
Looking into the [PatternFly style guide](https://www.patternfly.org/ux-writing/capitalization#red-hat-product-uis):

> Capitalize product feature names when they’re used as proper nouns or when they refer to a capitalized UI term (like a navigation item). Write them in lowercase when they’re used to describe generic concepts. For example:

> “Add a source by going to Settings > Sources.” "Check the Sources table for status."
> 
> Button text: "Add source"
> 	
> “Sources” is only capitalized when it directly refers to a subsection, feature, or location in the UI. "Source" is lowercase in the button text because button labels should always be in sentence case.

If I'm reading this right, it would mean lowercase for product/LE/CV/repository when used in regular sentences in docs.

apinnick:
I am closing this PR pending discussion about Product vs product.

PR #3145, opened 2024-07-17
Vale

apinnick:

* [ ] Foreman 3.6/Katello 4.8


apinnick:
I have a feeling this will not work because I forked the repo. See the Vale GitHub action [documentation](https://redhat-documentation.github.io/vale-at-red-hat/docs/main/user-guide/using-vale-github-action/):

> WARNING: Because of GitHub token permissions, the Vale GitHub Action cannot post annotations to pull requests from a forked repository.

apinnick:
Looks like we might have to use a script instead. See [documentation](https://redhat-documentation.github.io/vale-at-red-hat/docs/main/user-guide/using-vale-with-a-continuous-integration-service/).

Lennonka:
> Looks like we might have to use a script instead. See [documentation](https://redhat-documentation.github.io/vale-at-red-hat/docs/main/user-guide/using-vale-with-a-continuous-integration-service/).

I don't know about that. Possibly. Perhaps we should also turn it into a build target in the [Makefile](https://github.com/theforeman/foreman-documentation/blob/master/guides/Makefile)?

Otherwise, I hope someone more knowledgeable would help. @ekohl @evgeni ?

evgeni:
I think I need more context to understand what's going on here.

Lennonka:
@evgeni We want to add Vale, which is a linter for the RH style, to our GitHub PR checks.
Not sure how. Found [Vale as GitHub action](https://redhat-documentation.github.io/vale-at-red-hat/docs/main/user-guide/using-vale-github-action/) in the Vale docs.

.github/workflows/lint-with-vale.yml
```diff
@@ -0,0 +1,21 @@
...
+#         with:
+#           filter_mode: diff_context
+#           vale_flags: "--no-exit --minAlertLevel=error"
+#           reporter: github-pr-review
+#           fail_on_error: true
+#         env:
+#           # Required, set by GitHub actions automatically:
+#           # https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret
+#           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+#           REVIEWDOG_GITHUB_API_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
ekohl, line 21: Please ensure line endings in files

ekohl (commented):
GitHub only picks up new actions from inside the repository. One workaround I've found is to temporarily push your branch to upstream. Once the action is known, you can delete your branch and continue working from your fork.

.vale.ini
```diff
@@ -0,0 +1,38 @@
...
+# Use local Vocab terms
+Vale.Terms = YES
+Vale.Avoid = YES
+
+# Ignore files in dirs starting with `.` to avoid raising errors for `.vale/fixtures/*/testinvalid.adoc` files
+[[!.]*.adoc]
+
+BasedOnStyles = RedHat, AsciiDoc
+
+# Match INI files. See: https://docs.errata.ai/vale/scoping
```
evgeni, line 22: This URL goes to a domain grabber. Did you mean https://vale.sh/docs/topics/scoping/?

evgeni:
> GitHub only picks up new actions from inside the repository. One workaround I've found is to temporarily push your branch to upstream. Once the action is known, you can delete your branch and continue working from your fork.

That's not true for new workflows. Only when the repo didn't have actions at all before.

See https://github.com/theforeman/foreman-documentation/pull/3147/

.github/workflows/lint-with-vale.yml
```diff
@@ -0,0 +1,26 @@
...
+          filter_mode: diff_context
+          vale_flags: "--no-exit --minAlertLevel=error"
+          reporter: github-pr-review
+          fail_on_error: true
+          files: guides/common/
+        env:
+          # Required, set by GitHub actions automatically:
+          # https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+          REVIEWDOG_GITHUB_API_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
evgeni, line 26: you can skip the env, the action fetches `secrets.GITHUB_TOKEN` itself, and you only need to override it if you want to give it a different token.

apinnick:
It seems to work now. The check 'failed' after 1m and the annotations show the 2 errors I would expect.

.github/workflows/lint-with-vale.yml
```diff
@@ -0,0 +1,21 @@
+---
+name: Linting with Vale on pull request
```
Lennonka, line 2:
```suggestion
name: linter
```

.github/workflows/lint-with-vale.yml
```diff
@@ -0,0 +1,21 @@
...
+on:
+  pull_request:
+    paths:
+      - 'guides/common/*.adoc'
+      - '.github/workflows/lint-with-vale.yml'
+
+jobs:
+  vale:
+    name: Linting with Vale
```
Lennonka, line 11:
```suggestion
    name: lint-with-vale
```

Lennonka (changes requested):
Can we give it names consistent with other checks, please?

.github/workflows/lint-with-vale.yml
apinnick: I do not recommend this because it is possible to set up multiple linters. Specific job names make debugging easier.

.github/workflows/lint-with-vale.yml
apinnick: Not sure why the name should be changed.

.github/workflows/lint-with-vale.yml
ekohl: Right now the check shows up as `Linting with Vale on pull request / Linting with Vale (pull_request)`. That's composed of:
* `Linting with Vale on pull request` (this line)
* `Linting with Vale` (name on line 11)
* `pull_request` (trigger, line 4)

You can see there's a lot of  duplication.

I'd still consider simplifying to:
```suggestion
name: Vale
```

And then make line 11 simply `Lint`. The result will be `Vale / Lint (pull_request)`. On the checks tab all jobs will be grouped under `Vale` as well (though there is only 1 job now).

You can also consider renaming the file to `vale.yml`. IMHO the `lint-with` is redundant and you can simply make the whole file just for all actions that Vale can do.

apinnick: Done

.github/workflows/vale.yml
```diff
@@ -0,0 +1,21 @@
...
+    name: linter
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: errata-ai/vale-action@reviewdog
+        with:
+          filter_mode: diff_context
+          vale_flags: "--no-exit --minAlertLevel=error"
+          reporter: github-pr-review
+          fail_on_error: false
```
ekohl, line 20: Don't we actually want the check to fail if there's an error? Otherwise they're just warnings, right?

apinnick: "fail_on_error: true" blocks merge. I don't recommend it until this feature has been thoroughly tested. [Documentation](https://redhat-documentation.github.io/vale-at-red-hat/docs/main/user-guide/using-vale-github-action/)

apinnick: Yes, it's true that they are just warnings, but if you make a Vale pass mandatory, a merge might be blocked until someone can add an exception.

apinnick: Or we could make it mandatory and see how it goes....

maximiliankolb:
For the first month, I'd suggest to **not** block PRs if Vale fails. After all maintainers and regular contributors have some experience, I am open/in favor of making the check mandatory.

.github/workflows/vale.yml
apinnick: That's what I thought. :-) So I will resolve this comment.

apinnick: > A failing check shows up, but doesn't block merge unless you make it a required check. Just as the linkchecker today doesn't block merges but does clearly show.

I will add an error just to make sure that the Vale documentation is accurate.

Lennonka (approved):
Looks great!

maximiliankolb (approved):
Thanks Avital!

Lennonka:
Regarding the naming. If you look at the checks, you can see that the "check class" goes first and then goes the specific "check name". For Vale, it's the other way around. But it's not a blocker.
image

.github/workflows/vale.yml
```diff
@@ -0,0 +1,21 @@
+---
+name: Vale
+on:
+  pull_request:
+    paths:
+      - 'guides/common/*.adoc'
```
ekohl, line 6: Is this recursive? I'd be worried this is globbing just on `guides/common/*.adoc` instead of `guides/common/**/*.adoc`.

apinnick: It found errors in a guilds/common/module/<file>.

apinnick:
> Regarding the naming. If you look at the checks, you can see that the "check class" goes first and then goes the specific "check name". For Vale, it's the other way around. But it's not a blocker. 

@Lennonka No, it doesn't work this way. If you look at the other workflow files, "deploy" or "lnkchecker" is the name of the YAML file and "check-html", etc., is the name of the job. The first element of the name does not refer to a 'class' or 'stage'. I could change it to "Vale / Vale" but I'm not sure how that would be any better.

ekohl (approved):
I think the current name is correct. It's indeed `Workflow / Job (Trigger)` and to me this is the best option.

.github/workflows/vale.yml
ekohl: I think this was actually a problem. We can see in https://github.com/theforeman/foreman-documentation/pull/3165 that it's not triggered.

This PR also modified the other trigger (`vale.yml`) so it wasn't a problem (and vale doesn't care about it).

Looking at https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-based-on-files-changed-in-a-pull-request they explicitly use `**.js`.

PR #3152, opened 2024-07-19
Add command for new rake task

bangelic:

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/con_performing-post-upgrade-tasks.adoc
+++ b/guides/common/modules/con_performing-post-upgrade-tasks.adoc
@@ -10,13 +10,20 @@ This information allows Katello to display manifest labels, annotations, and inf
 As a result, migrations must be performed to pull this content from manifests into the database.
 
 ifdef::katello[]
-This migration takes time, so if you depend on container content and need minimal upgrade downtime, run the following command on {ProjectServer} for a pre-migration:
+This migration takes time, so if you depend on container content and need minimal upgrade downtime, run the following commands on {ProjectServer} for a pre-migration:
 
 [options="nowrap" subs="attributes"]
 ----
 # foreman-maintain advanced procedure run pulpcore-container-handle-image-metadata
 ----
 This command migrates data in the background without any need for downtime and reduces future upgrade downtime.
+
+You will also need to run this command:
+[options="nowrap" subs="attributes"]
+----
+# run foreman-rake katello:import_container_manifest_labels
+----
+This command allows the container image API to display manifest labels, annotations, and if the manifest represents bootable or flatpak content.
 endif::[]
 
 ifdef::satellite,orcharhino[]
--- a/guides/common/modules/proc_performing-post-upgrade-tasks.adoc
+++ b/guides/common/modules/proc_performing-post-upgrade-tasks.adoc
@@ -10,16 +10,17 @@ This information allows Katello to display manifest labels, annotations, and inf
 As a result, migrations must be performed to pull this content from manifests into the database.
 
 ifdef::katello[]
+This migration takes time, so if you depend on container content and need minimal upgrade downtime, use this procedure to migrate data.
+
 .Procedure
-. This migration takes time, so if you depend on container content and need minimal upgrade downtime, enter the following commands in a `tmux` window on {ProjectServer} for a pre-migration:
+. Enter the following command in a `tmux` window on {ProjectServer} for a pre-migration.
+This command migrates data while {Project} is running without any need for downtime and reduces future upgrade downtime:
 +
 [options="nowrap" subs="attributes"]
 ----
 # {foreman-maintain} advanced procedure run pulpcore-container-handle-image-metadata
 ----
-This command migrates data while {Project} is running without any need for downtime and reduces future upgrade downtime.
-
-. To allow the container image API to display manifest labels, annotations, and if the manifest represents bootable or flatpak content, enter the following command:
+. If the manifest represents bootable or flatpak content, allow the container image API to display manifest labels, annotations by entering the following command:
 +
 [options="nowrap" subs="attributes"]
 ----
--- a/guides/common/modules/proc_performing-post-upgrade-tasks.adoc
+++ b/guides/common/modules/proc_performing-post-upgrade-tasks.adoc
@@ -15,7 +15,7 @@ ifdef::katello[]
 +
 [options="nowrap" subs="attributes"]
 ----
-# foreman-maintain advanced procedure run pulpcore-container-handle-image-metadata
+# {foreman-maintain} advanced procedure run pulpcore-container-handle-image-metadata
 ----
 This command migrates data while {Project} is running without any need for downtime and reduces future upgrade downtime.
 
--- a/guides/common/modules/proc_performing-post-upgrade-tasks.adoc
+++ b/guides/common/modules/proc_performing-post-upgrade-tasks.adoc
@@ -11,21 +11,20 @@ As a result, migrations must be performed to pull this content from manifests in
 
 ifdef::katello[]
 .Procedure
-. This migration takes time, so if you depend on container content and need minimal upgrade downtime, run the following commands on {ProjectServer} for a pre-migration:
+. This migration takes time, so if you depend on container content and need minimal upgrade downtime, enter the following commands in a `tmux` window on {ProjectServer} for a pre-migration:
 +
 [options="nowrap" subs="attributes"]
 ----
 # foreman-maintain advanced procedure run pulpcore-container-handle-image-metadata
 ----
-This command migrates data in the background without any need for downtime and reduces future upgrade downtime.
+This command migrates data while {Project} is running without any need for downtime and reduces future upgrade downtime.
 
-. You will also need to run this command:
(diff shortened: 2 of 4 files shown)


guides/common/modules/con_performing-post-upgrade-tasks.adoc
```diff
@@ -10,13 +10,20 @@ This information allows Katello to display manifest labels, annotations, and inf
...
 [options="nowrap" subs="attributes"]
 ----
 # foreman-maintain advanced procedure run pulpcore-container-handle-image-metadata
 ----
 This command migrates data in the background without any need for downtime and reduces future upgrade downtime.
+
+You will also need to run this command:
+[options="nowrap" subs="attributes"]
+----
+# run foreman-rake katello:import_container_manifest_labels
```
ekohl, line 24: Why `run`?
```suggestion
# foreman-rake katello:import_container_manifest_labels
```

asteflova (commented):
Now that the module includes two two commands, how about turning it into a proper procedure? That would include using a numbered list for the two commands and perhaps also renaming the file to "proc_*.adoc".

guides/common/modules/proc_performing-post-upgrade-tasks.adoc
```diff
@@ -10,13 +10,22 @@ This information allows Katello to display manifest labels, annotations, and inf
...
-This migration takes time, so if you depend on container content and need minimal upgrade downtime, run the following command on {ProjectServer} for a pre-migration:
-
+.Procedure
+. This migration takes time, so if you depend on container content and need minimal upgrade downtime, run the following commands on {ProjectServer} for a pre-migration:
++
 [options="nowrap" subs="attributes"]
 ----
 # foreman-maintain advanced procedure run pulpcore-container-handle-image-metadata
 ----
 This command migrates data in the background without any need for downtime and reduces future upgrade downtime.
```
ekohl, line 20: General question: the command runs in the foreground? If it can take a long time, should we recommend running it in `tmux` like we do upgrades? That way you can detach the process. Or can you rerun it multiple times if it stopped?

guides/common/modules/proc_performing-post-upgrade-tasks.adoc
```diff
@@ -10,13 +10,22 @@ This information allows Katello to display manifest labels, annotations, and inf
...
+
+. You will also need to run this command:
```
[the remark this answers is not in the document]
bangelic: @ianballou Can you elaborate on this?

guides/common/modules/proc_performing-post-upgrade-tasks.adoc
bangelic: @ianballou Can you help with this as well?

guides/common/modules/proc_performing-post-upgrade-tasks.adoc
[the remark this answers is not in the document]
ianballou: The commands must run in the order:
1) `pulpcore-container-handle-image-metadata`
2) `import_container_manifest_labels`

They have to be run sequentially. The commands only need to be run one time together.

guides/common/modules/proc_performing-post-upgrade-tasks.adoc
ianballou: The "in the background" part is a bit misleading I realize -- the original intent was to communicate that it can run while Foreman is still operating normally. 
If the process stops, you should be able to run the command over and over again, but I think it's worth saying to run it in a tmux window.

guides/common/modules/proc_performing-post-upgrade-tasks.adoc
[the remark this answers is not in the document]
ekohl: Then why isn't the rake task part of the maintain procedure?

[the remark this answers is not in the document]
ianballou: It was originally, but it was taken out because it should only run on smart proxies, not Foreman itself. Perhaps it could've still lived in there but only ran if the system was a smart proxy.  https://github.com/theforeman/foreman_maintain/pull/877/files

guides/common/modules/proc_performing-post-upgrade-tasks.adoc
```diff
@@ -10,13 +10,21 @@ This information allows Katello to display manifest labels, annotations, and inf
 As a result, migrations must be performed to pull this content from manifests into the database.
 
 ifdef::katello[]
-This migration takes time, so if you depend on container content and need minimal upgrade downtime, run the following command on {ProjectServer} for a pre-migration:
+.Procedure
+. This migration takes time, so if you depend on container content and need minimal upgrade downtime, enter the following commands in a `tmux` window on {ProjectServer} for a pre-migration:
```
asteflova, line 14: This line actually sort of introduces both steps, right? In that case, it should not be part of either step. It would be a good fit for an introductory paragraph -- either before or `.Procedure`. What do you think?

guides/common/modules/proc_performing-post-upgrade-tasks.adoc
```diff
@@ -10,13 +10,21 @@ This information allows Katello to display manifest labels, annotations, and inf
...
+# {foreman-maintain} advanced procedure run pulpcore-container-handle-image-metadata
+----
+This command migrates data while {Project} is running without any need for downtime and reduces future upgrade downtime.
 
+. To allow the container image API to display manifest labels, annotations, and if the manifest represents bootable or flatpak content, enter the following command:
++
 [options="nowrap" subs="attributes"]
 ----
-# foreman-maintain advanced procedure run pulpcore-container-handle-image-metadata
+# foreman-rake katello:import_container_manifest_labels
```
asteflova, line 26: It would be good to describe the commands in a consistent way. Right now, in the first step, you start with the command and follow with a description of what the command does; in the second step, you do things the other way. (Looking at the surrounding procedures, they start with a description.)

guides/common/modules/proc_performing-post-upgrade-tasks.adoc
```diff
@@ -10,13 +10,21 @@ This information allows Katello to display manifest labels, annotations, and inf
...
+.Procedure
+. This migration takes time, so if you depend on container content and need minimal upgrade downtime, enter the following commands in a `tmux` window on {ProjectServer} for a pre-migration:
++
+[options="nowrap" subs="attributes"]
+----
+# {foreman-maintain} advanced procedure run pulpcore-container-handle-image-metadata
+----
+This command migrates data while {Project} is running without any need for downtime and reduces future upgrade downtime.
 
+. To allow the container image API to display manifest labels, annotations, and if the manifest represents bootable or flatpak content, enter the following command:
```
asteflova, line 22: You can drop "enter the command".

But more importantly, is "if the manifest represents [...]" a part of a decision the user must make? Is this a command that users only run when this condition is true?

If that's the case, you should probably lead with that: "If the manifest represents [...], allow the container image [...]".

guides/common/modules/proc_performing-post-upgrade-tasks.adoc
bangelic: I agree. I am going to move it before the procedure steps.

guides/common/modules/proc_performing-post-upgrade-tasks.adoc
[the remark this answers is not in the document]
ekohl: > It was originally, but it was taken out because it should only run on Katello itself, not smart proxies. Perhaps it could've still lived in there but only ran if the system was a Katello server. https://github.com/theforeman/foreman_maintain/pull/877/files

foreman-maintain can constrain it only to be run if katello is present.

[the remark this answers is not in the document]
ekohl: Sorry for the partial reply, but I now see it was introduced as a separate procedure. So if anything this PR should call `foreman-maintain` to do it and let that run `foreman-rake`. We should treat `foreman-rake` as an advanced feature that we don't steer users to.

That also makes me think more about this: do we need to run the `ContainerHandleImageMetadata` procedure on all machines with Pulp? If so, does it mean you also need to run `IndexKatelloRepositoriesContainerMetatdata` after you run `ContainerHandleImageMetadata` on a content proxy?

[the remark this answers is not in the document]
ianballou: All machines indeed need both the rake script and the pulp-container migration run.

> So if anything this PR should call `foreman-maintain` to do it and let that run `foreman-rake`.

I looked into this, and for some reason the `IndexKatelloRepositoriesContainerMetatdata` procedure was not available in the list of advanced procedures to run, unlike the Pulp container one.  Since the release is already out in the wild, I figured we should document what's actually possible.

guides/common/modules/proc_performing-post-upgrade-tasks.adoc
bangelic: Moved it before the procedure and reworded.

bangelic:
Can this one be merged or does need another review? @asteflova

maximiliankolb (approved):
LGTM.

PR #3149, opened 2024-07-18
Review adding AD as an external authentication source

asteflova:
### What changes are you introducing?

After reviewing the existing procedure for adding Active Directory as an external authentication source when the base system of the Foreman server is connected directly to AD, I propose these changes:

* Drop steps related to privilege separation for Apache
* Drop a couple of other instructions that are now obsolete
* Drop adding more detailed explanations of the steps the procedure recommends
* Add a list of authentication features available to AD users
* Other edits for clarity and to set the context of the whole scenario

~~The current procedure for direct AD integration + configuring AD as an external authentication source + using GSS Proxy for privilege separation is faulty. For example, the steps don't really achieve Apache keytab separation as promised, they combine Samba-based and SSSD-based AD join, and some of them could be replaced with a reference to FreeIPA/IdM docs to make sure we don't need to fret about keeping in sync with SSSD development.~~

~~In its latest state (after quite a few changes in course), this PR proposes a complete restart to how we document the use case of direct AD integration -- and to document a minimum viable workflow to achieve direct AD integration and adding AD as an external authentication source:~~

~~* First, refer users to FreeIPA/IdM docs for the integration to AD (referencing SSSD as [the recommended way to join an AD domain](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html-single/integrating_rhel_systems_directly_with_windows_active_directory/index#overview-of-direct-integration-using-sssd_connecting-rhel-systems-directly-to-ad-using-sssd) and omitting references to Samba)~~
~~* Then, advise users on how to configure AD as an external authentication source~~

~~This means that the procedures no longer attempt to configure privilege separation with GSS Proxy. If needed, we can add steps to that purpose later after we agree on the necessary minimum proposed in this PR.~~

### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

https://issues.redhat.com/browse/SAT-22855 requests making changes to how the existing procedure on direct AD integration + configuring AD as an external authentication source handles privilege separation for Apache with GSS proxy. While investigating that scenario, it turned out that proper privilege separation may currently not be possible to achieve. Therefore, the agreed-upon approach is to first document the basic steps for AD enrollment + authentication source configuration and then, in another PR, continue the work by adding steps to achieve proper privilege separation.

~~This PR originated with https://issues.redhat.com/browse/SAT-22855 and then... escalated.~~

[...]

PR #3162, opened 2024-07-24
Drop --now after restarting gssproxy

asteflova:
https://issues.redhat.com/browse/SAT-26357 requests clarifying a procedure step to restart and enable gssproxy.service.

Doing a bit of archeology, I found the commit that introduced `--now` to that step: https://github.com/theforeman/foreman-documentation/commit/a4be0ab11f53cc4ff1302b41816c41dd40874daa  It seems that in case of this particular step, asking users to restart && enable (without --now) might help prevent some raised eyebrows.

* [x] Foreman 3.9/Katello 4.11 (Satellite 6.15; orcharhino 6.8/6.9)
* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_configuring-direct-ad-integration-with-gss-proxy.adoc
+++ b/guides/common/modules/proc_configuring-direct-ad-integration-with-gss-proxy.adoc
@@ -70,7 +70,7 @@ euid = __ID_of_Apache_User__
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
 # systemctl restart gssproxy
-# systemctl enable --now gssproxy
+# systemctl enable gssproxy
 ----
 . To configure the Apache server to use the `gssproxy` service, create a `systemd` drop-in file and add the following content to it:
 +


asteflova:
@ekohl Can you please review? You authored https://github.com/theforeman/foreman-documentation/commit/a4be0ab11f53cc4ff1302b41816c41dd40874daa (see the PR description for an explanation).

guides/common/modules/proc_configuring-direct-ad-integration-with-gss-proxy.adoc
```diff
@@ -70,7 +70,7 @@ euid = __ID_of_Apache_User__
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
 # systemctl restart gssproxy
-# systemctl enable --now gssproxy
+# systemctl enable gssproxy
```
Lennonka, line 73: Isn't `# systemctl enable --now gssproxy` equivalent to:
```
# systemctl enable gssproxy
# systemctl start gssproxy
```
Because we haven't started that service after installing it (I don't see that sort of command [in the previous procedure](https://docs.theforeman.org/nightly/Installing_Server/index-satellite.html#Enrolling_Server_with_the_AD_Server_satellite)), there;s no reason to restart it, but rather simply start it.
Therefore, I'd lean towards this edit instead:
```
- # systemctl restart gssproxy # remove
  # systemctl enable --now gssproxy # keep as is
```

Lennonka (commented):
My two cents.

asteflova:
> My two cents. The two commands duplicate an action, that's why they raise eyebrows probably. Suggesting an alternative edit.

Ah, let's wait for https://github.com/theforeman/foreman-documentation/pull/3149 first, then. Because that introduces changing /etc/gssproxy/http.keytab. Restarting the service will be required. But not just yet, it seems.

guides/common/modules/proc_configuring-direct-ad-integration-with-gss-proxy.adoc
ekohl: https://github.com/theforeman/foreman-documentation/pull/1827#discussion_r1025343810 is the last time we discussed this. I can also confirm that I have a host that runs gssproxy even without IPA set up.

I can see this:
```console
$ systemctl show -p Wants auth-rpcgss-module
Wants=rpc-gssd.service gssproxy.service
$ systemctl show -p Wants nfs-client.target
Wants=rpc-statd-notify.service auth-rpcgss-module.service remote-fs-pre.target
```

And somehow on my system it pulls in `nfs-client.target` into the default `multi-user.target`.

So even though I don't have `/etc/krb5.keytab` it still ends up starting `gssproxy.service`.

I'd argue the restart is correct and `--now` is redundant.

asteflova:
> > My two cents. The two commands duplicate an action, that's why they raise eyebrows probably. Suggesting an alternative edit.
> 
> Ah, let's wait for #3149 first, then. Because that introduces changing /etc/gssproxy/http.keytab. Restarting the service will be required. But not just yet, it seems.

#3149 is nowhere near merging but based on the comments above, that doesn't have to be a blocker. Clarifying the starting/restarting/enabling of gssproxy.service was requested by an external user in https://issues.redhat.com/browse/SAT-26357 so it would be nice to be able to get the fix in.

ekohl (approved):

Lennonka (approved):

asteflova:

ded4218073..9d86041ebf  3.11 -> 3.11
85611ea549..3160b68057  3.10 -> 3.10
bbd2967195..76338f55b5  3.9 -> 3.9
fdabd10304..46e9412026  3.8 -> 3.8
d995cc3eea..583a69942d  3.7 -> 3.7
008b6aeee3..8f6139cfc8  3.6 -> 3.6
838bd144fd..227475c4c8  3.5 -> 3.5

PR #3165, opened 2024-07-25
Fix broken link to RHEL 8 doc

apinnick:
Also updated Vale workflow because it was not running.

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/.github/workflows/vale.yml
+++ b/.github/workflows/vale.yml
@@ -3,7 +3,7 @@ name: Vale
 on:
   pull_request:
     paths:
-      - 'guides/common/*.adoc'
+      - 'guides/common/**.adoc'
       - '.github/workflows/vale.yml'
 
 jobs:
--- a/guides/common/modules/ref_system-requirements.adoc
+++ b/guides/common/modules/ref_system-requirements.adoc
@@ -107,7 +107,7 @@ ifndef::satellite[]
 {RHEL} clones are not being actively tested in FIPS mode.
 If you require FIPS, consider using {RHEL}.
 endif::[]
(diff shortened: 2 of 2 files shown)


.github/workflows/vale.yml
```diff
@@ -3,8 +3,7 @@ name: Vale
 on:
   pull_request:
     paths:
-      - 'guides/common/*.adoc'
-      - '.github/workflows/vale.yml'
+      - '**.adoc'
```
ekohl, line 6: Please separate these changes, at least to its own commit.

apinnick: Done.

.github/workflows/vale.yml
```diff
@@ -3,8 +3,7 @@ name: Vale
 on:
   pull_request:
     paths:
-      - 'guides/common/*.adoc'
-      - '.github/workflows/vale.yml'
```
maximiliankolb, line 7: If not absolutely required, I'd like to keep this in to ensure vale runs if we change the vale GHA itself.

maximiliankolb (commented):
one suggestion; URL for RHEL8 works :heavy_check_mark:

.github/workflows/vale.yml
apinnick: Not sure I understand. You want me to revert these lines? I thought that they were causing the check not to run.

ekohl, line 7: Like @maximiliankolb I think it's good to run a workflow if you make changes to it. I think it should be:
```suggestion
      - 'guides/common/**.adoc'
      - '.github/workflows/vale.yml'
```

maximiliankolb: Only line 7; I'd always run an action if the action itself is changed.

apinnick: Will give this a try.

maximiliankolb (approved):
Now sure why the change in `vale.yml` is necessary. I cannot find the result for your PR, only the result for all of foreman-documentation. vale seems to run though :heavy_check_mark:

apinnick:
Merged and cherrypicked:

b358de4480..f707d9cbdd  3.11 -> 3.11
02d23303af..7e66719dc8  3.10 -> 3.10

PR #3174, opened 2024-07-29
Vale "slash" rule reenabled. Config changes trigger workflow

apinnick:
Enabled "RedHat.Slash" rule and updated paths in GitHub workflow vale.yml file.

* [ ] Foreman 3.6/Katello 4.8


.github/workflows/vale.yml
```diff
@@ -3,7 +3,6 @@ name: Vale
 on:
   pull_request:
     paths:
-      - 'guides/common/**.adoc'
```
ekohl, line 6: This means it won't trigger on `adoc` files anymore, only when `valie.yml` is changed. I don't think that's what you intended. Instead I think we should add other vale related files to this list.

ekohl (changes requested):

.github/workflows/vale.yml
apinnick: I added an error .adoc file after reverting this line, but Vale did not check the new file.

apinnick: I see the problem. Vale is only going to report actual errors. I'm going to raise the minimum report level from "error" to "warning" to see how that goes.

apinnick: Works now! "warning" level was not a success, so I reverted to "error" level reporting and made sure I had real errors.

.github/workflows/vale.yml
```diff
@@ -5,6 +5,7 @@ on:
     paths:
       - 'guides/common/**.adoc'
       - '.github/workflows/vale.yml'
+      - '.vale/styles/config/**.txt'
```
ekohl, line 8: How about just triggering on any vale change?
```suggestion
      - '.vale/**'
      - .vale.ini
```

ekohl: But this is only when we trigger the job. IMHO anything in `.vale` affects how Vale behaves so we should trigger CI on it. There's no reason to be more specific. `.vale/**` should still trigger when `.vale/styles/config/vocabularies/Foreman/accept.txt` changes.

apinnick: OK, you might have a point. I have never tested this. I will make the change and see what happens.

ekohl (commented):
The workflow changes look good to me, but could you rebase? It looks like it includes some changes that we merged to master already.

maximiliankolb (approved):
if you squash: LGTM

apinnick:
> if you squash: LGTM

Squashed!

ekohl (commented):
I'd prefer a more descriptive commit message. https://cbea.ms/git-commit/ is IMHO still the best reference on the topic. If I'm pedantic, I'd even suggest two commits because each change is logically speaking separate but I don't think others enforce that.

apinnick:
> I'd prefer a more descriptive commit message. https://cbea.ms/git-commit/ is IMHO still the best reference on the topic. If I'm pedantic, I'd even suggest two commits because each change is logically speaking separate but I don't think others enforce that.

I will update the commit message but I don't think breaking the squashed commit is necessary.

maximiliankolb (approved):

PR #3178, opened 2024-08-01
Mention --full-help in installation section

ehelms:

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_configuring-installation.adoc
+++ b/guides/common/modules/proc_configuring-installation.adoc
@@ -13,7 +13,6 @@ If you lose connection to the shell where the installation command is running, s
 
 * Use the `{installer-scenario} --help` command to display the available options and any default values.
 * Use the `{installer-scenario} --full-help` command to display advanced options.
-If you do not specify any values, the default values are used.
 
 * Specify a meaningful value for the option: `--foreman-initial-organization`.
 This can be your company name.
--- a/guides/common/modules/proc_configuring-installation.adoc
+++ b/guides/common/modules/proc_configuring-installation.adoc
@@ -12,6 +12,7 @@ If you lose connection to the shell where the installation command is running, s
 .Considerations
 
 * Use the `{installer-scenario} --help` command to display the available options and any default values.
+* For advanced installation options, use `{installer-scenario} --full-help`.
 If you do not specify any values, the default values are used.
 
 * Specify a meaningful value for the option: `--foreman-initial-organization`.
--- a/guides/common/modules/proc_configuring-installation.adoc
+++ b/guides/common/modules/proc_configuring-installation.adoc
@@ -12,7 +12,7 @@ If you lose connection to the shell where the installation command is running, s
 .Considerations
 
 * Use the `{installer-scenario} --help` command to display the available options and any default values.
-* For advanced installation options, use `{installer-scenario} --full-help`.
+* Use the `{installer-scenario} --full-help` command to display advanced options.
 If you do not specify any values, the default values are used.
 
 * Specify a meaningful value for the option: `--foreman-initial-organization`.
--- a/guides/common/modules/proc_configuring-installation.adoc
+++ b/guides/common/modules/proc_configuring-installation.adoc
@@ -11,7 +11,7 @@ If you lose connection to the shell where the installation command is running, s
 
 .Considerations
 
-* Use the `{installer-scenario} --help` command to display the available options and any default values.
+* Use the `{installer-scenario} --help` command to display the most commonly used options and any default values.
 * Use the `{installer-scenario} --full-help` command to display advanced options.
 
 * Specify a meaningful value for the option: `--foreman-initial-organization`.


guides/common/modules/proc_configuring-installation.adoc
```diff
@@ -12,6 +12,7 @@ If you lose connection to the shell where the installation command is running, s
 .Considerations
 
 * Use the `{installer-scenario} --help` command to display the available options and any default values.
```
ekohl, line 14: I wonder if this line should mention that it lists only the commonly used options, but I couldn't come up with a decent phrasing.

asteflova:
We've had a few PRs recently that use `.Additional resources` (placed at the very end of the procedure) to point users to --help commands. This approach follows the [modular documentation template for procedures](https://redhat-documentation.github.io/modular-docs/#con-creating-procedure-modules_writing-mod-docs). For example: https://github.com/theforeman/foreman-documentation/blob/6fa7a2d10c1aaff4fcce0253ee5b1a9662422531/guides/common/modules/proc_creating-a-user.adoc?plain=1#L47-L49

If you're willing to give this a try, the wording could go like this:

```
.Additional resources
* For a list of commonly used options and their default values, use the `{installer-scenario} --help` command.
* For a list of advanced installation options, use `{installer-scenario} --full-help`.
```

The caveat is that the rest of the file does _not_ follow the template so including --help commands at the end might make them look out of place. I'd argue that it's still more fitting than keeping them in `.Considerations`.

asteflova:
When creating a doc PR based on a ticket, can you please include a link to the ticket? In this case, it's https://issues.redhat.com/browse/SAT-25655  It's useful for a reviewer to be able to check the context of the bug report or possibly any customer case that's attached.

ehelms:
I am not a fan of shoving the mention of help and full-help to the bottom of the document where they can get lost. For this use case, I personally like them where they are.

guides/common/modules/proc_configuring-installation.adoc
maximiliankolb:
```suggestion
* Use the `{installer-scenario} --help` command to display the most commonly used options and any default values.
```

maximiliankolb (approved):
made one small suggestion. Either way, I like the addition -> LGTM.

guides/common/modules/proc_configuring-installation.adoc
```diff
@@ -12,6 +12,7 @@ If you lose connection to the shell where the installation command is running, s
...
+* For advanced installation options, use `{installer-scenario} --full-help`.
```
asteflova, line 15:
```suggestion
* Use the `{installer-scenario} --full-help` command to display advanced options.
```

This is to keep the list item parallel to the one before, this helps users navigate the text more quickly.

guides/common/modules/proc_configuring-installation.adoc
```diff
@@ -12,6 +12,7 @@ If you lose connection to the shell where the installation command is running, s
...
 If you do not specify any values, the default values are used.
```
asteflova, line 16:
```suggestion
```
This used to be part of the first bullet item. You could move the line back to the first bullet but I recommend to drop it. (Default values are always used by default, no need to explain it.)

asteflova:
> I am not a fan of shoving the mention of help and full-help to the bottom of the document where they can get lost. For this use case, I personally like them where they are.

Fair enough. I still believe it would work quite well but that would require more changes to the structure of the surrounding text... and neither one of us really wants to get into that right now :)

I reviewed again, this time focusing only on your addition and refraining from comments on the other aspects like placement and scannability.

asteflova (approved):
You're not requesting cherry-picks. Is that alright?

maximiliankolb (approved):

asteflova:
Before I merge this, I would like to clarify the cherry-picks. From my point of view (having checked the previous doc versions and calling --full-help on previous versions), it looks like this should be cherry-picked all the way down to 3.5 because we are supposed to maintain older doc versions too for Satellite customers. It looks like your point of view @ehelms is that you don’t want this cherry-picked. Can you please help me understand what the situation is?

ehelms:
> Before I merge this, I would like to clarify the cherry-picks. From my point of view (having checked the previous doc versions and calling --full-help on previous versions), it looks like this should be cherry-picked all the way down to 3.5 because we are supposed to maintain older doc versions too for Satellite customers. It looks like your point of view @ehelms is that you don’t want this cherry-picked. Can you please help me understand what the situation is?

You can cherry pick it if you want. I didn't select cherry picks because I didn't feel it needed to be picked.

asteflova:
> You can cherry pick it if you want. I didn't select cherry picks because I didn't feel it needed to be picked.

It's not really a question of 'wanting to' (cherry-picking so much is obviously a bit of a pain) but rather a question of properly maintaining all supported versions :) Thanks for confirming!

asteflova:

64c5705170..0f68328613  3.11 -> 3.11
81b1f04956..7800f5c976  3.10 -> 3.10
e29f23ef8b..4b62fe4c08  3.9 -> 3.9
01da73c87f..afedcd515f  3.8 -> 3.8
818571283b..385d32af61  3.7 -> 3.7
5f84290fe5..f5dab8ce1f  3.6 -> 3.6
6f3b8aa8a5..b234d4026f  3.5 -> 3.5

PR #3183, opened 2024-08-05
Drop plugin installation instructions on Satellite

ekohl:
Satellite has a set of mandatory plugins and the satellite meta-package ensures these are installed. Previously some were ensured by the installer but starting Satellite 6.16 this list is complete. The next step is to remove them from the installer, but some documentation procedures still told users to use these redundant options.

While this should be safe to cherry pick to Satellite 6.15, I'm a bit hesitant. Technically you'd rely on defaults in the installer to be correct while in 6.16 the packaging guarantees it. On the other hand, for ansible we we only `foreman-proxy-plugin-ansible` being enabled by default.

For now I've only selected the upstream supported versions to reduce potential cherry pick issues. There's also a small change that provides the instructions for azure on foreman-deb, but we probably don't publish that guide anyway.

Link: https://issues.redhat.com/browse/SAT-26276

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/assembly_getting-started-with-ansible.adoc
+++ b/guides/common/assembly_getting-started-with-ansible.adoc
@@ -4,7 +4,9 @@ include::modules/con_supported-ansible-versions.adoc[leveloffset=+1]
 
 include::modules/proc_configuring-your-project-to-run-ansible-roles.adoc[leveloffset=+1]
 
+ifndef::satellite[]
 include::modules/proc_enabling-ansible-integration-with-project.adoc[leveloffset=+1]
+endif::[]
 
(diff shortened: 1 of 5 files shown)


maximiliankolb (approved):
diff LGTM. Did you also check plugins such as OpenSCAP?

ehelms (approved):

ekohl:
> diff LGTM. Did you also check plugins such as OpenSCAP?

Yes, that already has the correct conditionals.

ekohl:
e002af9ca417..e5c30c9ccca1  3.10 -> 3.10
   19904cdb00aa..5499fdfc446f  3.11 -> 3.11

PR #3187, opened 2024-08-05
update the FQDN when doing the doing the monitoring configuration

Imaanpreet:

* [ ] Foreman 3.6/Katello 4.8


Imaanpreet:
Problem Statement

Here is the link

https://docs.redhat.com/en/documentation/red_hat_satellite/6.15/html-single/monitoring_satellite_performance/index#configuring-pcp-data-collection_monitoring

and here is the recommendation

```
cd /var/lib/pcp/pmdas/openmetrics
echo "https://satellite.example.com/metrics" > config.d/foreman.url
./Install 
```
 Instead, it could be something like

```
cd /var/lib/pcp/pmdas/openmetrics
echo "https://$(hostname -f)/metrics" > config.d/foreman.url
./Install 
```
 
The reason is because, in general, the customer will copy/paste the command and will move on, as consequence, the implementation will fail, once the FQDN is different.

Lennonka:
I don't think this is a good idea, especially for API requests, because those can be run from other machines than the Foreman server.

ekohl (changes requested):
Implementation note: you also didn't include `$()` so it's not clear that it is a shell command. But that's a small detail.

This is not something we should do. The `foreman-example-com` attribute is used in a very different way. For example, we say on the Capsule point it to `foreman-example-com`. If you change that to `hostname -f` then it will break.

Our convention is to use underscores to indicate a user should change the actual value.

Further looking into the future: I want to get rid of the hostname equals service name. At some point I think we should support that your server is named `satellite01.example.com` (and if it's load balanced also `satellite02.example.com`) while the service runs on `satellite.example.com`. The `foreman-example-com` attribute is supposed to be the service name, not the hostname.

ekohl:
I'd even suggest that if you really want to do this (and that's a big if) then we should modify the procedure to avoid the attribute.

Imaanpreet:
Hello @ekohl , @Lennonka & @maximiliankolb : thank you for looking in to the issue. 

Sorry for the typo (for not adding $ while creating the PR) @ekohl. 

Before creating the PR , I knew this we need to come to an agreement what to do with this request.  So the goal of this PR is to avoid the failure and we can do this by adding an another line as @Lennonka mentioned. I can totally agree with @ekohl avoid changing the attribute.

maximiliankolb:
docs triage: Closing this PR. We cannot make this change as proposed: Docs and esp. attribute usage are consistent: we always have "foreman.example.com" as single FQDN for your Foreman instance.

Regarding HA/load balancing: please see https://github.com/theforeman/foreman-documentation/pull/3187#pullrequestreview-2220814483

@Imaanpreet If you have further concerns, please reopen the PR or ping us on Matrix.

Imaanpreet:
Hello @maximiliankolb : can we just add a line by replace the satellite/foreman hostname with FQDN for example - 

"Be aware that you will need to replace the value of _{foreman-example-com}_ with the FQDN of your server"

[here](https://github.com/theforeman/foreman-documentation/blob/master/guides/common/modules/proc_configuring-pcp-data-collection.adoc?plain=1#L57)

This way we keep the hardcoded value and we don't need to change it. What are your thoughts about it?

maximiliankolb:
@Imaanpreet Maybe some after line 63 similar to

````
# guides/common/modules/snip_replace-fqdn-projectserver.adoc
1:Replace _{foreman-example-com}_ with the FQDN of your {ProjectServer}.
````

Imaanpreet:
> @Imaanpreet Maybe after line 63 similar to
> 
> ```
> # guides/common/modules/snip_replace-fqdn-projectserver.adoc
> 1:Replace _{foreman-example-com}_ with the FQDN of your {ProjectServer}.
> ```

looks perfect.

Imaanpreet:
@maximiliankolb, Hello !
Have you updated the change? can I update and give you the new PR ?

Issue #3140, opened 2024-07-17
Capitalize "remote execution" or not?


Lennonka:
Should we capitalize "remote execution" or not?

- The term is not yet defined in the [RH SSG](https://redhat-documentation.github.io/supplementary-style-guide/#_r).
- The capitalization of the term is not yet unified across the docs.

I'm leaning towards not capitalized. What do you think, @theforeman/documentation?

maximiliankolb:
I have no strong opinion but I like the approach of discussing, standardizing, and enforcing a common way. I can help best in phase two and three.

Lennonka:
@apinnick Should we decapitalize this term as well? :)

AkshayGadhaveRH:
Since remote execution is referred to as ReX by a lot of users,  decapitalising it makes it look neater IMO.

apinnick:
My preferences are "remote execution" and "REX". 

There is no rule that says we only capitalize the first letters in acronyms. I think "ReX" looks too 'fussy' and would be difficult to enforce. Users already use "REX".

bangelic:
I agree with @apinnick.

Lennonka:
Thanks, everyone. When I'm back from my vacation, I'll fix it.

apinnick:
[RHSSG PR](https://github.com/redhat-documentation/supplementary-style-guide/pull/542) created. Please review. Thanks!

I will request a separate review in the style council Slack chat.

PR #3193, opened 2024-08-06
Refs #37601 - Refresh/deploy CA cert on hosts

Lennonka:
This is for refreshing the self-signed Foreman CA cert on hosts.

Redmine issue: [#37601](https://projects.theforeman.org/issues/37601)

* [x] Foreman 3.12
* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/assembly_refreshing-the-ca-certificate-on-hosts.adoc
+++ b/guides/common/assembly_refreshing-the-ca-certificate-on-hosts.adoc
@@ -1,9 +1,7 @@
 include::modules/con_refreshing-the-ca-certificate-on-hosts.adoc[]
 
-ifdef::katello,orcharhino,satellite[]
 include::modules/proc_deploying-the-ca-certificate-on-a-host-by-using-script-rex.adoc[leveloffset=+1]
 
 include::modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc[leveloffset=+1]
-endif::[]
 
 include::modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc[leveloffset=+1]
--- a/guides/common/assembly_refreshing-the-ca-certificate-on-hosts.adoc
+++ b/guides/common/assembly_refreshing-the-ca-certificate-on-hosts.adoc
@@ -1,0 +1,7 @@
+include::modules/con_refreshing-the-ca-certificate-on-hosts.adoc[]
+
+include::modules/proc_deploying-the-ca-certificate-on-a-host-by-using-script-rex.adoc[leveloffset=+1]
+
+include::modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc[leveloffset=+1]
+
+include::modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc[leveloffset=+1]
--- a/guides/common/assembly_refreshing-the-ca-certificate-on-hosts.adoc
+++ b/guides/common/assembly_refreshing-the-ca-certificate-on-hosts.adoc
@@ -1,7 +1,9 @@
 include::modules/con_refreshing-the-ca-certificate-on-hosts.adoc[]
 
+ifdef::katello,orcharhino,satellite[]
 include::modules/proc_deploying-the-ca-certificate-on-a-host-by-using-script-rex.adoc[leveloffset=+1]
 
 include::modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc[leveloffset=+1]
+endif::[]
 
 include::modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc[leveloffset=+1]
--- a/guides/common/modules/con_refreshing-the-ca-certificate-on-hosts.adoc
+++ b/guides/common/modules/con_refreshing-the-ca-certificate-on-hosts.adoc
@@ -9,3 +9,16 @@ For more information, see {AdministeringDocURL}planning-for-ca-certificate-renew
 ifdef::katello,orcharhino,satellite[]
 If you have already changed the CA certificate on {ProjectServer} without using the temporary dual CA certificate file, you must refresh the certificate on hosts manually because the scripted variant will not recognize {ProjectServer}.
 endif::[]
+
+[IMPORTANT]
+====
+You only must redeploy the CA certificate if you are not using a certification authority that is well known.
+
+You can list well-known certificate authorities as follows:
+
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+awk -v cmd='openssl x509 -noout -subject' '
+    /BEGIN/{close(cmd)};{print | cmd}' < /etc/ssl/certs/ca-certificates.crt
+----
+====
--- a/guides/common/modules/con_refreshing-the-ca-certificate-on-hosts.adoc
+++ b/guides/common/modules/con_refreshing-the-ca-certificate-on-hosts.adoc
@@ -2,3 +2,10 @@
 = Refreshing the CA certificate on hosts
 
 When you change the CA certificate on your {ProjectServer}, you must refresh the CA certificate on your hosts.
+
+Ensure that you use a temporary dual CA certificate file for uninterrupted operation.
+For more information, see {AdministeringDocURL}planning-for-ca-certificate-renewal[Planning for CA certificate renewal] in _{AdministeringDocTitle}_.
+
+ifdef::katello,orcharhino,satellite[]
+If you have already changed the CA certificate on {ProjectServer} without using the temporary dual CA certificate file, you must refresh the certificate on hosts manually because the scripted variant will not recognize {ProjectServer}.
+endif::[]
--- a/guides/common/modules/con_refreshing-the-ca-certificate-on-hosts.adoc
+++ b/guides/common/modules/con_refreshing-the-ca-certificate-on-hosts.adoc
@@ -1,0 +1,4 @@
+[id="refreshing-the-ca-certificate-on-hosts"]
+= Refreshing the CA certificate on hosts
+
+When you change the CA certificate on your {ProjectServer}, you must refresh the public SSL key of the CA certificate on your hosts.
--- a/guides/common/modules/con_refreshing-the-ca-certificate-on-hosts.adoc
+++ b/guides/common/modules/con_refreshing-the-ca-certificate-on-hosts.adoc
@@ -1,4 +1,4 @@
 [id="refreshing-the-ca-certificate-on-hosts"]
 = Refreshing the CA certificate on hosts
 
-When you change the CA certificate on your {ProjectServer}, you must refresh the public SSL key of the CA certificate on your hosts.
+When you change the CA certificate on your {ProjectServer}, you must refresh the CA certificate on your hosts.
--- a/guides/common/modules/con_refreshing-the-ca-certificate-on-hosts.adoc
+++ b/guides/common/modules/con_refreshing-the-ca-certificate-on-hosts.adoc
@@ -6,19 +6,9 @@ When you change the CA certificate on your {ProjectServer}, you must refresh the
 Ensure that you use a temporary dual CA certificate file for uninterrupted operation.
 For more information, see {AdministeringDocURL}planning-for-ca-certificate-renewal[Planning for CA certificate renewal] in _{AdministeringDocTitle}_.
 
-ifdef::katello,orcharhino,satellite[]
 If you have already changed the CA certificate on {ProjectServer} without using the temporary dual CA certificate file, you must refresh the certificate on hosts manually because the scripted variant will not recognize {ProjectServer}.
-endif::[]
 
 [IMPORTANT]
 ====
-You only must redeploy the CA certificate if you are not using a certification authority that is well known.
-
-You can list well-known certificate authorities as follows:
-
-[options="nowrap" subs="+quotes,verbatim,attributes"]
-----
-awk -v cmd='openssl x509 -noout -subject' '
-    /BEGIN/{close(cmd)};{print | cmd}' < /etc/ssl/certs/ca-certificates.crt
-----
+You only must redeploy the CA certificate if you use a self-signed CA certificate.
 ====
--- a/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
+++ b/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
@@ -19,10 +19,14 @@ Redeploy the CA certificate when you change it on {ProjectServer}.
 +
 [options="nowrap" subs="+quotes,verbatim,attributes"]
 ----
-http://_{foreman-example-com}_/unattended/public/foreman_ca_refresh
+https://_{foreman-example-com}_/unattended/public/foreman_ca_refresh
 ----
 +
 Replace _{foreman-example-com}_ with the hostname of your {ProjectServer}.
++
+You can use HTTP when the CA certificate is expired.
+However, {Team} recommends preparing the certificate renewal in advance.
+For more information, see {AdministeringDocURL}renewing-the-ca-certificate[Renewing the CA certificate] in _{AdministeringDocTitle}_.
 . Optional: Click *Next* and configure advanced fields and scheduling as you require.
 . Click *Run on selected hosts*.
 
--- a/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
+++ b/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
@@ -2,11 +2,12 @@
 = Deploying the CA certificate on a host by using Ansible REX
 
 You can use remote execution (REX) with the Ansible provider to deploy the CA certificate.
-Redeploy the CA certificate when you change it on {ProjectServer}.
 
 .Prerequisites
 * The host is registered to {Project}.
 * Remote execution is enabled on the host.
+* The CA certificate has been changed on {ProjectServer}.
+For more information, see {AdministeringDocURL}planning-for-ca-certificate-renewal[Planning for CA certificate renewal] in _{AdministeringDocTitle}_.
 
 .Procedure
 . In the {ProjectWebUI}, navigate to *Monitor* > *Jobs*.
@@ -26,7 +27,6 @@ Replace _{foreman-example-com}_ with the hostname of your {ProjectServer}.
 +
 You can use HTTP when the CA certificate is expired.
 However, {Team} recommends preparing the certificate renewal in advance.
-For more information, see {AdministeringDocURL}planning-for-ca-certificate-renewal[Planning for CA certificate renewal] in _{AdministeringDocTitle}_.
 . Optional: Click *Next* and configure advanced fields and scheduling as you require.
 . Click *Run on selected hosts*.
 
--- a/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
+++ b/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
@@ -26,13 +26,7 @@ Replace _{foreman-example-com}_ with the hostname of your {ProjectServer}.
 . Optional: Click *Next* and configure advanced fields and scheduling as you require.
 . Click *Run on selected hosts*.
 
-.Verification
-* The following command succeeds on your host:
-+
-[options="nowrap" subs="+quotes,verbatim,attributes"]
-----
-curl https://_{foreman-example-com}_
-----
+include::snip_verification-ca-set-correctly.adoc[]
 
 [role="_additional-resources"]
 .Additional resources
--- a/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
+++ b/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
@@ -19,7 +19,7 @@ include::snip_prerequisites-deploying-ca-cert-rex.adoc[]
 https://_{foreman-example-com}_/unattended/public/foreman_ca_refresh
 ----
 +
-Replace _{foreman-example-com}_ with the hostname of your {ProjectServer}.
+include::snip_replace-fqdn-projectserver.adoc[]
 +
 You can use HTTP when the CA certificate is expired.
 However, {Team} recommends preparing the certificate renewal in advance.
--- a/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
+++ b/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
@@ -1,0 +1,39 @@
+[id="deploying-the-ca-certificate-on-a-host-by-using-ansible-rex"]
+= Deploying the CA certificate on a host by using Ansible REX
+
+You can use remote execution (REX) with the Ansible provider to deploy the CA certificate.
+Redeploy the CA certificate when you change it on {ProjectServer}.
+
+.Prerequisites
+* The host is registered to {Project}.
+* Remote execution is enabled on the host.
+
+.Procedure
+. In the {ProjectWebUI}, navigate to *Monitor* > *Jobs*.
+. Click *Run Job*.
+. From the *Job category* list, select `Ansible Commands`.
+. From the *Job template* list, select `Download and execute a script`.
+. Click *Next*.
+. Select hosts on which you want to execute the job.
+. In the *url* field, enter the following URL:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+http://_{foreman-example-com}_/unattended/public/foreman_ca_refresh
+----
++
+Replace _{foreman-example-com}_ with the hostname of your {ProjectServer}.
+. Optional: Click *Next* and configure advanced fields and scheduling as you require.
+. Click *Run on selected hosts*.
+
+.Verification
+* The following command succeeds on your host:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+curl https://_{foreman-example-com}_
+----
+
+[role="_additional-resources"]
+.Additional resources
+* xref:executing-a-remote-job_managing-hosts[]
--- a/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
+++ b/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
@@ -3,11 +3,7 @@
 
 You can use remote execution (REX) with the Ansible provider to deploy the CA certificate.
 
-.Prerequisites
-* The host is registered to {Project}.
-* Remote execution is enabled on the host.
-* The CA certificate has been changed on {ProjectServer}.
-For more information, see {AdministeringDocURL}planning-for-ca-certificate-renewal[Planning for CA certificate renewal] in _{AdministeringDocTitle}_.
+include::snip_prerequisites-deploying-ca-cert-rex.adoc[]
 
 .Procedure
 . In the {ProjectWebUI}, navigate to *Monitor* > *Jobs*.
--- a/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
+++ b/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
@@ -26,7 +26,7 @@ Replace _{foreman-example-com}_ with the hostname of your {ProjectServer}.
 +
 You can use HTTP when the CA certificate is expired.
 However, {Team} recommends preparing the certificate renewal in advance.
-For more information, see {AdministeringDocURL}renewing-the-ca-certificate[Renewing the CA certificate] in _{AdministeringDocTitle}_.
+For more information, see {AdministeringDocURL}planning-for-ca-certificate-renewal[Planning for CA certificate renewal] in _{AdministeringDocTitle}_.
 . Optional: Click *Next* and configure advanced fields and scheduling as you require.
 . Click *Run on selected hosts*.
 
--- a/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
+++ b/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
@@ -22,7 +22,6 @@ https://_{foreman-example-com}_/unattended/public/foreman_ca_refresh
 include::snip_replace-fqdn-projectserver.adoc[]
 +
 You can use HTTP when the CA certificate is expired.
-However, {Team} recommends preparing the certificate renewal in advance.
 . Optional: Click *Next* and configure advanced fields and scheduling as you require.
 . Click *Run on selected hosts*.
 
--- a/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-script-rex.adoc
+++ b/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-script-rex.adoc
@@ -1,0 +1,39 @@
+[id="deploying-the-ca-certificate-on-a-host-by-using-script-rex"]
+= Deploying the CA certificate on a host by using Script REX
+
+You can use remote execution (REX) with the Script provider to deploy the CA certificate.
+Redeploy the CA certificate when you change it on {ProjectServer}.
+
+.Prerequisites
+* The host is registered to {Project}.
+* Remote execution is enabled on the host.
+
+.Procedure
+. In the {ProjectWebUI}, navigate to *Monitor* > *Jobs*.
+. Click *Run Job*.
+. From the *Job category* list, select `Commands`.
+. From the *Job template* list, select `Download and run a script`.
+. Click *Next*.
+. Select hosts on which you want to execute the job.
+. In the *url* field, enter the following URL:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+http://_{foreman-example-com}_/unattended/public/foreman_ca_refresh
+----
++
+Replace _{foreman-example-com}_ with the hostname of your {ProjectServer}.
+. Optional: Click *Next* and configure advanced fields and scheduling as you require.
+. Click *Run on selected hosts*.
+
+.Verification
+* The following command succeeds on your host:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+curl https://_{foreman-example-com}_
+----
+
+[role="_additional-resources"]
+.Additional resources
+* xref:executing-a-remote-job_managing-hosts[]
--- a/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-script-rex.adoc
+++ b/guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-script-rex.adoc
@@ -2,13 +2,8 @@
 = Deploying the CA certificate on a host by using Script REX
 
 You can use remote execution (REX) with the Script provider to deploy the CA certificate.
-Redeploy the CA certificate when you change it on {ProjectServer}.
 
-.Prerequisites
-* The host is registered to {Project}.
-* Remote execution is enabled on the host.
-* The CA certificate has been changed on {ProjectServer}.
-For more information, see {AdministeringDocURL}planning-for-ca-certificate-renewal[Planning for CA certificate renewal] in _{AdministeringDocTitle}_.
+include::snip_prerequisites-deploying-ca-cert-rex.adoc[]
 
 .Procedure
 . In the {ProjectWebUI}, navigate to *Monitor* > *Jobs*.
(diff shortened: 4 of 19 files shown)


guides/common/modules/con_refreshing-ca-certificates-on-hosts.adoc
```diff
@@ -0,0 +1,4 @@
+[id="refreshing-ca-certificates-on-hosts"]
+= Refreshing CA certificates on hosts
+
+When you change the CA certificate on your {SmartProxy}, in cases such as when you rename your {SmartProxy} or when you configure {SmartProxy} with custom SSL certificates, you have to refresh the public SSL keys of the CA certificate on your hosts.
```
Lennonka, line 4: @ShimShtein This is probably incorrect as well.

guides/common/modules/con_refreshing-the-ca-certificate-on-hosts.adoc
Lennonka, line 1: @ShimShtein I've changed the headings to reflect that there's a single CA cert. I hope it's correct.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
```diff
@@ -0,0 +1,57 @@
...
+Replace _{foreman-example-com}_ with the hostname of your {ProjectServer}.
+. Copy the certificate to the truststore:
+ifdef::client-content-dnf[]
+* On {EL}:
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+cp foreman_ca_cert.crt /etc/pki/ca-trust/source/anchors
+----
+endif::[]
+ifdef::client-content-apt[]
```
Lennonka, line 28: @maximiliankolb Are these attributes used correctly?

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
```diff
@@ -0,0 +1,33 @@
...
+. Click *Run Job*.
+. From the *Job category* list, select `Ansible Commands`.
+. From the *Job template* list, select `Download and execute a script`.
+. Click *Next*.
+. Select hosts on which you want to execute the job.
+. In the *url* field, enter the following URL:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+http://_{foreman-example-com}_/unattended/public/foreman_ca_refresh
```
maximiliankolb, line 22:
```suggestion
https://_{foreman-example-com}_/unattended/public/foreman_ca_refresh
```

unless this is intentional because we assume that the cert is expired?

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-script-rex.adoc
```diff
@@ -0,0 +1,33 @@
...
+. Click *Run Job*.
+. From the *Job category* list, select `Commands`.
+. From the *Job template* list, select `Download and run a script`.
+. Click *Next*.
+. Select hosts on which you want to execute the job.
+. In the *url* field, enter the following URL:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+http://_{foreman-example-com}_/unattended/public/foreman_ca_refresh
```
maximiliankolb, line 22:
```suggestion
https://_{foreman-example-com}_/unattended/public/foreman_ca_refresh
```

same as above.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
maximiliankolb, line 1: Please prefix the commands with either `$` or `#`.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
maximiliankolb: Yes, based on the rendered docs: LGTM.

guides/common/modules/snip_verification-ca-set-correctly.adoc
```diff
@@ -0,0 +1,9 @@
+.Verification
+* The following command succeeds on your host:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+curl https://_{foreman-example-com}_
```
maximiliankolb, line 6:
```suggestion
$ curl -I https://_{foreman-example-com}_
```

To reduce the stdout.

guides/common/modules/snip_verification-ca-set-correctly.adoc
```diff
@@ -0,0 +1,9 @@
...
+----
++
+Replace _{foreman-example-com}_ with the hostname of your {ProjectServer}.
```
maximiliankolb, line 9:
```suggestion
```

Well yes (unless you meant "{SmartProxy}"? I'd drop this. It might even be wrong if hosts are registered to Smart Proxy Servers. I am unsure about this.

I have not found a Smart Proxy API endpoint that does not require authentication. The only alternative I found was `curl -I http://smart-proxy.example.com/pub/` which returns HTTP code 200.

What do you think @ShimShtein ?

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
ShimShtein: We have two scenarios when we will need to run the script:
1. When the user prepares for the change in advance.
2. When the CA already expired/changed on the Foreman server.

I think the full procedure to replace the CA if the user prepares in advance (1) would be:
1. Add a new cert to Foreman's CA
2. Run this REX job to have the dual CA published to the host
3. Remove the old CA from Foreman
4. Run this REX job to replace the dual CA with only the new one.

In this case the `https` endpoint will still be valid.

In case the CA has already been replaced with a new cert (2), the host has no way to verify Foreman's identity, so the `https` endpoint will not work, and the only way to contact the server is with `http`.

guides/common/modules/snip_verification-ca-set-correctly.adoc
ShimShtein: I will need to defer to @ekohl. 
I **think** once Katello is installed, it creates a reverse-proxy to Foreman server, and maybe we can use it to forward requests from the client through proxy.

Any thoughts on this one?

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
Lennonka: Perhaps I should add the general steps to docs as well...

Lennonka:
Rebased.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
Lennonka: I've added a new procedure for CA cert renewal and used it as a recommendation.

guides/common/modules/snip_verification-ca-set-correctly.adoc
Lennonka: Isn't the point here that `{ProjectServer}` is the CA regardless to what the host is registered?

@maximiliankolb I'm adding this line because I have a bug report from our support that users tend to copy-past without editing the command accordingly as discussed in this PR #3187. I can hide it for orcharhino or include it only for Satellite if you wish.

Lennonka:
@ShimShtein What about this? https://docs.theforeman.org/nightly/Installing_Server/index-katello.html#deploying-a-custom-ssl-certificate-to-hosts_foreman

guides/common/modules/proc_planning-for-ca-certificate-renewal.adoc
Lennonka, line 1: I'm not sure I'm using correct terminology. Please check.

guides/common/modules/snip_verification-ca-set-correctly.adoc
maximiliankolb: Maybe you're right about the CA. I am OK keeping it for all build targets.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-script-rex.adoc
```diff
@@ -0,0 +1,37 @@
...
+. Click *Run Job*.
+. From the *Job category* list, select `Commands`.
+. From the *Job template* list, select `Download and run a script`.
+. Click *Next*.
+. Select hosts on which you want to execute the job.
+. In the *url* field, enter the following URL:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+https://_{foreman-example-com}_/unattended/public/foreman_ca_refresh
```
Lennonka, line 22: @ShimShtein This template will refresh the CA cert if and only if Katello is present, right?

guides/common/modules/proc_planning-for-ca-certificate-renewal.adoc
```diff
@@ -0,0 +1,16 @@
+[id="planning-for-ca-certificate-renewal"]
+= Planning for CA certificate renewal
+
+If you need to update SSL certificates in the Certification Authority (CA) certificate on your {ProjectServer}, use a temporary dual CA certificate to retain the HTTPS connections to your {ProjectServer} during the renewal.
+
+.Procedure
+. Add the new SSL certificate to the CA certificate of {ProjectServer} and keep the old SSL certificate in the CA.
+. Renew the certificates on {ProjectServer} and any {SmartProxyServer}s.
```
maximiliankolb, line 8:
```suggestion
. Renew the certificates on {ProjectServer} and any {SmartProxyServers}.
```

see https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-base.adoc#L145

guides/common/modules/proc_planning-for-ca-certificate-renewal.adoc
```diff
@@ -0,0 +1,16 @@
...
+. Deploy the dual CA certificate on hosts.
+. Replace the dual CA certificate on {ProjectServer} with a new CA certificate that contains only the new SSL certificate.
+. Renew the certificates on {ProjectServer} and any {SmartProxyServer}s.
```
maximiliankolb, line 11:
```suggestion
. Renew the certificates on {ProjectServer} and any {SmartProxyServers}.
```

as above.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
```diff
@@ -0,0 +1,37 @@
+[id="deploying-the-ca-certificate-on-a-host-by-using-ansible-rex"]
+= Deploying the CA certificate on a host by using Ansible REX
+
+You can use remote execution (REX) with the Ansible provider to deploy the CA certificate.
+Redeploy the CA certificate when you change it on {ProjectServer}.
```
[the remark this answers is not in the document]
Lennonka: Perhaps this should be a prerequisite together with the link to the Planning section from below.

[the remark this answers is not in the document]
maximiliankolb: Yes, makes sense to me. Also, maybe as ".Next steps" in Changing the CA on Foreman Server".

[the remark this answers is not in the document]
Lennonka: There isn't a separate procedure for that, but the link is already included in the sections for renewing SSL certs.

guides/common/modules/proc_planning-for-ca-certificate-renewal.adoc
```diff
@@ -0,0 +1,16 @@
+[id="planning-for-ca-certificate-renewal"]
+= Planning for CA certificate renewal
+
+If you need to update SSL certificates in the Certification Authority (CA) certificate on your {ProjectServer}, use a temporary dual CA certificate to retain the HTTPS connections to your {ProjectServer} during the renewal.
```
ShimShtein, line 4:
```suggestion
If you need to update Certification Authority (CA) certificate on your {ProjectServer}, add the new CA certificate and use a temporary dual CA certificate file to retain the HTTPS connections to your {ProjectServer} during the renewal.
```

guides/common/modules/proc_planning-for-ca-certificate-renewal.adoc
```diff
@@ -0,0 +1,16 @@
...
+
+.Procedure
+. Add the new SSL certificate to the CA certificate of {ProjectServer} and keep the old SSL certificate in the CA.
```
ShimShtein, line 7:
```suggestion
. Add the new SSL certificate to the CA certificate file on {ProjectServer} and keep the old SSL certificate.
```

guides/common/modules/proc_planning-for-ca-certificate-renewal.adoc
```diff
@@ -0,0 +1,16 @@
...
+. Renew the certificates on {ProjectServer} and any {SmartProxyServers}.
+. Deploy the dual CA certificate on hosts.
+. Replace the dual CA certificate on {ProjectServer} with a new CA certificate that contains only the new SSL certificate.
```
ShimShtein, line 10:
```suggestion
. Remove the old certificate from the CA certificates file on {ProjectServer}, so the CA certificate file contains only the new SSL certificate.
```

guides/common/modules/proc_renewing-a-custom-ssl-certificate-on-server.adoc
```diff
@@ -39,12 +39,11 @@ Do not delete the certificate files after you deploy the certificate.
...
-====
-
 .Verification
 . Access the {ProjectWebUI} from your local machine.
-For example, https://{foreman-example-com}.
+For example, `https://{foreman-example-com}`.
 . In your browser, view the certificate details to verify the deployed certificate.
+
+.Next steps
+* If you have changed the CA certificate on {ProjectServer}, refresh the CA certificate on your hosts.
```
ShimShtein, line 48: If the user already changed the CA cert, then he needs to refresh the certs manually, and the automatic procedure will probably fail (the hosts won't recognize the Foreman server).

guides/common/modules/snip_verification-ca-set-correctly.adoc
ShimShtein: Actually, if the Smart Proxy uses certificates that were issued by the Foreman server (and are validated by the same CA), the user can also access `https://smart.proxy.example.com:9090/features` (a proxy endpoint). If the CA is not set on the host, the communication will fail, since the host won't be able to recognize the certificate presented by the Smart Proxy.

ShimShtein (commented):
@ekohl Can you please take a look that I didn't mislead in the planning section?

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-script-rex.adoc
Lennonka: In that case I exclude the REX variants for non-Katello builds.

guides/common/modules/proc_renewing-a-custom-ssl-certificate-on-server.adoc
Lennonka: But then it doesn't make sense to tell the users that they can access the template over HTTP if they can't use the procedure at all?

guides/common/modules/snip_verification-ca-set-correctly.adoc
Lennonka: @ShimShtein Does it matter whether the user uses the hostname of ProjectServer or SmartProxyServer for the verification? Or does he have to check both?

ShimShtein:
> @ShimShtein What about this? https://docs.theforeman.org/nightly/Installing_Server/index-katello.html#deploying-a-custom-ssl-certificate-to-hosts_foreman Should that also use the templates?

@Lennonka , I didn't change the `katello-rhsm-consumer` script yet. I suppose once we change it (it's already tracked in our Jira), we will need to change this part too.
We need to understand what would be the alternative to using this script.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-script-rex.adoc
```diff
@@ -0,0 +1,38 @@
+[id="deploying-the-ca-certificate-on-a-host-by-using-script-rex"]
+= Deploying the CA certificate on a host by using Script REX
+
+You can use remote execution (REX) with the Script provider to deploy the CA certificate.
+Redeploy the CA certificate when you change it on {ProjectServer}.
+
+.Prerequisites
```
maximiliankolb, line 7: this block looks identical to guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc -> I suggest to use a snippet for that.

If the job template is the only diff, then maybe you can use an attribute for the anchor, title, and job template and even reuse the whole file. This could also be a future improvement.

guides/common/modules/snip_verification-ca-set-correctly.adoc
ShimShtein: It depends on user's network configuration. In some use cases, hosts have only access to the proxy, in others - the hosts have direct connection to the server.

guides/common/modules/proc_renewing-a-custom-ssl-certificate-on-server.adoc
ShimShtein: If the CA is already changed, then most chances are that the server won't be able to execute Remote Execution jobs on the host. This is why only the manual workflow with HTTP endpoint would be possible. As in ssh to the host, download the script from HTTP endpoint and run it.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-script-rex.adoc
Lennonka: I've snippetized the Prerequisites. The rest seems to have too many variables.

ShimShtein (approved):
LGTM

Lennonka:
@ehelms Hi Eric, can you please take a look? Especially at the _Planning_ section.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
```diff
@@ -0,0 +1,63 @@
...
+
+You can deploy the CA certificate on the host manually by rendering a public provisioning template, which will provide the certificate to your host.
+
+.Prerequisites
+* You have root access on your host.
+* The CA certificate has been changed on {ProjectServer}.
+For more information, see {AdministeringDocURL}planning-for-ca-certificate-renewal[Planning for CA certificate renewal] in _{AdministeringDocTitle}_.
+
+.Procedure
+. Log in to you host by using SSH.
```
ehelms, line 12: Do we care how they access their host? Maybe this should drop the SSH or use `such as SSH`?

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
```diff
@@ -0,0 +1,63 @@
...
+However, {Team} recommends preparing the certificate renewal in advance.
+. Copy the certificate to the truststore:
+ifdef::client-content-dnf[]
+ifndef::satellite[]
+* On {EL}:
+endif::[]
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+# cp {project-context}_ca_cert.crt /etc/pki/ca-trust/source/anchors
```
ehelms, line 32: This is not necessary.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
```diff
@@ -0,0 +1,63 @@
...
+You can deploy the CA certificate on the host manually by rendering a public provisioning template, which will provide the certificate to your host.
+
+.Prerequisites
+* You have root access on your host.
+* The CA certificate has been changed on {ProjectServer}.
+For more information, see {AdministeringDocURL}planning-for-ca-certificate-renewal[Planning for CA certificate renewal] in _{AdministeringDocTitle}_.
+
+.Procedure
+. Log in to you host by using SSH.
+. Download the certificate on your host:
```
ehelms, line 13: Similar as above, perhaps we should say get the certificate on your host using insecure HTTP, SCP or some other secure method?

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
ehelms: And is only for some uses cases, like container content access.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
```diff
@@ -0,0 +1,63 @@
+[id="deploying-the-ca-certificate-on-a-host-manually"]
+= Deploying the CA certificate on a host manually
```
ehelms, line 2: I do not see us updating the CA certificate stored under `/etc/rhsm/` for use by subscription-manager.

guides/common/modules/proc_planning-for-ca-certificate-renewal.adoc
```diff
@@ -0,0 +1,16 @@
+[id="planning-for-ca-certificate-renewal"]
+= Planning for CA certificate renewal
+
+If you need to update Certification Authority (CA) certificate on your {ProjectServer}, add the new CA certificate and use a temporary dual CA certificate file to retain the HTTPS connections to your {ProjectServer} during the renewal.
```
ehelms, line 4:
```suggestion
If you need to update the Certification Authority (CA) certificate on your {ProjectServer}, add the new CA certificate and use a temporary dual CA certificate file to retain the HTTPS connections to your {ProjectServer} during the renewal.
```

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
```diff
@@ -0,0 +1,33 @@
...
+. Click *Next*.
+. Select hosts on which you want to execute the job.
+. In the *url* field, enter the following URL:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+https://_{foreman-example-com}_/unattended/public/foreman_ca_refresh
+----
++
+Replace _{foreman-example-com}_ with the hostname of your {ProjectServer}.
```
Lennonka, line 22:
```suggestion
Replace _{foreman-example-com}_ with the FQDN of your {ProjectServer}.
```

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
ShimShtein: @ehelms  Are you sure? If we update the certs manually we need to copy the cert to the truststore, how it will get to the truststore otherwise?

ShimShtein: And if you are talking generally, I suppose having the manual procedure outlined is a good thing.

ehelms: My point is, the cert only needs to go in the truststore if it needs to go in there. We shouldn't be blindly putting it in the truststore in *all* cases, just the ones where it's required.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
Lennonka: This is the standard way we provide guidelines for manual config on hosts. I'm sure that advanced users would be able to adapt it as they need.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
Lennonka: @ehelms When does the CA cert need to go to the truststore?

Lennonka:
@ehelms Since you haven't commented on the _Planning_ section, I'm assuming that it's okay.

guides/common/assembly_refreshing-the-ca-certificate-on-hosts.adoc
ekohl, line 1: Vanilla Foreman doesn't care about the CA certificate (AFAIK we don't support REX pull mode on it) so you can hide everything in this assembly.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
```diff
@@ -0,0 +1,33 @@
+[id="deploying-the-ca-certificate-on-a-host-by-using-ansible-rex"]
+= Deploying the CA certificate on a host by using Ansible REX
```
ekohl, line 2: Do we use the term "Ansible REX" elsewhere? I'd lean to keeping it simple use `using Remote Exection` instead (as we do elsewhere).

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+You can use remote execution (REX) with the Ansible provider to deploy the CA certificate.
```
ekohl, line 4: We're rather inconsistent about capitalizing remote execution. I see it's not listed under Capitalization in `CONTRIBUTING.md`. What's the guidance on this?

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
```diff
@@ -0,0 +1,33 @@
...
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+https://_{foreman-example-com}_/unattended/public/foreman_ca_refresh
+----
++
+include::snip_replace-fqdn-projectserver.adoc[]
++
+You can use HTTP when the CA certificate is expired.
+However, {Team} recommends preparing the certificate renewal in advance.
```
ekohl, line 25: We should avoid the `recommends` phrasing.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
```diff
@@ -0,0 +1,63 @@
...
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+# cp {project-context}_ca_cert.crt /usr/local/share/ca-certificates/
+----
+endif::[]
+. Update the truststore:
+ifdef::client-content-dnf[]
+ifndef::satellite[]
+* On {EL}:
+endif::[]
```
ekohl, line 47: I'd combine those steps into 2 larger `ifdef` blocks. So rather than
```asciidoc
ifdef::client-content-dnf[]
. copy step
endif::[]
ifdef::client-content-apt[]
. copy step
endif::[]
ifdef::client-content-dnf[]
. update step
endif::[]
ifdef::client-content-apt[]
. update step
endif::[]
```

I'd prefer

```asciidoc
ifdef::client-content-dnf[]
. copy step
. update step
endif::[]
ifdef::client-content-apt[]
. copy step
. update step
endif::[]
```

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
ehelms: In the next line, we are showing them how to download the CA certificate. Which is going to be inherently insecure, and users may not realize that because they trust our documentation. I think this is an area where we should be more clear, and use the right language in order to express that **care** needs to be taken.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
ekohl: When the server CA is not already in the trust store. Let's say your server uses Let's Encrypt (or any other well known CA) for certificates, then you can expect it to already be in the trust store and adding it really is redundant.

ehelms: @ianballou Is this still true for docker? for podman? Do we have to add the CA to the system wide trust store? Or can those services be configured individually to use the CA certificate?

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
```diff
@@ -0,0 +1,33 @@
...
+. Click *Run Job*.
+. From the *Job category* list, select `Ansible Commands`.
+. From the *Job template* list, select `Download and execute a script`.
+. Click *Next*.
+. Select hosts on which you want to execute the job.
+. In the *url* field, enter the following URL:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+https://_{foreman-example-com}_/unattended/public/foreman_ca_refresh
```
ekohl, line 19: If you go this route, you must mention that for isolated hosts this must be retrieved via a Smart Proxy (AKA Capsule) because they may not be able to access Foreman.

It would be better to implement a dedicated REX job template and that the user selects. Then the user doesn't need to input any fields and we can rely on REX to properly transfer the script. Even if the certificate expired (because SSH push mode doesn't need them) and when the host is isolated.

The same goes for the Script REX procedure.

ekohl (changes requested):

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
ShimShtein: @ekohl if Foreman uses a well-known CA, then the whole process is redundant. 

@Lennonka What do you think, should we add a "precodition" that this whole process (the process of updating hosts with the new CA) is relevant if the certificate that is used by Foreman is not trusted by default?

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
ShimShtein: > It would be better to implement a dedicated REX job template and that the user selects. Then the user doesn't need to input any fields and we can rely on REX to properly transfer the script. Even if the certificate expired (because SSH push mode doesn't need them) and when the host is isolated.
The same goes for the Script REX procedure.

Created a tracker to properly track it: https://projects.theforeman.org/issues/37773

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
Lennonka: If they plan the replacement of their custom CA cert properly as described in the _Planning_ section and mentioned in Prerequisites, what reasons are there left for the download to be insecure?

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
ehelms: @ekohl with our dropping of API access in the reverse proxy, will isolated hosts have access to this endpoint? do we need to add this endpoint to our list of allowed endpoints?

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
ehelms: Under those conditions yes - but this snippet is not specific to a user having done that. This is a generic "how to deploy a CA on a host manually" section based on how it's titled. Thus, I feel we shouldn't assume anything and give users all the info and tools to handle all of the possibilities.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
ekohl: > @ekohl with our dropping of API access in the reverse proxy, will isolated hosts have access to this endpoint? do we need to add this endpoint to our list of allowed endpoints?

We already have the [templates module](https://github.com/theforeman/smart-proxy/tree/develop/modules/templates) which proxies the unattended templates (also for kickstarts etc). I think the [current API](https://github.com/theforeman/smart-proxy/blob/develop/modules/templates/templates_unattended_api.rb) should cover it, but not 100% sure. This needs to be verified.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
Lennonka: I think we [agreed to decapitalize it](https://github.com/theforeman/foreman-documentation/issues/3140).

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
Lennonka: Actually, yes, this one doesn't assume the Planning. I mixed it up. On it!

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
Lennonka: I needed to distinguish the Ansible/Script provider and the title becomes too long if I spell it out.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
Lennonka: I'll just remove the line. We have it in Prereqs.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
Lennonka: I'd prefer to keep the order of the steps.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
Lennonka: @ehelms So how exactly should the user get the CA certificate to the host securely?

Lennonka: Assuming the endpoint isn't secure, then the user should just copy (scp) the cert from ProjectServer's `/var/html/www/pub` to the host instead of using the endpoint at all?
cc @ShimShtein

guides/common/modules/con_refreshing-the-ca-certificate-on-hosts.adoc
```diff
@@ -0,0 +1,24 @@
...
+You only must redeploy the CA certificate if you are not using a certification authority that is well known.
+
+You can list well-known certificate authorities as follows:
+
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+awk -v cmd='openssl x509 -noout -subject' '
+    /BEGIN/{close(cmd)};{print | cmd}' < /etc/ssl/certs/ca-certificates.crt
+----
+====
```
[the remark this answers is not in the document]
ShimShtein: On my developer machine `/etc/ssl/certs` folder contains more than one file with different filenames. 

```
for f in /etc/ssl/certs/*.crt; do awk -v cmd='openssl x509 -noout -subject' '
    /BEGIN/{close(cmd)};{print | cmd}' < ${f}; done;
```
will work for all files in the directory.

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-ansible-rex.adoc
ShimShtein: @ekohl Almost. We are missing `get "/:kind/:template" do` variant to get it working

ShimShtein: PR: https://github.com/theforeman/smart-proxy/pull/904

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-manually.adoc
Lennonka: Done. Please, review.

Lennonka:
@ShimShtein @ehelms Are we happy now?

Lennonka:
Unless there are further comments within the next 24 hours, I suggest we merge it.

maximiliankolb (approved):
Thanks Lena, LGTM.

Lennonka:
The fact that it's for self-signed CA certs only seemed important, so I've added it to the titles of the Planning module and the assembly.

@maximiliankolb Can you please give it one more go?

maximiliankolb (approved):
Yes, still LGTM.

Lennonka:
Cherry-picked:
- dad55b28eb..4880f92d6a  3.12 -> 3.12

guides/common/modules/proc_deploying-the-ca-certificate-on-a-host-by-using-script-rex.adoc
```diff
@@ -0,0 +1,32 @@
...
+= Deploying the CA certificate on a host by using Script REX
+
+You can use remote execution (REX) with the Script provider to deploy the CA certificate.
+
+include::snip_prerequisites-deploying-ca-cert-rex.adoc[]
+
+.Procedure
+. In the {ProjectWebUI}, navigate to *Monitor* > *Jobs*.
+. Click *Run Job*.
+. From the *Job category* list, select `Commands`.
```
apinnick, line 11: @Lennonka I realize that this has already been merged. I found this PR when I was trying to figure out the RN for this feature. UI elements and options should be marked up as bold, not with backticks.

PR #3197, opened 2024-08-07
Upstreamize Hammer guide

Lennonka:
Offering the Hammer guide to upstream.

* [x] Foreman 3.12

### DIFF
--- a/guides/common/modules/con_introduction-to-hammer.adoc
+++ b/guides/common/modules/con_introduction-to-hammer.adoc
@@ -65,7 +65,7 @@ There are three methods of hammer authentication:
 
 The hammer configuration file method is recommended when running commands automatically.
 For example, running {Project} maintenance commands from a cron job.
-When running commands manually, Red{nbsp}Hat recommends using the hammer authentication session and providing credentials with each command.
+When running commands manually, {Team} recommends using the hammer authentication session and providing credentials with each command.
 
 [[sect-CLI_Guide-Authentication-Hammer_Authentication_Session]]
 === Hammer authentication session


guides/doc-Hammer_CLI/topics/con_introduction-to-hammer.adoc
```diff
@@ -0,0 +1,457 @@
...
+endif::[]
+ifndef::satellite[]
+??? is there an alternative for Foreman/Katello?
+endif::[]
+. Enable the {Project} Utils module:
++
+[options="nowrap" subs="verbatim,quotes,attributes"]
+----
+# dnf module enable satellite-utils:el8
+----
```
Lennonka, line 197: I need help here. :)

guides/doc-Hammer_CLI/topics/con_introduction-to-hammer.adoc
```diff
@@ -0,0 +1,457 @@
...
+# hammer auth logout
+----
+
+[[sect-CLI_Guide-Authentication-Hammer_Configuration_File]]
+=== Hammer configuration file
+
+If you ran the {Project} installation with `--foreman-initial-admin-username` and `--foreman-initial-admin-password` options, credentials you entered are stored in the `~/.hammer/cli.modules.d/foreman.yml` configuration file, and hammer does not prompt for your credentials.
+
+You can also add your credentials to the `~/.hammer/cli.modules.d/foreman.yml` configuration file manually:
```
ekohl, line 122: Might be good to add that the installer won't update the file and if you changed your credentials then you need to do that manually. You already have the instructions.

guides/doc-Hammer_CLI/topics/con_introduction-to-hammer.adoc
```diff
@@ -0,0 +1,457 @@
...
++
+[options="nowrap" subs="verbatim,quotes,attributes"]
+----
+# dnf module enable satellite-utils:el8
+----
+. Install Hammer Ruby gem:
++
+[options="nowrap" subs="verbatim,quotes,attributes"]
+----
+# {client-package-install-el8} rubygem-hammer_cli_katello
```
ekohl, line 202: I'd argue this instruction is wrong. It installs a single plugin to Hammer that ends up pulling in some others. For satellite I think the instruction should be to install `satellite-cli`, but I'm not 100% sure we publish that in our utils repo.

guides/doc-Hammer_CLI/topics/con_introduction-to-hammer.adoc
```diff
@@ -0,0 +1,457 @@
...
+----
++
+. Edit the `:host:` entry in the `/etc/hammer/cli.modules.d/foreman.yml` file to include the {Project} IP address or FQDN.
```
ekohl, line 205: I first thought the IP is wrong, but then think we can even reduce it further.
```suggestion
. Set the `:host:` entry in the `/etc/hammer/cli.modules.d/foreman.yml` file to the {Project} URL.
```

guides/doc-Hammer_CLI/topics/con_introduction-to-hammer.adoc
```diff
@@ -0,0 +1,457 @@
...
+For example, set the number of entries displayed per request in the Hammer output by changing the following line:
+
+----
+:per_page: 30
+----
+
+This setting is an equivalent of the `--per-page` Hammer option.
+
+[[sect-CLI_Guide-Configuring_Hammer_Logging]]
+== Configuring Hammer logging
```
ekohl, line 292: This is specifically for [hammer-cli-foreman-admin](https://github.com/theforeman/hammer-cli-foreman-admin) which is a very unusual plugin. Normally Hammer is a tool that talks to the Foreman API over the network. hammer-cli-foreman-admin actually operates on the server itself. I'd love to get rid of it.

We already have a section that can change the log levels (https://docs.theforeman.org/nightly/Administering_Project/index-katello.html#Logging_and_Reporting_Problems_admin). I'd be my preference to use that instead as the single way to configure logging.

guides/doc-Hammer_CLI/topics/con_introduction-to-hammer.adoc
ekohl: For upstream it's the exact same repository as the server so you should be able to reuse that module

guides/doc-Hammer_CLI/topics/con_introduction-to-hammer.adoc
Lennonka: I checked on Ohsnap and yes, we publish the `satellite-cli` package in the Utils repo.

guides/common/attributes-titles.adoc
```diff
@@ -7,6 +7,7 @@
 :ContentManagementDocTitle: Managing content
 :ConvertingHostRHELDocTitle: Converting a host to RHEL
 :DeployingAWSDocTitle: Deploying {ProjectName} on Amazon Web Services
+:HammerDocTitle: Hammer CLI reference
```
Lennonka, line 10: @apinnick I would like to have your opinion.

I want to change the title, but I'm not sure whether to call it _Hammer CLI reference_ or _Using Hammer CLI_. In downstream, we put it into the Reference category, but it contains procedures as well. What do you think?

Lennonka:
I've made an attempt to rewrite the standalone installation procedure for upstream.
And incorporated all suggestions.

guides/common/attributes-titles.adoc
apinnick: For d/s, I would prefer "Hammer CLI guide" so that it is consistent with the existing "API Guide" (not sure why that title capitalizes "guide" -- will check). 

I would not use the contents  alone for determining the title. All API and CLI guides contain procedures for installing, configuring, setting log levels, etc. 

API and tool guides tend to be named either "Using the 'x' tool" or "X reference guide". Personally, I do not think it matters which format is used as long as the guide names are consistent. You could change the API guide to "Using the Satellite REST API" and call this guide "Using the Hammer CLI tool". :-)  

Consistency is the most important consideration, in my opinion.

Lennonka: I intend to change the title of the API guide too #3212 . I don't like having "guide" in the title, when we're avoiding it everywhere else.

apinnick: +1 I also don't like "guide" in the title.

Lennonka:
I've moved the assembly and modules where they belong. I suggest merging it as is and making further changes in other PRs.

maximiliankolb (commented):
* I noticed that many modules start with a level 2 heading. I am OK with this for now and can change it later in a follow-up PR.
* Unsure what "aqua-background" is for/ why "org" and "loc" has to be blue/yellow.
* Unsure if we _want_ to keep some rather long examples of stdout.

I had a glance at most changes but not all. I appreciate the effort to bring it upstream. All my suggestion can be applied by you or me after merging.

Merging now and improving later is perfectly fine for me. Any other concerns or approval?

Lennonka:
> * Unsure what "aqua-background" is for/ why "org" and "loc" has to be blue/yellow.

It's sort of a flag that is introduced and explained in General information. It can be done diffetently, but for now I'm keeping it as is.

guides/common/assembly_hammer-reference.adoc
Lennonka, line 1: I'm not sure what to do with this file.

The problem is that the reference should be generated separately for Foreman, Katello, orcharhino and Satellite.
If I generate it in downstream and keep this file in upstream, our sync will overwrite it from upstream again. Therefore, it would be better to remove it in upstream.
However, if I don't include it in upstream, the build will not pass because it's required by `master.adoc`.

So the solution might be to exclude this file from syncing in downstream. Will you be able to do that as well, @maximiliankolb?

maximiliankolb: Yeah, I can handle this. No worries.

maximiliankolb:
@Lennonka Just to ensure that I understand the intended workflow:

* for upstream Foreman and Foreman+Katello, we can check in the output `ref_*` files from `hammer full-help --md > hammer_full_help.md && ./generate-hammer-reference.sh hammer_full_help.md`
* for downstream, we can do the same on Satellite/orcharhino and check in those files to downstream-only branches?

I ran this on orcharhino and got 95 files in total. Do you envision generating the references files on a minimal installation, with select plugins, or with all available plugins in Satellite/orcharhino?

Lennonka:
> * for upstream Foreman and Foreman+Katello, we can check in the output `ref_*` files from `hammer full-help --md > hammer_full_help.md && ./generate-hammer-reference.sh hammer_full_help.md`
> * for downstream, we can do the same on Satellite/orcharhino and check in those files to downstream-only branches?

That is correct.

> I ran this on orcharhino and got 95 files in total. Do you envision generating the references files on a minimal installation, with select plugins, or with all available plugins in Satellite/orcharhino?

In Satellite, we have all plugins enabled AFAIK, so that's a given. For Foreman and Foreman+Katello I would go for generating the full help on an instance with all Hammer plugins enabled, either with or without Katello plugin accordingly. I think it's better to include all plugins regardless of what the user has enabled. Better to have it and not use it than not have it when needed :)

guides/common/modules/con_introduction-to-hammer.adoc
```diff
@@ -0,0 +1,437 @@
...
+In either case, hammer requires {Project} credentials for authentication.
+There are three methods of hammer authentication:
+
+* Hammer authentication session
+* Storing credentials in the hammer configuration file
+* Providing credentials with each hammer command
+
+The hammer configuration file method is recommended when running commands automatically.
+For example, running {Project} maintenance commands from a cron job.
+When running commands manually, Red{nbsp}Hat recommends using the hammer authentication session and providing credentials with each command.
```
maximiliankolb, line 68:
```suggestion
When running commands manually, {Team} recommends using the hammer authentication session and providing credentials with each command.
```

maximiliankolb (changes requested):
One small request.

I had a look at the built docs without reading/testing everything. I think we can merge this as is and then improve later.

I will have a closer look at this as soon as I integrate this into downstream docs.

As there are no other comments, I say LGTM; please merge after a final tiny edit & green GHA.

Thanks again @Lennonka for this effort! :rocket:

Lennonka:
Cherry-picked:
- cee50ae19d..c750bf7da3  3.12 -> 3.12

PR #3200, opened 2024-08-08
Document dns_nsupdate provider in managing DNS section

ekohl:
This takes the idea that there should be a central place with documentation on all DNS providers. Not spread over multiple guides. It starts with the `dns_nsupdate` provider, but this should be completed with all of the providers.

This is split off from https://github.com/theforeman/foreman-documentation/pull/2933 to keep the scope limited. Infoblox turned out to be much harder than expected.

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_configuring-dns-nsupdate.adoc
+++ b/guides/common/modules/proc_configuring-dns-nsupdate.adoc
@@ -1,0 +1,20 @@
+[id="configuring_dns_nsupdate_{context}"]
+= Configuring dns_nsupdate
+
+The _dns_nsupdate_ DNS provider manages DNS records using the `nsupdate` utility.
+You can use _dns_nsupdate_ with any DNS server compatible with https://www.rfc-editor.org/rfc/rfc2136[RFC2136].
+By default, _dns_nsupdate_ installs the ISC BIND server.
+For installation without ISC BIND, see xref:configuring-external-dns_{context}[].
+
+.Procedure
+* Configure `dns_nsupdate`:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# {foreman-installer} \
+--foreman-proxy-dns true \
+--foreman-proxy-dns-provider nsupdate \
+--foreman-proxy-dns-managed true \
+--foreman-proxy-dns-zone _example.com_ \
+--foreman-proxy-dns-reverse _2.0.192.in-addr.arpa_
+----
--- a/guides/common/assembly_managing-dns-on-smart-proxies.adoc
+++ b/guides/common/assembly_managing-dns-on-smart-proxies.adoc
@@ -1,5 +1,7 @@
 include::modules/con_managing-dns-by-using-smartproxy.adoc[]
 
+include::modules/proc_configuring-dns-nsupdate.adoc[leveloffset=+1]
+
 ifndef::satellite[]
 include::modules/proc_configuring-dns-libvirt.adoc[leveloffset=+1]
 
--- a/guides/common/modules/con_managing-dns-by-using-smartproxy.adoc
+++ b/guides/common/modules/con_managing-dns-by-using-smartproxy.adoc
@@ -17,8 +17,8 @@ ifndef::satellite[]
 * `dns_libvirt` {endash} Dnsmasq DNS via libvirt API.
 For more information, see xref:configuring_dns_libvirt_{context}[].
 endif::[]
-* `dns_nsupdate` {endash} Dynamic DNS update by using nsupdate.
-For more information, see {ProvisioningDocURL}Using_Infoblox_as_DHCP_and_DNS_Providers_provisioning[Using Infoblox as DHCP and DNS Providers] in _{ProvisioningDocTitle}_.
+* `dns_nsupdate` {endash} Dynamic DNS update using nsupdate.
+For more information, see xref:configuring_dns_nsupdate_{context}[].
 * `dns_nsupdate_gss` {endash} Dynamic DNS update with GSS-TSIG.
 For more information, see xref:configuring-dynamic-dns-update-with-gss-tsig-authentication_{context}[].
 ifndef::satellite[]


guides/common/modules/proc_configuring-dns-nsupdate.adoc
```diff
@@ -0,0 +1,19 @@
+[id="configuring_dns_nsupdate_{context}"]
+= Configuring dns_nsupdate
+
+The _dns_nsupdate_ DNS provider manages DNS records using nsupdate.
+It can be used with any https://www.rfc-editor.org/rfc/rfc2136[RFC2136] compatible nameserver, but by default ISC BIND is installed.
```
ekohl, line 5: Interesting: RFC2136 has 1 mention of nameserver and the rest is name server.

ekohl: The rabbit hole goes deeper. https://www.rfc-editor.org/rfc/rfc8499#section-6 suggests to me that it's now preferred to use DNS server

maximiliankolb (approved):
Checked `foreman-installer` on Foreman 3.11: option is present :heavy_check_mark: 

I believe you wanted to add additional information; feel free to extend your commit message or get additional technical ACK if you deem necessary.

LGTM from my side.

guides/common/modules/proc_configuring-dns-nsupdate.adoc
```diff
@@ -0,0 +1,19 @@
+[id="configuring_dns_nsupdate_{context}"]
+= Configuring dns_nsupdate
+
+The _dns_nsupdate_ DNS provider manages DNS records using nsupdate.
```
asteflova, line 4: Just to add markup and a basic introduction of what nsupdate is.
```suggestion
The _dns_nsupdate_ DNS provider manages DNS records using the `nsupdate` utility.
```

guides/common/modules/proc_configuring-dns-nsupdate.adoc
```diff
@@ -0,0 +1,19 @@
...
+It can be used with any https://www.rfc-editor.org/rfc/rfc2136[RFC2136] compatible DNS server, but by default ISC BIND is installed.
```
asteflova, line 5:
```suggestion
You can use _dns_nsupdate_ with any https://www.rfc-editor.org/rfc/rfc2136[RFC2136]-compatible DNS server, but by default ISC BIND is installed.
```
or maybe this would be better to avoid the need to use the hyphen:
```suggestion
You can use _dns_nsupdate_ with any DNS server compatible with https://www.rfc-editor.org/rfc/rfc2136[RFC2136], but by default ISC BIND is installed.
```
You know what, let's split this into two sentences and clarify the second one (I really am not sure who does the installing, I'm going to assume it's dns_nsupdate):
```suggestion
You can use _dns_nsupdate_ with any DNS server compatible with https://www.rfc-editor.org/rfc/rfc2136[RFC2136].
By default, _dns_nsupdate_ installs the ISC BIND server.
```

guides/common/modules/proc_configuring-dns-nsupdate.adoc
```diff
@@ -0,0 +1,19 @@
+[id="configuring_dns_nsupdate_{context}"]
+= Configuring dns_nsupdate
```
asteflova, line 2: I'd like to see more details in the heading about the use case. I'm going to make a guess as to what it could be, feel free to change it to something more appropriate.
```suggestion
= Configuring _dns_nsupdate_ as the DNS provider for {SmartProxy}
```
The idea is to include useful keywords into the heading and to avoid a short heading that doesn't say much about the section ([RH Supplementary Style Guide](https://redhat-documentation.github.io/supplementary-style-guide/#_principle_3_titles_and_headings)).

guides/common/modules/proc_configuring-dns-nsupdate.adoc
```diff
@@ -0,0 +1,19 @@
...
+
+The _dns_nsupdate_ DNS provider manages DNS records using nsupdate.
+It can be used with any https://www.rfc-editor.org/rfc/rfc2136[RFC2136] compatible DNS server, but by default ISC BIND is installed.
+For installation without ISC BIND, see xref:configuring-external-dns_{context}[].
+
+.Procedure
+* You can use `{foreman-installer}` to configure `dns_nsupdate`:
```
asteflova, line 9:
```suggestion
* Use `{foreman-installer}` to configure `dns_nsupdate`:
```
or even
```suggestion
* Configure `dns_nsupdate`:
```
because users can see that they are using the installer, no need to tell them twice. Instead, you _could_ provide more details or a bit of context about the actual command example, but then again, that's all included in --help already so this is probably just fine as is.

ekohl:
I should have noted that we already sort of have this documented in https://docs.theforeman.org/nightly/Installing_Server/index-katello.html#reverting-to-internal-dns-service_foreman (where I just noticed we have a dead anchor) and https://docs.theforeman.org/nightly/Installing_Server/index-katello.html#configuring-dns-dhcp-and-tftp_foreman. My primary aim was to provide readers with an easy way to set up only DNS instead of configuring DHCP & TFTP as well since they don't need to run on the same server. My general recommendation is to aim for a single DNS Smart Proxy unless you have a specific need for multiple.

guides/common/modules/proc_configuring-dns-nsupdate.adoc
asteflova: Oh, I hadn't noticed the other headings and sections. I'm actually not sure now: I still think those extra details in the heading(s) would help but then again, this PR is about adding one additional provider. So feel free to disregard my suggestions and just mimic the existing structure.

guides/common/modules/proc_configuring-dns-nsupdate.adoc
ekohl: I think it's an often repeating thing. We have many procedures that are essentially a single step that is "run the installer with these arguments". I never quite know what's best because that header in the procedure almost feels redundant.

guides/common/modules/proc_configuring-dns-nsupdate.adoc
ekohl: I kept it simple for now.

ekohl:
Rebased to resolve conflicts and I included @asteflova's suggestions, except the header one.

asteflova (approved):

PR #3209, opened 2024-08-12
Add comment to update json in downstream

Lennonka:

### DIFF
--- a/guides/common/attributes-satellite.adoc
+++ b/guides/common/attributes-satellite.adoc
@@ -8,6 +8,10 @@
 :BaseURL: {RHDocsBaseURL}red_hat_satellite/{ProjectVersion}/html-single/
 
 // URLs (published on Red Hat Portal)
+// IMPORTANT: When you change any guide URLs, also update the following:
+// - guides/upstream_filename_to_satellite_link.json
+// - lib/foreman_theme_satellite/documentation.rb in RedHatSatellite/foreman_theme_satellite
+// - downstream_filename_to_link.json in downstream
 :AdministeringDocURL: {BaseURL}administering_red_hat_satellite/index#
 :ConfiguringLoadBalancerDocURL: {BaseURL}configuring_capsules_with_a_load_balancer/index#
 :ContentManagementDocURL: {BaseURL}managing_content/index#


maximiliankolb (approved):

guides/common/attributes-satellite.adoc
```diff
@@ -8,6 +8,7 @@
 :BaseURL: {RHDocsBaseURL}red_hat_satellite/{ProjectVersion}/html-single/
 
 // URLs (published on Red Hat Portal)
+// IMPORTANT: When you change any guide URLs, also update the json file in downstream
```
ekohl, line 11: Given today we have multiple places, please also list https://github.com/theforeman/foreman-documentation/blob/master/guides/upstream_filename_to_satellite_link.json and https://github.com/RedHatSatellite/foreman_theme_satellite/blob/develop/lib/foreman_theme_satellite/documentation.rb

ekohl (changes requested):

guides/common/attributes-satellite.adoc
```diff
@@ -8,6 +8,10 @@
 :BaseURL: {RHDocsBaseURL}red_hat_satellite/{ProjectVersion}/html-single/
 
 // URLs (published on Red Hat Portal)
+// IMPORTANT: When you change any guide URLs, also update the following:
+// - guides/upstream_filename_to_satellite_link.json
+// - lib/foreman_theme_satellite/documentation.rb in RedHatSatellite/foreman_theme_satellite
```
ShimShtein, line 13: Actually you don't have to update this one right away. Once the new TOC will be generated, it will detect the incompatible links automatically. 
To fix the incompatibility a pr with the new toc and a fix to documentation.rb should be created.

ekohl: Ideally we would sync that mapping so we don't maintain the same thing in 3 places, but I explicitly asked for this because it tracks the related places. I don't think a docs writer monitors foreman_theme_satellite and may not be aware of it. After https://github.com/RedHatSatellite/foreman_theme_satellite/pull/58 it will have a mapping that must be maintained.

ekohl (approved):

Lennonka:
I've updated the commit message to reflect the changes.

PR #3211, opened 2024-08-12
Migrate PR requirements to the PR template

asteflova:
What: I'm proposing to move the expected contents of a PR from CONTRIBUTING.md directly to the PR template. This PR only moves two bullet points (the _what_ and the _why_) because I find that the rest is unclear or very difficult to enforce anyway. If you feel we should include the dropped bullet points as well, feel free to suggest more headings.

Why: Any documentation works best when presented at the moment when users need it. And I don't think adding these two points to the PR template makes it too cumbersome.

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,9 @@
+#### What changes are you introducing?
+
+#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)
+
+#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)
+
 
 * [x] I am okay with my commits getting squashed when you merge this PR.
 * [ ] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,6 +4,7 @@
 
 #### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)
 
+#### Checklists
 
 * [x] I am okay with my commits getting squashed when you merge this PR.
 * [ ] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,7 +10,7 @@ We respect each others time and energy spent on Foreman documentation.
 ## Maintainers
 
 * help less experienced community members with git, Github, PRs, asciidoc, and asciidoctor.
-* expect a basic effort in contributions.
+* expect a basic effort in contributions, including the [pull request template](PULL_REQUEST_TEMPLATE.md) being filled out.
 * only merge PRs if the Github Actions are green.
 * read the proposed change and make friendly and helpful comments.
 * ping community members with more experience if they are unsure about specific details.
@@ -49,8 +49,6 @@ See [Capitalization](#Capitalization).
 
 Please explain why someone else should merge your proposed change:
 
-* What has been changed?
-* Why has it been changed?
 * Are there any alternatives to my PR?
 * Are there any regressions or downsides to my PR?
 * Does my PR make multiple changes and therefore require multiple commits?
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,15 +45,6 @@ See [LICENSE](LICENSE).
 * [ ] I am familiar to proper capitalization for project-specific terminology.
 See [Capitalization](#Capitalization).
 
-## Crafting Great Pull Requests
-
-Please explain why someone else should merge your proposed change:
-
-* Are there any alternatives to my PR?
-* Are there any regressions or downsides to my PR?
-* Does my PR make multiple changes and therefore require multiple commits?
-Ideally, each individual change correlates to an individual commit.
-
 ## Capitalization
 
 All headings are sentence case.


CONTRIBUTING.md
```diff
@@ -45,17 +45,6 @@ See [LICENSE](LICENSE).
 * [ ] I am familiar to proper capitalization for project-specific terminology.
 See [Capitalization](#Capitalization).
 
-## Crafting Great Pull Requests
```
maximiliankolb, line 48: sad to see this section gone.

maximiliankolb (approved):
Is there any way to see the rendered GHA PR template? diff in PR template LGTM

asteflova:
> Is there any way to see the rendered GHA PR template? diff in PR template LGTM

I don't know but the markup is a shameless copy of https://github.com/Katello/katello/blob/master/.github/PULL_REQUEST_TEMPLATE.md?plain=1 so the Katello PR template should give you an idea of what it will look like.

CONTRIBUTING.md
asteflova: [I feel your pain.](https://www.goodreads.com/quotes/79715-in-writing-you-must-kill-all-your-darlings)

asteflova: I chose to include the bare minimum in the template at the first iteration, but as hinted in the PR description, I was hoping for a conversation. So thank you for starting it!

Looking at the bullet points I propose to drop:

- Lines 54-55: Are these questions something that we really expect contributors to answer? With most PRs, I’ve seen _reviewers_ ask these questions and propose alternative solutions to PR authors. I think that’s more useful than making less experienced contributors sweat over trying to come up with an alternative just for the sake of meeting an expectation.

- For lines 56-57: I've just noticed that this is actually included [CONTRIBUTING.md](https://github.com/theforeman/foreman-documentation/blob/f0d81c58c7943f6349bf556dd267768052dafcb1/CONTRIBUTING.md?plain=1#L42), which I'd say means we can safely drop it here. But on a more general note: While I understand the value in this guideline, I also know that I’m one of the people who have a very hard time following it; more importantly, I don't think it has been a blocker in any of my PRs. So, is it something that maintainers should require of contributors? And actively remind them of it as part of reviews?

As for https://cbea.ms/git-commit/, that's already included in [CONTRIBUTING.md](https://github.com/theforeman/foreman-documentation/blob/f0d81c58c7943f6349bf556dd267768052dafcb1/CONTRIBUTING.md?plain=1#L37 ).

ekohl: Ah right, I only looked at the diff and that wasn't visible. Perhaps the two chapters should be merged? So there's just a single section. As you say, there was already a lot of overlap.

> * Lines 54-55: Are these questions something that we really expect contributors to answer? With most PRs, I’ve seen _reviewers_ ask these questions and propose alternative solutions to PR authors. I think that’s more useful than making less experienced contributors sweat over trying to come up with an alternative just for the sake of meeting an expectation.

I think this is more a case of SHOULD (in the [rfc2119](https://www.rfc-editor.org/rfc/rfc2119) sense). If you have thought about those, write them down. It saves reviewers asking the question.

> * For lines 56-57: I've just noticed that this is actually included [CONTRIBUTING.md](https://github.com/theforeman/foreman-documentation/blob/f0d81c58c7943f6349bf556dd267768052dafcb1/CONTRIBUTING.md?plain=1#L42), which I'd say means we can safely drop it here. But on a more general note: While I understand the value in this guideline, I also know that I’m one of the people who have a very hard time following it; more importantly, I don't think it has been a blocker in any of my PRs. So, is it something that maintainers should require of contributors? And actively remind them of it as part of reviews?

I think it's a skill you can develop, but it takes practice. Properly separating your changes is hard and there are many "it depends" cases.

With code there's a general rule that every git commit in itself should be valid: it builds and works. I think that also applies to docs. So every commit needs to be at least as big to consistently change it. Concrete example: if you change an attribute, every instance of that attribute must still make sense.

Then for how small should it be? If you take it back to a commit message: once you start using `and` in the summary (first line of the commit message) you're probably combining things that don't need to be combined.

Another is to look at how a reviewer might look at them. Try to separate out things. For example, if you move some text around and make changes to it, that's often hard for a reviewer to verify. Splitting that up in a "move" commit and a "modify" commit it's easier to verify what you're changing.

asteflova: My approach to writing is often based on collecting different pieces of information from different sources (people, testing, man pages, other docs). Sometimes those sources contradict each other, in which case it’s very important to create commits that track "this person told me this" and "that person told me this" rather than commits that "move" and commits that “modify”. In the end, my approach absolutely results in PRs that mix together multiple changes and where looking at the commit history might make you want to cry.

There might very well be a better way to do things but very few people actually try. Which brings me to the main point I’m trying to make: When a maintainer sees a PR that doesn’t follow this guideline, do they ask the author to fix it? No, they don’t. Which made me question why we have contribution guidelines that we don’t even expect anyone to follow.

This is not just a pet peeve of mine (okay, it is…) but it has practical implications too: There were times when I wanted to point out to contributors that they were not following certain guidelines in CONTRIBUTING.md. But I couldn’t because what if that person correctly noted that most of the guidelines were regularly being ignored anyway?

And to conclude what has at this point become an internal monologue: I think the answer might lie in the proper usage of MUST, MUST NOT, and SHOULD in CONTRIBUTING.md. I will explore that approach in another PR.

asteflova: In any case, I'm adding one more line to the PR template to cover the dropped bullet points. Phrasing them as a question ("Do you see any downsides?") still gives any contributor the chance to reply "No".

ekohl: > My approach to writing is often based on collecting different pieces of information from different sources (people, testing, man pages, other docs). Sometimes those sources contradict each other, in which case it’s very important to create commits that track "this person told me this" and "that person told me this" rather than commits that "move" and commits that “modify”. In the end, my approach absolutely results in PRs that mix together multiple changes and where looking at the commit history might make you want to cry.

And development is often similar. For non-trivial changes it's hardly ever a straight line from A to B.

There are different visions on how you should structure reviews. Some argue a PR should show the complete evolution of all changes, including all mistakes. Others prefer to only see the end result. There's also a middle ground where while it's being reviewed it's the complete evolution and in the end you clean it up.

I learned to do code review in a system where every single commit was reviewed (first mailing lists, later [gerrit](https://www.gerritcodereview.com/)). I'm also a bit of a perfectionist, so I'm in the camp where every commit tells a story. Then you can look through the git log and easily see how the repository evolved. It also makes cherry picking easier.

In the end it's important that the maintainers as a whole can work effectively, which is why having these discussions is great.

It's fair to say that guidelines that are never followed are useless, but guidelines are also not absolute. In the ideal situation for every guideline we'd have a motivation of why it is useful and why we have it. If you know the goal, it's easier to determine if you can deviate.

> In any case, I'm adding one more line to the PR template to cover the dropped bullet points. Phrasing them as a question ("Do you see any downsides?") still gives any contributor the chance to reply "No".

If we look at Katello's template they have:

> Considerations taken when implementing this change?

That's precisely the section where you can add any thoughts (like downsides) you've had.

To me some of the worst PRs are complex changes without a description of what it does, why it does it and how. Then I effectively have to reverse engineer what the author tried to do. And conversely, some of the better ones are where the author helps me to understand their goals. If they tried the obvious solution and it didn't work for some reason, share that's why you went for a complex solution because as a reviewer that's likely what I'll ask anyway.

asteflova:
@maximiliankolb I've made more changes based on the conversation with @ekohl. Does your ack still apply? You can see what the new template looks like in action in https://github.com/theforeman/foreman-documentation/pull/3214.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,10 @@
+#### What changes are you introducing?
+
+#### Why are you introducing these changes? (Explain or link to a Jira ticket)
```
ekohl, line 3: I generally don't like issues that don't have context and only link to some bug tracker. Can we separate those out into a References section? References can be to any bug tracker, related PRs (which can even be dependencies) or external documentation.

My aim would be to encourage people to always explain why they're doing something. Jira issues can describe a bug, but that can be a very long conversation. Rather than having the reviewer read up on the whole history it's easier if the PR author summarizes it.

Of course, for a trivial issue like a typo fix it's obvious and I don't expect anyone to fill it in.

asteflova: > I generally don't like issues that don't have context and only link to some bug tracker. Can we separate those out into a References section? References can be to any bug tracker, related PRs (which can even be dependencies) or external documentation.
> 
> My aim would be to encourage people to always explain why they're doing something. Jira issues can describe a bug, but that can be a very long conversation. Rather than having the reviewer read up on the whole history it's easier if the PR author summarizes it.

When I wrote line 3, I was again thinking of the art of the possible. What I too often see on other projects is that contributors don't even bother to fill out the PR template. That's very likely to happen here too, in which case giving the option to only link to an issue is a compromise: from a contributor's perspective, it's easy; from a reviewer's perspective, it's better than nothing.

> Of course, for a trivial issue like a typo fix it's obvious and I don't expect anyone to fill it in.

I understand what you mean but then again, everybody doesn't have the same understanding of what "trivial" means. A typo? Sure! But how about adding a command option? To some, it may be trivial too so they wouldn't be inclined to fill in the section; to me, it's not trivial at all because it's also a question of duplicating man pages or `--help` outputs and as a reviewer, I need to know what problem is actually being solved because that also affects where to place that option... So I would expect at least a word or two for every PR.

ekohl: > When I wrote line 3, I was again thinking of the art of the possible. What I too often see on other projects is that contributors don't even bother to fill out the PR template. That's very likely to happen here too, in which case giving the option to only link to an issue is a compromise: from a contributor's perspective, it's easy; from a reviewer's perspective, it's better than nothing.

My issue is with Jira. That's a Red Hat thing that upstream shouldn't really care about. It can be an interesting reference, but nothing upstream should care about. So if anything I'd go with something like this.
```suggestion
#### Why are you introducing these changes? (Explanation, links to references, issues, etc)
```

> I understand what you mean but then again, everybody doesn't have the same understanding of what "trivial" means. A typo? Sure! But how about adding a command option? To some, it may be trivial too so they wouldn't be inclined to fill in the section; to me, it's not trivial at all because it's also a question of duplicating man pages or `--help` outputs and as a reviewer, I need to know what problem is actually being solved because that also affects where to place that option... So I would expect at least a word or two for every PR.

I was trying to indicate that they fall in the SHOULD category, not MUST.

ekohl (approved):

maximiliankolb (approved):
LGTM

PR #3213, opened 2024-08-13
Clarify the support level for the Quarkus distribution of Keycloak

asteflova:
The default distribution of Keycloak is now based on Quarkus. While we have the procedures for Quarkus-based Keycloak in place, we are still not clear on whether we should label the Wildfly-based Keycloak procedures as deprecated or whether we can actually support the Quarkus-based Keycloak procedures.

This PR introduces a note that describes the support level and provides links to docs about migrating Keycloak.

https://issues.redhat.com/browse/SAT-27194

UPDATE: I've been adding commits as the discussion around the support levels evolved. The latest state doesn't use the term "deprecated" for the procedures to configure the Wildfly distribution of Keycloak (because it's actually Keycloak that was deprecated, not the procedure in Satellite docs) and also excludes the Quarkus-based procedures from Satellite builds.

* [x] Foreman 3.12/Katello 4.14
* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/assembly_configuring-external-authentication.adoc
+++ b/guides/common/assembly_configuring-external-authentication.adoc
@@ -19,6 +19,7 @@ ifdef::satellite[]
 :!keycloak:
 endif::[]
 
+ifndef::satellite[]
 // The following ifdef sets :keycloak: to `Red Hat build of Keycloak` for Satellite builds only. For, non-Satellite builds, :keycloak: stays set to `Keycloak`.
 ifdef::satellite[]
 :keycloak: {keycloak-quarkus}
@@ -32,6 +33,7 @@ ifndef::parent-context[:!context:]
 // The following ifdef unsets :keycloak: because in Satellite builds, it's always either {keycloak-wildfly} or {keycloak-quarkus}.
 ifdef::satellite[]
 :!keycloak:
+endif::[]
 endif::[]
 
 include::assembly_configuring-active-directory-as-an-external-identity-provider-for-project.adoc[leveloffset=+1]
--- a/guides/common/modules/con_configuring-keycloak-wildfly-authentication-for-project.adoc
+++ b/guides/common/modules/con_configuring-keycloak-wildfly-authentication-for-project.adoc
@@ -1,4 +1,4 @@
-[id="configuring-keycloak-authentication-for-project_{context}"]
+[id="configuring-keycloak-wildfly-authentication-for-project_{context}"]
 = Configuring {keycloak-wildfly} authentication for {Project}
 
 {keycloak} is an open-source identity and access management solution that provides authentication features, such as single sign-on functionality, user federation, or centralized authentication management.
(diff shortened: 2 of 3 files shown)


asteflova:
In its current state, we can use this PR if the Quarkus distribution of Keycloak is supported in Satellite.

If we revert the last commit, we can use this PR if Quarkus-based Keycloak is not supported in Satellite.

asteflova:
I'm removing the deprecation notice (because it's actually the Wildfly distribution of Keycloak that is being deprecated rather than a component of Foreman) and I'm also excluding the Quarkus-based procedures from Satellite builds (because while we have tested them and they work, we also want more time before we commit to full support).

apinnick (approved):
LGTM

guides/common/assembly_configuring-external-authentication.adoc
```diff
@@ -19,6 +19,7 @@ ifdef::satellite[]
 :!keycloak:
 endif::[]
 
+ifndef::satellite[]
 // The following ifdef sets :keycloak: to `Red Hat build of Keycloak` for Satellite builds only. For, non-Satellite builds, :keycloak: stays set to `Keycloak`.
```
maximiliankolb, line 23: I think you could drop lines 23-27 + 33-36.

guides/common/modules/con_configuring-keycloak-wildfly-authentication-for-project.adoc
```diff
@@ -1,6 +1,17 @@
-[id="configuring-keycloak-authentication-for-project_{context}"]
+[id="configuring-keycloak-wildfly-authentication-for-project_{context}"]
 = Configuring {keycloak-wildfly} authentication for {Project}
 
+ifndef::satellite[]
+[NOTE]
+====
+The default {keycloak} distribution is now based on Quarkus.
+
+{keycloak-quarkus} replaces {keycloak-wildfly} in {Project} deployments. For information about configuring {keycloak-quarkus} authentication, see xref:configuring-keycloak-quarkus-authentication-for-project_keycloak-quarkus[].
```
maximiliankolb, line 9:
```suggestion
{keycloak-quarkus} replaces {keycloak-wildfly} in {Project} deployments.
For information about configuring {keycloak-quarkus} authentication, see xref:configuring-keycloak-quarkus-authentication-for-project_keycloak-quarkus[].
```

guides/common/assembly_configuring-external-authentication.adoc
asteflova: I could, but it's possible we will want to re-include lines 22-37 for Satellite which would be a lot easier if those extra conditionals stayed in place. This is all only temporary until we can safely remove anything related to the Wildfly distribution of Keycloak.

asteflova:

* ce13632f9b..89c1a95938  3.12 -> 3.12
* 6d89c559c6..64fe5dbac0  3.11 -> 3.11

PR #3215, opened 2024-08-15
Add pmproxy to list of monitoring services to restart

ehelms:

This is based on feedback from https://issues.redhat.com/browse/SAT-26588

### DIFF
--- a/guides/common/modules/proc_configuring-pcp-data-collection.adoc
+++ b/guides/common/modules/proc_configuring-pcp-data-collection.adoc
@@ -64,5 +64,5 @@ endif::[]
 . Restart PCP to begin data collection:
 +
 ----
-# systemctl restart pmcd pmlogger
+# systemctl restart pmcd pmlogger pmproxy
 ----


ekohl:
@ehelms I'd prefer it if you kept the cherry picks section or explained why you think it shouldn't be picked. This looks like a bug and that's generally something we cherry pick into stable releases. Especially since it was reported as an issue.

evgeni (approved):

ehelms:
> @ehelms I'd prefer it if you kept the cherry picks section or explained why you think it shouldn't be picked. This looks like a bug and that's generally something we cherry pick into stable releases. Especially since it was reported as an issue.

I think that's where I get confused. The cherry pick pops up for me as the PR opener but I do not have an opinion on where it should be cherry picked to. I suppose someone else could mark it off but then it would appear as if I did it?

asteflova:
> I think that's where I get confused. The cherry pick pops up for me as the PR opener but I do not have an opinion on where it should be cherry picked to. I suppose someone else could mark it off but then it would appear as if I did it?

Thanks Eric, that's actually very good feedback on our contributing process. I'm currently reviewing our contribution guidelines and will take this into consideration too.

evgeni:
   c4dcff5070..a83e068f42  3.10 -> 3.10
   bf562aef70..5efe423c20  3.11 -> 3.11
   a4d4fa43a7..6733e717dc  3.9 -> 3.9

PR #3216, opened 2024-08-15
Add /var/lib/pgsql to list of high-bandwidth, low-latency directories

ehelms:

Feedback from https://issues.redhat.com/browse/SAT-24917

### DIFF
--- a/guides/common/modules/ref_storage-guidelines.adoc
+++ b/guides/common/modules/ref_storage-guidelines.adoc
@@ -9,7 +9,7 @@ This is a requirement for the `puppetserver` service to work.
 
 * Because most {ProductName} data is stored in the `/var` directory, mounting `/var` on LVM storage can help the system to scale.
 
-* Use high-bandwidth, low-latency storage for the `/var/lib/pulp/` and PostgreSQL `/var/lib/pgsql/` directories.
+* Use high-bandwidth, low-latency storage for the `/var/lib/pulp/` and PostgreSQL `{postgresql-lib-dir}` directories.
 As {ProjectName} has many operations that are I/O intensive, using high latency, low-bandwidth storage causes performance degradation.
 
 ifdef::satellite[]
--- a/guides/common/modules/ref_storage-guidelines.adoc
+++ b/guides/common/modules/ref_storage-guidelines.adoc
@@ -9,7 +9,7 @@ This is a requirement for the `puppetserver` service to work.
 
 * Because most {ProductName} data is stored in the `/var` directory, mounting `/var` on LVM storage can help the system to scale.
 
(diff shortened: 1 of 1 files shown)


guides/common/modules/ref_storage-guidelines.adoc
```diff
@@ -9,7 +9,7 @@ This is a requirement for the `puppetserver` service to work.
 
 * Because most {ProductName} data is stored in the `/var` directory, mounting `/var` on LVM storage can help the system to scale.
 
-* Use high-bandwidth, low-latency storage for the `/var/lib/pulp/` directories.
+* Use high-bandwidth, low-latency storage for the `/var/lib/pulp/` and PostgreSQL `/var/lib/pgsql/` directories.
 As {ProjectName} has many operations that are I/O intensive, using high latency, low-bandwidth storage causes performance degradation.
 Ensure your installation has a speed in the range 60{range}80 Megabytes per second.
```
ekohl, line 14: I think this is outdated now. Modern storage is easily able to do 100x this (current fastest NVME single drivers are in the 10 GB/s+ range so getting to the 200x). I think the high-bandwidth and low latency should cover it already
```suggestion
```

maximiliankolb (approved):
Verified directory path on Foreman 3.11/Katello 4.13. LGTM.

I am OK dropping this sentence as suggested by Ewoud but also OK to merge as is.

guides/common/modules/ref_storage-guidelines.adoc
```diff
@@ -9,9 +9,8 @@ This is a requirement for the `puppetserver` service to work.
 
 * Because most {ProductName} data is stored in the `/var` directory, mounting `/var` on LVM storage can help the system to scale.
 
-* Use high-bandwidth, low-latency storage for the `/var/lib/pulp/` directories.
+* Use high-bandwidth, low-latency storage for the `/var/lib/pulp/` and PostgreSQL `/var/lib/pgsql/` directories.
```
ekohl, line 12: Thanks for @maximiliankolb for reminding me we need to check the path. There's actually an attribute for it that has a different value on Debian.
```suggestion
* Use high-bandwidth, low-latency storage for the `/var/lib/pulp/` and PostgreSQL `{postgresql-lib-dir}` directories.
```

ekohl (changes requested):

maximiliankolb (approved):
Retested, looks correct to me:

````
# tested on Foreman 3.11 with very little content
# du -ksh /var/lib/pgsql
186M	/var/lib/pgsql
# du -ksh /var/lib/pgsql/data/
186M	/var/lib/pgsql/data/
````

-> almost all storage is in "data", so I conclude that it's the proper attribute.

Thanks Eric and Ewoud.

ekohl (approved):

Issue #2667, opened 2024-01-11
Document compute resources and their provisioning capabilities


ekohl:
Every Foreman compute resource has a list of capabilities. 2 are special, and they're `build` and `image`. These can be found in the code.

So there's:
* Foreman core
  * [ec2](https://github.com/theforeman/foreman/blob/c10969136f90584dc5ead7e5e3a4a53c938728db/app/models/compute_resources/foreman/model/ec2.rb#L43-L45) image
  * [libvirt](https://github.com/theforeman/foreman/blob/c10969136f90584dc5ead7e5e3a4a53c938728db/app/models/compute_resources/foreman/model/libvirt.rb#L40-L42) build & image
  * [openstack](https://github.com/theforeman/foreman/blob/c10969136f90584dc5ead7e5e3a4a53c938728db/app/models/compute_resources/foreman/model/openstack.rb#L34-L36) image
  * [ovirt](https://github.com/theforeman/foreman/blob/c10969136f90584dc5ead7e5e3a4a53c938728db/app/models/compute_resources/foreman/model/ovirt.rb#L38-L40) build & image
  * [vmware](https://github.com/theforeman/foreman/blob/c10969136f90584dc5ead7e5e3a4a53c938728db/app/models/compute_resources/foreman/model/vmware.rb#L49-L51) build & image
* Plugins
  * [azure](https://github.com/theforeman/foreman_azure_rm/blob/e2357a7408bab67ef73b117a742bfb4f6fd69504/app/models/foreman_azure_rm/azure_rm.rb#L87-L89) image
  * [google](https://github.com/theforeman/foreman_google/blob/44dc7d20405ebd3c1500e91bdcbbd860a061c69d/app/models/foreman_google/gce.rb#L18-L20) image

Build is essentially the bare metal provisioning method (PXE, HTTPBoot, etc). Image means booting up an image and then customizing it (like with cloud-init).

There may be more compute resource plugins, but the method is always the same.

dgoetz:
Another Plugin is [Proxmox](https://github.com/theforeman/foreman_fog_proxmox/blob/master/app/models/foreman_fog_proxmox/proxmox.rb#L56-L58) using build and image.

PR #3217, opened 2024-08-15
Add diagram for provisioning method selection

Lennonka:
Will be included in the docs later.


ekohl (changes requested):
You can do network-based provisioning on VMs as well and that arrow is missing. See https://github.com/theforeman/foreman-documentation/issues/2667 for which compute resources support this.

maximiliankolb (commented):
According to downstream docs, you can also do network-based deployments for VMs: https://docs.orcharhino.com/or/docs/sources/guides/almalinux/provisioning_hosts/introduction.html#Provisioning_Methods

ekohl (commented):
The arrows to image based and network based overlap, which implies it doesn't matter. Is there a way to get those arrows to separate visually?

Lennonka:
> The arrows to image based and network based overlap, which implies it doesn't matter. Is there a way to get those arrows to separate visually?

@ekohl I can do that, but the message will still be the same. If it matters and the user has to make a decision, then what do they base their decision on?

Lennonka:
Feel free to suggest further improvements in a subsequent PR. Thank you!

PR #3222, opened 2024-08-16
Ext auth methods review

asteflova:

Adding the keywords "single sign on-on" and "two-factor authentication" to a few headings and introductions in assemblies related to external identity providers.

When it comes to the procedures to configure external authentication sources, we need to emphasize use cases (SSO, 2FA) over the implementation (FreeIPA, Keycloak). This is because we have users who are above all looking to implement a certain type of authentication; these users will start by searching for the keywords "SSO" or "2FA" and only then start thinking about whether to implement them through FreeIPA or Keycloak.

https://issues.redhat.com/browse/SAT-27101

I'm making no changes to the AD provider user story because for that, a more thorough review is coming.

We are also considering splitting off `assembly_configuring-external-authentication-and-enabling-single-sign-on-and-two-factor-authentication.adoc` into a separate guide to make it even more easier for users to find the content.

#### Checklists

* [ ] Foreman 3.6/Katello 4.8


guides/common/modules/con_configuring-external-authentication-and-enabling-single-sign-on-and-two-factor-authentication.adoc
```diff
@@ -1,8 +1,10 @@
 [id="Configuring_External_Authentication_{context}"]
```
maximiliankolb, line 1:
```suggestion
[id="configuring-external-authentication-and-enabling-single-sign-on-and-two-factor-authentication_{context}"]
```

maximiliankolb (commented):
one small suggestion, rest of the diff LGTM. Handing over to Avital as per ticket.

asteflova:
> one small suggestion, rest of the diff LGTM. Handing over to Avital as per ticket.

In the ticket, I'm asking Avital for guidance on splitting the assemblies off to a new guide. Her review is welcome for this PR too, obviously, but not strictly necessary. Which makes me wonder @maximiliankolb: Is a general "LGTM" comment like yours the same as a formal approval? (I've been wondering about this for a while now and this seems like a good opportunity to clarify the issue.)

I think it's beneficial to require a formal ack on a PR before merging, considering that this is a project where maintainers regularly merge their own PRs. In my opinion, it helps ensure a certain level of accountability. However, I'm not sure if there's a specific convention for this in our repo. What do you think?

maximiliankolb:
> > one small suggestion, rest of the diff LGTM. Handing over to Avital as per ticket.
> 
> In the ticket, I'm asking Avital for guidance on splitting the assemblies off to a new guide. Her review is welcome for this PR too, obviously, but not strictly necessary. Which makes me wonder @maximiliankolb: Is a general "LGTM" comment like yours the same as a formal approval? (I've been wondering about this for a while now and this seems like a good opportunity to clarify the issue.)

If I write "LGTM" as a comment, then it's a "LGTM but I am not 100% sure. If you're confident, please merge as is.". I will reflect about the way I do upstream reviews.

> I think it's beneficial to require a formal ack on a PR before merging, considering that this is a project where maintainers regularly merge their own PRs. In my opinion, it helps ensure a certain level of accountability. However, I'm not sure if there's a specific convention for this in our repo. What do you think?

I only commented on the diff which I deemed good. I did not test this myself and I saw that you've talked with Avital in the ticket and requested a review from Marek. I wanted to express :+1: to both Avital and Marek to make it easier for them.

asteflova:
Thanks a lot for explaining!

> I did not test this myself

Now this is interesting. I would not even expect a fellow writer to test any changes. I was under the impression that a writer's peer review is to check for style issues and Dev/QE review covers the technical part. It's also how I've been using the review labels: `needs style review`/`style review done` and `needs tech review`/`tech review done`.

apinnick:
I think it makes sense to move the content to a separate guide in the future.

ares (approved):
I think this is going towards the right direction. Couple of observations:

1) Wildfly based keycloak wasn't updated with wording, while it also provides SSO and 2FA, was this intentional?
2) While this adds information to existing headers, it does not change the structure of the guide much. After reading the whole text, I guess the challenge here is, that various method provides only parts of the functionality. So I wonder, would it help to build some sort of a table/matrix for which providers offer SSO, 2FA, and external IdP? If that's the first thing mentioned in the "[5. Configuring external authentication and enabling single sign-on and two-factor authentication](https://theforeman-foreman-documentation-preview-pr-3222.surge.sh/nightly/Installing_Server/index-katello.html#configuring-external-authentication-and-enabling-single-sign-on-and-two-factor-authentication_foreman)", it would likely inform the user, which sections they should read from chapter 5.

None from the above blocks the PR as these are good improvements already.

asteflova:
Thanks for the feedback, Marek!

> 1. Wildfly based keycloak wasn't updated with wording, while it also provides SSO and 2FA, was this intentional?

With the Wildfly distribution in legacy mode now, I intentionally left the whole chapter out. I don't want to draw any more attention to it.

> 
> 2. While this adds information to existing headers, it does not change the structure of the guide much. After reading the whole text, I guess the challenge here is, that various method provides only parts of the functionality. So I wonder, would it help to build some sort of a table/matrix for which providers offer SSO, 2FA, and external IdP? If that's the first thing mentioned in the "[5. Configuring external authentication and enabling single sign-on and two-factor authentication](https://theforeman-foreman-documentation-preview-pr-3222.surge.sh/nightly/Installing_Server/index-katello.html#configuring-external-authentication-and-enabling-single-sign-on-and-two-factor-authentication_foreman)", it would likely inform the user, which sections they should read from chapter 5.

The trouble with overview sections like the one you describe is that they duplicate information that is documented in, for example, actual procedures. They are also at risk of getting outdated (sometimes writers update only the actual procedures and forget to update the overviews). To some extent, that was the case of the external authentication overview section that is included in the older versions of the document: [5. Configuring external authentication](https://docs.theforeman.org/3.10/Installing_Server/index-katello.html#Configuring_External_Authentication_foreman). Summarizing features offered by different products is also a big maintenance risk: we shouldn't commit to maintaining up-to-date information about other product offerings.

A well-designed table of contents (ToC) along with informative introductory sections should meet the goal that you describe too, but without the additional maintenance burden: By looking at a list of chapters and sections, users should be able to get a reasonably good idea of what they can expect and what options they have. I'm not entirely opposed to adding a new section with a summary/table/matrix, but I want to make sure I explore the ToC-only solution properly first.

I agree that the current solution still leaves room for improvement. I expect things might get clearer once we have a separate guide on external authentication (which is my next step after this PR). A guide has a title, subtitle, abstract; all this metadata is also a good place to address some of the existing gaps. And if that proves to be insufficient, we can think of additional changes.

asteflova:
Rebased to allow merging.

> one small suggestion, rest of the diff LGTM.

This covers style review.

> None from the above blocks the PR as these are good improvements already.

This covers tech review.

asteflova:

44d39d689c..ea2fcfbe7b  3.12 -> 3.12
50d9c8c69d..c50e378436  3.11 -> 3.11

PR #3223, opened 2024-08-17
Correct Smart Proxy upgrade instructions on foreman-el

ekohl:

On foreman-el there is no PostgreSQL present so there is no DNF module switching.

Eventually users should be able to follow the upgrade guide and actually get an upgraded system.

Note there are no upgrade instructions for EL9 and the setting up repository step for foreman-deb is completely missing.

It's also interesting that it upgrades the OS packages first and then the Foreman packages. That feels redundant to me.

@evgeni any thoughts on recommending foreman-maintain like we do in Satellite? That would help abstract away many of the differences and align on process.

#### Checklists

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_upgrading-smartproxy-server.adoc
+++ b/guides/common/modules/proc_upgrading-smartproxy-server.adoc
@@ -44,30 +44,31 @@ For information on backups, see {AdministeringDocURL}backing-up-{project-context
 ifdef::foreman-el,katello[]
 . Update repositories:
 +
+ifdef::katello[]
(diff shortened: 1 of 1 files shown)


evgeni:
Well, that's sadly not correct.
You need the foreman module for foreman-proxy.rpm, and that pulls the PostgreSQL Module, so you gotta switch.

I had to when upgrading puppet01.

ekohl:
Ah, you're right. I'm so glad we'll drop modularity soon when we drop EL8 in 3.13

PR #3224, opened 2024-08-19
Add more information about online and offline backup

AkshayGadhaveRH:
Currently the documentation does not talk about the usecases for the different types of backups. Users are bound to get confused about which procedure to use when backing up. Adding information about usecases for online and offline backups.

JIRA links:
- Backup documentation improvement: https://issues.redhat.com/browse/SAT-23077
- Online backups are for production: https://issues.redhat.com/browse/SAT-26312

#### Checklists

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_performing-a-full-backup.adoc
+++ b/guides/common/modules/proc_performing-a-full-backup.adoc
@@ -12,7 +12,7 @@ All {Project} services are shut down during an offline backup.
 * Online backup
 +
 With online backups, only services that affect the consistency of the backup are shut down while the backup process is running.
-For production usecases, use online backups. 
+Online backups require additional time to ensure the consistency of the backup. 
 
 For more information about each of these methods, you can view the usage statements for each backup method.
 
--- a/guides/common/modules/proc_performing-a-full-backup.adoc
+++ b/guides/common/modules/proc_performing-a-full-backup.adoc
@@ -8,8 +8,9 @@ There are two main methods of backing up {ProjectServer}:
 * Offline backup
 * Online backup
 +
-With online backups, the {Project} services are online while the backup process is running.
-The {Project} services are shut down during an offline backup, resulting in a more consistent backup.
+With online backups, only services that affect the consistency of the backup are shut down while the backup process is running.
+For production usecases, use online backups. 
+All {Project} services are shut down during an offline backup.
 For more information about each of these methods, you can view the usage statements for each backup method.
 
 .Offline backups
--- a/guides/common/modules/proc_performing-a-full-backup.adoc
+++ b/guides/common/modules/proc_performing-a-full-backup.adoc
@@ -8,6 +8,8 @@ There are two main methods of backing up {ProjectServer}:
 * Offline backup
 * Online backup
 +
+With online backups, the {Project} services are online while the backup process is running.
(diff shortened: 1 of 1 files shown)


AkshayGadhaveRH:
@evgeni could you check and let me know if it looks good?

guides/common/modules/proc_performing-a-full-backup.adoc
```diff
@@ -8,6 +8,9 @@ There are two main methods of backing up {ProjectServer}:
 * Offline backup
 * Online backup
 +
+With online backups, only services that affect the consistency of the backup are shut down while the backup process is running.
+For production usecases, use online backups.
```
evgeni, line 12: So, when should I use offline ones? ;-)

asteflova (changes requested):
Please @AkshayGadhaveRH add a link to the related Jira issue to the PR description.

You are adding information about both types of backup (online and offline) to the bullet point that describes one type (online). Please reconsider. The desired state is for each type of backup to be described under its corresponding label. In this case, those labels could be either the bullets on lines 8 and 9 or the discrete headings on lines 13 and 19.

guides/common/modules/proc_performing-a-full-backup.adoc
AkshayGadhaveRH: @evgeni should I remove this? I wasn't sure about this

evgeni: I would, yeah.

Both backup methods are fine for production usecases, that was the point of the changes we implemented.

The tradeoff for the user is:
- When doing an offline backup, all services are down and we can just assume the data copied is consistent (= we perform no checks)
- When doing an online backup, only some services (which we know to produce inconsistencies) are down. As there still can be something interfering with the consistency, we perform checks (and may restart data-copy operations) which take additional time.

evgeni (approved):
Ack on the message, I'll leave the exact wording to Aneta!

guides/common/modules/proc_performing-a-full-backup.adoc
```diff
@@ -6,8 +6,14 @@
 There are two main methods of backing up {ProjectServer}:
 
 * Offline backup
++
+All {Project} services are shut down during an offline backup.
+
 * Online backup
 +
+With online backups, only services that affect the consistency of the backup are shut down while the backup process is running.
```
ehelms, line 14: This is rather vague from a user's perspective. Do we have a place where we go into more detail about what is possible and what is not possible during online backup capability wise?

evgeni: We do not

AkshayGadhaveRH: We can create a concept module for this with more detailed description.

ehelms: Let's either do that as part of this or create a tracker to tackle it as a follow up.

evgeni:
Small not on the cherry picks: please don't pick it further than 3.12, older releases do not have the online-backup improvements this description relies on.

asteflova (approved):

evgeni:
   f818ad45f5..c9111f2173  3.12 -> 3.12

PR #3225, opened 2024-08-19
Add procedure to upgrade Foreman on Deb

maximiliankolb:
Content is based on Foreman Manual for 3.11:
https://theforeman.org/manuals/3.11/index.html#3.6Upgradeto3.11

Make "Upgrade Server" available for "Foreman on Debian/Ubuntu".

Before, it was only available in the Foreman Manual. Before we can make docs.theforeman.org official, we need to provide basic upgrade instructions for the Community.

Downsides: we now have two places where we define the set of operating systems for `foreman-deb` -> I have added comments to both files.

Follow up task: Add instructions to upgrade Smart Proxy Servers on Debian/Ubuntu

#### Checklists

### DIFF
--- a/guides/common/modules/con_upgrade-paths.adoc
+++ b/guides/common/modules/con_upgrade-paths.adoc
@@ -15,7 +15,9 @@ The high-level steps in upgrading {Project} to {ProjectVersion} are as follows:
 . Upgrade {ProjectServer} to {ProjectVersion}.
 For more information, see xref:Project_Upgrade_Considerations_{context}[].
 . Upgrade all {SmartProxyServers} to {ProjectVersion}.
+ifndef::foreman-deb[]
 For more information, see xref:upgrading_{smart-proxy-context}_server_{context}[].
+endif::[]
 
 ////
 ifdef::satellite[]
--- a/guides/common/modules/con_upgrade-paths.adoc
+++ b/guides/common/modules/con_upgrade-paths.adoc
@@ -18,12 +18,3 @@ For more information, see xref:Project_Upgrade_Considerations_{context}[].
 ifndef::foreman-deb[]
 For more information, see xref:upgrading_{smart-proxy-context}_server_{context}[].
 endif::[]
-
-////
-ifdef::satellite[]
-During an upgrade of {ProjectServer}, you must observe the correct upgrade path depending on your network environment:
-
-.Overview of {ProjectServer} upgrade paths in connected and disconnected network environments
-image::common/upgrade-paths-satellite.png[Overview of {ProjectServer} upgrade paths in connected and disconnected network environments]
-endif::[]
-////
--- a/guides/common/modules/proc_upgrading-a-connected-project-server.adoc
+++ b/guides/common/modules/proc_upgrading-a-connected-project-server.adoc
@@ -97,49 +97,9 @@ Replace _My_Distribution_Code_Name_ with the proper distribution code name based
 [options="nowrap" subs="attributes"]
 ----
 # apt-get update
-# apt-get --only-upgrade install ruby\* foreman\*
+# apt-get upgrade
 ----
-. Migrate the database:
-+
-[options="nowrap" subs="attributes"]
-----
-# foreman-rake db:migrate
-# foreman-rake db:seed
-----
-. Clear the cache and existing database sessions:
-+
-[options="nowrap" subs="attributes"]
-----
-# foreman-rake tmp:cache:clear
-# foreman-rake db:sessions:clear
-----
-. Optional: Reclaim space on your PostgreSQL database:
-+
-[options="nowrap" subs="attributes"]
-----
-# su - postgres -c "vacuumdb --full --dbname=foreman"
-----
-. If you have used `{foreman-installer}` to set up your existing {ProjectServer}, {Team} recommends running it again after upgrading.
-However, `{foreman-installer}` can modify config files so this might overwrite your custom changes.
-+
-* You can run the installer in noop mode to see what would be changed:
-+
-[options="nowrap" subs="+quotes,verbatim,attributes"]
-----
-# {foreman-installer} --noop --verbose
-----
-+
-You might see errors such as the following:
-+
-[options="nowrap" subs="+quotes,verbatim,attributes"]
-----
-/Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[{foreman-example-com}]: Could not evaluate: Connection refused - connect(2)
-----
-+
-These errors mean that the `apache2` service is being stopped.
-You can safely ignore them.
-+
-* After reviewing the changes the installer will make, run it again:
+. Run `{foreman-installer}`:
 +
 [options="nowrap" subs="+quotes,verbatim,attributes"]
 ----
--- a/guides/common/modules/proc_upgrading-a-connected-project-server.adoc
+++ b/guides/common/modules/proc_upgrading-a-connected-project-server.adoc
@@ -30,8 +30,9 @@ include::snip_warning-maintain-config-noop.adoc[]
 +
 [options="nowrap" subs="attributes"]
 ----
-# {foreman-installer} --foreman-proxy-dns-managed=false \
---foreman-proxy-dhcp-managed=false
+# {foreman-installer} \
+--foreman-proxy-dhcp-managed=false \
+--foreman-proxy-dns-managed=false
 ----
 . In the {ProjectWebUI}, navigate to *Hosts* > *Discovered hosts*.
 On the Discovered Hosts page, power off and then delete the discovered hosts.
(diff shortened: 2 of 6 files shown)


guides/common/modules/con_upgrade-paths.adoc
```diff
@@ -14,8 +14,10 @@ The high-level steps in upgrading {Project} to {ProjectVersion} are as follows:
 
 . Upgrade {ProjectServer} to {ProjectVersion}.
 For more information, see xref:Project_Upgrade_Considerations_{context}[].
+ifndef::foreman-deb[]
```
maximiliankolb, line 17: alternative is to flip line 17 and 18.

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
```diff
@@ -76,14 +77,76 @@ ifdef::katello[]
...
+Replace _My_Distribution_Codename_ with "bullseye" if you run {ProjectServer} on Debian 11, "jammy" for Ubuntu 22.04, and "focal" for Ubuntu 20.04.
+// List of supported OS for foreman-deb has to match "guides/common/modules/proc_configuring-repositories-foreman-deb.adoc"
+. Update to {Project} {ProjectVersion}:
++
+[options="nowrap" subs="attributes"]
+----
+# apt-get update
+# apt-get --only-upgrade install ruby\* foreman\*
+----
+. Migrate the database on {ProjectServer}:
```
maximiliankolb, line 98: Should I drop "on {ProjectServer}" everywhere because it's obvious?

guides/common/modules/con_upgrade-paths.adoc
asteflova: For Debian, this only shows one item in the list. Which is not great and flipping 17 and 18 could help. Have considered, as an alternative, simply dropping the "High-level upgrade steps" list for all builds? The section is named "Upgrade paths", so I would say that an overview of the upgrade does not belong there anyway. Additionally, from looking at the table of contents and considering the overall flow of the document, I think users should be able to follow the upgrade steps just fine without this summary.

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
asteflova: All the steps should be performed on {ProjectServer}, is that correct? You could just write it once at the very beginning:

```
.Upgrade {ProjectServer}
On your {ProjectServer}:

<step 1>
```

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
```diff
@@ -76,14 +77,76 @@ ifdef::katello[]
...
+. Update repositories in `/etc/apt/sources.list.d/foreman.list` on your {ProjectServer} to consume content for {Project} {ProjectVersion}:
++
+[options="nowrap" subs="attributes"]
+----
+deb http://deb.theforeman.org/ _My_Distribution_Codename_ {ProjectVersion}
+deb http://deb.theforeman.org/ plugins {ProjectVersion}
+----
++
+// List of supported OS for foreman-deb has to match "guides/common/modules/proc_configuring-repositories-foreman-deb.adoc"
+Replace _My_Distribution_Codename_ with `bullseye` if you run {ProjectServer} on Debian 11, `jammy` for Ubuntu 22.04, and `focal` for Ubuntu 20.04.
```
asteflova, line 90: I'd like to see all the items in the list use the same structure (parallelism). It tripped me up the first time I read this. How about this:

```suggestion
Replace _My_Distribution_Code_Name_ with the proper distribution code name based on where your run {ProjectServer}:

* `bullseye` for Debian 11
* `jammy` for Ubuntu 22.04
* `focal` for Ubuntu 20.04
```

Not sure if you really need the list.

Also, "code name" based on [Merriam Webster](https://www.merriam-webster.com/dictionary/code%20name).

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
```diff
@@ -76,14 +77,76 @@ ifdef::katello[]
...
+# foreman-rake db:sessions:clear
+----
+. Optional: Reclaim space on your PostgreSQL database on {ProjectServer}:
++
+[options="nowrap" subs="attributes"]
+----
+# su - postgres -c "vacuumdb --full --dbname=foreman"
+----
+. If you have used `{foreman-installer}` to set up your existing {ProjectServer}, {Team} recommends running it again after upgrading.
+* `{foreman-installer}` can modify config files so this may overwrite your custom changes.
```
asteflova, line 119:
```suggestion
. If you have used `{foreman-installer}` to set up your existing {ProjectServer}, {Team} recommends running it again after upgrading.
However, `{foreman-installer}` can modify config files so this might overwrite your custom changes.
```

Technically, I think line 119 is more tightly related to line 118 than 120. Which is why I'm suggesting this.

You could even consider using an admonition. I know that they are frowned upon in this team :) but they do exist for a reason. Your call.

```suggestion
. If you have used `{foreman-installer}` to set up your existing {ProjectServer}, {Team} recommends running it again after upgrading.
+
[IMPORTANT]
====
`{foreman-installer}` can modify config files so this might overwrite your custom changes.
====
```

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
```diff
@@ -76,14 +77,76 @@ ifdef::katello[]
...
+You can run the installer in noop mode so you can see what would be changed:
```
asteflova, line 120:
```suggestion
You can run the installer in noop mode to see what would be changed:
```

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
```diff
@@ -76,14 +77,76 @@ ifdef::katello[]
...
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+# {foreman-installer} --noop --verbose
+----
++
+You may see errors such as `/Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[{foreman-example-com}]: Could not evaluate: Connection refused - connect(2)` due to apache2 service being stopped.
+These can be safely ignored.
```
asteflova, line 128: Just a suggestion to make this easier to parse.

```suggestion
You might see errors such as the following:
+
----
/Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[{foreman-example-com}]: Could not evaluate: Connection refused - connect(2)
 ----
 +
These errors mean that the `apache2` service is being stopped.
You can safely ignore them.
```

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
```diff
@@ -76,14 +77,76 @@ ifdef::katello[]
...
+* To apply these changes, run the installer again without options:
```
asteflova, line 129: I don't think "applying changes" is an apt enough description for what we're asking users to do. How about this?

```suggestion
* After reviewing the changes the installer will make, run it again:
```

guides/common/modules/con_upgrade-paths.adoc
maximiliankolb: I have flipped 17 and 18. I think the concept would be almost empty without those two steps. I am OK with the following solutions for my PR:

* keep as is
* drop 11-end
* drop the whole file because without line 11-end, the file does not really show any "upgrade paths" as the title suggests. I would then move line 7+8 for all build targets to another file because we should not really recommend users to skip versions during upgrades.

What do you think is best?

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
maximiliankolb: Applied.

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
maximiliankolb: Dropped "on {ProjectServer}"; the procedure is named ".Upgrade {ProjectServer}".

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
maximiliankolb: Applied your first suggestion without admonition.

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
maximiliankolb: Applied.

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
maximiliankolb: Applied.

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
maximiliankolb: Applied.

maximiliankolb (commented):
Applied all feedback by Anet. Ready for re-review.

guides/common/modules/con_upgrade-paths.adoc
asteflova: I have a slight preference towards dropping the file, but  that option would also require the most effort from you. Considering that the goal is to make progress with the migration from the Manual, "keep as is" will work just fine too :)

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
```diff
@@ -76,14 +77,89 @@ ifdef::katello[]
...
+* `jammy` for Ubuntu 22.04
+* `focal` for Ubuntu 20.04
+. Update to {Project} {ProjectVersion}:
++
+[options="nowrap" subs="attributes"]
+----
+# apt-get update
+# apt-get --only-upgrade install ruby\* foreman\*
+----
+. Migrate the database:
```
evgeni, line 102: I would prefer to take this opportunity and drop all non-installer setups from the docs.
Then the upgrade becomes:
1. stop services
2. apt update/apt upgrade
3. foreman-installer

evgeni: (we do so in EL-docs already.

maximiliankolb (commented):
Rebased to "master" and applied the suggestion to drop all non-foreman-installer steps. Kindly asking for a re-review @evgeni

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
```diff
@@ -76,6 +77,41 @@ ifdef::katello[]
...
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+# {foreman-installer}
+----
+. Restart all {Project} services:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+# {foreman-maintain} service restart
```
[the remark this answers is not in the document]
maximiliankolb: Removed this code block; added a command to stop all services before performing the package upgrade.

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
```diff
@@ -76,6 +77,41 @@ ifdef::katello[]
...
+----
+# {foreman-maintain} service stop
+----
+. Update to {Project} {ProjectVersion}:
++
+[options="nowrap" subs="attributes"]
+----
+# apt-get update
+# apt-get upgrade
+----
```
evgeni, line 107: The flow here is a bit weird:
* change repo config (apt)
* stop services (foreman)
* update packages (apt)

While technically correct (changing repo config doesn't further alter the system), I'd still go with:
* stop services (foreman)
* change repo config (apt)
* update packages (apt)

As it creates less context switches.

maximiliankolb: Makes sense, fixed.

evgeni (approved):
I don't like the fact that we duplicate the OS list, but this makes the procedure much more actionable, so I'd err on the side of including it.

maximiliankolb:
   c9111f2173..98bb3e662e  3.12 -> 3.12
   cc9e654c90..f5617e7147  3.11 -> 3.11

ekohl:
Thanks @maximiliankolb for getting this done!

PR #3233, opened 2024-08-21
Cherrypick missed commits into 3.12 master

ColeHiggins2:
A few commits were missed during 3.12 branching. This PR is a cherrypick of those commits into 3.12 master


ekohl (approved):
@asteflova since you noticed it, I think this addresses your concerns. Mind verifying?

asteflova (approved):

PR #3239, opened 2024-08-22
Add organization to export repo without error

bangelic:
The command for exporting a repository resulted in 
error until a flag
for organization was added. The organization's name has 
been added to
the command to fix the error. See https://issues.redhat.com/browse/SAT-26583.

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_exporting-a-repository.adoc
+++ b/guides/common/modules/proc_exporting-a-repository.adoc
@@ -34,14 +34,14 @@ For more information, see xref:Download_Policies_Overview_{context}[].
 * Ensure that you synchronize products that you export to the required date.
 
 .Procedure
-. Export a repository using the product name and repository name:
+. Export a repository using the product, repository, and organization names:
 +
 [options="nowrap" subs="+quotes"]
 ----
 # hammer content-export complete repository \
---name="_My_Repository_" \
---organization="_My_Organization" \
---product="_My_Product_"
+--name="My_Repository" \
+--product="My_Product"  \
+--organization="My_Organization"
 ----
 +
 [NOTE]
--- a/guides/common/modules/proc_exporting-a-repository.adoc
+++ b/guides/common/modules/proc_exporting-a-repository.adoc
@@ -39,9 +39,9 @@ For more information, see xref:Download_Policies_Overview_{context}[].
 [options="nowrap" subs="+quotes"]
 ----
 # hammer content-export complete repository \
---name="My_Repository" \
---product="My_Product"  \
---organization="My_Organization"
+--name="_My_Repository_" \
+--product="_My_Product_"  \
+--organization="_My_Organization_"
 ----
 +
 [NOTE]
--- a/guides/common/modules/proc_exporting-a-repository.adoc
+++ b/guides/common/modules/proc_exporting-a-repository.adoc
@@ -34,7 +34,7 @@ For more information, see xref:Download_Policies_Overview_{context}[].
 * Ensure that you synchronize products that you export to the required date.
 
 .Procedure
-. Export a repository using the product, repository, and organization names:
+. Export a repository:
 +
 [options="nowrap" subs="+quotes"]
 ----
--- a/guides/doc-Installing_Server_Disconnected/master.adoc
+++ b/guides/doc-Installing_Server_Disconnected/master.adoc
@@ -26,10 +26,6 @@ include::common/modules/proc_configuring-server-to-consume-content-from-a-custom
 
 include::common/assembly_configuring-inter-server-synchronization.adoc[leveloffset=+2]
 
-include::common/assembly_importing-kickstart-repositories.adoc[leveloffset=+2]
-
-include::common/assembly_importing-the-project-client-name-repository.adoc[leveloffset=+2]
-
 include::common/modules/proc_configuring-pull-based-transport-for-remote-execution.adoc[leveloffset=+2]
 
 include::common/modules/proc_enabling-power-management-on-hosts.adoc[leveloffset=+2]
--- a/guides/doc-Installing_Server_Disconnected/master.adoc
+++ b/guides/doc-Installing_Server_Disconnected/master.adoc
@@ -26,6 +26,10 @@ include::common/modules/proc_configuring-server-to-consume-content-from-a-custom
 
 include::common/assembly_configuring-inter-server-synchronization.adoc[leveloffset=+2]
 
+include::common/assembly_importing-kickstart-repositories.adoc[leveloffset=+2]
+
+include::common/assembly_importing-the-project-client-name-repository.adoc[leveloffset=+2]
+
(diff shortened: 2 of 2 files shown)


maximiliankolb (changes requested):
I cannot reproduce this/don't know what this change fixes. Tested on Foreman 3.11:

````
# hammer content-export complete repository --name="AlmaLinux Client 8" --organization="Example" --product="AlmaLinux Client"
[.....................................................................................................................................................................................................] [100%]
Generated /var/lib/pulp/exports/Example/Export-AlmaLinux_Client_8-23/1.0/2024-08-23T15-33-03-02-00/metadata.json
# echo $?
0
# mv -i /var/lib/pulp/exports/Example/Export-AlmaLinux_Client_8-23/ /var/lib/pulp/exports/Example/Export-AlmaLinux_Client_8-23_test1

# hammer content-export complete repository --name="AlmaLinux Client 8" --product="AlmaLinux Client" --organization="Example"
[.....................................................................................................................................................................................................] [100%]
Generated /var/lib/pulp/exports/Example/Export-AlmaLinux_Client_8-23/2.0/2024-08-23T15-35-11-02-00/metadata.json
# echo $?
0

# cat /var/lib/pulp/exports/Example/Export-AlmaLinux_Client_8-23_test1/1.0/2024-08-23T15-33-03-02-00/metadata.json | jq > out1.json
# cat /var/lib/pulp/exports/Example/Export-AlmaLinux_Client_8-23/2.0/2024-08-23T15-35-11-02-00/metadata.json | jq > out2.json
# diff out1.json out2.json 
````

````diff
42c42
<     "major": 1,
---
>     "major": 2,
67c67
<   "toc": "export-01917f70-cea0-7e11-9533-13e880eec9f8-20240823_1333-toc.json"
---
>   "toc": "export-01917f72-bf16-706a-9126-62b725778ecf-20240823_1335-toc.json"
````

To my understanding, this means that the exported repositories are identical irrespective of the order of the Hammer arguments.

**Comment by jeremylenz** [comment]
@maximiliankolb My understanding of the bug is that the documentation only has `--name` and `--product` and does not include `--organization`. So if you run it again without the `--organization` option you should reproduce the error.

File: guides/common/modules/proc_exporting-a-repository.adoc [hunk 1]
```diff
@@ -34,14 +34,14 @@ For more information, see xref:Download_Policies_Overview_{context}[].
 * Ensure that you synchronize products that you export to the required date.
 
 .Procedure
-. Export a repository using the product name and repository name:
+. Export a repository using the product, repository, and organization names:
```
asteflova, line 37: This is a valid change that ensures all options passed with the command below are listed. Alternatively, to prevent issues like this in the future, you could also consider something like this:

```suggestion
. Export a repository:
```

guides/common/modules/proc_exporting-a-repository.adoc
```diff
@@ -34,14 +34,14 @@ For more information, see xref:Download_Policies_Overview_{context}[].
...
 +
 [options="nowrap" subs="+quotes"]
 ----
 # hammer content-export complete repository \
---name="_My_Repository_" \
---organization="_My_Organization" \
---product="_My_Product_"
+--name="My_Repository" \
+--product="My_Product"  \
+--organization="My_Organization"
```
asteflova, line 44: This does not address the problem reported in https://issues.redhat.com/browse/SAT-26583.

asteflova (commented):
https://issues.redhat.com/browse/SAT-26583 reports the error for Satellite 6.14. Looking into the commit history for the file in question, https://github.com/theforeman/foreman-documentation/commit/ddd2ab3d40db6e91538b399c837ec0661492e9e6 added `--organization` to the command but was cherry-picked only down to 3.9. Therefore, the problem now exists only on branches 3.8 and lower so to address the ticket, I recommend looking into those branches.

maximiliankolb:
Thanks @jeremylenz I do not have access to the SAT issue but I'll retest this after Brian has adjusted the PR.

guides/common/modules/proc_exporting-a-repository.adoc
bangelic: @asteflova can you clarify?

I was talking with @jeremylenz about this on Slack and he said from what he was reading, the issue was outlined in your comment. He said this PR is comparing it to the master branch and that it only shows I'm changing the order of the organization and product lines, since the `--organization="My_Organization"` is already there.

He says if this PR is cherry-picked into older branches as well (3.8?) it may result in the `--organization="My_Organization"` being added, rather than reordered.

asteflova: That's a good point @bangelic. The --organization flag is missing on branches 3.8 and lower; but adding it to master and cherry-picking all the way down will ensure that the procedure will be consistent across versions, which is a good thing.

I'd like you to change one tiny thing: It seems you have accidentally dropped the italics for replaceables on lines 42-44. Can you re-add them? Then your PR should be good to merge.

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -26,10 +26,6 @@ include::common/modules/proc_configuring-server-to-consume-content-from-a-custom
 
 include::common/assembly_configuring-inter-server-synchronization.adoc[leveloffset=+2]
 
-include::common/assembly_importing-kickstart-repositories.adoc[leveloffset=+2]
-
-include::common/assembly_importing-the-project-client-name-repository.adoc[leveloffset=+2]
-
```
asteflova, line 32: This wasn't here before, was it? Please drop this part of your PR.

asteflova (changes requested):
Ok, when reviewing the whole set of changes again, I noticed that something seems to have sneaked in. Please address that @bangelic.

asteflova:
> Thanks @jeremylenz I do not have access to the SAT issue but I'll retest this after Brian has adjusted the PR.

It's too bad https://issues.redhat.com/browse/SAT-26583 isn't public, I think tickets like that should be open. Please see https://github.com/theforeman/foreman-documentation/pull/3239#pullrequestreview-2273261458 for a summary: the `--organization` flag is missing on lower branches.

guides/doc-Installing_Server_Disconnected/master.adoc
bangelic: I think this is part of a different PR I'm doing. Not sure why this showed up.

asteflova (approved):

asteflova:

* 96fbc82b74..2853e588ff  3.12 -> 3.12
* b37775ae04..e2eee0a2f3  3.11 -> 3.11
* 72b7266614..841d2daa26  3.10 -> 3.10
* ce170b622c..c9f468ff05  3.9 -> 3.9
* 1a3c9d8197..fb0766362c  3.8 -> 3.8
* 2f15ee5050..f348417648  3.7 -> 3.7
* 71ec6b26d6..d2e9988baf  3.6 -> 3.6
* a260b697f8..b6c65b1fc3  3.5 -> 3.5

PR #3242, opened 2024-08-26
Configure host for global registration

Lennonka:

A host needs to be configured a certain way before it can be registered by using Global Registration and that isn't properly documented at the present time.

We have a _Configure a host for registration_ module, but we don't use it in the _Managing Hosts_ guide currently, which we should. Originally, it covered only the NTP synchronization.

Also, we should provide a prerequisite so that the user can make a secured curl call right away. The CA certificate config is now also included in the pre-registration configuration.

Because it's better to tell the user how the host must be configured for registration beforehand instead of explaining it when it's "too late".

Depends on #3193

#### Checklists

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/assembly_registering-hosts.adoc
+++ b/guides/common/assembly_registering-hosts.adoc
@@ -11,6 +11,8 @@ include::modules/con_registering-hosts-by-using-global-registration.adoc[levelof
 
 include::modules/ref_global-parameters-for-registration.adoc[leveloffset=+2]
 
+include::modules/proc_configuring-a-host-for-registration.adoc[leveloffset=+2]
+
 include::modules/proc_registering-a-host.adoc[leveloffset=+2]
 
 include::modules/proc_customizing-host-registration-by-using-snippets.adoc[leveloffset=+2]
--- a/guides/common/modules/proc_configuring-a-host-for-registration.adoc
+++ b/guides/common/modules/proc_configuring-a-host-for-registration.adoc
@@ -2,24 +2,79 @@
 = Configuring a host for registration
 
 Configure your host for registration to {ProjectServer} or {SmartProxyServer}.
+You can use a configuration management tool to configure multiple hosts at once.
 
 .Prerequisites
 * The host must be using a supported operating system.
 For more information, see xref:supported-clients-in-registration_{context}[].
+ifdef::satellite,orcharhino,katello[]
+* {ProjectServer} and any {SmartProxyServers} must have a time-synchronization tool enabled and running and be synchronized with the same NTP server.
+For more information, see the following resources:
+** link:{InstallingServerDocURL}synchronizing-the-system-clock-with-chronyd_{project-context}[Synchronizing the system clock with chronyd] in _{InstallingServerDocTitle}_
+ifdef::satellite[]
+** link:{InstallingServerDisconnectedDocURL}synchronizing-the-system-clock-with-chronyd_{project-context}[Synchronizing the system clock with chronyd] in _{InstallingServerDisconnectedDocTitle}_
+endif::[]
+** link:{InstallingSmartProxyDocURL}synchronizing-the-system-clock-with-chronyd_{smart-proxy-context}[Synchronizing the system clock with chronyd] in _{InstallingSmartProxyDocTitle}_
+endif::[]
 
 .Procedure
-* Ensure that a time-synchronization tool is enabled and running on the host.
+ifdef::satellite,orcharhino,katello[]
+. Enable and start a time-synchronization tool on your host.
+The host must be synchronized with the same NTP server as {ProjectServer} and any {SmartProxyServers}.
 ifdef::client-content-dnf[]
-** For {EL} 7 and newer:
+** On {EL} 7 and later:
 +
 ----
 # systemctl enable --now chronyd
 ----
-** For {EL} 6:
+** On {EL} 6:
 +
 ----
 # chkconfig --add ntpd
 # chkconfig ntpd on
 # service ntpd start
+----
+endif::[]
+endif::[]
+. Deploy the SSL CA file on your host so that the host can make a secured registration call.
+.. Find where {Project} stores the SSL CA file by navigating to *Administer* > *Settings* > *Authentication* and locating the value of the *SSL CA file* setting.
+.. Transfer the SSL CA file to your host securely, for example by using `scp`.
+.. Login to your host by using SSH.
+.. Copy the certificate to the truststore:
+ifdef::client-content-dnf[]
+ifndef::satellite[]
+* On {EL}:
+endif::[]
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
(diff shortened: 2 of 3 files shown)


Lennonka:
Pushed fix - missing `endif`.

guides/common/assembly_registering-hosts.adoc
```diff
@@ -11,6 +11,10 @@ include::modules/con_registering-hosts-by-using-global-registration.adoc[levelof
 
 include::modules/ref_global-parameters-for-registration.adoc[leveloffset=+2]
 
+ifdef::katello,orcharhino,satellite[]
+include::modules/proc_configuring-a-host-for-registration.adoc[leveloffset=+2]
+endif::[]
```
Lennonka, line 16: I'm including this procedure for non-Katello products only because time synchronization wasn't required and I was told that vanilla Foreman doesn't care about the certificate.

Lennonka:
Open to suggestions :)

guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -58,40 +58,14 @@ A {SmartProxy} behind a load balancer takes precedence over a {SmartProxy} selec
...
-+
-To do this, find where {Project} stores the CA file by navigating to *Administer* > *Settings* > *Authentication* and locating the value of the *SSL CA file* setting.
-+
-Copy the CA file to the `/etc/pki/ca-trust/source/anchors/` directory on hosts and enter the following commands:
-+
-[options="nowrap", subs="+quotes,attributes"]
-----
-# update-ca-trust enable
-# update-ca-trust
-----
```
Lennonka, line 80: My main point was to move this into an action that has to be done as a prerequisite to registration so that the user can make a secure call right away.

guides/common/assembly_registering-hosts.adoc
stejskalleos: > for non-Katello products

Isn't it the opposite, `ifdef::katello,orcharhino,satellite[]` is for katello, right?

Lennonka: @stejskalleos That's correct.

guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -4,10 +4,10 @@
 You can register a host by using registration templates and set up various integration features and host tools during the registration process.
 
 .Prerequisites
-* Your user account has a role assigned that has the `create_hosts` permission.
+* Your user account has a role assigned that grants the `create_hosts` permission.
 * You must have root privileges on the host that you want to register.
 ifdef::satellite,orcharhino,katello[]
-* {ProjectServer}, any {SmartProxyServers}, and all hosts must be synchronized with the same NTP server, and have a time synchronization tool enabled and running.
+* You have configured the host for registration.
```
stejskalleos, line 10: Can we make it as a link to the `guides/common/modules/proc_configuring-a-host-for-registration.adoc` section

guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -58,40 +58,14 @@ A {SmartProxy} behind a load balancer takes precedence over a {SmartProxy} selec
 ifdef::katello,satellite,orcharhino[]
 . In the *Activation Keys* field, enter one or more activation keys to assign to hosts.
 endif::[]
-. Optional: Select the *Insecure* option, if you want to make the first call insecure.
-During this first call, hosts download the CA file from {Project}.
+. Optional: Select the *Insecure* option, if you want to make the registration call insecure.
```
stejskalleos, line 61: I would rephrase it:
It's not like users want to make it insecure. It's because the CA file is not in the host's "trusted store."
So I would say something like, "If the CA is not trusted by the host, use insecure for the initial call. Then, the registration will download the CA automatically and use a secure connection for the next calls and communication."

Lennonka: @stejskalleos That's exactly the situation I'm trying to prevent by having the user to configure the host before registration. The pre-registration config involves adding the CA cert to the trust store. Are there any other use cases for an insecure call?

maximiliankolb: If users SSH into a host to run the curl/wget command, then they'd also have a way to securely deploy the Foreman CA cert. With guides/common/modules/proc_configuring-a-host-for-registration.adoc, which I like, there should not be a valid use case for `Insecure` anymore.

I also think the explanation by Leos makes sense; it's IMHO better than "click insecure to make X insecure" -> "Select insecure if your host does not trust the Foreman CA prior to the registration".

guides/common/modules/proc_registering-a-host.adoc
maximiliankolb: As long as Foreman Web UI offers the Insecure checkbox, we should keep the procedure on how to import the Foreman CA cert. Instead of deleting it, maybe move it to a separate procedure which we could specify as prerequisite?

maximiliankolb (commented):
some suggestions;

guides/common/modules/proc_configuring-a-host-for-registration.adoc
```diff
@@ -2,24 +2,42 @@
...
 # service ntpd start
 ----
 endif::[]
+. If you use a self-signed {Project} CA certificate, deploy the CA certificate on your host so that the host can make a secured registration call.
+ifdef::managing-hosts[]
+For more information, see xref:deploying-the-ca-certificate-on-a-host-manually[].
+endif::[]
+ifndef::managing-hosts[]
+{ManagingHostsDocURL}deploying-the-ca-certificate-on-a-host-manually[Deploying the CA certificate ona  host manually] in _{ManagingHostsDocTitle}_.
+endif::[]
```
Lennonka, line 43: @maximiliankolb FYI, this is the step to configure the CA cert. Since there's a procedure how to do that I'm just linking to it instead of duplicating the content.

Lennonka: On second thought, I think I'm mixing up two different certs. The original procedure talks about locating the **SSL CA file** setting, which refers to a different cert.

Lennonka:
I have rewritten the configuring procedure because earlier I was referring users to an incorrect certificate file.

I have also successfully tested the configuration + secure registration on RHEL 9. :heavy_check_mark:

maximiliankolb (approved):
LGTM style-wise.

Lennonka:
I'm setting the `tech review done` flag because I've tested it successfully. ^^

Lennonka:
Rebased.

Lennonka:
Cherry-picked:
- cd06a1e2b0..de15ce94e8  3.12 -> 3.12
- bc2ad19748..34db4e3ad0  3.11 -> 3.11
- 3d4cead6da..398118deca  3.10 -> 3.10
- caca9c03b6..530deeb3fd  3.9 -> 3.9

There were conflicts on 3.8, therefore I'm not cherrypicking it further.

PR #3243, opened 2024-08-28
Manually set global Git configuration in proc to sync templates with an existing repo

asteflova:
Attempting to update the Git global config for the `foreman` user fails due to insufficient permissions. Adding steps to manually create the config file as well as set the appropriate permissions on the file ensures that it gets updated as expected.

https://issues.redhat.com/browse/SAT-27104

The proposed solution might be considered hacky by some but it looks like [SAT-27349](https://issues.redhat.com/browse/SAT-27349) should make the whole section obsolete in a future version.

#### Checklists

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_synchronizing-templates-with-an-existing-repository.adoc
+++ b/guides/common/modules/proc_synchronizing-templates-with-an-existing-repository.adoc
@@ -5,6 +5,25 @@ Use this procedure to synchronize templates between your {ProjectServer} and an 
 
 .Procedure
 . If you want to use HTTPS to connect to the repository and you use a self-signed certificate authentication on your Git server:
+.. Create a new directory under the `/usr/share/foreman/` directory to store the Git configuration for the certificate:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+# mkdir -p /usr/share/foreman/.config/git
+----
+.. Create a file named `config` in the new directory:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+# touch /usr/share/foreman/.config/git/config
+----
+.. Allow the `foreman` user access to the `.config` directory:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+# chown -R foreman /usr/share/foreman/.config
+----
++
 .. Validate the certificate:
 +
 [options="nowrap" subs="+quotes,verbatim,attributes"]
--- a/guides/common/modules/proc_synchronizing-templates-with-an-existing-repository.adoc
+++ b/guides/common/modules/proc_synchronizing-templates-with-an-existing-repository.adoc
@@ -23,7 +23,6 @@ Use this procedure to synchronize templates between your {ProjectServer} and an 
 ----
 # chown --recursive foreman /usr/share/foreman/.config
 ----
-+
 .. Update the Git global configuration for the `foreman` user with the path to your self-signed CA certificate:
 +
 [options="nowrap" subs="+quotes,verbatim,attributes"]
--- a/guides/common/modules/proc_synchronizing-templates-with-an-existing-repository.adoc
+++ b/guides/common/modules/proc_synchronizing-templates-with-an-existing-repository.adoc
@@ -9,7 +9,7 @@ Use this procedure to synchronize templates between your {ProjectServer} and an 
 +
 [options="nowrap" subs="+quotes,verbatim,attributes"]
 ----
-# mkdir -p /usr/share/foreman/.config/git
+# mkdir --parents /usr/share/foreman/.config/git
 ----
 .. Create a file named `config` in the new directory:
 +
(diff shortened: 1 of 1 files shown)


asteflova:
@adamruzicka Can you please review?

guides/common/modules/proc_synchronizing-templates-with-an-existing-repository.adoc
```diff
@@ -4,13 +4,33 @@
...
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+# touch /usr/share/foreman/.config/git/config
+----
+.. Allow the `foreman` user access to the `.config` directory:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+# chown -R foreman /usr/share/foreman/.config
```
ekohl, line 24: I always prefer long options to make them self documenting
```suggestion
# chown --recursive foreman /usr/share/foreman/.config
```

guides/common/modules/proc_synchronizing-templates-with-an-existing-repository.adoc
```diff
@@ -4,13 +4,33 @@
...
 ----
-# sudo -u foreman git config --global http.sslCAPath _Path_To_My_Certificate_
+# mkdir -p /usr/share/foreman/.config/git
 ----
-. If you want to use SSH to connect to the repository, perform the following steps:
+.. Create a file named `config` in the new directory:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+# touch /usr/share/foreman/.config/git/config
```
ekohl, line 18: In my own home directory I see it has mode 0600 and I think in some cases git is unhappy with a world readable file, but I'm not entirely sure if you also need to chmod.

ekohl (commented):
I wonder if it makes sense to ship an empty config file with the correct permissions by default so I opened https://github.com/theforeman/foreman-packaging/pull/11180. There's some room for discussion of that implementation and we don't have it available everywhere immediately so it doesn't block merging this.

ekohl:
> A little bit of context: we're planning some changes to foreman_templates so that the http proxy configured in foreman's settings would be implicitly used, without the user having to go faff around with git config so the whole procedure along with the changes proposed here may become irrelevant.

The HTTP proxy configuration still doesn't set the trusted CA certificates. If you have a self signed certificate for your private gitlab then that doesn't help you, right? Or should we recommend setting up the trusted certificate system wide and avoid the specific git configuration?

guides/common/modules/proc_synchronizing-templates-with-an-existing-repository.adoc
asteflova: 0600 does sound like a good idea and looking at [16.2. Using repository sources](https://theforeman-foreman-documentation-preview-pr-3243.surge.sh/nightly/Administering_Project/index-satellite.html#using-repository-sources_admin) as a whole, I don't think it breaks anything further down the line..? I feel like we're already asking users to jump through too many hoops here but on the other hand, this could be a security concern.

@adamruzicka Would you be okay with my adding `# chmod 0600 /usr/share/foreman/.config/git/config`?

guides/common/modules/proc_synchronizing-templates-with-an-existing-repository.adoc
asteflova: Makes sense, let's do it for the whole procedure then.

asteflova:
Agreed, fixing the code would be preferable than making users do the legwork of manual configuration, but considering the context (explained in the PR description and by Adam above), my recommendation was to fix this in the docs.

@ekohl Since you opened the foreman-packaging PR (thanks for that, by the way!), can you please help me figure out which branches should the doc fix go to? Should I cherry-pick only to 3.12 and below because for 3.13 and above, the file will be shipped by default?

guides/common/modules/proc_synchronizing-templates-with-an-existing-repository.adoc
adamruzicka: Yes, fine by me

adamruzicka:
> The HTTP proxy configuration still doesn't set the trusted CA certificates. If you have a self signed certificate for your private gitlab then that doesn't help you, right? Or should we recommend setting up the trusted certificate system wide and avoid the specific git configuration?

I wanted to keep this brief, but since you're asking about it, I'll elaborate. The full plan is to offer an option to pick a proxy from a list of available proxies in the template sync form. The list will be populated with the default proxy as set in settings and all the currently visible proxies configured in infrastructure > http proxies, but will default to the one in settings. For the ones in infrastructure > http proxies, you can paste in a CA cert in case you need it.

guides/common/modules/proc_synchronizing-templates-with-an-existing-repository.adoc
ekohl: > 0600 does sound like a good idea and looking at [16.2. Using repository sources](https://theforeman-foreman-documentation-preview-pr-3243.surge.sh/nightly/Administering_Project/index-satellite.html#using-repository-sources_admin) as a whole, I don't think it breaks anything further down the line..? I feel like we're already asking users to jump through too many hoops here but on the other hand, this could be a security concern.

The reason I ask is that I vague recall that git ignores the config if the permissions are too wide, but I can't reproduce it now.

If it's not needed, please ignore my suggestion. 100% agreed the procedure is already too many steps.

ekohl:
> @ekohl Since you opened the foreman-packaging PR (thanks for that, by the way!), can you please help me figure out which branches should the doc fix go to? Should I cherry-pick only to 3.12 and below because for 3.13 and above, the file will be shipped by default?

I'd suggest to merge it first to master, pick it down to all branches you want. Once the packaging change becomes available we'll submit a docs update and then pick it to the same branches where the packaging update was picked to. We should also keep downstream in mind for that, so I probably wouldn't pick it back further than 3.10.

guides/common/modules/proc_synchronizing-templates-with-an-existing-repository.adoc
asteflova: Okay, thanks everyone. Let's not add it then. (For the record, I did try searching for anything related to git complaining about .gitignore being readable for everyone, but didn't find anything that should prevent us from sticking to what the step currently looks like.)

asteflova:
I'm going to assume there are no technical issues with the current state of this PR so I'm setting `tech review done`. Style review would be appreciated now.

asteflova:
> [...] Once the packaging change becomes available we'll submit a docs update [...]

Another assumption I'm going to make: "We'll submit a docs update" means that you intend to track this on your side. If that's not the case, let me know and I'll file a Jira for the docs team and/or myself.

guides/common/modules/proc_synchronizing-templates-with-an-existing-repository.adoc
```diff
@@ -4,25 +4,45 @@
...
+----
+.. Allow the `foreman` user access to the `.config` directory:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+# chown --recursive foreman /usr/share/foreman/.config
+----
++
```
maximiliankolb, line 26:
```suggestion
```

should not be necessary. unless ...?

maximiliankolb (approved):
one tiny suggestion, rest LGTM. Feel free to merge with or without.

asteflova:

89c1a95938..a5564467f8  3.12 -> 3.12
64fe5dbac0..f3230b9e24  3.11 -> 3.11
3287c30c17..5e6337a2a6  3.10 -> 3.10
57dc4c0221..56f0d6755b  3.9 -> 3.9
68d15e29bb..58c89b3c64  3.8 -> 3.8
2d586624a8..db84943f20  3.7 -> 3.7
763d727f02..cfd432d08c  3.6 -> 3.6
649bfa2f7f..ad13123c07  3.5 -> 3.5

PR #3247, opened 2024-08-30
Update client repository to satellite-6-client-2

AkshayGadhaveRH:
- The newer versions support puppet-agent 8.
- The client repo is versionless and older versions use puppet-agent 7.
- To work around this, `satellite-6-client-2` is created for the later versions with puppet-agent 8.
- Removed the snippet linked in `Installing and configuring Puppet agent manually` as this snippet is used in multiple modules.

JIRA: https://issues.redhat.com/browse/SAT-27092

#### Checklists

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/attributes-satellite.adoc
+++ b/guides/common/attributes-satellite.adoc
@@ -138,13 +138,13 @@
 
 // For a release
 :RepoSatelliteVersion: {ProjectVersion}
-:project-client-2-name: satellite-6-client-2
+:project-client-name: satellite-6-client-2
 
 // RHEL 9 Satellite repos
 :RepoRHEL9ServerSatelliteCapsuleProjectVersion: satellite-capsule-{RepoSatelliteVersion}-for-rhel-9-x86_64-rpms
 :RepoRHEL9ServerSatelliteMaintenanceProjectVersion: satellite-maintenance-{RepoSatelliteVersion}-for-rhel-9-x86_64-rpms
 :RepoRHEL9ServerSatelliteServerProjectVersion: satellite-{RepoSatelliteVersion}-for-rhel-9-x86_64-rpms
-:RepoRHEL9ServerSatelliteToolsProjectVersion: {project-client-2-name}-for-rhel-9-<arch>-rpms
+:RepoRHEL9ServerSatelliteToolsProjectVersion: {project-client-name}-for-rhel-9-<arch>-rpms
 
 // RHEL 8 Satellite repos
 :RepoRHEL8ServerSatelliteCapsuleProjectVersion: satellite-capsule-{RepoSatelliteVersion}-for-rhel-8-x86_64-rpms
@@ -152,14 +152,14 @@
 :RepoRHEL8ServerSatelliteMaintenanceProjectVersionPrevious: satellite-maintenance-{ProjectVersionPrevious}-for-rhel-8-x86_64-rpms
 :RepoRHEL8ServerSatelliteServerProjectVersion: satellite-{RepoSatelliteVersion}-for-rhel-8-x86_64-rpms
 :RepoRHEL8ServerSatelliteServerProjectVersionPrevious: satellite-{ProjectVersionPrevious}-for-rhel-8-x86_64-rpms
-:RepoRHEL8ServerSatelliteToolsProjectVersion: {project-client-2-name}-for-rhel-8-<arch>-rpms
+:RepoRHEL8ServerSatelliteToolsProjectVersion: {project-client-name}-for-rhel-8-<arch>-rpms
 :RepoRHEL8ServerSatelliteUtils: satellite-utils-{RepoSatelliteVersion}-for-rhel-8-x86_64-rpms
 
 // RHEL 7 Satellite repos
 :RepoRHEL7ServerSatelliteMaintenanceProjectVersion: rhel-7-server-satellite-maintenance-{RepoSatelliteVersion}-rpms
 :RepoRHEL7ServerSatelliteServerProjectVersion: rhel-7-server-satellite-{RepoSatelliteVersion}-rpms
 :RepoRHEL7ServerSatelliteServerProjectVersionPrevious: rhel-7-server-satellite-{ProjectVersionPrevious}-rpms
-:RepoRHEL7ServerSatelliteToolsProjectVersion: {project-client-2-name}-for-rhel-7-<arch>-rpms
+:RepoRHEL7ServerSatelliteToolsProjectVersion: {project-client-name}-for-rhel-7-<arch>-rpms
 :RepoRHEL7ServerSatelliteUtils: rhel-7-server-satellite-utils-{RepoSatelliteVersion}-rpms
 // Used in downstream guides
 :RepoRHEL7ServerSoftwareCollections: rhel-server-rhscl-7-rpms
--- a/guides/common/attributes-satellite.adoc
+++ b/guides/common/attributes-satellite.adoc
@@ -144,7 +144,8 @@
 :RepoRHEL9ServerSatelliteCapsuleProjectVersion: satellite-capsule-{RepoSatelliteVersion}-for-rhel-9-x86_64-rpms
 :RepoRHEL9ServerSatelliteMaintenanceProjectVersion: satellite-maintenance-{RepoSatelliteVersion}-for-rhel-9-x86_64-rpms
 :RepoRHEL9ServerSatelliteServerProjectVersion: satellite-{RepoSatelliteVersion}-for-rhel-9-x86_64-rpms
-:RepoRHEL9ServerSatelliteToolsProjectVersion: {project-client-2-name}-for-rhel-9-<arch>-rpms
+:RepoRHEL9ServerSatelliteToolsProjectVersion: satellite-client-6-for-rhel-9-<arch>-rpms
+:RepoRHEL9ServerSatelliteTools2ProjectVersion: rhel-9-server-{project-client-2-name}-rpms
 
 // RHEL 8 Satellite repos
 :RepoRHEL8ServerSatelliteCapsuleProjectVersion: satellite-capsule-{RepoSatelliteVersion}-for-rhel-8-x86_64-rpms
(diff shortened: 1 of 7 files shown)


guides/common/attributes-satellite.adoc
```diff
@@ -136,27 +136,28 @@
...
 // For a release
 :RepoSatelliteVersion: {ProjectVersion}
+:project-client-2-name: satellite-6-client-2
 
 // RHEL 9 Satellite repos
 :RepoRHEL9ServerSatelliteCapsuleProjectVersion: satellite-capsule-{RepoSatelliteVersion}-for-rhel-9-x86_64-rpms
 :RepoRHEL9ServerSatelliteMaintenanceProjectVersion: satellite-maintenance-{RepoSatelliteVersion}-for-rhel-9-x86_64-rpms
 :RepoRHEL9ServerSatelliteServerProjectVersion: satellite-{RepoSatelliteVersion}-for-rhel-9-x86_64-rpms
-:RepoRHEL9ServerSatelliteToolsProjectVersion: satellite-client-6-for-rhel-9-<arch>-rpms
+:RepoRHEL9ServerSatelliteToolsProjectVersion: {project-client-2-name}
```
ekohl, line 145: I'm pretty sure there's one repo built for each RHEL version so I suspect it'll something like this:
```suggestion
:RepoRHEL9ServerSatelliteToolsProjectVersion: {project-client-2-name}-for-rhel-9-<arch>-rpms
```

@ehelms can you confirm?

ehelms: Confirmed

guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
maximiliankolb, line 1: Please revert this change. Instead, you can either adjust the attribute in `guides/common/modules/snip_prerequisite-project-client-repository-enabled.adoc`, or, if necessary, change the attribute value in https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-satellite.adoc#L164.

guides/common/attributes-satellite.adoc
```diff
@@ -136,27 +136,28 @@
 
 // For a release
 :RepoSatelliteVersion: {ProjectVersion}
+:project-client-2-name: satellite-6-client-2
```
maximiliankolb, line 139: I am unsure if it's a good idea to have the number as part of the attribute name.

maximiliankolb (changes requested):

guides/common/attributes-satellite.adoc
AkshayGadhaveRH: @maximiliankolb should I add the `2` to the repo labels?

guides/common/modules/proc_installing-and-configuring-puppet-agent-manually.adoc
maximiliankolb: @AkshayGadhaveRH Maybe reverting is not an option if you explicitely only want version 2 in this procedure. But I still wonder if it makes sense to reference different client versions in different procedures.

guides/common/attributes-satellite.adoc
maximiliankolb: Hmm, I again wonder if my original comment was a good idea. Maybe your solution works best? We already use "project-client-name". I am OK as is as well as adding `-2` to all three attributes.

maximiliankolb (commented):
sorry for the confusing comments!

Lennonka:
Aren't we supposed to support both repos at this time? So instead of replacing it, we should add it?

ekohl:
Good question. I think we want users to use the latest version. I think there's one edge case: if you have n-1 Capsules then they may run Puppetserver 7. You're only supposed to use agents that are at most as new as the server, so having Puppet 8 on clients with a Capsule 6.15 should be unsupported.

Not all Capsules will run Puppetserver, but out of simplicity I'd say that you need to have upgraded the server (Satellite or Capsule, depending on how the host is configured) to 6.16 to use `satellite-6-client-2`. Otherwise you need to stick to the old version.

It's possible to have a mixed environment where some hosts are connected to Satellite 6.16 using `satellite-6-client-2` while other hosts are connected to a Capsule 6.15 using the old version. If the user is careful I don't see how that would be a problem, but generally I'd say users should upgrade their entire deployment (Satellite and all Capsules) before migrating to `satellite-6-client-2`.

Should that be covered in the upgrading guide as well? Probably also a release note, though I know we don't capture Satellite release notes here.

asteflova:
> Good question. [...] I think there's one edge case: [...] If the user is careful I don't see how that would be a problem, but generally I'd say users should [...]

From our d/s perspective, we should not try to document every possible scenario. We should stick to the most user-friendly (and, I suppose, least error-prone) solutions. Without understanding all the details of this particular situation, I just wanted to chime in with my 2 cents: If there is a recommended scenario among many other different scenarios that could go badly if users aren't careful, we should document the recommended one to steer users towards those user stories that we actually want to support.

TL;DR: We shouldn't document everything that's possible. We should document the things that we want users to do.

ekohl:
> TL;DR: We shouldn't document everything that's possible. We should document the things that we want users to do.

I agree, which is why I emphasized on the simple solutions. But we must think about the edge cases because customers will come up with them. If we don't clearly document that users of n-1 first need to upgrade, it's likely that some customer will attempt to do so.

More concrete, we have this line in our Satellite upgrade guide:

> Note that you can upgrade Capsules separately from Satellite.

And this line:

> Note that Satellite Server upgraded from 6.15 to 6.16 can use Capsule Servers still at 6.15.


That will now have an asterisk that there are some limitations if you do.

Lennonka:
I rather meant that in Sat 6.16 we support Puppet agent 7 in the Client 1 repo alongside Puppet agent 8 in the Client 2 repo. So we need to add a new attribute instead of replacing the existing value.

AkshayGadhaveRH:
@Lennonka I'll retain the old attribute as well.

guides/common/attributes-satellite.adoc
```diff
@@ -136,27 +136,28 @@
...
-:RepoRHEL8ServerSatelliteToolsProjectVersion: satellite-client-6-for-rhel-8-<arch>-rpms
+:RepoRHEL8ServerSatelliteToolsProjectVersion: {project-client-2-name}-for-rhel-8-<arch>-rpms
 :RepoRHEL8ServerSatelliteUtils: satellite-utils-{RepoSatelliteVersion}-for-rhel-8-x86_64-rpms
 
 // RHEL 7 Satellite repos
 :RepoRHEL7ServerSatelliteMaintenanceProjectVersion: rhel-7-server-satellite-maintenance-{RepoSatelliteVersion}-rpms
 :RepoRHEL7ServerSatelliteServerProjectVersion: rhel-7-server-satellite-{RepoSatelliteVersion}-rpms
 :RepoRHEL7ServerSatelliteServerProjectVersionPrevious: rhel-7-server-satellite-{ProjectVersionPrevious}-rpms
-:RepoRHEL7ServerSatelliteToolsProjectVersion: rhel-7-server-satellite-client-6-rpms
+:RepoRHEL7ServerSatelliteToolsProjectVersion: {project-client-2-name}-for-rhel-7-<arch>-rpms
```
evgeni, line 160: I would have expected `rhel-7-server-{project-client-2-name}-rpms` or something similar here.

ehelms: > I would have expected `rhel-7-server-{project-client-2-name}-rpms` or something similar here.

Agreed, @AkshayGadhaveRH can you fix it?

AkshayGadhaveRH: Done

AkshayGadhaveRH:
@evgeni @ehelms where should the new repos replace the old ones?

Lennonka:
> where should the new repos replace the old ones?

We need to update the following:

- https://docs.theforeman.org/nightly/Managing_Configurations_Puppet/index-satellite.html - sections about Puppet agent - I'm not sure if we want to provide guidance on upgrading Puppet agent 7 to Puppet agent 8, which might require another procedure
- https://docs.theforeman.org/nightly/Managing_Security_Compliance/index-satellite.html#deploying-compliance-policies_security-compliance - sections 9.5 and 9.6

guides/common/modules/snip_prerequisite-puppet-project-client-repository-enabled.adoc
Lennonka, line 1: The build is failing because you have renamed this snippet, but it has other occurences and now it cannot be found there. But it's unnecessary to rename this snippet because it is not limited to Puppet. Please, undo the renaming.

Lennonka (changes requested):

Lennonka:
FYI With recent updates from the RH engineering, the Client 2 repo will not be provided in 6.16.0 GA. It will be provided later. So we need to monitor it and merge this PR later.

AkshayGadhaveRH:
Hey @Lennonka fixed it. I had preserved the old snippet as well, but I think the build was failing because the `project-client-2-name` attribute was not defined for non-Satellite builds. Added these to the attributes and it seems to build fine now.

Lennonka:
@AkshayGadhaveRH That's good spotting, but I would prefer that we just edit the original snippet instead of adding a new one. It isn't necessary to add the new one because the old one can be reused with edits.

AkshayGadhaveRH:
@Lennonka would it be better to manipulate the attribute for puppet related modules so that it renders to Client-2 in that case?

Lennonka:
@AkshayGadhaveRH I think we should clarify first, whether we want to tell users that they can use either Client 1 or Client 2 repos, or we tell them only about Client 2, in 6.16.

@ehelms Can you please clarify?

ehelms:
> @AkshayGadhaveRH I think we should clarify first, whether we want to tell users that they can use either Client 1 or Client 2 repo, or we tell them only about Client 2, in 6.16.
> 
> @ehelms Can you please clarify?

There is no client 2 right now, this change is not needed for the moment. So we can just close it till we are ready.

AkshayGadhaveRH:
Closing this as the changes have only been made to the puppeterver and the installer making these doc changes unnecessary.

AkshayGadhaveRH:
Closing this as the changes have only been made to the puppeterver and the installer making these doc changes unnecessary.

PR #3261, opened 2024-09-04
Change checksum number

bangelic:
The checksum number `sha1` is no longer secure and modern systems support file checksums greater than `sha1`. It has been replaced with the correct checksum values.

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_creating-a-custom-file-type-repository.adoc
+++ b/guides/common/modules/proc_creating-a-custom-file-type-repository.adoc
@@ -74,7 +74,7 @@ By default, the repository is published through HTTP.
 |====
 | *Option* | *Description*
 
-| `--checksum-type` _sha_version_                 | Repository checksum (either `sha1` or `sha256`)
+| `--checksum-type` _sha_version_                 | Repository checksum (either `sha256`, `sha384`, or `sha512`)
 | `--download-policy` _policy_name_       | Download policy for repositories (either `immediate` or `on_demand`)
 | `--gpg-key-id` _gpg_key_id_                 | GPG key numeric identifier
 | `--gpg-key` _gpg_key_name_                  | Key name to search by


ekohl (approved):
I thinking this is the first time I read "checksum number". The common term is hashing algorithm. You're correct that sha1 is an insecure algorithm and the sha2 family (`sha256`, `sha384` & `sha512`) should be used instead.

This reflects https://github.com/Katello/katello/commit/c98e3271dcb5abedec9fcfe03b52481a2e4a9a95 which is indeed only present on 4.14 and newer.

asteflova (approved):

asteflova:

8818dd1d2a..ba81cc5170  3.12 -> 3.12

asteflova:
For the record, this is why I'm not a fan of copy-pasting information from man pages or --help pages :) We just end up maintaining the same information in two places without offering additional value.

(@bangelic This is not a commentary on your effort! It's great that you keep the docs up-to-date, this is not something that you could have influenced. Thank you for the PR.)

ekohl:
> For the record, this is why I'm not a fan of copy-pasting information from man pages or --help pages :) We just end up maintaining the same information in two places without offering additional value.

If you duplicate it (and I emphasize the **if**), IMHO you need to have an automated way to update it. For example, https://github.com/theforeman/theforeman.org/tree/gh-pages/scripts/installer has scripts to automatically retrieve the correct info and provide it in the same format. It could certainly be better by working even more with 100% generated files that are included, but it's reasonably easy to maintain.

Manually keeping things in sync is a recipe for problems.

asteflova:
We publish this information in an automated way through a script in the Hammer CLI guide. It has a section on [hammer repository create](https://docs.redhat.com/en/documentation/red_hat_satellite/6.15/html-single/hammer_cli_guide/index#hammer-repository-create) (Satellite 6.15 so it still lists sha1). Furthermore, now that I compare the section on hammer repository create in [Managing Content](https://docs.theforeman.org/nightly/Managing_Content/index-katello.html#Creating_a_Custom_File_Type_Repository_content-management) and [Hammer CLI Guide](https://docs.redhat.com/en/documentation/red_hat_satellite/6.15/html-single/hammer_cli_guide/index#hammer-repository-create), it seems that the list of options in Managing Content is far from complete, presumably because no one has been adding new options.

PR #3254, opened 2024-09-03
Remove qpid ports

ehelms:
#### Checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.12/Katello 4.14 (Satellite 6.16)
* [x] Foreman 3.11/Katello 4.13
* [x] Foreman 3.10/Katello 4.12
* [x] Foreman 3.9/Katello 4.11 (Satellite 6.15; orcharhino 6.8/6.9/6.10)
* [ ] Foreman 3.8/Katello 4.10
* [ ] Foreman 3.7/Katello 4.9 (Satellite 6.14)
* [ ] Foreman 3.6/Katello 4.8
* [ ] Foreman 3.5/Katello 4.7 (Satellite 6.13; orcharhino 6.6/6.7)
* We do not accept PRs for Foreman older than 3.5.

PR #3268, opened 2024-09-09
Remove qpid port from load balancing diagram

maximiliankolb:
Refs/follow up to  https://github.com/theforeman/foreman-documentation/pull/3254

@ehelms You can either cherry-pick from my PR or we can merge both PRs in succession.


ehelms (approved):
Let's just merge them independently, they are both correct and need to be done.

maximiliankolb:
triage: tech review enough; @maximiliankolb will merge

PR #3270, opened 2024-09-10
Mitigate installing yggdrasil from EPEL

maximiliankolb:

Mitigate issue on hosts with EPEL enabled when installing REX pull client.

If you install the pull provider for remote execution from Foreman Clients on hosts running AlmaLinux, Amazon Linux, CentOS, Oracle Linux, Red Hat Enterprise Linux, or Rocky Linux, you install the yggdrasil package. Because EPEL also contains a package called “yggdrasil”, you have to ensure that the package comes from the Foreman Client and not EPEL in case it is enabled on your host.

This is a verified solution from ATIX Service Portal.

Another alternative solution is to use `dnf --disablerepo="_My_EPEL_Repository_" install yggdrasil`.

#### Checklists

* [ ] Foreman 3.6/Katello 4.8


guides/common/modules/proc_configuring-a-host-to-use-the-pull-client.adoc
```diff
@@ -21,6 +21,13 @@ ifdef::client-content-dnf[]
 ----
 # {client-package-install-el8} katello-pull-transport-migrate
 ----
+ifndef::satellite[]
++
+Ensure that your host installs `yggdrasil` from {project-client-name}.
+If you have EPEL enabled on your host, you have to lock the version by using `echo "yggdrasil-_My_Package_Version_" >> /etc/dnf/plugins/versionlock.list` or filter the package by using a content view filter.
+You can find the `yggdrasil` package version on {project-client-name}.
+For more information, see {ContentManagementDocURL}Creating_a_Content_Filter_for_Yum_Content_content-management[Creating a content filter for Yum content] in _{ContentManagementDocTitle}_.
```
asteflova, line 29: This seems like too much information to be included as an afterthought after the `katello-pull-transport-migrate` step. Do you see a way to break it down into multiple steps perhaps?

ekohl (commented):
Further context: https://github.com/theforeman/foreman_ygg_worker/pull/21 is working on supporting yggdrasil 0.4.

I indeed see it's in EPEL 8 & 9: https://src.fedoraproject.org/rpms/yggdrasil

Probably better to add `excludepkgs=yggdrasil` in `epel.repo`. If you have synced content views, you can exclude it there.

Though yggdrasil 0.4.2 is also in CentOS Stream 9 now, so there you have a similar problem. I wonder if we should set a more narrow version dependency in `foreman_ygg_worker`. Let's have that discussion in https://github.com/theforeman/foreman-packaging/pull/11232.

cc @evgeni @adamruzicka

guides/common/modules/proc_configuring-a-host-to-use-the-pull-client.adoc
maximiliankolb: I have split this into two equally possible sub-steps. Is this what you had in mind?

asteflova: Yes! Seeing the information broken down into smaller chunks like this makes it a lot less intimidating at first glance. (Three-level nesting is not great but IMO that's the lesser evil.)

guides/common/modules/proc_configuring-a-host-to-use-the-pull-client.adoc
```diff
@@ -21,6 +21,15 @@ ifdef::client-content-dnf[]
 ----
 # {client-package-install-el8} katello-pull-transport-migrate
 ----
+ifndef::satellite[]
++
+If you have EPEL enabled on your host, ensure that your host installs `yggdrasil` from {project-client-name}:
++
+*** Lock the version by using `echo "yggdrasil-_My_Package_Version_" >> /etc/dnf/plugins/versionlock.list`.
```
asteflova, line 28: Have you considered including the command in a code block as usual? Seeing it inline like this is a bit weird.

maximiliankolb: I did not consider this up until now. Applied.

asteflova (approved):
Looks good style-wise.

ekohl (changes requested):
I'd prefer a technical solution to it. As you can see in https://github.com/theforeman/foreman-packaging/pull/11232#issuecomment-2348506114 it's possible to properly enforce the requirement. Right now that's waiting on final approval and cherry picking.

maximiliankolb:
@ekohl Should I reopen this PR against 3.11 or the newest version that does not receive a backport? I can even make it non-Satellite and/or non-orcharhino for versions that receive backported bugfixes.

ekohl:
I intend to backport it to 3.10 (oldest supported version). I'll check internally about Satellite

maximiliankolb:
@ekohl Is that PR still needed for Foreman nightly? I saw that https://github.com/theforeman/foreman-packaging/pull/11232 had been merged.

ekohl:
@adamruzicka any thoughts on this?

ekohl:
Linking the relevant discussion about removal from EPEL: https://bugzilla.redhat.com/show_bug.cgi?id=2274334. That mentions it didn't make it into RHEL 9.5 so it's still available in EPEL for now.

maximiliankolb:
triage: close with no longer necessary.

PR #3271, opened 2024-09-10
Add procedure to troubleshoot permission issues

maximiliankolb:
Refs #30527

See "definitions/checks/foreman/check_duplicate_permission.rb" in foreman_maintain

small procedure to verify permission issues and how to resolve them.

to help users fix permission issues.

#### Checklists

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_troubleshooting-permission-issues.adoc
+++ b/guides/common/modules/proc_troubleshooting-permission-issues.adoc
@@ -2,14 +2,14 @@
 = Troubleshooting permission issues
 
 {Project} upgrades perform pre-upgrade checks.
-If your {Project} upgrade fails due to permission issues, you can identify and remedy them on your {ProjectServer}.
-
-You can perform this procedure if you receive an error similar to the following one:
+If the pre-upgrade check discovers permission issues, it fails with an error similar to the following one:
 
 [options="nowrap", subs="verbatim,quotes,attributes"]
 ----
 2024-01-29T20:50:09 [W|app|] Could not create role 'Ansible Roles Manager': ERF73-0602 [Foreman::PermissionMissingException]: some permissions were not found:
 ----
+
+If you see an error like this on your {ProjectServer}, identify and remedy the permission issues.
 
 .Procedure
 . On your {ProjectServer}, identify permission issues:


guides/common/modules/proc_troubleshooting-permission-issues.adoc
```diff
@@ -0,0 +1,26 @@
+[id="troubleshooting-permission-issues"]
+= Troubleshooting permission issues
+
+You can perform a health check on {ProjectServer} to find permission issues.
```
asteflova, line 4: "Permission issues" can mean a lot of things, not just fixing duplicate permissions. So using "permission issues" in the heading and intro here is too vague and can be misleading. Can you consider a more precise description of what the section contains?

guides/doc-Administering_Project/master.adoc
```diff
@@ -66,6 +66,9 @@ include::common/assembly_using-foreman-webhooks.adoc[leveloffset=+1]
 
 include::common/assembly_searching-and-bookmarking.adoc[leveloffset=+1]
 
+[appendix]
+include::common/modules/proc_troubleshooting-permission-issues.adoc[leveloffset=+1]
+
```
asteflova, line 71: "Managing users and roles" seems like a better fit because that assembly covers procedure-related permissions.

guides/common/modules/proc_troubleshooting-permission-issues.adoc
asteflova, line 4: Can you include a bit more context into the introduction? How will users know that they are in fact dealing with permission issues? Is there a specific action failure, an error message, etc.? Or something that  you expect users would try to look up with a full text search?

guides/common/modules/proc_troubleshooting-permission-issues.adoc
```diff
@@ -0,0 +1,26 @@
...
+----
+. Fix duplicate permission issues:
++
+[options="nowrap", subs="verbatim,quotes,attributes"]
+----
+# foreman-rake db:seed
+----
+
+.Verification
+* Rerun the check to ensure that there are no missing or duplicate permissions:
```
asteflova, line 21:
```suggestion
* Rerun the check to ensure no missing or duplicate permissions remain:
```
"There is/are" is considered an empty phrase.

guides/common/modules/proc_troubleshooting-permission-issues.adoc
```diff
@@ -0,0 +1,26 @@
+[id="troubleshooting-permission-issues"]
+= Troubleshooting permission issues
+
+You can perform a health check on {ProjectServer} to find permission issues.
+
+.Procedure
+. On your {ProjectServer}, find duplicate permissions:
```
asteflova, line 7: Lines 7 and 21 seem to inconsistently describe what the command does. On line 7, you write that the purpose of the command is to "find duplicate permissions"; on line 21, it's "missing or duplicate permissions". Can you try to describe the command in both places the same way?

guides/common/modules/proc_troubleshooting-permission-issues.adoc
maximiliankolb: I have slightly extended the introduction and referenced the redmine issue in my commit message. Should I include parts of the error message?

maximiliankolb (commented):
Rebased to "master" and applied several suggestions.

guides/common/modules/proc_troubleshooting-permission-issues.adoc
asteflova: The revised introduction along with the move to Upgrading provide sufficient context. You can resolve this thread.

asteflova: I like the new introduction, thanks.

Regarding the error message: If the message is something you expect users might first see in their environment and then search for what it means, then yes, including it is a good idea. If you think this scenario is unlikely, then including the message is not necessary. Ultimately, your call :)

asteflova (approved):
Looks good style-wise.

guides/common/modules/proc_troubleshooting-permission-issues.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+{Project} upgrades perform pre-upgrade checks.
+If your {Project} upgrade fails due to permission issues, you can identify and remedy them on your {ProjectServer}.
+
+You can perform this procedure if you receive an error similar to the following one:
+
+[options="nowrap", subs="verbatim,quotes,attributes"]
+----
+2024-01-29T20:50:09 [W|app|] Could not create role 'Ansible Roles Manager': ERF73-0602 [Foreman::PermissionMissingException]: some permissions were not found:
+----
```
asteflova, line 12: I'd like to find a way to avoid self-referential wording ("This procedure").

Also, it's not the upgrade that fails, is it? I'd like to assume the error is displayed in the pre-upgrade check phase, before the upgrade itself even starts.

I tried shuffling words around a bit and would suggest something like this. What do you think?

```suggestion
If the pre-upgrade check discovers permission issues, it fails with an error similar to the following one:

[options="nowrap", subs="verbatim,quotes,attributes"]
----
2024-01-29T20:50:09 [W|app|] Could not create role 'Ansible Roles Manager': ERF73-0602 [Foreman::PermissionMissingException]: some permissions were not found:
----

If you see an error like this on your {ProjectServer}, identify and remedy the permission issues.
```

maximiliankolb: Yes, only the pre-upgrade check fails, but that means as a user, the upgrade process fails.

Applied.

asteflova (approved):

sbernhard (approved):

maximiliankolb:
   94729165f5..d79b3706ef  3.12 -> 3.12
   00494309f6..0d40788422  3.11 -> 3.11

ekohl:
Late to the party, but in https://github.com/theforeman/foreman-documentation/pull/3867#pullrequestreview-2858295107 I had a look at this procedure. I very much think it's a bug, but not related to https://projects.theforeman.org/issues/30527.

aneta-petrova:
However, it seems that running `rake db:seed` is precisely what the logs recommended when the error occurs: https://projects.theforeman.org/issues/30527

PR #3303, opened 2024-09-20
Update security key generation command for DHCP config

Lennonka:

Replacing `dnssec-keygen` with `tsig-keygen`.
The `HMAC-MD5` algorithm is not supported by `dnssec-keygen` on EL 9, but the algorithm is required for Foreman.

The `tsig-keygen` command also works on EL 8.

Because of EL 9 support. The `tsig-keygen` command should be preferred on EL 8 as well.

https://issues.redhat.com/browse/SAT-27986

I have tested that the `tsig-keygen` command is available on both RHEL 9 and 8, and gives the required output.

Unlike `dnssec-keygen`, which generates files, `tsig-keygen` only outputs text on the console and doesn't create any files. Therefore, no files have to be deleted afterwards.

#### Checklists

* [x] Foreman 3.6/Katello 4.8


Lennonka:
@evgeni Technical ack, please? As the verification sufficient as mentioned in the description?

evgeni:
The diff matches what we discussed, yes.

I'd prefer if someone could actually setup an external DHCP using this procedure and verify it works, because I did not.

Lennonka:
I've requested QE testing on my team.

shubhamsg199:
Hello,
`tsig-keygen` works. Tested configuring the external DHCP server with it, I didn't see any issues with RHEL8/9

evgeni (approved):
formal ACK etc ;-)

guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
```diff
@@ -32,16 +32,9 @@ include::snip_firewalld.adoc[]
...
-# dnssec-keygen -a HMAC-MD5 -b 512 -n HOST omapi_key
-----
-+
-As a result, a key pair that consists of two files is created in the current directory.
-. Copy the secret hash from the key:
-+
-[options="nowrap"]
-----
-# grep ^Key Komapi_key.+*.private | cut -d ' ' -f2
+# tsig-keygen -a hmac-md5 omapi_key
```
asteflova, line 35: "omapi_key" would normally be a replaceable but it looks like Foreman expects the key to be named exactly this..? Do you think it would be worth letting users know? I'm thinking that perhaps you could update the step lead-in sentence to something like "Generate a security key named omapi_key:".

guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
```diff
@@ -32,16 +32,9 @@ include::snip_firewalld.adoc[]
...
 ----
+. Copy the key.
 . Edit the `dhcpd` configuration file for all subnets and add the key.
```
asteflova, line 38:
```suggestion
. Edit the `dhcpd` configuration file for all subnets and add the key generated by `tsig-keygen`.
```
This way, you can drop one step.

asteflova (approved):
Two comments. Please consider whether you want to tweak the procedure based on my suggestions, otherwise, looks good.

guides/common/modules/proc_configuring-an-external-dhcp-server.adoc
evgeni: The name *can* be changed, but needs to be configured in other places (look for `--foreman-proxy-plugin-dhcp-remote-isc-key-name`).

asteflova: Perhaps a replaceable named _My_Omapi_Key_Name_ then, used in both procedures? _My_Something_ seems to be the standard throughout the docs set.

It's especially confusing in the step to modify `/etc/foreman-proxy/settings.d/dhcp.yml` in the other procedure where some values are labeled as replaceables (_My_Something_) but omapi_key isn't. That's actually what made me think this was the only permitted name for the key.

Lennonka:
After careful consideration, I've decided not to make further changes. I don't think we have to document every option. Providing just the key name that works out of the box seems sufficient.

Lennonka:
Cherry-picked:
- 3d5e5d4295..b4097cfa40  3.12 -> 3.12
- 0ddfeaa68d..977bcad362  3.11 -> 3.11
- 858cc14a08..9b2c98d222  3.10 -> 3.10
- b6faa93d1b..2c53c6e9f2  3.9 -> 3.9
- 05257b8211..3afdb52b6d  3.8 -> 3.8
- e855f380fb..6ac92302d5  3.7 -> 3.7
- 3bea1ac680..e45349c0be  3.6 -> 3.6
- 5f5c6de9b8..caa6138878  3.5 -> 3.5

PR #3311, opened 2024-09-23
Update Hammer reference for 3.12/6.16

asteflova:

I'm adding Hammer reference generated based on Foreman 3.12/Satellite 6.16 by using guides/scripts/generate-hammer-reference.sh.

The script provides an automated way of generating an up-to-date list of Hammer commands, subcommands, and options for a particular Foreman version.

We are not supposed to modify the generated assembly and modules, but I had to fix a couple of build failures related to apparent attributes like `%{package}`. I addressed this in a separate commit.

I found one other small issue with the generated content but because it has been there in earlier d/s-only versions of the guide (see the description for `--subscriptions` in [3.2.2. activation-key add-subscription](https://docs.redhat.com/en/documentation/red_hat_satellite/6.15/html-single/hammer_cli_guide/index#hammer-activation-key-add-subscription)), I suggest to leave this as is. The proper fix is to update the script.

#### Checklists

* [ ] Foreman 3.6/Katello 4.8


asteflova:
Vale can't check the diff:

`reviewdog: fail to get diff: GET https://api.github.com/repos/theforeman/foreman-documentation/pulls/3311: 406 Sorry, the diff exceeded the maximum number of lines (20000)
`

so I propose to ignore the Vale check. The content is generated from --help pages anyway and those are not subject to style guides and Vale checks.

asteflova:
And it looks like the script generates "Satellite" everywhere, understandably. I suppose that means I should exclude the reference for non-Satellite builds. Thoughts?

Lennonka:
> And it looks like the script generates "Satellite" everywhere, understandably. I suppose that means I should exclude the reference for non-Satellite builds. Thoughts?

Thank you for taking on this task.

Actually, the idea was to update the reference directly in downstream for Satellite and orcharhino, and keep upstream for Foreman only because one needs the branded output from Hammer help. I suggest that you close this PR and re-open it downstream.

I have yet to ensure that our sync will not overwrite the reference modules. cc @maximiliankolb I forgot about those! :)

PR #3313, opened 2024-09-23
Remove attaching Satellite Infrastructure subscription

Lennonka:

Removal of attaching Satellite Infrastructure subscription

The procedure is only needed for SCA-disabled organizations, but in Foreman 3.12/Satellite 6.16, all organizations are SCA-enabled and SCA is enforced.

https://issues.redhat.com/browse/SAT-27606

I didn't find an include of this module in _Installing Server Disconnected_. Therefore, no removal there.

#### Checklists

* [x] Foreman 3.6/Katello 4.8


Lennonka:
@evgeni Hi! Can you please review this PR? Should I also remove it from [System requirements](https://docs.theforeman.org/nightly/Installing_Server/index-satellite.html#system-requirements_satellite)?

evgeni:
I am not exactly an expert in subs (:eyes: @jeremylenz @sideangleside), but will try my best here :)

This is what I know (or think to know, heh):
* Support for non-SCA manifests (= the subscriptions that are responsible for the content the Katello/Satellite *clients* receive) has been removed in Katello 4.12 (and thus any Satellite release based on this or later versions).
* Installing packages directly from cdn.redhat.com (after using `subscription-manager register` etc) is not affected by this, as that's a RHEL feature, not Katello/Satellite
* If I read https://access.redhat.com/articles/transition_of_subscription_services_to_the_hybrid_cloud_console correctly, *everyone* will be switched to SCA-mode in Oct/Nov, so you won't need `subscription-manager attach` to obtain specific subs anymore, even if you were using non-SCA mode against the CDN before.

For the docs/installation process, this means:
* You still need to *own* a Satellite Infra subscription (that's what "Your Satellite must have the Red Hat Satellite Infrastructure Subscription manifest in your Customer Portal" in the system requirements is about)
* You don't have to *attach* that subscription using `subscription-manager attach …`
* This applies to *all* Satellite versions, not only the ones that don't support non-SCA mode for *their* clients.

TL;DR:
* your patch is correct
* it needs to be cherry-picked to all supported releases, not only 4.14/6.16
* I am not sure whether this should happen now, or closer to the actual "force-SCA" date

Lennonka:
Thank you. However, I would prefer additional ack from @jeremylenz. Are we sure about the cherry-picking to older Satellites?

jeremylenz:
Right, if we're running `subscription-manager attach` on the Satellite and not on a host, then it doesn't matter whether the Satellite is SCA-only or not. I agree with @evgeni that this affects all Satellite versions, and we no longer need to attach the Satellite infrastructure subscription anywhere once all RH accounts are SCA.

As to the specific date that RH accounts will all be SCA, we can ask @sideangleside. :eyes:

sideangleside:
The date when Red Hat accounts switch to SCA only is on/around 25-Oct-2024, as documented  [here](https://access.redhat.com/articles/transition_of_subscription_services_to_the_hybrid_cloud_console#important-upcoming-dates-and-releases-2) 

Also, `subscription-manager` will, in accounts where SCA is enabled, simply 'no-op' the request to attach. 

```
subscription-manager attach --pool "8a85f99c6895cc9901689a87d3f01323"
Ignoring the request to attach. Attaching subscriptions is disabled for organization "XXXXX" because Simple Content Access (SCA) is enabled.


```

Lennonka:
Does it mean I should wait until Oct 25th to merge this PR?

jeremylenz:
> * I am not sure whether this should happen now, or closer to the actual "force-SCA" date

My vote would be to just do it now. thoughts from others welcome

Lennonka:
@ehelms Eric, can I merge and cherry pick this now?

evgeni:
> > * I am not sure whether this should happen now, or closer to the actual "force-SCA" date
> 
> My vote would be to just do it now. thoughts from others welcome

So would be mine!

Lennonka:
Alright then, let's party! :tada:

Lennonka:
Cherry-picked:
- d6a76d1457..0ca4ae8495  3.12 -> 3.12
- 977bcad362..4bb5eb5e2f  3.11 -> 3.11
- 9b2c98d222..9ecce9d65c  3.10 -> 3.10
- 2c53c6e9f2..33751dd1c8  3.9 -> 3.9
- 3afdb52b6d..5eee4460bd  3.8 -> 3.8
- 6ac92302d5..ca9e615630  3.7 -> 3.7
- e45349c0be..a36828baaa  3.6 -> 3.6
- caa6138878..e1d652dcae  3.5 -> 3.5

PR #3329, opened 2024-09-30
Document CLI, Ansible, and API for registering a host

asteflova:
#### What changes are you introducing?

I'm describing alternative ways of registering a host: CLI, Ansible, API.

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

https://issues.redhat.com/browse/SAT-28377 requested adding the Hammer command and API resource that correspond to the existing web UI procedure as users weren't aware of their existence and were unable to look them up. Then on top of that, prompted by a conversation with @evgeni, I thought it might be a good idea to add a reference to the corresponding Ansible module too.

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)

The CLI, Ansible, and API "procedures" are very minimal by design. Users have very good reference documentation available to them (`--help`, `ansible-doc`, and built-in API reference) and I don't want to duplicate it in the guides.

I'm requesting cherry-picks down to 3.7 only because that's what https://github.com/theforeman/foreman-documentation/pull/3143, which targets the same procedure, did.

#### Checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.12/Katello 4.14 (Satellite 6.16)
* [x] Foreman 3.11/Katello 4.13
* [x] Foreman 3.10/Katello 4.12
* [x] Foreman 3.9/Katello 4.11 (Satellite 6.15; orcharhino 6.8/6.9/6.10)
* [x] Foreman 3.8/Katello 4.10
* [x] Foreman 3.7/Katello 4.9 (Satellite 6.14)
* [ ] Foreman 3.6/Katello 4.8
* [ ] Foreman 3.5/Katello 4.7 (Satellite 6.13; orcharhino 6.6/6.7)
* We do not accept PRs for Foreman older than 3.5.

PR #2865, opened 2024-03-06
Reduce doc bloat in procedure for creating a user

asteflova:
During peer review for https://github.com/theforeman/foreman-documentation/pull/2861, there was a suggestion to include a piece of information in the procedure module for creating a user. It's a good idea, but more work needs to be done within that procedure modules to make it happen.

EDIT: In this PR, I propose the following changes to the procedure to create a user.

* Update UI labels (it looks like they changed in the web UI, but not in the docs)
* Delete steps that are clear from the web UI (like "use enter First Name and Surname to enter first name and surname")
* Split the procedure into two parts: one that shows how to find the page to create a user; another that highlights notable account details to enter
* Minor edits to the CLI part of the procedure

This is an effort to reduce maintenance burden (the fewer UI labels are documented, the less chance there is of the docs becoming outdated), usability (users don't need UI instructions that are this detailed), and scannability (being able to identify the really important parts of the procedure).

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.10/Katello 4.12
* [x] Foreman 3.9/Katello 4.11 (planned Satellite 6.15)
* [ ] Foreman 3.8/Katello 4.10
* [ ] Foreman 3.7/Katello 4.9 (Satellite 6.14)
* [ ] Foreman 3.6/Katello 4.8
* [ ] Foreman 3.5/Katello 4.7 (Satellite 6.13; orcharhino 6.6/6.7)
* [ ] Foreman 3.4/Katello 4.6 (EL8 only)
* [ ] Foreman 3.3/Katello 4.5 on EL7 & EL8 (Satellite 6.12 on EL8 only; orcharhino 6.4/6.5 on EL8 only)
* We do not accept PRs for Foreman older than 3.3.

PR #3330, opened 2024-09-30
Reduce doc bloat in registering a host

asteflova:

I cross-checked the existing web UI procedure for registering a host with the page in the web UI and identified the parts of the documented procedure that can be dropped because they are actually documented in the web UI. Additionally, I'm making minor changes to the wording in order to make the procedure even shorter.

The current procedure is _very_ long (22 steps!). Making sure we don't duplicate the web UI elements (obvious input fields, tooltips, etc.) will go a long way towards making the procedure shorter.

I wanted to do this because in https://github.com/theforeman/foreman-documentation/pull/3329, I'm introducing information on how to perform the action from the CLI, using Ansible, or using API. Without trimming the web UI procedure, the new additions would be lost at the end of a huge web UI procedure.

In editing the procedure, I followed the template for web UI procedures established in https://github.com/theforeman/foreman-documentation/pull/2865.

I'm requesting cherry-picks down to 3.7 only because that's what https://github.com/theforeman/foreman-documentation/pull/3143, which targets the same procedure, did.

#### Checklists

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_registering-a-host.adoc
+++ b/guides/common/modules/proc_registering-a-host.adoc
@@ -42,7 +42,7 @@ endif::[]
 .Procedure
 . In the {ProjectWebUI}, navigate to *Hosts* > *Register Host*.
 . Enter the details for how you want the registered hosts to be configured.
-. Click *Generate*.
+. Click *Generate* to generate a `curl` command.
 . Run the `curl` command as `root` on the host that you want to register.
 After registration completes, any Ansible roles assigned to a host group you specified when configuring the registration template will run on the host.
 
--- a/guides/common/modules/proc_registering-a-host.adoc
+++ b/guides/common/modules/proc_registering-a-host.adoc
@@ -51,9 +51,9 @@ After registration completes, any Ansible roles assigned to a host group you spe
 
 The registration details that you can specify include the following:
 
-* Optional: On the *General* tab, in the *{SmartProxy}* field, select the {SmartProxy} to register hosts through.
+* On the *General* tab, in the *{SmartProxy}* field, you can select the {SmartProxy} to register hosts through.
 A {SmartProxy} behind a load balancer takes precedence over a {SmartProxy} selected in the {ProjectWebUI} as the content source of the host.
-* Optional: On the *General* tab, select the *Insecure* option to make the first call insecure.
+* On the *General* tab, you can select the *Insecure* option to make the first call insecure.
 During this first call, hosts download the CA file from {Project}.
 Hosts will use this CA file to connect to {Project} with all future calls making them secure.
 +
@@ -87,9 +87,9 @@ The following is an example of the `curl` command with the `--insecure` option:
 ----
 # curl -sS --insecure https://{foreman-example-com}/register ...
 ----
-* Optional: On the *Advanced* tab, in the *Repository* field, list repositories to be added before the registration is performed.
+* On the *Advanced* tab, in the *Repository* field, you can list repositories to be added before the registration is performed.
 You do not have to specify repositories if you provide them in an activation key.
-* Optional: On the *Advanced* tab, in the *Token lifetime (hours)* field, change the validity duration of the JSON Web Token (JWT) that {Project} uses for authentication.
+* On the *Advanced* tab, in the *Token lifetime (hours)* field, you can change the validity duration of the JSON Web Token (JWT) that {Project} uses for authentication.
 The duration of this token defines how long the generated `curl` command works.
 +
 Note that {Project} applies the permissions of the user who generates the `curl` command to authorization of hosts.
--- a/guides/common/modules/proc_registering-a-host.adoc
+++ b/guides/common/modules/proc_registering-a-host.adoc
@@ -87,7 +87,7 @@ The following is an example of the `curl` command with the `--insecure` option:
 ----
 # curl -sS --insecure https://{foreman-example-com}/register ...
 ----
-* On the *Advanced* tab, in the *Repository* field, you can list repositories to be added before the registration is performed.
+* On the *Advanced* tab, in the *Repositories* field, you can list repositories to be added before the registration is performed.
 You do not have to specify repositories if you provide them in an activation key.
 * On the *Advanced* tab, in the *Token lifetime (hours)* field, you can change the validity duration of the JSON Web Token (JWT) that {Project} uses for authentication.
 The duration of this token defines how long the generated `curl` command works.
--- a/guides/common/modules/proc_registering-a-host.adoc
+++ b/guides/common/modules/proc_registering-a-host.adoc
@@ -42,6 +42,9 @@ endif::[]
 .Procedure
 . In the {ProjectWebUI}, navigate to *Hosts* > *Register Host*.
 . Enter the details for how you want the registered hosts to be configured.
+ifdef::katello,satellite,orcharhino[]
+. On the *General* tab, in the *Activation Keys* field, enter one or more activation keys to assign to hosts.
+endif::[]
 . Click *Generate* to generate a `curl` command.
 . Run the `curl` command as `root` on the host that you want to register.
 After registration completes, any Ansible roles assigned to a host group you specified when configuring the registration template will run on the host.
--- a/guides/common/modules/proc_registering-a-host.adoc
+++ b/guides/common/modules/proc_registering-a-host.adoc
@@ -41,26 +41,16 @@ endif::[]
 
 .Procedure
 . In the {ProjectWebUI}, navigate to *Hosts* > *Register Host*.
-. Optional: Select a different *Organization*.
-. Optional: Select a different *Location*.
-. Optional: From the *Host Group* list, select the host group to associate the hosts with.
-Fields that inherit value from *Host group*: *Operating system*, *Activation Keys* and *Lifecycle environment*.
-+
-If your host group has any Ansible roles assigned, the Ansible roles will run on your host upon the registration.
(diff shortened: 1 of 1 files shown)


guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -41,26 +41,16 @@ endif::[]
...
-[NOTE]
-====
-A {SmartProxy} behind a load balancer takes precedence over a {SmartProxy} selected in the {ProjectWebUI} as the host's content source.
-====
-ifdef::katello,satellite,orcharhino[]
-. In the *Activation Keys* field, enter one or more activation keys to assign to hosts.
-endif::[]
-. Optional: Select the *Insecure* option, if you want to make the first call insecure.
+. Enter the details for how you want the registered hosts to be configured.
+. Click *Generate*.
```
apinnick, line 45:
```suggestion
. Click *Generate* to generate a `curl` command.
```

guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -41,26 +41,16 @@ endif::[]
...
+After registration completes, any Ansible roles assigned to a host group you specified when configuring the registration template will run on the host.
+
+The registration details that you can specify include the following:
+
+* Optional: On the *General* tab, in the *{SmartProxy}* field, select the {SmartProxy} to register hosts through.
+A {SmartProxy} behind a load balancer takes precedence over a {SmartProxy} selected in the {ProjectWebUI} as the content source of the host.
+* Optional: On the *General* tab, select the *Insecure* option to make the first call insecure.
 During this first call, hosts download the CA file from {Project}.
 Hosts will use this CA file to connect to {Project} with all future calls making them secure.
 +
```
apinnick, line 56: I know this is outside the scope of your changes, but it looks like the CA file manual copy/install is a separate procedure. You might want to break it out so that it has its own procedure module if it does not already exist.

If the procedure is used frequently, it could be turned into a snippet. Your call.

guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -41,26 +41,16 @@ endif::[]
...
-ifdef::katello,satellite,orcharhino[]
-. In the *Activation Keys* field, enter one or more activation keys to assign to hosts.
-endif::[]
-. Optional: Select the *Insecure* option, if you want to make the first call insecure.
+. Enter the details for how you want the registered hosts to be configured.
+. Click *Generate*.
+. Run the `curl` command as `root` on the host that you want to register.
+After registration completes, any Ansible roles assigned to a host group you specified when configuring the registration template will run on the host.
+
+The registration details that you can specify include the following:
```
apinnick, line 49:
```suggestion
You can configure the following registration options:
```
If they are all optional, this lead-in sentence means that you can remove "Optional:" from the steps. 

I think steps are easier to ready if they do not begin with "Optional:", so I try to keep their use to a minimum.

apinnick (commented):
Big improvement!!! I made a few suggestions.

guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -41,26 +41,16 @@ endif::[]
...
-+
-[NOTE]
-====
-A {SmartProxy} behind a load balancer takes precedence over a {SmartProxy} selected in the {ProjectWebUI} as the host's content source.
-====
-ifdef::katello,satellite,orcharhino[]
-. In the *Activation Keys* field, enter one or more activation keys to assign to hosts.
-endif::[]
-. Optional: Select the *Insecure* option, if you want to make the first call insecure.
+. Enter the details for how you want the registered hosts to be configured.
```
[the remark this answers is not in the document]
asteflova: The mandatory fields are marked with an asterisk:

Screenshot from 2024-10-01 11-30-33

I would argue that this is enough and there is no need to duplicate the same information in the guide.

Brevity is one reason to do it this way. Another is maintainability: We wouldn't need to worry about making sure we list all the mandatory fields -- the web UI already does!

Does this address your concern?

guides/common/modules/proc_registering-a-host.adoc
asteflova: > If they are all optional, this lead-in sentence means that you can remove "Optional:" from the steps.

However, what if we need to add a required step some time in the future and forget to update the lead-in sentence? In reality (and under pressure) writers sometimes don't read all the surrounding text to make sure it's still valid. So I'd prefer to use a sentence that will work under all circumstances, no matter what kind of option (optional/required) gets added to the list of options. WDYT?

> I think steps are easier to ready if they do not begin with "Optional:", so I try to keep their use to a minimum.

Agreed. I could also just drop all the "Optional:" bits..? But then again, the rest of each sentence would make it sound as if the option is required:

> Optional: On the General tab, in the Capsule field, select the Capsule to register hosts through.

vs

> On the General tab, in the Capsule field, select the Capsule to register hosts through.

guides/common/modules/proc_registering-a-host.adoc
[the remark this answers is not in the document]
apinnick: I still think it's better to document a mandatory field even when it is marked in the UI. It's a single step, so brevity is not a concern in this case. 

In my experience, maintainability is not a major concern because mandatory fields tend not to change. (Same cannot be said about options.) Sometimes a writer forgets to update prerequisites. If a step says "Enter the activation keys.", this alerts the user to a possible prerequisite. 

I tend to read through procedures before performing them and I like to know what I'm doing before I start. That's my personal preference. Other users might not feel that way. 

As I said, it is your call.

guides/common/modules/proc_registering-a-host.adoc
asteflova: I agree, the description doesn't look right within the procedure. But you are right, it's out of scope of this PR. Let's file an issue.

guides/common/modules/proc_registering-a-host.adoc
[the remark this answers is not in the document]
asteflova: That makes sense, thanks for explaining your point of view. I'll add it.

guides/common/modules/proc_registering-a-host.adoc
asteflova: https://github.com/theforeman/foreman-documentation/issues/3335

guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -94,87 +87,16 @@ The following is an example of the `curl` command with the `--insecure` option:
...
-* In the *Repository* field, enter a repository to be added before the registration is performed.
-ifdef::client-content-dnf[]
-For Red Hat family distributions, enter the URL of the repository, for example `\http://rpm.example.com/`.
-endif::[]
-ifdef::client-content-apt[]
-For Debian OS families, enter the whole line of list file content, for example `deb \http://deb.example.com/ bookworm 1.0`.
-endif::[]
-* Optional: In the *Repository GPG key URL* field, specify the public key to verify the signatures of GPG-signed packages.
-It needs to be specified in the ASCII form with the GPG public key header.
-ifdef::katello,orcharhino,satellite[]
```
maximiliankolb, line 136: I'd like to keep this; at least for non-satellite builds. IMHO it add value because it explains how to insert the URLs and what type of GPG pub key URL.

Also, line 90" in the repository field" is wrong because by now, it's a set of repositories.

Foreman 3.12:

image

If you argue that we can drop it because the tooltip explains the fields sufficiently, then I'm also OK.

maximiliankolb (commented):
Overall, I very much appreciate the change! I have on change request "repository field".

guides/common/modules/proc_registering-a-host.adoc
asteflova: > If you argue that we can drop it because the tooltip explains the fields sufficiently, then I'm also OK.

Yes :) For transparency (and to show that I did my due diligence and am not just deleting things on a whim :)), this is the tooltip for the `Repository` field:

`A repository to be added before the registration is performed. For Red Hat and SUSE family distributions, this should be the URL of the repository, e.g. 'http://rpm.example.com/'. For Debian OS families, it's the whole list file content, e.g. 'deb http://deb.example.com/ buster 1.0'.`

Tooltip for the `Repository GPG key URL` field:

`If packages are GPG signed, the public key can be specified here to verify the packages signatures. It needs to be specified in the ascii form with the GPG public key header.
`

Information-wise, these tooltips match what we currently document. Wording-wise, they don't follow our style guides but that doesn't justify duplicating the information.

asteflova: > Also, line 90" in the repository field" is wrong because by now, it's a set of repositories.

Good catch! I changed the line to refer to `Repositories` field.

guides/common/modules/proc_registering-a-host.adoc
asteflova: I kept the original lead-in sentence, but dropped "Optional:" and rephrased each option description to use "you can" instead of a plain imperative. I hope that should address the concern while still keeping the wording precise.

maximiliankolb (approved):
LGTM

asteflova:

d4cffa3be6..f5f04b34f6  3.12 -> 3.12
c7e7f7bc95..7802f22eb9  3.11 -> 3.11 (minor conflicts here and below)
9fdcce827e..3d4cead6da  3.10 -> 3.10
884d19b00f..caca9c03b6  3.9 -> 3.9

I didn't cherry pick further because there was more than one conflict and I'd have to check the web UI for all the versions if I wanted to be sure. IMO not worth it.

PR #3350, opened 2024-10-07
Drop Internet Explorer reference

asteflova:

I'm removing a reference to configuring Internet Explorer for Kerberos SSO. After that, the remaining content in the module can't really stand on its own so I'm merging it with another module.

Internet Explorer is being deprecated for years now. Its latest version is still officially supported, but in a very limited way with Microsoft Edge now being the recommended browser for Windows users.

Initially, I thought about replacing the entry about IE with a similar one about Edge. But then I thought that we probably don't need to (or even shouldn't) document third-party software unless absolutely necessary. Microsoft Edge users should use Microsoft docs for that.

#### Checklists

* [ ] Foreman 3.6/Katello 4.8


asteflova:
I believe Microsoft's support policy for Internet Explorer should be enough to set `tech review done`: https://learn.microsoft.com/en-us/lifecycle/faq/internet-explorer-microsoft-edge

maximiliankolb (approved):

asteflova:

ec31851a1a..1f5f8ecead  3.12 -> 3.12
423106cd95..b88843390a  3.11 -> 3.11

PR #3357, opened 2024-10-08
Add PostgreSQL space requirement in prerequisites

AkshayGadhaveRH:
As PostgreSQL is upgraded from 12 to 13 during the upgrade process, it is necessary to maintain enough space for this. Adding this as a prerequisite to the upgrade.

Ref: https://github.com/theforeman/foreman_maintain/blob/master/definitions/checks/disk/available_space_postgresql13.rb

JIRA: https://issues.redhat.com/browse/SAT-28667

#### Checklists

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/con_upgrading-prerequisites.adoc
+++ b/guides/common/modules/con_upgrading-prerequisites.adoc
@@ -17,7 +17,7 @@ Upgrading {SmartProxy} takes approximately 10{range}30 minutes.
 * Ensure that you have sufficient storage space on your server.
 For more information, see {InstallingServerDocURL}Preparing_your_Environment_for_Installation_{project-context}[Preparing your Environment for Installation] in _{InstallingServerDocTitle}_ and {InstallingSmartProxyDocURL}preparing-environment-for-capsule-installation[Preparing your Environment for Installation] in _Installing {SmartProxyServer}_.
 ifndef::foreman-deb[]
-* Ensure that you have at least the same amount of free space as that consumed by `{postgresql-lib-dir}/data`.
+* Ensure that you have at least the same amount of free space on `{postgresql-lib-dir}` as that consumed by `{postgresql-data-dir}`.
 Upgrading to {Project} {ProjectVersion} involves a PostgreSQL 12 to PostgreSQL 13 upgrade.
 The contents of `{postgresql-lib-dir}` are backed up during the PostgreSQL upgrade. 
 endif::[]
--- a/guides/common/modules/con_upgrading-prerequisites.adoc
+++ b/guides/common/modules/con_upgrading-prerequisites.adoc
@@ -16,9 +16,11 @@ Upgrading {SmartProxy} takes approximately 10{range}30 minutes.
 
 * Ensure that you have sufficient storage space on your server.
 For more information, see {InstallingServerDocURL}Preparing_your_Environment_for_Installation_{project-context}[Preparing your Environment for Installation] in _{InstallingServerDocTitle}_ and {InstallingSmartProxyDocURL}preparing-environment-for-capsule-installation[Preparing your Environment for Installation] in _Installing {SmartProxyServer}_.
-* Ensure that you have at least the same amount of free space as that consumed by `/var/lib/pgsql/data` for PostgreSQL upgrade.
-The upgrade involves PostgreSQL 12 to PostgreSQL 13 upgrade.
-The existing postgreSQL data is backed up during the PostgreSQL upgrade. 
+ifndef::foreman-deb[]
+* Ensure that you have at least the same amount of free space as that consumed by `/var/lib/pgsql/data`.
+Upgrading to {Project} {ProjectVersion} involves a PostgreSQL 12 to PostgreSQL 13 upgrade.
+The contents of `{postgresql-lib-dir}` are backed up during the PostgreSQL upgrade. 
+endif::[]
 * Back up your {ProjectServer} and all {SmartProxyServers}.
 For more information, see {AdministeringDocURL}backing-up-{project-context}-server-and-{smart-proxy-context}_admin[Backing Up {ProjectServer} and {SmartProxyServer}] in _{AdministeringDocTitle}_.
 * Plan for updating any scripts you use that contain {Project} API commands because some API commands differ between versions of {Project}.
--- a/guides/common/modules/con_upgrading-prerequisites.adoc
+++ b/guides/common/modules/con_upgrading-prerequisites.adoc
@@ -17,6 +17,8 @@ Upgrading {SmartProxy} takes approximately 10{range}30 minutes.
 * Ensure that you have sufficient storage space on your server.
 For more information, see {InstallingServerDocURL}Preparing_your_Environment_for_Installation_{project-context}[Preparing your Environment for Installation] in _{InstallingServerDocTitle}_ and {InstallingSmartProxyDocURL}preparing-environment-for-capsule-installation[Preparing your Environment for Installation] in _Installing {SmartProxyServer}_.
 * Ensure that you have at least the same amount of free space as that consumed by `/var/lib/pgsql/data` for PostgreSQL upgrade.
+The upgrade involves PostgreSQL 12 to PostgreSQL 13 upgrade.
+The existing postgreSQL data is backed up during the PostgreSQL upgrade. 
 * Back up your {ProjectServer} and all {SmartProxyServers}.
 For more information, see {AdministeringDocURL}backing-up-{project-context}-server-and-{smart-proxy-context}_admin[Backing Up {ProjectServer} and {SmartProxyServer}] in _{AdministeringDocTitle}_.
 * Plan for updating any scripts you use that contain {Project} API commands because some API commands differ between versions of {Project}.
--- a/guides/common/modules/con_upgrading-prerequisites.adoc
+++ b/guides/common/modules/con_upgrading-prerequisites.adoc
@@ -19,7 +19,7 @@ For more information, see {InstallingServerDocURL}Preparing_your_Environment_for
 ifndef::foreman-deb[]
 * Ensure that you have at least the same amount of free space on `{postgresql-lib-dir}` as that consumed by `{postgresql-data-dir}`.
 Upgrading to {Project} {ProjectVersion} involves a PostgreSQL 12 to PostgreSQL 13 upgrade.
-The contents of `{postgresql-lib-dir}` are backed up during the PostgreSQL upgrade. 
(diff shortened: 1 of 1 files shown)


guides/common/modules/con_upgrading-prerequisites.adoc
```diff
@@ -16,6 +16,7 @@ Upgrading {SmartProxy} takes approximately 10{range}30 minutes.
 
 * Ensure that you have sufficient storage space on your server.
 For more information, see {InstallingServerDocURL}Preparing_your_Environment_for_Installation_{project-context}[Preparing your Environment for Installation] in _{InstallingServerDocTitle}_ and {InstallingSmartProxyDocURL}preparing-environment-for-capsule-installation[Preparing your Environment for Installation] in _Installing {SmartProxyServer}_.
+* Ensure that you have at least the same amount of free space as that consumed by `/var/lib/pgsql/data` for PostgreSQL upgrade.
```
ekohl, line 19: This should be RPM-only and use the attribute we already have
```suggestion
ifndef::foreman-deb[]
* Ensure that you have at least the same amount of free space as that consumed by `{postgresql-data-dir}` for PostgreSQL upgrade.
endif::[]
```

ekohl (changes requested):

guides/common/modules/con_upgrading-prerequisites.adoc
ahumbe: @ekohl hi, Instead of saying "some amount of free space" can we provide any rough estimate about the required free space?  Like if the existing size of /var/lib/pgsql/data is 15G then you need at least XX G free on the partition. As some customers have dedicated partition for /var/lib/pgsql/ .

ekohl: @ahumbe it's the same amount because it's literally copied to a backup location, so in your example it'd be at least 15G free.

We have implemented this in a technical check as well: our installer hook checks exactly that in https://github.com/theforeman/foreman-installer/blob/67702be227056c29bbe9e34089009d69bb3a6758/hooks/boot/06-postgresql-upgrade-extensions.rb#L67-L81 and foreman_maintain has it in https://github.com/theforeman/foreman_maintain/blob/master/definitions/checks/disk/available_space_postgresql13.rb.

Now that I read that, I see it uses `/var/lib/pgsql` for the checks so it should be:
```suggestion
ifndef::foreman-deb[]
* Ensure that you have at least the same amount of free space as that consumed by `{postgresql-lib-dir}` for PostgreSQL upgrade.
endif::[]
```

ekohl (approved):
Tech ack, but someone please do the style part. Given @ahumbe had a question about it I'm not sure it's clear enough.

asteflova:
Looking at the comment in https://github.com/theforeman/foreman-documentation/pull/3357/files/ae505f817e93a84552f4538c87ef59d7aad4e133#r1792231647, I think Ashish might have simply misread the line but to be fair, "same amount" and "some amount" are very similar and this could happen to users too.

While the new bullet point explains the actual requirement, it doesn't explain the context. (What PostgreSQL upgrade? Why does PostgreSQL need upgrading? When does it happen? Should _I_ be the one to perform the upgrade?) I'm not suggesting that all these questions need to be answered -- that would be too many words for a simple prerequisite. But adding just a little bit more detail about what is happening during the upgrade would help prevent users from asking these questions in the first place.

More importantly, the related release notes ([upstream](https://github.com/theforeman/theforeman.org/pull/2134/files) and [downstream](https://issues.redhat.com/browse/SAT-23369)) mention that users can delete `/var/lib/pgsql/data-old/` after the upgrade completes. If this is valid, then I'd recommend including it as a step in the upgrade procedure.

guides/common/modules/con_upgrading-prerequisites.adoc
```diff
@@ -16,6 +16,9 @@ Upgrading {SmartProxy} takes approximately 10{range}30 minutes.
...
```
asteflova, line 19: To avoid repeating "PostgreSQL upgrade" more than necessary:
```suggestion
* Ensure that you have at least the same amount of free space as that consumed by `/var/lib/pgsql/data`.
```

guides/common/modules/con_upgrading-prerequisites.adoc
```diff
@@ -16,6 +16,9 @@ Upgrading {SmartProxy} takes approximately 10{range}30 minutes.
...
+The upgrade involves PostgreSQL 12 to PostgreSQL 13 upgrade.
```
asteflova, line 20: On this line, you are talking about two different upgrades: Project and PostgreSQL. I recommend to specify each of them to prevent confusion. This would be one way to do it:
```suggestion
Upgrading to {Project} {ProjectVersion} involves a PostgreSQL 12 to PostgreSQL 13 upgrade.
```

guides/common/modules/con_upgrading-prerequisites.adoc
```diff
@@ -16,6 +16,9 @@ Upgrading {SmartProxy} takes approximately 10{range}30 minutes.
...
+The existing postgreSQL data is backed up during the PostgreSQL upgrade.
```
asteflova, line 21: At the very least, capitalization:
```suggestion
The existing PostgreSQL data is backed up during the PostgreSQL upgrade. 
```
But perhaps also use consistent terminology when referring to the data directory to help users connect the dots more easily:
```suggestion
The contents of `/var/lib/pgsql/data` is backed up during the PostgreSQL upgrade. 
```

asteflova (changes requested):
@AkshayGadhaveRH It seems you may have lost the changes requested by @ekohl (ifndef and attribute). Please re-check the conversation you have resolved.

guides/common/modules/con_upgrading-prerequisites.adoc
```diff
@@ -16,6 +16,9 @@ Upgrading {SmartProxy} takes approximately 10{range}30 minutes.
 
 * Ensure that you have sufficient storage space on your server.
 For more information, see {InstallingServerDocURL}Preparing_your_Environment_for_Installation_{project-context}[Preparing your Environment for Installation] in _{InstallingServerDocTitle}_ and {InstallingSmartProxyDocURL}preparing-environment-for-capsule-installation[Preparing your Environment for Installation] in _Installing {SmartProxyServer}_.
+* Ensure that you have at least the same amount of free space as that consumed by `/var/lib/pgsql/data`.
+Upgrading to {Project} {ProjectVersion} involves a PostgreSQL 12 to PostgreSQL 13 upgrade.
+The contents of `/var/lib/pgsql/data` is backed up during the PostgreSQL upgrade.
```
Lennonka, line 21:
```suggestion
ifndef::foreman-deb[]
* Ensure that you have at least the same amount of free space as that consumed by `/var/lib/pgsql/data`.
Upgrading to {Project} {ProjectVersion} involves a PostgreSQL 12 to PostgreSQL 13 upgrade.
The contents of `{postgresql-lib-dir}` is backed up during the PostgreSQL upgrade.
endif::[]
```
As requested by Ewoud.

Lennonka (changes requested):
Still missing the condition and attribute for the path.

guides/common/modules/con_upgrading-prerequisites.adoc
```diff
@@ -16,6 +16,11 @@ Upgrading {SmartProxy} takes approximately 10{range}30 minutes.
 
 * Ensure that you have sufficient storage space on your server.
 For more information, see {InstallingServerDocURL}Preparing_your_Environment_for_Installation_{project-context}[Preparing your Environment for Installation] in _{InstallingServerDocTitle}_ and {InstallingSmartProxyDocURL}preparing-environment-for-capsule-installation[Preparing your Environment for Installation] in _Installing {SmartProxyServer}_.
+ifndef::foreman-deb[]
+* Ensure that you have at least the same amount of free space as that consumed by `/var/lib/pgsql/data`.
```
asteflova, line 20: The attribute is useful here too, actually:
```suggestion
* Ensure that you have at least the same amount of free space as that consumed by `{postgresql-lib-dir}/data`.
```

ekohl: There's `postgresql-data-dir` for `{postgresql-lib-dir}/data` already. If you want to be explicit, you could even write
```suggestion
* Ensure that you have at least the same amount of free space on `{postgresql-lib-dir}` as that consumed by `{postgresql-data-dir}`.
```

Otherwise people could check for free space on `/` while they have a `/var/lib/pgsql` partition.

asteflova (approved):
I think we're good here but considering the number of back-and-forths this PR has gone through, I'd appreciate it if someone else could re-review as well and then handle the merging.

guides/common/modules/con_upgrading-prerequisites.adoc
```diff
@@ -16,6 +16,11 @@ Upgrading {SmartProxy} takes approximately 10{range}30 minutes.
 
 * Ensure that you have sufficient storage space on your server.
 For more information, see {InstallingServerDocURL}Preparing_your_Environment_for_Installation_{project-context}[Preparing your Environment for Installation] in _{InstallingServerDocTitle}_ and {InstallingSmartProxyDocURL}preparing-environment-for-capsule-installation[Preparing your Environment for Installation] in _Installing {SmartProxyServer}_.
+ifndef::foreman-deb[]
+* Ensure that you have at least the same amount of free space on `{postgresql-lib-dir}` as that consumed by `{postgresql-data-dir}`.
+Upgrading to {Project} {ProjectVersion} involves a PostgreSQL 12 to PostgreSQL 13 upgrade.
+The contents of `{postgresql-lib-dir}` are backed up during the PostgreSQL upgrade.
```
ekohl, line 22: I think technically it backs up the data dir, not the lib dir. That's which is why you need the same amount of space as the data dir.
```suggestion
The contents of `{postgresql-data-dir}` are backed up during the PostgreSQL upgrade. 
```

ekohl (approved):
@evgeni mind checking since I made suggestions?

evgeni (approved):
ACK on the diff.
Please also pick to 3.11 as that's where the upgrade happens for Foreman/Katello

apinnick:
@AkshayGadhaveRH Rebasing should fix the Vale problem.

Lennonka (approved):
Looking good!

Lennonka:
Cherry picked:
- c5fa1d4ca1..cf5909c951  3.12 -> 3.12
- b51c68b0a0..e3425d7bf5  3.11 -> 3.11

PR #3374, opened 2024-10-15
Mention that the external database should be upgraded

AkshayGadhaveRH:
The external database should be upgraded from PostgreSQL 12 to 13. Mentioning this in the ext DB upgrade chapter.

JIRA: https://issues.redhat.com/browse/SAT-28273

#### Checklists

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/proc_upgrading-a-connected-project-server.adoc
+++ b/guides/common/modules/proc_upgrading-a-connected-project-server.adoc
@@ -46,6 +46,7 @@ ifdef::satellite[]
 # {foreman-maintain} self-upgrade
 ----
 
+. If you are using an external database, upgrade your database to PostgreSQL 13.
 . Use the health check option to determine if the system is ready for upgrade.
 When prompted, enter the hammer admin user credentials to configure `{foreman-maintain}` with hammer credentials.
 These changes are applied to the `/etc/foreman-maintain/foreman-maintain-hammer.yml` file.
@@ -158,6 +159,7 @@ endif::[]
 ----
 # dnf -y module switch-to postgresql:13
 ----
+. If you are using an external database, upgrade your database to PostgreSQL 13.
 . On {EL} 8: Ensure the module streams are enabled:
 +
 [options="nowrap" subs="attributes"]
@@ -183,8 +185,5 @@ endif::[]
 ----
 # {foreman-installer}
 ----
-endif::[]
-ifdef::foreman-el,katello,satellite[]
-. If you are using an external database, upgrade your database to PostgreSQL 13.
 endif::[]
 include::snip_steps-needs-reboot.adoc[]
--- a/guides/common/modules/proc_upgrading-a-connected-project-server.adoc
+++ b/guides/common/modules/proc_upgrading-a-connected-project-server.adoc
@@ -152,13 +152,6 @@ ifdef::katello[]
 # dnf module disable pulpcore
 ----
 endif::[]
-. On {EL} 8: Switch to the PostgreSQL 13 module:
-+
-[options="nowrap" subs="+quotes,attributes"]
-----
-# dnf -y module switch-to postgresql:13
-----
-. If you are using an external database, upgrade to PostgreSQL 13.
 . On {EL} 8: Ensure the module streams are enabled:
 +
 [options="nowrap" subs="attributes"]
@@ -185,7 +178,13 @@ endif::[]
 # {foreman-installer}
 ----
 endif::[]
-ifdef::satellite[]
-. If you are using an external database, upgrade to PostgreSQL 13.
+ifdef::foreman-el,katello,satellite[]
+. On {EL} 8: Switch to the PostgreSQL 13 module:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# dnf -y module switch-to postgresql:13
+----
+. If you are using an external database, upgrade your database to PostgreSQL 13.
 endif::[]
 include::snip_steps-needs-reboot.adoc[]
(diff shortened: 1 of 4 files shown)


Lennonka (changes requested):
Wouldn't it be better as a second prerequisite?
Something like "Install PostgreSQL version 13 on the new RHEL host"?

AkshayGadhaveRH:
@Lennonka agreed. Makes sense to me.

Lennonka (approved):
LGTM

Lennonka:
I've set both "review done" labels because it's correct that PG 13 is needed for 3.12/6.16 and I have reviewed the style as well.

AkshayGadhaveRH:
@asteflova I thought I'd selected `needs tech review`. Sorry, my bad.

evgeni (changes requested):
What we need to tell our users (but do not today) is that they have to upgrade PostgreSQL from 12 to 13 on their external DB server *during* the upgrade to 6.16/3.12(3.11).

We have the "On {EL} 8: Switch to the PostgreSQL 13 module" step -- either before of after that we should say something like "If you're running an external database setup, upgrade the external database to PostgreSQL 13".

guides/common/modules/proc_upgrading-the-external-database.adoc
```diff
@@ -8,6 +8,7 @@ You can upgrade an external database from {EL} 7 to {EL} 8 while upgrading {Proj
 ifdef::katello,orcharhino,satellite[]
 For more information, see {InstallingServerDocURL}using-external-databases_{project-context}[Using External Databases with {Project}].
 endif::[]
+* Install PostgreSQL version 13 on the new {EL} host.
```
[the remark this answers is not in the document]
AkshayGadhaveRH: ack @evgeni so we should be talking about EL8 to 9 upgrade for external DB, right?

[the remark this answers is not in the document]
evgeni: Absolutely.

AkshayGadhaveRH:
@evgeni does this plan look good? (also, small question for you)

1. The documentation should mention that the EL version of the DB should be the same as that of the Satellite.
2. We should mention that the postgresql version should be upgraded for EL8.

> _Will this involve creating backup of old (pgsql12 + RHEL 8) and restoring that on fresh (pgsql13 + EL8)?_

3. The external DB upgrade chapter should talk about EL8 to EL9 upgrade of the external DB OS. RHEL9 already has postgresql by default.

> _Edit existing `proc_upgrading-the-external-database.adoc` module._

evgeni:
1. The OS *should* be updated, but it's not necessary for the upgrade of the main server.
2. The PostgreSQL*must* be upgraded to 13, which is mandatory for the upgrade of the server.
3. We don't document *how* these upgrades are to be done, that's up to the local admin.

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
```diff
@@ -158,6 +158,7 @@ endif::[]
 ----
 # dnf -y module switch-to postgresql:13
 ----
+If you are using an external database, upgrade to PostgreSQL 13.
```
evgeni, line 161: This should be an own step, and it should be rendered for Satellite too (which it currently is not because for Satellite we use foreman-maintain which takes care of all those switches)

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
```diff
@@ -184,4 +185,7 @@ endif::[]
 # {foreman-installer}
 ----
 endif::[]
+ifdef::satellite[]
+. If you are using an external database, upgrade to PostgreSQL 13.
+endif::[]
```
evgeni, line 190: This needs to happen before `satellite-maintain upgrade check`, as that one will already complain about the wrong version.

AkshayGadhaveRH: Should I just move this to the end for both Foreman and Satellite @evgeni ?

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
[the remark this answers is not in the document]
AkshayGadhaveRH: Added steps for disconnected as well.

Lennonka:
Looks good to me.

@evgeni Do we have a technical ack?

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
```diff
@@ -184,4 +178,13 @@ endif::[]
 # {foreman-installer}
 ----
 endif::[]
+ifdef::foreman-el,katello,satellite[]
+. On {EL} 8: Switch to the PostgreSQL 13 module:
```
evgeni, line 182: For Katello/Foreman, this *has* to happen before "Ensure the module streams are enabled" -- otherwise the rest of the upgrade will fail.

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
```diff
@@ -184,4 +178,13 @@ endif::[]
...
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# dnf -y module switch-to postgresql:13
+----
+. If you are using an external database, upgrade your database to PostgreSQL 13.
```
evgeni, line 188: This needs to happen before "Run the installer" for Foreman/Katello, and before "maintain check" for Satellite, otherwise the upgrade will fail.

evgeni (changes requested):

guides/common/modules/proc_upgrading-a-connected-project-server.adoc
evgeni: Twice in the same procedure? No.
But Foreman/Katello and Satellite procedures *are* quite different, and each of them needs that step.

AkshayGadhaveRH:
@Lennonka @evgeni is this good to go?

evgeni (approved):

PR #3376, opened 2024-10-16
Add reproducer for Hammer CLI reference

maximiliankolb:
#### What changes are you introducing?

docs on how to create/update/review Hammer CLI reference files for katello.

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

as a preparation to another PR to add a Hammer CLI reference for latest greatest Katello.

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)

The first and only goal is to create a reproducer for Hammer CLI reference files for Foreman+Katello "develop" with all possible Hammer CLI plugins.

#### Checklists

* [ ] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [ ] Foreman 3.12/Katello 4.14 (Satellite 6.16)
* [ ] Foreman 3.11/Katello 4.13
* [ ] Foreman 3.10/Katello 4.12
* [ ] Foreman 3.9/Katello 4.11 (Satellite 6.15; orcharhino 6.8/6.9/6.10)
* [ ] Foreman 3.8/Katello 4.10
* [ ] Foreman 3.7/Katello 4.9 (Satellite 6.14)
* [ ] Foreman 3.6/Katello 4.8
* [ ] Foreman 3.5/Katello 4.7 (Satellite 6.13; orcharhino 6.6/6.7)
* We do not accept PRs for Foreman older than 3.5.

PR #3377, opened 2024-10-16
Add Hammer CLI reference files for katello

maximiliankolb:

Add Hammer CLI reference files for katello.

They were missing in foreman-documentation before.

Please only review the third commit. Commit one is from https://github.com/theforeman/foreman-documentation/pull/3376 and you can reproduce the second commit by using an Ansible playbook.

**Please note that this PR does not yet contain docs for all Hammer CLI plugins! See https://github.com/theforeman/foreman-documentation/pull/3376#issuecomment-2417028868** 

#### Checklists


maximiliankolb:
Thanks Lena! I propose to keep this PR open until at least Monday, and also to discuss/merge/review/ACK https://github.com/theforeman/foreman-documentation/pull/3376 first.

Lennonka:
@maximiliankolb Do we have to block the reference because of the reproducer? I personally wouldn't block it because you don't have to use the reproducer to produce the reference :))

PR #3382, opened 2024-10-18
Clean up external links in procedure to manage hosts using Cockpit

asteflova:

* Dropping links to Cockpit docs from the list of prerequisites in Managing hosts using Cockpit
* Grouping the remaining external links under Additional resources in order to make them easier to scan

https://issues.redhat.com/browse/SAT-28721 reported that a link to RHEL 9 was missing from the Prerequisites. @adamruzicka suggested to drop the links rather than update the list.

N/A

#### Checklists

* [ ] Foreman 3.6/Katello 4.8


asteflova:
@adamruzicka Can you please take a look to make sure the rest of the section looks okay?

guides/common/modules/proc_managing-and-monitoring-hosts-using-cockpit.adoc
```diff
@@ -5,14 +5,7 @@ You can access the {Cockpit} UI through the {ProjectWebUI} and use the functiona
...
 * You have enabled {the-Cockpit} in {Project}.
-* You have installed {the-Cockpit} on the host that you want to view:
-ifndef::satellite,orcharhino[]
-** For more information, see https://cockpit-project.org/running.html[Running Cockpit].
-endif::[]
-ifdef::satellite[]
-** For {RHEL}{nbsp}8, see {RHELDocsBaseURL}8/html/managing_systems_using_the_rhel_8_web_console/getting-started-with-the-rhel-8-web-console_system-management-using-the-rhel-8-web-console#installing-the-web-console_getting-started-with-the-rhel-8-web-console[Installing the web console] in _Managing systems using the RHEL{nbsp}8 web console_.
-** For {RHEL}{nbsp}7, see {RHELDocsBaseURL}7/html/managing_systems_using_the_rhel_7_web_console/getting-started-with-the-rhel-web-console_system-management-using-the-rhel-7-web-console#installing-the-web-console_getting-started-with-the-web-console[Installing the web console] in _Managing systems using the RHEL{nbsp}7 web console_.
-endif::[]
+* You have installed {the-Cockpit} on the host that you want to view.
```
adamruzicka, line 8: Should this mention foreman ships with a job template to do this?

guides/common/modules/proc_managing-and-monitoring-hosts-using-cockpit.adoc
```diff
@@ -22,8 +15,17 @@ For more information, see xref:Distributing_SSH_Keys_for_Remote_Execution_{conte
 
 You can now access the full range of features available for host monitoring and management, for example, {LoraxCompose}, through {the-Cockpit}.
 
+.Additional resources
+ifndef::satellite,orcharhino[]
+* For more information about installing {the-Cockpit}, see https://cockpit-project.org/running.html[Running Cockpit].
```
adamruzicka, line 20: My issue with the official guide on installing/running cockpit contain a lot of steps which are unnecessary - we just need the packages to be there.

asteflova: I reworded the reference to be as generic as possible and avoid suggesting that users need to do any extra steps to "install" anything. Other than that, we want to help users find more details about Cockpit: using generic wording and keeping the links in the "Additional resources" section should help with that.

Note that I dropped the other set of links from the Prerequisites section, I believe those were the misleading ones.

guides/common/modules/proc_managing-and-monitoring-hosts-using-cockpit.adoc
asteflova: @adamruzicka What is the template called? I can't find it.

adamruzicka: `Service Action - Enable Web Console`, sadly it is only there if the users have the ansible plugin enabled
https://github.com/theforeman/foreman_ansible/blob/master/app/views/foreman_ansible/job_templates/service_action_-_enable_web_console.erb

asteflova: Thanks! I found it. I think it's a good idea to let users know about the template but I'm not sure how to best do it. Let's see what other writers think about adding a note.

guides/common/modules/proc_managing-and-monitoring-hosts-using-cockpit.adoc
```diff
@@ -5,14 +5,15 @@ You can access the {Cockpit} UI through the {ProjectWebUI} and use the functiona
...
-endif::[]
+* You have installed {the-Cockpit} on the host that you want to view.
++
+[NOTE]
+====
+You can use a job template to install {the-Cockpit}.
+To view the template in {ProjectWebUI}, navigate to *Monitor* > *Jobs* and click *Run Job*.
+From the *Job category* list, select *Ansible Services*.
+From the *Job template* list, select *Service Action {endash} Enable Web Console*.
+====
```
asteflova, line 16: I think this might be too much for a prerequisite but then again, I can't think of a better way to point users to the template. Does anyone have suggestions?

maximiliankolb: I propose to move these steps to a procedure to which we can link.

asteflova: I'm not trying to describe a procedure, just the place where users can find the template. Do you have an alternative solution? What I'm trying to do is let users know that this template exists. I can't think of a way to do that without guiding them through the web UI.

asteflova: The procedure itself would be [13. Configuring and setting up remote jobs](https://docs.theforeman.org/nightly/Managing_Hosts/index-katello.html#Configuring_and_Setting_Up_Remote_Jobs_managing-hosts). But how to best point to the particular template to use?

maximiliankolb: I see this as a new 9.2 "Enabling Cockput on hosts": https://docs.theforeman.org/nightly/Managing_Hosts/index-katello.html#host-management-and-monitoring-by-using-cockpit or similar.

The procedure itself does not need to duplicate all steps for REX jobs, but reference a procedure and mention the correct job category+job template.

asteflova: I don't know, I'd still like to avoid adding a whole new procedure. I wouldn't want to set a precedent for creating a whole new module every time I'd like to reference a job template/ansible module/etc. I still hope there is a simpler way to achieve that, one that doesn't eventually contribute to doc bloat :)

How about now? Just a note that the template is there and which category it's included in?

maximiliankolb: Yes, I like this much better.

maximiliankolb (approved):
Two minor optional suggestions. Overall LGTM

asteflova:
Hi @adamruzicka are you okay with the changes now? I cleaned up the links to try to make it clear that we don't link to the Cockpit docs for any installation steps and I added a reference to the job template.

maximiliankolb (approved):

guides/common/modules/proc_managing-and-monitoring-hosts-using-cockpit.adoc
```diff
@@ -22,8 +18,17 @@ For more information, see xref:Distributing_SSH_Keys_for_Remote_Execution_{conte
...
+* For more information about using {LoraxCompose} through {the-Cockpit}, see the following documents:
+** {RHELDocsBaseURL}9/html/composing_a_customized_rhel_system_image/creating-system-images-with-composer-web-console-interface_composing-a-customized-rhel-system-image#accessing-composer-gui-in-the-rhel-8-web-console_creating-system-images-with-composer-web-console-interface[_Accessing Image Builder GUI in the RHEL{nbsp}9 web console_]
+** {RHELDocsBaseURL}8/html/composing_a_customized_rhel_system_image/creating-system-images-with-composer-web-console-interface_composing-a-customized-rhel-system-image#accessing-composer-gui-in-the-rhel-8-web-console_creating-system-images-with-composer-web-console-interface[_Accessing Image Builder GUI in the RHEL{nbsp}8 web console_]
+** {RHELDocsBaseURL}7/html/image_builder_guide/chap-documentation-image_builder-test_chapter_4#sect-Documentation-Image_Builder-Chapter4[_Accessing Image Builder GUI in the RHEL{nbsp}7 web console_]
```
adamruzicka, line 33: nitpick:
When following the rhel7 link, one lands at `Accessing Image Builder GUI in the RHEL 7 web console`, however both rhel 8 and 9 take you to `Accessing the RHEL image builder dashboard in the RHEL web console`, should the link be name exactly as the section of the docs it targets?

asteflova: You are right and this is a good catch, thank you. I updated the RHEL 8 and 9 links to match the respective section headings.

asteflova:
After implementing Adam's latest comment, I realized that a few more tweaks are needed to properly describe the links to RH docs. I tried to find a good way to do that although it's not easy to find a solution that doesn't feel too cumbersome when dealing with three different links per resource.

asteflova:

1f0eadc31c..5afcb65a5c  3.12 -> 3.12

PR #3388, opened 2024-10-23
Add Debian 12 (Bookworm) 

evgeni:

Adding Debian 12 to docs

We now support Debian 12

#### Checklists

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/doc-Release_Notes/topics/foreman.adoc
+++ b/guides/doc-Release_Notes/topics/foreman.adoc
@@ -4,7 +4,14 @@
 [id="foreman-headline-features"]
 == Headline Features
 
+ifndef::foreman-deb[]
 There are no highlights with Foreman {ProjectVersion}.
+endif::[]
+ifdef::foreman-deb[]
+=== Running Foreman on Debian 12 (Bookworm)
+
+Foreman now supports running on Debian 12 (Bookworm).
+endif::[]
 
 [id="foreman-upgrade-warnings"]
 == Upgrade Warnings
--- a/guides/common/modules/con_projectserver-operating-system.adoc
+++ b/guides/common/modules/con_projectserver-operating-system.adoc
@@ -1,7 +1,7 @@
 [id="ProjectServer-Operating-System_{context}"]
 = {ProjectServer} operating system
 
-{Project} has packages for {EL} 8 and 9, Debian 11 and Ubuntu 22.04.
+{Project} has packages for {EL} 8 and 9, Debian 11, Debian 12 and Ubuntu 22.04.
 Katello plugin packages, which provide content management capabilities, are only available for {EL}.
 
 {Team} only packages {Project} for x86_64.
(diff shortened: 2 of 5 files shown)


evgeni:
Technically the "drop extra heading" commit is an unrelated bugfix that could (should?) be picked to older branches too.

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -4,7 +4,14 @@
 [id="foreman-headline-features"]
 == Headline Features
 
+ifndef::foreman-deb[]
 There are no highlights with Foreman {ProjectVersion}.
+endif::[]
+ifdef::foreman-deb[]
+=== Running Foreman on Debian 12 (Bookworm)
+
+Foreman now supports running on Debian 12 (Bookworm).
```
ekohl, line 13: I'd like to support it with Foreman 3.12 now that puppetserver is available to give users a better upgrade path. You already made sure we have packages so it should be easy.

ekohl (commented):
I'm wondering about Debian 11 removal. Would it be too aggressive if we drop it with 3.13 already?

guides/doc-Release_Notes/topics/foreman.adoc
evgeni: should this be read: "drop the release note here and add it in 3.12" or "keep it here but *also* pick it to 3.12"?

evgeni:
> I'm wondering about Debian 11 removal. Would it be too aggressive if we drop it with 3.13 already?

I think that's too aggressive, as I don't expect people to re-read release notes on .z releases *and* this would shorten the time between announcement and removal of support to just ~2 months?

guides/doc-Release_Notes/topics/foreman.adoc
ekohl: I'm leaning to the latter: introduce it as a headline feature here, but also note it for the future patch releases. That way users who read them as a patch release know they can benefit from it. My reasoning behind that is that I wouldn't read headline features when I upgrade a patch release because I'd assume they don't change.

evgeni: Sounds like no changes are needed from me, cool!

ekohl (approved):

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -4,7 +4,14 @@
 [id="foreman-headline-features"]
 == Headline Features
 
+ifndef::foreman-deb[]
 There are no highlights with Foreman {ProjectVersion}.
+endif::[]
+ifdef::foreman-deb[]
+=== Running Foreman on Debian 12 (Bookworm)
```
Lennonka, line 11:
```suggestion
=== Running Foreman on Debian 12
```
I'd drop this detail from the heading because it's repetitive, but it isn't a blocker. Feel free to disagree :)

Lennonka (approved):
Looks good to me!

evgeni:
Playing the "disagree" card, mostly to keep consistency with other places (also outside docs.theforeman.org) where we "always" write "Debian X (Release Name)"

PR #3408, opened 2024-10-31
Document high-level steps for upgrading

asteflova:

* Adding a high-level summary of the upgrade process loosely based on [Upgrade paths](https://docs.theforeman.org/3.3/Upgrading_and_Updating/index-katello.html#upgrade_paths) in version 3.3. ~~("Loosely" because rather than re-create the original list of steps, I rather propose to add a very general summary in its place and include the other, more detailed steps as part of the assembly that actually describes upgrading.)~~
* Removing the introduction for the upgrade assembly (`guides/common/modules/con_upgrading-project.adoc`) because it's out-of-date.
* Enhancing the upgrade procedures with a few bits of information based on the old Upgrade paths section and the removed assembly introduction, to make sure we still keep the important parts.

https://issues.redhat.com/browse/SAT-29113 points out that the high-level upgrade steps are missing in the latest version and requests adding it.

~~If I get an ack for this, I'd like to open another PR for 3.11 and below, but without the part about upgrading to EL 9.~~

#### Checklists

* [ ] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/modules/con_upgrade-paths.adoc
+++ b/guides/common/modules/con_upgrade-paths.adoc
@@ -2,14 +2,49 @@
 == Upgrade paths
 
 You can upgrade to {ProjectName} {ProjectVersion} from {ProjectName} {ProjectVersionPrevious}.
+The high-level steps in upgrading {Project} to {ProjectVersion} are as follows:
 
 ifdef::satellite[]
-{ProjectServer}s and {SmartProxyServers} on earlier versions must first be upgraded to {Project} {ProjectVersionPrevious}.
+. Ensure that your {ProjectServer}s and {SmartProxyServers} have been upgraded to {Project} {ProjectVersionPrevious}.
 For more information, see {UpgradingPreviousDocURL}[_{UpgradingPreviousDocTitle}_] or {UpgradingDisconnectedPreviousDocURL}[_{UpgradingDisconnectedPreviousDocTitle}_].
 endif::[]
 
-Upgrading {Project} to {ProjectVersion} includes upgrading your {ProjectServer} and all {SmartProxyServers} to {ProjectVersion}.
 ifndef::foreman-deb[]
-After the upgrade, you can also upgrade the operating system of your {ProjectServer} and {SmartProxyServers} to {EL}{nbsp}9.
+. Upgrade your {ProjectServer}:
+.. Upgrade your {ProjectServer} to {ProjectVersion}.
+.. Optional: Upgrade the operating system on your {ProjectServer} to {EL}{nbsp}9.
+ifdef::satellite[]
+.. Synchronize the new {ProjectVersion} repositories.
 endif::[]
-For more information on the steps of the upgrade process, see xref:Upgrading_{project-context}_{context}[].
+. Upgrade your {SmartProxyServers}:
+.. Upgrade all {SmartProxyServers} to {ProjectVersion}.
+.. Optional: Upgrade the operating system on your {SmartProxyServers} to {EL}{nbsp}9.
+endif::[]
+
+ifdef::foreman-deb[]
+. Upgrade your {ProjectServer} to {ProjectVersion}.
+. Upgrade all {SmartProxyServers} to {ProjectVersion}.
+endif::[]
+
+ifndef::foreman-deb[]
+[NOTE]
+====
+Although upgrading the operating system of your {ProjectServer} and {SmartProxyServers} to {EL}{nbsp}9 is an optional part of the upgrade process, you will need to upgrade to {EL}{nbsp}9 before you can upgrade to the next {Project} version after {ProjectVersion}.
+====
+endif::[]
+
+{SmartProxies} at version {ProjectVersionPrevious} will keep working with your upgraded {ProjectServer} {ProjectVersion}.
+After you upgrade {ProjectServer} to {ProjectVersion}, you can upgrade your {SmartProxies} separately over multiple maintenance windows.
+ifdef::foreman-el,katello,satellite[]
+For more information, see xref:Upgrading_Proxies_Separately_from_Server_{context}[].
+endif::[]
+
+{Project} services are shut down during the upgrade. Make sure to plan for the required downtime.
+The upgrade process duration might vary depending on your hardware configuration, network speed, and the amount of data that is stored on the server.
+
+* Upgrading {ProjectServer} takes approximately 1{range}2 hours.
+* Upgrading {SmartProxyServer} takes approximately 10{range}30 minutes.
+
+.Additional resources
+
+* For more information on the steps of the upgrade process, see xref:Upgrading_{project-context}_{context}[].
--- a/guides/common/modules/con_upgrade-paths.adoc
+++ b/guides/common/modules/con_upgrade-paths.adoc
@@ -13,24 +13,28 @@ ifndef::foreman-deb[]
 . Upgrade your {ProjectServer}:
 .. Upgrade your {ProjectServer} to {ProjectVersion}.
 .. Optional: Upgrade the operating system on your {ProjectServer} to {EL}{nbsp}9.
++
+[NOTE]
+====
+Although upgrading the operating system of your {ProjectServer} to {EL}{nbsp}9 is optional, you will need to do it before you can upgrade to the next {Project} version after {ProjectVersion}.
+====
++
 ifdef::satellite[]
 .. Synchronize the new {ProjectVersion} repositories.
 endif::[]
 . Upgrade your {SmartProxyServers}:
 .. Upgrade all {SmartProxyServers} to {ProjectVersion}.
 .. Optional: Upgrade the operating system on your {SmartProxyServers} to {EL}{nbsp}9.
++
+[NOTE]
+====
+Although upgrading the operating system of your {SmartProxyServers} to {EL}{nbsp}9 is optional, you will need to do it before you can upgrade to the next {Project} version after {ProjectVersion}.
+====
 endif::[]
 
 ifdef::foreman-deb[]
 . Upgrade your {ProjectServer} to {ProjectVersion}.
 . Upgrade all {SmartProxyServers} to {ProjectVersion}.
-endif::[]
-
-ifndef::foreman-deb[]
-[NOTE]
-====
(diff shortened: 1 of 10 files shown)


guides/common/modules/con_upgrade-paths.adoc
```diff
@@ -8,13 +8,8 @@ ifdef::satellite[]
 For more information, see {UpgradingPreviousDocURL}[_{UpgradingPreviousDocTitle}_] or {UpgradingDisconnectedPreviousDocURL}[_{UpgradingDisconnectedPreviousDocTitle}_].
 endif::[]
 
-.High-level upgrade steps
-
-The high-level steps in upgrading {Project} to {ProjectVersion} are as follows:
-
-. Upgrade {ProjectServer} to {ProjectVersion}.
-For more information, see xref:Project_Upgrade_Considerations_{context}[].
-. Upgrade all {SmartProxyServers} to {ProjectVersion}.
```
asteflova, line 17: When the high-level overview was formatted as a numbered list, it gave the impression that it represented a complete list of steps required for an upgrade. That was certainly not the case and I don't think we should even try to give the complete list here (that would duplicate the actual upgrade chapter). Rewording the list into a normal para a suggested below reduces the risk of misunderstanding and will make maintaining this section easier.

guides/common/modules/snip_prerequisites-upgrading-project-server.adoc
```diff
@@ -12,6 +13,10 @@ Cloning is the recommended method because that prevents them being overwritten i
 To confirm if a template has been edited, you can view its *History* before you upgrade or view the changes in the audit log after an upgrade.
 In the {ProjectWebUI}, navigate to *Monitor* > *Audits* and search for the template to see a record of changes made.
 If you use the export method, restore your changes by comparing the exported template and the default template, manually applying your changes.
+ifdef::satellite[]
+* Optional: Clone your {ProjectServer} to test the upgrade.
```
asteflova, line 17: I'm not sure if testing the upgrade really is the only purpose of cloning the server but it's what the corresponding chapter in Administering Project suggests.

guides/common/modules/con_upgrade-paths.adoc
ekohl: I'm not sure this is better. I think the previous text tried to make it clear that you would first upgrade your `{ProjectServer}` and then you would upgrade all your proxy servers, but that maybe wasn't obvious. It may also not have been obvious that you can spread those steps out over multiple maintenance windows since there is n-1 support.

Your new text puts it in a single sentence which to me implies it is a single operation in a single maintenance window.

I think it's something like:

* Upgrade `{ProjectServer}` to `{ProjectVersion}`
  * If on EL8, at some point before upgrading to the next version: migrate to EL9
* For every `{SmartProxyServer}: find a nice maintenance window and upgrade it to `{ProjectVersion}`. Has to be completed before upgrading `{ProjectServer}` to the next version
  * If on EL8, at some point before upgrading to the next version: migrate to EL9

Now that probably isn't the most user friendly description, but I hope you at least understand it and can work it into something consumable.

guides/common/modules/snip_prerequisites-upgrading-project-server.adoc
ekohl: Can be both. I think we also support upgrades via cloning, but normally I look at @evgeni for those questions.

guides/common/modules/con_upgrade-paths.adoc
ekohl: To add to this, in the resulting document you have an overview that states (for Katello):

> Review prerequisites and available upgrade paths below before upgrading your current Foreman installation to Foreman nightly.
>
> Note that you can upgrade Smart Proxies separately from Foreman. For more information, see [Upgrading Smart Proxies separately from Foreman](https://theforeman-foreman-documentation-preview-pr-3408.surge.sh/nightly/Upgrading_Project/index-katello.html#Upgrading_Proxies_Separately_from_Server_upgrading-connected).

Then there's an Upgrade paths section. That last sentence is something that IMHO belongs in the Upgrade paths section, but then the overview is very short. Should they be combined?

asteflova: Well, the whole thing makes a lot more sense to me now. Thanks! I'll keep working on it.

asteflova: > To add to this, in the resulting document you have an overview that states (for Katello):
> 
> > Review prerequisites and available upgrade paths below before upgrading your current Foreman installation to Foreman nightly.
> > Note that you can upgrade Smart Proxies separately from Foreman. For more information, see [Upgrading Smart Proxies separately from Foreman](https://theforeman-foreman-documentation-preview-pr-3408.surge.sh/nightly/Upgrading_Project/index-katello.html#Upgrading_Proxies_Separately_from_Server_upgrading-connected).
> 
> Then there's an Upgrade paths section. That last sentence is something that IMHO belongs in the Upgrade paths section, but then the overview is very short. Should they be combined?

Yes. Similarly, there is a prerequisite a bit further along that describes how long the upgrade takes and that users need to properly plan for it. That also seems closely related and might work better together with the other related information in Upgrade paths.

guides/common/modules/snip_prerequisites-upgrading-project-server.adoc
asteflova: To give a bit of context: I'm adding this because the [3.3 version of the document](https://docs.theforeman.org/3.3/Upgrading_and_Updating/index-satellite.html#upgrade_paths) states cloning the server as the first step. But I want to add a reason for doing it, especially since by making it optional we are asking users to decide whether they want to do it.

asteflova:
I reworked the Upgrade paths section to give an overview of the upgrade process, outlining that the upgrade can be executed in multiple phases. I'll appreciate honest feedback about whether I'm making things better or worse.

guides/common/modules/con_upgrade-paths.adoc
```diff
@@ -2,19 +2,49 @@
...
+ifndef::foreman-deb[]
+. Upgrade your {ProjectServer}:
+.. Upgrade your {ProjectServer} to {ProjectVersion}.
+.. Optional: Upgrade the operating system on your {ProjectServer} to {EL}{nbsp}9.
+ifdef::satellite[]
+.. Synchronize the new {ProjectVersion} repositories.
+endif::[]
+. Upgrade your {SmartProxyServers}:
+.. Upgrade all {SmartProxyServers} to {ProjectVersion}.
+.. Optional: Upgrade the operating system on your {SmartProxyServers} to {EL}{nbsp}9.
```
ekohl, line 21: Should these link to the specific sections? Then you can drop the "Additional resources" part at the end.

ekohl (commented):
I think it's an improvement. I'd welcome @apinnick's thoughts on it as well. It's very tempting to continue restructuring, but we need to draw the line somewhere.

For example, I'd say section 1.2 Prerequisites still feels odd. While I certainly think that we should mention reading the release notes is needed, the disk space prerequisites probably belong in the specific upgrade sections. Also, the "plan for updating any scripts" is less of a prerequisite and more of a planning step.

guides/common/modules/con_upgrade-paths.adoc
asteflova: I'd rather not. What I dislike about the [3.3 version of Upgrade paths](https://docs.theforeman.org/3.3/Upgrading_and_Updating/index-katello.html#upgrade_paths) is that it basically creates a whole new, but incomplete table of contents for how to upgrade. I'm concerned that if we include links here, users might then follow only the linked sections and skip or overlook the others.

What I'm going for here is a high-level overview without links, just to prepare users for what they should expect in general terms. And then if anyone wants to follow the link in Additional resources, they will be taken to the beginning of the upgrade chapter, which reduces the risk of them skipping parts of the process.

asteflova:
> I'd welcome @apinnick's thoughts on it as well. It's very tempting to continue restructuring, but we need to draw the line somewhere.
> 
> For example, I'd say section 1.2 Prerequisites still feels odd. While I certainly think that we should mention reading the release notes is needed, the disk space prerequisites probably belong in the specific upgrade sections. Also, the "plan for updating any scripts" is less of a prerequisite and more of a planning step.

I agree that there are opportunities for additional improvements but I am trying very hard to ignore them. I need to limit this PR to describing the high-level overview of the upgrade process and also get it done in a timely manner.

asteflova:
I requested @apinnick's review as suggested above. Avital, as you review the PR, can you please think about which changes need to be addressed for this PR to be good enough to merge and which changes could perhaps be postponed for later? Perhaps it might be a good idea to create an epic similar to the one you created for the installation guide.

ekohl:
That's precisely my thinking. You can probably spend months on improving this document, but it's good to deliver some smaller improvements in time for 6.16.

guides/common/modules/con_upgrade-paths.adoc
asteflova: However, it might help if I move the link to the introduction. If my goal is to make sure users don't overlook it, it's probably a reasonable thing to do.

apinnick (approved):
I think this is good enough for current documentation, so I am approving it.

At the same time, I also have reservations about the prerequisites section, partly because it combines what I consider "planning considerations" with technical prerequisites. 

In addition, I find the relationship between "1.2 Prerequisites" and "2.1 Sat Server upgrade considerations" rather murky. I thought I created a Jira issue to fix this. Will check.

asteflova:
@ekohl and/or @evgeni, do you have any further comments?

guides/common/modules/snip_prerequisites-upgrading-project-server.adoc
evgeni: The user can certainly do both:
* create a clone, use that for testing the upgrade, then upgrade the original box and throw away the clone
* create a clone, upgrade it, test it, promote it to "the real thing" and throw away the original

asteflova: Thanks! I added a few words to describe the "why". I'm pretty sure this is not the best place to add this information but my immediate goal is to add a step about cloning + explain when & why to clone so I think this is good enough for now. Tagging @apinnick in case she wants to look into this when planning the upgrade guide improvements.

ekohl:
Like others I think this is a good step forward and appreciate that we've also identified future work

asteflova:

8536dca72c..639de86b30  3.12 -> 3.12

PR #3411, opened 2024-11-01
Salt repo update

asteflova:

Replacing links to repo.saltproject.io with up-to-date links and dropping parts of the obsolete Salt content.

The old URL for Salt package repositories is no longer available, which breaks quite a few things on our side. More details in https://saltproject.io/blog/salt-project-package-repo-migration-and-guidance/.

@maximiliankolb was kind enough to provide a patch to take my PR further than just updating the broken links.

~~I'm not sure at all whether the new links are the right ones, all I can verify is that they work. Whether they actually take users where they need to go is a different matter.~~

#### Checklists

* [x] Foreman 3.6/Katello 4.8

### DIFF
--- a/guides/common/attributes-base.adoc
+++ b/guides/common/attributes-base.adoc
@@ -187,7 +187,7 @@
 :client-content-content-view-label: AlmaLinux_9
 :client-content-product-label: {client-content-content-view-label}
 :client-content-repository-label: BaseOS
-:client-salt-minion-repository-url: https://repo.saltproject.io/py3/
+:client-salt-minion-repository-url: https://packages.broadcom.com/ui/repos/tree/General/saltproject-rpm
 // Foreman Server and Smart Proxy Server
 :project-minimum-memory: 4 GB
 :project-package-check-update: dnf check-update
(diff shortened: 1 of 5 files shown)


ekohl:
> I know that it's a workaround, but that's the best I can do right now. According to Pulp Devs, the Yum repo by Broadcom is not a valid Yum repo to pulp_rpm which means that we cannot only update the URLs.

That's nasty, but I have little faith in Broadcom to fix this.

guides/common/attributes-base.adoc
```diff
@@ -187,7 +187,7 @@
 :client-content-content-view-label: AlmaLinux_9
 :client-content-product-label: {client-content-content-view-label}
 :client-content-repository-label: BaseOS
-:client-salt-minion-repository-url: https://repo.saltproject.io/py3/
+:client-salt-minion-repository-url: https://packages.broadcom.com/ui/repos/tree/General/saltproject-rpm
```
maximiliankolb, line 190: My patch was wrong, sorry for that. I believe that the attribute is indeed unused, you were right the first time.

asteflova: It was used in `guides/common/modules/proc_providing-the-salt-client-to-salt-minions.adoc` but your patch deleted it. So I dropped it from here too.

maximiliankolb (approved):
Thanks Anet, diff LGTM!

asteflova:
The tech ack was implied in the conversation so I believe we have that covered as well.

asteflova:
Thank you @maximiliankolb!

PR #3439, opened 2024-11-07
Add release for 3.13-rc1

ColeHiggins2:


ekohl (approved):
Kind of hard to pick headline features based on this. I'd also like to clean up the issues, but we can do that for RC2.

PR #3450, opened 2024-11-13
Fix selection diagram based on whether Foreman manages DHCP

Lennonka:

Update to the provisioning method selection.
Changing the choices of network configuration based on whether Foreman manages DHCP

If user's Foreman does not manage DHCP, the user _must_ use full-host bootdisk provisioning because the other installer-based methods require DHCP to be managed by Foreman.

It does not depend on whether the IP address is static or not.

#### Checklists

* [ ] Foreman 3.13/Katello 4.15 (Satellite 6.17)
* [ ] Foreman 3.11/Katello 4.13 (orcharhino 6.11 on EL8 only)
* [ ] Foreman 3.6/Katello 4.8


Lennonka:
@stejskalleos Technical ack, please? :)

stejskalleos:
I'm busy with wrapping up release work + I'm preparing for Cfgmngt camp 25, so after that I'll take a look.

stejskalleos (changes requested):
Do we have to mention the `Installer-based provisioning` part in the `Network configuration section`? If not, what else would it be?

Also, for the image-based provisioning, we still use the installer in the image, so technically it should be `Installer-based Image-based provisioning`, which sounds weird, but it's precisely that.

IMO, we should not mention the `installer-based provisioning` part

Lennonka:
I've rephrased it from "Foreman manages DHCP" to "Foreman configures DHCP".

stejskalleos (approved):

maximiliankolb (approved):

PR #3452, opened 2024-11-13
Deprecate running on Debian 11

ekohl:

Add a release note that Foreman 3.14 will drop the support for running on Debian 11.

I still need to open an RFC for this, but this allows RC1 to be announced with it already.

This is only for 3.13.

#### Checklists


evgeni (approved):
The linkchecker suggests we still link to el8 somewhere, but that's unrelated to the change at hand

ColeHiggins2 (approved):
ack

Issue #190, opened 2020-06-03
Avoid /etc/foreman-installer/scenarios.d


ekohl:
This directory will be moved in the future (probably `/var/lib/foreman-installer`) to signify the user is not supposed to modify these files. The way I look at it is that it's similar to suggesting open a database console and `SELECT * FROM settings` rather than going to the settings page.

Would it be better to suggest `foreman-installer --full-help` and let users look at `(default: $VALUE)` instead? The benefit is that it's not scenario specific and always the same command. Note that `--scenario $scenario` only needs to be passed on the first installation. After that a scenario has been selected and will always be used.

maximiliankolb:
somewhat related: https://projects.theforeman.org/issues/36957

maximiliankolb:
triage: let's do this soon.

PR #3438, opened 2024-11-07
Retire branches older than 3.7

Lennonka:
#### What changes are you introducing?

Retiring branches older than 3.7

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

Because [Satellite 6.13 becomes EOL on November 30, 2024](https://access.redhat.com/support/policy/updates/satellite/)

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)

@maximiliankolb orcharhino okay with that?

#### Checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into: N/A

### Merge after November 30!

PR #3484, opened 2024-12-03
Avoid or clarify /etc/foreman-installer/scenarios.d references

asteflova:

Removing one reference to `scenarios.d/` and clarifying two others in a way that (hopefully) ensures that users will not update its contents.

EDIT: One of the references to `scenarios.d/` was resolved separately in https://github.com/theforeman/foreman-documentation/pull/3486.

https://github.com/theforeman/foreman-documentation/issues/190

I couldn't find a way to completely avoid mentioning scenarios.d/ in all of the existing occurrences.

#### Checklists

* [x] Foreman 3.11/Katello 4.13 (orcharhino 6.11 on EL8 only)

### DIFF
--- a/guides/common/modules/proc_resetting-custom-ssl-certificate-to-default-self-signed-certificate-on-project.adoc
+++ b/guides/common/modules/proc_resetting-custom-ssl-certificate-to-default-self-signed-certificate-on-project.adoc
@@ -11,10 +11,10 @@
 
 .Verification
 ifndef::orcharhino[]
-Ensure the following parameters in `/etc/foreman-installer/scenarios.d/{project-context}-answers.yaml` have no values:
+Verify that the following parameters in `/etc/foreman-installer/scenarios.d/{project-context}-answers.yaml` have no values:
 endif::[]
 ifdef::orcharhino[]
-Ensure the following parameters in `/etc/foreman-installer/scenarios.d/katello-answers.yaml` have no values:
+Verify that the following parameters in `/etc/foreman-installer/scenarios.d/katello-answers.yaml` have no values:
 endif::[]
 
 * server_cert:
--- a/guides/common/modules/ref_glossary-of-terms-used-in-project.adoc
+++ b/guides/common/modules/ref_glossary-of-terms-used-in-project.adoc
@@ -42,6 +42,7 @@ For more information about automating {Project} using {Project} Ansible collecti
 [[Answer_file]]
 Answer file:: A configuration file that defines settings for an installation scenario.
 Answer files are defined in the YAML format and stored in the `/etc/foreman-installer/scenarios.d/` directory.
+To see the default values for installation scenario parameters, use the `{foreman-installer} --full-help` command on your {ProjectServer}.
 
 [[ARF_report]]
 ARF report::


guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
```diff
@@ -75,10 +75,26 @@ To update the certificates on a currently running Katello installation, run:
 endif::[]
 +
 Note that you must not access or modify `/root/ssl-build`.
-. From the output of the `katello-certs-check` command, depending on your requirements, enter the `{foreman-installer}` command that installs a new {Project} with custom SSL certificates or updates certificates on a currently running {Project}.
+. Run the command to update certificates on your {ProjectServer}:
 +
-If you are unsure which command to run, you can verify that {Project} is installed by checking if the file `/etc/foreman-installer/scenarios.d/.installed` exists.
```
asteflova, line 80: At this point in the process, we expect users to have an existing server installation. So I don't think we need the reference for how to check whether a server is installed.

guides/common/modules/proc_resetting-custom-ssl-certificate-to-default-self-signed-certificate-on-project.adoc
```diff
@@ -11,10 +11,10 @@
 
 .Verification
 ifndef::orcharhino[]
-Ensure the following parameters in `/etc/foreman-installer/scenarios.d/{project-context}-answers.yaml` have no values:
```
asteflova, line 14: I checked `--full-help` and didn't find "current" values for these parameters. Only the default ones but that's not what we want users to check here. However, I believe "Ensure" is not the right action verb to use here; rather than encourage users to delete the values in the file, we want them to make sure that no values are currently set.

guides/common/modules/ref_glossary-of-terms-used-in-project.adoc
```diff
@@ -42,6 +42,7 @@ For more information about automating {Project} using {Project} Ansible collecti
 [[Answer_file]]
 Answer file:: A configuration file that defines settings for an installation scenario.
 Answer files are defined in the YAML format and stored in the `/etc/foreman-installer/scenarios.d/` directory.
```
asteflova, line 44: This `scenarios.d` reference seems safe enough to keep.

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
```diff
@@ -75,10 +75,27 @@ To update the certificates on a currently running Katello installation, run:
...
-If the file exists, run the second `{foreman-installer}` command that updates certificates.
+ifdef::satellite[]
+[options="nowrap", subs="+quotes,attributes"]
+----
+# {foreman-installer} --scenario satellite \
```
ekohl, line 83: If you have an existing installation, the `--scenario` option is not needed because the installer will remember the last used scenario. As you stated earlier, we expect the user to have an existing installation.

Taking a step back and looking at the whole procedure, I'd suggest to drastically simplify it. The installer has [a hook](https://github.com/theforeman/foreman-installer/blob/develop/hooks/pre_commit/20-certs_update.rb) that runs `katello-certs-check` itself, so we can actually drop step 1 altogether.

At that point, the instructions won't need the `ifdef` at all and overall is much shorter.

As for cherry picking, https://projects.theforeman.org/issues/36567 was a significant bug so I'd take 3.6 as the lowest version that's safe. After https://github.com/theforeman/foreman-documentation/pull/3438 is merged, that shouldn't be a problem then.

That's perhaps a change that should be made before this in its own PR.

ekohl (commented):
I appreciate you picking this up.

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-satellite-server.adoc
asteflova: Ok, thanks! Let's try to look at this in https://github.com/theforeman/foreman-documentation/pull/3486

maximiliankolb (commented):
Reviewed 2/3 files which LGTM. Maybe we can merge https://github.com/theforeman/foreman-documentation/pull/3486 first? I'd be happy to re-review after you rebase.

asteflova:
https://github.com/theforeman/foreman-documentation/pull/3486 has been merged and I rebased, which takes care of one of the references to `scenarios.d/` that I set out to fix. That's no longer a part of this PR.

I'll set `tech review done` because I believe all of @ekohl's is covered for the remaining two `scenarios.d/` references. In the end, I'm actually not removing either of them, just clarifying the context and wording.

@maximiliankolb Can you please re-review?

maximiliankolb (approved):
LGTM

asteflova:

* c17ac39d59..0a15d580fd  3.13 -> 3.13
* c444a9b252..8c6432a093  3.12 -> 3.12
* f8684fb9ee..563e291422  3.11 -> 3.11
* 23cc9ab98f..0a1f7e38c8  3.10 -> 3.10

Conflicts start on 3.9 because the content looks a bit different there. I checked and concluded that we can leave it as.

PR #2145, opened 2023-04-13
Add SecureBoot support for arbitrary operating systems to "Grub2 UEFI" PXE loaders

jloeser:
This adds a section about new SecureBoot support. It only works in combination with the following patchset:

https://github.com/theforeman/foreman/pull/9864
https://github.com/theforeman/smart-proxy/pull/877

RFC: https://community.theforeman.org/t/add-secureboot-support-for-arbitrary-distributions/32601/1

PR #3488, opened 2024-12-04
Add Secure Boot for Satellite and RHEL

Lennonka:

Adding Secure Boot support for Satellite and RHEL.

[SAT-23035](https://issues.redhat.com/browse/SAT-23035) Epic

- Please, take a look at the preview.
- I'm trying to reuse the module from #2145, but I'm not entirely sure it fits.

#### Checklists

Please cherry-pick my commits into: N/A - for 3.14

### DIFF
--- a/guides/common/assembly_using-network-boot-to-provision-hosts.adoc
+++ b/guides/common/assembly_using-network-boot-to-provision-hosts.adoc
@@ -122,7 +122,6 @@ endif::[]
 :!shim_efi_downloaded_package_name:
 :!shim_efi_package_name:
 :!shim_efi_tmp_binary_path:
-endif::[]
 
 include::modules/proc_creating-hosts-with-pxe-boot-provisioning.adoc[leveloffset=+1]
 
--- a/guides/common/assembly_using-network-boot-to-provision-hosts.adoc
+++ b/guides/common/assembly_using-network-boot-to-provision-hosts.adoc
@@ -12,7 +12,6 @@ include::modules/ref_required-boot-order-for-network-boot.adoc[leveloffset=+1]
 
 include::modules/con_configuring-smartproxy-for-secure-boot.adoc[leveloffset=+1]
 
-ifndef::satellite[]
 :extract_deb_prefix: cd /tmp && ar x /tmp
 :extract_deb_xz_suffix: && tar -xf data.tar.xz && cd -
 :extract_deb_zst_suffix: && tar --use-compress-program=unzstd -xf data.tar.zst && cd -
@@ -22,6 +21,7 @@ ifndef::satellite[]
 :parent-client-pkg-ext: {client-pkg-ext}
 :secureboot-os-name: My_Operating_System_In_Lowercase
 
+ifndef::satellite[]
 :client-os-context: almalinux
 :client-os: AlmaLinux
 :client-pkg-ext: rpm
--- a/guides/common/assembly_using-network-boot-to-provision-hosts.adoc
+++ b/guides/common/assembly_using-network-boot-to-provision-hosts.adoc
@@ -66,7 +66,24 @@ include::modules/proc_configuring-smart-proxy-to-provision-secure-boot-enabled-h
 :extract_grub: {extract_rpm_prefix}/{grub_efi_downloaded_package_name} {extract_rpm_suffix}
 :extract_shim: {extract_rpm_prefix}/{shim_efi_downloaded_package_name} {extract_rpm_suffix}
 include::modules/proc_configuring-smart-proxy-to-provision-secure-boot-enabled-hosts.adoc[leveloffset=+2]
+endif::[]
 
+:client-os-context: rhel
+:client-os: {RHEL}
+:client-pkg-ext: rpm
+:grub_efi_download_url: https://access.redhat.com/downloads/content/package-browser[Package browser] on the Red{nbsp}Hat Customer Portal
+:grub_efi_downloaded_package_name: grub2-efi-x64.rpm
+:grub_efi_package_name: grub2-efi-x64
+:grub_efi_tmp_binary_path: /tmp/boot/efi/EFI/{client-os-context}/grubx64.efi
+:shim_efi_download_url: https://access.redhat.com/downloads/content/package-browser[Package browser] on the Red{nbsp}Hat Customer Portal
+:shim_efi_downloaded_package_name: shim-x64.rpm
+:shim_efi_package_name: shim-x64
+:shim_efi_tmp_binary_path: /tmp/boot/efi/EFI/{client-os-context}/shimx64.efi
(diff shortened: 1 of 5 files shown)


Lennonka:
Rebased.

maximiliankolb (commented):
diff and rendered docs LGTM.

The only minor improvement that I see is "_My_Operating_System_in_Lowercase_". Maybe you want to introduce an attribute for that? Your call.
I am currently unsure why we didn't do that in the first place. If you're interested, I could ping people that implemented this in the beginning of January.

guides/common/assembly_using-pxe-to-provision-hosts.adoc
```diff
@@ -21,6 +20,7 @@ ifndef::satellite[]
 :parent-client-pkg-ext: {client-pkg-ext}
 :secureboot-os-name: My_Operating_System_In_Lowercase
```
Lennonka, line 21: @maximiliankolb There actually is an attribute for that but it isn't being modified anywhere :laughing: 
Perhaps it's a good idea to request a PR to do that for the other OSs.
I'm not sure what the value is supposed to be, e.g. is `rhel` good enough? Does the directory name even matter as long as it's consistent in the whole procedure?

maximiliankolb: Yeah, let's keep it consistent for now, and then think/ask if we can overwrite this attribute for each OS in a follow-up PR.

nofaralfasi (commented):
If I understand correctly, this PR adds instructions on configuring Secure Boot for other OS versions. However, we decided this is no longer necessary, as the same GRUB and shim can be used across multiple RHEL versions.
cc: @stejskalleos

guides/common/modules/con_using-pxe-to-provision-hosts.adoc
```diff
@@ -43,22 +43,23 @@ In {Project} provisioning, the PXE loader option defines the DHCP `filename` opt
 * For BIOS systems, select the *PXELinux BIOS* option to enable a provisioned host to download the `pxelinux.0` file over TFTP.
 * For UEFI systems, select the *Grub2 UEFI* option to enable a TFTP client to download `grubx64.efi` file, or select the *Grub2 UEFI HTTP* option to enable an UEFI HTTP client to download `grubx64.efi` with the HTTP Boot feature.
 
-ifndef::satellite[]
 {ProjectName} supports UEFI Secure Boot.
 SecureBoot PXE loaders enable a client to download the `shim.efi` bootstrap boot loader that then loads the signed `grubx64.efi`.
 Use the *Grub2 UEFI SecureBoot* PXE loader for PXE-boot provisioning or *Grub2 UEFI HTTPS SecureBoot* for HTTP-boot provisioning.
```
Lennonka, line 48: @nofaralfasi So I'm assuming that these lines are valid for Satellite and I will have to enable them.

guides/common/modules/con_using-pxe-to-provision-hosts.adoc
```diff
@@ -43,22 +43,23 @@ In {Project} provisioning, the PXE loader option defines the DHCP `filename` opt
...
 
 By default, you can provision operating systems from the vendor of the operating system of your {ProjectServer} on Secure Boot enabled hosts.
 To provision operating systems on Secure Boot enabled hosts from different vendors, you have to provide signed shim and GRUB2 binaries provided by the vendor of your operating system.
```
Lennonka, line 51: @nofaralfasi But we will probably need a different wording here because we don't expect users to use Satellite with other vendors, right? What should we tell them instead?

nofaralfasi: So we should remove line 51.

guides/common/assembly_using-network-boot-to-provision-hosts.adoc
```diff
@@ -81,6 +98,7 @@ include::modules/proc_configuring-smart-proxy-to-provision-secure-boot-enabled-h
 :extract_grub: {extract_deb_prefix}/{grub_efi_downloaded_package_name} {extract_deb_zst_suffix}
 :extract_shim: {extract_deb_prefix}/{shim_efi_downloaded_package_name} {extract_deb_xz_suffix}
 include::modules/proc_configuring-smart-proxy-to-provision-secure-boot-enabled-hosts.adoc[leveloffset=+2]
+endif::[]
```
maximiliankolb, line 101:
```suggestion
```

I think this is covered by line 125. Alternative: add 101 and drop 125.

maximiliankolb (changes requested):
build is broken due to an uneven amount of "ifdef" and "endif". Overall, LGTM!

guides/common/modules/con_using-pxe-to-provision-hosts.adoc
```diff
@@ -43,22 +43,27 @@ In {Project} provisioning, the PXE loader option defines the DHCP `filename` opt
 * For BIOS systems, select the *PXELinux BIOS* option to enable a provisioned host to download the `pxelinux.0` file over TFTP.
 * For UEFI systems, select the *Grub2 UEFI* option to enable a TFTP client to download `grubx64.efi` file, or select the *Grub2 UEFI HTTP* option to enable an UEFI HTTP client to download `grubx64.efi` with the HTTP Boot feature.
 
-ifndef::satellite[]
 {ProjectName} supports UEFI Secure Boot.
 SecureBoot PXE loaders enable a client to download the `shim.efi` bootstrap boot loader that then loads the signed `grubx64.efi`.
 Use the *Grub2 UEFI SecureBoot* PXE loader for PXE-boot provisioning or *Grub2 UEFI HTTPS SecureBoot* for HTTP-boot provisioning.
 
+ifdef::satellite[]
+By default, you can provision the RHEL version of your {ProjectServer} on Secure Boot enabled hosts.
```
nofaralfasi, line 51: Is it intentional that `RHEL` is not enclosed in curly brackets here?

nofaralfasi (approved):

guides/common/modules/con_using-pxe-to-provision-hosts.adoc
Lennonka: Yes, it is. I'm using the abbreviation for better readability.

asteflova: It's okay to use abbreviations. Typically, we expand abbreviations upon first use in a section or chapter to introduce its meaning (`Red Hat Enterprise Linux (RHEL)`) and then it's prefectly fine to keep using the abbreviation only (`RHEL`). Not sure what the current practice is with RHEL itself, that particular abbreviation is so common that it might not need expanding on each first use :shrug:

maximiliankolb (approved):

PR #2409, opened 2023-08-31
Use update-ca-trust to import LDAP certificate

ekohl:
It's also not needed to restart the service.

@evgeni this is what you mentioned. Is this correct? If so, it probably should be picked back to all releases.

* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.8/Katello 4.10
* [x] Foreman 3.7/Katello 4.9 (planned Satellite 6.14)
* [x] Foreman 3.6/Katello 4.8
* [x] Foreman 3.5/Katello 4.7 (Satellite 6.13)
* [x] Foreman 3.4/Katello 4.6 (EL8 only)
* [x] Foreman 3.3/Katello 4.5 on EL7 & EL8 (Satellite 6.12 on EL8 only; orcharhino 6.4/6.5 on EL8 only)
* [x] Foreman 3.2/Katello 4.4 on EL7 & EL8
* [x] Foreman 3.1/Katello 4.3 on EL7 & EL8 (Satellite 6.11 EL7/8; orcharhino 6.3 on EL7/8)
* We do not accept PRs for Foreman older than 3.1.

PR #3499, opened 2024-12-06
Update procedure to configure TLS for LDAP

asteflova:

I'm updating the procedure to configure TLS for LDAP authentication:
* Restarting httpd is not needed
* We can also simplify the instructions for uploading the LDAP certificate

I noticed https://issues.redhat.com/browse/SAT-19821 and https://github.com/theforeman/foreman-documentation/pull/2409/ and thought "why not?"

#### Checklists

A separate PR will probably be needed for some of the older branches due to conflicts but https://github.com/theforeman/foreman-documentation/pull/2409/ suggests that we should get the fixes there too.

* [x] Foreman 3.11/Katello 4.13 (orcharhino 6.11 on EL8 only)

### DIFF
--- a/guides/common/modules/proc_configuring-tls-for-secure-ldap.adoc
+++ b/guides/common/modules/proc_configuring-tls-for-secure-ldap.adoc
@@ -26,7 +26,6 @@ The filename extensions `.cer` and `.crt` are only conventions and can refer to 
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-# update-ca-trust enable
-# update-ca-trust
+# update-ca-trust extract
 ----
 . Delete the downloaded LDAP certificate from the temporary location on your {ProjectServer}.
--- a/guides/common/modules/proc_configuring-tls-for-secure-ldap.adoc
+++ b/guides/common/modules/proc_configuring-tls-for-secure-ldap.adoc
@@ -29,3 +29,8 @@ The filename extensions `.cer` and `.crt` are only conventions and can refer to 
 # update-ca-trust extract
 ----
 . Delete the downloaded LDAP certificate from the temporary location on your {ProjectServer}.
+
+ifndef::orcharhino[]
+.Additional resources
+* For more information about adding certificates to the system truststore, see link:{RHELDocsBaseURL}9/html/securing_networks/using-shared-system-certificates_securing-networks[Using shared system certificates] in _{RHEL}{nbsp}9 Securing networks_.
+endif::[]
--- a/guides/common/modules/proc_configuring-tls-for-secure-ldap.adoc
+++ b/guides/common/modules/proc_configuring-tls-for-secure-ldap.adoc
@@ -15,27 +15,18 @@ endif::[]
 You will remove the certificate when finished.
 +
 The filename extensions `.cer` and `.crt` are only conventions and can refer to DER binary or PEM ASCII format certificates.
-. Add the LDAP certificate to your CA trust list:
-.. Install the LDAP certificate in the `/etc/pki/tls/certs/` directory with the correct permissions:
+. Add the LDAP server certificate to the system trust store:
+.. Import the certificate:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-# install /tmp/_example.crt_ /etc/pki/tls/certs/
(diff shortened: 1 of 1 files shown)


ekohl (approved):
Thanks for taking over. Again a great simplification.

maximiliankolb (approved):
style-wise it LGTM; though I am not sure if that would work on Foreman on Debian/Ubuntu. But it's an improvement, so let's get it in.

ekohl:
> though I am not sure if that would work on Foreman on Debian/Ubuntu

Good call. It would not, but the previous instructions also didn't work. The Debian equivalent is to place the file in `/etc/ssl/certs` and run `update-ca-certificates`.

maximiliankolb:
Thanks Ewoud for providing the cmd/path. I will fix this in a follow-up PR so that it does not further block this PR. See https://github.com/theforeman/foreman-documentation/issues/3500

ekohl (approved):
FYI, this now matches https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/securing_networks/using-shared-system-certificates_securing-networks#adding-new-certificates_using-shared-system-certificates. Perhaps we should refer to it?

asteflova:
> FYI, this now matches https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/securing_networks/using-shared-system-certificates_securing-networks#adding-new-certificates_using-shared-system-certificates. Perhaps we should refer to it?

I saw that section and didn't think it would add much value. But why not, let's add a link.

asteflova:
I don't think we can _replace_ the steps on our side with a link (update-ca-trust **extract** is not part of the RHEL docs) but in https://issues.redhat.com/browse/SAT-19820, Evgeni does recommend adding a link so that's what I did. Except that I'd rather not link to a specific section, only to the whole chapter/assembly (chapter links shouldn't change as much as section links).

ekohl:
> update-ca-trust **extract** is not part of the RHEL docs

Perhaps that's a RHEL docs bug. Technically it works if you don't pass one, but there's a difference. Quoting `man update-ca-trust`:
> (absent/empty command)
> Same as the extract command described below. (However, the command may print fewer warnings, as this command is being run during rpm package installation, where non-fatal status output is undesired.)

That makes me think users should pass the `extract` command.

asteflova:
Great, thanks a lot, Mirek! Looking at the RHEL RNs that you linked:

> Using update-ca-trust without arguments is deprecated

which I believe confirms @ekohl's suspicion. If you do manage to apply the quick fix to update-ca-trust @mjahoda, I would appreciate it if you let me know because then I'd look into replacing the procedure steps on our side with a link to your guide.

However, for now, I think this PR is good enough to be merged. If anyone disagrees, please let me know.

neverpanic (approved):

ekohl:
> [3] https://issues.redhat.com/browse/RHEL-54695

This is how we (or at least, I) found out the hard way: https://github.com/voxpupuli/puppet-trusted_ca/commit/b3416c58e6a0d72ef8c49725a6193c3217b001cc & https://github.com/theforeman/puppet-foreman_proxy_content/commit/53187d9ede507c752d64e0a0780d2da89249d9b6 were the result of a very frustrating Friday when the change landed in CentOS Stream.

ekohl:
And as I merged I just noticed I should have squashed.

   563e29142275..ec409bd7baf8  3.11 -> 3.11
   8c6432a093c8..faf85d893aff  3.12 -> 3.12
   0a15d580fd14..f37aa7419fe3  3.13 -> 3.13

There's a merge conflict on 3.10 because 8069cf8518dc8799b98beeae8855b0e90a5d5541 wasn't picked.

asteflova:
I'll raise a follow-up PR to get the changes to 3.10 and below.

mjahoda:
JFYI, the quick fix (the addition of "extract") is available in the published RHEL 8 and 9 docs: https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/securing_networks/using-shared-system-certificates_securing-networks#adding-new-certificates_using-shared-system-certificates

asteflova:
That was very fast @mjahoda, thanks! I created an issue to update the docs on our side: https://github.com/theforeman/foreman-documentation/issues/3504

PR #3549, opened 2025-01-06
Declare a cipher unsupported for direct AD integration

asteflova:

I'm adding a note about the TLS_DHE_RSA_WITH_AES_256_GCM_SHA384 cipher suite being unsupported. For Satellite, I'm also adding a link to a KBase solution.

https://issues.redhat.com/browse/SAT-25882 reports errors in user environments that rely on the cipher.

N/A

#### Checklists

### DIFF
--- a/guides/common/modules/proc_configuring-the-active-directory-authentication-source-on-projectserver.adoc
+++ b/guides/common/modules/proc_configuring-the-active-directory-authentication-source-on-projectserver.adoc
@@ -3,6 +3,14 @@
 
 Enable Active Directory (AD) users to access {Project} by configuring the corresponding authentication provider on your {ProjectServer}.
 
+[NOTE]
+====
+{Project} does not support using the `TLS_DHE_RSA_WITH_AES_256_GCM_SHA384` cipher suite.
+ifdef::satellite[]
+For more information, see the Red{nbsp}Hat Knowledgebase solution link:https://access.redhat.com/solutions/4870221[API calls to Red Hat Satellite 6 fail intermittently on LDAP authentication].
+endif::[]
+====
+
 .Prerequisites
 * The base system of your {ProjectServer} must be joined to an Active Directory (AD) domain.
 To enable AD users to sign in with Kerberos single sign-on, use the System Security Services Daemon (SSSD) and Samba services to join the base system to the AD domain:


asteflova:
Hi @adamruzicka and @lhellebr, can you please review? Looking at the Jira (see this PR's description), I'm wondering whether we should also mention a specific Windows Server version that this applies to.

lhellebr:
More of a questions than suggestions:

1) Should we instead list supported ciphers positively? Sounds more sane than maintaining a list of everything we *don't* support.

2) Should we instead specify that the user should let the Satellite decide on what cipher to use, not changing the default configuration? But maybe not since the other side can already initiate the connection enforcing certain cipher?

adamruzicka:
> Should we instead list supported ciphers positively?

I'm not sure if that would be feasible. We were operating under the assumption that as long as both sides can agree on a cipher, it should work, until now we never really had to care about such low level details. Apparently there is a bug in the net-ldap/ruby-openssl/openssl or AD (or both) which makes the connections using this particular cipher occasionally break. Listing all that is supported would be better, if we actually knew. On a stock rhel9, `openssl ciphers -s` gives me 35 ciphers enabled by default, testing all of them would be rather time consuming.

> Should we instead specify that the user should let the Satellite decide on what cipher to use, not changing the default configuration? (But maybe not since the other side can already initiate the connection enforcing certain cipher?

Now this is where it gets interesting. Both sides have a list of allowed ciphers and they agree on one to use. Even with stock settings, `DHE-RSA-AES256-GCM-SHA384` is allowed to be used by default (at least on rhel side) but usually other ciphers from the list are preferred. 

If it works ootb, should we maybe reword this to state that this is only relevant when applying security hardening?

> OTOH Satellite should handle that and refuse such communication.

openssl does indeed handle this. The connection won't be established if both sides can't agree on connection parameters.

lhellebr:
> openssl does indeed handle this. The connection won't be established if both sides can't agree on connection parameters.

I know, but perhaps Satellite should disable it since it's known now not to be supported?

adamruzicka:
I don't think it is worth the effort. As far as I know we've only had one or two people run into this over the last ten years. We may reconsider in the future if people start running into it again, but for now I'd say having it documented is good enough

ekohl:
Looking at https://ciphersuite.info/cs/TLS_DHE_RSA_WITH_AES_256_GCM_SHA384/ it's declared weak, so IMHO it's best to disable it.

From reading this change itself, it's not obvious to me how I (as a user) would disable it. The linked KB article is also unclear. Do I need to perform any action on the client (=Satellite) or Server (=LDAP) side?

adamruzicka:
> Do I need to perform any action on the client (=Satellite) or Server (=LDAP) side?

On either end should be enough.

On the ldap side, you have to do whatever you have to do, depending on the implementation you have.

On client side (and on el*, no idea how this works on debian), you change `/etc/crypto-policies/back-ends/openssl.config` and the `CipherString` line in `/etc/crypto-policies/back-ends/opensslcnf.config`. The syntax is somewhat obscure and hard to get right. Maybe there are more sophisticated ways of going about this, but this already is way outside of what I'm sure about. Additionally, whatever you do in there will apply system-wide. Is this really something we want to do?

ekohl:
FYI: I think you can disable it system wide on RHEL 8 using https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/security_hardening/using-the-system-wide-cryptographic-policies_security-hardening#customizing-system-wide-cryptographic-policies-with-subpolicies_using-the-system-wide-cryptographic-policies (though https://www.redhat.com/en/blog/configuring-rhel-8-compliance-crypto-policy-related-cipher-block-chaining is IMHO a bit more concrete). Given it's a weak cipher, I'd expect it to be safe.

adamruzicka:
> Edit: to be explicit: don't modify the default policy but rather refine it by removing a single cipher.

That might be a good thing to do in general, but it still wouldn't address the original issue where the user modified the default policy to explicitly include this problematic cipher by hand. From brief testing on one of my machines, you can craft a configuration in crypto policy backends that will win over the subpolicy

ekohl:
At first I didn't see there was a customer case attached (because I wasn't logged in), so I assumed this cipher was enabled by default.

For those non-Red Hatters reading along: this really is a case where the user manually enabled an insecure cipher that led to problems.

There's only a limited amount we can do and we can't possibly enumerate all cases where the user did something that breaks. On the other hand, it led to problems and reading the case this is a relatively common occurrence. Now I wonder how to best deal with this.

I'd love to understand why users add this insecure cipher to their configuration and also don't follow the documentation on how to refine the policy properly.

asteflova:
Thanks for all the comments! I added another sentence to (hopefully) clarify that the cipher being enabled is a result of user actions and that in such situations, it needs to be disabled on either side.

I'm not adding any details on how exactly to disable the cipher because it being enabled is not the default state. This is not exactly a mainstream scenario so I'd prefer not to overdo how much information we provide. Let's assume that if users enabled it at some point, they can figure out how to disable it (I know that's not the safest of assumptions to make, but I think it might just be good enough in this situation).

adamruzicka:
> so I assumed this cipher was enabled by default.

Well, unless I'm misreading what I'm seeing, it is enabled by default. On a fresh RHEL9 machine:
```
# openssl ciphers -s | sed "s/:/\n/g"  | grep DHE-RSA-AES256-GCM-SHA384
ECDHE-RSA-AES256-GCM-SHA384
DHE-RSA-AES256-GCM-SHA384
```

but it is pretty low on the list.

asteflova:
There is an error that users encounter when the cipher is enabled and used:

`{ "error": { "message": "ERF77-7629 [Foreman::LdapException]: Error while connecting to 'server.com' LDAP server at 'ldap.example.com' during authentication ([Net::LDAP::Error]: Connection reset by peer - SSL_connect)" } }`

@adamruzicka Do you think it would be better to turn this around and document the error? So, more or less: "If you hit the following error, make sure that this cipher is disabled." Would that be a more suitable approach that the note that this PR currently proposes?

guides/common/modules/proc_configuring-the-active-directory-authentication-source-on-projectserver.adoc
```diff
@@ -3,6 +3,15 @@
 
 Enable Active Directory (AD) users to access {Project} by configuring the corresponding authentication provider on your {ProjectServer}.
 
+[NOTE]
+====
+{Project} does not support using the `TLS_DHE_RSA_WITH_AES_256_GCM_SHA384` cipher suite.
```
ekohl, line 8: I'm not sure we should call it "does not support". Perhaps something like:
```suggestion
The `TLS_DHE_RSA_WITH_AES_256_GCM_SHA384` cipher suite is known to cause incompatibilities.
```

ekohl (commented):
I think documenting exact errors is a good thing for troubleshooting. Sysadmins are trained in doing so (whether they know it or not). Today we mostly do so in KCS articles, but I think we should preempt errors if we know about them.

adamruzicka:
> Do you think it would be better to turn this around and document the error?

In general yes, but I'm afraid you can get the same error in other scenarios so it is not 100% reliable.

asteflova:
Based on the feedback, I tried to come up with something that would help both prevent the issue and troubleshoot it. Can you please re-review? Is this helpful, or did I just create a horrible [catdog](https://www.imdb.com/title/tt0154061/) and should just revert back to the note?

guides/common/modules/proc_configuring-the-active-directory-authentication-source-on-projectserver.adoc
```diff
@@ -105,6 +105,31 @@ $ curl -k -u : --negotiate https://{foreman-example-com}/users/extlogin
...
+# openssl s_client -connect _ldap.example.com_:636
+----
++
+If the `TLS_DHE_RSA_WITH_AES_256_GCM_SHA384` cipher is used, disable it on either the {ProjectServer} side or on the AD side.
+The `TLS_DHE_RSA_WITH_AES_256_GCM_SHA384` cipher is known to cause incompatibilities.
++
+ifdef::satellite[]
+For more information, see the Red{nbsp}Hat Knowledgebase solution link:https://access.redhat.com/solutions/4870221[API calls to Red Hat Satellite 6 fail intermittently on LDAP authentication].
+endif::[]
+ifndef::orcharhino[]
```
ekohl, line 129: Perhaps limit this to EL-based installs?
```suggestion
ifdef::foreman-el,katello[]
```

guides/common/modules/proc_configuring-the-active-directory-authentication-source-on-projectserver.adoc
```diff
@@ -105,6 +105,31 @@ $ curl -k -u : --negotiate https://{foreman-example-com}/users/extlogin
 <html><body>You are being <a href="{foreman-example-com}/hosts">redirected</a>.</body></html>
 ----
 
+.Troubleshooting
+* Connecting to the AD LDAP can sometimes fail with an error such as the following:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+{ "error": { "message": "ERF77-7629 [Foreman::LdapException]: Error while connecting to 'server.com' LDAP server at 'ldap.example.com' during authentication ([Net::LDAP::Error]: Connection reset by peer - SSL_connect)" } }
```
ekohl, line 113: Technically this is API output. I think the `message` part can also be seen in the UI. Probably if you use the `Test connection` button if it's 100% reproducible but I don't have the setup to test this.

asteflova: What if I clarify that the error message appears in the logs? That's how the issue was troubleshooted. Does it make sense?

adamruzicka: > if it's 100% reproducible

Sadly it's not, the failure is intermittent. From what I've seen I'd say each ldap lookup has a 1:50 chance of hitting this.

> What if I clarify that the error message appears in the logs?

That would make sense to me

asteflova:
@adamruzicka @ekohl @lhellebr Do you think that the current state is good enough? Is this something you feel comfortable approving?

ekohl (approved):
I think this is good. Long term I'd like to use th ERF error codes more. We used to maintain https://projects.theforeman.org/projects/foreman/wiki/ErrorCodes to make troubleshooting easier

asteflova:
Dear writers, can I get a style review?

maximiliankolb (approved):

PR #3550, opened 2025-01-07
Update /auth context path requirements for Keycloak

asteflova:

We need to update the procedures for configuring Keycloak as an authentication source for Foreman to reflect changes in `keycloak-httpd-client-install` version 1.3.

This PR reshuffles existing prerequisites for the setup (to better enable adding the new information) and then on top of that explains which steps users need to take depending on whether 1.3 or 1.2 is installed on their Foreman server.

Existing procedures for configuring Quarkus-based Keycloak as an authentication source currently contain a workaround because they are based on the fact that `keycloak-httpd-client-install` requires the Keycloak server to be initialized with `--http-relative-path=/auth`. That is no longer the case because with version 1.3, `keycloak-httpd-client-install` has been updated. Notable commits:

*   https://github.com/latchset/keycloak-httpd-client-install/commit/67a05be47ecabf9e0dcacdd85a9c050a2e2c89d4 removes /auth from URLs
*  https://github.com/latchset/keycloak-httpd-client-install/commit/ef7df4e33a298b37ae6632c1710042c52dab2a37 autodetects whether /auth is needed

Therefore, the support matrix that we are dealing with now involves two versions of Keycloak (one based on the Quarkus application server, the other based on the deprecated Wildfly server) and two versions of the `keycloak-httpd-client-install` utility (1.2 and 1.3).

Thanks to @ekohl for investigating the situation and recording his findings in https://issues.redhat.com/browse/SAT-29434.

#### Checklists

* [ ] Foreman 3.13/Katello 4.15

### DIFF
--- a/guides/common/modules/con_prerequisites-for-configuring-project-with-keycloak-quarkus-authentication.adoc
+++ b/guides/common/modules/con_prerequisites-for-configuring-project-with-keycloak-quarkus-authentication.adoc
@@ -10,11 +10,19 @@ On your {ProjectServer}:
 # {project-package-install} mod_auth_openidc keycloak-httpd-client-install python3-lxml
 ----
 // python3-lxml is only needed on EL8 because of https://issues.redhat.com/browse/RHEL-31496
+ifeval::["{context}" == "keycloak-quarkus"]
+* Check which `keycloak-httpd-client-install` version is installed:
++
+[options="nowrap", subs="verbatim,quotes,attributes"]
+----
+# rpm --query keycloak-httpd-client-install
+----
+endif::[]
 
 On the {keycloak} side, ensure you meet the following requirements:
 
 ifeval::["{context}" == "keycloak-quarkus"]
-* You must use a {keycloak} server that has been initialized with the `--http-relative-path=/auth` context path.
+* If `keycloak-httpd-client-install` version 1.2 or earlier is installed on your {ProjectServer}, you must use a {keycloak} server that has been initialized with the `--http-relative-path=/auth` context path.
 To access a {keycloak} server initialized with `--http-relative-path=/auth` from its web UI, go to `https://_{keycloak-example-com}_:8443/auth`.
 +
 [NOTE]
@@ -24,6 +32,8 @@ ifndef::orcharhino[]
 For more information about configuring a different context path, see the {RHDocsBaseURL}red_hat_build_of_keycloak/24.0/html-single/server_guide/index#reverseproxy-different-context-path-on-reverse-proxy[_{RHBK} Administration Guide_].
 endif::[]
 ====
++
+If `keycloak-httpd-client-install` version 1.3 or later is installed, your {keycloak} server does not need to be initialized with the `--http-relative-path=/auth` context path.
 endif::[]
 * Your {keycloak} server uses HTTPS instead of HTTP.
 * If the certificates or the CA are self-signed, they have been added to the end-user certificate truststore.
--- a/guides/common/modules/con_prerequisites-for-configuring-project-with-keycloak-quarkus-authentication.adoc
+++ b/guides/common/modules/con_prerequisites-for-configuring-project-with-keycloak-quarkus-authentication.adoc
@@ -1,8 +1,20 @@
 [id="prerequisites-for-configuring-{project-context}-with-keycloak-authentication_{context}"]
 = Prerequisites for configuring {Project} with {keycloak-quarkus} authentication
 
-* A {keycloak} account with administrative privileges.
-* A {keycloak} server that uses HTTPS instead of HTTP and has been initialized with the `--http-relative-path=/auth` context path.
+On your {ProjectServer}:
+
+* Install the packages required for registering a {keycloak} client:
++
+[options="nowrap", subs="verbatim,quotes,attributes"]
+----
+# {project-package-install} mod_auth_openidc keycloak-httpd-client-install python3-lxml
+----
+// python3-lxml is only needed on EL8 because of https://issues.redhat.com/browse/RHEL-31496
+
+On the {keycloak} side, ensure you meet the following requirements:
+
+ifeval::["{context}" == "keycloak-quarkus"]
+* You must use a {keycloak} server that has been initialized with the `--http-relative-path=/auth` context path.
 To access a {keycloak} server initialized with `--http-relative-path=/auth` from its web UI, go to `https://_{keycloak-example-com}_:8443/auth`.
 +
 [NOTE]
@@ -12,9 +24,12 @@ ifndef::orcharhino[]
 For more information about configuring a different context path, see the {RHDocsBaseURL}red_hat_build_of_keycloak/24.0/html-single/server_guide/index#reverseproxy-different-context-path-on-reverse-proxy[_{RHBK} Administration Guide_].
 endif::[]
 ====
-* If the certificates or the CA are self-signed, add them to the end-user certificate truststore.
-* A {keycloak} realm created for {Project} user accounts, for example `_{Project}_Realm_`.
-* Users imported or added to {keycloak}.
+endif::[]
+* Your {keycloak} server uses HTTPS instead of HTTP.
+* If the certificates or the CA are self-signed, they have been added to the end-user certificate truststore.
+* Your {keycloak} account has administrative privileges.
+* A realm is created on the {keycloak} server for {Project} user accounts, for example `_{Project}_Realm_`.
+* User accounts have been imported or added to {keycloak}.
 ifndef::orcharhino[]
 For more information on importing or creating users, see the {RHDocsBaseURL}red_hat_build_of_keycloak/24.0/html/server_administration_guide/user-storage-federation[_{RHBK} Administration Guide_].
 endif::[]
--- a/guides/common/modules/con_prerequisites-for-configuring-project-with-keycloak-wildfly-authentication.adoc
+++ b/guides/common/modules/con_prerequisites-for-configuring-project-with-keycloak-wildfly-authentication.adoc
@@ -1,11 +1,23 @@
 [id="prerequisites-for-configuring-{project-context}-with-keycloak-authentication_{context}"]
 = Prerequisites for configuring {Project} with {keycloak-wildfly} authentication
 
-* A {keycloak} account with administrative privileges.
-* A {keycloak} server that uses HTTPS instead of HTTP.
-* If the certificates or the CA are self-signed, ensure that they are added to the end-user certificate truststore.
-* A {keycloak} realm created for {Project} user accounts, for example `_{Project}_Realm_`.
-* Users imported or added to {keycloak}.
+On your {ProjectServer}:
+
+* Install the packages required for registering a {keycloak} client:
++
+[options="nowrap", subs="verbatim,quotes,attributes"]
+----
+# {project-package-install} mod_auth_openidc keycloak-httpd-client-install python3-lxml
+----
+// python3-lxml is only needed on EL8 because of https://issues.redhat.com/browse/RHEL-31496
+
+On the {keycloak} side, ensure you meet the following requirements:
+
+* Your {keycloak} server uses HTTPS instead of HTTP.
+* If the certificates or the CA are self-signed, they have been added to the end-user certificate truststore.
+* Your {keycloak} account has administrative privileges.
+* A realm is created on the {keycloak} server for {Project} user accounts, for example `_{Project}_Realm_`.
+* User accounts have been imported or added to {keycloak}.
 ifndef::orcharhino[]
 For more information about importing or creating users, see the {RHDocsBaseURL}red_hat_single_sign-on/7.6/html/server_administration_guide/assembly-managing-users_server_administration_guide#proc-creating-user_server_administration_guide[_{RHSSO} Server Administration Guide_].
 endif::[]
--- a/guides/common/modules/proc_registering-project-as-a-client-of-keycloak.adoc
+++ b/guides/common/modules/proc_registering-project-as-a-client-of-keycloak.adoc
@@ -20,13 +20,6 @@ endif::[]
 
 On your {ProjectServer}:
 
-. Install the packages required for registering a {keycloak} client:
-+
-[options="nowrap", subs="verbatim,quotes,attributes"]
-----
-# {project-package-install} mod_auth_openidc keycloak-httpd-client-install python3-lxml
-----
-// python3-lxml is only needed on EL8 because of https://issues.redhat.com/browse/RHEL-31496
 . Choose the authentication method you want {keycloak} users to use when authenticating to {Project}:
 * If you want users to authenticate by using the {ProjectWebUI}:
 .. Create a client for {Project}.


guides/common/modules/con_prerequisites-for-configuring-project-with-keycloak-quarkus-authentication.adoc
```diff
@@ -12,9 +32,14 @@ ifndef::orcharhino[]
 For more information about configuring a different context path, see the {RHDocsBaseURL}red_hat_build_of_keycloak/24.0/html-single/server_guide/index#reverseproxy-different-context-path-on-reverse-proxy[_{RHBK} Administration Guide_].
 endif::[]
 ====
-* If the certificates or the CA are self-signed, add them to the end-user certificate truststore.
-* A {keycloak} realm created for {Project} user accounts, for example `_{Project}_Realm_`.
-* Users imported or added to {keycloak}.
++
+If `keycloak-httpd-client-install` version 1.3 or later is installed, your {keycloak} server does not need to be initialized with the `--http-relative-path=/auth` context path.
```
lhellebr, line 36: Should this also be in `ifeval::["{context}" == "keycloak-quarkus"]`?

guides/common/modules/con_prerequisites-for-configuring-project-with-keycloak-quarkus-authentication.adoc
```diff
@@ -1,8 +1,28 @@
...
-* A {keycloak} server that uses HTTPS instead of HTTP and has been initialized with the `--http-relative-path=/auth` context path.
+On your {ProjectServer}:
+
+* Install the packages required for registering a {keycloak} client:
++
+[options="nowrap", subs="verbatim,quotes,attributes"]
+----
+# {project-package-install} mod_auth_openidc keycloak-httpd-client-install python3-lxml
+----
+// python3-lxml is only needed on EL8 because of https://issues.redhat.com/browse/RHEL-31496
```
ekohl, line 12: As noted in https://github.com/theforeman/foreman-documentation/issues/3675, it also affects EL9 and that's the only platform we now run on. Perhaps updated it while you're at it?
```suggestion
// python3-lxml is only needed on EL9 because of https://issues.redhat.com/browse/RHEL-31496
```

guides/common/modules/con_prerequisites-for-configuring-project-with-keycloak-wildfly-authentication.adoc
```diff
@@ -1,11 +1,23 @@
...
-* Users imported or added to {keycloak}.
+On your {ProjectServer}:
+
+* Install the packages required for registering a {keycloak} client:
++
+[options="nowrap", subs="verbatim,quotes,attributes"]
+----
+# {project-package-install} mod_auth_openidc keycloak-httpd-client-install python3-lxml
+----
+// python3-lxml is only needed on EL8 because of https://issues.redhat.com/browse/RHEL-31496
```
ekohl, line 12: And here
```suggestion
// python3-lxml is only needed on EL9 because of https://issues.redhat.com/browse/RHEL-31496
```

ekohl (commented):
Overall I think this looks valid.

guides/common/modules/con_prerequisites-for-configuring-project-with-keycloak-quarkus-authentication.adoc
aneta-petrova: Good catch! It looks like I actually decided to split the quarkus and wildfly prerequisites into separate files so we don't need any `ifeval`s here anymore. (With the new steps for `keycloak-httpd-client-install`, the prerequisites got so different that it seemed easier to me to just have two different files and risk some amount of content duplication.)

TL;DR: I dropped the `ifeval`s from this file and that should be enough.

aneta-petrova:
> Overall I think this looks valid.

Thanks for the review! I want to wait for QE input before adding the `tech review` ack but this means that I can now ask for `style review` to help move the PR along a bit.

guides/common/modules/con_prerequisites-for-configuring-project-with-keycloak-quarkus-authentication.adoc
```diff
@@ -12,9 +29,13 @@ ifndef::orcharhino[]
 For more information about configuring a different context path, see the {RHDocsBaseURL}red_hat_build_of_keycloak/24.0/html-single/server_guide/index#reverseproxy-different-context-path-on-reverse-proxy[_{RHBK} Administration Guide_].
 endif::[]
 ====
-* If the certificates or the CA are self-signed, add them to the end-user certificate truststore.
-* A {keycloak} realm created for {Project} user accounts, for example `_{Project}_Realm_`.
-* Users imported or added to {keycloak}.
++
+If `keycloak-httpd-client-install` version 1.3 or later is installed, your {keycloak} server does not need to be initialized with the `--http-relative-path=/auth` context path.
+* Your {keycloak} server uses HTTPS instead of HTTP.
+* If the certificates or the CA are self-signed, they have been added to the end-user certificate truststore.
```
maximiliankolb, line 35: suggestion (non-blocking): Use active voice.

* Who adds what where? I think you can copy this more of less from `guides/common/modules/proc_configuring-a-host-for-registration.adoc`.

guides/common/modules/con_prerequisites-for-configuring-project-with-keycloak-wildfly-authentication.adoc
```diff
@@ -1,11 +1,23 @@
...
+[options="nowrap", subs="verbatim,quotes,attributes"]
+----
+# {project-package-install} mod_auth_openidc keycloak-httpd-client-install python3-lxml
+----
+// python3-lxml is only needed on EL9 because of https://issues.redhat.com/browse/RHEL-31496
+
+On the {keycloak} side, ensure you meet the following requirements:
+
+* Your {keycloak} server uses HTTPS instead of HTTP.
+* If the certificates or the CA are self-signed, they have been added to the end-user certificate truststore.
```
maximiliankolb, line 17: same as above.

guides/common/modules/con_prerequisites-for-configuring-project-with-keycloak-quarkus-authentication.adoc
```diff
@@ -1,8 +1,25 @@
...
+* Check which `keycloak-httpd-client-install` version is installed:
++
+[options="nowrap", subs="verbatim,quotes,attributes"]
+----
+# rpm --query keycloak-httpd-client-install
+----
+
+On the {keycloak} side, ensure you meet the following requirements:
+
+* If `keycloak-httpd-client-install` version 1.2 or earlier is installed on your {ProjectServer}, you must use a {keycloak} server that has been initialized with the `--http-relative-path=/auth` context path.
```
MartinSpiessl, line 22: **suggestion(non-blocking)**: this sentence could somehow incorporate the other possibility that is explained in the note 3 lines down, i.e., alternatively one can adjust the context path that is used on the foreman via the file (`/etc/httpd/conf.d/foreman-openidc_oidc_keycloak_Foreman_Realm.conf).

Sometimes one cannot initialize the keycloak server with `--http-relative-path=/auth` because it is already setup in the company for other purposes as well.

The sentence currently reads as if there is no other way than setting the `--http-relative-path=/auth` option ("you must use...").

MartinSpiessl, line 22: **suggestion(non-blocking)**: FYI the `--http-relative-path=/auth` option is only needed for keycloak versions >17, before that having `/auth` was the default. Not sure if we want to explain this HERE, but in theory this would be confusing for people using such an old version of keycloak. Would be fine for me to leave this bit out, but if you think it is worth adding this, then something like this would be more accurate:

> If `keycloak-httpd-client-install` version 1.2 or earlier is installed on your {ProjectServer}, **and you are using keycloak in version is 18 or newer**, you must...

guides/common/modules/con_prerequisites-for-configuring-project-with-keycloak-quarkus-authentication.adoc
aneta-petrova: This might be a bit trickier because with multiple systems (Foreman, Keycloak), you might also have different admins. So the "who" might not be the same person, which made me think passive was more appropriate. (I remember checking the style guides when I was writing this, passive voice should be fine in these situations.)

guides/common/modules/con_prerequisites-for-configuring-project-with-keycloak-quarkus-authentication.adoc
aneta-petrova: We currently (try to) capture it by documenting the >17 and <17 Keycloaks separately. Except that we are using 'Quarkus-based Keycloak' and 'Wildfly-based Keycloak':

* https://docs.theforeman.org/nightly/Configuring_User_Authentication/index-katello.html#prerequisites-for-configuring-foreman-with-keycloak-authentication_keycloak-quarkus
* https://docs.theforeman.org/nightly/Configuring_User_Authentication/index-katello.html#prerequisites-for-configuring-foreman-with-keycloak-authentication_keycloak-wildfly

Because it's the underlying application platform (Quarkus/Wildfly) that is different. Your feedback makes me wonder whether distinguishing based on the platform is enough, in other words, whether we shouldn't also mention the version. But hopefully not: The change in the platform was a significant milestone that AFAIK was communicated extensively on the Keycloak side.

aneta-petrova: Thanks! I missed that before. Reworded to make it clear that these are alternatives.

MartinSpiessl: I tought the API change with removal of `/auth` was done in version 18, but I looked it up again, and it was done in version 17 (source: https://www.keycloak.org/migration/migrating-to-quarkus), my bad! This source also sounds like this only affects the quarkus-based distribution (no idea whether this is also the case for the "legacy" keycloak 17 that is still wild-fly based, I doubt they would put breaking changes into a "legacy" version), so I think this solves my comment here, it seems to be fine to imply that the `/auth` is removed if a user is using a quarkus-based keycloak.

I did some recherche on whether the distinction quarkus vs. wild-fly is good, here are my findings. This is a bit off-topic though, and might be better discussed elsewhere (in a separate issue?):

Since version 17, the **default** keycloak distribution is based on quarkus (source: https://www.keycloak.org/docs/latest/release_notes/index.html#quarkus-distribution-is-now-fully-supported). 
This does sound like one can still use wild-fly! Indeed there are "legacy" versions that are using wild-fly for 17,18, and 19 (source: https://mvnrepository.com/artifact/org.keycloak/keycloak-server-dist). So the distinction wild-fly vs. quarkus let's us deduce whether version >=17 or <17 is used **only when** we exclude these legacy versions.

Wild-fly seems to be removed from version >=20 onwards (source: https://www.keycloak.org/docs/latest/release_notes/index.html#wildfly-distribution-removed), so from there on things get a lot easier to document because we can just assume keycloak is always quarkus-based.
If users use a "legacy" keycloak based on wild-fly with versions 17,18, or 19, this is a really weird edge case, not sure if it is worth addressing this, I also doubt anyone has tested this.

It would be interesting to see the keycloak lifecycle to get some idea on which versions need to be considered in a current documentation. The open-source keycloak project only fixes stuff in the latest major version:
- https://github.com/keycloak/keycloak/discussions/10449#discussioncomment-2249604
- https://www.keycloak.org/security.html

The "Red Hat build of Keycloak" has concrete dates for up to when a version is supported, but it does not even mention versions before 22, and it seems like only the 2 latest major versions get support:
- https://access.redhat.com/support/policy/updates/red_hat_build_of_keycloak_notes

I hope nobody runs a keycloak older than version 20 in production nowadays (experience tells me there are always people doing exactly this though), so we could just demand that a keycloak >= version 20 is used which would also simplify the documentation.

IMHO: from a security-perspective, we should not incentivize people to use such an old keycloak version with their foreman and put some decent version requirement like >=20 in the docs, which would simplify the documentation.

aneta-petrova:
@lhellebr, to address your feedback:

> the documentation asks me to set OIDC issuer and JWKs URL to URIs including `/auth` even though the instance may be configured without `/auth`.

The latest commit adds "If using /auth context path" and "If not using /auth context path" conditions through the Quarkus/RHBK procedures. I think this change should address this part of your feedback, right?

> I need to include the port - it may be logical but it should be documented, especially since RHBK documentation uses 8443 by default.

I'm afraid I'll need a little more help with this part. Can you please explain what changes you'd like to see?

lhellebr:
Docs says: `OIDC Issuer: Set to https://keycloak.example.com/auth/realms/Foreman_Realm`
What I really had to set: `https://<rhbk_fqdn>:8443/realms/master`

That's not surprising because my RHBK runs on port 8443 while standard HTTPS port is 443. Port 8443 is default by RHBK documentation. I obviously used that in client registration command but I wasn't sure whether OIDC issuer and JWKs URL contain that port as well. Turns out they do and the documentation didn't tell me about that.

aneta-petrova:
Now I see it! Thanks. The port was missing in some of the Quarkus/RHBK paths. I'm adding it to a few places in the latest commit. Can you please re-review @lhellebr?

guides/common/modules/proc_configuring-a-project-client-to-provide-hammer-cli-authentication-with-keycloak.adoc
```diff
@@ -13,8 +13,16 @@ ifeval::["{context}" == "keycloak-wildfly"]
 * Ensure that the *Access Type* setting in the {Project} client in the {keycloak-wildfly} web UI is set to *public*.
 For more information, see xref:configuring-the-{project-context}-client-in-keycloak_keycloak-wildfly[].
 endif::[]
+ifeval::["{context}" == "keycloak-quarkus"]
+* If you initialized your Keycloak server without the `--http-relative-path=/auth` context path, obtain the values to configure {Project} settings from the following URL: `https://_{keycloak-example-com}_/realms/_{Project}_Realm_/.well-known/openid-configuration`.
+Replace _{Project}_Realm_ with the name of the {keycloak} realm created for your {Project} server.
```
maximiliankolb, line 18:
```suggestion
Replace _{Project}_Realm_ with the name of the {keycloak} realm created for your {ProjectServer}.
```

lhellebr:
Hi,
Since you added the port 8443 somewhere, I think you need to add it everywhere for consistency, e.g.:

```
keycloak-httpd-client-install --app-name foreman-openidc \
--keycloak-server-url "https://keycloak.example.com:8443"
...
```

```
hammer auth login oauth \
--oidc-token-endpoint 'https://keycloak.example.com:8443/realms/Foreman_realm/protocol/openid-connect/token' \
--oidc-authorization-endpoint 'https://keycloak.example.com:8443'
...
```
^ here mind there are two places where port needs to be specified

Of course the user doesn't need to specify port when they've set up their RHBK instance to run on 443 but the docs seem to use 8443 by default.

aneta-petrova:
Thanks @lhellebr! I added the port to the two places. I checked the rest and confirm that these seem to be the only two missing.

lhellebr:
ACK. Thanks, Aneta :)

aneta-petrova:

be2fe81965..f3d58a7669  3.14 -> 3.14

PR #3552, opened 2025-01-07
Add note about resources searchable by title

Lennonka:

Adding a note to global registration for users who need to edit the registration command

New feature allows users to search certain resources not only by ID but also by name.
This feature was introduced to ease transition from the bootstrap script to global registration.
Documents https://github.com/theforeman/foreman/pull/10351
[SAT-28832](https://issues.redhat.com/browse/SAT-28832) (private)

- I couldn't find any better place.
- I think a formal Note is in order to emphasize this info but feel free to disagree.

#### Checklists

### DIFF
--- a/guides/common/modules/proc_registering-a-host.adoc
+++ b/guides/common/modules/proc_registering-a-host.adoc
@@ -96,12 +96,16 @@ The scope of the JWTs is limited to the registration endpoints only and cannot b
 [NOTE]
 ====
 {Project} generates the registration command with parameters that search resources by ID.
-However, you can edit the registration command to search the following resources by name:
+You can edit the registration command to search the following resources by title:
 
-Organization:: URL fragment example: `organization=My+Organization`
-Location:: URL fragment example: `location=My+Location`
-Host group:: URL fragment example: `hostgroup=My+Host+Group`
-Operating system:: URL fragment example: `operatingsystem=My+Operating+System`
+Organization:: URL fragment example: `organization=My%20Organization`
+Location:: URL fragment example: `location=My%20Location`
+Host group:: If a host group is nested, include the parent group separated with the slash character (`/`).
++
+URL fragment example: `hostgroup=Parent%20Group%2FMy%20Host%20Group`
+Operating system:: URL fragment example: `operatingsystem=My%20Operating%20System`
+
+The parameter values must be URL encoded.
 ====
 
 .CLI procedure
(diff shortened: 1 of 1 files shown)


guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -93,6 +93,17 @@ Therefore, do not delete, block, or change permissions of the user during the to
 +
 The scope of the JWTs is limited to the registration endpoints only and cannot be used anywhere else.
 
+[NOTE]
+====
+{Project} generates the registration command with parameters that search resources by ID.
+However, you can edit the registration command to search the following resources by name:
```
stejskalleos, line 99:
```suggestion
However, you can edit the registration command to search the following resources by title:
```

We use `title` instead od `name`. For example:
Operating system `RHEL 9.5` is the title, and `RHEL` is the name.
With host groups, if the host group has a parent, the tile would be `Parent / Child group.`

amolpati30:
@Lennonka 
I have tested using the following URL fragment, and host registration works as expected:
hostgroup=test1%2Ftest2
location=Default%20Location
operatingsystem=RHEL%209.5
organization=Default%20Organization

amolpati30:
We can also register a host using the following parameters. Do you think this should be added to the documentation?
location=Default+Location
operatingsystem=RHEL+9.5
organization=Default+Organization

maximiliankolb:
what's the use case to edit this by hand? Why make users change the way Foreman/Katello searches for resources?
Is the alternative an API endpoint that accepts org/loc/os/host group/ak and returns the registration command?

Lennonka:
> what's the use case to edit this by hand? Why make users change the way Foreman/Katello searches for resources? Is the alternative an API endpoint that accepts org/loc/os/host group/ak and returns the registration command?

@stejskalleos Can you please clarify this?

stejskalleos:
> what's the use case to edit this by hand? 

It was a feature request by one of the Foreman users.

Lennonka:
@stejskalleos @amolpati30 Ack?

Lennonka:
@maximiliankolb The use case is that one user generates a registration command, edits it and forwards it to other users who run it on hosts, for example, in various locations.
We do not force them to use this feature, we're just giving them the option to use it.

amolpati30:
Ack from QE :+1:

stejskalleos (approved):
:green_apple:

maximiliankolb (approved):
ACK; works for me. I was just curious to _why_ users would want this.

PR #3554, opened 2025-01-08
Drop instructions to update grub2-efi

ekohl:

As the title says: it drops the manual instruction to install/update the grub2-efi package.

As far as I can tell this was needed in the past because RHEL 7.9 contained an updated grub2-efi with support for HTTPBoot. Now that EL7 has been dropped, this is irrelevant and any installed version will do. The installer when enabling the tftp feature the already guarantees that the package is installed so the instruction can be dropped.

I only selected Foreman 3.13 as a cherry pick so we don't modify existing Satellite releases, but it's probably safe there as well.

#### Checklists


maximiliankolb (approved):
diff LGTM, I cannot comment much about the package itself. I know that Foreman/Katello 3.13+ runs on EL9 only. So the argument that we can remove it because this is a relic from the time when Smart Proxies ran on EL7 makes sense to me.

apinnick:
@Lennonka FYI

Lennonka (approved):

Lennonka:
Cherry picked:
- ea5fd5cedb..3843d31d13  3.13 -> 3.13

PR #3565, opened 2025-01-14
Remove unused text

maximiliankolb:
~same as ten minutes ago. Kindly asking for another review~ @asteflova 

While working on fixing a broken xref, Anet and I found out that the whole section is not in use -> let's drop it. You can see that the GHA bot edited the first comment but there's still no changed file.


maximiliankolb:
I cannot find the actual broken link but I suspect that it'd be somewhere around https://docs.redhat.com/en/documentation/red_hat_satellite/6.16/html-single/installing_satellite_server_in_a_connected_network_environment/index#enabling-the-project-client-name-repository_satellite if Satellite docs were built based on "master".

asteflova:
The link is in a part of the module that has an ifeval for disconnected scenarios but if I'm looking at this correctly, then the disconnected guide doesn't even include the module. So the link is not even used anywhere..? Can you remove the whole ifeval section?

maximiliankolb:
Testing ... :crossed_fingers:

PR #3568, opened 2025-01-14
Remove unused procedure to remove keycloak authentication

ekohl:

See summary

Since 50b9023ea0779c82d9b399e428cc1fc1eb159559 this procedure isn't included anymore.

I'm not sure if this is really the right solution. An explicit procedure can be much better than a note somewhere in a procedure to enable it.

I also didn't look at any stable branches and IMHO it doesn't need to.

#### Checklists


asteflova (approved):

asteflova:
> I'm not sure if this is really the right solution. An explicit procedure can be much better than a note somewhere in a procedure to enable it.

You would probably find that opinions vary from documentarian to documentarian. I prefer to avoid maintaining whole procedures dedicated to a single command because other documentation sources exist for this purpose (man pages, --help pages, references like our Hammer guide). And we can reference those.

Besides a note, a simple para or a bullet point under `.Additional resources` or some such section also works. I'm not saying that it's great, just that maintaining a whole module is worse IMO.

ekohl:
I'd say there's a trade off. Being consistent is important and a separate chapter in the navigation means users generally know how to find it. If it's sometimes within in a chapter, sometimes its own then that can be frustrating. Given that some removal procedures are non-trivial and deserve their own chapter I'd lean to always have your own chapter.

@apinnick or others with an opinion on this?

asteflova:

d332f751b7..aee4dbe282  3.13 -> 3.13
ff0009c1b9..0d47aebafd  3.12 -> 3.12
c074636cf4..5ed1b494c1  3.11 -> 3.11

PR #3569, opened 2025-01-14
Deprecate iPXE in Satellite

Lennonka:

Deprecating the iPXE feature in Satellite

[SAT-28823](https://issues.redhat.com/browse/SAT-28823) (public)

- For Satellite 6.17

#### Checklists

### DIFF
--- a/guides/common/modules/con_using-ipxe-to-reduce-provisioning-times.adoc
+++ b/guides/common/modules/con_using-ipxe-to-reduce-provisioning-times.adoc
@@ -5,6 +5,13 @@ iPXE is an open-source network-boot firmware.
 It provides a full PXE implementation enhanced with additional features, such as booting from an HTTP server.
 For more information about iPXE, see link:https://ipxe.org[iPXE website].
 
+ifdef::satellite[]
+[IMPORTANT]
+====
+The iPXE network boot is **deprecated** and will be removed in a future release.
+====
+endif::[]
+
 You can use iPXE if the following restrictions prevent you from using PXE:
 
 * A network with unmanaged DHCP servers.


asteflova:
I like how https://github.com/theforeman/foreman-documentation/blob/master/guides/common/modules/proc_registering-clients-using-the-bootstrap-script.adoc includes the deprecation notice in the heading because it makes everything clear just from looking at the ToC. But I don't think we have an established standard so if you prefer the IMPORTANT box, that's also fine.

guides/common/modules/con_using-ipxe-to-reduce-provisioning-times.adoc
```diff
@@ -5,6 +5,13 @@ iPXE is an open-source network-boot firmware.
 It provides a full PXE implementation enhanced with additional features, such as booting from an HTTP server.
 For more information about iPXE, see link:https://ipxe.org[iPXE website].
 
+ifdef::satellite[]
```
asteflova, line 8: Just curious: Satellite-only? If the feature is to be removed from code, then it won't be available in any flavor, will it?

asteflova (approved):

guides/common/modules/con_using-ipxe-to-reduce-provisioning-times.adoc
Lennonka: The iPXE code is external from the Foreman code base (well except for provisioning templates and PXE loaders). For Satellite, iPXE is provided by RHEL OS packages. These iPXE packages will no longer be included in RHEL 10. For upstream users, it doesn't stop them from installing iPXE from somewhere else and continuing using it. I'm not aware that we will be removing the functionality from Foreman.

ekohl: Note that it mentions that they're removing a lot of drivers, not iPXE entirely:
>  Instead, iPXE now depends on the platform firmware to provide a NIC driver for the network boot. The /usr/share/ipxe/ipxe-snponly-x86_64.efi and /usr/share/ipxe/undionly.kpxe iPXE binary files are the part of the ipxe-bootimgs package and use the NIC driver provided by the platform firmware.

I'm not familiar enough with the matter to say for sure what this means for end users because I don't know how well supported the platform firmware is.

ares (approved):

guides/common/modules/con_using-ipxe-to-reduce-provisioning-times.adoc
sbernhard: Should there be a community board discussion regarding this major change? I guess, it will affect a lot of scenarios of users?

ekohl: > Should there be a community board discussion regarding this major change? I guess, it will affect a lot of scenarios of users?

This specific PR is only affecting Satellite and the change @Lennonka mentioned is what happens in EL10 so Foreman isn't the right project. To be explicit: Foreman has never packaged iPXE and always consumed it from somewhere else.

PR #3588, opened 2025-01-17
Provide env var to use gh CLI in GHA

maximiliankolb:

Provide env vars to use `gh` in GHA

Original error message:
````
  Run gh run download '' -n pr
    gh run download '' -n pr
    shell: /usr/bin/bash -e {0}
  gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable. Example:
    env:
      GH_TOKEN: ${{ github.token }}
````

Refs 71b572445ab8a4ab156e5a9e66463ce57b76b6fa (PR 3587 on GitHub)

#### Checklists

no cherry-picks.

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -7,9 +7,6 @@ on:
     types:
       - completed
 
-env:
-  GH_TOKEN: ${{ github.token }}
-
 jobs:
   preview-failed:
     if: github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion != 'success'
@@ -19,6 +16,8 @@ jobs:
 
       - name: Download metadata artifact
         run: gh run download '${{ github.event.workflow_run.id }}' -n pr
+        env:
+          GH_TOKEN: ${{ github.token }}
 
       - name: Unzip artifact
         run: unzip pr.zip
@@ -40,6 +39,8 @@ jobs:
 
       - name: Download metadata artifact
         run: gh run download '${{ github.event.workflow_run.id }}' -n pr
+        env:
+          GH_TOKEN: ${{ github.token }}
 
       - name: Unzip artifact
         run: unzip pr.zip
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: .github/workflows/preview.yml [hunk 1]
```diff
@@ -7,6 +7,9 @@ on:
     types:
       - completed
 
+env:
+  GH_TOKEN: ${{ github.token }}
```
ekohl, line 11: From a security point of view I wonder if we should set this at the top level or not. I'd lean to being as specific as possible. @evgeni any thoughts?

evgeni: That'd mean adding it to every `run` that has `gh`?

maximiliankolb: Can we somehow use CODEOWNERS to prevent changes from non-maintainers in `.github/*` in PRs?

Applied in c4ffa1acaf4ab8951a276ea233df9f8567eac213

evgeni (approved):

.github/workflows/preview.yml
maximiliankolb: True. But still a nice angle: PRs that change/introduce GHA that run on PRs and use that token. I can only hope it's similar to GitLab and tied to your user account.

maximiliankolb:
We are one step further: https://github.com/theforeman/foreman-documentation/actions/runs/12830400048/job/35778499772

````
Run unzip pr.zip
  unzip pr.zip
  shell: /usr/bin/bash -e {0}
unzip:  cannot find or open pr.zip, pr.zip.zip or pr.zip.ZIP.
Error: Process completed with exit code 9.
````

.github/workflows/preview.yml
ekohl: > That'd mean adding it to every `run` that has `gh`?

More that any npm package we happen to pull in via `surge` also can access the env var.

evgeni:
Ooooh, `gh run download` does  unpack the artifact… handy?

PR #3591, opened 2025-01-17
Rely on gh run download to extract the artifacts

ekohl:

Turns out gh already unpacks the artifact so the unzip steps are no longer needed. This also means we need multiple indivual commands because they need to be extract to different directories.

https://github.com/theforeman/foreman-documentation/pull/3588#issuecomment-2598421476

I promise I'm trying to make things easier ;)

I also verified I can download manually:

```console
$ gh run download --repo theforeman/foreman-documentation '12830208840' --name foreman-docs-web-master --dir ~/tmp-docs/preview
$ gh run download --repo theforeman/foreman-documentation '12830208840' --name foreman-docs-html-provide_env_var --dir ~/tmp-docs/preview/master
$ ls ~/tmp-docs
preview
$ ls ~/tmp-docs/preview/
404.html  CNAME  css  favicon.ico  img  index.html  js  master  release
```

#### Checklists

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -19,9 +19,6 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
 
-      - name: Unzip artifact
-        run: unzip pr.zip
-
       - name: Read PR data
         run: echo "PR_DATA=$(cat ./pr.json)" >> $GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


maximiliankolb (approved):
Looks sane. Let's try it out.

.github/workflows/preview.yml
```diff
@@ -66,14 +55,18 @@ jobs:
 
       - name: Create preview layout
         run: |
-          mkdir -p preview/${{ fromJSON(env.PR_DATA).target_name }}
-          unzip -d preview foreman-docs-web-master.zip
-          unzip -d preview/${{ fromJSON(env.PR_DATA).target_name }} foreman-docs-html-${{ fromJSON(env.PR_DATA).branch_name }}.zip
+          gh run download '${{ github.event.worfklow_run.id }}' --name 'foreman-docs-web-master' --dir preview
```
ekohl, line 58: Due to a merge conflict, Worf is back. Surprisingly, https://github.com/theforeman/foreman-documentation/actions/runs/12831182509/job/35780996640 passed:
> gh run download '' --name 'foreman-docs-web-master' --dir preview

The preview and diff look correct so perhaps the workflow ID is not needed at all and it's smart enough to detect it?

ekohl: Answering my own question from `man gh-run-download`:
> By default, this command downloads the latest artifact created and uploaded through GitHub Actions. Because workflows can  delete  or  overwrite  artifacts, <run-id> must be used to select an artifact from a specific workflow run.

So it worked because it was the last run, but that's not guaranteed to be correct. https://github.com/theforeman/foreman-documentation/pull/3593 should address this.

Overall I think this is now almost done, even though it was bumpy.

maximiliankolb: Hmm, this is a good sign to me: https://github.com/theforeman/foreman-documentation/pull/3592#issuecomment-2598537737

So yes, we should either fix the typo and ignore that it's not really necessary or drop it?

PR #3598, opened 2025-01-20
Drop references to servers on RHEL 8

aneta-petrova:

I'm removing a few references to running servers and proxy servers on RHEL 8. Some of them are links to RHEL 8 docs in places where it's now only RHEL 9 that should be linked.

I checked all guides to see if there is anything left that we should remove.

N/A

#### Checklists

### DIFF
--- a/guides/common/modules/proc_configuring-a-cross-forest-trust-between-freeipa-and-active-directory-for-project.adoc
+++ b/guides/common/modules/proc_configuring-a-cross-forest-trust-between-freeipa-and-active-directory-for-project.adoc
@@ -5,7 +5,7 @@ When your {FreeIPA} deployment includes a cross-forest trust with Active Directo
 
 .Prerequisites
 * An existing {FreeIPA} server with a cross-forest trust with AD established.
-For more information, see link:{RHELDocsBaseURL}8/html-single/installing_identity_management/index#installing-trust-between-idm-and-ad_installing-identity-management[{RHEL}{nbsp}8 _Installing trust between IdM and AD_].
+For more information, see link:{RHELDocsBaseURL}9/html/installing_trust_between_idm_and_ad[{RHEL}{nbsp}9 _Installing trust between IdM and AD_].
 
 .Procedure
 
--- a/guides/common/modules/proc_installing-the-load-balancer.adoc
+++ b/guides/common/modules/proc_installing-the-load-balancer.adoc
@@ -1,7 +1,7 @@
 [id="Installing_the_Load_Balancer_{context}"]
 = Installing the load balancer
 
-The following example provides general guidance for configuring an HAProxy load balancer using {EL} 9 or {EL} 8.
+The following example provides general guidance for configuring an HAProxy load balancer using {EL} 9.
 However, you can install any suitable load balancing software solution that supports TCP forwarding.
 
 .Procedure
--- a/guides/common/modules/proc_configuring-the-active-directory-authentication-source-on-projectserver.adoc
+++ b/guides/common/modules/proc_configuring-the-active-directory-authentication-source-on-projectserver.adoc
@@ -68,10 +68,7 @@ ad_gpo_map_service = +foreman
 ----
 ifndef::orcharhino[]
 +
-For more information on GPOs, see the following documents:
-+
-* link:{RHELDocsBaseURL}9/html/integrating_rhel_systems_directly_with_windows_active_directory/managing-direct-connections-to-ad_integrating-rhel-systems-directly-with-active-directory#how-sssd-interprets-gpo-access-control-rules_applying-group-policy-object-access-control-in-rhel[How SSSD interprets GPO access control rules] in _Integrating RHEL systems directly with Windows Active Directory (RHEL{nbsp}9)_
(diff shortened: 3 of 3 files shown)


guides/common/modules/proc_installing-the-load-balancer.adoc
```diff
@@ -1,21 +1,21 @@
...
 ----
-# {client-package-install-el8} haproxy
+# {client-package-install-el9} haproxy
 ----
 . Install the following package that includes the `semanage` tool:
 +
 [options="nowrap" subs="attributes"]
 ----
-# {client-package-install-el8} policycoreutils-python-utils
+# {client-package-install-el9} policycoreutils-python-utils
```
asteflova, line 18: I'm not sure about these. I thought that a load balancer needs to be installed on a smart proxy, but these procedures use client-package-install-el8. Should I leave this whole file as it was?

ekohl: I think `client-package-install-el8` was wrong in the first place. Since we dropped EL7 we should really simply it back to just `server-package-install`, which is defined different for EL vs deb but otherwise always the same.

guides/common/modules/proc_configuring-a-cross-forest-trust-between-freeipa-and-active-directory-for-project.adoc
```diff
@@ -5,7 +5,7 @@ When your {FreeIPA} deployment includes a cross-forest trust with Active Directo
 
 .Prerequisites
 * An existing {FreeIPA} server with a cross-forest trust with AD established.
-For more information, see link:{RHELDocsBaseURL}8/html-single/installing_identity_management/index#installing-trust-between-idm-and-ad_installing-identity-management[{RHEL}{nbsp}8 _Installing trust between IdM and AD_].
+For more information, see link:{RHELDocsBaseURL}/html/installing_trust_between_idm_and_ad[{RHEL}{nbsp}9 _Installing trust between IdM and AD_].
```
ekohl, line 8: Shouldn't this include the RHEL version number?
```suggestion
For more information, see link:{RHELDocsBaseURL}9/html/installing_trust_between_idm_and_ad[{RHEL}{nbsp}9 _Installing trust between IdM and AD_].
```

guides/common/modules/con_kickstart-workflow.adoc
```diff
@@ -14,4 +14,4 @@ When you run a {ProjectName} Kickstart script, the script performs the following
 . It runs user defined snippets.
 
 .Additional resources
-For more information about Kickstart, see {RHELDocsBaseURL}8/html/automatically_installing_rhel/automated-installation-workflow_rhel-installer[Automated installation workflow] in _Automatically installing RHEL{nbsp}8_.
+For more information about Kickstart, see {RHELDocsBaseURL}9/html/automatically_installing_rhel/automated-installation-workflow_rhel-installer[Automated installation workflow] in _Automatically installing RHEL{nbsp}9_.
```
maximiliankolb, line 17: This is about Clients! I think your update is great but does not match your PR. Technically, we could still list EL7/8/9 here. Bumping to 9 as latest major RHEL is :+1:

guides/common/modules/proc_installing-the-load-balancer.adoc
maximiliankolb: Are you sure? My understanding is that you install HAProxy on a fresh EL 8 host which does not yet use `satellite-maintain` to install packages? See https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-satellite.adoc#L102

guides/common/modules/con_kickstart-workflow.adoc
asteflova: My bad, let's move it to https://github.com/theforeman/foreman-documentation/pull/3601

guides/common/modules/proc_installing-the-load-balancer.adoc
ekohl: I think you're right, but that's why the client vs server is so confusing. HAProxy will be a server and probably not a client for Foreman (though it might)

asteflova: It looks like we should track this separately so I'd appreciate your review here: https://github.com/theforeman/foreman-documentation/pull/3605.

I'll drop the change from this PR.

guides/common/modules/proc_installing-the-load-balancer.adoc
```diff
@@ -1,7 +1,7 @@
 [id="Installing_the_Load_Balancer_{context}"]
 = Installing the load balancer
 
-The following example provides general guidance for configuring an HAProxy load balancer using {EL} 9 or {EL} 8.
+The following example provides general guidance for configuring an HAProxy load balancer using {EL} 9.
```
asteflova, line 4: Should I drop this change too? Or do we want to stick to EL 9 only here?

Lennonka (approved):
LGTM

ehelms (approved):

aneta-petrova:

* 017e596faa..890af69780  3.13 -> 3.13

PR #3608, opened 2025-01-23
Create a new guide for provisioning virtual machines (2 of 3)

Lennonka:

Splitting _Provisioning hosts_ into 3 new guides:

1. Provisioning machines by using an installer - includes network boot and boot disks, which are both installer-based provisioning, plus Discovery, which also uses network boot 
2. **Provisioning virtual machines** - includes virtualization providers, image-based provisioning, and directives for installer-based provisioning if available _**<--- We are here.**_
3. Provisioning cloud instances - includes cloud infrastructure providers and image-based provisioning only

To mitigate pain points identified in [SAT-26089](https://issues.redhat.com/browse/SAT-26089) comments (public).

- _Provisioning hosts_ will be destroyed after all 3 guides from above will have been created.
- Currently, I am only modifying the structure. Further, detailed improvements will be done in subsequent PRs. Therefore, the new guide is terminologically inconsistent.
  **Reviewers, please, focus on the title and heading structure in this PR.** Using preview first is highly recommended.
- May require updates of documentation links in the web UI.

#### Checklists


guides/doc-Provisioning_Virtual/docinfo.xml
```diff
@@ -0,0 +1,17 @@
+<title>{ProvisioningVirtualDocTitle}</title>
+<subtitle>Provision virtual machines on virtualization infrastructure by using images</subtitle>
+<productname>Red Hat Satellite</productname>
+<productnumber>{ProjectVersion}</productnumber>
+<abstract>
+  <para>
+  TODO
```
Lennonka, line 7: Note to self: TODO

guides/common/modules/snip_warning-destroy-vm-on-host-delete.adoc
```diff
@@ -0,0 +1,11 @@
+[WARNING]
+====
+By default, if a host entry that is associated with a virtual machine is deleted, the virtual machine will remain on the compute resource.
+
+To delete the virtual machine on the compute resource when the associated host entry is deleted:
+
+. Navigate to *Administer > Settings* and select the *Provisioning* tab.
+. Change the *Destroy associated VM on host delete* setting to *Yes*.
```
maximiliankolb, line 8:
```suggestion
If you want {Project} to delete VMs on the compute resource when the associated host entry is deleted, set the *Destroy associated VM on host delete* setting to *Yes*.
```

suggestion/question (non-blocking): Just an idea to reword this slightly. It's a snippet, so I though that instead of two steps, we could just name the relevant setting.

Because it's a snippet, I agree that it's OK to lose the Hammer CLI command compared to guides/common/modules/proc_removing-a-virtual-machine-upon-host-deletion.adoc.

guides/doc-Provisioning_Virtual/master.adoc
```diff
@@ -0,0 +1,51 @@
+include::common/attributes.adoc[]
+include::common/header.adoc[]
+:context: provisioning-virtual
+:provisioning-virtual:
+
+= {ProvisioningVirtualDocTitle}
+
+// This guide is not ready for stable releases
```
maximiliankolb, line 8: question: Do you have reasons to believe that the guide would not work for unstable and/or foreman-el/foreman-deb?

maximiliankolb (commented):
Thanks for your effort Lena! Diff looks very reasonable to me but tech ACK is appreciated.

guides/common/modules/con_provisioning-virtual-machines-on-kubevirt.adoc
```diff
@@ -26,7 +26,7 @@ include::snip_prerequisites-common-compute-resource.adoc[]
 * A {SmartProxyServer} managing a network on the {Kubernetes} cluster.
 Ensure that no other DHCP services run on this network to avoid conflicts with {SmartProxyServer}.
 For more information about network service configuration for {SmartProxyServers}, see {ProvisioningDocURL}preparing-networking[Preparing networking] in _{ProvisioningDocTitle}_.
+* Ensure the provisioning user has adequate permissions to provision hosts.
```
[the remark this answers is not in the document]
Lennonka: Actually, there's a snippet with common prereqs for all compute resource. It should go there.

guides/doc-Provisioning_Virtual/master.adoc
Lennonka: It probably should work for all builds, since images are not part of Katello. However, right now I'm not sure it works at all and I would prefer to revisit this question later :))

guides/doc-Provisioning_Virtual/master.adoc
```diff
@@ -0,0 +1,51 @@
...
+
+include::common/modules/con_provisioning-contexts.adoc[leveloffset=+1]
+
+include::common/modules/proc_setting-the-provisioning-context.adoc[leveloffset=+2]
+
+include::common/modules/proc_using-novnc-to-access-virtual-machines.adoc[leveloffset=+1]
+
+include::common/assembly_provisioning-virtual-machines-kvm.adoc[leveloffset=+1]
+
+include::common/assembly_provisioning-virtual-machines-rhv.adoc[leveloffset=+1]
```
apinnick, line 36: Are we still supporting RHV? Chris Roberts told me to remove the RHV/oVirt content from the virt-who guide.

guides/doc-Provisioning_Virtual/master.adoc
```diff
@@ -0,0 +1,51 @@
...
+
+include::common/assembly_provisioning-virtual-machines-vmware.adoc[leveloffset=+1]
+
+ifndef::satellite[]
+include::common/assembly_provisioning-virtual-machines-proxmox.adoc[leveloffset=+1]
+endif::[]
+
+include::common/assembly_provisioning-virtual-machines-kubevirt.adoc[leveloffset=+1]
```
apinnick, line 44: I think we can remove this since the kubevirt plugin no longer works and will be removed in 6.17.

Lennonka:
@stejskalleos Please, answer the questions above and check the TOC of the guide.

guides/common/assembly_security-settings.adoc
```diff
@@ -1,11 +1,9 @@
 include::modules/con_security-settings.adoc[]
 
+ifndef::provisioning-virtual[]
 include::modules/proc_configuring-the-security-token-validity-duration.adoc[leveloffset=+1]
+endif::[]
```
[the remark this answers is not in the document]
stejskalleos: It is.
The machine will still ask Foreman to generate the Kickstart default template.

* Disabling tokens in the settings will remove the requirement though. To generate the template, you must ensure the host can be authenticated in Foreman. Usually, this is done by MAC / IP.
* If the image already contains generated kickstart, you don't have to call Foreman for the generated template; you don't need the token.
* Otherwise you need it.

stejskalleos:
> TOC of the guide

What's TOC?

guides/doc-Provisioning_Virtual/master.adoc
stejskalleos: I vote for making this guide "working" for all stable releases. I'll keep it in mind during the review

guides/common/modules/proc_creating-compute-profiles.adoc
```diff
@@ -13,8 +13,12 @@ A default installation of {ProjectName} contains three predefined profiles:
 
 You can apply compute profiles to all supported compute resources:
 
+ifdef::provisioning,provisioning-cloud[]
```
stejskalleos, line 16: Are these flags used during the documentation generation?
What's the reasoning for not including these blocks in the end docs?

guides/common/modules/con_provisioning-virtual-machines-on-kubevirt.adoc
```diff
@@ -27,6 +27,4 @@ include::snip_prerequisites-common-compute-resource.adoc[]
 Ensure that no other DHCP services run on this network to avoid conflicts with {SmartProxyServer}.
 For more information about network service configuration for {SmartProxyServers}, see {ProvisioningDocURL}preparing-networking[Preparing networking] in _{ProvisioningDocTitle}_.
 
-.Additional resources
-* For a list of permissions a non-admin user requires to provision hosts, see xref:permissions-required-to-provision-hosts_provisioning[].
```
stejskalleos, line 31: Note to myself: snip_prerequisites-common-compute-resource.adoc

stejskalleos (changes requested):

stejskalleos (changes requested):
I went over the [index-katello](https://theforeman-foreman-documentation-preview-pr-3608.surge.sh/nightly/Provisioning_Virtual/index-katello.htm), here are my comments:

**1. Introduction to provisioning**
I would move this whole section into a new Intro guide or something similar. Or are we going to repeat this introduction section in every provisioning guide?

**2. Preparing client platforms**
Can we move it to the separate guide? It's not limited to virtual provisioning, but applies to all provisioning guides.

**2.1 Creating an operating system for...**
Can we remove the whole section? Creating an operating system in Foreman is not difficult; it's quite easy. Or at least remove four examples and have one generic.

Idea: Sections `5. Security settings`, `6. Provisioning contexts` & `7. Using noVNC to access virtual machines could be moved after the provisioning of virtual machines ... guides. They are not essential for the setup. WDYT?

Note for other reviewers: I focused solely on the content structure; I did not review the content itself, so my review is not a technical ACK.

guides/common/modules/proc_creating-compute-profiles.adoc
Lennonka: There will be a third guide for provisioning cloud instances.

guides/common/assembly_security-settings.adoc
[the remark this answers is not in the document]
Lennonka: Image-based provisioning does not render Kickstart default template. It renders only "finish" or "cloud-init" templates AFAIK.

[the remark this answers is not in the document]
stejskalleos: > Which Kickstart default template? 

Sorry for the confusion. The authentication between the host and Foreman when requesting templates is not limited to Kickstart templates only, but applies to every provisioning template kind, of course.

stejskalleos:
> Chapter 1 is repeated in every guide, yes. I wanted to avoid having a 4th guide for general introduction.

IMHO, I don't think having a special introduction guide is a bad thing, quite the contrary, I would welcome it. Currently, the guide begins with the sentence: "_a process that starts with a bare physical or virtual machine_." Why talk about physical machines when the guide is literally called "Provisioning Virtual machines". I vote for intro guide.

> Chapter 2 is also repeated in all 3 guides precisely for that reason.

We could move it to the new provisioning intro guide. (as mentioned above)

> Well, users have to set their provisioning context before they start creating hosts, don't they?

Well, they don't, kinda. From the docs: _If you have not selected an organization and location to use, the menu displays: Any Organization and Any Location._

By default, the Default org and loc are selected, so unless they touch it, they don't have to care about the context. Changing & configuring contexts might be optional for some users.

apinnick:
@stejskalleos @Lennonka 
Chapter 1 is already in the [Planning guide](https://docs.redhat.com/en/documentation/red_hat_satellite/6.16/html-single/overview_concepts_and_deployment_considerations/index#Provisioning-Methods-in-Satellite_planning). You can link to it instead of repeating it in each guide.

Chapter 2 is very generic. In my experience, users tend to ignore generic procedures and go straight to the relevant sections like VMware or KVM. If these procedures are necessary, I recommend rewriting them to make them more concise (obvious settings like "Description", "Name", "Family" do not need to be documented). Then put them into the sections with the providers.

"Creating Operating Systems" procedure looks like it's for adding a custom operating system, so you might want to update the title and rewrite the introduction.

guides/doc-Provisioning_Virtual/master.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+include::common/assembly_provisioning-virtual-machines-vmware.adoc[leveloffset=+1]
+
+ifndef::satellite[]
+include::common/assembly_provisioning-virtual-machines-proxmox.adoc[leveloffset=+1]
+endif::[]
+
+include::common/assembly_provisioning-virtual-machines-kubevirt.adoc[leveloffset=+
+
+include::common/modules/proc_using-novnc-to-access-virtual-machines.adoc[leveloffset=+1]1]
```
maximiliankolb, line 36:
```suggestion
include::common/modules/proc_using-novnc-to-access-virtual-machines.adoc[leveloffset=+1]
```

guides/doc-Provisioning_Virtual/master.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+include::common/assembly_provisioning-virtual-machines-kvm.adoc[leveloffset=+1]
+
+include::common/assembly_provisioning-virtual-machines-vmware.adoc[leveloffset=+1]
+
+ifndef::satellite[]
+include::common/assembly_provisioning-virtual-machines-proxmox.adoc[leveloffset=+1]
+endif::[]
+
+include::common/assembly_provisioning-virtual-machines-kubevirt.adoc[leveloffset=+
```
maximiliankolb, line 34:
```suggestion
include::common/assembly_provisioning-virtual-machines-kubevirt.adoc[leveloffset=+1]
```

maximiliankolb:
>     * Chapter 2 is also repeated in all 3 guides precisely for that reason.
>       @maximiliankolb What do you think about reducing the number of procedures 2.2-2.5? Do you want to keep them?

I want to reduce them. I looked into a way to show default OS repositories in tables inhttps://github.com/theforeman/foreman-documentation/pull/3804. The info to create Operating System entries in Foreman/Katello would be a logical next step. I will continue with this if/when there's consensus on how to handle default OS content in 3804.

Another thing I found on a local build: chapter 10 mentions oVirt but oVirt is not in the list of compute resources. Probably related to/done with https://github.com/theforeman/foreman-documentation/pull/3800

> 2.1 Creating an operating system for...
Can we remove the whole section? Creating an operating system in Foreman is not difficult; it's quite easy. Or at least remove four examples and have one generic.

Well... The generic example is aimed at RHEL but I want to show examples for Debian/Ubuntu/SLES/EL too. As mentioned above, I suggest to keep it for now with this PR and give me some time to rework it.

ekohl (changes requested):
I want to weigh in whether we want this guide in the first place because I don't think we should.

Foreman has a goal to make provisioning as similar as possible, whether you provision on bare metal or on virtual machines. This guide goes against that spirit.

I can see sufficient differences between image based deployments and installer based deployments that you can split those guides, but VM based provisioning is wrong IMHO. Especially if you also want to add a guide for cloud deployments. Clouds are VMs and should be treated similar.

When you deploy on VMs you can, depending on the hypervisor use both installer based provisioning and image based provisioning. This is the wrong separation.

Lennonka:
@ekohl Can you please help me figure out what to do about this?

Lennonka:
We will likely go in a different direction. Closing this.

PR #3614, opened 2025-01-24
Move DR overview and reuse it in Admin

aneta-petrova:

* Moving disaster recovery overview from a separate assembly in the Planning guide to the deployment path assembly
* Including the disaster recovery assembly in the Admin guide
* Renaming the disaster recovery assembly

Except for the renaming, the content stays the same.

As I started documenting the first of the DR scenarios (https://github.com/theforeman/foreman-documentation/pull/3615), this structure seemed to work better.

I'm not entirely sure about the placement in the Admin guide and want to use this PR to discuss it.

#### Checklists


asteflova:
@apinnick What do you think is the best place for disaster planning + testing + recovery? This PR places the assembly in the Admin guide -- at the very end because considering how long that guide is, I think it's kind of pointless to play with the order of the chapters.

I considered a separate guide and.. I don't know.. maybe?

(To explain a bit more: For now, this is still just an assembly with an overview. I will be adding detailed procedures for each of the DR scenarios. You can see a preview of the first scenario here if it helps https://github.com/theforeman/foreman-documentation/pull/3615. Also, just in case it isn't clear from the PR: My plan is to add the scenarios to Admin only, not to Planning.)

asteflova:
Let's skip tech review because this just moves things around.

maximiliankolb (approved):

guides/common/modules/con_overview-of-recommended-disaster-recovery-plans.adoc
```diff
@@ -54,3 +54,16 @@ You must ensure that content synchronization and content view creation are synch
 Content drift occurs when available content deviates from the intended state defined by a content view.
 If you fail to prevent content drift, expect inconsistency in the content that is available to hosts.
 endif::[]
+
+ifeval::["{context}" == "planning"]
```
Lennonka, line 58: I'd strongly recommend using the `:planning:` attribute instead of context. We want to get rid of contexts where they aren't needed for reuse.
```suggestion
ifdef::planning[]
```
Please, define it in the `master.adoc`.

Lennonka (approved):
LGTM. My suggestion is not blocking your PR but highly recommended :)

guides/common/modules/con_overview-of-recommended-disaster-recovery-plans.adoc
maximiliankolb: This is a very nice idea which simplifies things a lot! Thanks Lena.

PR #3617, opened 2025-01-27
Replace ping command and output with dig command and output for DNS resolution

Levi-Leah:

Replaced the ping command and its related output with the dig command, as dig is more appropriate for demonstrating DNS resolution.

Since the goal is to demonstrate DNS resolution, the dig command is a more accurate and reliable tool for this purpose.

Bug [SAT-28901](https://issues.redhat.com/browse/SAT-28901) (public)

Alternatively, ping command can be replaced with nslookup command.

#### Checklists

### DIFF
--- a/guides/common/modules/proc_verifying-dns-resolution.adoc
+++ b/guides/common/modules/proc_verifying-dns-resolution.adoc
@@ -32,7 +32,7 @@ my_system.domain.com.			0	IN	A	XXX.XXX.X.XX
 my_system.domain.com.			0	IN	A	XX.XX.XXX.X
 ----
 
-. To avoid discrepancies with static and transient host names, set all the host names on the system with the following command:
+. To avoid discrepancies with static and transient host names, set all the host names on the system:
 +
 [options="nowrap" subs="+quotes"]
 ----
--- a/guides/common/modules/proc_verifying-dns-resolution.adoc
+++ b/guides/common/modules/proc_verifying-dns-resolution.adoc
@@ -9,33 +9,30 @@ Verify the full forward and reverse DNS resolution using a fully-qualified domai
 +
 [options="nowrap"]
 ----
-# ping -c1 localhost
-# ping -c1 `hostname -f` # my_system.domain.com
+# dig localhost
+# dig `hostname -f` # my_system.domain.com
 ----
 +
-Successful name resolution results in output similar to the following:
+When name resolution succeeds, the output resembles the following:
 +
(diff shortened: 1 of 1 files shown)


guides/common/modules/proc_verifying-dns-resolution.adoc
```diff
@@ -9,33 +9,30 @@ Verify the full forward and reverse DNS resolution using a fully-qualified domai
...
+# dig  `hostname -f`
+; <<>> DiG 9.18.30 <<>> my_system.domain.com
+...
+;; ANSWER SECTION:
+my_system.domain.com.			0	IN	A	XXX.XXX.X.XX
+my_system.domain.com.			0	IN	A	XX.XX.XXX.X
 ----
 
-. To avoid discrepancies with static and transient host names, set all the host names on the system by entering the following command:
+. To avoid discrepancies with static and transient host names, set all the host names on the system with the following command:
```
aneta-petrova, line 35: This might work even better:
```suggestion
. To avoid discrepancies with static and transient host names, set all the host names on the system:
```

aneta-petrova:
@Levi-Leah In the cherry-pick list, you selected Satellite versions only. I don't think there is a need to do that, is there? If not, then please check all versions, even the upstream-only ones.

ekohl (commented):
Dig is not installed by default. I'm not sure what would be best here. I'm considering a more engineering solution where we provide a check instead.

@evgeni any preference?

evgeni:
:+1: on a check -- never liked the `ping` calls here (and yeah, `dig` is not installed by default, and we should probably not use it)

ekohl:
I suppose the best place for that check is in the installer. We already have https://github.com/theforeman/foreman-installer/blob/develop/checks/hostname.rb so I suppose we should enhance it to actually catch the problematic situation.

Should we also do something in [foreman_maintain](https://github.com/theforeman/foreman_maintain)? If we become stricter it may break upgrades.

My preference would be to make it an engineering task first (given there's no hurry; this has been unchanged for ages) and then once the check is in place we update this PR, probably by replacing the whole verification step.

aneta-petrova:
To summarize the conversation in https://issues.redhat.com/browse/SAT-28901 (private): The proper way to address the issue is a code change in the installer. The ultimate goal is to remove this entire documentation section -- after the mentioned code change is implemented. That makes this PR no longer necessary. Thanks @Levi-Leah for investigating the original issue!

PR #3645, opened 2025-02-05
Dissociate Smart Proxies unmanaged IPv6

Lennonka:

Adding important instructions for subnets and domains in IPv6 networks, where DHCP and DNS cannot be managed by Foreman
...and a couple of cosmetic improvements

- Users need to know this if they work in IPv6 networks.
- "Once no proxies are associated, the system will allow creating hosts without IPs specified." [SAT-30604](https://issues.redhat.com/browse/SAT-30604)

#### Checklists

### DIFF
--- a/guides/common/modules/proc_adding-a-domain.adoc
+++ b/guides/common/modules/proc_adding-a-domain.adoc
@@ -22,6 +22,11 @@ If you experience timeouts during DNS conflict resolution, check the following s
 * The domain name must have a Start of Authority (SOA) record available from {ProjectServer}.
 * The system resolver in the `/etc/resolv.conf` file must have a valid and working configuration.
 
+[IMPORTANT]
+====
+In an IPv6 network with unmanaged DNS, do not assign DNS {SmartProxy} to your domain.
+====
+
 To use the CLI instead of the {ProjectWebUI}, see the xref:cli-adding-a-domain_{context}[].
 
 .Procedure
--- a/guides/common/modules/proc_adding-a-domain.adoc
+++ b/guides/common/modules/proc_adding-a-domain.adoc
@@ -44,14 +44,13 @@ For example, user defined Boolean or string parameters to use in templates.
 .CLI procedure
 * Use the `hammer domain create` command to create a domain:
 +
-[options="nowrap" subs="+quotes"]
+[options="nowrap" subs="+quotes,attributes"]
 ----
 $ hammer domain create \
 --description "_My_Domain_" \
---dns-id _My_DNS_ID_ \
+--dns-id _My_DNS_{smart-proxy-context}_ID_ \
 --locations "_My_Location_" \
 --name "_my-domain.tld_" \
 --organizations "_My_Organization_"
 ----
 
-In this example, the `--dns-id` option uses `1`, which is the ID of your integrated {SmartProxy} on {ProjectServer}.
--- a/guides/common/modules/proc_adding-a-subnet.adoc
+++ b/guides/common/modules/proc_adding-a-subnet.adoc
@@ -43,13 +43,13 @@ For example, user defined Boolean or string parameters to use in templates.
 .CLI procedure
 * Create the subnet with the following command:
 +
-[options="nowrap" subs="+quotes"]
+[options="nowrap" subs="+quotes,attributes"]
 ----
 $ hammer subnet create \
 --boot-mode "DHCP" \
 --description "_My_Description_" \
---dhcp-id _My_DHCP_ID_ \
---dns-id _My_DNS_ID_ \
+--dhcp-id _My_DHCP_{smart-proxy-context}_ID_ \
+--dns-id _My_DNS_{smart-proxy-context}_ID_ \
 --dns-primary "192.168.140.2" \
 --dns-secondary "8.8.8.8" \
 --domains "_my-domain.tld" \
@@ -64,8 +64,3 @@ $ hammer subnet create \
 --tftp-id _My_TFTP_ID_ \
 --to "192.168.140.250"
 ----
-
-[NOTE]
-====
-In this example, the `--dhcp-id`, `--dns-id`, and `--tftp-id` options use 1, which is the ID of the integrated {SmartProxy} in {ProjectServer}.
-====
--- a/guides/common/modules/proc_adding-a-subnet.adoc
+++ b/guides/common/modules/proc_adding-a-subnet.adoc
@@ -22,7 +22,7 @@ To use the CLI instead of the {ProjectWebUI}, see the xref:cli-adding-a-subnet_{
 . In the *Primary DNS server* field, enter a primary DNS for the subnet.
 . In the *Secondary DNS server*, enter a secondary DNS for the subnet.
 . From the *IPAM* list, select the method that you want to use for IP address management (IPAM).
-For more information about IPAM, see xref:preparing-networking[].
+For more information about IPAM, see xref:Network_Resources_{context}[].
 +
 . Enter the information for the IPAM method that you select.
 ifdef::satellite,orcharhino[]
--- a/guides/common/modules/proc_adding-a-subnet.adoc
+++ b/guides/common/modules/proc_adding-a-subnet.adoc
@@ -4,6 +4,11 @@
 You must add information for each of your subnets to {ProjectServer} because {Project} configures interfaces for new hosts.
 To configure interfaces, {ProjectServer} must have all the information about the network that connects these interfaces.
 
+[IMPORTANT]
+====
+In an IPv6 network with unmanaged DHCP and DNS, do not assign DHCP {SmartProxy} and Reverse DNS {SmartProxy} to your subnet.
+====
+
 To use the CLI instead of the {ProjectWebUI}, see the xref:cli-adding-a-subnet_{context}[].
 
 .Procedure


ShimShtein:
ACK, makes sense

maximiliankolb (approved):
LGTM

ekohl:
I think it's a bug this can be done in the first place: the DHCP management is IPv4-only and we know the subnet type

guides/common/modules/proc_adding-a-subnet.adoc
```diff
@@ -4,6 +4,11 @@
 You must add information for each of your subnets to {ProjectServer} because {Project} configures interfaces for new hosts.
 To configure interfaces, {ProjectServer} must have all the information about the network that connects these interfaces.
 
+[IMPORTANT]
+====
+In an IPv6 network with unmanaged DHCP and DNS, do not assign DHCP {SmartProxy} and Reverse DNS {SmartProxy} to your subnet.
```
ekohl, line 9: This is IMHO a bug that the user can select this in the first place. Is there work tracked to make that impossible?

guides/common/modules/proc_adding-a-domain.adoc
```diff
@@ -22,6 +22,11 @@ If you experience timeouts during DNS conflict resolution, check the following s
 * The domain name must have a Start of Authority (SOA) record available from {ProjectServer}.
 * The system resolver in the `/etc/resolv.conf` file must have a valid and working configuration.
 
+[IMPORTANT]
+====
+In an IPv6 network with unmanaged DNS, do not assign DNS {SmartProxy} to your domain.
```
ekohl, line 27: I don't understand this sentence. Why shouldn't a user do this? Why is it IPv6 specific?

guides/common/modules/proc_adding-a-subnet.adoc
ShimShtein: I wouldn't mark it as a bug yet. We may start supporting DHCPv6 management in the future. The same goes for DNS, maybe we will be able to support managed DNS for IPv6 networks later on (for example in retroactive mode, where we update the DNS once we get the facts from the host).

guides/common/modules/proc_adding-a-domain.adoc
ShimShtein: While it's not IPv6 specific, it is relevant here. If a user assigns a DNS proxy to the subnet/domain, Foreman will require IP address, which cannot be set in IPv6 scenario (at least for the foreseeable future).

ekohl: The Smart Proxy can manage AAAA records and IPv6 PTR records so I still don't see why it's a problem.

Having said that, for unmanaged DNS you should **never** set a DNS Smart Proxy anyway so I'm still deeply confused.

guides/common/modules/proc_adding-a-subnet.adoc
ekohl: Until we develop the feature, we should make it impossible set unsupported options. Because of that I'd qualify it as a bug.

guides/common/modules/proc_adding-a-domain.adoc
ShimShtein: I think making it clear that the user should not set the DNS smart proxy as long as there is no IPv6 set on the machine is useful here. Otherwise the host will require setting an IPv6 address for the host, which is not currently possible (we only support unmanaged DHCPv6 for provisioning).

guides/common/modules/proc_adding-a-subnet.adoc
ShimShtein: Let's create an issue on Redmine then. Sounds fine by me.

guides/common/modules/proc_adding-a-domain.adoc
ekohl: But users can assign an IPv6 address statically or rely on EUI64. For dual stack that is fine.

Lennonka:
So I can merge this, right?

guides/common/modules/proc_adding-a-subnet.adoc
aneta-petrova: It looks like this thread will be resolved by an issue on the code side. I'll resolve this thread then (but it would be nice if you could add a link to the Redmine issue for reference).

guides/common/modules/proc_adding-a-domain.adoc
ShimShtein: Yes they can assign addresses, but not set the proxy. The proxy will just cause errors.

aneta-petrova: @ekohl Do you have a suggestion for how to rewrite the note to be clearer from your perspective? If not, I'd prefer we go ahead and merge.

ekohl: The whole sentence needs to be dropped IMHO. For DNS it should work. If not, that's a bad bug.

@ShimShtein perhaps you're confused with DHCP, which indeed can't work with IPv6. But there's code that should prevent it from being a problem. I'm not saying it works, but there was the intention.

https://github.com/theforeman/foreman/commit/990dee696f58b9fa6035b239152223b073999488 was the first step to introduce IPv6 support. There it implemented [`supports_ipam_mode?`](https://github.com/theforeman/foreman/blob/2a81d51f7793f82bc5687952789e716b278083ff/app/models/subnet.rb#L407-L409), which relies on `supported_ipam_modes`. That is [delegated](https://github.com/theforeman/foreman/blob/2a81d51f7793f82bc5687952789e716b278083ff/app/models/subnet.rb#L123 to the class).

If we go to the specific class, you can see that for [IPv4 it lists `dhcp`](https://github.com/theforeman/foreman/blob/2a81d51f7793f82bc5687952789e716b278083ff/app/models/subnet/ipv4.rb#L28-L30) while for [IPv6 it doesn't](https://github.com/theforeman/foreman/blob/2a81d51f7793f82bc5687952789e716b278083ff/app/models/subnet/ipv6.rb#L26-L28).

To wrap this up you have the [`belongs_to_proxy`](https://github.com/theforeman/foreman/blob/2a81d51f7793f82bc5687952789e716b278083ff/app/models/subnet.rb#L45-L50) statement to add the DHCP proxy. Note the `if` that should return `true` for IPv4 while returning `false` for IPv6.

If this code works as intended, I can't see how Foreman would ever attempt to send DHCP related API calls to the proxy even if it somehow was assigned.

guides/common/modules/proc_adding-a-subnet.adoc
ekohl: Unresolving because reverse DNS should work. There is explicit code for it and I'd like to see some reproducer that shows it's broken before documenting it. Then we can, while we fix the bug, document a known issue.

aneta-petrova:
Unsetting `tech review` because there is still interest in continuing the discussion.

guides/common/modules/proc_adding-a-domain.adoc
ShimShtein: The problem here is not the API calls, but the validation on the `ip6` field. Once the domain has a DNS proxy assigned to it, any host that will use that domain will require either `ip` or `ip6` field to be not empty.
If we are talking about IPv6-only network, and since we can't manage DHCPv6 (at least for now) there won't be an option to save the host record.

guides/common/modules/proc_adding-a-subnet.adoc
ekohl: Interesting, but I'm having trouble wrapping my head around why this is an issue. Is there a bug report with the reproducer steps? Is it something like:

* User creates a new host
* User doesn't select a compute resource which can provide an IPv6 address
* User adds a NIC
* User selects IPv6 subnet with reverse DNS set
* User doesn't set an IPv6 address

Now the user is presented with an error, but how is that a problem? When is selecting an IPv6 subnet without setting an IPv6 address valid?

If you are provisioning with IPv6 with unmanaged DHCPv6 then I'd assume you don't set the IPv6 subnet and instead rely on it to be set later when it's set via facts: https://github.com/theforeman/foreman/blob/2a81d51f7793f82bc5687952789e716b278083ff/app/models/host/base.rb#L512-L517

guides/common/modules/proc_adding-a-domain.adoc
ekohl: AFAIK Foreman has never supported provisioning where the IP wasn't known in advance. Shouldn't the workflow recommend users to use EUI-64 where the IP address is based on the mac address? It's a reasonable expectation that in an unmanaged DHCP environment the assigned address is exactly that.

You could even mandate with unmanaged DHCP that SLAAC is also used to assign IP addresses.

Lennonka:
@ShimShtein @ekohl What are we gonna do about this?

ShimShtein:
@ekohl I would say that we should take the discussion to our forum, and unblock this PR. 
IMHO it just describes the current state, and how to work with it, so it gives enough benefit. Later on we can improve the documentation further.

ekohl:
There are 3 changes in this PR:
* Stylistic of using the correct DNS/TFTP ID
* Changing the xref
* Adding warnings for IPv6 management

The first 2 looks correct to me, but I still maintain that the warnings for IPv6 management are invalid.

Combining multiple changes in the same PR is generally a bad thing because if there's an issue in one of the changes means that none of the changes get in. That's why I always hammer on separating changes when possible.

My recommendation would be to submit the first 2 changes as a separate PR and keep this open to resolve the discussion. I think having asking more people for input on the forum to get more opinions is a good suggestion from @ShimShtein.

PR #3651, opened 2025-02-10
Add note on enabling minor version of BaseOS and AppStream RPMs

Levi-Leah:

Ads a note to enable minor version of BaseOS and AppStream RPMs.

Based on the the leapp supported upgrade path we currently need 9.5 minor repositories; upgrade fails otherwise

Bug [SAT-29652](https://issues.redhat.com/browse/SAT-29652) (public)

#### Checklists

### DIFF
--- a/guides/common/modules/proc_upgrading-project-or-proxy-in-place-using-leapp.adoc
+++ b/guides/common/modules/proc_upgrading-project-or-proxy-in-place-using-leapp.adoc
@@ -20,11 +20,15 @@ ifdef::satellite[]
 +
 Red Hat Enterprise Linux 9 for x86_64 - BaseOS (RPMs):: `{RepoRHEL9BaseOS}`
 +
-Ensure this repository is enabled for the latest minor version, as specified in the {RHELDocsBaseURL}9/html-single/upgrading_from_rhel_8_to_rhel_9/index#con_supported-upgrade-paths_upgrading-from-rhel-8-to-rhel-9[Supported upgrade paths] in _Upgrading from RHEL 8 to RHEL 9_.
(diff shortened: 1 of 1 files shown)


evgeni:
We document that people should run `subscription-manager release --unset` after the upgrade (in the connected case, and capsules are always considered "connected"), if they *only* have the 9.Y repos synced, this will break for them. So *technically* you need both repo versions available: "9" and "9.Y".

aneta-petrova:
Hi @evgeni, I'd like to wrap this PR up on Levi's behalf. I don't know if you've already acked this one (I don't see an ack here or in Jira). Can you please take a look and let me know if this is good to merge so that I can be sure?

guides/common/modules/proc_upgrading-project-or-proxy-in-place-using-leapp.adoc
```diff
@@ -20,8 +20,16 @@ ifdef::satellite[]
 +
 Red Hat Enterprise Linux 9 for x86_64 - BaseOS (RPMs):: `{RepoRHEL9BaseOS}`
 +
+Ensure this repository is enabled for the latest minor and major versions.
```
evgeni, line 23: This is technically correct, but I wonder if the reader will understand what we mean.

Take this screenshot (and ignore the ARM part, I have x86 repos already enabled, so they are not displayed in the list)
image

They need to enable "aarch64 9" (note there is no dot-something!) and "aarch64 9.5"

Wonder if the RHEL docs have a name for those repositories.

aneta-petrova:
Because it looks like more significant changes are needed, let's continue this in https://github.com/theforeman/foreman-documentation/pull/3708.

PR #3657, opened 2025-02-12
Simplify flavor+version navigation paths

aneta-petrova:

This is an attempt to simplify the way users can navigate docs.theforeman.org, namely by:

* Removing the current flavor dropdowns from the top menu bar and replacing them with new version pages, each with collapsible lists
* Removing a few links from the main landing page and adding them to the top menu bar instead

Wherever possible, I'm making use of native AsciiDoc features (hence the use of collapsibles) so that the result is easy to understand and modify by any writer familiar with AsciiDoc.

I'm trying to address these issues that currently affect docs.theforeman.org:

* Users can't use full-text search to find the guide they need
* Navigating between the flavors and versions is not straightforward enough

The idea is to split the navigation path into two steps: First choose the version, then choose the flavor. This is to give users one choice at a time. With the existing navigation, they can choose either the version or the flavor at the beginning of their navigation path, which is confusing.

#### Checklists

### DIFF
--- a/web/content/js/nav.js
+++ b/web/content/js/nav.js
@@ -23,18 +23,6 @@ function buildNavigation() {
 </button>
 <ul class="nav-menu">
   <li class="nav-item"><a href="https://theforeman.org/">About Foreman</a></li>`
-    + navBuilds.map(function(build){
-      return(
-        `<li class="nav-item dropdown">
-          <a href="#" data-action="dropdown-toggle">${build.title}</a>
-            <div class="dropdown-menu">`
-            + build.guides.map(function(guide){
-              const url = `/${currentVer}/${guide.path}/${build.filename}`;
-              return `<div class="dropdown-div"><a class="dropdown-item" href="${url}">${guide.title}</a></div>`;
-            }).join("")
-            +`</div>
-        </li>`
-      )}).join("")
     + `<li class="nav-item dropdown">
         <a href="#" data-action="dropdown-toggle">Version ${currentVer}</a>
         <div class="dropdown-menu dropdown-menu-left">`
--- a/web/content/index.adoc.erb
+++ b/web/content/index.adoc.erb
@@ -12,9 +12,9 @@ For official documentation, see link:https://theforeman.org/manuals/latest/index
 For Foreman plugin overview, see link:https://theforeman.org/plugins/[Plugins].
 ====
 
-== Navigating the docs.theforeman.org site
+=== Navigating the docs.theforeman.org site
 
-Use the top menu bar to select your release version and flavor.
+Start by selecting your release version in the top menu bar.
 
 For questions, visit the https://community.theforeman.org/c/support/10[support forum].
 For contribution, visit the https://github.com/theforeman/foreman-documentation[GitHub repository].
--- a/web/releases/2.4.adoc
+++ b/web/releases/2.4.adoc
@@ -2,7 +2,7 @@
 :KATELLO_VER: <%= @item[:katello] %>
 
 == Foreman {FOREMAN_VER} and Katello {KATELLO_VER}
-.Show all guides
+.Show documentation
 .Upgrading
 * link:/{FOREMAN_VER}/Upgrading_and_Updating/index-katello.html[Upgrading and Updating]
 
--- a/web/releases/2.4.adoc
+++ b/web/releases/2.4.adoc
@@ -9,7 +9,7 @@
 .Quickstart
 * link:/{FOREMAN_VER}/Quickstart_Guide/index-katello.html[Quickstart Guide]
 
-.Deployment and installation
+.Deploying Foreman and Katello
 * link:/{FOREMAN_VER}/Planning_Guide/index-katello.html[Planning Guide]
 * link:/{FOREMAN_VER}/Installing_Server_on_Red_Hat/index-katello.html[Installing Katello Server]
 * link:/{FOREMAN_VER}/Installing_Proxy_on_Red_Hat/index-katello.html[Installing Smart Proxy with Content]
@@ -17,11 +17,11 @@
 * link:/{FOREMAN_VER}/Deploying_on_AWS/index-katello.html[Deploying Foreman on Amazon Web Services]
 * link:/{FOREMAN_VER}/Application_Centric_Deployment/index-katello.html[Application Centric Deployment]
 
-.Server administration
+.Administering Foreman server
 * link:/{FOREMAN_VER}/Administering_Red_Hat_Satellite/index-katello.html[Administering Foreman]
 * link:/{FOREMAN_VER}/Content_Management_Guide/index-katello.html[Content Management Guide]
 * link:/{FOREMAN_VER}/Configuring_Ansible/index-katello.html[Configuring Foreman to use Ansible]
 
-.Host administration
+.Administering hosts
 * link:/{FOREMAN_VER}/Managing_Hosts/index-katello.html[Managing Hosts Guide]
 * link:/{FOREMAN_VER}/Provisioning_Guide/index-katello.html[Provisioning Guide]
--- a/web/releases/2.4.adoc
+++ b/web/releases/2.4.adoc
@@ -2,17 +2,26 @@
 :KATELLO_VER: <%= @item[:katello] %>
 
 == Foreman {FOREMAN_VER} and Katello {KATELLO_VER}
+.Show all guides
+.Upgrading
+* link:/{FOREMAN_VER}/Upgrading_and_Updating/index-katello.html[Upgrading and Updating]
 
-Use the top menu bar for navigation for all guides.
+.Quickstart
+* link:/{FOREMAN_VER}/Quickstart_Guide/index-katello.html[Quickstart Guide]
 
-=== Quickstart
+.Deployment and installation
+* link:/{FOREMAN_VER}/Planning_Guide/index-katello.html[Planning Guide]
+* link:/{FOREMAN_VER}/Installing_Server_on_Red_Hat/index-katello.html[Installing Katello Server]
+* link:/{FOREMAN_VER}/Installing_Proxy_on_Red_Hat/index-katello.html[Installing Smart Proxy with Content]
+* link:/{FOREMAN_VER}/Configuring_Load_Balancer/index-katello.html[Configuring Smart Proxies with a Load Balancer]
+* link:/{FOREMAN_VER}/Deploying_on_AWS/index-katello.html[Deploying Foreman on Amazon Web Services]
+* link:/{FOREMAN_VER}/Application_Centric_Deployment/index-katello.html[Application Centric Deployment]
 
-* link:/{FOREMAN_VER}/Quickstart_Guide/index-katello.html[Katello on RHEL/CentOS Quickstart Guide]
+.Server administration
+* link:/{FOREMAN_VER}/Administering_Red_Hat_Satellite/index-katello.html[Administering Foreman]
+* link:/{FOREMAN_VER}/Content_Management_Guide/index-katello.html[Content Management Guide]
+* link:/{FOREMAN_VER}/Configuring_Ansible/index-katello.html[Configuring Foreman to use Ansible]
 
-=== Installation
-
-* link:/{FOREMAN_VER}/Installing_Server_on_Red_Hat/index-katello.html[Installing Katello on RHEL/CentOS]
-
-=== Upgrading
-
-* link:/{FOREMAN_VER}/Upgrading_and_Updating/index-katello.html[Upgrading and Updating Katello on RHEL/CentOS]
+.Host administration
+* link:/{FOREMAN_VER}/Managing_Hosts/index-katello.html[Managing Hosts Guide]
+* link:/{FOREMAN_VER}/Provisioning_Guide/index-katello.html[Provisioning Guide]
--- a/web/releases/2.5.adoc
+++ b/web/releases/2.5.adoc
@@ -2,17 +2,25 @@
 :KATELLO_VER: <%= @item[:katello] %>
 
 == Foreman {FOREMAN_VER} and Katello {KATELLO_VER}
+.Upgrading
+* link:/{FOREMAN_VER}/Upgrading_and_Updating/index-katello.html[Upgrading and Updating]
 
-Use the top menu bar for navigation for all guides.
+.Quickstart
+* link:/{FOREMAN_VER}/Quickstart_Guide/index-katello.html[Quickstart Guide]
 
-=== Quickstart
+.Deployment and installation
+* link:/{FOREMAN_VER}/Planning_Guide/index-katello.html[Planning Guide]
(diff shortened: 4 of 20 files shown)


aneta-petrova:
Notes from today's triage: Generally, ack, this is seen as an improvement over the current website navigation.

Current blocker: No all version files (the ones linked from the version dropdown) are currently populated with a list of guides. I will add this.

Also, I want to experiment with different options for how to present the collapsible lists, to find out if it's possible to keep the Katello list uncollapsed by default.

web/content/js/nav.js
```diff
@@ -12,7 +12,7 @@ function buildNavigation() {
 
   return `<nav>
   <a href="/">
-    <img class="logo" src="/img/helmet.svg" alt="Home">
+    <img class="logo" src="/img/helmet.png" alt="Home">
```
ekohl, line 15: What's the reason to replace the SVG with a PNG?

ekohl (commented):
In general I agree with the drop down menus being gone because they're getting too long and don't scale.

In https://github.com/theforeman/foreman-documentation/pull/1204 I played around with having those landing pages within the branches themselves. It may be worth investigating that again. I think long term it can easy maintenance.

aneta-petrova:
> In #1204 I played around with having those landing pages within the branches themselves. It may be worth investigating that again. I think long term it can easy maintenance.

That is what worries me most about the solution I propose: That maintaining a list of guide links manually for each version is heavy maintenance. But then again, I try to keep in mind that we also need a solution that is easy to understand and update by any writer.

ekohl:
My idea comes to have something like `guides/index.adoc` in every branch and then somehow gather all of those when building the website. That way the index lives very close to the actual guides. For me it's hard to estimate if that would be easy to understand by any writer, but I think it would because (as you also said) it's using tooling writers are already familiar and in a directory you already work in.

To be clear: I don't mean we need to block your effort on it and it can be done as a follow up.

web/content/js/nav.js
aneta-petrova: The reason is that I got carried away. I'm about to break this commit and a couple of more into a separate PR and will make a note in there to address this.

ekohl: Fair enough. Happens to me all the time :)

aneta-petrova:
> Also, I want to experiment with different options for how to present the collapsible lists, to find out if it's possible to keep the Katello list uncollapsed by default.

@maximiliankolb To follow up on this, these are two things I tried:

Screenshot From 2025-02-13 18-35-50

Screenshot From 2025-02-13 18-36-05

I have to say that I still prefer the current, collapsed by default. To me, it still seems like the simplest and cleanest solution, and I don't think the one extra click is too much of an issue:

Screenshot From 2025-02-13 18-36-18

I'm going to proceed with the collapsed-by-default approach and, as discussed at today's triage, we can always improve later :)

ekohl:
Something I thought about was taking more of the tabs approach, but no idea if it's actually feasible/usable.

Basically have a header that's `Foreman $version` and then 3 big tabs:

* Foreman on EL
* Foreman on Debian/Ubuntu
* Katello

Depending on the tab, you see the list of guides. You can also consider adding logos to the tabs for additional visibility. The asciidoc tab plugin can use cookies to store the user's preference so if you navigate to other versions it should remember what the user has.

aneta-petrova:
I couldn't get tabs to work: I tried on my own and then also when I checked out your web branch. I never got the tabs, always just the example block. That's when I thought of collapsible lists as an acceptable alternative.

aneta-petrova:
I'd say this is now ready for review. Instructions for testing the website locally are in /web/README.md

This is what it looks like:

* The main page. Note the missing flavor dropdowns, they are not necessary anymore thanks to the collapsibles on individual version pages (see later screenshots):

Screenshot From 2025-04-30 10-06-23

* Collapsed list of guides for one of the versions:

Screenshot From 2025-04-30 10-07-20

* The same page, with one list expanded:

Screenshot From 2025-04-30 10-07-48

* The version dropdown remains unchanged:

Screenshot From 2025-04-30 10-08-24

aneta-petrova:
I'd appreciate a review especially from @maximiliankolb and @ekohl. Can you please take a look when you have a moment?

maximiliankolb:
> I'd say this is now ready for review. Instructions for testing the website locally are in /web/README.md

I could not get this to fully work locally, so thanks for the screenshots.

My feedback

* "Show all guides" > reword to "Documentation for Foreman _Foreman Version_/Katello _Katello Version_ on _OS_" or similar. Technically, "Release notes" are not a guide. I have no strong opinion.
* "Deployment and installation": > "Deploying Foreman/Katello"
* "Server administration": > "Administrating Foreman/Katello (I think "server" is kinda unfortunate here because readers might confuse this with "I want to manage my RHEL 9 server" which is a _client_ in the context of F/K.)
* "Host administration": > "Administrating hosts"
* I see "managing content" in the section about Hosts, not about F/K maintenance/installation.
* ACD is also part of Hosts
* In `web/releases/nightly.adoc`, we could use attributes for titles _if_ we set the build flavour for every collapsible.

Overall, I like it a lot!

aneta-petrova:
>     * "Show all guides" > reword to "Documentation for Foreman _Foreman Version_/Katello _Katello Version_ on _OS_" or similar. Technically, "Release notes" are not a guide. I have no strong opinion.

I think "Documentation for [...]" would be way too many words for such a small (albeit important) UI element. An alternative would be "Show all documentation". But then again, in-app content like web UI descriptions or Hammer --help pages are also documentation so that would also not be entirely precise :upside_down_face: Additionally, I must say I do consider Release notes a guide :shrug: 

For now no change and I'm sticking to "Show all guides". I'm willing to reconsider if there are further objections.

> 
>     * "Deployment and installation": > "Deploying Foreman/Katello"

Applied.

> 
>     * "Server administration": > "Administrating Foreman/Katello (I think "server" is kinda unfortunate here because readers might confuse this with "I want to manage my RHEL 9 server" which is a _client_ in the context of F/K.)

How about "Administering Foreman server" to make sure users don't confuse this with any other server? This is because I like the distinction between "server" and "hosts". Therefore, I used "Administering F server", let me know if you think it doesn't work.

> 
>     * "Host administration": > "Administrating hosts"

Applied.

> 
>     * I see "managing content" in the section about Hosts, not about F/K maintenance/installation.

It's in "Administering Foreman" now.

> 
>     * ACD is also part of Hosts

This is in "Deploying" now.

> 
>     * In `web/releases/nightly.adoc`, we could use attributes for titles _if_ we set the build flavour for every collapsible.

A good idea (and I do miss the option to use attributes for titles) but then as part of the release process, _someone_ would have to replace those attributes with hard-coded title names, right? I don't think it's a good idea to add that step to the process. So no change.

> 
> 
> Overall, I like it a lot!

Yay!

web/content/js/nav.js
```diff
@@ -23,18 +23,6 @@ function buildNavigation() {
...
-          <a href="#" data-action="dropdown-toggle">${build.title}</a>
-            <div class="dropdown-menu">`
-            + build.guides.map(function(guide){
-              const url = `/${currentVer}/${guide.path}/${build.filename}`;
-              return `<div class="dropdown-div"><a class="dropdown-item" href="${url}">${guide.title}</a></div>`;
-            }).join("")
-            +`</div>
-        </li>`
-      )}).join("")
     + `<li class="nav-item dropdown">
```
ekohl, line 26: You can drop this bit as well:
```
` + `
```

web/content/js/nav.js
```diff
@@ -23,18 +23,6 @@ function buildNavigation() {
 </button>
 <ul class="nav-menu">
   <li class="nav-item"><a href="https://theforeman.org/">About Foreman</a></li>`
-    + navBuilds.map(function(build){
```
ekohl, line 26: Now that you no longer need `navBuilds` you can also remove the `const navBuilds` declaration.

ekohl (commented):
One thing I'm not sure about is the additional clicks. Right now I very often navigate between different guides in the top bar. You go from 2 clicks on the same page on the navigation bar to:

* Click on the version number
* Remember the same version number
* Find the right flavor
* Click show all guides
* Click the guide I want

Perhaps a thought is to at least separate the current version from all versions. That way you end up in the top bar with:

* About Foreman
* Version x.y
* All versions

It would be even better if we also have flavor specific pages. Effectively you could have these URLs:
* `/release/$release` for all 3 flavors
* `/release/$release/foreman-deb` for just the Foreman on Debian guides
* `/release/$release/foreman-deb` for just the Foreman on EL guides
* `/release/$release/katello` for just the Katello guides

Then in the top navigation you would have some Javascript to figure out what best to link to.

This may involve a lot of duplication, which is undesirable. Having reusable bits would be ideal. Not sure how much we can use `include` in adoc to achieve that.

> An unresolved concern that I have is whether I can now delete all *.json files under /web/releases 🤷 Was their sole purpose to build the old navigation menu or are they needed for something else too?

They are used: in https://github.com/theforeman/foreman-documentation/blob/master/web/lib/releases.rb they are read and used. First of all to detect whether the pages themselves should be created but also to provide some context. Most notable, the Foreman and Katello release versions. Some examples where it's still used:

https://github.com/theforeman/foreman-documentation/blob/2dc3c666bf8e5307dc073d5c39113ed986606574/web/content/index.adoc.erb#L34-L36

https://github.com/theforeman/foreman-documentation/blob/2dc3c666bf8e5307dc073d5c39113ed986606574/web/content/index.adoc.erb#L49-L53

https://github.com/theforeman/foreman-documentation/blob/2dc3c666bf8e5307dc073d5c39113ed986606574/web/content/js/versions.js.erb#L1

Having said that, we can probably trim down the `versions.js` file a lot if you don't need the individual menu items there. To do so, remove the `navBuilds` bit from `nav.js`. Also drop `guides_links` from `lib/releases.rb`. Then in `releases/*.json` you can remove `builds`. If you want to be lazy, you can script it using `jq`:
```sh
for file in web/releases/*.json ; do
  jq 'del(.builds)' "$file" > "$file.tmp"
  mv "$file.tmp" "$file"
done
```

Having said all of that, it may not be a bad idea to think about how to restructure it so the JSON files have the sections in them and reflect that in the adoc files. My end goal was to remove all the release specific release adoc files and instead only maintain the JSON files. Then a **further** step would be to store that specific JSON file in the release branch instead of `master`.

I'd be happy to further elaborate my previous thoughts so you can at least weigh them off.

maximiliankolb:
> I think "Documentation for [...]" would be way too many words for such a small (albeit important) UI element. An alternative would be "Show all documentation". But then again, in-app content like web UI descriptions or Hammer --help pages are also documentation so that would also not be entirely precise 🙃 Additionally, I must say I do consider Release notes a guide 🤷

Maybe "Show all documentation" or "Show documentation"? Either way, I am also cool with "guides".

aneta-petrova:
> Maybe "Show all documentation" or "Show documentation"? Either way, I am also cool with "guides".

"Show documentation" is something I can very easily live with :) Thanks, applied.

aneta-petrova:
> One thing I'm not sure about is the additional clicks.

There is more clicking, true. What I was going for was more clicks, but each of them a mindless choice (https://www.goodreads.com/quotes/755485-it-doesn-t-matter-how-many-times-i-have-to-click) because that's my current issue with the usability side of the navigation: you may need very few clicks but as a new user, you have to think hard which click is the one that you need.

I get, however, that for users who already know how to navigate, this is not exactly a change for the better. I would still argue that an interface that tries to pack too many things into it to avoid extra clicks shouldn't be the preferred choice. I don't know the history of the website but I can imagine the current navigation worked quite well for versions 3.7 and earlier (https://docs.theforeman.org/release/3.7/) where there was only one flavor. Version 3.8 adds two additional flavors and that's when the website IMO starts to try to accommodate too many things at once.

>     * Remember the same version number
> [...]
> Perhaps a thought is to at least separate the current version from all versions. That way you end up in the top bar with:
> 
>     * About Foreman
> 
>     * Version x.y
> 
>     * All versions

Current version ("Version x.y") stays selected in the version dropdown at all times so you don't need to remember it. Which gives you easy access to your preferred version and you don't need a separate "Version x.y" item in the top bar. Or am I missing something and you meant something else?

> It would be even better if we also have flavor specific pages. Effectively you could have these URLs:
> 
>     * `/release/$release` for all 3 flavors
> 
>     * `/release/$release/foreman-deb` for just the Foreman on Debian guides
> 
>     * `/release/$release/foreman-deb` for just the Foreman on EL guides
> 
>     * `/release/$release/katello` for just the Katello guides
> 
> 
> Then in the top navigation you would have some Javascript to figure out what best to link to.

I'm afraid that "have some Javascript" is out of my skill zone ;) You have good ideas on how to improve my attempt but I'm afraid I must stay limited to more or less what's already included in this PR.

web/content/js/nav.js
aneta-petrova: Can I really? Because it causes the top menu bar to disappear.

web/content/js/nav.js
aneta-petrova: This also causes the top bar to disappear.

ekohl:
> Current version ("Version x.y") stays selected in the version dropdown at all times so you don't need to remember it. Which gives you easy access to your preferred version and you don't need a separate "Version x.y" item in the top bar. Or am I missing something and you meant something else?

My idea of making it fixed in the same place is that you then can mindlessly click in the same place instead of having to find the right version in a drop down

maximiliankolb:
@aneta-petrova What do you think about demoing this (again)? Either at the Foreman Community Demo or at the monthly docs triage meeting. We could invite others to give feedback. If there is no further feedback, we will probably have to do some kind of vote to reach a conclusion.

I am cool merging this or not, but I am not okay with letting it just sit there.

ekohl:
@aneta-petrova I should have mentioned that we're currently in the middle of branching (https://community.theforeman.org/t/foreman-3-15-branching-process/43138) and I'd like to get that wrapped up first. Especially since you'd touch those files here.

aneta-petrova:
> @aneta-petrova What do you think about demoing this (again)? Either at the Foreman Community Demo or at the monthly docs triage meeting.

I'm not sure about doing another demo. A bolder move would be to merge (if we resolve the conversation on this PR! I wouldn't agree with merging without addressing the conversation here), announce the change, invite others to test it... and then just revert if there are significant objections. The reason is that it would be good to let users try the change for themselves: a demo can't do that and we don't have any kind of test environment, but reverting should always be easy.

> @aneta-petrova I should have mentioned that we're currently in the middle of branching (https://community.theforeman.org/t/foreman-3-15-branching-process/43138) and I'd like to get that wrapped up first. Especially since you'd touch those files here.

Sure! That said, I'd like to reach a resolution before the next branching :)

ekohl:
https://github.com/theforeman/foreman-documentation/pull/3915 implements a compromise.

aneta-petrova:
Closing in favor of https://github.com/theforeman/foreman-documentation/pull/3915. Thanks @ekohl!

PR #3658, opened 2025-02-12
Use foreman-installer for hammer resource quota

bastian-src:

Install the hammer cli package for the `foreman_resource_quota` via `foreman-installer` instead of `dnf`.

`foreman-installer` is the preferred installation method and becomes available with https://github.com/theforeman/foreman-installer/pull/1007

#### Checklists

no cherry-picks: only foreman_resource_quota is part of 3.13; the Hammer CLI plugin is not available in nightly.


bastian-src:
LGTM, thanks! :dolphin:

PR #3615, opened 2025-01-24
Add DR virtualization scenario

aneta-petrova:
#### What changes are you introducing?

Adding steps for disaster recovery by virtualizing Project Server.

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

A follow-up to https://github.com/theforeman/foreman-documentation/pull/3558
Somewhat depends on https://github.com/theforeman/foreman-documentation/pull/3614

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)

~~A lot of blanks but I'm hoping that opening a PR and getting feedback will help fill them in.~~

#### Checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [ ] Foreman 3.13/Katello 4.15 (EL9 only)
* [ ] Foreman 3.12/Katello 4.14 (Satellite 6.16)
* [ ] Foreman 3.11/Katello 4.13 (orcharhino 6.11 on EL8 only; orcharhino 7.0 on EL8+EL9)
* [ ] Foreman 3.10/Katello 4.12
* [ ] Foreman 3.9/Katello 4.11 (Satellite 6.15; orcharhino 6.8/6.9/6.10)
* [ ] Foreman 3.8/Katello 4.10
* [ ] Foreman 3.7/Katello 4.9 (Satellite 6.14)
* We do not accept PRs for Foreman older than 3.7.

PR #3558, opened 2025-01-09
Add overview of preparing for DR

asteflova:
#### What changes are you introducing?

I'm adding a list of recommended disaster recovery plans. Each DR plan description follows the same template to make it easier to compare them and help users decide which one to choose.

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

N/A

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)

As a next step, I'd like to provide more detailed instructions for each of the DR methods. Ideally with three procedures for each: 1. Steps to prepare, 2. Steps to test, 3. Steps to recover.

#### Checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [ ] Foreman 3.13/Katello 4.15 (EL9 only)
* [ ] Foreman 3.12/Katello 4.14 (Satellite 6.16)
* [ ] Foreman 3.11/Katello 4.13 (orcharhino 6.11 on EL8 only; orcharhino 7.0 on EL8+EL9)
* [ ] Foreman 3.10/Katello 4.12
* [ ] Foreman 3.9/Katello 4.11 (Satellite 6.15; orcharhino 6.8/6.9/6.10)
* [ ] Foreman 3.8/Katello 4.10
* [ ] Foreman 3.7/Katello 4.9 (Satellite 6.14)
* We do not accept PRs for Foreman older than 3.7.

PR #3679, opened 2025-02-18
Add DR with active-passive and external storage

aneta-petrova:

Adding a new disaster recovery scenario: Active and passive server with external storage.

A follow-up to https://github.com/theforeman/foreman-documentation/pull/3558
The new scenario follows the structure established in https://github.com/theforeman/foreman-documentation/pull/3615 for consistency

Now that I'm working on the second scenario, I can already see opportunities for reuse or snippets. However, I think it will be better to deal with this after all the planned scenarios are in place.

#### Checklists

### DIFF
--- a/guides/common/modules/con_disaster-recovery-with-active-and-passive-project-server-and-external-storage.adoc
+++ b/guides/common/modules/con_disaster-recovery-with-active-and-passive-project-server-and-external-storage.adoc
@@ -1,0 +1,6 @@
+[id="disaster-recovery-with-active-and-passive-{project-context}-server-and-external-storage"]
+= Disaster recovery with active and passive {ProjectServer} and external storage
+
+In this scenario, you maintain two {ProjectServer}s and you store critical data externally on a network-attached storage (NAS).
+The primary server is active while the secondary server remains passive.
+If the primary server fails, you attach the NAS to your secondary server and turn it into your new primary server.
--- a/guides/common/modules/con_disaster-recovery-with-active-and-passive-project-server-and-external-storage.adoc
+++ b/guides/common/modules/con_disaster-recovery-with-active-and-passive-project-server-and-external-storage.adoc
@@ -1,6 +1,6 @@
 [id="disaster-recovery-with-active-and-passive-{project-context}-server-and-external-storage"]
 = Disaster recovery with active and passive {ProjectServer} and external storage
 
-In this scenario, you maintain two {ProjectServer}s and you store critical data externally on network-attached storage (NAS).
+To prepare for disaster recovery, you can configure two {ProjectServer}s and store critical data externally on shared storage.
 The primary server is active while the secondary server remains passive.
-If the primary server fails, you attach the NAS to your secondary server and turn it into your new primary server.
+If the primary server fails, the shared storage is attached to your secondary server, which turns the secondary server into your new primary server.
--- a/guides/common/modules/con_disaster-recovery-with-active-and-passive-project-server-and-external-storage.adoc
+++ b/guides/common/modules/con_disaster-recovery-with-active-and-passive-project-server-and-external-storage.adoc
@@ -1,6 +1,6 @@
 [id="disaster-recovery-with-active-and-passive-{project-context}-server-and-external-storage"]
 = Disaster recovery with active and passive {ProjectServer} and external storage
 
-In this scenario, you maintain two {ProjectServer}s and you store critical data externally on a network-attached storage (NAS).
+In this scenario, you maintain two {ProjectServer}s and you store critical data externally on network-attached storage (NAS).
 The primary server is active while the secondary server remains passive.
 If the primary server fails, you attach the NAS to your secondary server and turn it into your new primary server.
--- a/guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
+++ b/guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
@@ -21,8 +21,8 @@ The source server remains your primary active server, while the clone server bec
 Perform this test in an isolated staging environment:
 // Does this make sense? We don't want users testing this in production, right?
 
-. Stop the active server to mimic failure.
-// How? foreman-maintain service stop?
+. Mimic a full outage on the active server.
+You can do this by turning the machine off, halting the virtual machine (VM) if your server runs on a VM, or isolating the machine by using a firewall.
 . Start the passive server.
 // How? foreman-maintain service start?
 . Verify that your passive server can access the data stored on your NAS.
--- a/guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
+++ b/guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
@@ -1,0 +1,31 @@
+[id="preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage"]
+= Preparing for disaster recovery with active and passive {ProjectServer} with external storage
+
+Create your passive {ProjectServer} as a clone of your active {ProjectServer}.
+Make sure that both servers connect to the database content on your network-attached storage (NAS).
+
+.Procedure
+. Replicate the `/var/lib/pulp` and `{postgresql-lib-dir}` directories from the active {ProjectServer} to your NAS.
+// Do we want to keep this generic like the virtualization scenario, or is this a good place to provide example rsync commands or something?
+. Clone your active {ProjectServer}.
+See xref:cloning_satellite_server[].
++
+Keep the source server powered on and power off the new cloned server.
+The source server remains your primary active server, while the clone server becomes the secondary passive server.
+. Determine how you want to attach the database content on the NAS to your passive server:
+* If you mount the storage directly by both your active and passive server, the servers will always see the same, up-to-date content.
+* If you mount the storage only to your active server and replicate the storage periodically to another location, the passive server will access the data only if it takes over as the active server.
+// This would be a good place to talk more about the pros and cons of these two alternatives, wouldn't it? What would be those pros and cons?
+
+.Verification
+Perform this test in an isolated staging environment:
+// Does this make sense? We don't want users testing this in production, right?
+
+. Stop the active server to mimic failure.
+// How? foreman-maintain service stop?
+. Start the passive server.
+// How? foreman-maintain service start?
+. Verify that your passive server can access the data stored on your NAS.
+. Assess the functionality of the test {ProjectServer}.
+See xref:Retrieving_the_Status_of_Services_{context}[].
+. Perform these verification checks regularly.
--- a/guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
+++ b/guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
@@ -6,7 +6,6 @@ Make sure that both servers connect to the database content on your shared stora
 
 .Procedure
 . Replicate the `/var/lib/pulp` and `{postgresql-lib-dir}` directories from the active {ProjectServer} to your shared storage.
-// Do we want to keep this generic like the virtualization scenario, or is this a good place to provide example rsync commands or something?
 . Clone your active {ProjectServer}.
 See xref:cloning_satellite_server[].
 . Keep the source server powered on.
@@ -20,7 +19,6 @@ The source server remains your active primary server, while the clone server bec
 
 .Verification
 Perform this test in an isolated staging environment:
-// Does this make sense? We don't want users testing this in production, right?
 
 . Mimic a full outage on the active server.
 You can do this by turning the machine off, halting the virtual machine (VM) if your server runs on a VM, or isolating the machine by using a firewall.
--- a/guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
+++ b/guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
@@ -2,19 +2,20 @@
 = Preparing for disaster recovery with active and passive {ProjectServer} with external storage
 
 Create your passive {ProjectServer} as a clone of your active {ProjectServer}.
-Make sure that both servers connect to the database content on your network-attached storage (NAS).
+Make sure that both servers connect to the database content on your shared storage.
 
 .Procedure
-. Replicate the `/var/lib/pulp` and `{postgresql-lib-dir}` directories from the active {ProjectServer} to your NAS.
+. Replicate the `/var/lib/pulp` and `{postgresql-lib-dir}` directories from the active {ProjectServer} to your shared storage.
 // Do we want to keep this generic like the virtualization scenario, or is this a good place to provide example rsync commands or something?
 . Clone your active {ProjectServer}.
 See xref:cloning_satellite_server[].
+. Keep the source server powered on.
+Power off the new cloned server.
 +
-Keep the source server powered on and power off the new cloned server.
-The source server remains your primary active server, while the clone server becomes the secondary passive server.
-. Determine how you want to attach the database content on the NAS to your passive server:
-* If you mount the storage directly by both your active and passive server, the servers will always see the same, up-to-date content.
-* If you mount the storage only to your active server and replicate the storage periodically to another location, the passive server will access the data only if it takes over as the active server.
+The source server remains your active primary server, while the clone server becomes the passive secondary server.
+. Determine how you want to attach the database content on the shared storage to your passive server:
+* If you mount the storage directly on both your active and passive server, the servers will always see the same, up-to-date content.
+* If you mount the storage only on your active server, the passive server will access the data only if it takes over as the active server.
 // This would be a good place to talk more about the pros and cons of these two alternatives, wouldn't it? What would be those pros and cons?
 
 .Verification
@@ -25,7 +26,7 @@ Perform this test in an isolated staging environment:
 You can do this by turning the machine off, halting the virtual machine (VM) if your server runs on a VM, or isolating the machine by using a firewall.
 . Start the passive server.
 // How? foreman-maintain service start?
-. Verify that your passive server can access the data stored on your NAS.
+. Verify that your passive server can access the data stored on your shared storage.
 . Assess the functionality of the test {ProjectServer}.
 See xref:Retrieving_the_Status_of_Services_{context}[].
 . Perform these verification checks regularly.
--- a/guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
+++ b/guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
@@ -2,7 +2,7 @@
 = Preparing for disaster recovery with active and passive {ProjectServer} with external storage
 
 Create your passive {ProjectServer} as a clone of your active {ProjectServer}.
-Make sure that both servers connect to the database content on your shared storage.
+Ensure that the `/var/lib/pulp` and `{postgresql-lib-dir}` directories on your shared storage are available to both servers.
 
 .Procedure
 . Replicate the `/var/lib/pulp` and `{postgresql-lib-dir}` directories from the active {ProjectServer} to your shared storage.
--- a/guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
+++ b/guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
@@ -29,3 +29,8 @@ Perform this test in an isolated staging environment:
 . Assess the functionality of the test {ProjectServer}.
 See xref:Retrieving_the_Status_of_Services_{context}[].
 . Perform these verification checks regularly.
+
+ifdef::satellite[]
+.Additional resources
+* For more information on mounting directories, see link:{RHELDocsBaseURL}9/html-single/managing_file_systems/index#mounting-file-systems-on-demand_managing-file-systems[Mounting file systems on demand] in _{RHEL}{nbsp}9 Managing file systems_.
+endif::[]
--- a/guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
+++ b/guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
@@ -21,9 +21,8 @@ The source server remains your active primary server, while the clone server bec
 Perform this test in an isolated staging environment:
 
 . Mimic a full outage on the active server.
-You can do this by turning the machine off, halting the virtual machine (VM) if your server runs on a VM, or isolating the machine by using a firewall.
-. Start the passive server.
-// How? foreman-maintain service start?
+To make sure the active server is inaccessible, you can turn the machine off, halt the virtual machine (VM) if your server runs on a VM, or isolate the machine by using a firewall.
+. Switch DNS records of the active server with the DNS records of the passive server.
 . Verify that your passive server can access the data stored on your shared storage.
 . Assess the functionality of the test {ProjectServer}.
 See xref:Retrieving_the_Status_of_Services_{context}[].
--- a/guides/common/modules/proc_recovering-from-disaster-with-active-and-passive-project-server-with-external-storage.adoc
+++ b/guides/common/modules/proc_recovering-from-disaster-with-active-and-passive-project-server-with-external-storage.adoc
@@ -1,0 +1,17 @@
+[id="recovering-from-disaster-with-active-and-passive-server-with-external-storage"]
+= Recovering from disaster with active and passive server with external storage
+
+If your active {ProjectServer} fails, detach it from the network-attached storage (NAS) and make sure your passive server can access the data stored on the NAS.
+This turns the passive server into your new active server.
+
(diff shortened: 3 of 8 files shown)


aneta-petrova:
@ehelms Can you please review another DR scenario? I am once again making a lot of barely-educated guesses and I added several questions as comments so please keep that in mind and read carefully :) Also, I'm not sure what level of detail we are aiming for with this scenario: Do we want to keep things high-level like the virtualization scenario or can we/should we provide more detailed instructions?

guides/common/modules/con_disaster-recovery-with-active-and-passive-project-server-and-external-storage.adoc
```diff
@@ -0,0 +1,6 @@
+[id="disaster-recovery-with-active-and-passive-{project-context}-server-and-external-storage"]
+= Disaster recovery with active and passive {ProjectServer} and external storage
+
+In this scenario, you maintain two {ProjectServer}s and you store critical data externally on a network-attached storage (NAS).
```
ehelms, line 4:
```suggestion
In this scenario, you maintain two {ProjectServer}s and you store critical data externally on network-attached storage (NAS).
```

I think if we were referencing a device we'd use `a network-attached storage device`.

guides/common/modules/con_disaster-recovery-with-active-and-passive-project-server-and-external-storage.adoc
```diff
@@ -0,0 +1,6 @@
...
+The primary server is active while the secondary server remains passive.
+If the primary server fails, you attach the NAS to your secondary server and turn it into your new primary server.
```
ehelms, line 6: @evgeni They could attach it to both, and then flip it to write for the passive server when it needs to become the primary? Or is this a better /safer method?

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
```diff
@@ -0,0 +1,31 @@
+[id="preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage"]
+= Preparing for disaster recovery with active and passive {ProjectServer} with external storage
+
+Create your passive {ProjectServer} as a clone of your active {ProjectServer}.
+Make sure that both servers connect to the database content on your network-attached storage (NAS).
+
+.Procedure
+. Replicate the `/var/lib/pulp` and `{postgresql-lib-dir}` directories from the active {ProjectServer} to your NAS.
```
ehelms, line 8: We can reference the documentation that discusses using NAS here as that should be a prerequisite for this DR scenario.

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
```diff
@@ -0,0 +1,31 @@
...
+* If you mount the storage directly by both your active and passive server, the servers will always see the same, up-to-date content.
+* If you mount the storage only to your active server and replicate the storage periodically to another location, the passive server will access the data only if it takes over as the active server.
+// This would be a good place to talk more about the pros and cons of these two alternatives, wouldn't it? What would be those pros and cons?
+
+.Verification
+Perform this test in an isolated staging environment:
+// Does this make sense? We don't want users testing this in production, right?
+
+. Stop the active server to mimic failure.
+// How? foreman-maintain service stop?
```
ehelms, line 25: Better to turn the machine off (or halt the VM) to mimic a full outage. Alternatively, they could have a DR firewall test scenario that walls off the machine.

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
```diff
@@ -0,0 +1,31 @@
...
+. Start the passive server.
+// How? foreman-maintain service start?
```
ehelms, line 27: I can see two scenarios:

 1. A passive server that is running all the time but inaccessible. In this case the user would simply ensure storage is attached properly and flip DNS.
 2. A passive server that is something like an image, that gets turned up, storage attached and then update is run.
 3. A machine that is spun up, restore is run with storage attached.

@evgeni Vet me here

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
aneta-petrova: I seem to be having trouble finding something in RH docs that applies but I'll keep looking.

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
aneta-petrova: Added.

guides/common/modules/con_disaster-recovery-with-active-and-passive-project-server-and-external-storage.adoc
```diff
@@ -0,0 +1,6 @@
+[id="disaster-recovery-with-active-and-passive-{project-context}-server-and-external-storage"]
+= Disaster recovery with active and passive {ProjectServer} and external storage
+
+In this scenario, you maintain two {ProjectServer}s and you store critical data externally on network-attached storage (NAS).
```
rh-max, line 4: `In this scenario` refers to what's described in the section heading, right? Section body is then dependent on section heading, which I see seldom. If there are no guidelines against this, looks good.

guides/common/modules/con_disaster-recovery-with-active-and-passive-project-server-and-external-storage.adoc
```diff
@@ -0,0 +1,6 @@
...
+The primary server is active while the secondary server remains passive.
+If the primary server fails, you attach the NAS to your secondary server and turn it into your new primary server.
```
rh-max, line 6: I don't know the exact mechanism, but it might be that between this:

  `attach the NAS to your secondary server`

and this:

  `turn it into your new primary server`

some intermediate event(s) is/are happening, even if automatic. In that case it's worth mentioning, because otherwise this seems to me to have a gap in the sequence of events.

In addition to the potentially-existing event I mentioned, it might also make sense to explicitly say `make it active` right before the `and turn it into your new primary server` part.

Disregard these suggestions if you intended to only mention user actions here, though.

guides/common/modules/ref_prerequisites-disaster-recovery-with-active-and-passive-project-server-and-external-storage.adoc
```diff
@@ -0,0 +1,9 @@
+[id="prerequisites-disaster-recovery-with-active-and-passive-project-server-and-external-storage"]
+= Prerequisites
+
+* Review xref:overview-of-recommended-disaster-recovery-plans[] to make sure that this disaster recovery plan works for you.
+* Review {InstallingServerDocURL}storage-requirements_{project-context}[Storage requirements] and {InstallingServerDocURL}storage-guidelines_{project-context}[Storage guidelines] in _{InstallingServerDocTitle}_.
+Ensure that your network-attached storage meets the requirements to hold the contents of `/var/lib/pulp` and `{postgresql-lib-dir}`.
```
rh-max, line 6:
```suggestion
Ensure that your network-attached storage meets the requirements of holding the contents of `/var/lib/pulp` and `{postgresql-lib-dir}`.
```

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
```diff
@@ -0,0 +1,36 @@
...
+Create your passive {ProjectServer} as a clone of your active {ProjectServer}.
+Make sure that both servers connect to the database content on your network-attached storage (NAS).
+
+.Procedure
+. Replicate the `/var/lib/pulp` and `{postgresql-lib-dir}` directories from the active {ProjectServer} to your NAS.
+// Do we want to keep this generic like the virtualization scenario, or is this a good place to provide example rsync commands or something?
+. Clone your active {ProjectServer}.
+See xref:cloning_satellite_server[].
++
+Keep the source server powered on and power off the new cloned server.
```
rh-max, line 13: `and` might imply that instruct the reader to do these in a particular sequence. I don't think you do, so perhaps this would be better:

`Keep the source server powered on. Keep the new cloned server powered off.`

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
```diff
@@ -0,0 +1,36 @@
...
+The source server remains your primary active server, while the clone server becomes the secondary passive server.
```
rh-max, line 14: The `primary active server` phrasing implies that there is another active server that is not primary. Better go with `active primary server`.

rh-max, line 14: And same with `secondary passive server` -- better use `passive secondary server`.

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
```diff
@@ -0,0 +1,36 @@
...
+. Determine how you want to attach the database content on the NAS to your passive server:
+* If you mount the storage directly by both your active and passive server, the servers will always see the same, up-to-date content.
```
rh-max, line 16: `mount ... by` -> `mount ... on`, otherwise it seems like both of these mountings serve some greater purpose, and not just the purpose of these two servers interacting with the data.

Additionally, `mount by` is almost never used, it's usually `mount on` and less frequently `mount at`.

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
```diff
@@ -0,0 +1,36 @@
...
+* If you mount the storage only to your active server and replicate the storage periodically to another location, the passive server will access the data only if it takes over as the active server.
```
rh-max, line 17:
```suggestion
* If you mount the storage only on your active server and replicate the storage periodically to another location, the passive server will access the data only if it takes over as the active server.
```

rh-max, line 17: The `replicate the storage periodically to another location` part doesn't really make a difference AFAIU, it's just to ensure redundancy of data on the NAS. Hence, it has no bearing on when the passive server accesses the data.

However, I suppose the reason you've put it here is to help readers identify the specific plan out of the selection of plans in `21.1. Overview of recommended disaster recovery plans`, so keeping it is fine.

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
```diff
@@ -0,0 +1,36 @@
...
+// Does this make sense? We don't want users testing this in production, right?
+
+. Mimic a full outage on the active server.
+You can do this by turning the machine off, halting the virtual machine (VM) if your server runs on a VM, or isolating the machine by using a firewall.
+. Start the passive server.
+// How? foreman-maintain service start?
+. Verify that your passive server can access the data stored on your NAS.
+. Assess the functionality of the test {ProjectServer}.
+See xref:Retrieving_the_Status_of_Services_{context}[].
+. Perform these verification checks regularly.
```
rh-max, line 31: Maybe remove this step, and just add the word `regularly` to the intro of the verification procedure?

rh-max: So the intro would read:
`Regularly perform this test in an isolated staging environment:`

guides/common/modules/proc_recovering-from-disaster-with-active-and-passive-project-server-with-external-storage.adoc
```diff
@@ -0,0 +1,17 @@
...
+If your active {ProjectServer} fails, detach it from the network-attached storage (NAS) and make sure your passive server can access the data stored on the NAS.
+This turns the passive server into your new active server.
+
+.Procedure
+. Verify that the failed active server is powered off or fully detached from the NAS.
+This ensures that the failed server cannot keep writing to the NAS, which helps prevent further data corruption on the connected databases.
+// Is foreman-maintain service stop enough?
+. Switch DNS records of the active server with the DNS records of the passive server.
+This ensures that hosts remain connected and you do not need to re-register them.
+. If your NAS was mounted by both your active and passive servers, your passive server can already access the data.
```
rh-max, line 13:
```suggestion
. If your NAS was mounted on both your active and passive servers, your passive server can already access the data.
```

guides/common/modules/proc_recovering-from-disaster-with-active-and-passive-project-server-with-external-storage.adoc
```diff
@@ -0,0 +1,17 @@
+[id="recovering-from-disaster-with-active-and-passive-server-with-external-storage"]
+= Recovering from disaster with active and passive server with external storage
+
+If your active {ProjectServer} fails, detach it from the network-attached storage (NAS) and make sure your passive server can access the data stored on the NAS.
+This turns the passive server into your new active server.
+
+.Procedure
+. Verify that the failed active server is powered off or fully detached from the NAS.
+This ensures that the failed server cannot keep writing to the NAS, which helps prevent further data corruption on the connected databases.
```
rh-max, line 9: `further data corruption` might not be entirely precise -- if it wasn't writing anything before failing, then there was no data corruption, in which case `further` would be slightly misleading (worst case scenario, a very literally-minded reader might take some unnecessary actions on their supposedly corrupted database). I've no good ideas how to phrase it though, perhaps something like this?

  `which helps prevent data corruption on connected databases from starting or worsening`
  or
  `which helps prevent connected databases from experiencing the start or worsening of data corruption`

rh-max (changes requested):
Very clear and well-structured overall, I have only a few suggestions and considerations. :+1: 

Also, I'll just paste here what I found in section `21.1. Overview of recommended disaster recovery plans`:


1)
```
Choose a disaster recovery plan that best helps ensure the continuity of Satellite services in your deployment.

Virtualizing your Satellite Server ...
```

This communicates the idea _"my disaster recovery plan is virtualizing the Satellite Server"_, but virtualizing by itself is not the solution. I'd make it more precise:

`Snapshots of virtualized Satellite Server`

This also aligns better with the non-action style of the subsequent plan suggestions.


2)
`This method is suitable if you can run Satellite on top of virtualization.`

This is partly subjective, but to me choosing this phrasing:

`run Satellite on top of virtualization` (phrasing 1)

instead of something like this:

`run Satellite virtualized` (phrasing 2a)

or

`run Satellite in a virtual machine` (phrasing 2b)

implies that scenario described in phrasing 1 has some additional layer between Satellite and virtualization, and scenario described in phrasings 2a and 2b does not. But maybe most readers have more context, which rules out that extra non-existent layer.

guides/common/modules/con_disaster-recovery-with-active-and-passive-project-server-and-external-storage.adoc
evgeni: Depends on how adventurous the user is.
Both attaching and flipping ro/rw is an action you need to perform and as such doesn't give you any real benefit one way or another.
If on the other hand you can keep the storage attached in rw mode all the time, but ensure that absolutely no way the app starts, you win a bit of time/process steps.

Is it worth it? idk.

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
evgeni: 1. you can't count ;-)
2. when you run active/passive your nodes need to be identical and the only change you perform is "start the service" (however that is exactly defined)
3. I guess it should be something like your 1, but "inaccessible" means "services are down" (even if the machine is not accessed, running services would mean workers pick up work from the pool etc, not what you want!)

guides/common/modules/con_disaster-recovery-with-active-and-passive-project-server-and-external-storage.adoc
evgeni, line 4: I am not sure why, but my brain insists that "NAS" is the wrong name here as it associates it with those tiny, "couple of hard-drives" boxes from QNAP, Synology etc, which is exactly not what you'd use in a datacenter. And it then goes on saying it should be "SAN" (Storage Area Network) instead.

The reality is muddier. A SAN is effectively "here is a disk, exposed over the network, the client is handling the filesystem" while a NAS is "here is a filesystem, you can store files here" [1]. So of you export your super expensive NetApp, Hitachi, whatever via NFS, you get a NAS. And if you export your cheap QNAP, Synology, whatever via iSCSI you get a SAN.

Should we say "store data on shared storage" and let the interpretation to the reader?

[1] https://en.wikipedia.org/wiki/Network-attached_storage#Comparing_with_SAN

aneta-petrova:
> Also, I'll just paste here what I found in section `21.1. Overview of recommended disaster recovery plans`:

Let's track these in https://github.com/theforeman/foreman-documentation/issues/3704.

guides/common/modules/proc_recovering-from-disaster-with-active-and-passive-project-server-with-external-storage.adoc
aneta-petrova: How about just dropping the `which [...]` part? Simply saying that we don't want the failed server writing to the storage should be enough.

guides/common/modules/proc_recovering-from-disaster-with-active-and-passive-project-server-with-external-storage.adoc
aneta-petrova: Addressed here and on line 14.

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
aneta-petrova: I'm wondering whether all of this goes into too much detail for a testing/verification procedure. As an alternative, I could also drop line 27, which would take users directly to line 29 to tell them 'verify your passive server', however they would decide to do it. Is that just me being lazy, or would it actually make sense like this?

aneta-petrova:
@ehelms @evgeni Thanks for all your feedback so far! I implemented most of it, except for one comment thread that remains open with a follow-up question. Can you please re-review? This scenario feels more complex to me than the virtualization one, so I can't really tell whether we are getting close enough to the desired outcome.

aneta-petrova:
It looks like the preview in https://github.com/theforeman/foreman-documentation/pull/3679#issuecomment-2664807362 didn't get updated after the latest set of commits. Just FYI for potential reviewers.

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
ehelms: Experience tells me that these details are questions users will ask, and this guide is intended to help reduce the questions being asked by codifying our knowledge. Perhaps we say as @evgeni did in #3 - by stating inaccessible and then giving some examples.

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
```diff
@@ -0,0 +1,35 @@
+[id="preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage"]
+= Preparing for disaster recovery with active and passive {ProjectServer} with external storage
+
+Create your passive {ProjectServer} as a clone of your active {ProjectServer}.
```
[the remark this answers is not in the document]
evgeni: You mean packages? Configurations? Data?
The latter is handled by the shared storage, but I agree we don't have a good solution for configurations and packages.

guides/common/modules/con_disaster-recovery-with-active-and-passive-project-server-and-external-storage.adoc
```diff
@@ -0,0 +1,6 @@
+[id="disaster-recovery-with-active-and-passive-{project-context}-server-and-external-storage"]
+= Disaster recovery with active and passive {ProjectServer} and external storage
```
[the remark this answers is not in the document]
rh-max: @aneta-petrova It's good. Another option is something like `Disaster recovery for active and passive {ProjectServer} with external storage`. This way, there are no chained `and`s.

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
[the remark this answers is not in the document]
aneta-petrova: The way I'm interpreting this is:

* Data: "handled by the shared storage", which means no doc action is required (the shared storage solution is what this PR adds)
* Configurations + packages: "no good solution", which means no action is currently possible

Please let me know if I'm missing something.

aneta-petrova:
Feedback from style review implemented, which means `style review done`.

Rebased, which pulled in the change in https://github.com/theforeman/foreman-documentation/pull/3705 and allowed me to include the procedure on retrieving status of services cleanly.

aneta-petrova:
@ehelms and/or @evgeni, do you have any further comments?

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
```diff
@@ -0,0 +1,34 @@
...
+Make sure that both servers connect to the database content on your shared storage.
```
evgeni, line 5: What does "database content" mean here?

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+.Procedure
+. Replicate the `/var/lib/pulp` and `{postgresql-lib-dir}` directories from the active {ProjectServer} to your shared storage.
+. Clone your active {ProjectServer}.
+See xref:cloning_satellite_server[].
```
evgeni, line 10: I just realized: the cloning process involves a DB-restore, which is probably not what people want.
What would probably work is: clone with `/var/lib/pgsql` being local, then turn things off and replace it with shared storage.
But that's fragile :/

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
aneta-petrova: I was trying to come up with a short way to describe `/var/lib/pulp` and `/var/lib/pgsql`. What's a good alternative here? Just "content"? Or just "data"? Or perhaps be specific like this:

```suggestion
Ensure that both servers connect to the contents of `/var/lib/pulp` and `/var/lib/pgsql` on your shared storage.
```

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
aneta-petrova: When you say that it's "fragile", you're actually kind of saying that that's probably not what we want either :) So what would be a good solution?

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
evgeni: How about
```suggestion
Ensure that `/var/lib/pulp` and `/var/lib/pgsql` on your shared storage is available to both servers.
```

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
evgeni: Yeah… The deployment consists of three things: packages (rpms), config (/etc) and data (/var/lib/). To get a working deployment you need all of them. As the data is quickly moving, we place it on shared storage so it can be picked up from various places. The packages and config are not, so we can keep them on "traditional" storage, but they still need to match.

The easiest way to create a matching set is a "clone" (in the general meaning of the term), but satellite-clone does more than cloning rpms/configs, it also does data.

How we can get to a good clone of the slowly moving parts only: I need to think more about it :D

ehelms (approved):
I am good to merge the current state.

guides/common/modules/proc_preparing-for-disaster-recovery-with-active-and-passive-project-server-with-external-storage.adoc
aneta-petrova: I created issue https://github.com/theforeman/foreman-documentation/issues/3714 where we can track this. According to https://github.com/theforeman/foreman-documentation/pull/3679#pullrequestreview-2662882118, this open item doesn't block merging.

aneta-petrova:

a8cd1ffbdf..f564f69511  3.14 -> 3.14

PR #3692, opened 2025-02-25
Update xrefs that are broken in satellite d/s builds

aneta-petrova:

Adding labels to a few xref cross-links.

https://issues.redhat.com/browse/SAT-31298 reports a few xrefs that appear broken in Satellite builds. Satellite d/s docs tooling doesn't support multiple xrefs included on a single line, unless their labels are explicitly defined.

Another solution would be to include each xref on a separate line, for example as bullet point lists. But adding the label should work too (example: [9.4 Setting the global out-of-sync interval](https://docs.redhat.com/en/documentation/red_hat_satellite/6.16/html-single/managing_configurations_by_using_puppet_integration/index#setting-the-global-out-of-sync-interval_managing-configurations-puppet)).

#### Checklists

### DIFF
--- a/guides/common/modules/proc_configuring-project-to-use-ldap.adoc
+++ b/guides/common/modules/proc_configuring-project-to-use-ldap.adoc
@@ -23,8 +23,7 @@ Configure an LDAP authentication source to enable users to log in to {Project} w
 +
 For TLS encrypted connections, select *LDAPS* to enable encryption.
 . On the *Account* tab, enter the account information and domain name details.
-+
-See xref:Example_Settings_for_LDAP_Connections_{context}[] and xref:Example_LDAP_Filters_{context}[].
+For more information, see xref:Example_Settings_for_LDAP_Connections_{context}[Example settings for LDAP connections] and xref:Example_LDAP_Filters_{context}[Example LDAP filters].
 . On the *Attribute mappings* tab, map LDAP attributes to {Project} attributes.
 . On the *Locations* tab, select the locations you want {Project} to assign to users created from the LDAP authentication source.
 These locations are available to users after they log in for the first time.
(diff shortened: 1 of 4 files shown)


aneta-petrova:
> This looks like an issue with downstream docs frameworks. As far as I can tell, it previously affected two xrefs in a single line. So far, the solution was always to use bullet points (one per link).
> 
> I like this much better than adding titles to xrefs. Would you mind investigating this?

Sure, I can change all of these to bullet points. But first I'd like to know why you prefer bulleted lists :) I'm not saying no, it's just that when I compared the two options, bullet points seemed to make the text longer and used up a lot more words, which is why I decided to go with the labels. I don't have a strong preference myself but would like to understand your point of view better. @maximiliankolb

aneta-petrova:
How about this @maximiliankolb? With the latest commit, I'm using the list workaround for xrefs to other sections. I'm only keeping the explicit label workaround for references to CLI/API/Ansible procedures. The reason is that I don't expect the CLI/API/Ansible headings to change (so I think defining their labels explicitly is safe enough) + listing these as individual bullets would take up way too much space (considering that they only point to places within that same module, that's IMO not worth making the procedure that much longer).

maximiliankolb (commented):
I like the lists a lot, because they indicate that there are multiple links users can choose from depending on the method on how they interact with Foreman: WebUI, API, CLI.

not blocking your PR: Would like to see this for the other two files too.

aneta-petrova:
I checked a few other places where we reference multiple procedures and found this: https://github.com/theforeman/foreman-documentation/blob/e9dd66a8d3b0d82ace6a0c756c0d05a0c5de42e0/guides/common/modules/proc_invalidating-your-own-jwts.adoc?plain=1#L6  In order to be consistent, I added blank lines in between the xrefs.

aneta-petrova:
@maximiliankolb reached out to me privately to let me know that I should consider his latest comment an ack. That covers `style review`. Let's merge.

aneta-petrova:

* d844b1b15e..8e5f92371a  3.14 -> 3.14
* 450e869788..8faf9f9ea7  3.13 -> 3.13
* ac68a5a1d0..54f202c833  3.12 -> 3.12

PR #3699, opened 2025-02-26
Restructure required RHEL repos + 9, 10 to Managing content

Lennonka:

Splitting off required repos from _Enabling RH repos_ into a reference module
And adding RHEL 9 and 10 repos
Adding Kickstart repos needed for provisioning

- Splitting off because the list would become too long
- Adding RHEL 9 because it was missing, RHEL 10 will soon be available

Bug [SAT-31200](https://issues.redhat.com/browse/SAT-31200) (public)

#### Checklists

### DIFF
--- a/guides/common/assembly_importing-content.adoc
+++ b/guides/common/assembly_importing-content.adoc
@@ -170,6 +170,8 @@ endif::[]
 
 include::modules/proc_enabling-red-hat-repositories.adoc[leveloffset=+1]
 
+include::modules/ref_required-red-hat-repositories.adoc[leveloffset=+2]
+
 include::modules/proc_synchronizing-repositories.adoc[leveloffset=+1]
 
 include::modules/proc_synchronizing-all-repositories-in-an-organization.adoc[leveloffset=+1]
--- a/guides/common/assembly_importing-content.adoc
+++ b/guides/common/assembly_importing-content.adoc
@@ -170,8 +170,6 @@ endif::[]
 
 include::modules/proc_enabling-red-hat-repositories.adoc[leveloffset=+1]
 
-include::modules/ref_required-red-hat-repositories.adoc[leveloffset=+2]
-
 include::modules/proc_synchronizing-repositories.adoc[leveloffset=+1]
 
 include::modules/proc_synchronizing-all-repositories-in-an-organization.adoc[leveloffset=+1]
--- a/guides/common/modules/ref_required-red-hat-repositories.adoc
+++ b/guides/common/modules/ref_required-red-hat-repositories.adoc
@@ -23,8 +23,6 @@ You require both:
 
 For {RHEL} 7 hosts::
 You require the *{RHEL} 7 Server (RPMs)* repository.
-+
-The difference between associating {RHEL} operating system release version with either *7Server* repositories or *7._y_* repositories is that *7Server* repositories contain all the latest updates while *{RHEL} 7._y_* repositories stop getting updates after the next minor version release.
 
 .Additional repositories for provisioning
 To provision hosts by using the Anaconda installer with Kickstart, you additionally need the following repositories:
@@ -49,4 +47,3 @@ You require both:
 
 For {RHEL} 7 hosts::
 You require the *{RHEL} 7 Server (Kickstart)* repository.
-Note that Kickstart repositories only have minor versions.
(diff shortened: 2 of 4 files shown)


guides/common/assembly_importing-content.adoc
```diff
@@ -170,6 +170,8 @@ endif::[]
 
 include::modules/proc_enabling-red-hat-repositories.adoc[leveloffset=+1]
 
+include::modules/ref_required-red-hat-repositories.adoc[leveloffset=+2]
```
Lennonka, line 173: Not sure if this should go to an appendix. WDYT?

aneta-petrova: That might help, yes. What would you think about converting the contents into a table too?

Or... you could keep it in the procedure module because these are prerequisites, right? Things that users _always_ need when following the procedure? So they might not appreciate being redirected to a different section, only to have to return immediately after. But I understand your concern about the list being too long now.

It might be a good use case for tabs, too: https://docs.theforeman.org/nightly/Installing_Server/index-foreman-deb.html#configuring-repositories_foreman Tabs would help keep the list more compact and then it would work better in its original place.

Lennonka: I don't like the idea of a table for this case.
No, they aren't formal prerequisites. The procedure is generic for enabling repos. The reference tells the user which repos they need for what purposes so that they know what repos they should enable.

How are tabs handled downstream?

Lennonka:
@stejskalleos Please, tech review.

guides/common/modules/ref_required-red-hat-repositories.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+For {RHEL} 8 hosts::
+You require both:
+
+* *{RHEL} 8 for x86_64 - BaseOS (Kickstart)*
+* *{RHEL} 8 for x86_64 - AppStream (Kickstart)*
+
+For {RHEL} 7 hosts::
+You require the *{RHEL} 7 Server (Kickstart)* repository.
+Note that Kickstart repositories only have minor versions.
```
stejskalleos, line 52: Nit: This sounds weird to me, it took me a while to understand that it means you cannot use the `latest` version.

guides/common/modules/ref_required-red-hat-repositories.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+* *{RHEL} 8 for x86_64 - BaseOS (RPMs)*
+* *{RHEL} 8 for x86_64 - AppStream (RPMs)*
+
+For {RHEL} 7 hosts::
+You require the *{RHEL} 7 Server (RPMs)* repository.
++
+The difference between associating {RHEL} operating system release version with either *7Server* repositories or *7._y_* repositories is that *7Server* repositories contain all the latest updates while *{RHEL} 7._y_* repositories stop getting updates after the next minor version release.
+
+.Additional repositories for provisioning
```
stejskalleos, line 29: Since the repos are the same for RHEL-{8,9,10}, do we have to mention them specifically?

stejskalleos (commented):
Technical side ok, just two comments

guides/common/modules/ref_required-red-hat-repositories.adoc
Lennonka: I'll probably drop this. RH is not releasing new RHEL 7 versions anymore.

maximiliankolb (approved):
style-wise LGTM

aneta-petrova (approved):
One comment with a few suggestions, but all non-blocking.

maximiliankolb (approved):

aneta-petrova:

14bc207b3e..aadebce55f  3.14 -> 3.14

guides/common/modules/ref_required-red-hat-repositories.adoc
```diff
@@ -0,0 +1,37 @@
+[id="required-red-hat-repositories"]
+= Required Red Hat repositories
+
+You need the following repositories to manage hosts with {RHEL}.
+
+For {RHEL} 10 hosts::
+* *{RHEL} 10 for x86_64 - BaseOS (RPMs)*
```
apinnick, line 7: I realize this is a little late because your PR has been merged. I was looking through this section in the Pantheon preview build and found the bold font hard to read. Is there any reason not to use regular font for the repos?

PR #3709, opened 2025-03-05
Consolidate 'keycloak realm' replaceables

aneta-petrova:

Replacing two (now obsolete) replaceables for Keycloak realm with `{Project}_Realm`.

To be consistent with the other references to a Keycloak realm for Foreman, which already use `{Project}_Realm`: https://github.com/theforeman/foreman-documentation/blob/6239e9211bb1a59750d195247290dad8a0969e4e/guides/common/modules/proc_configuring-a-project-client-to-provide-hammer-cli-authentication-with-keycloak.adoc?plain=1#L17

I seem to have missed these in https://github.com/theforeman/foreman-documentation/commit/50b9023ea0779c82d9b399e428cc1fc1eb159559 where I originally updated the realm replaceable.

#### Checklists

### DIFF
--- a/guides/common/attributes-base.adoc
+++ b/guides/common/attributes-base.adoc
@@ -90,7 +90,6 @@
 :keycloak-quarkus: Quarkus-based Keycloak
 :keycloak: Keycloak
 :keycloak-example-com: keycloak.example.com
-:keycloak-realm: KEYCLOAK_REALM
 // We need Satellite attributes here because some upstream builds refer to RH SSO and RH BK documentation.
 :RHSSO: Red{nbsp}Hat Single Sign-On
 :RHBK: Red{nbsp}Hat build of Keycloak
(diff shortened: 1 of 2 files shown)


ekohl:
For me (and perhaps this is my programmer background) it's easy to see that it's exactly the same value. Perhaps I'm just not used to reading `_value_` as `$value`

I'll also say that I still regret that users have to do this manually rather than providing them as installer arguments. Then the whole problem would be reduced to passing the value once in an installer argument and it'd be obvious that an attribute is not needed.

maximiliankolb (approved):
Makes sense to me.

ekohl:
If you both feel this is the better path then don't let me hold it back.

maximiliankolb:
If it was part of the installer and needed no/less docs, then I think we'd make this change in foreman-documentation immediately. For now, I consider this PR an improvement to docs maintenance.

aneta-petrova:

a2cf11c53b..9b05712a5e  3.14 -> 3.14
5f1990edf8..92437788ef  3.13 -> 3.13
90d3b30e8a..a5c26f5409  3.12 -> 3.12

PR #3720, opened 2025-03-10
Add upgrade warning for `ansible_tower_fqdn` removal

nofaralfasi:
Add a release note about the parameter change, explaining that
`ansible_tower_fqdn` has been replaced and highlighting the default API path behavior.

See `Update Ansible Tower API URL for AAP 2.5+ compatibility` PR: https://github.com/theforeman/foreman/pull/10445

#### Checklists

### DIFF
--- a/guides/doc-Release_Notes/topics/foreman.adoc
+++ b/guides/doc-Release_Notes/topics/foreman.adoc
@@ -34,6 +34,14 @@ Clients will remain supported.
 For more details, see the https://community.theforeman.org/t/drop-debian-11-ruby-2-7-and-nodejs-14-support-in-foreman-3-14/40503[removal RFC].
 endif::[]
 
+=== {awx} Parameter Change
+The `ansible_tower_fqdn` parameter has been removed and replaced with `ansible_tower_api_url`.
+This new parameter includes both the previous `ansible_tower_fqdn` value and the required API path for {awx}.
+
+By default, the API path is set to `/api/controller/v2`.
+If you are using an older {awx} version, update the API path manually to `/api/v2`.
+If you're unsure which version you have, check your instance's API endpoints to confirm the correct path.
+
 [id="foreman-deprecations"]
 == Deprecations
 


guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -34,6 +34,11 @@ Clients will remain supported.
 For more details, see the https://community.theforeman.org/t/drop-debian-11-ruby-2-7-and-nodejs-14-support-in-foreman-3-14/40503[removal RFC].
 endif::[]
 
+=== {awx} Parameter Change
```
Lennonka, line 37: Is it a host parameter or a global setting?

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -34,6 +34,11 @@ Clients will remain supported.
...
+The `ansible_tower_fqdn` parameter has been removed and replaced with `ansible_tower_api_url`. This new parameter includes both the previous `ansible_tower_fqdn` value and the required API path for {awx}.
+
+By default, the API path is set to `/api/controller/v2`. If you are using an older version of {awx} that requires a different API path, you will need to update it manually to `/api/v2`.
```
ekohl, line 40: This is something I missed in the code (which is why I love user focused release notes as a secondary check): today we use `/api/v2`. This also implies that all users upgrading to Foreman 3.14 need to have AAP 2.5+ or perform a manual update.

I'm unsure how to best phrase that, but it's good to call out.

guides/doc-Release_Notes/topics/foreman.adoc
ekohl: Perhaps worth to call out AAP 2.5+ compatibility? Not sure if AWX and AAP maintain the same versioning, but something like this perhaps?
```suggestion
===  Parameter change for {awx} 2.5+ compatibility
```

Lennonka: I'm confused but I'm sure that Ewoud and other writers will help you to finish it.

ekohl: > Is it a host parameter or a global setting?

It's parameters that affect hosts. That means it can be a global parameter, but also a hostgroup parameter or a host parameter. However, not a global setting.

nofaralfasi: > Perhaps worth to call out AAP 2.5+ compatibility? Not sure if AWX and AAP maintain the same versioning, but something like this perhaps?

I couldn't find any references to AAP in our documentation, so I wasn't sure if we should explicitly mention it. However, specifying the supported AAP versions would make this change clearer. Should we include it?

guides/doc-Release_Notes/topics/foreman.adoc
nofaralfasi: Do you mean that the current phrasing is not enough?

aneta-petrova: Would one of these suggestions help?

At the very least, this is required for style review:
```suggestion
By default, the API path is set to `/api/controller/v2`.
If you are using an older version of {awx} that requires a different API path, update it manually to `/api/v2`.
```
But even better, can we be specific about the versions (is 2.3 the right one?):
```suggestion
By default, the API path is set to `/api/controller/v2`.
If you are using {awx}{nbsp}2.3, update the API path manually to `/api/v2`.
```
And according to certain resources downstream, we could also recommend users to avoid upgrading altogether:
```suggestion
By default, the API path is set to `/api/controller/v2`.
If you are using {awx}{nbsp}2.3, update the API path manually to `/api/v2` or avoid upgrading to {Project} {nbsp}3.14 until you upgrade to {awx}{nbsp}2.5.
```
The last suggestion sounds a lot like a known issue. I don't know how those are handled upstream but it could be an option too.

aneta-petrova: Is it okay that the Ansible config guide still references `ansible_tower_fqdn`? https://docs.theforeman.org/3.14/Managing_Configurations_Ansible/index-katello.html#Configuring_Provisioning_Callback_for_a_Host_ansible

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -34,6 +34,11 @@ Clients will remain supported.
 For more details, see the https://community.theforeman.org/t/drop-debian-11-ruby-2-7-and-nodejs-14-support-in-foreman-3-14/40503[removal RFC].
 endif::[]
 
+=== {awx} Parameter Change
+The `ansible_tower_fqdn` parameter has been removed and replaced with `ansible_tower_api_url`. This new parameter includes both the previous `ansible_tower_fqdn` value and the required API path for {awx}.
```
aneta-petrova, line 38:
```suggestion
The `ansible_tower_fqdn` parameter has been removed and replaced with `ansible_tower_api_url`.
This new parameter includes both the previous `ansible_tower_fqdn` value and the required API path for {awx}.
```

guides/doc-Release_Notes/topics/foreman.adoc
ekohl: Good catch. It isn't because that's going away. Thanks for proving the point that I often make: release notes are a great time to review a feature.

ekohl: > But even better, can we be specific about the versions (is 2.3 the right one?):

I'd prefer a variation on this. Say something like:
```suggestion
By default, the API path is set to `/api/controller/v2` as used by {awx}{nbsp}2.5 and newer.
If you are using an older version of {awx}, update the API path manually to `/api/v2`.
```

guides/doc-Release_Notes/topics/foreman.adoc
nofaralfasi: That's why I didn't explicitly mention it. If I understood you correctly, we should explicitly mention versioning for AAP to ensure clarity.

guides/doc-Release_Notes/topics/foreman.adoc
nofaralfasi: I agree, because the same applies for AAP 2.4 (using the old path `/api/v2`).

guides/doc-Release_Notes/topics/foreman.adoc
ekohl: I think it helps readers, but wasn't quite sure how to best phrase it.

guides/common/modules/proc_configuring-provisioning-callback-for-a-host.adoc
```diff
@@ -47,8 +47,8 @@ For more information about job templates, see http://docs.ansible.com/automation
 |====
 |Name |Value |Description
 |`ansible_tower_provisioning` |true |Enables Provisioning Callback.
-|`ansible_tower_fqdn` |_controller.example.com_ |The fully qualified domain name (FQDN) of your {awx}.
-Do not add `https` because this is appended by {Project}.
+|`ansible_tower_api_url` |_https://controller.example.com/api/controller/v2_ |Defines the fully qualified domain name (FQDN) of your {awx} with the required API path.
+For {awx}{nbsp}2.4 or older, use `/api/v2` instead of `/api/controller/v2`.
```
ekohl, line 51: I think it's best to first submit this to `master` and then cherry pick it. Otherwise we'll have a regression on the next version.

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -34,6 +34,13 @@ Clients will remain supported.
 For more details, see the https://community.theforeman.org/t/drop-debian-11-ruby-2-7-and-nodejs-14-support-in-foreman-3-14/40503[removal RFC].
 endif::[]
 
+=== {awx} Parameter Change
+The `ansible_tower_fqdn` parameter has been removed and replaced with `ansible_tower_api_url`.
+This new parameter includes both the previous `ansible_tower_fqdn` value and the required API path for {awx}.
+
+By default, the API path is set to `/api/controller/v2` as used by {awx}{nbsp}2.5 and newer.
```
ekohl, line 41: This now expands to `AWX` in upstream (and is unused in downstream). Given the different versioning I now think it'd be best to expand it manually
```suggestion
By default, the API path is set to `/api/controller/v2` as used by Ansible Automation Controller{nbsp}2.5 and newer.
```

guides/common/modules/proc_configuring-provisioning-callback-for-a-host.adoc
```diff
@@ -47,8 +47,8 @@ For more information about job templates, see http://docs.ansible.com/automation
 |====
 |Name |Value |Description
 |`ansible_tower_provisioning` |true |Enables Provisioning Callback.
-|`ansible_tower_fqdn` |_controller.example.com_ |The fully qualified domain name (FQDN) of your {awx}.
-Do not add `https` because this is appended by {Project}.
+|`ansible_tower_api_url` |_https://controller.example.com/api/controller/v2_ |Defines the fully qualified domain name (FQDN) of your {awx} with the required API path.
```
ekohl, line 50: Looking at writers here for the exact phrasing. I mostly care that this is no longer the FQDN but rather a URL (so including a scheme and path).
```suggestion
|`ansible_tower_api_url` |_https://controller.example.com/api/controller/v2_ | Defines the URL of your {awx}, including the required API path.
```

guides/doc-Release_Notes/topics/foreman.adoc
nofaralfasi: After reviewing the AWX versions on https://github.com/ansible/awx/releases, I believe we need to distinguish between AWX and AAP, as they have different versioning.

nofaralfasi: I couldn't find any mention of the new API path in their docs https://ansible.readthedocs.io/projects/awx/en/latest/rest_api/browseable.html.

ekohl: I just noticed that `{awx}` expands to `Ansible Automation Controller`. That abbreviates to AAC, not AAP. Looking at https://docs.redhat.com/en/documentation/red_hat_ansible_automation_platform/2.5 I'd also expect it to be Ansible Automation Platform. It was changed in ec21fdfae90872042875e6956f73b9457df1350e.

I also get the impression that Controller is a sub-project of platform. Upstream for that part appears to be https://docs.ansible.com/automation-controller/latest/html/release-notes/index.html.

Not sure what my conclusion is, other than to be utterly confused.

nofaralfasi: We can add a note advising users to verify the correct API path for their instance if they're unsure about their 
version.

guides/doc-Release_Notes/topics/foreman.adoc
nofaralfasi: But AAC and AAP are two different things. See https://access.redhat.com/support/policy/updates/ansible-automation-platform#packages-and-versions.

guides/doc-Release_Notes/topics/foreman.adoc
ekohl: I think that's best. Let's not spend too much time fine tuning the perfect answer here.

ekohl (commented):
I think something went wrong in changing the base branch because there are a lot of unrelated changes. In my experience that's a GH feature to avoid.

nofaralfasi:
> I think something went wrong in changing the base branch because there are a lot of unrelated changes. In my experience that's a GH feature to avoid.

Yeah, it took me some time to fix that. Hope it's ok now.

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -12,6 +12,14 @@ There are no highlights with Foreman {ProjectVersion}.
 // If this section would be empty otherwise, uncomment the following line:
 There are no upgrade warnings with Foreman {ProjectVersion}.
```
ekohl, line 13: There now is an upgrade warning
```suggestion
```

guides/doc-Release_Notes/topics/foreman.adoc
lumarel: Also giving my 2 cents here now 🙂 
I think the differentiation has to be made more broad overall, as far as I can see in my installs of AWX, the API path never changed, this must be a AAP specific change.
At least AWX 24.6.1, which was the last version before the big rewrite, which is going on since last year, still uses `/api/v2`.

I suspect due to the multi-project nature of AAP a reverse-proxy applied sub-path got added in AAP 2.5.

lumarel:
As already mentioned above, I don't see any change to `/api/controller/v2/` for AWX,
also made sure to spin up a version of the current development version now, and also there it's still `/api/v2/`

So must be a AAP only change 👍🏻

guides/common/modules/proc_configuring-provisioning-callback-for-a-host.adoc
```diff
@@ -47,8 +47,9 @@ For more information about job templates, see http://docs.ansible.com/automation
 |====
 |Name |Value |Description
 |`ansible_tower_provisioning` |true |Enables Provisioning Callback.
-|`ansible_tower_fqdn` |_controller.example.com_ |The fully qualified domain name (FQDN) of your {awx}.
-Do not add `https` because this is appended by {Project}.
+|`ansible_tower_api_url` |_https://controller.example.com/api/controller/v2_ | Defines the URL of your {awx}, including the required API path.
+For older versions of {awx}, use `/api/v2` instead of `/api/controller/v2`.
```
maximiliankolb, line 51: Older than which version?

guides/common/modules/proc_configuring-provisioning-callback-for-a-host.adoc
```diff
@@ -47,8 +47,9 @@ For more information about job templates, see http://docs.ansible.com/automation
...
+If unsure, check your instance's API endpoints to verify the correct path.
```
maximiliankolb, line 52:
```suggestion
If unsure, check the API endpoints of your instance to verify the correct path.
```

guides/common/modules/proc_configuring-provisioning-callback-for-a-host.adoc
nofaralfasi: AAP 2.5 - See the discussion here: https://github.com/theforeman/foreman-documentation/pull/3720#discussion_r1987691773.  

**Summary:** These changes are required for AAP 2.5+. If you're using an older AAP version or AWX (instead of AAP), these changes don't apply. In that case, you'll need to manually set the `ansible_tower_api_url` endpoint to `/api/v2` instead of `/api/controller/v2`.

nofaralfasi:
Can someone from the docs team approve and merge this PR? @maximiliankolb @aneta-petrova

maximiliankolb (approved):
style-wise LGTM

maximiliankolb:
* If the code change is in Foreman 3.14 already, then you'll need to drop any changes to the release notes on this PR and open another PR with the same changes against 3.14. Refs https://github.com/theforeman/foreman-documentation/pull/3720#discussion_r2006006880
* We can merge the change in `guides/common/modules/proc_configuring-provisioning-callback-for-a-host.adoc` to "master" and then cherry-pick to 3.14.
* I take Lukas' comment as tech ACK/test. Refs https://github.com/theforeman/foreman-documentation/pull/3720#issuecomment-2744410916

cc @nofaralfasi

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -34,6 +34,14 @@ Clients will remain supported.
 For more details, see the https://community.theforeman.org/t/drop-debian-11-ruby-2-7-and-nodejs-14-support-in-foreman-3-14/40503[removal RFC].
 endif::[]
 
+=== {awx} Parameter Change
+The `ansible_tower_fqdn` parameter has been removed and replaced with `ansible_tower_api_url`.
+This new parameter includes both the previous `ansible_tower_fqdn` value and the required API path for {awx}.
+
+By default, the API path is set to `/api/controller/v2`.
+If you are using an older {awx} version, update the API path manually to `/api/v2`.
```
aneta-petrova, line 42: This might raise questions about which version you mean. I have no idea here either but I'm going to suggest something and hope you will correct me :) My point is to be more specific:

```suggestion
If you are using an {awx} version earlier than X.Y, update the API path manually to `/api/v2`.
```

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -34,6 +34,14 @@ Clients will remain supported.
...
+If you're unsure which version you have, check your instance's API endpoints to confirm the correct path.
```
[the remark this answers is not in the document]
nofaralfasi: No, it means the user's AAP/AWX instance.

guides/doc-Release_Notes/topics/foreman.adoc
nofaralfasi: See previous discussion here https://github.com/theforeman/foreman-documentation/pull/3720#discussion_r2009961092

maximiliankolb: I dont't quite understand this. Are you saying that it does not make any sense from a tech perspective to mention the version? (aka. declining the suggestion)? Please let me know; then I will merge as is.

nofaralfasi: This is the Jira ticket if you need more context https://issues.redhat.com/browse/SAT-30761
The issue is that we have no mentioning of the AAP in our docs, only AAC (Ansible Automation Controller) for downstream, or AWX for upstream.
However, this change is required for AAP only. AWX remains without any changes.

maximiliankolb (approved):

PR #3724, opened 2025-03-11
Deprecate oVirt

stejskalleos:
oVirt integration is going to be deprecated in 3.15 and removed in 3.16

#### Why are you introducing these changes?
https://community.theforeman.org/t/rfc-deprecation-and-removal-of-ovirt-compute-resource/42527
https://projects.theforeman.org/issues/38266
https://issues.redhat.com/browse/SAT-12005

#### Anything else to add? 
@maximiliankolb Atix folks probably take over the ownership and extract the code to the plugin, so there should be a doc variant for your docs.

### DIFF
--- a/guides/common/modules/proc_performing-post-upgrade-tasks.adoc
+++ b/guides/common/modules/proc_performing-post-upgrade-tasks.adoc
@@ -1,6 +1,9 @@
 [id="performing-post-upgrade-tasks_{context}"]
 = Performing post-upgrade tasks
 
+* Optional: If you are utilizing the oVirt compute resource, it's crucial to migrate your data by executing the `foreman-rake ovirt:drop` task.
+Remember that with the upcoming release of version 3.16, oVirt will be officially phased out, and this migration will be automatically managed during the upgrade process.
+
 * Optional: If the default provisioning templates have been changed during the upgrade, recreate any templates cloned from the default templates.
 If the custom code is executed before and/or after the provisioning process, use custom provisioning snippets to avoid recreating cloned templates.
 For more information about configuring custom provisioning snippets, see {ProvisioningDocURL}Creating_Custom_Provisioning_Snippets_provisioning[Creating Custom Provisioning Snippets] in _{ProvisioningDocTitle}_.
--- a/guides/doc-Release_Notes/topics/foreman.adoc
+++ b/guides/doc-Release_Notes/topics/foreman.adoc
@@ -15,4 +15,10 @@ There are no upgrade warnings with Foreman {ProjectVersion}.
 [id="foreman-deprecations"]
 == Deprecations
 
-There are no deprecations with Foreman {ProjectVersion}.
+// There are no deprecations with Foreman {ProjectVersion}
+
+=== Compute resources - {oVirt}
+
+* The {oVirt} compute resource has been deprecated and will be removed in the 3.16 release.
+* Migrate your data with the `foreman-rake ovirt:drop` task.
+* https://community.theforeman.org/t/rfc-deprecation-and-removal-of-ovirt-compute-resource/42527[RFC: Deprecation and removal of oVirt Compute Resource]


guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -15,4 +15,6 @@ There are no upgrade warnings with Foreman {ProjectVersion}.
 [id="foreman-deprecations"]
 == Deprecations
 
-There are no deprecations with Foreman {ProjectVersion}.
+=== Compute resources - oVirt
```
maximiliankolb, line 18:
```suggestion
=== Compute resources - oVirt

```

Please add an empty line between the heading and the list.

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -15,4 +15,6 @@ There are no upgrade warnings with Foreman {ProjectVersion}.
...
+* The oVirt compute resource has been deprecated and will be removed in a future release.
+* Migrate your data with the "foreman-rake ovirt:drop" task.
```
maximiliankolb, line 20:
```suggestion
* Migrate your data with the `foreman-rake ovirt:drop` task.
```

suggestion: Use emphasis for inline code snippet.

guides/doc-Release_Notes/topics/foreman.adoc
maximiliankolb, line 18: FYI: There's an attribute for oVirt: https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-base.adoc#L107 I am OK as is; I am also OK using oVirt in the heading and then explaining that it affects oVirt itself and any downstream products such as RHV and OLVM.

maximiliankolb (commented):
Thanks for the PR. I will keep you updated if there's any news from ATIX about oVirt+Foreman/Katello.

guides/doc-Release_Notes/topics/foreman.adoc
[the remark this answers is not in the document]
aneta-petrova: ... or just what @maximiliankolb said :) Looks like he was faster with his review.

stejskalleos:
Updated formatting. About the rake task description, I'll describe it in the release notes as well, once the https://github.com/theforeman/foreman/pull/10469 get merged.

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -15,4 +15,7 @@ There are no upgrade warnings with Foreman {ProjectVersion}.
 [id="foreman-deprecations"]
 == Deprecations
 
-There are no deprecations with Foreman {ProjectVersion}.
+=== Compute resources - {oVirt}
+
+* The {oVirt} compute resource has been deprecated and will be removed in a future release.
```
ekohl, line 20: A concrete timeline is always appreciated, if available.

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -15,4 +15,7 @@ There are no upgrade warnings with Foreman {ProjectVersion}.
 [id="foreman-deprecations"]
 == Deprecations
 
-There are no deprecations with Foreman {ProjectVersion}.
+=== Compute resources - {oVirt}
```
ekohl, line 18: We usually refer to the RFC so please include a link to https://community.theforeman.org/t/rfc-deprecation-and-removal-of-ovirt-compute-resource/42527.

ekohl (commented):
Please don't use `Fixes #SAT-something`. That is really only used for Redmine.

ekohl:
> @maximiliankolb Atix folks probably take over the ownership and extract the code to the plugin, so there should be a doc variant for your docs.

If there is a new plugin and that's used in upstream, then IMHO it shouldn't be deprecated. Instead we should point out (at least in upstream) that it's going be migrated.

Until that's decided I don't think we should merge this. In other words, only once consensus is reached in https://community.theforeman.org/t/rfc-deprecation-and-removal-of-ovirt-compute-resource/42527 should we proceed with steps.

stejskalleos:
In the [RFC](https://community.theforeman.org/t/rfc-deprecation-and-removal-of-ovirt-compute-resource/42527/12) there are no updates regarding to migrating oVirt to plugin, so the plan goes forward and in 3.16 we are removing it without any future support.

@ekohl, what must be done before merging it into the 3.15 doc?

stejskalleos:
Rebased & Updated. @Lennonka review please

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -15,4 +15,10 @@ There are no upgrade warnings with Foreman {ProjectVersion}.
 [id="foreman-deprecations"]
 == Deprecations
 
-There are no deprecations with Foreman {ProjectVersion}.
+// There are no deprecations with Foreman {ProjectVersion}
+
+=== Compute resources - {oVirt}
+
+* The {oVirt} compute resource has been deprecated and will be removed in a the 3.16 release.
+* Migrate your data with the `foreman-rake ovirt:drop` task.
```
ekohl, line 23: We have found that it's a bad practice to have procedures in release notes. Things users must do belong in the upgrade guide. That may be something we do in the Foreman 3.16 upgrade guide and not 3.15 though.

For the writers, have you considered what would be best here?

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -15,4 +15,10 @@ There are no upgrade warnings with Foreman {ProjectVersion}.
...
+* https://community.theforeman.org/t/rfc-deprecation-and-removal-of-ovirt-compute-resource/42527
```
Lennonka, line 24:
```suggestion
* https://community.theforeman.org/t/rfc-deprecation-and-removal-of-ovirt-compute-resource/42527[RFC: Deprecation and removal of oVirt Compute Resource]
```
I'd prefer a labeled link to a naked one. :)

guides/doc-Release_Notes/topics/foreman.adoc
Lennonka: cc @aneta-petrova @AkshayGadhaveRH

aneta-petrova: Answering the question because I was CC'ed (and it would be a bit rude not to reply :)): No, I personally have not considered what would be best.

To add a bit more, I don't entirely agree with the conclusion that it's a bad practice, although AFAIK I was not part of that conversation so I might not know all the arguments that were shared.

I also don't agree with the statement that line 23 is a procedure. It seems more of a heads-up than a procedure to me, which is why I'd be fine with keeping it here.

Adding a step to the upgrade guide would probably be a good idea but for us, these types of tasks typically originate downstream from the Satellite team developing the feature. Then the RN could possibly link to the upgrade guide for more details.

ekohl: > To add a bit more, I don't entirely agree with the conclusion that it's a bad practice, although AFAIK I was not part of that conversation so I might not know all the arguments that were shared.

I'll refine it a bit: mandatory procedures that must be done in an upgrade process do not belong in the release notes.

This particular case is a bit of an edge case because in this release it's not mandatory.

I'll share my opinion. In Foreman 3.15 we can add it as an optional post-upgrade step to fully remove oVirt from an installation. Then in Foreman 3.16 we should move that optional step to a mandatory pre-upgrade step. In addition on the engineering side we should implement a check that prevents upgrades if the mandatory pre-upgrade step wasn't executed.

This way we encourage users to prepare in 3.15 and be ready, but make sure they can't go wrong in upgrading to 3.16.

The release notes should make users aware of the strategy and possibly point to the specific procedures.

> I also don't agree with the statement that line 23 is a procedure. It seems more of a heads-up than a procedure to me, which is why I'd be fine with keeping it here.

IMHO it's incomplete. A complete procedure would also remove the `foreman-ovirt` package so users can no longer create oVirt compute resources either. Downstream that wouldn't work since `foreman-ovirt` is always installed but upstream it's an optional package.

stejskalleos: @aneta-petrova @Lennonka so should I add steps to post-upgrade instructions? If yes could you please help me where and how can I do it?

stejskalleos:
rebased and updated

guides/doc-Release_Notes/topics/foreman.adoc
aneta-petrova: @stejskalleos Sure! In the Upgrade guide, we have a whole section for post-upgrade tasks: https://docs.theforeman.org/nightly/Upgrading_Project/index-katello.html#performing-post-upgrade-tasks_upgrading-connected
The corresponding file: https://github.com/theforeman/foreman-documentation/blob/master/guides/common/modules/proc_performing-post-upgrade-tasks.adoc

This section already has one "Optional:" bullet point so I think you could add another one for oVirt removal.

guides/common/modules/proc_performing-post-upgrade-tasks.adoc
```diff
@@ -1,6 +1,9 @@
 [id="performing-post-upgrade-tasks_{context}"]
 = Performing post-upgrade tasks
 
+* Optional: If you are utilizing the oVirt compute resource, it's crucial to migrate your data by executing the `foreman-rake ovirt:drop` task.
```
aneta-petrova, line 4: Suggestions to address style requirements: Attribute for oVirt, avoiding contractions ("it's"), and keep it simple by using plain imperative voice ("migrate your data" rather than "it is crucial to migrate your data").

Suggestion to improve clarity: Where do I run it? On the server, right?

```suggestion
* Optional: If you are utilizing the {oVirt} compute resource, migrate your data by executing the `foreman-rake ovirt:drop` task on your {ProjectServer}.
```

guides/common/modules/proc_performing-post-upgrade-tasks.adoc
```diff
@@ -1,6 +1,9 @@
...
+Remember that with the upcoming release of version 3.16, oVirt will be officially phased out, and this migration will be automatically managed during the upgrade process.
```
aneta-petrova, line 5: I wonder if we should avoid advertising specific changes for a particular release (because plans can change). May I suggest wording that's a bit more on the vague side, like "In a future version"?

```suggestion
In a future version, {oVirt} will be officially phased out, and this migration will be automatically managed during the upgrade process.
```

Lennonka:
Since 3.15 has been branched, we don't want this on master anymore. I've filed two other PRs instead:

- #3870 
- #3871 

Closing this one.

PR #3728, opened 2025-03-12
Update notes for 3.14.0 GA

ColeHiggins2:
Adding Headline Features for 3.14.0 Release notes

### DIFF
--- a/guides/doc-Release_Notes/topics/foreman.adoc
+++ b/guides/doc-Release_Notes/topics/foreman.adoc
@@ -4,7 +4,17 @@
 [id="foreman-headline-features"]
 == Headline Features
 
-There are no highlights with Foreman {ProjectVersion}.
+=== IPV6 Support
+Satellite deployment and operation in IPv6 networks is now fully supported. Dual stack deployments have not been tested and are not supported.
+
+Provisioning over IPv6 is supported on bare metal hosts.
+For compute resources, you can define the machine outside Satellite on the compute resource and then you can provision the machine as bare metal in Satellite.
+
+=== Support for Secure Boot provisioning
+Provide support for Secure Boot provisioning on bare-metal, VMware vSphere and Libvirt across multiple operating systems.
+
+=== Invalidating JWT for global registration
+This will give users the ability to edit user permissions and invalidate JWT tokens for other users, invalidate self's token, and admins can invalidate selfs and other user tokens
 
 [id="foreman-upgrade-warnings"]
 == Upgrade Warnings
--- a/guides/doc-Release_Notes/topics/foreman.adoc
+++ b/guides/doc-Release_Notes/topics/foreman.adoc
@@ -18,8 +18,8 @@ New virtual machines with Secure Boot enabled will be created with the default k
 
 === Invalidate JWT for global registration
 
-This will give users the ability to edit user permissions and invalidate JWT tokens for otherers can now invalidate their own tokens for global registration.
-Users with the `edit_users` permission can also invalidate all tokens for all users in a single action.users, invalidate self's token, and admins can invalidate selfs and other user tokens
+Users can now invalidate their own tokens for global registration.
+Users with the `edit_users` permission can also invalidate all tokens for all users in a single action.
 
 [id="foreman-upgrade-warnings"]
 == Upgrade Warnings
--- a/guides/doc-Release_Notes/topics/foreman.adoc
+++ b/guides/doc-Release_Notes/topics/foreman.adoc
@@ -9,7 +9,7 @@
 Deployment and operation in IPv6-only networks is now fully supported.
 
 Provisioning over IPv6 is supported on bare metal hosts.
-For compute resources, you can define the machine outside Satellite on the compute resource and then you can provision the machine as bare metal in Satellite.
+For compute resources, you can define the machine outside Foreman on the compute resource and then you can provision the machine as bare metal in Foreman.
 
 === Secure Boot provisioning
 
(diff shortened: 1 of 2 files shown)


ColeHiggins2:
@ekohl can you review/tag anyone else you think might be useful to get feedback from :)

ekohl (changes requested):
I started to review https://github.com/theforeman/theforeman.org/pull/2221, but the same notes apply. I should have started the review here.

@Lennonka I'd appreciate if you take a look as well

cc @stejskalleos @nofaralfasi @girijaasoni

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -4,7 +4,20 @@
 [id="foreman-headline-features"]
 == Headline Features
 
-There are no highlights with Foreman {ProjectVersion}.
+=== IPv6 Support
+
+Deployment and operation in IPv6-only networks is now fully supported.
+
+Provisioning over IPv6 is supported on bare metal hosts.
+For compute resources, you can define the machine outside Satellite on the compute resource and then you can provision the machine as bare metal in Satellite.
```
maximiliankolb, line 12:
```suggestion
For compute resources, you can define the machine outside Foreman on the compute resource and then you can provision the machine as bare metal in Foreman.
```

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -4,7 +4,20 @@
 [id="foreman-headline-features"]
 == Headline Features
 
-There are no highlights with Foreman {ProjectVersion}.
+=== IPv6 Support
+
+Deployment and operation in IPv6-only networks is now fully supported.
+
+Provisioning over IPv6 is supported on bare metal hosts.
+For compute resources, you can define the machine outside Foreman on the compute resource and then you can provision the machine as bare metal in Foreman.
```
ekohl, line 12: Is there any reference on why image mode in IPv6-only wouldn't work?

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -4,7 +4,20 @@
...
+
+=== Secure Boot provisioning
+
+Support for Secure Boot provisioning on bare-metal, VMware vSphere and Libvirt across multiple operating systems.
```
ekohl, line 16: @nofaralfasi is this a good addition? I want to also imply that VMs will be created with Secure Boot enabled.
```suggestion
Provisioning with Secure Boot is now supported on bare-metal, VMware vSphere and Libvirt across multiple operating systems.
New virtual machines will be created with the default keys enrolled.
```

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -4,7 +4,20 @@
...
+
+=== Invalidate JWT for global registration
+
+This will give users the ability to edit user permissions and invalidate JWT tokens for other users, invalidate self's token, and admins can invalidate selfs and other user tokens
```
ekohl, line 20: In release notes you are talking to the user, not about the user. When you write "This will give users" you're talking in the future tense, but the change has already landed so it should be written in the present tense.

The exact phrasing can probably be best reviewed by a writer, but how about:

```suggestion
Users can now invalidate their own tokens for global registration.
Users with the `edit_users` permission can also invalidate all tokens for all users in a single action.
```

guides/doc-Release_Notes/topics/foreman.adoc
nofaralfasi: I don't quite follow. Are you referring to all new VMs being created with the default keys enrolled (which isn't the case)? Or are you talking specifically about new VMs with Secure Boot enabled? If it's the latter, why is it important to mention that here? Isn't that implicit?

nofaralfasi: LGTM. Thanks!

guides/doc-Release_Notes/topics/foreman.adoc
ColeHiggins2: not sure tbh, just what I was given

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -4,7 +4,22 @@
...
+
+=== Secure Boot provisioning
+
+Provisioning with Secure Boot is now supported on bare-metal, VMware vSphere and Libvirt across multiple operating systems.
+New virtual machines with Secure Boot enabled will be created with the default keys enrolled.
+
+=== Invalidate JWT for global registration
+
+This will give users the ability to edit user permissions and invalidate JWT tokens for otherers can now invalidate their own tokens for global registration.
+Users with the `edit_users` permission can also invalidate all tokens for all users in a single action.users, invalidate self's token, and admins can invalidate selfs and other user tokens
```
ekohl, line 22: I'm not sure what happened to this sentence my suggestion, but this doesn't read well to me.

The first sentence is also just wrong in saying that it gives users the ability to edit user permissions. It doesn't. Users that have the `edit_users` permission can now invalidate tokens for other users. In my suggestion I tried to split it up into 2 clear cases: something that's true for all users and something that's true for users with a specific permission.

Also, I've never seen `self's` before and it really confuses me. I'm not a native speaker, but if I look at https://www.merriam-webster.com/dictionary/self I see no definition that fits. Shouldn't there be a possessive pronoun in this context?

ColeHiggins2:
@ekohl I might have missed your original suggestion. please check again

ekohl (approved):

PR #3775, opened 2025-04-09
Fix vague AK parameter instruction step

Lennonka:

Introducing a snippet for a repeating instruction step that was too vague in most places of provisioning

Specific parameter is needed and docs failed to mention it.

https://issues.redhat.com/browse/SAT-32160

#### Checklists

* [x] Foreman 3.11/Katello 4.13 (orcharhino 6.11 on EL8 only; orcharhino 7.0 on EL8+EL9; orcharhino 7.1 with Leapp)

### DIFF
--- a/guides/common/modules/snip_step-parameter-ak.adoc
+++ b/guides/common/modules/snip_step-parameter-ak.adoc
@@ -1,4 +1,4 @@
-. Click the *Parameters* tab.
-Add the `kt_activation_keys` parameter and enter the name of the activation key as the value.
+. On the *Parameters* tab, click *Add parameter*.
+Add a parameter named `kt_activation_keys`, select the *string* type, and enter the name of the activation key as the value.
 The activation key has to belong to the same organization as your host.
 You can also enter a comma-separated list of multiple activation keys.
--- a/guides/common/modules/snip_step-parameter-ak.adoc
+++ b/guides/common/modules/snip_step-parameter-ak.adoc
@@ -1,0 +1,4 @@
+. Click the *Parameters* tab.
+Add the `kt_activation_keys` parameter and enter the name of the activation key as the value.
+The activation key has to belong to the same organization as your host.
+You can also enter a comma-separated list of multiple activation keys.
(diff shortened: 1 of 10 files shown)


guides/common/modules/snip_step-parameter-ak.adoc
```diff
@@ -0,0 +1,3 @@
+. Click the *Parameters* tab.
+Add the `kt_activation_keys` parameter and enter the name of the activation key as the value.
```
stejskalleos, line 2: I know that Katello is now suggesting available keys under the current organization, but would it be helpful to mention that the keys must be under the same organization as the host is or is going to be?

Lennonka: Yes, that's a good point because I don't think that Foreman is aware that the parameter value is an AK when it's entered as a parameter.

guides/common/modules/snip_step-parameter-ak.adoc
```diff
@@ -0,0 +1,3 @@
...
+You can also enter a comma-separated list of multiple activation keys.
```
Lennonka, line 3: @stejskalleos Is this sentence accurate? It was a guess :)

stejskalleos: The thing is that on some pages, like the host group, there is a UI for selecting activation keys, so you don't have to write them manually. But for some pages, AFAIK, the input is just a text field, where you have to write them manually.

cc @jeremylenz

jeremylenz: There's an Activation Keys tab for hostgroups, but not for creating or editing hosts. The Activation Keys tab is nice UI for adding the parameter, so the end result is always the same. The comma-separated list will work fine. :)

Lennonka:
@stejskalleos Ack?

stejskalleos (approved):

guides/common/modules/snip_step-parameter-ak.adoc
```diff
@@ -0,0 +1,4 @@
+. Click the *Parameters* tab.
+Add the `kt_activation_keys` parameter and enter the name of the activation key as the value.
```
maximiliankolb, line 2:
```suggestion
Add a parameter named `kt_activation_keys`, select the *string* type, and enter the name of the activation key as the value.
```

suggestion: Add type of the parameter. I am unsure about the wording, so I had a look at existing docs: https://github.com/theforeman/foreman-documentation/blob/master/guides/common/modules/proc_installing-and-configuring-puppet-agent-during-host-registration.adoc?plain=1#L52

maximiliankolb (approved):
Thanks Lena, LGTM.

amolpati30:
Ack

PR #3782, opened 2025-04-14
Clean up master.adoc in Planning

aneta-petrova:

Include "overview & concepts" and "deployment planning" as assemblies rather than AsciiDoc parts.

To clean up master.adoc a bit and also to resolve errors about too many level 0 headings.

I'm adding a new part/assembly to the Planning guide and this would help keep things a bit more tidy.

#### Checklists

### DIFF
--- a/guides/common/assembly_foreman-deployment-planning.adoc
+++ b/guides/common/assembly_foreman-deployment-planning.adoc
@@ -1,0 +1,7 @@
+include::modules/con_foreman-deployment-planning.adoc[leveloffset=+1]
+
+include::assembly_deployment-path.adoc[leveloffset=+2]
+
+include::assembly_common-deployment-scenarios.adoc[leveloffset=+2]
+
+include::assembly_provisioning-requirements.adoc[leveloffset=+2]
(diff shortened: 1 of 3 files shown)


maximiliankolb (approved):
split LGTM, though I would prefer to use a single "leveloffset=+1" in "master.adoc" for those assemblies and have the first file in the assemblies start without a leveloffset.

Looking at the built docs:
image

Your PR is clearly an improvement for maintenance and we should merge this, but maybe we can improve it further in a follow-up PR to reduce the number of nested modules?

aneta-petrova:
> split LGTM, though I would prefer to use a single "leveloffset=+1" in "master.adoc" for those assemblies and have the first file in the assemblies start without a leveloffset.

Like this? Can you please re-review?

> Your PR is clearly an improvement for maintenance and we should merge this, but maybe we can improve it further in a follow-up PR to reduce the number of nested modules?

That was the benefit of the AsciiDoc parts -- this wasn't a problem with them. I agree that this kind of improvement would be nice but I'm not volunteering :upside_down_face:

maximiliankolb:
> Like this? Can you please re-review?

Yes, LGTM

maximiliankolb (approved):
Thanks.

PR #3760, opened 2025-04-01
Updating find_unused_modules script

apinnick:
#### What changes are you introducing?

Added a line to detect snippets included in modules.

Added a comment that script does not work on file includes that contain attributes. User will have to ignore those files in the output.

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

Snippets included in modules were reported as false positives.

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)

I tested this with a fake snippet. It was detected.

No style review required.

#### Checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [ ] Foreman 3.14/Katello 4.16
* [ ] Foreman 3.13/Katello 4.15 (EL9 only)
* [ ] Foreman 3.12/Katello 4.14 (Satellite 6.16)
* [ ] Foreman 3.11/Katello 4.13 (orcharhino 6.11 on EL8 only; orcharhino 7.0 on EL8+EL9; orcharhino 7.1 with Leapp)
* [ ] Foreman 3.10/Katello 4.12
* [ ] Foreman 3.9/Katello 4.11 (Satellite 6.15; orcharhino 6.8/6.9/6.10)
* [ ] Foreman 3.8/Katello 4.10
* [ ] Foreman 3.7/Katello 4.9 (Satellite 6.14)
* We do not accept PRs for Foreman older than 3.7.

PR #3790, opened 2025-04-16
Implement find_unused_modules in CI

ekohl:

This rewrites find_unused_modules into Ruby and uses asciidoctor to determine which files are unused.

It also adds a workflow that GitHub Actions runs. To make that easier to consume it emits GitHub Annotations with warnings.

It came up in https://github.com/theforeman/foreman-documentation/pull/3760 that the current script has false positives. My version shouldn't have that, though it now does emit some unused files.

#### Checklists


guides/scripts/find_unused_modules
```diff
@@ -0,0 +1,49 @@
...
+
+unused = (all_modules - absolute_included).map do |path|
+  Pathname.new(path).relative_path_from(doc_root).to_s
+end
+
+if unused.any?
+  puts unused
+  if ENV.key?('GITHUB_ACTIONS')
+    unused.each do |file|
+      puts "::warning file=#{file}::Module is unused"
```
ekohl, line 46: Should this be an error? And if so, should we also fail the workflow so it becomes visible?

We'd need to resolve the current unused files, but I think that's best.

Lennonka:
@ekohl Great idea! Would you also consider extending the script to check for unused assemblies as well?

ekohl:
> Would you also consider extending the script to check for unused assemblies as well?

I did think about this yes but forgot to do it. Now added as an additional glob pattern.

guides/scripts/find_unused_modules
Lennonka: Gotcha

Lennonka (approved):
LGTM

apinnick (approved):

apinnick:
I will merge today so that I can see whether there are any problems.

apinnick:
Tested the GitHub action with my [test PR](https://github.com/theforeman/foreman-documentation/pull/3791). It works.

PR #3794, opened 2025-04-22
Remove snippet that is only used once

maximiliankolb:

Move content from snippet to procedure

This snippet is only included once in upstream and downstream docs. Therefore, to allow checking unused modules via GHA, I have moved the content from the snippet to the actual place the snippet was included.

This technically does not ensure that the content is rendered in previews for orcharhino on GitHub.

Refs PR 3792 on GitHub

#### Checklists


guides/common/modules/proc_registering-a-host.adoc
```diff
@@ -55,7 +55,53 @@ endif::[]
 endif::[]
 ifdef::orcharhino[]
 ifdef::suse_linux_enterprise_server[]
```
ekohl, line 57: I wondered if we could somehow set `suse_linux_enterprise_server` instead.

Btw, I recently learned you combine ifdefs with `and` by using `+` (https://docs.asciidoctor.org/asciidoc/latest/directives/ifdef-ifndef/#checking-multiple-attributes) so you can remove a lot of nested `ifdef` statements in this document:
```asciidoc
ifdef::orcharhino+suse_linux_enterprise_server[]
```

maximiliankolb: I would have to test if that works for downstream docs too, but yet. I am currently investigating how to handle Client OS-specific content. I would like to keep it as is for now and then, possibly, fix all of those instances in a follow-up PR.

ekohl: I would like it if we upstream had some sort of preview what's rendered downstream. Right now that really isn't true for orcharhino and that also makes reviews harder.

Lennonka (approved):

PR #3799, opened 2025-04-23
Add documentation on security features

jherrman:

Providing a conceptual module on the available security features that a user can avail of in order to harden their deployment.

Based on https://issues.redhat.com/browse/SAT-30719

#### Checklists

### DIFF
--- a/guides/common/assembly_preparing-environment-for-installation.adoc
+++ b/guides/common/assembly_preparing-environment-for-installation.adoc
@@ -29,6 +29,8 @@ include::modules/proc_verifying-dns-resolution.adoc[leveloffset=+1]
 
 ifdef::katello,orcharhino,satellite[]
 include::modules/proc_tuning-with-predefined-profiles.adoc[leveloffset=+1]
+
+include::modules/con_security-features.adoc[leveloffset=+1]
 endif::[]
 
 ifdef::parent-context[:context: {parent-context}]
--- a/guides/common/modules/con_security-features.adoc
+++ b/guides/common/modules/con_security-features.adoc
@@ -1,0 +1,39 @@
+:_newdoc-version: 2.18.4
+:_template-generated: 2025-04-23
+:_mod-docs-content-type: CONCEPT
+
+[id="security-features_{context}"]
+= Security features in {Project}
+
+When deploying {Project}, consider implementing security features that enhance the overall security of your system.
+{Project} supports multiple security mechanisms, including FIPS mode, `fapolicyd`, and SELinux, to provide additional layers of protection.
+
+Federal Information Processing Standards (FIPS) Mode::
+{Project} supports running on FIPS-enabled RHEL hosts to comply with security standards for cryptographic modules. 
+However, {Project} itself does not currently support running in FIPS mode.
++
+FIPS mode enforces the use of validated cryptographic algorithms and modules.
+Enabling FIPS mode ensures that all cryptographic operations within the system adhere to strict security standards.
+
+File Access Policy Daemon (`fapolicyd`)::
+By enabling `fapolicyd` on your {Project} Server, you can monitor and control file and directory access, providing an additional security layer.
++
+This feature helps prevent unauthorized code execution and enhances system integrity. You can enable or disable `fapolicyd` on both {Project} and {ProjectServer} at any time, depending on your security requirements.
++
+For additional information about `fapolicyd`, see link:https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/security_hardening/assembly_blocking-and-allowing-applications-using-fapolicyd_security-hardening[Blocking and allowing applications by using fapolicyd] in _Security hardening_.
+
+Security-Enhanced Linux (SELinux)::
+{Project} supports SELinux, a mandatory access control system that restricts system access and reduces the risk of security vulnerabilities.
+SELinux enforces policies that define which processes can access specific system resources, providing protection against unauthorized access and exploitation.
++
+Running {Project} with SELinux in enforcing mode is recommended to maintain a high-security environment.
+It ensures that access control policies are strictly enforced, minimizing the risk of privilege escalation or unauthorized actions.
+
+Additional Security Recommendations::
+Regularly updating security policies and configurations, as well as monitoring system logs and security policies, can help detect and respond to potential threats effectively.
+
+////
+[role="_additional-resources"]
+.Additional resources
+* TODO - maybe link to procedure modules on enabling the individual security features, when there are any
+////


guides/common/modules/con_security-features.adoc
```diff
@@ -0,0 +1,39 @@
...
+:_mod-docs-content-type: CONCEPT
+
+[id="security-features_{context}"]
+= Security features in {Project}
+
+When deploying {Project}, consider implementing security features that enhance the overall security of your system.
+{Project} supports multiple security mechanisms, including FIPS mode, `fapolicyd`, and SELinux, to provide additional layers of protection.
+
+Federal Information Processing Standards (FIPS) Mode::
+{Project} supports running on FIPS-enabled RHEL hosts to comply with security standards for cryptographic modules.
```
AkshayGadhaveRH, line 12:
```suggestion
{Project} supports running on FIPS-enabled {RHEL} hosts to comply with security standards for cryptographic modules. 
```

guides/common/modules/con_security-features.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+Federal Information Processing Standards (FIPS) Mode::
+{Project} supports running on FIPS-enabled RHEL hosts to comply with security standards for cryptographic modules. 
+However, {Project} itself does not currently support running in FIPS mode.
++
+FIPS mode enforces the use of validated cryptographic algorithms and modules.
+Enabling FIPS mode ensures that all cryptographic operations within the system adhere to strict security standards.
+
+File Access Policy Daemon (`fapolicyd`)::
+By enabling `fapolicyd` on your {Project} Server, you can monitor and control file and directory access, providing an additional security layer.
```
AkshayGadhaveRH, line 19:
```suggestion
By enabling `fapolicyd` on your {ProjectServer}, you can monitor and control file and directory access, providing an additional security layer.
```

guides/common/modules/con_security-features.adoc
```diff
@@ -0,0 +1,39 @@
...
++
+This feature helps prevent unauthorized code execution and enhances system integrity. You can enable or disable `fapolicyd` on both {Project} and {ProjectServer} at any time, depending on your security requirements.
++
+For additional information about `fapolicyd`, see link:https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/security_hardening/assembly_blocking-and-allowing-applications-using-fapolicyd_security-hardening[Blocking and allowing applications by using fapolicyd] in _Security hardening_.
```
AkshayGadhaveRH, line 23:
```suggestion
For additional information about `fapolicyd`, see link:{RHELDocsBaseURL}9/html/security_hardening/assembly_blocking-and-allowing-applications-using-fapolicyd_security-hardening[Blocking and allowing applications by using fapolicyd] in _Security hardening_.
```

AkshayGadhaveRH (changes requested):
Made some small suggestions.

aneta-petrova:
I'm wondering whether this would be better placed somewhere else. Perhaps the Planning guide? The installation guide and especially its prerequisite sections should be kept short and to-the-point, very action-focused. The security features overview is not really that. It's more like a list of 'things that Foreman can do' rather than a list of 'things you need to do before installing Foreman'.

We've recently redesigned the structure of the Planning guide: https://docs.theforeman.org/nightly/Planning_for_Project/index-katello.html I'll admit I don't see a good place to move the security features overview section there right now but I wonder whether that's because I actually don't fully understand its use case (that use case is probably not 'Make sure you deploy Foreman in a secure way', because the module offers too little on that).

Then again, the installation guide already has useful on FIPS, SELinux, and fapolicyd: see 1.1. System requirements and 3.3. Optional: Using fapolicyd on Foreman server. That information is simply just buried within walls of text. Which makes me wonder: Is there a way to make the existing descriptions more visible (and perhaps enhancing them with new information) rather than add a new module that tries to explain the same concepts differently? I'm a bit worried that we're adding new content to try to fix current inefficiencies rather than addressing the root cause of these inefficiencies.

jherrman:
@aneta-petrova , you're right that this module is not a great match for the installation prep chapter. The Planning guide is definitely a better document to include this in. I'd appreciate your advice on the placing, though.

On the downstream side, I see there is an assembly on [Deployment considerations](https://docs.redhat.com/en/documentation/red_hat_satellite/6.16/html-single/overview_concepts_and_deployment_considerations/index#chap-Architecture_Guide-Deployment_Considerations), which seems like a good fit, but does not appear [in the Foreman version](https://docs.theforeman.org/nightly/Planning_for_Project/index-katello.html#foreman-deployment-planning), nor in [the source](https://github.com/theforeman/foreman-documentation/blob/master/guides/common/assembly_foreman-deployment-planning.adoc), for some reason. Is there something I'm missing in terms of the conditional content syntax, perhaps?

As for the connection between this module and the information in the Installation guide(s), I added a couple of links to make it more explicit. Admittedly, you could make an argument that we don't need this additional conceptual info at all (whether the current content on SELinux, fapolicyd, and FIPS really is inefficient is not something that https://issues.redhat.com/browse/SAT-30719 addresses), but as a planning consideration, I'd argue it's still valuable, even at the cost of a bit of duplication.

guides/common/modules/con_security-features.adoc
```diff
@@ -0,0 +1,43 @@
...
+By enabling `fapolicyd` on your {Project} Server, you can monitor and control file and directory access, providing an additional security layer.
++
+This feature helps prevent unauthorized code execution and enhances system integrity. You can enable or disable `fapolicyd` on both {Project} and {SmartProxyServer} at any time, depending on your security requirements.
++
+For additional information about `fapolicyd`, see {RHELDocsBaseURL}9/html/security_hardening/assembly_blocking-and-allowing-applications-using-fapolicyd_security-hardening[Blocking and allowing applications by using fapolicyd] in _Security hardening_.
++
+For instructions on enabling `fapolicyd`, see {InstallingServerDocURL}installing_satellite_server_in_a_connected_network_environment/installing_server_connected_{project-context}#using-fapolicyd-on-server_{project-context}[Using fapolicyd on Satellite Server
+] in _{InstallingServerDocTitle}_.
+
+https://docs.redhat.com/en/documentation/red_hat_satellite/6.16/html/installing_satellite_server_in_a_connected_network_environment/installing_server_connected_satellite#using-fapolicyd-on-server_satellite
```
ekohl, line 30: Debug left over?
```suggestion
```

guides/common/assembly_preparing-environment-for-installation.adoc
```diff
@@ -29,6 +29,8 @@ include::modules/proc_verifying-dns-resolution.adoc[leveloffset=+1]
 
 ifdef::katello,orcharhino,satellite[]
 include::modules/proc_tuning-with-predefined-profiles.adoc[leveloffset=+1]
+
+include::modules/con_security-features.adoc[leveloffset=+1]
```
ekohl, line 33: Looking at the chapter I think it all applies to any EL build
```suggestion
endif::[]

ifndef::foreman-deb[]
include::modules/con_security-features.adoc[leveloffset=+1]
```

guides/common/modules/con_security-features.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+Security-Enhanced Linux (SELinux)::
+{Project} supports SELinux, a mandatory access control system that restricts system access and reduces the risk of security vulnerabilities.
+SELinux enforces policies that define which processes can access specific system resources, providing protection against unauthorized access and exploitation.
++
+Running {Project} with SELinux in enforcing mode is recommended to maintain a high-security environment.
+It ensures that access control policies are strictly enforced, minimizing the risk of privilege escalation or unauthorized actions.
++
+Using SELinux in enforcing mode is required for {Project} Server to be installed on your environment. For additional information, see {InstallingServerDocURL}preparing_your_environment_for_installation_{project-context}#system-requirements_{project-context} in _{InstallingServerDocTitle}_.
```
ekohl, line 39: This contradicts each other: first it's recommended but then it's required. In the installation guide we say:
> SELinux must be enabled, either in enforcing or permissive mode.

In other words, it's currently recommended (and IMHO that's still correct).

ekohl (commented):
@aneta-petrova already covered what I wanted to say: it belongs more in planning. In the content journey @apinnick did we discussed this and concluded we should move it out of installation guides and into planning guides. Quoting the JIRA issue:

> We could create a short concept module called "Security considerations features"  in the "Preparing for installation" section of the Installation guide. Eventually, it would be moved to the "Planning" section, which has not yet been created.

This doesn't quite cover what was all discussed so I'll elaborate.

When you install we automatically adopt to your environment. When we detect fapolicyd we automatically install the required packages. selinux is always installed. Similar with FIPS we make it work. Users who want to use any of these features already know how they need to prepare their base OS install. In other words, no user action is needed.

What they do need to know is the that it exists and if so, what the limitations are. That belongs in planning.

I would prefer to add this new module to the planning guide (where is a good question) and drop the SELinux and FIPS security prerequisites from the installation guide. Additionally we can drop the "Optional: Using fapolicyd on Foreman server" section altogether.

guides/common/modules/con_security-features.adoc
```diff
@@ -0,0 +1,43 @@
+:_newdoc-version: 2.18.4
+:_template-generated: 2025-04-23
+:_mod-docs-content-type: CONCEPT
+
```
Lennonka, line 4:
```suggestion
```
Pls, remove.
FYI, there's the `-D` option in newdoc to avoid this metadata header.

Lennonka (changes requested):

apinnick:
@jherrman Can you move this to the deployment section of the Planning guide?

guides/common/assembly_foreman-overview-and-concepts.adoc
Lennonka, line 1: This isn't the "deployment section".

[This is the deployment section.](https://theforeman-foreman-documentation-preview-pr-3799.surge.sh/nightly/Planning_for_Project/index-satellite.html#satellite-deployment-planning)

Lennonka (changes requested):

guides/common/modules/con_security-features.adoc
```diff
@@ -0,0 +1,37 @@
+[id="security-features_{context}"]
+= Security features in {Project}
```
Lennonka, line 2:
```suggestion
= Security hardening in {Project}
```
This seems to be the standard term used in RHEL docs.

Lennonka (changes requested):

guides/common/modules/con_security-features.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+File Access Policy Daemon (`fapolicyd`)::
+By enabling `fapolicyd` on your {ProjectServer}, you can monitor and control file and directory access, providing an additional security layer.
++
+This feature helps prevent unauthorized code execution and enhances system integrity. You can enable or disable `fapolicyd` on both {Project} and {SmartProxyServer} at any time, depending on your security requirements.
++
+For additional information about `fapolicyd`, see {RHELDocsBaseURL}9/html/security_hardening/assembly_blocking-and-allowing-applications-using-fapolicyd_security-hardening[Blocking and allowing applications by using fapolicyd] in _Security hardening_.
++
+For instructions on enabling `fapolicyd`, see {InstallingServerDocURL}installing_satellite_server_in_a_connected_network_environment/installing_server_connected_{project-context}#using-fapolicyd-on-server_{project-context}[Using fapolicyd on Satellite Server
+] in _{InstallingServerDocTitle}_.
```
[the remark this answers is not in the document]
jherrman: Good catch, will fix.

guides/common/modules/con_security-features.adoc
jherrman: It is an established term indeed, but I don't think it would be a good heading here. The module is a conceptual/referential list of the available security features, rather than a specific guide how to set them up them. Also, using a gerund heading for a non-procedural module should be avoided according to our style guides.

guides/common/assembly_foreman-overview-and-concepts.adoc
jherrman: The "overview and concepts" assembly seems to me a somewhat better fit content-wise, but you're right that it is fairly overloaded. I'll move the module over, then.

guides/common/modules/con_security-features.adoc
ekohl: I'd agree it's not a good heading. Many of these are more compliance than hardening. For example, FIPS only has a certainly list of allowed ciphers. Newer ciphers (which are arguably better) may not be on the list until a new standard is allowed. Effectively FIPS can make a system less secure, but compliant. An example we had in the past: FIPS didn't allow the use of MD5, but RHEL 6 only had MD5 checksums for RPMs. The solution: disable checksums altogether.

IMHO nobody should use FIPS unless you're mandated to.

guides/common/modules/con_security-features.adoc
```diff
@@ -0,0 +1,42 @@
+[id="security-features_{context}"]
+= Security features in {Project}
+
+When deploying {Project}, consider implementing security features that enhance the overall security of your system.
+{Project} supports multiple security mechanisms, including FIPS mode, `fapolicyd`, and SELinux, to provide additional layers of protection.
+
+Federal Information Processing Standards (FIPS) Mode::
+{Project} supports running on FIPS-enabled {RHEL} hosts to comply with security standards for cryptographic modules. 
+However, {Project} itself does not currently support running in FIPS mode.
```
ekohl, line 9: I think we usually say `{Project}` itself is not FIPS certified.
```suggestion
However, {Project} itself is not FIPS certified.
```

guides/common/modules/con_security-features.adoc
```diff
@@ -0,0 +1,42 @@
...
+For more information, see {RHELDocsBaseURL}9/html/security_hardening/switching-rhel-to-fips-mode_security-hardening[Switching RHEL to FIPS mode] in _{RHEL}{nbsp}9 Security hardening_.
+
+File Access Policy Daemon (`fapolicyd`)::
+By enabling `fapolicyd` on your {ProjectServer}, you can monitor and control file and directory access, providing an additional security layer.
++
+This feature helps prevent unauthorized code execution and enhances system integrity. You can enable or disable `fapolicyd` on both {Project} and {SmartProxyServer} at any time, depending on your security requirements.
++
+For additional information about `fapolicyd`, see {RHELDocsBaseURL}9/html/security_hardening/assembly_blocking-and-allowing-applications-using-fapolicyd_security-hardening[Blocking and allowing applications by using fapolicyd] in _Security hardening_.
++
+For instructions on enabling `fapolicyd`, see {InstallingServerDocURL}installing_satellite_server_in_a_connected_network_environment/installing_server_connected_{project-context}#using-fapolicyd-on-server_{project-context}[Using fapolicyd on Satellite Server] in _{InstallingServerDocTitle}_.
```
ekohl, line 23: I'm not sure we need to link there. The original goal was to drop the chapter in the installation guide, though https://issues.redhat.com/browse/SAT-30719 doesn't cover that.

For a user it's sufficient to run `dnf install fapolicyd`, either before or after installation. https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/security_hardening/assembly_blocking-and-allowing-applications-using-fapolicyd_security-hardening (where you already link to in the paragraph above) does exactly that. In other words: there's nothing in our procedures that adds value. And the group of users who use `fapolicyd` will already know what `fapolicyd` is. Those who don't know it, don't need it.

The conclusion was that we should mention in the planning guide that we support fapolicyd (and will ensure the right policy is installed). Then drop `guides/common/modules/proc_installing-fapolicyd-on-server.adoc` and all references to it.

IMHO that should be part of this PR.

guides/common/modules/con_security-features.adoc
```diff
@@ -0,0 +1,42 @@
+[id="security-features_{context}"]
+= Security features in {Project}
+
+When deploying {Project}, consider implementing security features that enhance the overall security of your system.
+{Project} supports multiple security mechanisms, including FIPS mode, `fapolicyd`, and SELinux, to provide additional layers of protection.
+
+Federal Information Processing Standards (FIPS) Mode::
```
ekohl, line 7: I'd like a note that you can't enable FIPS on an already installed `{Project}`. Today we have that in our installation guides, but it's better in the planning guide. FIPS mode is neither a prerequisite nor is enabling FIPS part of the installation procedure. Once you have that, you can drop the FIPS mode parts from our installation guides.

guides/common/modules/con_security-features.adoc
```diff
@@ -0,0 +1,42 @@
...
+For instructions on enabling `fapolicyd`, see {InstallingServerDocURL}installing_satellite_server_in_a_connected_network_environment/installing_server_connected_{project-context}#using-fapolicyd-on-server_{project-context}[Using fapolicyd on Satellite Server] in _{InstallingServerDocTitle}_.
+
+Security-Enhanced Linux (SELinux)::
+{Project} supports SELinux, a mandatory access control system that restricts system access and reduces the risk of security vulnerabilities.
+SELinux enforces policies that define which processes can access specific system resources, providing protection against unauthorized access and exploitation.
++
+Running {Project} with SELinux in enforcing mode is recommended to maintain a high-security environment.
+It ensures that access control policies are strictly enforced, minimizing the risk of privilege escalation or unauthorized actions.
++
+Using SELinux is required for {Project} Server to be installed on your environment. For additional information, see {InstallingServerDocURL}preparing_your_environment_for_installation_{project-context}#system-requirements_{project-context}[System requirements] in _{InstallingServerDocTitle}_.
```
ekohl, line 32: Like SELinux and fapolicyd, the idea was to move it from the installation guide to the planning guide.

Also, technically it's not required but for Satellite it must be enabled to consider the system supported. It may be in permissive mode, though enforcing mode is recommended. For upstream I'd follow those recommendations.

Here the argument to move it isn't as strong because you can consider it a prerequisite.

guides/common/modules/con_security-features.adoc
jherrman: I see, I'll work on including that in the next draft, then.

guides/common/modules/ref_system-requirements.adoc
```diff
@@ -93,6 +93,7 @@ ifdef::foreman-el,katello,satellite[]
 .SELinux mode
 SELinux must be enabled, either in enforcing or permissive mode.
 Installation with disabled SELinux is not supported.
+For more information, see {PlaningDocURL}security-features_planning[Security features] in _{PlanningDocTitle}_.
```
evgeni, line 96:
```suggestion
For more information, see {PlanningDocURL}security-features_planning[Security features] in _{PlanningDocTitle}_.
```

guides/common/modules/con_security-features.adoc
```diff
@@ -0,0 +1,52 @@
+[id="security-features"]
+= Security features in {Project}
+
+When deploying {Project}, consider implementing security features that enhance the overall security of your system.
+{Project} supports multiple security mechanisms to provide additional layers of protection, which you can enable during or after the installation of {Project}, if your use case requires them.
```
Lennonka, line 5: Should we mention that these security considerations are applicable to both ProjectServer and SmartProxyServers?

guides/common/assembly_foreman-deployment-planning.adoc
```diff
@@ -5,3 +5,7 @@ include::assembly_deployment-path.adoc[leveloffset=+1]
 include::assembly_common-deployment-scenarios.adoc[leveloffset=+1]
 
 include::assembly_provisioning-requirements.adoc[leveloffset=+1]
+
+ifndef::foreman-deb[]
```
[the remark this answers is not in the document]
jherrman: Good point, fapolicy is already "ifndef'd" in the module, so it should hopefully be safe to include the module for Debian and Ubuntu as well.

guides/common/modules/con_security-features.adoc
```diff
@@ -0,0 +1,52 @@
...
+====
+{Project} supports DEFAULT and FIPS crypto-policies.
+The FUTURE crypto-policy is not supported for {Project} and {SmartProxy} installations.
+The FUTURE policy is a stricter forward-looking security level intended for testing a possible future policy.
+For more information, see {RHELDocsBaseURL}9/html/security_hardening/using-the-system-wide-cryptographic-policies_security-hardening[Using system-wide cryptographic policies] in _{RHEL}{nbsp}9 Security hardening_.
+====
+
+// we do not package "fapolicy" for Foreman on Debian/Ubuntu
+ifndef::foreman-deb[]
+File Access Policy Daemon (`fapolicyd`)::
```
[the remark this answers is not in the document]
Lennonka: Oh, I see. Have I mistaken it with SELinux perhaps?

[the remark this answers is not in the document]
jherrman: Yep, that's already in :-)

guides/common/modules/con_security-features.adoc
ekohl: I think that reads well to me.

guides/common/assembly_foreman-deployment-planning.adoc
```diff
@@ -5,3 +5,5 @@ include::assembly_deployment-path.adoc[leveloffset=+1]
 include::assembly_common-deployment-scenarios.adoc[leveloffset=+1]
 
 include::assembly_provisioning-requirements.adoc[leveloffset=+1]
+
+include::modules/con_security-considerations.adoc[leveloffset=+1]
```
ekohl, line 9: I think none of the security features we mention here are supported on Debian: FIPS, fapolicyd and SELinux are all EL only. At least for Foreman.

Lennonka: @jherrman Sorry for confusing you, Jirka. Ewoud is more knowledgeable than me in this area. ^^

Lennonka (approved):
LGTM

guides/common/modules/con_security-features.adoc
Lennonka: @jherrman Actually, can you address this? I missed it during review.

Lennonka (changes requested):
Please, address my last comment above.

guides/common/modules/con_security-features.adoc
jherrman: @Lennonka , already did: https://github.com/theforeman/foreman-documentation/pull/3799/files#diff-5a886518aa8c58822600953b3d423d55082b31748c85b9f4810099e61c3de04aR9

Is this what you had in mind?

Cheers,
J.

Lennonka: Oh, yes, that. Thanks!

Lennonka (approved):

jherrman:
@Lennonka @ekohl @maximiliankolb , @AkshayGadhaveRH , as far as I can tell, this should be ready to merge.

Unless any further tweaks to the updates are required, could you help merge this, before any more conflicts rear their ugly heads? :-)

Thanks!

PR #3841, opened 2025-05-06
Remove instructions to place GCE key in /etc/foreman

ekohl:

Remove the instructions to place the key in `/etc/foreman`.

Hammer uploads the key to Foreman (just like in the UI procedure) and never read again. This means there's no point in storing the file in /etc/foreman with specific permissions.

It was introduced in 40b118063bf980dbd92469793abfc1229b691d99 (https://github.com/theforeman/foreman-documentation/pull/1949).

I changed the Hammer command to run as root for two reasons: the scp command also uses root and the installer sets up Hammer for root by default.  1137d3a971688a8bfef350962fdb1bc9afc17cae changed it to non-root and this goes against that.

An alternative is to change the instruction to a prerequisite to have Hammer set up and the key file present. Then you only need to list the Hammer command.

#### Checklists

### DIFF
--- a/guides/common/modules/proc_adding-a-google-gce-connection.adoc
+++ b/guides/common/modules/proc_adding-a-google-gce-connection.adoc
@@ -22,38 +22,12 @@ For more information, see https://cloud.google.com/iam/docs/creating-managing-se
 .CLI procedure
 . In Google GCE, generate a service account key in JSON format.
 For more information, see https://cloud.google.com/iam/docs/creating-managing-service-account-keys[Create and manage service account keys] in the GCE documentation.
-. Copy the file from your local machine to {ProjectServer}:
-+
-[options="nowrap", subs="+quotes,verbatim,attributes"]
-----
-# scp _My_GCE_Key_.json root@{foreman-example-com}:/etc/foreman/_My_GCE_Key_.json
-----
-. On {ProjectServer}, change the owner for your service account key to the `foreman` user:
-+
-[options="nowrap", subs="+quotes,verbatim,attributes"]
-----
-# chown root:foreman /etc/foreman/_My_GCE_Key_.json
-----
-. On {ProjectServer}, configure permissions for your service account key to ensure that the file is readable:
-+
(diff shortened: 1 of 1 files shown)


ekohl:
@maximiliankolb did you have a look at my thoughts on whether we need the `scp` command at all and the run hammer as root or non-root option?

maximiliankolb:
> @maximiliankolb did you have a look at my thoughts on whether we need the `scp` command at all and the run hammer as root or non-root option?

I had a glance and did not really have any input on it.

On second though: Maybe we should make it more explicit by prefixing the file name with `/root/`? This would indicate that a) the key is in the home dir of the root user and therefore not readable by others, and b) that you need to run the Hammer CLI command as root.

ekohl:
Should we drop the `scp` instruction and make it a prerequisite? Perhaps also add a recommendation to remove the file after uploading it?

maximiliankolb:
Maybe a prerequisite to create the JSON file on *.google.com and then in the procedure to either upload it via browser or scp it to Foreman Server?

ekohl:
Perhaps maybe this is good enough now and we can iterate on it when we feel the need to?

maximiliankolb (approved):

maximiliankolb:
@ekohl This suggestion is still open: https://github.com/theforeman/foreman-documentation/pull/3841#issuecomment-2857319044 But I can also look into this after merging this PR.

Lennonka (approved):
I'm wondering if we could suggest an alternative path to the user that would promote a good practice where to put the key. Consider this optional for this PR.

Lennonka:
And I think it might be a good idea for some to test the new procedure.

maximiliankolb:
triage: kindly asking for a test from team rocket. cc @Lennonka

Lennonka:
In that case, the location of the key is irrelevant and this probably doesn't need testing.

Lennonka:
@ekohl Friendly reminder to add the step to remove the key

PR #3855, opened 2025-05-14
Improve pre-upgrade documentation

aneta-petrova:

This PR reviews the parts of the Upgrade guide that describe what happens before users start upgrading. I'm splitting this content into "Planning" and "Procedure prerequisites".

I'm proposing one new assembly:

* "Preparing for Project upgrade": This groups all information relevant before users roll up their sleeves to start the actual upgrade process.

And then three different prerequisite-like sections:

* "Planning Project upgrade": This is part of "Preparing for Project upgrade". This section lists general considerations + actions to be taken when preparing for the upgrade process as a whole.
* "Prerequisites" in the procedure "Upgrading a ProjectServer": This part lists very specific actions to be taken right before following the server upgrade procedure.
* "Prerequisites" in the procedure "Upgrading a SmartProxy": This part lists very specific actions to be taken right before following the proxy upgrade procedure.

Suboptimal upgrade experience. I shared a more detailed explanation and justification in https://issues.redhat.com/browse/SAT-29153

Before, the pre-upgrade information was included in various multiple sections throughout the guide and there was quite a lot of duplication. Reviewing the pre-upgrade information and grouping it into fewer sections will guide users through the pre-upgrade process more smoothly.

The diff shows a lot of deletions but the only ~~two~~ items I'm actually removing are ~~one obsolete prerequisite and another irrelevant prereqs~~ items that are obsolete now. ~~Both~~ All removals are highlighted in ~~a separate commit~~ separate commits for easier review. The rest of the removed lines is just me removing duplicate content (there was a lot of it).

#### Checklists


guides/common/modules/con_available-upgrade-paths.adoc
```diff
@@ -0,0 +1,44 @@
...
+endif::[]
+ifdef::foreman-deb[]
+. Upgrading your {ProjectServer} to {ProjectVersion}.
+. Upgrading your {SmartProxyServers} to {ProjectVersion}.
+endif::[]
+ifdef::orcharhino[]
+Because {SmartProxies} at version {ProjectVersionPrevious} will keep working with your upgraded {ProjectServer} {ProjectVersion}, you can upgrade your {SmartProxies} separately over multiple maintenance windows.
+endif::[]
+ifndef::orcharhino[]
+You can upgrade your {SmartProxies} separately over multiple maintenance windows because {SmartProxies} at version {ProjectVersionPrevious} and {ProjectVersionPrevious-Previous} will keep working with your upgraded {ProjectServer} {ProjectVersion}.
```
ekohl, line 26: Aren't the benefits also explained in the bullet points below? Perhaps leave them out here and only list them in the bullet points?

ekohl (commented):
I like the shortening of the guide. It was way too long. Possibly still is, but I also know this is the first step.

I'm looking at @evgeni and @ehelms here, but also don't want to blow up the scope too much @aneta-petrova.

I'd suggest to make the Katello and Satellite upgrade procedures the same. We should also make sure the Smart Proxy installation guide always tells users to sync content and install that way. Benefits of alignment are hopefully clear.

I'm not sure where Orcharino stands here (:wave: @maximiliankolb), but ideally it would be as similar as possible.

Then on `foreman-el` and `foreman-deb` it will also be the same. The concrete instructions will differ only where we set up repositories. We have started to look more at `foreman-maintain` to do that work so we could even further align there, but I suspect that's out of scope for now.

aneta-petrova:
> I'm looking at @evgeni and @ehelms here, but also don't want to blow up the scope too much @aneta-petrova.

The scope for now is only "things users need to know/do before they upgrade".

> I'd suggest to make the Katello and Satellite upgrade procedures the same.

I'll note your suggestion to align the actual upgrade procedures in the downstream ticket I have (https://issues.redhat.com/browse/SAT-34002). That will help me remember when I get to the actual upgrade part.

> We should also make sure the Smart Proxy installation guide always tells users to sync content and install that way.

Just for the sake of accountability: I'm looking at upgrades now, not installation. So I don't expect to take any action based on this. (also to avoid blowing up the scope too much and all :))

ekohl:
> Just for the sake of accountability: I'm looking at upgrades now, not installation. So I don't expect to take any action based on this. (also to avoid blowing up the scope too much and all :))

But I think it is related. You will confuse users if we tell them that you upgrade by syncing content but never instructed them to set it up in the first place. If so, we need to think about a migration to get them into the newly preferred path. Or not do it at all.

guides/common/modules/con_planning-project-upgrade.adoc
```diff
@@ -0,0 +1,38 @@
...
+= Planning {Project} upgrade
+
+Upgrading to {Project} {ProjectVersion} affects your entire {Project} infrastructure.
+Plan carefully before proceeding.
+
+* Read the {ProjectName} {ProjectVersion} {ReleaseNotesDocURL}[Release Notes].
+* Ensure that you have sufficient storage space.
+For more information, see {InstallingServerDocURL}Preparing_your_Environment_for_Installation_{project-context}[Preparing your Environment for Installation] in _{InstallingServerDocTitle}_ and {InstallingSmartProxyDocURL}preparing-environment-for-capsule-installation[Preparing your Environment for Installation] in _Installing {SmartProxyServer}_.
+* Ensure that your firewall configuration is up-to-date.
+For more information, see {InstallingServerDocURL}Port_and_firewall_requirements_{project-context}[Port and firewall requirements] in _{InstallingServerDocTitle}_ and {InstallingServerDisconnectedDocURL}Port_and_firewall_requirements_{project-context}[Port and firewall requirements] in _{InstallingServerDisconnectedDocTitle}_.
```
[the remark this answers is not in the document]
aneta-petrova: Agreed. But just to be sure @ekohl, I can drop both the bullet points on storage space and firewall config, right? That would be lines 8-11.

My reasoning is that we're talking about upgrades, and users should already have sufficient storage space + appropriate firewall setup from the time they did the initial installation. And if these ever change from the previous version, it should be mentioned in the release notes and preferably also as a specific, actionable step in the upgrade procedure.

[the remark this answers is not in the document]
ekohl: Exactly that on both paragraphs.

guides/common/modules/con_available-upgrade-paths.adoc
```diff
@@ -0,0 +1,44 @@
...
+= Available upgrade paths
+
+Upgrading to {ProjectName} {ProjectVersion} from {ProjectName} {ProjectVersionPrevious} includes the following high-level steps:
+
+ifdef::satellite[]
+. Upgrading your {ProjectServer}s and {SmartProxyServers} to {Project} {ProjectVersionPrevious}.
+endif::[]
+ifndef::foreman-deb[]
+. Upgrading your {ProjectServer}:
+.. Upgrading your {ProjectServer} to {ProjectVersion}.
```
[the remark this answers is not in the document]
aneta-petrova: I looked at simplifying the high-level summary. Based on the previous version of the module and your suggestions, it boils down to one (shorter) summary for Deb+EL+orcharhino and one (longer) summary for Katello+Satellite. @maximiliankolb please take a look here to make sure the simplified list accurately describes the process for orcharhino.

guides/common/modules/con_available-upgrade-paths.adoc
aneta-petrova: Addressed.

guides/common/modules/con_planning-project-upgrade.adoc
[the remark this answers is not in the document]
aneta-petrova: Dropped.

guides/common/modules/con_planning-project-upgrade.adoc
```diff
@@ -0,0 +1,38 @@
...
+After you successfully test the upgrade on the clone, you can repeat the upgrade on your primary {ProjectServer} and discard the clone, or you can promote the clone to your primary {ProjectServer} and discard the previous primary {ProjectServer}.
+For more information, see {AdministeringDocURL}cloning_satellite_server[Cloning {ProjectServer}] in _{AdministeringDocTitle}_.
+endif::[]
+ifndef::foreman-deb[]
+* If your {ProjectServer} runs on a RHEL system that is not operating in FIPS mode, you cannot upgrade it to a RHEL system that is operating in FIPS mode.
+In this situation, you must install {Project} on a freshly provisioned RHEL system in FIPS mode.
+endif::[]
+ifdef::satellite[]
+ifeval::["{mode}" == "connected"]
+* If you are using a self-registered {Project}, you cannot upgrade it.
```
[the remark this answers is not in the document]
aneta-petrova: Dropped.

guides/common/modules/con_planning-project-upgrade.adoc
```diff
@@ -0,0 +1,38 @@
...
+In this situation, you must migrate your {Project} to the Red{nbsp}Hat Content Delivery Network (CDN).
+endif::[]
+endif::[]
+
+{Project} services are shut down during the upgrade.
+Ensure to plan for the required downtime.
+The upgrade process duration might vary depending on your hardware configuration, network speed, and the amount of data that is stored on the server.
+
+* Upgrading {ProjectServer} takes approximately 1{range}2 hours.
```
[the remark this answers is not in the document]
aneta-petrova: It sounds like your "30 minutes" is more of a guess :) so I hesitate to update this. Perhaps @evgeni or @lpramuk could chime in with their estimates as well? How do you think we should update lines 37 and 38?

guides/common/modules/snip_prerequisites-upgrading-project-server.adoc
```diff
@@ -1,42 +1,21 @@
...
-* Review xref:upgrading_prerequisites_{context}[].
-* Note that you can upgrade {SmartProxies} separately from {Project}.
-For more information, see xref:Upgrading_Proxies_Separately_from_Server_{context}[].
-* Review and update your firewall configuration.
-For more information, see {InstallingServerDocURL}Preparing_your_Environment_for_Installation_{project-context}[Preparing your environment for installation] in _{InstallingServerDocTitle}_.
-ifdef::katello,orcharhino,satellite[]
-* Ensure that you do not delete the manifest from the Customer Portal or in the {ProjectWebUI} because this removes all the entitlements of your content hosts.
+.Prerequisites
+ifdef::satellite[]
+* Ensure that all {ProjectServer}s are on the same version.
```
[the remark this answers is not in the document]
aneta-petrova: @evgeni What do you think? Is what Ewoud describes the correct strategy for disconnected installs?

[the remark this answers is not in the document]
evgeni: I think the first part is correct, this is only needed if you have more than one server that need to share content via an export.
Now, the tricky part comes that there are different ways to share content, and only the export/import one really needs this lockstep (but I am not sure whether other matrix entries are tested properly).

The hosed part: it depends, as your second server will usually be a consumer (either of the first, or an ISO), you should be able to upgrade it without problems, as the incompatible content export is not used here.

guides/common/modules/con_planning-project-upgrade.adoc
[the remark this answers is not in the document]
ekohl: If we look at https://ci.theforeman.org/view/Foreman%20Nightly/job/foreman-nightly-rpm-pipeline/ then we see the entire test pipeline takes 35 minutes. That includes an upgrade from n-2 to n-1 and then another upgrade from n-1 to n. Granted, that's on an empty installation. With data it could be slower, but I think 1 to 2 hours is unlikely.

If we look at https://ci.theforeman.org/view/Foreman%20Nightly/job/katello-nightly-rpm-pipeline/ then we see the whole upgrade pipeline takes about 1:30. Again, this includes n-2 to n-1 and n-1 to n. This pipeline also includes installation of a proxy and upgrade that twice too.

The precise timing for a job tells me that with (a mostly empty) Katello install a project server upgrade step (n-2 to n-1 or n-1 to n) takes about 6 minutes while the proxy server step is about 4 minutes. It's really the test suite that takes the most time (50 minutes), followed by the server installation (10 minutes).

I think the 1 to 2 hours estimate is really dramatic, but may be true for slow hardware on large installations. It's really a worst case, not an average. Feedback @ehelms received from customers also pointed that steps around it (like backups) usually take more time than the upgrade itself.

Taking a step back, do we need to give an estimate? I understand why people like it, but if it's so inaccurate, does it really help users? I fear it may scare them and postpone upgrades.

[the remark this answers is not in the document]
evgeni: IMHO there are way too many factors to give a useful estimate.

An empty Foreman upgrades in 3-4 minutes.
An empty Katello in 5-6 minutes.
An empty Luna in 8-10 minutes.
(All assuming there are no other OS-level upgrades that need to happen and reasonable hardware underneath).

I have an production Satellite that upgrades in 30-ish minutes.

But it's totally possible that an upgrade with more data will take longer.

guides/common/modules/con_available-upgrade-paths.adoc
```diff
@@ -0,0 +1,39 @@
+[id="available-upgrade-paths"]
+= Available upgrade paths
```
evgeni, line 2: Are those really different paths?

To me it feels like it's always the same path:
1. ensure you run version X
2. upgrade server to X+1
3. upgrade the proxies to X+1

Sure, you can walk the last step at a different pace, but it's still the same path?

Compare to el8-to-el9 upgrade, where you can pick "in-place", "backup/restore", (and "rebuild") that use totally different technology, have different pro/contras yet ultimately get you to the same place ("system running on el9")

Having the overview here is good, the naming just somehow implies the user has a choice, while it's really limited.

guides/common/modules/con_planning-project-upgrade.adoc
```diff
@@ -0,0 +1,28 @@
...
+endif::[]
+ifdef::satellite[]
+* Optional: You can test the upgrade on a clone of your {ProjectServer}.
+After you successfully test the upgrade on the clone, you can repeat the upgrade on your primary {ProjectServer} and discard the clone, or you can promote the clone to your primary {ProjectServer} and discard the previous primary {ProjectServer}.
+For more information, see {AdministeringDocURL}cloning_satellite_server[Cloning {ProjectServer}] in _{AdministeringDocTitle}_.
+endif::[]
+ifndef::foreman-deb[]
+* If your {ProjectServer} runs on a RHEL system that is not operating in FIPS mode, you cannot upgrade it to a RHEL system that is operating in FIPS mode.
+In this situation, you must install {Project} on a freshly provisioned RHEL system in FIPS mode.
+endif::[]
```
evgeni, line 21: while technically true, I always thought it doesn't belong into the upgrade docs. the user also can't enable fips after the fact during normal operations.

I recall we wanted to have a "security considerations" guide, that's where it belongs.

guides/common/modules/con_available-upgrade-paths.adoc
aneta-petrova: Renamed to "Upgrade path overview"

guides/common/modules/con_planning-project-upgrade.adoc
aneta-petrova: At first I thought this was the result of a customer case request. But actually it was part of the initial import: https://github.com/theforeman/foreman-documentation/commit/84e812e1326 so it's very old.
And I do think that we have this covered in the installation guide: https://github.com/theforeman/foreman-documentation/blob/028b74684e59bcb042835d20da4713e095cd67ed/guides/common/modules/ref_system-requirements.adoc?plain=1#L107-L115, which is being reviewed by Jirka in https://github.com/theforeman/foreman-documentation/pull/3799 in an effort to turn it into a "security considerations" section somewhere in the docs set.

Overall, I think we can remove this.

guides/common/modules/snip_prerequisites-upgrading-project-server.adoc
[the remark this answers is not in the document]
aneta-petrova: I updated the prereq to only show up for the disconnected scenario and to limit it to sync by using exports.

I'm adding anything about which server to update first because based on Evgeni's comment, the order doesn't matter that much.

guides/common/modules/con_planning-project-upgrade.adoc
evgeni: Some more archeology with zombies later: this was added as part of https://bugzilla.redhat.com/show_bug.cgi?id=1671445 (the bug itself doesn't contain much data, sorry) as in Satellite 6.5 FIPS was a new feature, but users could not go from a non-FIPS 6.4 to a FIPS 6.5, and that's what this was warning against.

My calendar says this is safe to remove from the upgrade docs indeed.

guides/common/modules/con_planning-project-upgrade.adoc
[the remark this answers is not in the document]
ekohl: I agree with that, but perhaps we can discuss it with RH's customer support on what they think?

@maximiliankolb I'd be interested in your opinion too.

guides/common/modules/con_planning-project-upgrade.adoc
```diff
@@ -0,0 +1,18 @@
...
+Upgrading to {Project} {ProjectVersion} affects your entire {Project} infrastructure because {Project} services are shut down during the upgrade.
+The upgrade process duration varies depending on your hardware configuration, network speed, and the amount of data that is stored on the server.
+Plan carefully before proceeding.
+
+* Read the {ProjectName} {ProjectVersion} {ReleaseNotesDocURL}[Release Notes].
+* Consider whether any of your scripts need updating.
+Some {Project} API endpoints, Hammer CLI commands, and modules from the {Project} Ansible Collection can differ between versions of {Project}.
+ifdef::satellite[]
+For information about changes in these tools, see the {ProjectName} {ProjectVersion} {ReleaseNotesDocURL}[Release Notes].
+endif::[]
```
ekohl, line 13: On line 8 we also point to the release notes and this feels like a repetition. I wonder if there's a way to combine them.

aneta-petrova: I wondered myself and couldn't figure out a way. The first reference is generic. This second one talks about a specific piece of information that's included in the release notes, and only downstream. That's why I kept them separate. If anyone has a better suggestion, I'll consider it. For now, no action.

ekohl: Perhaps move away from bullet points into a paragraph?

Read the {ProjectName} {ProjectVersion} {ReleaseNotesDocURL}[Release Notes].
This helps you plan your upgrade by pointing out noteworthy changes.
For example, {Project} API endpoints, Hammer CLI commands, and modules from the {Project} Ansible Collection can differ between versions of {Project}.
Any custom integrations, such as scripts may need to be updated.

The above IMHO doesn't need any conditionals and explains how to view the release notes. I'm sure you can improve wording.

guides/common/modules/con_upgrade-path-overview.adoc
```diff
@@ -0,0 +1,39 @@
...
+ifdef::katello,satellite[]
+. Ensuring that your {ProjectServer}s and {SmartProxyServers} are running on {Project} {ProjectVersionPrevious}.
+. Upgrading your {ProjectServer}:
+.. Upgrading your {ProjectServer} to {ProjectVersion}.
+.. Synchronizing the new {ProjectVersion} repositories.
+. Upgrading your {SmartProxyServers} to {ProjectVersion}.
+endif::[]
+
+ifndef::orcharhino[]
+You can upgrade your {SmartProxies} separately over multiple maintenance windows because {SmartProxies} at version {ProjectVersionPrevious} and {ProjectVersionPrevious-Previous} will keep working with your upgraded {ProjectServer} {ProjectVersion}.
```
bangelic, line 19:
```suggestion
You can upgrade your {SmartProxies} separately over multiple maintenance windows because versions {ProjectVersionPrevious} and {ProjectVersionPrevious-Previous} remain compatible with your upgraded {ProjectServer} {ProjectVersion}.
```

guides/common/modules/con_upgrade-path-overview.adoc
```diff
@@ -0,0 +1,39 @@
...
+endif::[]
+ifdef::orcharhino[]
+You can upgrade your {SmartProxies} separately over multiple maintenance windows because {SmartProxies} at version {ProjectVersionPrevious} will keep working with your upgraded {ProjectServer} {ProjectVersion}.
```
bangelic, line 22:
```suggestion
You can upgrade your {SmartProxies} separately over multiple maintenance windows because {SmartProxies} running version {ProjectVersionPrevious} remain compatible with {ProjectServer} {ProjectVersion}.
```

guides/common/modules/con_upgrade-path-overview.adoc
```diff
@@ -0,0 +1,39 @@
...
+This allows you to upgrade all {SmartProxies} one after another without any outage.
+endif::[]
+
+ifdef::orcharhino[]
+{SmartProxies} at version {ProjectVersionPrevious} retain all of their previous functionality.
+endif::[]
+ifndef::orcharhino[]
+{SmartProxies} at version {ProjectVersionPrevious} and {ProjectVersionPrevious-Previous} retain all of their previous functionality.
+endif::[]
+New functionality added in the {ProjectVersion} release will not work until you upgrade {SmartProxies} to {ProjectVersion}.
```
bangelic, line 39:
```suggestion
New functionality added in the {ProjectVersion} release is available only after you upgrade your {SmartProxies} to {ProjectVersion}.
```

guides/common/modules/proc_upgrading-smartproxy-server.adoc
```diff
@@ -1,33 +1,22 @@
...
-* Ensure that you synchronize the required repositories on {ProjectServer}.
-For more information, see xref:synchronizing_the_new_repositories_{context}[].
+After you have upgraded your {ProjectServer} and synchronized the required repositories, you can start upgrading your {SmartProxyServers}.
 endif::[]
+You can upgrade from any minor version of {SmartProxyServer} {ProjectVersionPrevious} to {ProjectVersion}.
+
+.Prerequisites
 ifdef::katello,orcharhino,satellite[]
-* If you use content views to control updates to the base operating system of {SmartProxyServer}, update those content views with new repositories, publish, and promote their updated versions.
+* If you use content views to control updates to the base operating system of {SmartProxyServer}, or for {SmartProxyServer} repository, update those content views with new repositories, publish, and promote their updated versions.
```
bangelic, line 11:
```suggestion
* If you use content views to control updates to the {SmartProxyServer} base operating system or repository, update those content views with new repositories, then publish and promote their updated versions.
```

bangelic (changes requested):

guides/common/modules/con_planning-project-upgrade.adoc
aneta-petrova: I don't think dropping the conditional works because upstream notes actually don't list the changes, we only publish the changelists downstream (https://content-preview.docs.redhat.com/en/documentation/red_hat_satellite/6.17/html-single/release_notes/index#changes-in-satellite-installer-parameters-hammer-cli-api-and-ansible_satellite). That reference is something I can only include for satellite.

Also, I like how both actions have their own bullet point: 1. Read the notes. 2. Think about your integrations. Both are actions users should carefully consider so I believe both deserve their own emphasis.

I agree that including the RN link twice is not the most elegant solution but it's functional. And it's also only for downstream satellite so any potential impact is limited.

guides/common/modules/con_planning-project-upgrade.adoc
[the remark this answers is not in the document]
aneta-petrova: > if it's so inaccurate, does it really help users

> there are way too many factors to give a useful estimate.

The usefulness of the estimate is not in giving users accurate data. Rather than tell them how long it will take, I think of it as telling them how long it will *not* take. In other words, by giving users an estimate of 1-2 hours, we're also telling them that it won't be 1-2 days.

I asked questions downstream and got confirmation that the documented estimate is used in conversations with customers.

If there is consensus that 1-2 hours is too much, we can update the estimates to make them sound less scary. What I tried with the latest update is to differentiate between "average installations" and "very large installations". Can you please take a look? Would something like this work?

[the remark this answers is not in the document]
evgeni: It probably opens the question "what is average" and "what is large", but yeah, I think that should work.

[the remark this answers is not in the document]
aneta-petrova: > It probably opens the question "what is average" and "what is large", but yeah, I think that should work.

True, but hopefully it at least shows that 1-2 hours is not a typical scenario, which is what I suspect @ekohl's objection was largely about.

I'm resolving this thread but if there are further suggestions or concerns, please reopen.

aneta-petrova:
> But I think it is related. You will confuse users if we tell them that you upgrade by syncing content but never instructed them to set it up in the first place. If so, we need to think about a migration to get them into the newly preferred path. Or not do it at all.

I opened https://github.com/theforeman/foreman-documentation/pull/3856 to address this on the Proxy Installation guide's side.

aneta-petrova:
Hi @ekohl and @evgeni, do you have any further comments from the technical point of view? If not, do you feel like you can give this an ack?

Hi @bangelic, a similar question for you :) Do you have any further comments from the style point of view?

In the meantime, I also asked @maximiliankolb to take a look to make sure I don't break anything for orcharhino.

guides/common/modules/con_planning-project-upgrade.adoc
```diff
@@ -0,0 +1,24 @@
+[id="planning-{project-context}-upgrade"]
+= Planning {Project} upgrade
+
+Upgrading to {Project} {ProjectVersion} affects your entire {Project} infrastructure.
+Plan carefully before proceeding.
+
+* Read the {ProjectName} {ProjectVersion} {ReleaseNotesDocURL}[Release Notes].
+* Consider whether any of your scripts need updating.
```
evgeni, line 8:
```suggestion
* Consider whether any of your integrations need updating.
```

IMHO this better encapsulates the fact that it can be bash scripts, ansible playbooks or some other api-based stuff.

evgeni (approved):

maximiliankolb (commented):
@aneta-petrova  I left some comments. Overall LGTM

About minor vs. patch version: I would like to get some more input from either Evgeni or Ewoud if possible.

guides/common/modules/con_planning-project-upgrade.adoc
```diff
@@ -0,0 +1,24 @@
+[id="planning-{project-context}-upgrade"]
+= Planning {Project} upgrade
+
+Upgrading to {Project} {ProjectVersion} affects your entire {Project} infrastructure.
+Plan carefully before proceeding.
+
+* Read the {ProjectName} {ProjectVersion} {ReleaseNotesDocURL}[Release Notes].
```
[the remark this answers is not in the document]
aneta-petrova: Applied.

guides/common/modules/con_upgrading-project.adoc
```diff
@@ -1,4 +1,4 @@
 [id="Upgrading_{project-context}_{context}"]
 = Upgrading {ProjectName}
 
-Use the following procedures to upgrade your existing {ProjectName} to {ProjectName} {ProjectVersion}.
+You can upgrade from any patch version of {ProjectServer} {ProjectVersionPrevious} to {ProjectServer} {ProjectVersion}.
```
aneta-petrova, line 4: @evgeni Maximilian pointed out this line in his review (and also line 7 in `guides/common/modules/proc_upgrading-smartproxy-server.adoc`).

For context, this is based on the current version of the upgrade guide where in another (now deleted) section, we show this: https://github.com/theforeman/foreman-documentation/blob/2b04b6d2bae80df1d85a38ce9b18b7f2db514039/guides/common/modules/con_project-server-upgrade-considerations.adoc?plain=1#L5

Can you confirm that the current statement about upgrading "from any patch version" being possible is okay?

bangelic (approved):
LGTM

guides/common/modules/con_upgrading-project.adoc
ekohl: Not @evgeni but it really is `patch`, not `minor`.

evgeni: If the question is "are all of the following upgrades allowed?", then the answer is yes:
- Foreman 3.11.0 to 3.12(.latest)
- Foreman 3.11.4 to 3.12(.latest)
- Katello 4.14.0 to 4.15(.latest)
- Katello 4.14.3 to 4.15(.latest)
- Satellite 6.16.0 to 6.17(.latest)
- Satellite 6.16.5.1 to 6.16(.latest)

If the question is "what does major/minor/patch mean exactly in Foreman/Katello/Satellite", then I have no good answer, as we do not follow SemVer, so really every version (3.11, 4.14, 6.16) is major, and these get patches (3.11.4, 4.14.3, 6.16.5), and there is not really a "minor" thing.

ekohl: Good point that we don't really follow semver

aneta-petrova:
I believe the last lingering issue around versioning has been resolved. Unless there is further feedback, I'll merge tomorrow.

aneta-petrova:
Merged to "master", no cherry-picks.

PR #3874, opened 2025-05-21
clarify that it can be *any* system that can run the ansible modules

evgeni:
#### Checklists

### DIFF
--- a/guides/common/modules/proc_creating-a-playbook-with-modules-from-project-ansible-collection.adoc
+++ b/guides/common/modules/proc_creating-a-playbook-with-modules-from-project-ansible-collection.adoc
@@ -18,7 +18,7 @@ The following examples use Ansible vault to manage sensitive data.
 ** Kerberos ticket, if your {ProjectServer} is configured to use {FreeIPA} as an external authentication source
 * A system that can reach the {Project} API.
 This is the system that you will execute your playbook against.
-It can be {ProjectServer} itself.
+It can be the {ProjectServer} itself, a {SmartProxyServer}, or any other system in your environment.
 The following examples use `localhost`.
 
 .Procedure


aneta-petrova (approved):

aneta-petrova:
Style ack from me. And I've seen the downstream conversation that this change originated from, which -- in addition to Evgeni raising the PR -- should be good enough for tech ack.

ekohl (approved):
I think it's fair to say that step 2.1 (installing the modules) is a prerequisite and that covers exactly what's needed for all 3 system flavors (ProjectServer, SmartProxyServer or $random_system).

evgeni:
> I think it's fair to say that step 2.1 (installing the modules) is a prerequisite and that covers exactly what's needed for all 3 system flavors (ProjectServer, SmartProxyServer or $random_system).

Technically, no, that's not required. On the target system you only need Python and `python-requests`.

Step 2.1 says to install the modules on the system you execute `ansible-playbook` on, that's not (necessarily) the same system you target with the `hosts` directive in the playbook.

ekohl:
Hmm yes, good point. So, just for my own sanity, we have:

* System that runs `ansible-playbook`
  * Requires `$collection`
* Target host in playbook
  * Requires `python-requests`
  * Requires access to `$foreman` service
* `$foreman` service

All 3 can be the same, but they can also all be different.

The Ansible modules themselves show an error if `requests` is unavailable so I think it's OK to use the current requirements and not specify `requests` explicitly.

evgeni:
Correct, you're allowed to keep your sanity (for now)

evgeni:
Another thought I just had: if the user doesn't have a system that has direct access to the Foreman API, the modules also can utilize a (http) proxy.

Is that something we want documented here?

ekohl:
And you said I was allowed to keep my sanity. That "for now" did a lot of heavy lifting.

Questions I have for HTTP proxy suppor are:

* Does Kerberos auth work through the proxy or do we need to document a limitation?
* Do we need a code example?

evgeni:
I would expect it to work, but I haven't tested it.

evgeni:
Oh yeah, I wasn't arguing to add it *in this PR*, just… something we should add in general.

evgeni:
   49cd231557..3ba27d8bf8  3.14 -> 3.14

evgeni:
and https://github.com/theforeman/foreman-documentation/pull/3881 for the proxy part

PR #3876, opened 2025-05-21
Update the external DB upgrade section

AkshayGadhaveRH:
In this PR:

- Remove the outdated information about migrating to a fresh pgsql 13 DB.
- Add information about the postgresql-setup utility.

Currently the external DB upgrade process asks the user to install a new OS and migrate the DB. However, this is no longer needed, and the DB can be upgraded in-place.

JIRA: https://issues.redhat.com/browse/SAT-33169

#### Checklists

* [X] Foreman 3.12/Katello 4.14 (Satellite 6.16; orcharhino 7.2 on EL9 only)


aneta-petrova:
I'm looking at _2.2. Upgrading a connected Satellite Server_, which has this step:

* `9. If you are using an external database, upgrade your database to PostgreSQL 13.`

Now that the process of upgrading to PosgreSQL 13 is so simple (or at least its description is :)), it makes me wonder if this is a good opportunity to just delete `guides/common/modules/proc_upgrading-the-external-database.adoc` and move its new contents to that step. Do you think it could work? I like how it would simplify how we explain the upgrade process -- we would no longer be asking users to jump between two sections.

AkshayGadhaveRH:
@aneta-petrova I agree. The module can be dropped and we can turn this into a step.
I'll make those changes in the next commit.

guides/common/modules/proc_upgrading-the-external-database.adoc
```diff
@@ -1,34 +1,6 @@
...
-+
-[options="nowrap", subs="+quotes,verbatim,attributes"]
-----
-# {foreman-installer} \
---foreman-db-host _newpostgres.example.com_ \
---katello-candlepin-db-host _newpostgres.example.com_ \
---foreman-proxy-content-pulpcore-postgresql-host _newpostgres.example.com_
-----
+* You can upgrade an external database from PostgreSQL 12 to PostgreSQL 13 without upgrading the underlying operating system from {EL} 8 to {EL} 9. 
+* You can use the `pg_upgrade` utility to upgrade from PostgreSQL 12 to PostgreSQL 13.
```
ekohl, line 5: I'm not sure this line really adds anything because it's just repeating the line above it with more implementation details. I'd suggest to drop it.
```suggestion
```

guides/common/modules/proc_upgrading-the-external-database.adoc
```diff
@@ -1,34 +1,6 @@
...
-. Restore the backup on the new server.
-ifdef::katello,satellite,orcharhino[]
-. Correct the permissions on the `evr` extension:
-+
-[options="nowrap", subs="+quotes,verbatim,attributes"]
-----
-# runuser -l postgres -c \
-"psql -d foreman -c \"UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='foreman') WHERE extname='evr';\""
-----
-endif::[]
```
ekohl, line 24: This part is really important and specific because it's a Katello extension, not part of standard PostgreSQL. You **must** do this after upgrading to PostgreSQL 13 or users will run into problems in the migration to drop it.

ekohl (changes requested):

guides/common/modules/proc_upgrading-the-external-database.adoc
```diff
@@ -1,34 +1,15 @@
...
-Otherwise reconfigure {Project} to use the new name:
-+
-[options="nowrap", subs="+quotes,verbatim,attributes"]
-----
-# {foreman-installer} \
---foreman-db-host _newpostgres.example.com_ \
---katello-candlepin-db-host _newpostgres.example.com_ \
---foreman-proxy-content-pulpcore-postgresql-host _newpostgres.example.com_
-----
+* Note that you can upgrade the underlying operating system from {EL} 8 to {EL} 9 separately.
```
AkshayGadhaveRH, line 15: @aneta-petrova @ekohl do you think it makes sense to add this sentence here?

guides/common/modules/proc_upgrading-the-external-database.adoc
```diff
@@ -1,34 +1,15 @@
...
-* Create a new {EL} 9 based host for PostgreSQL server that follows the external database on {EL} 9 documentation.
-ifdef::katello,orcharhino,satellite[]
-For more information, see {InstallingServerDocURL}using-external-databases_{project-context}[Using External Databases with {Project}].
-endif::[]
-* Install PostgreSQL version 13 on the new {EL} host.
-
-.Procedure
-. Create a backup.
-. Restore the backup on the new server.
+* If you use an external database, use the `pg_upgrade` utility to upgrade from PostgreSQL 12 to PostgreSQL 13 without upgrading the underlying operating system from {EL} 8 to {EL} 9.
```
ekohl, line 4: I think the method isn't that relevant and in the line below you already give a solution. In the end there's nothing wrong with alternative methods like upgrading the external DB to PostgreSQL 13, but we only give the supported path of in place upgrading.
```suggestion
* If you use an external database, upgrade from PostgreSQL 12 to PostgreSQL 13.
```

ekohl (changes requested):
The placement of this procedure is wrong because you must first upgrade the external DB to PostgreSQL 13 before upgrading the `ProjectServer`.

If you look at https://theforeman-foreman-documentation-preview-pr-3876.surge.sh/3.12/Upgrading_Project/index-foreman-el.html#upgrading_a_connected_foreman_server_upgrading-connected is already has a step:

> 10. If you are using an external database, upgrade your database to PostgreSQL 13.

You are now rewriting the optional step to upgrade the OS from EL 8 to EL9 to be the mandatory step 10.

For Katello (https://docs.theforeman.org/3.12/Upgrading_Project/index-katello.html#upgrading_a_connected_foreman_server_upgrading-connected) it's step 12.

When I read the original issue what we should have done is to expand on that step with the link to `{RHELDocsBaseURL}8/html/deploying_different_types_of_servers/using-databases#fast-upgrade-using-the-pg_upgrade-tool_migrating-to-a-rhel-8-version-of-postgresql` and that would probably have been sufficient.

AkshayGadhaveRH:
@ekohl @aneta-petrova I think it will be easier if we keep this module and point the user to the module instead of adding the steps within the the upgrade section, at least for this release. Otherwise it would be a little difficult to navigate the upgrade module itself.

ekohl:
> In my opinion, that's way too much jumping :)

:+1: on that you shouldn't jump forward and then back. Jumps should be forward only.

> The alternative I'm suggesting is:
> 
>     * If you are using an external database, upgrade your database to PostgreSQL 13:
> 
>     * * Upgrade from PostgreSQL 12 to PostgreSQL 13.
> 
>     * * Correct the permissions on the evr extension.
> 
> 
> The step is now longer and contains a few substeps, but this way users could finish the upgrade by following a single procedure rather than jump between sections.

You can add in a procedure before where you upgrade the external DB from PostgreSQL 12 to 13 before you upgrade Satellite itself.

The high level steps would become:

1. Upgrade your external DB server (if any)
1.1. Upgrade to PostgreSQL 13 (effectively the module you have now)
1.2. Optional: Upgrade the OS to EL9 (effectively the module that already was there)
3. Upgrade your `{ProjectServer}`
4. Upgrade your `{SmartProxyServers}`

aneta-petrova:
Thanks @AkshayGadhaveRH! Ack from the style perspective on the placement of the new lines. Now might be a good idea for a tech review and then we can wrap up the rest of the style review :)

ekohl (changes requested):
Procedure wise this feels awkward. You start a procedure on one host and then in the middle you switch over to another host. Then you switch back to the original host. All while this could be step 3 in the procedure (right after making a backup). Especially upgrading the whole external DB OS can be a fairly long procedure (which is why it was separate). That's what I tried to say in https://github.com/theforeman/foreman-documentation/pull/3876#issuecomment-2911607062 with the new high level overview.

guides/common/modules/proc_upgrading-the-external-database.adoc
```diff
@@ -1,34 +1,15 @@
...
-Otherwise reconfigure {Project} to use the new name:
-+
-[options="nowrap", subs="+quotes,verbatim,attributes"]
-----
-# {foreman-installer} \
---foreman-db-host _newpostgres.example.com_ \
---katello-candlepin-db-host _newpostgres.example.com_ \
---foreman-proxy-content-pulpcore-postgresql-host _newpostgres.example.com_
-----
+. Optional: Upgrade the external database server from {EL} 8 to {EL} 9.
```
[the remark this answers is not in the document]
AkshayGadhaveRH: I am on board with adding more instructions on how to upgrade from EL8 to EL9 (as we have these in 3.14). 
I have created a separate assembly for Upgrading the DB and introduced the `proc_upgrading-the-external-database-operating-system.adoc` module into this branch.

guides/common/modules/proc_upgrading-postgresql-on-the-external-database.adoc
```diff
@@ -0,0 +1,14 @@
+[id="Upgrading_PostgreSQL_on_the_External_Database_{context}"]
+= Upgrading PostgreSQL on the external database
+
+. If you use an external database, upgrade the external database to PostgreSQL 13.
+For more information, see {RHELDocsBaseURL}8/html/deploying_different_types_of_servers/using-databases#fast-upgrade-using-the-pg_upgrade-tool_migrating-to-a-rhel-8-version-of-postgresql[Fast upgrade using the pg_upgrade utility] in _Deploying different types of servers_.
```
ekohl, line 5: I think "Deploying different types of servers" isn't a very clear description. You're right it's the document title, but I wonder if we should be explicit that it's a RHEL 8 guide?

guides/common/modules/proc_upgrading-the-external-database-operating-system.adoc
```diff
@@ -0,0 +1,30 @@
+[id="upgrading-the-external-database-operating-system"]
+= Upgrading the external database operating system
+
+Optional: You can upgrade the external database from {EL} 8 to {EL} 9 using backup and restore.
+
+.Prerequisites
+* Create a host running {EL} 9 for PostgreSQL server that follows the external database on {EL} 9 documentation.
```
ekohl, line 7: I think prerequisites should be a list of items, not a list of steps.
```suggestion
* A host running {EL} 9 for PostgreSQL server that follows the external database on {EL} 9 documentation.
```

ekohl (approved):
From a technical point of view ACK. Some minor notes inline that are more stylistic.

Another question: should this also be reflected in the Upgrade paths section at the start? Also looking at @aneta-petrova here.

And I'd like to ask @evgeni or @ehelms for a technical review since I made quite a few suggestions.

guides/common/modules/proc_upgrading-postgresql-on-the-external-database.adoc
aneta-petrova: In a number of other places, we specify the version of a RHEL guide a bit differently: in _{RHEL}{nbsp}X Guide Name_. Like here: https://github.com/theforeman/foreman-documentation/blob/8b1b052611eccfddfc4a615a2a745ca869e61e3b/guides/common/modules/ref_system-requirements.adoc?plain=1#L38  It would be better to stay consistent.

aneta-petrova, line 5: This procedure module needs an introduction and the `.Procedure` heading. Both are mandatory elements of procedure modules.

Now I'm thinking about how to do this without too much duplication. How about this:

```suggestion
= Upgrading PostgreSQL on the external database

If you use an external database, upgrade the external database to PostgreSQL 13.

.Procedure
. Follow the instructions in {RHELDocsBaseURL}8/html/deploying_different_types_of_servers/using-databases#fast-upgrade-using-the-pg_upgrade-tool_migrating-to-a-rhel-8-version-of-postgresql[Fast upgrade using the pg_upgrade utility] in _{RHEL}{nbsp}8 Deploying different types of servers_ to upgrade the external database.
```

guides/common/modules/proc_upgrading-the-external-database-operating-system.adoc
```diff
@@ -0,0 +1,30 @@
+[id="upgrading-the-external-database-operating-system"]
+= Upgrading the external database operating system
+
+Optional: You can upgrade the external database from {EL} 8 to {EL} 9 using backup and restore.
```
aneta-petrova, line 4: `Optional:` is AFAIK typically used in procedure steps. In introductions, I most often see it written as a regular part of the sentence, like this:

```suggestion
Optionally, you can upgrade the external database from {EL} 8 to {EL} 9 using backup and restore.
```

The reason is that an introduction is just regular text so it should follow a normal sentence flow.

aneta-petrova:
> Another question: should this also be reflected in the Upgrade paths section at the start? Also looking at @aneta-petrova here.

This is a valid suggestion @AkshayGadhaveRH. If you look at the "Upgrade paths" section, it should now include a new step that says something along the lines of "If you use an external database, upgrade the external database to PostgreSQL 13 and optionally upgrade the underlying operating system from Red Hat Enterprise Linux 8 to Red Hat Enterprise Linux 9."

guides/common/modules/proc_upgrading-postgresql-on-the-external-database.adoc
AkshayGadhaveRH: I used `If you use an external database, upgrade the external database and rectify the permissions.` 
WDYT?

guides/common/modules/proc_upgrading-postgresql-on-the-external-database.adoc
```diff
@@ -0,0 +1,17 @@
+[id="Upgrading_PostgreSQL_on_the_External_Database_{context}"]
+= Upgrading PostgreSQL on the external database
+
+If you use an external database, upgrade the external database and rectify the permissions.
```
evgeni, line 4: Can the "and rectify the permissions" part be hidden behind "ifndef::foreman-el" (or ifdef::katello_etc)? Otherwise the introduction says "do these two things" but then there is just one step to do.

evgeni:
tech-ack. my comments are purely on wording/formatting :)

guides/common/modules/proc_upgrading-postgresql-on-the-external-database.adoc
```diff
@@ -0,0 +1,17 @@
...
+
+.Procedure
+. If you use an external database, upgrade the external database to PostgreSQL 13.
+For more information, see {RHELDocsBaseURL}8/html/deploying_different_types_of_servers/using-databases#fast-upgrade-using-the-pg_upgrade-tool_migrating-to-a-rhel-8-version-of-postgresql[Fast upgrade using the pg_upgrade utility] in _{RHEL}{nbsp}8 Deploying different types of servers_.
```
aneta-petrova, line 8: There is no need to repeat the "If..." condition here if it's now included in the procedure introduction. It applies to the whole procedure, not just the first step.

My earlier suggestion was this:

```suggestion
. Follow the instructions in {RHELDocsBaseURL}8/html/deploying_different_types_of_servers/using-databases#fast-upgrade-using-the-pg_upgrade-tool_migrating-to-a-rhel-8-version-of-postgresql[Fast upgrade using the pg_upgrade utility] in _{RHEL}{nbsp}8 Deploying different types of servers_ to upgrade the external database.
```

An alternative might be something like this:

```suggestion
. Upgrade the external database to PostgreSQL 13.
For more information, see {RHELDocsBaseURL}8/html/deploying_different_types_of_servers/using-databases#fast-upgrade-using-the-pg_upgrade-tool_migrating-to-a-rhel-8-version-of-postgresql[Fast upgrade using the pg_upgrade utility] in _{RHEL}{nbsp}8 Deploying different types of servers_.
```

I'm sure there are other ways of writing this but the main point is not to repeat the "If..." condition.

AkshayGadhaveRH:
Hey @aneta-petrova is this good to merge?

aneta-petrova (approved):

PR #3905, opened 2025-05-29
Update Satellite server topology diagram

aneta-petrova:

While the original request was to change some information in the existing diagram and add more information, I don't consider that feasible. Instead, I'm proposing to replace the existing complex  architecture diagrams with simpler diagrams accompanied by further explanations and lists in textual form.

Information suitable for a diagram:
* Relationships between components
* Hierarchical structure of components
* Flow of data

Information not suitable for a diagram:
* Information that needs to be scannable, like lists of ports or components
* Information that needs to be searchable, for example with full-text search
* Anything that requires a more detailed explanation, such as which plugins can be optional

With this PR, I'm including new diagrams that include information suitable for a diagram. I'm moving information not suitable for a diagram into appropriate sections as text.

~~* Drop port 8443~~
~~* Foreman talks to the Foreman Proxy on the server~~
~~* Include PostgreSQL and Redis~~
~~* Move Pulp out of the "Internal Capsule" box and put it alongside Foreman and Candlepin~~
~~* Drop the term "Isolated Capsule" in favor of "Capsule"~~
~~* Drop the term "Isolated Host" in favor of "Host"~~
~~* Add Registration to Foreman Proxy~~
~~* Drop port 8443~~

https://issues.redhat.com/browse/SAT-21492

The core reason to simplify the diagram is to improve clarity and usability. An overly complex diagram can be overwhelming, which makes it difficult for users to quickly grasp the system's essentials. By moving certain types of information to a textual format, we can make the diagram more effective at its primary job: visualizing relationships between components. The accompanying text then takes care of the secondary objective: providing specific information and detailed lists.

Existing issues that I observe in the diagram:
* The list of services is expected to change, and it will be quicker to update them as text rather than in a diagram.
* Too many arrows and crossing lines make it difficult to extract information from the old diagrams.
This is the information that I'm breaking out of the diagram and instead including it in the appropriate sections.
* The section the diagrams are included in is named "Smart Proxy networking". The current diagram tries to encompass much more than that.

#### Checklists

* [x] Foreman 3.14/Katello 4.16 (Satellite 6.17)
* [x] Foreman 3.12/Katello 4.14 (Satellite 6.16; orcharhino 7.2 on EL9 only)


aneta-petrova:
Hi @ehelms, can you please take a look at the updated diagram? I did my best to implement the suggestions from https://issues.redhat.com/browse/SAT-21492. I was unclear about these two:

> Apache is integral part of the architecture and I think it's instructive to include it so folks understand how the communication routes through it for services

I don't know where exactly to add Apache. Can you help?

> We should include PostgreSQL and Redis

For now, I added Redis only to Foreman. Should I add it to both Foreman Proxies too?

Note that this is a WIP version so the elements are not aligned properly etc. This is because I want to first make sure to collect feedback and acks on technical accuracy. Only then will I start finalizing the design itself.

aneta-petrova:
Thanks @ehelms! I added the missing components reported in the Jira. Can you please re-review? Are all the components in the appropriate places now?

In https://github.com/theforeman/foreman-documentation/pull/3905/commits/d3e37667b5ef41aa6a2a45b86a03f045ae812657, I also dropped the port numbers from the diagram and replaced them with a textual link to the section on port requirements in the Installation guide. This helps us maintain this information in one place only and in text form, which is easier. Additionally, I believe it makes the diagram itself look cleaner and less intimidating at first glance. Let me know if you have any concerns about this particular change.

ehelms:
A few notes on what we should try to show:

- I think we could drop the plugins listed inside of Foreman
- Apache should sit on top of and across Foreman, Pulp, Candlepin
- Postgresql and Redis should be outside of the Foreman Proxy

One thing our port/firewall documentation doesn't appear to do is talk about it in terms of the components in this diagram. That makes it hard to connect the diagram with the port table.

aneta-petrova:
> One thing our port/firewall documentation doesn't appear to do is talk about it in terms of the components in this diagram. That makes it hard to connect the diagram with the port table.

Is this something we tell users to do or expect them to do? To open or close ports to enable or prevent communication between the individual components? I think users are only expected to open ports as described in https://docs.theforeman.org/nightly/Installing_Server/index-katello.html#Enabling_Connections_from_a_Client_to_Server_foreman, aren't they? If so, then including details on internal communications (ports between individual components) doesn't represent any actionable items for users, which would support dropping the port numbers + reducing the number of arrows. The current diagram seems more like a jungle of arrows and numbers than a clear and useful diagram :)

ehelms:
> Is this something we tell users to do or expect them to do? To open or close ports to enable or prevent communication between the individual components? I think users are only expected to open ports as described in https://docs.theforeman.org/nightly/Installing_Server/index-katello.html#Enabling_Connections_from_a_Client_to_Server_foreman, aren't they? If so, then including details on internal communications (ports between individual components) doesn't represent any actionable items for users, which would support dropping the port numbers + reducing the number of arrows. The current diagram seems more like a jungle of arrows and numbers than a clear and useful diagram :)

Users ask these kind of questions as they like to understand what is talking to what. And I think through docs we can capture that information and prevent users raising issues or forum questions. 

A table works.

aneta-petrova:
The feedback I received internally says that this isn't the right approach either. Hopefully, after more clarification, we'll be able to fix the diagram in a way that is acceptable from both the docs and tech perspective. Until that clarification happens, I'm closing this PR.

ekohl (commented):
I was writing up a review before you closed it so I'm still going to post that.

PR #3625, opened 2025-01-29
Restructure boot disks in provisioning

Lennonka:
#### What changes are you introducing?

Splitting the boot disk intel into multiple modules and laying out a structure that is easier to follow

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

- To make it more beginner friendly
- Following up on #3472 

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)

- We should consider adding more information about the boot disk types and their use cases.

#### Checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into: N/A

PR #3472, opened 2024-11-25
Create a new guide for installer-based provisioning (1 guide of 3)

Lennonka:
#### What changes are you introducing?

Splitting _Provisioning hosts_ into 3 new guides:

1. **Provisioning machines by using an installer** - includes network boot and boot disks, which are both installer-based provisioning, plus Discovery, which also uses network boot _**<--- We are here.**_
2. Provisioning virtual machines - includes virtualization providers, image-based provisioning, and directives for installer-based provisioning if available
3. Provisioning cloud instances - includes cloud infrastructure providers and image-based provisioning only

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

To mitigate pain points identified in [SAT-26089](https://issues.redhat.com/browse/SAT-26089) comments (public).

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)

- _Provisioning hosts_ will be destroyed after all 3 guides from above will have been created.
- I've removed `_{context}` where it was needed for module reuse. Unfortunately, these changes were extensive to make the build work and this PR is as small as it gets.
- Currently, I am only modifying the structure. Further, detailed improvements will be done in subsequent PRs. Therefore, the new guide is terminologically inconsistent.
  **Reviewers, please, focus on the title and heading structure in this PR.** The rest of the affected guides are just link updates. Using preview first is highly recommended.
- Will require updates of documentation links in the web UI.

#### Checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into: N/A - for Foreman 3.14

PR #3922, opened 2025-06-10
Incorporate improvements to original provisioning

Lennonka:

Adding provisioning improvements to the original _Provisioning hosts_ guide that can be reused from other PRs: #3472, #3608, #3625
In particular:

- Splitting the chapter _Configuring provisioning resources_ into multiple, more structured chapters per a set of resources _Prepare `<resource set>`_ to make them more findable.
- Moving security settings to the chapter _Security settings_
- Splitting the chapter _Using PXE to provision host_ into _Using network boot to provision hosts_ and _Using boot disks to provision hosts_.
- Moving the chapter _Using iPXE_ under _Using network boot_.
- Removing an outdated initialization script appendix.
- ~~Replacing host destruction on VM deletion procedure with a warning snippet.~~ moved to #3937 

- We've decided to move in a different direction with the guide split. We'll create content journeys to decide whether and how the guide should be split. In the mean time, we would like to deliver smaller improvements.
- These improvements should help make the guide more beginner friendly.

- Using **preview** to get the general idea of the suggested structure is highly recommended.
- We don't have to use all of them in case they aren't helpful.
- Dropping the _Provisioning with installer_ guide in #3923 
- **Target version: 3.16**

#### Checklists


guides/common/modules/snip_warning-destroy-vm-on-host-delete.adoc
```diff
@@ -0,0 +1,8 @@
+[WARNING]
+====
+By default, if a host entry that is associated with a virtual machine is deleted, the virtual machine will remain on the compute resource.
```
apinnick, line 3:
```suggestion
By default, deleting a host entry associated with a virtual machine does not delete the virtual machine on the compute resource.
```
Suggestion for removing passive voice and future tense.

guides/common/modules/snip_warning-destroy-vm-on-host-delete.adoc
```diff
@@ -0,0 +1,8 @@
...
+
+If you want {Project} to destroy virtual machines on the compute resource when the associated host entry is deleted, navigate to *Administer* > *Settings* > *Provisioning* and set the *Destroy associated VM on host delete* setting to *Yes*.
+
+To avoid destroying virtual machines in this situation, disassociate the virtual machine from {Project} without removing it from the compute resource or keep the setting set to *No*.
```
apinnick, line 7: I don't think you need this sentence because you have already explained the default behavior.

Lennonka:
Description updated. cc @apinnick Anything that's missing?

guides/common/modules/snip_warning-destroy-vm-on-host-delete.adoc
Lennonka: A small suggestion:
```suggestion
By default, deleting a host entry associated with a virtual machine does not destroy the virtual machine on the compute resource.
```

Lennonka:
Rebased.

stejskalleos (approved):
LGTM, haven't found any issues, but the final ack should be done by DOC person who understands the guide structures better than I do.

ekohl (commented):
Some things that come to mind. Many of these weren't changed by you here, but I'm going to dump them. Feel free to split them off to some other issue because it may be more of the start of a content journey.

Looking at the foreman-el build first. Note I only got to the point of Provisioning VMs on libvirt.

# Supported cloud providers

https://theforeman-foreman-documentation-preview-pr-3922.surge.sh/nightly/Provisioning_Hosts/index-foreman-el.html#supported-cloud-providers

Should each entry link to the "Provisioning cloud instances on X" chapter we have?

# Supported virtualization infrastructures

https://theforeman-foreman-documentation-preview-pr-3922.surge.sh/nightly/Provisioning_Hosts/index-foreman-el.html#supported-virtualization-infrastructures

Same as above for cloud providers.

# Preparing provisioning content

https://theforeman-foreman-documentation-preview-pr-3922.surge.sh/nightly/Provisioning_Hosts/index-foreman-el.html#preparing-provisioning-content

It lists Lorax Composer image as the first entry, but I think this isn't the primary way users provision. I'd suggest to move this lower. Installation medium is something you need to set up first anyway. It's also something that is only going to work if you use Anaconda so it must be used in combination with the Red Hat OS family installer setups.

# Preparing client platforms

https://theforeman-foreman-documentation-preview-pr-3922.surge.sh/nightly/Provisioning_Hosts/index-foreman-el.html#preparing-client-platforms

I don't like that we have many procedures that are essentially all the same. Creating Debian 12 & 11 and Ubuntu 24.04 & 22.04 are effectively all the same, just different on what you need to fill in where.

I know you didn't change this, but could we have a single "Creating operating system" procedure and then an appendix with a table that tells users what they can fill in?

# Facts and NIC filtering

https://theforeman-foreman-documentation-preview-pr-3922.surge.sh/nightly/Provisioning_Hosts/index-foreman-el.html#facts-and-nic-filtering_provisioning

Why is this in the provisioning guide? IMHO this belongs in the administration guide.

# Optimizing performance by removing NICs from database

https://theforeman-foreman-documentation-preview-pr-3922.surge.sh/nightly/Provisioning_Hosts/index-foreman-el.html#facts-and-nic-filtering_provisioning

Same as `Facts and NIC filtering` above: belongs in the admin guide.

# Options in managed DHCPv4

https://theforeman-foreman-documentation-preview-pr-3922.surge.sh/nightly/Provisioning_Hosts/index-foreman-el.html#options-in-managed-dhcpv4

This belongs in the Integrating provisioning services guide.

I'd argue we can drop the whole chapter because it's already covered by https://docs.theforeman.org/nightly/Integrating_Provisioning_Infrastructure_Services/index-foreman-el.html#configuring-tftp-integration.

# Options in unmanaged DHCPv6

https://theforeman-foreman-documentation-preview-pr-3922.surge.sh/nightly/Provisioning_Hosts/index-foreman-el.html#options-in-unmanaged-dhcpv6

Similar to DHCPv4 this needs to be moved to the Integrating provisioning services guide.

# Prerequisites for image-based provisioning

https://theforeman-foreman-documentation-preview-pr-3922.surge.sh/nightly/Provisioning_Hosts/index-foreman-el.html#Prerequisites_for_Image_Based_Provisioning_provisioning

> Images that use the finish post-boot configuration scripts require a managed DHCP server, for example, your Foreman server or Smart Proxy servers. The host must be created with a subnet associated with a DHCP Smart Proxy, and the IP address of the host must be a valid IP address from the DHCP range.

Is this really true? If you assign static IP addresses then I'd expect it to work without DHCP.

In fact, various compute resources actually provide the IP address and it can't work with DHCP. This whole chapter needs a thorough review because it looks factually wrong.

# Multiple subnets or domains using installer

https://theforeman-foreman-documentation-preview-pr-3922.surge.sh/nightly/Provisioning_Hosts/index-foreman-el.html#Multiple_Subnets_or_Domains_Using_Installer_provisioning

Needs to move to the new Integrating provisioning services guide.

# DNS options for network configuration

https://theforeman-foreman-documentation-preview-pr-3922.surge.sh/nightly/Provisioning_Hosts/index-foreman-el.html#DNS_Options_For_Network_Configuration_provisioning

Needs to be integrated into the new Integrating provisioning services guide. I think it's already largely redundant, possibly completely.

# TFTP options for network configuration

Needs to be dropped since it's already covered by the new Integrating provisioning services guide.

# Changing the provisioning password hashing algorithm

https://theforeman-foreman-documentation-preview-pr-3922.surge.sh/nightly/Provisioning_Hosts/index-foreman-el.html#Changing_the_Provisioning_Password_Hashing_Algorithm_provisioning

I think we should mention that SHA512 is recommended since EL5 and became default in EL6. Lowering it to SHA256 will lower security, not raise it. Since Foreman 3.12 the default changed from SHA256 to SHA512. Note it didn't update existing operating systems on upgrade.

Quoting https://github.com/theforeman/foreman/commit/e2dee7d6f7ffb505d317ca1b205f302ccd622063

> SHA256 is no longer considered supported and is prohibitted (sic) by some policies

So I think we should drop this chapter.

# Using network boot to provision hosts

https://theforeman-foreman-documentation-preview-pr-3922.surge.sh/nightly/Provisioning_Hosts/index-foreman-el.html#using-network-boot-to-provision-hosts

This lists all the various ways to set up Secure Boot, but can't we just point to Configuring Smart Proxy for Secure Boot and provide the table there.

Note that all the procedures there are written are for the **admin** persona, not a user persona. Because of that you could argue it should move to the new Integrating provisioning services guide and then be listed as a prerequisite.

# Using boot disks to provision hosts

https://theforeman-foreman-documentation-preview-pr-3922.surge.sh/nightly/Provisioning_Hosts/index-foreman-el.html#using-boot-disks-to-provision-hosts

In upstream we don't ship the bootdisk plugin by default and this procedure misses instructions to enable it. In Satellite it's enabled by default so there it isn't an issue.

# Provisioning virtual machines on KVM (libvirt)

https://theforeman-foreman-documentation-preview-pr-3922.surge.sh/nightly/Provisioning_Hosts/index-foreman-el.html#Provisioning_Virtual_Machines_on_KVM_kvm-provisioning

I don't like the focus on KVM. In Foreman we always talk about libvirt, not KVM. KVM is an internal implementation detail.

It's also confusing that in the prerequisites there's:

> Optional: The examples in these procedures use the root user for KVM. If you want to use a non-root user on the KVM server, you must add the user to the libvirt group on the KVM server:

But this is only required if libvirt is on the same machine as Foreman.

For upstream it's also missing the instruction to install the libvirt compute resource. You can use the same procedure as "Installing VMware plugin" but the command is `foreman-installer --enable-foreman-compute-libvirt`.

> Configuring Foreman server for KVM connections

This should be rephrased to mention it's only for adding **remote** libvirt connections. It's not needed if libvirt is on the same system.

guides/doc-Provisioning_Hosts/master.adoc
```diff
@@ -17,17 +17,25 @@ endif::[]
...
 include::common/assembly_introduction-to-provisioning.adoc[leveloffset=+1]
 
-include::common/assembly_configuring-provisioning-resources.adoc[leveloffset=+1]
+include::common/assembly_preparing-provisioning-content.adoc[leveloffset=+1]
+
+include::common/assembly_preparing-client-platforms.adoc[leveloffset=+1]
+
+include::common/assembly_preparing-templates-for-provisioning.adoc[leveloffset=+1]
 
 include::common/assembly_preparing-networking.adoc[leveloffset=+1]
```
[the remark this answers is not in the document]
ekohl: Possibly. Long term you could say that networking becomes more and more a prerequisite that an admin needs to address and it makes sense to list prerequisites first.

AkshayGadhaveRH (commented):
Only a couple observations, otherwise LGTM!

Lennonka:
> Some things that come to mind. Many of these weren't changed by you here, but I'm going to dump them. Feel free to split them off to some other issue because it may be more of the start of a content journey.

@ekohl FYI, I've filed sub-issues for each of your reported issues under #4015 

I propose to merge this PR as is and work on those issues in following PRs.

guides/doc-Provisioning_Hosts/master.adoc
```diff
@@ -17,17 +17,25 @@ endif::[]
...
 
-include::common/assembly_using-pxe-to-provision-hosts.adoc[leveloffset=+1]
+include::common/assembly_security-settings.adoc[leveloffset=+1]
 
-include::common/assembly_using-ipxe-to-reduce-provisioning-times.adoc[leveloffset=+1]
+include::common/modules/con_provisioning-contexts.adoc[leveloffset=+1]
 
-include::common/assembly_discovering-hosts-on-a-network.adoc[leveloffset=+1]
+include::common/modules/proc_setting-the-provisioning-context.adoc[leveloffset=+2]
```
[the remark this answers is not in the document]
Lennonka: That's reasonable. I'll do this in this PR.

Lennonka:
Rebased.

Lennonka:
Is anyone against merging this?

I'm not sure what to do about testing. Probably not everything at once, right?

maximiliankolb (approved):
Thanks Lena, this is a great improvement!

shubhamsg199 (approved):
Ack, LGTM

PR #3928, opened 2025-06-11
Retroactively add 4.16 warning about container naming

ianballou:
Adds a release warning for 4.16 as as response to https://community.theforeman.org/t/docker-image-repos-renamned-with-foreman-3-14-katello-4-16/43493

PR #3929, opened 2025-06-11
Add container image naming upgrade warning

ianballou:
Adds an upgrade warning in 4.17 in case users didn't see the late addition https://github.com/theforeman/foreman-documentation/pull/3928


ekohl (approved):
Good idea to call it out

PR #3946, opened 2025-06-19
Add review checklists to PR template

aneta-petrova:

Adding 'review checklists' to the PR template.

Discussions over what conditions need to be met for a PR to be considered reviewed have been popping up recently. For example: Should we request a second opinion for PRs filed by engineers? How to introduce the requirement for QE/testing review when the concept of QE reviews is not native to upstream? What should be the scope of a review anyway?

Standardizing expectations could be beneficial for all sides: both writers and reviewers. These checklists could also eliminate the need to manually assign the `needs testing`/`testing done` labels (because they include testing as part of the regular tech review).
~~EDIT: This might actually eliminate the tech and style review labels too!~~

A previous suggestion by @ekohl was to use [Github's saved replies](https://docs.github.com/en/get-started/writing-on-github/working-with-saved-replies/using-saved-replies) feature and maintain the lists manually for each reviewer. Adding them to the PR template would make the template even longer than it already is but it would mean we only have one place where we store it.

#### Checklists

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,7 +4,7 @@
 
 #### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)
 
-#### Checklists
+#### Contributor checklists
 
 * [x] I am okay with my commits getting squashed when you merge this PR.
 * [ ] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.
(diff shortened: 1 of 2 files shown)


ekohl:
> A previous suggestion by @ekohl was to use [Github's saved replies](https://docs.github.com/en/get-started/writing-on-github/working-with-saved-replies/using-saved-replies) feature and maintain the lists manually for each reviewer. Adding them to the PR template would make the template even longer than it already is but it would mean we only have one place where we store it.

One downside of the PR template is that it's in the original message and it's hard to figure out who ACKed it. I don't have a good solution either.

aneta-petrova:
> One downside of the PR template is that it's in the original message and it's hard to figure out who ACKed it. I don't have a good solution either.

Can I assume that only users with commit access and the PR author can check the boxes? If so, then that at least limits who can do it.

An alternative could be to keep using labels (for which GH always shows who set/unset them) and add the review checklists to [CONTRIBUTING.md](https://github.com/theforeman/foreman-documentation/blob/a5c1bf8cea174c825ff63938c1d4697da35b0844/CONTRIBUTING.md?plain=1#L15). But should we expect reviewers to read those guidelines? I wouldn't.

maximiliankolb (approved):
diff LGTM

aneta-petrova:
Considering that there has been one approval and no blocking comments in 2 weeks, I'll move forward with this. Next step: Update the tech review/style review part of contributing guidelines accordingly.

@Lennonka What's your opinion on potentially replacing the testing review/tech review/style review labels with checklists like these? IIRC you were the one who implemented the testing label so I'd like to know what your thoughts are.

aneta-petrova:
I updated the contribution guidelines to drop the (weak) definition of tech and style review that was available there. This is what the updated PR template (with a more detailed definition provided by the new checklists) replaces.

@maximiliankolb Can you please re-review?

maximiliankolb (commented):
diff still LGTM.

PR #3950, opened 2025-06-22
Add assembly attribute

apinnick:

Adding content type attribute to assemblies.

Required for tooling migration.

#### Checklists

### DIFF
--- a/guides/common/assembly_accessing-server.adoc
+++ b/guides/common/assembly_accessing-server.adoc
@@ -1,3 +1,5 @@
+:_mod-docs-content-type: ASSEMBLY
+
 include::modules/con_accessing-server.adoc[]
 
 include::modules/proc_logging-in-to-the-projectwebui.adoc[leveloffset=+1]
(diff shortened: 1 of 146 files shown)


apinnick:
@maximiliankolb Please review. Will this attribute create problems for other products?

guides/common/assembly_hammer-reference.adoc
```diff
@@ -1,3 +1,5 @@
+:_mod-docs-content-type: ASSEMBLY
+
 // DO NOT EDIT MANUALLY
```
maximiliankolb, line 3: I think this will be overwritten by `generate-hammer-reference.sh`. OK for now, but maybe we need a check or adjustment to this shellscript to not drop this change when we recreate the Hammer CLI reference.

maximiliankolb (approved):
One small comment that we could probably resolve by adding a comment/action item to the shellscript for now. Overall, LGTM.

guides/common/assembly_hammer-reference.adoc
apinnick: Good point. I forgot about auto-generated files. Will fix.

Lennonka: Thanks for catching this! I would prefer to update the script as part of this PR.

Lennonka (approved):
Thank you!

PR #3967, opened 2025-06-30
Check URLs on a weekly basis

maximiliankolb:

Our docs include links to various external sources, such as "ansible.com" and similar. To provide a good UX, we want to ensure that all external URLs work.

However, due to the active work of contributors, having a GHA that runs for each commit produces a lot of noice on our PRs. Therefore, we want to use a schedule in favor of checks on PRs.

By default, all maintains will get notified for failed checks.

To reduce noice on PRs and therefore improve maintaining Foreman docs.

#### Checklists

### DIFF
--- a/.github/workflows/linkchecker.yml
+++ b/.github/workflows/linkchecker.yml
@@ -1,11 +1,10 @@
 ---
-name: linkchecker
+name: "Check URLs"
 
-on:
-  pull_request:
-    branches:
-      - master
-      - "[0-9]+.[0-9]+"
+on:  # yamllint disable-line rule:truthy
+  schedule:
+    - cron: "18 4 * * sun"
+
 env:
   RUBY_VERSION: 3.1
   BUNDLE_WITHOUT: "nanoc"


ekohl (changes requested):
:-1: on removing it for PRs. If you add a link that's broken, then you should fix it. That's the whole idea behind CI.

Now what makes this flaky is that it's external websites so you don't control them.

You can mitigate that by only looking at the files that you changed and check those URLs. This speeds up the check and gives fewer false positives. On the other hand, now you need someone to check those results.

> By default, all maintains will get notified for failed checks.

I don't get notifications from failed scheduled runs because I turned it off. It created too much noise for me in the past. How do you intend to check failures? I think this will end in us just ignoring failures.

If you want to run this, I'd suggest to look into automation that creates an issue if broken links are detected.

Another note: if this is about finding broken links, would it be better to just let linktryer-checker loose on https://docs.theforeman.org/ without building it locally? Then it should find all versions and not just the branch that is being checked.

PR #3968, opened 2025-06-30
Add mod doc type attribute to script

apinnick:

Adding the mod docs type attribute to generate-hammer-reference script so that modules are generated with the attribute.

This attribute will be required if we try to set up a CI job to migrate foreman docs continuously to the d/s repo after we migrate to new tooling.

#### Checklists

### DIFF
--- a/guides/scripts/generate-hammer-reference.sh
+++ b/guides/scripts/generate-hammer-reference.sh
@@ -3,6 +3,7 @@
 SRC=
 ASMB_DIR=common # assembly destination folder
 MOD_DIR=modules
+MOD_TYPE=":_mod-docs-content-type: REFERENCE"
 ASMB_FILENAME=assembly_hammer-reference.adoc
 NOOP= # dry run
 DEBUG=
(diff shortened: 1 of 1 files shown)


guides/scripts/generate-hammer-reference.sh
```diff
@@ -154,7 +155,7 @@ cat "$SRC" | while read line ; do
             id=[id=\"$id_core\"]
             heading=`echo -e $line | sed 's/^### hammer/==/'`
```
apinnick, line 156: According to our mod doc guidelines, all modules should begin with level 1 headers. 

Would it be possible to change this to level 1 and add 'echo -e "include::modules/$filename[leveloffset=+2]\n" >>$asmb_file'?

Lennonka: Potentially yes, but the problem is that there are 5 levels of commands. If we structure the modules for 5 levels, the nesting of the modules will be too deep (>3 nesting levels of sections). We don't want that either.

Perhaps we could think a bit deeper about the design of the reference before we make any changes to the script.

Lennonka (approved):
Tested successfully.

PR #3975, opened 2025-07-03
Add a new guide for the new installer

aneta-petrova:

Introducing a new guide for the upcoming new installer:

* A new directory, named `doc-Installing_Server_New_Installer`.
* In that directory, an empty `master.adoc` that includes established attributes + an IMPORTANT box explaining that this is an experimental feature + the `:new-installer:` attribute that we can use in the future when including existing modules (if there is a need to differentiate between the old installer and the new installer).

The purpose of this PR is to introduce an empty guide and establish a preview. We will start adding the actual content in subsequent PRs.

I'm intentionally not adding the guide to the website navigation. It's too soon for that.

The long term plan is for this new installation guide to eventually replace the current installation guide -- when the new installer replaces the old installer.

#### Contributor checklists

#### Review checklists

Tech review (performed by an Engineer who did not author the PR; can be skipped if tech review is unnecessary):

* [ ] The PR documents a recommended, user-friendly path.
* [ ] The PR removes steps that have been made unnecessary or obsolete.
* [ ] Any steps introduced or updated in the PR have been tested to confirm that they lead to the documented end result.

Style review (by a Technical Writer who did not author the PR):

* [ ] The PR conforms with the team's style guidelines.
* [ ] The PR introduces documentation that describes a user story rather than a product feature.


guides/doc-Installing_Server_foremanctl/master.adoc
```diff
@@ -0,0 +1,15 @@
+include::common/attributes.adoc[]
+include::common/header.adoc[]
+:installing-satellite-server-connected:
```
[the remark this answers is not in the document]
aneta-petrova: If there is a need for upstream-friendly naming, that should be resolved in a separate PR that also targets the existing installation guide. Changing it here only could prevent us from reusing existing modules.

guides/doc-Installing_Server_foremanctl/master.adoc
```diff
@@ -0,0 +1,15 @@
...
+:context: {project-context}
+:mode: connected
+:ProductName: {ProjectServer}
+:foremanctl:
```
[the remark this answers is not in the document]
aneta-petrova: > I'm not sure that's needed. The attribute doesn't have a value and it would be used for conditionals only I assume.

Spot on! The purpose of the attribute is to be able to differentiate between "old installer" and "new installer" in existing modules.

guides/doc-Installing_Server_foremanctl/master.adoc
```diff
@@ -0,0 +1,15 @@
...
+
+= {InstallingServerDocTitle}
+
+[IMPORTANT]
+====
+The `foremanctl` installer is an experimental feature only.
+This guide provides early documentation, enabling you to test the functionality and provide feedback.
+====
```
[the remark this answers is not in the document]
aneta-petrova: https://github.com/theforeman/foreman-documentation/pull/3996 indeed makes it possible to use the snippet. Thanks!

aneta-petrova:
I'll keep asking questions until someone answers at least one of them :) If I avoid creating a whole new guide and just use an existing guide but with the new `foremanctl` attribute, how would I go about creating an HTML build for it? In other words, now we generate:

* https://docs.theforeman.org/nightly/Installing_Server/index-foreman-el.html
* https://docs.theforeman.org/nightly/Installing_Server/index-foreman-deb.html
* https://docs.theforeman.org/nightly/Installing_Server/index-katello.html
* https://docs.theforeman.org/nightly/Installing_Server/index-satellite.html

How would I make sure that we also generate this?

* https://docs.theforeman.org/nightly/Installing_Server/index-foremanctl.html

maximiliankolb:
I strongly recommend not adding a new flavour "foremanctl" next to all other five flavours. I think a new guide is the way to go.

This would require a huge effort to fix basically _all_ ifdef/ifndef statements.

Maybe we need to think of "foremanctl" as an attribute similar to "{SmartProxy}" which works well IMHO.

guides/doc-Installing_Server_foremanctl/docinfo.xml
```diff
@@ -0,0 +1,15 @@
+<title>{InstallingServerDocTitle}</title>
```
Lennonka, line 1: I'm wondering if this guide should get a more specific title, both upstream and downstream?
Considering they will be published side by side, it would be confusing to have two guides with the same title.

guides/doc-Installing_Server_foremanctl/master.adoc
```diff
@@ -0,0 +1,12 @@
+include::common/attributes.adoc[]
+include::common/header.adoc[]
+:installing-satellite-server-connected:
+:context: {project-context}
+:mode: connected
+:ProductName: {ProjectServer}
+:installer-foremanctl:
```
Lennonka, line 7: On second thought, given that we have "foreman-installer", we might want to name it a bit differently.
```suggestion
:install-with-foremanctl:
```
Non blocking, but better for clarity of the attribute purpose.

Lennonka (changes requested):
Perhaps the titles and links should also be defined in this PR?
Your call if you plan to do this later.

guides/doc-Installing_Server_foremanctl/docinfo.xml
aneta-petrova: Good point, how about we just add "by using foremanctl" for now? I think that should be good enough -- we're just setting up a very simple preview guide that we're not even adding to the navigation (yet), no need to come up with a definitive title at this stage.

Lennonka (changes requested):
You need to add new attributes, such as `:InstallingServerCTLDocTitle:`.
This would only change the existing titles.

guides/doc-Installing_Server_foremanctl/master.adoc
Lennonka: I meant for our clarity :) Still an optional recommendation.

guides/doc-Installing_Server_foremanctl/docinfo.xml
Lennonka, line 1:
```suggestion
<title>{InstallingServerCTLDocTitle}</title>
```

Lennonka (approved):
One nitpick, otherwise I'm happy with it.

PR #4048, opened 2025-07-22
xref formats test

apinnick:
testing xref format. See "9. Using boot disks to provision hosts".

### DIFF
--- a/guides/common/modules/con_using-boot-disks-to-provision-hosts.adoc
+++ b/guides/common/modules/con_using-boot-disks-to-provision-hosts.adoc
@@ -13,3 +13,6 @@ Using this ISO, the host can boot the installation media, connect to {SmartProxy
 THIS IS THE NEW LINK.
 For more information, see xref:common/modules/con_discovery-in-pxeless-mode.adoc#discovery-in-pxeless-mode[Discovery in PXE-less mode].
 
+THIS IS THE LINK WITHOUT THE ANCHOR ID.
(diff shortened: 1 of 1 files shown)


guides/common/modules/con_using-boot-disks-to-provision-hosts.adoc
```diff
@@ -9,4 +9,7 @@ This is also known as PXE-less provisioning and involves generating a boot ISO t
 Using this ISO, the host can boot the installation media, connect to {SmartProxy}, and install the operating system.
 
 {Project} also provides a PXE-less discovery service that operates without PXE-based services, such as DHCP and TFTP.
-For more information, see xref:discovery-in-pxeless-mode[].
+
+THIS IS THE NEW LINK.
+For more information, see xref:common/modules/con_discovery-in-pxeless-mode.adoc#discovery-in-pxeless-mode[Discovery in PXE-less mode].
```
maximiliankolb, line 14:
```suggestion
For more information, see xref:common/modules/con_discovery-in-pxeless-mode.adoc[Discovery in PXE-less mode].
```

Does this work as well? There's only one heading which is at the top, so specifying the anchor should not be necessary in this case. I imagine that it's mostly necessary for CLI procedures.

maximiliankolb: > we can't migrate procedure modules that contain more than one procedure

Is there a technical reason for this?

This requirement alone could mean a lot of work

PR #4091, opened 2025-07-31
Add line break to end of assembly

maximiliankolb:


ekohl (approved):
I wonder if Vale could warn against this or if we need to write out own check. IMHO we should make sure CI flags this as an issue that blocks merging.

PR #4234, opened 2025-09-10
Rework Hammer reference generation

Lennonka:

Updating the `generate-hammer-reference.sh` for the new Hammer reference

- Assembly moved to a new guide
- More granular modularization
- DITA compliance
- Decreased nesting

- The subcommands are now split into multiple `ref_` modules. The generation results in over **800 modules** in addition to the existing ones! Are we sure we want them inside `common/modules`?
- The generation workflow now requires manual steps to decrease nesting of a small number of subcommands.
- I've tested it downstream thoroughly.

#### Contributor checklists

* [x] Foreman 3.16

### DIFF
--- a/guides/scripts/generate-hammer-reference.sh
+++ b/guides/scripts/generate-hammer-reference.sh
@@ -437,7 +437,7 @@ echo -e "I: [End of input file]"
 echo
 
 # Process option details
-echo -e "Processing option details..."
+echo -e "Processing data representation module..."
 sort $details_tmp | uniq | \
     sed -E 's/^([A-Z\_]+)\s+/\n\1:: /' >>$details_file
 [ -n "$DEBUG" ] || rm $details_tmp # Delete tmp file
--- a/guides/scripts/generate-hammer-reference.sh
+++ b/guides/scripts/generate-hammer-reference.sh
@@ -10,7 +10,7 @@ DEBUG=
 
 print_help() {
     echo -e "Generate AsciiDoc reference from a Hammer full-help MarkDown file"
-    echo -e "Generates the master.adoc and one module per hammer subsubcommand (3 levels)."
+    echo -e "Generates $ASMB_FILENAME and one module per hammer subsubcommand (3 levels)."
     echo
     echo -e "Usage:"
     echo -e "    $0 [OPTIONS ...] SOURCE_FILE"
--- a/guides/scripts/generate-hammer-reference.sh
+++ b/guides/scripts/generate-hammer-reference.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/bash
 
 SRC=
-MOD_DIR=common/modules/
+MOD_DIR=common/hammer-reference/
 MOD_TYPE=":_mod-docs-content-type: REFERENCE"
 ROLE_ABSTRACT='[role="_abstract"]'
 ASMB_FILENAME=doc-Hammer_Reference/master.adoc
@@ -10,7 +10,7 @@ DEBUG=
 
 print_help() {
     echo -e "Generate AsciiDoc reference from a Hammer full-help MarkDown file"
-    echo -e "  Generates one assembly and one module per hammer command."
+    echo -e "Generates the master.adoc and one module per hammer subsubcommand (3 levels)."
     echo
     echo -e "Usage:"
     echo -e "    $0 [OPTIONS ...] SOURCE_FILE"
@@ -36,7 +36,7 @@ bye() {
 
 # Process user arguments
 if [ $# -eq 0 ]; then print_help; exit 1; fi
-while [ $# -gt 1 ]; do
+while [ $# -gt 0 ]; do
     case "$1" in
         --target-dir|-T)
             MOD_DIR="$2"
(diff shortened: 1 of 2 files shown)


guides/scripts/generate-hammer-reference.sh
```diff
@@ -72,24 +71,23 @@ done
...
+asmb_header='include::common/attributes.adoc[]\ninclude::common/header.adoc[]\n:context: hammer-reference\n:hammer-reference:\n\n= {HammerRefDocTitle}\n\nifdef::satellite[]\ninclude::common/modules/proc_providing-feedback-on-red-hat-documentation.adoc[leveloffset=+1]\nendif::[]\n\n// DO NOT EDIT MANUALLY!\n// Use the generate-hammer-reference.sh script to update\n'
+asmb_footer='include::common/modules/ref_hammer-option-details.adoc[leveloffset=+1]\n\nifndef::orcharhino,satellite[]\ninclude::common/ribbons.adoc[]\nendif::[]\n'
 
-# Create folders
-{ [ -n "$NOOP" ] || mkdir -p $mod_path ; } && echo -e "Created folders: " $mod_path
+# Create folders if they don't exist
+if [ -d "$mod_path" ]; then
+    echo "I: Module path exists: $mod_path"
+else
+    { [ -n "$NOOP" ] || mkdir -p $mod_path ; } && echo -e "I: Created path: $mod_path"
```
Lennonka, line 89: Should the script fail if the folder doesn't exist?

maximiliankolb: Probably, yes. It's an indication that the user is running the script in the wrong place, isn't it?

guides/scripts/generate-hammer-reference.sh
```diff
@@ -153,9 +151,13 @@ cat "$SRC" | while read line ; do
             [ -n "$DEBUG" ] && echo -e "I: [cmd-3]: $line"
             id_core=`echo -e $line | sed 's/^### //' | sed 's/ /-/g'`
             id=[id=\"$id_core\"]
-            heading=`echo -e $line | sed 's/^### hammer/==/'`
+            filename=ref_$id_core.adoc
```
maximiliankolb, line 154: What does "core" mean? Does that mean that we have modules with two underscores?

I would prefer a second/different prefix if possible. As a maintainer, I would probably want an easy way to exclude those automatically generated files when I make tree-wide changes.

guides/scripts/generate-hammer-reference.sh
```diff
@@ -169,9 +171,13 @@ cat "$SRC" | while read line ; do
             [ -n "$DEBUG" ] && echo -e "I: [cmd-4]: $line"
             id_core=`echo -e $line | sed 's/^#### //' | sed 's/ /-/g'`
             id=[id=\"$id_core\"]
-            heading=`echo -e $line | sed 's/^#### hammer/===/'`
+            filename=ref_$id_core.adoc
```
maximiliankolb, line 174: same as above?

guides/scripts/generate-hammer-reference.sh
```diff
@@ -200,7 +206,7 @@ cat "$SRC" | while read line ; do
         "Usage:")
             [ -n "$DEBUG" ] && echo -e "I: [Usage]ID: $id_core"
             [ -n "$NOOP" ] || [ -n "$ASSEMBLY_ONLY" ] || \
-                echo -e ".Usage" >>$mod_file
+                echo -e "*Usage*\n" >>$mod_file
```
maximiliankolb, line 209: Alternative: `Usage::` but I have not seen the output yet, so please take this comment with a grain of Salt.

guides/scripts/generate-hammer-reference.sh
```diff
@@ -223,15 +229,15 @@ cat "$SRC" | while read line ; do
         "Options:")
             [ -n "$DEBUG" ] && echo -e "I: [Options]ID: $id_core"
             [ -n "$NOOP" ] || [ -n "$ASSEMBLY_ONLY" ] || \
-                echo -e ".Options" >>$mod_file
+                echo -e "*Options*\n" >>$mod_file
```
maximiliankolb, line 232: Same as above.

guides/scripts/generate-hammer-reference.sh
```diff
@@ -223,15 +229,15 @@ cat "$SRC" | while read line ; do
...
             options=yes
             skip_section=
             continue
         ;;
         *"Predefined field sets"*)
             [ -n "$DEBUG" ] && echo -e "I: [Predefined field sets]ID: $id_core"
             [ -n "$NOOP" ] || [ -n "$ASSEMBLY_ONLY" ] || \
-                echo -e ".Predefined field sets" >>$mod_file
+                echo -e "*Predefined field sets*\n" >>$mod_file
```
maximiliankolb, line 240: Same as above.

maximiliankolb: Also line 258 and 269

maximiliankolb (commented):
I left some questions & suggestions. Is this script ready so that I can try it out?

apinnick:
I don't think I am qualified to review your changes.

guides/scripts/generate-hammer-reference.sh
maximiliankolb: Oh. Didn't realize that. Then it's OK as is. Alternative suggestion to emphasise that it's the name of a variable:

```suggestion
            filename=ref_${id_core}.adoc
```

guides/scripts/generate-hammer-reference.sh
Lennonka: @apinnick suggested the same, but I hesitate using it.

The definition list doesn't make a good replacement for "informal heading" or "block title". It's supposed to define terms, formally speaking.
The blocks below a definition list item need to be connected with `+` for proper indentation in the output. This situation doesn't seem to happen in the reference at the moment. But if it did, I'm not sure the script could insert `+` at the right places with its current logic.

So technically, there's nothing stopping us from using it.

Lennonka:
@maximiliankolb Yes, you can try it out! :tada:

apinnick:
> @apinnick You have access to our internal preview of the output. You could at least ack the new formatting.

Good point! Will take a look.

apinnick:
@Lennonka Might be a dumb question but where do I find it in the [theforeman-foreman-documentation-preview-pr-4234.surge.sh](https://theforeman-foreman-documentation-preview-pr-4234.surge.sh/)? I only see _Using the Hammer CLI tool_ for the different versions.

apinnick (approved):
The generated preview looks good!

Lennonka:
> I don't know why the warning appears/what it's telling me

I think there's a logical mistake in that expression and it's a false positive. I'll try to fix that :laughing:

Lennonka:
Cherry picked:
- d57248a3e2..69993b21ec  3.16 -> 3.16

PR #4239, opened 2025-09-11
Add hint about related setting for REX concurrency

maximiliankolb:

Fixes [#38247](https://projects.theforeman.org/issues/38247) (Missing docs on how to REX batching works)

Untested!

#### Contributor checklists

* [x] Foreman 3.18

### DIFF
--- a/guides/common/modules/proc_triggering-remote-execution-jobs-in-batches.adoc
+++ b/guides/common/modules/proc_triggering-remote-execution-jobs-in-batches.adoc
@@ -1,0 +1,18 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="triggering-remote-execution-jobs-in-batches"]
+= Triggering remote execution jobs in batches
+
+[role="_abstract"]
+You can control the number of tasks your {SmartProxy} recieves at once from starting a remote execution job.
+This helps you to prevent overloading your {SmartProxies} when starting remote execution jobs that target many hosts.
+By default, the batch size is set to `100`.
+
+.Procedure
+. In the {ProjectWebUI}, navigate to *Administer* > *Settings*.
+ifdef::satellite[]
+. On the *Tasks* tab, set *Allow {SmartProxy} batch tasks* to `Yes`.
+endif::[]
+ifndef::satellite[]
+. On the *Tasks* tab, set *Allow proxy batch tasks* to `Yes`.
+endif::[]
--- a/guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
+++ b/guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
@@ -42,7 +42,7 @@ Effective user password::
 Is used if effective user is different from the ssh user.
 
 Concurrency level::
-Defines the maximum number of jobs executed at once.
+Defines the maximum number of tasks executed at once.
 This can prevent overload of system resources in a case of executing the job on a large number of hosts.
 
 Execution ordering::
--- a/guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
+++ b/guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
@@ -47,15 +47,23 @@ Concurrency level::
 Defines the maximum number of tasks executed at once.
 This can prevent overload of system resources in a case of executing the job on a large number of hosts.
 +
+A concurrency level of one is identical to running remote execution jobs sequentially.
+For example, if you run HTTP servers behind a load balancer, you can set the concurrency level to one to update or configure them one at a time to ensure that there is always a host up and running.
++
 ifdef::satellite[]
-With *Allow {SmartProxy} batch tasks*
+By default, *Allow {SmartProxy} batch tasks* is set to `Yes`.
 endif::[]
 ifndef::satellite[]
-With *Allow proxy batch tasks*
+By default, *Allow proxy batch tasks* is set to `Yes`.
 endif::[]
-set to `Yes`, you can also control the number of tasks your {SmartProxy} receives at once from starting a remote execution job.
-This helps you to prevent overloading your {SmartProxies} when starting remote execution jobs that target many hosts.
-By default, the batch size is set to `100`.
+If you have many hosts registered to {SmartProxies}, enable batching to prevent overloading your {SmartProxies}.
+ifdef::satellite[]
+By default, *{SmartProxy} tasks batch size* is set to `100`.
+endif::[]
+ifndef::satellite[]
+By default, *Proxy tasks batch size* is set to `100`.
+endif::[]
+The batch size controls the number of tasks your {SmartProxy} receives at once from starting a remote execution job.
 You can find both settings under *Administer* > *Settings* on the *Tasks* tab.
 
 Execution ordering::
--- a/guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
+++ b/guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
@@ -54,20 +54,20 @@ A concurrency level of one is identical to running remote execution jobs sequent
 For example, if you run HTTP servers behind a load balancer, you can set the concurrency level to one to update or configure them one at a time to ensure that there is always a host up and running.
 +
 A high number of concurrent tasks may overload {SmartProxies}.
+If you have many hosts registered to {SmartProxies}, batching prevents overloading your {SmartProxies}.
 ifdef::satellite[]
 By default, *Allow {SmartProxy} batch tasks* is set to `Yes`.
 endif::[]
 ifndef::satellite[]
 By default, *Allow proxy batch tasks* is set to `Yes`.
 endif::[]
-If you have many hosts registered to {SmartProxies}, batching prevents overloading your {SmartProxies}.
+You can control the number of tasks your {SmartProxy} receives at once from starting a remote execution job by setting the batch size.
 ifdef::satellite[]
 By default, *{SmartProxy} tasks batch size* is set to `100`.
 endif::[]
 ifndef::satellite[]
 By default, *Proxy tasks batch size* is set to `100`.
 endif::[]
-The batch size controls the number of tasks your {SmartProxy} receives at once from starting a remote execution job.
 You can find both settings under *Administer* > *Settings* on the *Tasks* tab.
 
 Execution ordering::
--- a/guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
+++ b/guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
@@ -47,16 +47,20 @@ Concurrency level::
 Defines the maximum number of tasks executed at once.
 This can prevent overload of system resources in a case of executing the job on a large number of hosts.
 +
+If you set the concurrency level to two or higher, remote execution jobs on hosts are executed concurrently.
+For example, you can patch multiple hosts at the same time.
++
 A concurrency level of one is identical to running remote execution jobs sequentially.
 For example, if you run HTTP servers behind a load balancer, you can set the concurrency level to one to update or configure them one at a time to ensure that there is always a host up and running.
 +
+A high number of concurrent tasks may overload {SmartProxies}.
 ifdef::satellite[]
 By default, *Allow {SmartProxy} batch tasks* is set to `Yes`.
 endif::[]
 ifndef::satellite[]
 By default, *Allow proxy batch tasks* is set to `Yes`.
 endif::[]
-If you have many hosts registered to {SmartProxies}, enable batching to prevent overloading your {SmartProxies}.
+If you have many hosts registered to {SmartProxies}, batching prevents overloading your {SmartProxies}.
 ifdef::satellite[]
 By default, *{SmartProxy} tasks batch size* is set to `100`.
 endif::[]
--- a/guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
+++ b/guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
@@ -44,8 +44,19 @@ Effective user password::
 Is used if effective user is different from the ssh user.
 
 Concurrency level::
-Defines the maximum number of jobs executed at once.
+Defines the maximum number of tasks executed at once.
 This can prevent overload of system resources in a case of executing the job on a large number of hosts.
++
+ifdef::satellite[]
+With *Allow {SmartProxy} batch tasks*
+endif::[]
+ifndef::satellite[]
+With *Allow proxy batch tasks*
+endif::[]
+set to `Yes`, you can also control the number of tasks your {SmartProxy} receives at once from starting a remote execution job.
+This helps you to prevent overloading your {SmartProxies} when starting remote execution jobs that target many hosts.
+By default, the batch size is set to `100`.
+You can find both settings under *Administer* > *Settings* on the *Tasks* tab.
 
 Execution ordering::
 Determines the order in which the job is executed on hosts.
--- a/guides/common/assembly_configuring-and-setting-up-remote-jobs.adoc
+++ b/guides/common/assembly_configuring-and-setting-up-remote-jobs.adoc
@@ -124,6 +124,8 @@ endif::[]
 
 include::modules/con_remote-execution-reference.adoc[leveloffset=+1]
 
+include::modules/proc_triggering-remote-execution-jobs-in-batches.adoc[leveloffset=+2]
+
 include::modules/ref_advanced-settings-in-the-job-wizard.adoc[leveloffset=+2]
 
 include::modules/ref_extended-cron-line-syntax-in-remote-jobs.adoc[leveloffset=+2]


guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
```diff
@@ -43,6 +43,10 @@ Is used if effective user is different from the ssh user.
 Concurrency level::
 Defines the maximum number of jobs executed at once.
 This can prevent overload of system resources in a case of executing the job on a large number of hosts.
++
+If a batch size is set, {SmartProxies} run batches of concurrent remote execution jobs.
+For Ansible-based remote execution jobs, change the `foreman_ansible_proxy_batch_size` setting in the {ProjectWebUI} under *Administer* > *Settings* on the *Ansible* tab.
```
MartinSpiessl: not quite, `foreman_ansible_proxy_batch_size` is empty by default. Then `foreman_tasks_proxy_batch_size` will be taken also for ansible-based remote execution jobs. If you set `foreman_ansible_proxy_batch_size` to some value, then this one overrides the batch size for ansible-based jobs.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
```diff
@@ -43,6 +43,10 @@ Is used if effective user is different from the ssh user.
 Concurrency level::
 Defines the maximum number of jobs executed at once.
 This can prevent overload of system resources in a case of executing the job on a large number of hosts.
++
```
MartinSpiessl: not sure if batch sizes even belong under "Concurrency level", since concurrency level effectively seems to simply ignore the batch sizes. One would need to have to look into the actual code/logics to make sense of this I guess.

MartinSpiessl (commented):
I had a look again and left a few comments. What the foreman actually does is on a different page entirely, it might greatly deviate from what we can deduce from the help text of the settings!

MartinSpiessl:
> - Is concurrency level really ignored if foreman_tasks_proxy_batch_trigger is not enabled? My tests would indicate otherwise

I think I figured this part out:

> If this is set and proxy batch triggering is enabled, then tasks are triggered on the smart proxy in batches of size 1.

 This addition is meant as a hint that concurrency level will reduce the batch size to 1 as a side effect, but it will also work if proxy batch triggering is disabled.

adamruzicka:
> Are these descriptions of how these 3 work still correct?

`proxy_batch_trigger`, `proxy_batch_size` and what is the third?

> What is the difference between sending a task to a smart proxy and triggering it?

None really

> Is concurrency level really ignored if foreman_tasks_proxy_batch_trigger is not enabled? 
> This addition is meant as a hint that concurrency level will reduce the batch size to 1 as a side effect, but it will also work if proxy batch triggering is disabled.

Correct, although this behaviour might have changed since https://github.com/theforeman/foreman_remote_execution/pull/827 . It won't be hard limited to 1, but it will be something a number from 1 to `min($hosts_count, $concurrency_level, $proxy_batch_size)`, depending on timing.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
adamruzicka: > not sure if batch sizes even belong under "Concurrency level"

I'd say they don't. Batches are a completely separate concept which just happens to operate in the same area.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
```diff
@@ -43,6 +43,10 @@ Is used if effective user is different from the ssh user.
...
+If a batch size is set, {SmartProxies} run batches of concurrent remote execution jobs.
```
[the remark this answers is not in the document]
maximiliankolb: Changed jobs to tasks.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
```diff
@@ -43,6 +43,10 @@ Is used if effective user is different from the ssh user.
...
+For Ansible-based remote execution jobs, change the `foreman_ansible_proxy_batch_size` setting in the {ProjectWebUI} under *Administer* > *Settings* on the *Ansible* tab.
+For Script-based remote execution jobs, change the `foreman_tasks_proxy_batch_size` setting in the {ProjectWebUI} under *Administer* > *Settings* on the *Tasks* tab.
```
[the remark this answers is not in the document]
maximiliankolb: This is correct based on "master" in foreman-tasks:

````
$ cat lib/foreman_tasks/engine.rb | sed -n "80,89p"
            setting('foreman_tasks_proxy_batch_trigger',
                    type: :boolean,
                    description: N_('Allow triggering tasks on the smart proxy in batches'),
                    default: true,
                    full_name: N_('Allow proxy batch tasks'))
            setting('foreman_tasks_proxy_batch_size',
                    type: :integer,
                    description: N_('Number of tasks which should be sent to the smart proxy in one request, if foreman_tasks_proxy_batch_trigger is enabled'),
                    default: 100,
                    full_name: N_('Proxy tasks batch size'))
````

However, I don't quite understand yet a) why we should mention/explain this if the default value is correct/sane and b) I still cannot reproduce the user story behind this.

**Review by Lennonka — CHANGES_REQUESTED** [review]

**Comment by maximiliankolb** [comment]
Converted to draft until I know how to continue with this. Afterwards, I'll ask for another style review.

File: guides/common/modules/proc_running-remote-execution-jobs-in-batches.adoc [hunk 5]
```diff
@@ -0,0 +1,17 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="running-remote-execution-jobs-in-batches"]
+= Running remote execution jobs in batches
+
+[role="_abstract"]
+You can run remote execution jobs in batches.
+This helps you to finetune the characteristics of remote execution jobs.
```
MartinSpiessl: My understanding is
"This helps to control in how many requests the tasks for a job are sent to the proxy"

guides/common/modules/proc_running-remote-execution-jobs-in-batches.adoc
```diff
@@ -0,0 +1,17 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="running-remote-execution-jobs-in-batches"]
+= Running remote execution jobs in batches
```
MartinSpiessl: I think the point of this option is more about triggering than the actual running/execution, i.e., sending the tasks to the proxy. See also this issue https://projects.theforeman.org/issues/25160 where this option was added to fix [Bug #24138](https://projects.theforeman.org/issues/24138):

> Blocks Foreman Remote Execution - [Bug #24138](https://projects.theforeman.org/issues/24138): When running REX at scale, capsule is not able process all the incoming requests, leading to "Initialization error: RestClient::RequestTimeout - Request Timeout" in actions

guides/common/modules/proc_running-remote-execution-jobs-in-batches.adoc
maximiliankolb: I reworded this slightly.

guides/common/modules/proc_triggering-remote-execution-jobs-in-batches.adoc
```diff
@@ -0,0 +1,18 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="triggering-remote-execution-jobs-in-batches"]
+= Triggering remote execution jobs in batches
```
Lennonka: But changing the setting does not trigger anything?
I'd prefer a title like _Enabling remote execution processing in batches_.
Because it's about splitting the processing of jobs on many hosts into batched tasks if I understand correctly.

guides/common/modules/proc_triggering-remote-execution-jobs-in-batches.adoc
```diff
@@ -0,0 +1,18 @@
...
+
+[role="_abstract"]
+You can control the number of tasks your {SmartProxy} recieves at once from starting a remote execution job.
+This helps you to prevent overloading your {SmartProxies} when starting remote execution jobs that target many hosts.
+By default, the batch size is set to `100`.
```
Lennonka: The current abstract is not related to enabling the setting. The setting we're changing has nothing to do with the batch size.

Does the setting affect only REX or other tasks as well?

Lennonka (changes requested):

guides/common/modules/proc_triggering-remote-execution-jobs-in-batches.adoc
adamruzicka: > I'd prefer a title like Enabling remote execution processing in batches.


It is enabled by default, but "Disabling remote execution ..." doesn't sound all that good

Lennonka: Then why are we adding a procedure to enable it, when it's enabled by default?
Why would one disable it?
What are we actually trying to accomplish here? :exploding_head: 
Perhaps it should be a concept module or something instead of a procedure?

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
```diff
@@ -42,7 +42,7 @@ Effective user password::
 Is used if effective user is different from the ssh user.
 
 Concurrency level::
-Defines the maximum number of jobs executed at once.
+Defines the maximum number of tasks executed at once.
```
Lennonka, line 47: @maximiliankolb I would see it as a more detailed explanation under this definition list item instead of a procedure. Perhaps describe how batching works for the script and Ansible providers separately? No idea how this works tho.

guides/common/modules/proc_triggering-remote-execution-jobs-in-batches.adoc
```diff
@@ -0,0 +1,18 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="triggering-remote-execution-jobs-in-batches"]
+= Triggering remote execution jobs in batches
+
+[role="_abstract"]
+You can control the number of tasks your {SmartProxy} recieves at once from starting a remote execution job.
```
MartinSpiessl: typo: recieves -> receives

guides/common/modules/proc_running-remote-execution-jobs-in-batches.adoc
MartinSpiessl: lgtm

guides/common/modules/proc_running-remote-execution-jobs-in-batches.adoc
MartinSpiessl: Wording was changes from "running" to "triggering" => fine for me

MartinSpiessl (approved):
Looks good from my side

guides/common/modules/proc_triggering-remote-execution-jobs-in-batches.adoc
maximiliankolb: Fixed.

guides/common/modules/proc_triggering-remote-execution-jobs-in-batches.adoc
maximiliankolb: Good call; moved to ref module.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
maximiliankolb: I extended the description of this setting. What do you think now?

maximiliankolb (commented):
>  What are we actually trying to accomplish here? 🤯

I wish I knew!

Reworked the PR. 🙃

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
Lennonka: That looks better, but I'm missing a sentence that would relate the concurrency level to the batch size. At the moment, it isn't clear why we're mentioning the setting and the batch size under the concurrency level advanced REX setting.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
```diff
@@ -42,8 +42,18 @@ Effective user password::
...
 This can prevent overload of system resources in a case of executing the job on a large number of hosts.
++
+ifdef::satellite[]
+With *Allow {SmartProxy} batch tasks*
```
Lennonka: What is this (a global setting) and where do I find it?

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
```diff
@@ -42,8 +42,18 @@ Effective user password::
...
+endif::[]
+ifndef::satellite[]
+With *Allow proxy batch tasks*
+endif::[]
+set to `Yes`, you can control the number of tasks your {SmartProxy} receives at once from starting a remote execution job.
+This helps you to prevent overloading your {SmartProxies} when starting remote execution jobs that target many hosts.
+By default, the batch size is set to `100`.
```
Lennonka: How do I change the batch size? Is there another global setting for it?

I've found **Capsule tasks batch size** that has 100 as the default.

Lennonka (changes requested):

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
maximiliankolb: It's a global setting. You can find it under Administer > Settings > Tasks.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
maximiliankolb: Yes, there's another setting; "Proxy tasks batch size" in upstream.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
maximiliankolb: I honestly wish I could tell you. I am somehow trying since end of Febuary 2025...

To my understanding, concurrency level means "how many jobs run on hosts in parallel". The batch size is about how many jobs Foreman sends to Smart Proxy at once.

For now, I added an "also" to line 54.

maximiliankolb (commented):
Am I allowed to describe how to set the setting in the reference module?

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
Lennonka: Don't tell me, tell the users :laughing:

Lennonka:
> Am I allowed to describe how to set the setting in the reference module?

You don't have to write steps, you can just say something like "You can find it under Administer > Settings > Tasks."

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
maximiliankolb: Added.

maximiliankolb (commented):
Mentioned place of settings.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
Lennonka: You may want to revisit the linked Redmine issue and ask for clarifications in the Matrix dev channel :)

m-bucher: If I understand the code correctly, then concurrency level is a parameter for Dynflow and decides how many actual Tasks can run in parallel (Task execution-phase). 'run in parallel' refers to ForemanTask. So if you start a RemoteJob on 100 Hosts, this will result in 100 ForemanTasks. If the concurrency is set to 10, then Dynflow ensures that only 10 of them run at the same time.

While the batch_size specifies how many many Tasks are scheduled at once in foreman-tasks (Task creation-phase). The previously mentioned 100 RemoteJobs will result in 100 ForemanTasks, but these tasks are created in ForemanTask in batches of 10 (assuming the batch_size is set to 10).
If no concurrency level is set these tasks however will start running as soon as they are created successfully.

I did not fully understand whether concurrency-level setting will also schedule the Tasks in batches the same size as the concurrency level. That might be for @adamruzicka to answer :wink: 
If that is the case: `This also limits the batch-size to the specified value (see 'Allow proxy batch tasks').`

If not: `In contrast to the batch-size described at 'Allow proxy batch tasks', this only limits how many of the created tasks will be executed at the same time.`

maximiliankolb: FBP: example: webservers behind a LB: you update/configure one host at a time if you set this to one. then it runs sequentially and you always have a host up and running.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
```diff
@@ -42,8 +42,19 @@ Effective user password::
...
 Concurrency level::
-Defines the maximum number of jobs executed at once.
+Defines the maximum number of tasks executed at once.
 This can prevent overload of system resources in a case of executing the job on a large number of hosts.
++
+ifdef::satellite[]
+With *Allow {SmartProxy} batch tasks*
+endif::[]
+ifndef::satellite[]
+With *Allow proxy batch tasks*
```
maximiliankolb: FBP: batching was invented so that tasks are not sent invidivudally to avoid smart proxy dropping tasks.

maximiliankolb: higher batches = slower as a whole; user perceives a bigger latency

maximiliankolb: FBP: if you have a lot of hosts registersted to a SMart Proxy, you need batching. else, your SP just dies.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
```diff
@@ -42,8 +42,19 @@ Effective user password::
...
+endif::[]
+set to `Yes`, you can also control the number of tasks your {SmartProxy} receives at once from starting a remote execution job.
+This helps you to prevent overloading your {SmartProxies} when starting remote execution jobs that target many hosts.
```
maximiliankolb: fbp: drop line.

maximiliankolb (commented):
input from Adam and Markus.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
maximiliankolb: Applied.

maximiliankolb (commented):
Rebased to HEAD of "master" and applied feedback by Adam and Markus from the Foreman Birthday Party.

Lennonka:
> Rebased to HEAD of "master" and applied feedback by Adam and Markus from the Foreman Birthday Party.

That must have been a hell of a party :eyes: :grin:

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
```diff
@@ -42,8 +42,27 @@ Effective user password::
 Is used if effective user is different from the ssh user.
 
 Concurrency level::
-Defines the maximum number of jobs executed at once.
+Defines the maximum number of tasks executed at once.
 This can prevent overload of system resources in a case of executing the job on a large number of hosts.
++
+A concurrency level of one is identical to running remote execution jobs sequentially.
+For example, if you run HTTP servers behind a load balancer, you can set the concurrency level to one to update or configure them one at a time to ensure that there is always a host up and running.
```
Lennonka, line 54: What does it mean if I set the concurrency level of more than one? Do those tasks get distributed to several SmartProxy servers then? Or do they get distributed in another way like to CPU cores?

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
```diff
@@ -42,8 +42,27 @@ Effective user password::
...
++
+ifdef::satellite[]
+By default, *Allow {SmartProxy} batch tasks* is set to `Yes`.
+endif::[]
+ifndef::satellite[]
+By default, *Allow proxy batch tasks* is set to `Yes`.
+endif::[]
+If you have many hosts registered to {SmartProxies}, enable batching to prevent overloading your {SmartProxies}.
```
Lennonka: What does the setting mean? How does it relate to the concurrency level? What's the point of enabling it when it's enabled by default? Why shouldn't I disable it?

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
```diff
@@ -42,8 +42,27 @@ Effective user password::
 Is used if effective user is different from the ssh user.
 
 Concurrency level::
-Defines the maximum number of jobs executed at once.
+Defines the maximum number of tasks executed at once.
 This can prevent overload of system resources in a case of executing the job on a large number of hosts.
++
+A concurrency level of one is identical to running remote execution jobs sequentially.
+For example, if you run HTTP servers behind a load balancer, you can set the concurrency level to one to update or configure them one at a time to ensure that there is always a host up and running.
++
```
Lennonka, line 55: Still missing an intro sentence on how these settings relate to the concurrency level. Why are they relevant?

Lennonka (changes requested):
Thank you. We're getting closer, but we're not quite there yet.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
adamruzicka: When you run a remote execution job, you will get a single JobInvocation object, commonly referred to as "job". This object will spawn a single parent task, which then proceeds to spawn the individual per-host tasks so that every host within the job gets its own task.

Concurrency level, works strictly "inside Foreman", without involving any proxies or whatnot. It controls the way how the parent task spawns the per-host tasks to ensure that less than $concurrency_level per-host tasks exist and are "running" (even though the more accurate term would be "not finished") at any given time.

> What does it mean if I set the concurrency level of more than one?

Let's say you set it to 4. In the beginning the parent will spawn 4 per-host tasks. Those will run until completion. The parent periodically checks if there is any capacity left within the configured concurrency level. If 1 of the per-host tasks finished in the meantime, the parent will spawn 1 more. If 3 finished, the parent will spawn 3 and so on.

> Do those tasks get distributed to several SmartProxy servers then?

Eventually, maybe. The concurrency level has no effect on this. Concurrency level only works between the parent task and the per-host tasks. The whole proxy-delegation business happens later, but only between the per-host tasks and the smart-proxies.

> Or do they get distributed in another way like to CPU cores?

Let's not go to such low levels of the stack.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
adamruzicka: Historically, each per-host task made its own api call to the smart proxy to perform the delegation. This didn't really scale, as the proxy would eventually start discarding the incoming requests. To work around this, we introduced batching. Instead of letting each per-host task to talk to the proxy directly, we batch several of them and send them to the proxy in a single, bigger request.

> How does it relate to the concurrency level?

It doesn't directly relate to it, but it is a part of the same pipeline. It is just that a parent task gets spawned, spawns per-host tasks in a way so that they fit within the limit and the spawned per-host tasks then go through batching.

> Why shouldn't I disable it?

Because you'd invite back the original scale problem. Which is sort of fine if you have let's say less than 100 hosts. The defaults are sane enough to work reasonably well no matter your scale, but if you really really want the behaviour we had around foreman 2.4, you can disable it.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
maximiliankolb: I added an explanation about concurrency level of two or higher.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
maximiliankolb: Added an intro sentence.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
maximiliankolb: The reference modules tries to explain the setting.

As explained by Adam, the default values are good for most scenarios. However, is some cases, users might want to tweak those settings.

Reworded to emphasise that it's enabled and its effect; not that users can enable it.

maximiliankolb (commented):
Rebased to HEAD of "master" and applied all feedback.

sbernhard (approved):
LGTM

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
```diff
@@ -44,8 +44,31 @@ Effective user password::
...
+endif::[]
+If you have many hosts registered to {SmartProxies}, batching prevents overloading your {SmartProxies}.
+ifdef::satellite[]
+By default, *{SmartProxy} tasks batch size* is set to `100`.
+endif::[]
+ifndef::satellite[]
+By default, *Proxy tasks batch size* is set to `100`.
+endif::[]
+The batch size controls the number of tasks your {SmartProxy} receives at once from starting a remote execution job.
+You can find both settings under *Administer* > *Settings* on the *Tasks* tab.
```
Lennonka, line 71:
```suggestion
A high number of concurrent tasks may overload {SmartProxies}.
If you have many hosts registered to {SmartProxies}, batching prevents overloading your {SmartProxies}.
ifdef::satellite[]
By default, *Allow {SmartProxy} batch tasks* is set to `Yes`.
endif::[]
ifndef::satellite[]
By default, *Allow proxy batch tasks* is set to `Yes`.
endif::[]
You can control the number of tasks your {SmartProxy} receives at once from starting a remote execution job by setting the batch size.
ifdef::satellite[]
By default, *{SmartProxy} tasks batch size* is set to `100`.
endif::[]
ifndef::satellite[]
By default, *Proxy tasks batch size* is set to `100`.
endif::[]
You can find both settings under *Administer* > *Settings* on the *Tasks* tab.
```

Lennonka (commented):
Yes, yes! Very nice!
Just a small suggestion for the text flow.

Lennonka:
Btw, I'm wondering if we should put it into a separate reference module and cross reference it instead. But this is optional for this PR and if you feel it would be more suitable, you can do it as a follow up.

guides/common/modules/ref_advanced-settings-in-the-job-wizard.adoc
maximiliankolb: Applied.

maximiliankolb (commented):
Thanks @Lennonka I applied your suggestion. Ready for final ACK.

maximiliankolb:
> Btw, I'm wondering if we should put it into a separate reference module and cross reference it instead. But this is optional for this PR and if you feel it would be more suitable, you can do it as a follow up.

I would rather not do this right now; possibly ever.

Lennonka (approved):
That's okay, I'm quite happy with it as it is. You've done a great job!

maximiliankolb:
Thanks everyone for your help!

   92e85638e5..874593fbdc  3.19 -> 3.19
   632fad00d6..684fc4b6d2  3.18 -> 3.18

maximiliankolb:
cherry-pick to "5.0":    1fa4ed2fd6..133785c098  5.0 -> 5.0

PR #4296, opened 2025-09-24
Remove note about cloning and editing templates

### LINKED ISSUE #4256: Recommend custom pre/custom post snippets in favor of cloning+editing templates
**Affected files**: `guides/common/modules/proc_booting-virtual-machines.adoc` and `guides/common/modules/proc_chainbooting-ipxe-from-pxelinux.adoc`
**Task**: Reword the note about cloning and editing templates.
**Goal**: Users should always use `custom pre` and `custom post` snippets to run custom code during host registration/host provisioning.

**nightly docs**: https://docs.theforeman.org/nightly/Provisioning_Hosts/index-katello.html#running-custom-code-during-host-provisioning which explains the exact name the snippets must have to get picked up when Foreman rendered the templates.

maximiliankolb:

Users should never clone and edit templates. Instead, if they really need to run custom code before or after registration or provisioning workflows, they should use "custom pre" and "custom post" snippets as documented. This way they do not lose support by the community/vendor.

Also, there's configuration management.

Fixes #4256

Should we also drop the hint about cloning and editing in `guides/common/modules/proc_setting-up-job-templates.adoc`?
#### Contributor checklists

* [x] Foreman 3.16/Katello 4.18 (Satellite 6.18)

### DIFF
--- a/guides/common/modules/proc_booting-virtual-machines.adoc
+++ b/guides/common/modules/proc_booting-virtual-machines.adoc
@@ -18,13 +18,6 @@ endif::[]
 * You have configured your iPXE environment.
 For more information, see xref:Configuring_iPXE_Environment_{context}[].
 
-[NOTE]
-====
-You can use the original templates shipped in {Project} as described below.
-If you require modification to an original template, clone the template, edit the clone, and associate the clone instead of the original template.
-For more information, see xref:running-custom-code-during-host-provisioning-by-using-web-ui[].
-====
-
 .Procedure
 . In the {ProjectWebUI}, navigate to *Hosts* > *Templates* > *Provisioning Templates*.
 ifdef::satellite[]
@@ -60,3 +53,6 @@ endif::[]
 . Select the *Templates* tab.
 . In *Provisioning Templates*, click *Resolve* and verify that the *iPXE template* resolves to the required template.
 . Click *Submit* to save host settings.
+
+.Additional resources
+* xref:running-custom-code-during-host-provisioning-by-using-web-ui[]
--- a/guides/common/modules/proc_chainbooting-ipxe-from-pxelinux.adoc
+++ b/guides/common/modules/proc_chainbooting-ipxe-from-pxelinux.adoc
@@ -15,13 +15,6 @@ To use HTTP with iPXE, use the iPXE build with built-in drivers (`ipxe.lkrn`).
 * You have configured your iPXE environment.
 For more information, see xref:Configuring_iPXE_Environment_{context}[].
 
-[NOTE]
-====
-You can use the original templates shipped in {Project} as described below.
-If you require modification to an original template, clone the template, edit the clone, and associate the clone instead of the original template.
-For more information, see xref:running-custom-code-during-host-provisioning-by-using-web-ui[].
-====
-
 .Procedure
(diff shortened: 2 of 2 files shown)


guides/common/modules/proc_chainbooting-ipxe-from-pxelinux.adoc
```diff
@@ -15,13 +15,6 @@ To use HTTP with iPXE, use the iPXE build with built-in drivers (`ipxe.lkrn`).
 * You have configured your iPXE environment.
 For more information, see xref:Configuring_iPXE_Environment_{context}[].
 
-[NOTE]
-====
-You can use the original templates shipped in {Project} as described below.
-If you require modification to an original template, clone the template, edit the clone, and associate the clone instead of the original template.
-For more information, see xref:running-custom-code-during-host-provisioning-by-using-web-ui[].
```
Gauravtalreja1: I'm not sure if we should recommend users to not clone provisioning template, since those are locked by default and how users would add pre/post snippets to it without cloning?

guides/common/modules/proc_booting-virtual-machines.adoc
```diff
@@ -18,13 +18,6 @@ endif::[]
 * You have configured your iPXE environment.
 For more information, see xref:Configuring_iPXE_Environment_{context}[].
 
-[NOTE]
-====
-You can use the original templates shipped in {Project} as described below.
-If you require modification to an original template, clone the template, edit the clone, and associate the clone instead of the original template.
-For more information, see xref:running-custom-code-during-host-provisioning-by-using-web-ui[].
```
Gauravtalreja1: I think this isn't a correct link for the clone guide, which should be either https://docs.theforeman.org/nightly/Provisioning_Hosts/index-katello.html#cloning-provisioning-templates, right?
cc @stejskalleos @ShimShtein

guides/common/modules/proc_chainbooting-ipxe-from-pxelinux.adoc
Lennonka: As Max is saying, the custom pre/post snippets do not have to be added to templates because the code is there by default ([example](https://github.com/theforeman/foreman/blob/2e415fc3b02486fe53f5bcacbfcc8c90760576ad/app/views/unattended/provisioning_templates/provision/kickstart_default.erb#L222)). If the snippets don't exist, nothing happens. Once the user creates them, the templates pick them up automatically.

Therefore, the user doesn't have to clone and edit templates to add the snippets to the provisioning process.

guides/common/modules/proc_booting-virtual-machines.adoc
Lennonka: Why is this relevant when we're removing it?

guides/common/modules/proc_chainbooting-ipxe-from-pxelinux.adoc
stejskalleos: > If the snippets don't exist, nothing happens. Once the user creates them, the templates pick them up automatically.

Correct.

> Therefore, the user doesn't have to clone and edit templates to add the snippets to the provisioning process.

Also correct.

PR #4221, opened 2025-09-08
Add [role] element to introductions in user auth, puppet, and lb

aneta-petrova:
#### What changes are you introducing?

* Adding a role element to introductions in the guides for configuring user authentication, managing puppet, and load balancing.
* Adding an introduction to modules that currently don't have it.
* Make sure the abstracts are complete paragraphs (for example, that they don't end with a colon)
* Turn existing prerequisite and additional resources reference modules into snippets (based on the solution suggested in https://github.com/jhradilek/asciidoctor-dita-vale/issues/126)

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

The goal is to annotate the first paragraph of any module as an "abstract", which is supposed to be a short introduction summarizing the contents of the module. (https://github.com/jhradilek/asciidoctor-dita-vale/issues/110#issuecomment-3253175466)

The goal is not to ensure all abstracts are perfect. Only that the content is migrateable.

Follow up on https://github.com/theforeman/foreman-documentation/pull/4189, https://github.com/theforeman/foreman-documentation/pull/4178, and https://github.com/theforeman/foreman-documentation/pull/4175.

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)

I used [this](https://github.com/apinnick/scripts/blob/main/extract-doc-modules.sh) to extract the modules for each of the guides and this to add the element:

```
for file in *.adoc; do sed -i '/^= .*/{n;a\[role="_abstract"]
}' "$file"; done
```

The paragraphs annotated as "abstract" don't need to be perfect abstracts at this point. That's planned for later.

#### Contributor checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.16/Katello 4.18 (Satellite 6.18)
* [ ] Foreman 3.15/Katello 4.17
* [ ] Foreman 3.14/Katello 4.16 (Satellite 6.17; orcharhino 7.4)
* [ ] Foreman 3.13/Katello 4.15 (EL9 only)
* [ ] Foreman 3.12/Katello 4.14 (Satellite 6.16; orcharhino 7.2 on EL9 only; orcharhino 7.3)
[...]

PR #4315, opened 2025-09-30
Add style definition for abstracts

aneta-petrova:

Adding a CSS style definition for paragraphs annotated with [role=_abstract].

In order to address the requirements for Red Hat's downstream DITA pre-migration checks (https://github.com/jhradilek/asciidoctor-dita-vale/), we are now required to add abstract metadata to introductory paragraphs (http://github.com/theforeman/foreman-documentation/pull/4221). I thought it might be nice to use this metadata to visually separate abstract paragraphs on the Foreman side.

Screenshot From 2025-09-30 20-13-19

IMO this sort of visual separation improves readability (users will immediately know which part of a module is an abstract and where the actual fun begins). But also, seeing the abstract highlighted like this in preview might help us all write abstract as summaries (authors and reviewers will immediately see that an abstract is something else than the rest of the introduction).

For transparency: The CSS definition was generated by Gemini and I only made minor tweaks to it. (The prompt I used was to make the abstracts look similar to standard AsciiDoc [INFO] blocks.) Feel free to suggest different CSS attributes if you'd prefer something else.

#### Contributor checklists

### DIFF
--- a/web/content/css/_site.scss
+++ b/web/content/css/_site.scss
@@ -13,3 +13,11 @@
     margin-top: 70px!important;
   }
 }
+
+._abstract {
+  background-color: #f7f7f7;
+  border-left: 4px solid #025d8c;
+  padding: 15px;
+  margin-bottom: 1.25em;
+  border-radius: 4px;
+}


Lennonka:
That's a nice thought! However, I think that the suggested style is a bit too much.

- Do we want to draw readers' attention to abstract or to the "actual fun part"?
If it's the latter, then we should perhaps decrease the significance of abstracts instead of increasing it visually.
- Increased padding in comparison to normal paragraphs looks typographically bad.

Let me know what you think. I'm trying to figure out an alternative style, but let's think it through first.

web/content/css/_site.scss
```diff
@@ -13,3 +13,11 @@
     margin-top: 70px!important;
   }
 }
+
+._abstract {
+  background-color: #f7f7f7;
+  border-left: 4px solid #025d8c;
+  padding: 15px;
+  margin-bottom: 1.25em;
+  border-radius: 4px;
```
Lennonka, line 22: I would go with something very simple like decreasing the intensity of the font color:

```suggestion
  color: #555555;
```

Lennonka: However, I didn't check it out in a preview. :)

Lennonka:
> I actually prefer the first one with the background because I think the clearer visual separation actually helps users focus on the "fun" part of the module content. At the risk of overthinking this, just like the well-known [Ebbinghaus illusion](https://www.labvanced.com/content/research/en/blog/2022-03-ebbinghaus/) shows that human perception tends to focus on a smaller object, when looking at the preview, my brain at least tends to focus on the non-highlighted ("smaller") part of a give section.

My brain focuses on the bar on the left of the abstract paragraph...

aneta-petrova:
I look for existing examples elsewhere and found something I liked on GH docs: https://docs.github.com/en/migrations/importing-source-code/using-the-command-line-to-import-source-code/adding-locally-hosted-code-to-github?platform=linux  This uses larger font and a visually distinct color.

I'm not sure about the larger font for our use case but I liked the more visually distinct color:

Screenshot From 2025-10-01 17-44-33

For me, this meets the requirement for clear separation but there is no border. Would that work @Lennonka?

aneta-petrova:
About https://github.com/theforeman/foreman-documentation/pull/4318: I tried a border on the bottom myself at one point (https://github.com/theforeman/foreman-documentation/pull/4315#issuecomment-3356989600). To me, that's something that basically splits the section into two and personally I'm not a fan.

aneta-petrova:
Thanks @Lennonka! This was a great brainstorming session! :rocket: 

Let's keep this open for additional reviews. I think that if there is at least one more ack, that should be good enough. If on the other hand there are further concerns or suggestions, we can keep working on this.

maximiliankolb (approved):
I have no strong opinion about this. I personally would have prefered the small blue stripe on the side, but I'm also OK as is.

aneta-petrova:
I'm not comfortable merging without a more decisive preference for one of the options so I'll list the options that seem most popular and ask for more feedback on Matrix.

The question is: Which style do we want to use for highlighting abstracts in our guides?

Option nr 1 -- blue stripe and grey background:

01-abstract

Option nr 2 -- blue stripe:

02-abstract

Option nr 3 -- red text color:

04-abstract

maximiliankolb:
:two:

ekohl:
I'm also leaning to option 2.

Lennonka:
Good luck.

PR #4472, opened 2025-11-13
Add a Vale rule to report usage of downstream branded terms

aneta-petrova:

* Introducing a Vale rule to check for downstream branded terminology and suggest an attribute as a replacement
* Dropping the related table from the contributing guidelines

To automate the reviews of these errors and to make the contributing guidelines a little bit shorter.

The rule reports only downstream terms by design. In some cases, using "Foreman" is actually appropriate. Additionally, most of the cases I've seen as a reviewer were about PR authors (myself included) accidentally using "Satellite", which is what the rule is designed to report.

When running the check on the whole repo, the rule currently reports a few cases of "Satellite" being used, for example https://github.com/theforeman/foreman-documentation/blob/92fbccf15c626ce3294f69cb882adb99b1158c6d/guides/common/modules/proc_migrating-a-host-from-red-hat-network.adoc?plain=1#L4. I don't think this is a problem.

#### Contributor checklists

## Summary by Sourcery

Enforce use of AsciiDoc attributes instead of branded downstream terms by adding a Vale substitution rule and simplifying the contributing guidelines to remove the detailed terminology table.

New Features:
- Add a Vale rule to report downstream branded terms and suggest corresponding AsciiDoc attribute replacements

Documentation:
- Remove the detailed branded terminology table from CONTRIBUTING.md and reference the `{Project}` attribute instead

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -91,15 +91,7 @@ For inclusion in the content, export diagrams to SVG and place them as described
 ### AsciiDoc attributes for different build targets
 
 Because the content in this repository is shared between the upstream Foreman community and branded downstream products, many terms need to be written using AsciiDoc attributes.
-For example, never write "Foreman", "Satellite", or "orcharhino" words directly, but use the following attributes:
-
-| Attribute | Upstream value | Downstream by Red Hat | Downstream by ATIX |
-| -------- | -------------- | --------------------- | ------------------ |
-| {Project} | Foreman | Satellite | orcharhino |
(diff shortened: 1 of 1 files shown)


maximiliankolb (approved):
That's a great idea! I don't know how to run this locally yet, but I assume that you've tested it and it works.

> When running the check on the whole repo, the rule currently reports a few cases...

This is OK for me because we do not enforce a green Vale check yet. If we did, we might have to extend this rule, add some exceptions, and fix the issues. Not a blocker in my eyes.

Thanks Anet!

PR #4473, opened 2025-11-13
Add a Vale rule to report capitalization issues

aneta-petrova:

* Introducing a Vale rule to check for proper capitalization of terms specific to this project
* Dropping the list of capitalization rules from CONTRIBUTING.md

To automate the reviews of these errors and to make the contributing guidelines a little bit shorter.

There are three capitalization rules which do not have an equivalent in Vale: Deb (as content type), Yum (as content type), and Red Hat Network. I think we can live without these.

#### Contributor checklists

## Summary by Sourcery

Enforce project-specific term capitalization via a new Vale rule and streamline contributing guidelines

New Features:
- Add a Vale substitution rule to enforce proper capitalization of project-specific terms

Enhancements:
- Add Capitalization.yml in .vale/styles/foreman-documentation to define substitution rules for terms

Documentation:
- Remove the manual capitalization list from CONTRIBUTING.md

### DIFF
--- a/.vale/styles/foreman-documentation/Capitalization.yml
+++ b/.vale/styles/foreman-documentation/Capitalization.yml
@@ -3,7 +3,7 @@
 extends: substitution
 ignorecase: false
 level: error
-message: "Use '%s' instead of '%s' unless you are using the term in a context that requires different capitalization."
+message: "Use '%s' instead of '%s' unless you are using the term in a context that requires different capitalization. For example, different capitalization might be required in Web UI buttons, settings, or templates."
 action:
   name: replace
 swap:


.vale/styles/foreman-documentation/Capitalization.yml
```diff
@@ -0,0 +1,30 @@
+# Suggest proper capitalization
+---
+extends: substitution
+ignorecase: false
+level: error
+message: "Use '%s' instead of '%s' unless you are using the term in a context that requires different capitalization."
```
maximiliankolb: context means if you describe the Foreman WebUI such as buttons, settings, or templates.

Your call if you want to add this.

maximiliankolb (commented):
Also a great idea to automate this! Thanks Anet. I left one small suggestion and I did not test it myself. But LGTM overall.

> There are three capitalization rules which do not have an equivalent in Vale: Deb (as content type), Yum (as content type), and Red Hat Network. I think we can live without these.

Can we add them? (not blocking)

.vale/styles/foreman-documentation/Capitalization.yml
```diff
@@ -0,0 +1,31 @@
+# Suggest proper capitalization
+---
+extends: substitution
+ignorecase: false
+level: error
+message: "Use '%s' instead of '%s' unless you are using the term in a context that requires different capitalization. For example, different capitalization might be required in Web UI buttons, settings, or templates."
+action:
+  name: replace
+swap:
```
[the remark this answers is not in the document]
Lennonka: Cool. I thought it might help to change the order, but that doesn't seem to be the problem.
It just reports false positives for "Subscription" in "Subscription Manager" but I don't think we can work around it.
Sigh. Nevermind :)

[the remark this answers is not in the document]
maximiliankolb: Rename "Subscription Manager" to "`subscription-manager`" if possible?

.vale/styles/foreman-documentation/Capitalization.yml
```diff
@@ -0,0 +1,31 @@
+# Suggest proper capitalization
+---
+extends: substitution
```
Lennonka: Actually, extending the `substitution` check might not be the best solution.
There's a check for capitalization specifically: https://vale.sh/docs/checks/capitalization

It has exceptions :grin:

Lennonka: I don't know. The scoping might be problematic.

Just a thought.

apinnick:
I think it would be easier and more efficient to add terms to the foreman/vale 'accept' list than to change the capitalization rule. If you change the rule here, I suspect it would be overwritten because Vale normally syncs its images.

PR #4506, opened 2025-11-28
Add new build target for foremanctl

aneta-petrova:

* Adding a new `foremanctl` flavor for build targets `katello`, `orcharhino`, and `satellite`
* Adding the `foremanct-katello` build target to web navigation. The list of guides on the website now includes Quickstart and skeleton previews of two high-impact guides: server installation and upgrading (both empty, to be filled with foremanctl content later)
* Updating the Quickstart guide to drop tabs and leave only the foremanctl content in place (for foremanctl build targets) and old installer content (for the non-foremanctl build targets) (goodbye tabs :cry:)

To prepare the underlying documentation structure for developing `foremanctl` documentation in parallel to documentation based on the existing installer.

The expected future state when `foremanctl` becomes more mature is described in https://github.com/theforeman/foreman-documentation/pull/4506#discussion_r2589844939  This documentation PR is developed to support that future, while also allowing downstream teams to continuously review the state of their foremanctl-powered documentation sets.

Up until now, foremanctl documentation has been developed with the expectation that with a future version, Foreman will switch from using the current Puppet-based installer to the new containerized deployment tool named foremanctl. This is why in a previous PR https://github.com/theforeman/foreman-documentation/pull/3975, a new guide directory was created, with the expectation that it would at a certain point replace the existing guide.

Now it's become clear that the switch to foremanctl will be more gradual upstream and for some time, foremanctl will exists alongside non-foremanctl builds. This PR was raised to investigate an alternative approach where a new build target is created instead of a new separate guide.

Alternative approaches that have been considered:

* Developing foremanctl docs on a long-lived branch that would be continuously updated: This has the disadvantage of not being able to publish both foremanctl-based and foremanctl-free docs in parallel.
* Creating separate guides as in https://github.com/theforeman/foreman-documentation/pull/4087: This doesn't work well for parallel publishing and might require us to make changes to all guides twice, once in the foremanctl-based guide and then also in the foremanctl-free guide.
* Implementing tabs as in https://docs.theforeman.org/3.16/Quickstart/index-katello.html: This would not work well for situation where whole sections will disappear with foremanctl (for example, https://github.com/theforeman/foreman-documentation/blob/master/guides/common/modules/proc_starting-and-stopping-server.adoc might go away completely and also the installation guide's structure will change significantly)

#### Contributor checklists

## Summary by Sourcery

Introduce foremanctl-specific build targets and documentation structure alongside existing installer-based guides.

New Features:
- Add foremanctl build targets for katello, orcharhino, and satellite guides and include the foremanctl-katello variant in the web navigation.

Enhancements:
- Restructure documentation to support foremanctl variants via new attribute and repository snippet files instead of a separate Installing Server guide for foremanctl.

Build:
- Extend the top-level html make target and deployment workflow to build the new foremanctl-* documentation variants.

CI:
- Update the deploy GitHub Actions workflow to generate and publish the new foremanctl-based builds.

Documentation:
- Update README and web index to document and surface the new foremanctl build variants in the published site, and remove the standalone foremanctl Installing Server guide in favor of variant-based docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ See the [README in the `guides/` subdirectory](guides/README.md) for more inform
 
 To build both the static site and the guides for easy local testing, a global `Makefile` is provided in the root directory with the following targets:
 
-* `html`: builds HTML guides with all contexts (`foreman-el`, `foreman-deb`, `foremanctl-katello`, `foremanctl-orcharhino`, `foremanctl-satellite`, `katello`, `satellite`, and `orcharhino`)
+* `html`: builds HTML guides with all contexts (`foreman-el`, `foreman-deb`, `foremanctl-katello`, `foremanctl-orcharhino`, `foremanctl-satellite`, `katello`, `orcharhino`, and `satellite`)
 * `web`: builds static site using the `nanoc` tool
 * `compile`: compiles all content into a single directory `./result`
 * `serve`: serves the result directory via a python web server (the default target)
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ clean:
 	$(MAKE) -C guides/ clean
 	rm -rf $(DEST) web/output/
 
-html: build-foreman-el build-foreman-deb build-foremanctl-katello build-katello
+html: build-foreman-el build-foreman-deb build-foremanctl-katello build-foremanctl-orcharhino build-foremanctl-satellite build-katello build-orcharhino build-satellite
 
 build-%: FORCE prep
 	$(MAKE) -C guides/ html BUILD=$*


web/releases/nightly.json
```diff
@@ -129,6 +129,28 @@
           ["Hammer_Reference", "Hammer reference"]
         ]
       }
+    },
+    {
+      "title": "Containerized Foreman",
```
aneta-petrova: This is almost certainly not a good name for the new build target and I'm open to suggestions.

guides/common/attributes.adoc
```diff
@@ -37,6 +38,9 @@ ifeval::["{build}" == "foreman-deb"]
 :foreman-deb:
 :foreman:
 endif::[]
+ifeval::["{build}" == "foremanctl"]
+:foremanctl:
```
ekohl: Now that I see this I'm wondering: initially we discussed a more [feature flag](https://en.wikipedia.org/wiki/Feature_toggle) idea. My idea was that we can modify the `satellite` flavor to also define `foremanctl`. Seeing this here makes me worried we'll break stuff. On the other hand, perhaps that's unfounded.

Testing that idea might require some real content to be in place.

ekohl (commented):
This indeed looks very much like what we discussed and makes it easy for others to see what we're talking about. I'm positively surprised how small the whole diff really is.

guides/common/attributes.adoc
ekohl: We can add `:foremanctl:` to `attributes-satellite.adoc` to "flip" the feature on but that can have unintended consequences. That's what I had in my head as a theory and I'd love to come up with a way to test that theory though I doubt we can without filling up the `foremanctl` guide first.

ekohl: > Wouldn't that effectively make foremanctl docs development downstream first?

Maybe I wasn't clear. I was saying that we start building out the `foremanctl` flavor (build). Once downstream is ready, we "flip the switch" and **also** add the `foremanctl` attribute to the `satellite` flavor. And now I'm asking (out loud) if that is feasible. That doesn't mean the `foremanctl` flavor goes away in upstream because that's still a valid target. It just means that the `satellite` flavor will then be closer to the `foremanctl` flavor than the `katello` flavor.

aneta-petrova: I suspect previewing the changes might be trickier with what you're suggesting. Also extending the single `foremanctl` build target from being satellite-compliant to being fit for any other build target.

In the meantime https://github.com/theforeman/foreman-documentation/pull/4507 started building correctly. That approach shows a separate preview for each build target + flavor combination:

* Foreman EL
* Foreman Deb
* Foreman + Katello
* orcharhino
* Satellite
* Foreman EL and foremanctl
* Foreman Deb and foremanctl
* Foreman + Katello and foremanctl
* orcharhino and foremanctl
* Satellite and foremanctl

Then, once any existing build target goes into full foremanctl mode, we could just drop the pre-foremanctl build and replace it with the corresponding foremanctl build. So in a way, the same thing we intended to do when we created a separate foremanctl server installation guide (in https://github.com/theforeman/foreman-documentation/pull/3975), with the intention of eventually replacing the pre-foremanctl guide with it (in https://github.com/theforeman/foreman-documentation/pull/4087).

ekohl: > So to summarize if I understand correctly:

Those 3 points are what I currently assume yes

> What's not quite clear yet from the community users perspective: will they use RPM-based on their Debian/Ubuntu? I assume so, because a) it doesn't matter too much because it's containerized; and b) because the Foreman Community will not provide Deb-based container images. If that's true, then I would say that one `foremanctl` build flavor for upstream docs is OK with prerequisites as you've mentioned.

Also correct: we have [foreman-oci-images](https://github.com/theforeman/foreman-oci-images) for Foreman and Foreman Proxy containers. Then [pulp-oci-images](https://github.com/theforeman/pulp-oci-images) and [candlepin-oci-images](https://github.com/theforeman/candlepin-oci-images) speak for themselves. There has been some investigation into Debian in https://github.com/theforeman/foremanctl/pull/235 but nothing is final yet.

> In the long run:

This is what I'd like to see, but we'll have to see in practice if we can get there.

aneta-petrova: @maximiliankolb and I had a brief chat about this. I'd like Red Hatters to have a foremanctl-satellite preview available right from the start (we do have quite a few conditionals unique to Satellite and once we start adding foremanctl-based content, a preview will be useful to continuously check if everything looks okay in the Sat preview). Maximilian then asked for the same thing for orcharhino.

Other than that, one foremanctl upstream preview is where I'm taking this PR after the conversation. Thanks for everybody's input, I really appreciate it. I'll now continue to bring this PR into a ready-to-review state and also close the other PoC PR https://github.com/theforeman/foreman-documentation/pull/4507.

aneta-petrova:
I've made some changes to the initial PoC based on conversations with @ekohl and @maximiliankolb. The PR's description has been updated to reflect that and the PR itself is now ready for review. I'm looking forward to more comments, feedback, and thoughts. To summarize: In this PR, we are looking at what the future documentation structure will look like and how we can make sure it efficiently reflects the future state of things in the Foreman+Katello world as described in https://github.com/theforeman/foreman-documentation/pull/4506#discussion_r2589844939.

ekohl:
Yes, I meant https://github.com/theforeman/foreman-documentation/pull/4517. That's what I get for quickly trying to wrap up something before a meeting.

ekohl (approved):
One thing I noticed: in the "affected files" comment: we link to removed files too. That was a bit confusing. It's rather rare we remove files so probably not important to address.

I think this is now largely doing what we want and a solid base to build on. Now the the hard part starts: actually filling the content.

README.md
```diff
@@ -34,7 +34,7 @@ See the [README in the `guides/` subdirectory](guides/README.md) for more inform
 
 To build both the static site and the guides for easy local testing, a global `Makefile` is provided in the root directory with the following targets:
 
-* `html`: builds HTML guides with all contexts (`foreman-el`, `foreman-deb`, `katello`, `satellite`, and `orcharhino`)
+* `html`: builds HTML guides with all contexts (`foreman-el`, `foreman-deb`, `foremanctl-katello`, `foremanctl-orcharhino`, `foremanctl-satellite`, `katello`, `satellite`, and `orcharhino`)
```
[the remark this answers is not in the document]
aneta-petrova: I went with the definition in the README: `builds HTML guides with all contexts` so I added the missing items to the makefile.

PR #4600, opened 2026-01-20
Consume content without subscription-manager for orcharhino builds

maximiliankolb:

Adding an admonition about consuming content without using `subscription-manager` for orcharhino builds.

All content in Foreman+Katello that is considered "unprotected" can be consumed via HTTP/HTTPS/apt/dnf/zypper/pip/curl without "subscription-manager".

However, even if users do not register their host to Foreman+Katello, ATIX still considers it a managed host and therefore part of the orcharhino Subscription.

#### Contributor checklists

### DIFF
--- a/guides/common/modules/con_content-views-in-project.adoc
+++ b/guides/common/modules/con_content-views-in-project.adoc
@@ -76,6 +76,8 @@ endif::[]
 
 ifdef::orcharhino[]
 You can use these URLs to provide versioned {project-client-name}s during host registration.
+
+include::snip_important-consuming-content-requires-a-subscription.adoc[]
 endif::[]
 
 ifeval::["{context}" == "planning"]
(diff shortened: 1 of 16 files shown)


Lennonka (changes requested):
Snippets shouldn't be included in assemblies directly. Please move them to the concepts instead.

guides/common/modules/con_content-views-in-project.adoc
Lennonka: Do you really need it twice? I would just go with the second one.

guides/common/modules/snip_important-consuming-content-requires-a-subscription.adoc
```diff
@@ -0,0 +1,5 @@
+[IMPORTANT]
+====
+You can consume content from {Project} without registering your hosts to {Project} and/or without using `subscription-manager` to consume content, for example, in containerized build pipelines.
+Any host that consumes content from {ProjectServer} or {SmartProxyServers} requires a valid subscription.
```
Lennonka: Btw, could you be more specific about what kind of subscription? Is it a sub for the OS?

PR #4626, opened 2026-01-29
Add snippet content type

aneta-petrova:

* Adding the snippet content type to snippet files
* Changing the previous way of including the procedure for configuring repositories (which relied on resolving `{build}` dynamically) to instead including one separate procedure module per build. This is because I didn't find any other way of including the content type attribute.
* Making minor changes here and there (like adding `+`) to ensure the attribute doesn't break builds.

https://github.com/jhradilek/asciidoctor-dita-vale/blob/138759cd28e7781860d6259741f6d682256dfa72/README.md?plain=1#L91

N/A

#### Contributor checklists

* [x] Foreman 3.18/Katello 4.20

### DIFF
--- a/guides/common/modules/proc_configuring-repositories.adoc
+++ b/guides/common/modules/proc_configuring-repositories.adoc
@@ -33,6 +33,7 @@ include::snip_configuring-repositories-foremanctl-satellite.adoc[]
 endif::[]
 endif::[]
 
+ifndef::foremanctl[]
 ifdef::foreman-el,katello,satellite[]
 .Verification
 * Verify that the required repositories are enabled:
@@ -41,4 +42,5 @@ ifdef::foreman-el,katello,satellite[]
 ----
 # dnf repolist enabled
 ----
+endif::[]
 endif::[]


guides/common/modules/snip_configuring-repositories-katello.adoc
maximiliankolb: AFAIK this was a symlink before. Using fd/find sometimes trips over this.

aneta-petrova:
> I also strongly recommend that you/we talk with the people that create this requirement again.

I will talk to them. It will be good to get more understanding on the reasons behind the requirement. Then we can think about possible next steps for our team.

aneta-petrova:
Not defining the snippet content type causes asciidoctor-dita-vale errors. We haven't run into these before because https://github.com/apinnick/scripts/blob/main/extract-doc-modules.sh, which is part of this team's workflow, does not process snippets. But they will pop up during the migration, and that's going to block our progress.

Asciidoctor-dita-vale does not exclude the use of snippets and they are used on other teams at Red Hat as well. So another thing I want to do is to look at how they look in other repos, to see some more usage examples.

guides/common/modules/proc_configuring-repositories.adoc
```diff
@@ -0,0 +1,44 @@
...
+endif::[]
+ifdef::orcharhino[]
+include::snip_configuring-repositories-orcharhino.adoc[]
+endif::[]
+ifdef::satellite[]
+include::snip_configuring-repositories-satellite.adoc[]
+endif::[]
+endif::[]
+ifdef::foremanctl[]
+ifdef::foreman-katello[]
```
maximiliankolb:
```suggestion
ifdef::katello[]
```

I am not 100% sure, but I did not find any instance where we define "foreman-katello". Was this supposed to be a comma instead of a dash?

Looking at `ls -1 guides/common/attributes*` and `rg "html:" Makefile`, I assume that we only build foremanctl for the Katello-flavor.

aneta-petrova:
This is still WIP and I'm investigating the issues. It might take time and to prevent further reviews while I'm reviewing the changes myself, I'm converting this PR to draft and removing the review label for now.

PR #4679, opened 2026-02-25
Fix Podman proxy configuration leaking into containers

chris1984:
Only set https_proxy in the engine env, which is sufficient for container pulls, and disable http_proxy passthrough to containers to prevent them from routing inter-container traffic through the proxy.

* Update the Podman HTTP proxy configuration procedure to only set https_proxy in the [engine] env (removing http_proxy), and add a [containers] section with http_proxy=false to prevent proxy environment variables from leaking into containers.

* Current [Jira](https://issues.redhat.com/browse/SAT-40216) 
* The current documentation sets both http_proxy and https_proxy in the Podman engine env. Due to a Podman bug [RHEL-127541](https://issues.redhat.com/browse/RHEL-127541), these proxy settings are passed through to containers, causing inter-container communication to be routed through the proxy. 
* Setting https_proxy alone is sufficient for Podman to pull container images, and http_proxy=false under [containers] prevents the proxy settings from leaking into the container environment.

* Even with http_proxy=false, the https_proxy variable may still leak into containers but setting http_proxy=false mitigates the most disruptive behavior (inter-container traffic being blocked by the proxy).
* This aligns with the documented workaround from the bug report.

#### Contributor checklists

### DIFF
--- a/guides/common/modules/proc_configuring-podman-to-use-an-http-proxy.adoc
+++ b/guides/common/modules/proc_configuring-podman-to-use-an-http-proxy.adoc
@@ -16,10 +16,13 @@ For more information, see https://podman.io/docs/installation#installing-on-linu
 endif::[]
 
 .Procedure
-* Edit the `/etc/containers/containers.conf` file to include the following directive:
+* Edit the `/etc/containers/containers.conf` file to include the following directives:
 +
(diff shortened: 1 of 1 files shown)


Lennonka (changes requested):
Just a few style tweaks to make it pretty.

PR #4647, opened 2026-02-12
Review 'Managing security compliance' for CQA

aneta-petrova:
#### What changes are you introducing?

* Review assemblies to make sure each describes one user story
* Review abstracts to be descriptive and 50-300 chars long
* Review titles to be descriptive and 3-11 words long

#### Contributor checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.18/Katello 4.20 (Satellite 6.19)
* [x] Foreman 3.17/Katello 4.19
* [x] Foreman 3.16/Katello 4.18 (Satellite 6.18; orcharhino 7.6)
* [ ] Foreman 3.15/Katello 4.17
* [ ] Foreman 3.14/Katello 4.16 (Satellite 6.17; orcharhino 7.4; orcharhino 7.5)
* [ ] Foreman 3.13/Katello 4.15 (EL9 only)
* [ ] Foreman 3.12/Katello 4.14 (Satellite 6.16; orcharhino 7.2 on EL9 only; orcharhino 7.3)
* We do not accept PRs for Foreman older than 3.12.

PR #4687, opened 2026-03-02
Add auxiliary IDs to modules linked from Sat web UI

aneta-petrova:

* Adding a few auxiliary IDs per https://docs.asciidoctor.org/asciidoc/latest/sections/custom-ids/#assign-auxiliary-ids

https://github.com/theforeman/foreman-documentation/pull/4647 changed several IDs. Some of these changes broke links in the web UI. While https://github.com/RedHatSatellite/foreman_theme_satellite/pull/314 has been raised to change the links in the web UI, what we can do on our side is to add auxiliary IDs to keep the old links working.

N/A

#### Contributor checklists

### DIFF
--- a/guides/common/modules/con_configuration-management-with-puppet-in-project.adoc
+++ b/guides/common/modules/con_configuration-management-with-puppet-in-project.adoc
@@ -1,7 +1,7 @@
 :_mod-docs-content-type: CONCEPT
 
 [id="configuration-management-with-puppet-in-{project-context}"]
-= Configuration management with Puppet in {Project}
+= Configuration management with Puppet in {Project}[[introducing-configuration-management-by-using-puppet]]
 
 [role="_abstract"]
 Puppet is a configuration management tool to automate configurations of hosts across your infrastructure.


aneta-petrova:
> If there's already a PR to fix the anchors in foreman_theme_satellite, then please raise this PR against 3.18 (or 3.16 if you can still get it into the 3.18-based Satellite build).

That is a good question. @adamruzicka Is the plan to go forward with the foreman_theme_satellite PR? Originally I thought this PR would eliminate the need to update the links in the web UI, now I'm not so sure.

aneta-petrova:
Can you please take another look @adamruzicka? Would adding auxiliary IDs like this mean that the links in the web UI will keep working so that we don't have to change them in the web UI?

adamruzicka:
> mkolb: If there's already a PR to fix the anchors in foreman_theme_satellite, then please raise this PR against 3.18 (or 3.16 if you can still get it into the 3.18-based Satellite build).
> ...
> apetrova: Is the plan to go forward with the foreman_theme_satellite PR? Originally I thought this PR would eliminate the need to update the links in the web UI, now I'm not so sure.

I missed this one previously, I'll sleep on this one and let you know tomorrow

maximiliankolb (approved):
Thanks Anet, style-wise LGTM. I assume that we're going with example 1: https://docs.asciidoctor.org/asciidoc/latest/sections/custom-ids/#assign-auxiliary-ids

-> a single whitespace before; no whitespace after the brackets 👍

maximiliankolb:
> each branded url had three components - guide (managing_configurations_by_using_puppet_integration), chapter (introducing-configuration-management-by-using-puppet) and anchor (performing-configuration-management_managing-configurations-puppet).

Is this something downstream-specific?

To my understanding, each module in Foreman+Katello docs has **two** parts that a documentation button needs to know:

* NOT base URL because that's static for upstream/downstream products, for example, `https://docs.theforeman.org/`
* NOT version because that's defined via foreman, for example, `nightly`
* guide, for example, `Managing_Configurations_Puppet`
* NOT build because that's defined via scenario, for example, `katello`
* anchor, for example, `Overriding_Smart_Class_Parameter_for_Location_managing-configurations-puppet`

With that, Foreman/Katello/any plugin should be able to assemble the full URL: https://docs.theforeman.org/nightly/Managing_Configurations_Puppet/index-katello.html#Overriding_Smart_Class_Parameter_for_Location_managing-configurations-puppet

> With html-single, this stops being a problem as long as the guide name stays the same.

:+1: Should have re-read this before starting my reply. I am commenting anyway because the URL structure might still be interesting to some. If I'm wrong on the structure please tell me!

> as long as the guide name stays the same

We rarely rename a guide aka. a directory in `./guides/`. If we do, we could announce it.

PR #4754, opened 2026-04-06
Add katello repos to foremanctl installation doc

arvind4501:
Add katello-repos package installation step
When running `foremanctl deploy --add-feature hammer`, the installation fails because it tries to install hammer_cli_katello package but cannot find it since the Katello repository is not enabled. Installing   
  katello-repos-latest.rpm first ensures the Katello repositories are configured before attempting to install Katello-specific packages.
#### Contributor checklists

### DIFF
--- a/guides/common/modules/proc_configuring-repositories.adoc
+++ b/guides/common/modules/proc_configuring-repositories.adoc
@@ -73,6 +73,14 @@ ifdef::foremanctl[]
 ----
 # {package-install} https://yum.theforeman.org/releases/{ProjectVersion}/el{distribution-major-version}/x86_64/foreman-release.rpm
 ----
+ifdef::katello[]
+. Install the `katello-repos-latest.rpm` package:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {package-install} https://yum.theforeman.org/katello/{KatelloVersion}/katello/el{distribution-major-version}/x86_64/katello-repos-latest.rpm
+----
+endif::[]
 . Enable the required additional repositories:
 +
 [options="nowrap" subs="+quotes,attributes"]
--- a/guides/common/modules/proc_configuring-repositories.adoc
+++ b/guides/common/modules/proc_configuring-repositories.adoc
@@ -89,8 +89,7 @@ endif::[]
 ----
 endif::[]
 
-ifndef::foremanctl[]
-ifdef::foreman-el,katello,satellite[]
+ifdef::foreman-el,katello,satellite,foremanctl[]
 .Verification
 * Verify that the required repositories are enabled:
 +
@@ -98,5 +97,4 @@ ifdef::foreman-el,katello,satellite[]
 ----
 # dnf repolist enabled
 ----
-endif::[]
 endif::[]
--- a/guides/common/modules/proc_configuring-repositories.adoc
+++ b/guides/common/modules/proc_configuring-repositories.adoc
@@ -8,7 +8,6 @@
 Configure the required repositories.
 
 .Procedure
-ifndef::foremanctl[]
 ifdef::satellite[]
 . Disable all repositories:
 +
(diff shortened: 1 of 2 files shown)


archanaserver (approved):
LGTM

Gauravtalreja1 (approved):
LGTM, just a small change would be nice, could you be specific in your commit message/PR title about which guide needs update?

jafiala (approved):
LGTM

guides/common/modules/proc_configuring-repositories.adoc
ekohl: Since you're enabling repositories, isn't this verification also fine for `foremanctl`?

ekohl: Since https://github.com/theforeman/foreman-packaging/pull/12560 we package foremanctl in our repositories. For nightly this copr enablement is still OK for stable releases it should not be needed. I think we have some issue to properly track that in the release process, but I can't find that right now.

@evgeni do you happen to remember?

guides/common/modules/proc_configuring-repositories.adoc
```diff
@@ -73,6 +73,14 @@ ifdef::foremanctl[]
 ----
 # {package-install} https://yum.theforeman.org/releases/{ProjectVersion}/el{distribution-major-version}/x86_64/foreman-release.rpm
 ----
+ifdef::katello[]
```
ekohl: The procedure is now the same as what we have in `ifdef::foreman-el,katello[]` on line 40. Perhaps it should be split off into `snip_configuring-repositories-foreman-rpm.adoc` with (in pseudo-code):
```asciidoc
yum install foreman-release.rpm

ifdef::katello[]
yum install katello-repos-latest.rpm
endif::[]

ifdef::foremanctl[]
dnf copr enable @theforeman/foremanctl rhel-9-x86_64
endif::[]
ifndef::foremanctl[]
yum install puppet-release.rpm
endif::[]
```

Then in this file you'd have:
```asciidoc
ifdef::satellite[]
...
endif::[]
ifdef::foreman-deb[]
include::snip_configuring-repositories-foreman-deb.adoc[]
endif::[]
ifdef::foreman-el,katello,foremanctl[]
include::snip_configuring-repositories-foreman-rpm.adoc[]
endif::[]
```

Isn't that a lot easier to maintain?

ekohl (commented):
I don't want to have too detailed of an opinion, but this file contains a lot of conditionals so I'm going to share my opinions of the whole file. They don't have to be blockers for this PR, but consider it for the long term.

guides/common/modules/proc_configuring-repositories.adoc
evgeni: Nope

guides/common/modules/proc_configuring-repositories.adoc
arvind4501: Yes, and updated

guides/common/modules/proc_configuring-repositories.adoc
ekohl: Ok, then I'll assume we don't have it and I'll make sure we track it somewhere.

guides/common/modules/proc_configuring-repositories.adoc
```diff
@@ -37,52 +36,15 @@ ifdef::foreman-el,katello[]
...
-+
-[options="nowrap" subs="+quotes,attributes"]
-----
-# dnf copr enable @theforeman/foremanctl rhel-9-x86_64
-----
-endif::[]
 
-ifndef::foremanctl[]
-ifdef::foreman-el,katello,satellite[]
+ifdef::foreman-el,katello,satellite,foremanctl[]
```
ekohl: Not a blocker and don't know if it's correct but isn't this now easier?
```suggestion
ifndef::foreman-deb[]
```

guides/common/modules/proc_configuring-repositories.adoc
ekohl: Found it as a note in https://redhat.atlassian.net/browse/SAT-42499 as an acceptance criteria.

ekohl (approved):

amolpati30 (approved):

jafiala:
Actually @amolpati30 why is the content now separated into a snippet? Do we intend to include that snippet also somewhere else, where the procedure itself would not be a better fit?

Snippets are a little harder to maintain, for example, our validation complains that the .Procedure does not have any steps in it. (Which it does, but the checker does not follow the include into the snippet.)

Even if we are planning to have multiple "Configuring repos" for different scenarios, using ifdefs within the procedure, or even having multiple procedure files that would get included conditionally, would be a better structure.

arvind4501:
the snippet was introduced because the main file was having repeated code and lot of conditions which decreses readbility. IMHO now with snippet code is easier to maintain and looks fine.

we have https://github.com/theforeman/foreman-documentation/blob/07869a5d0cc9408580c2e0f77225b66964cbc6e1/guides/common/modules/proc_configuring-repositories.adoc?plain=1#L31 which includes snippet for debian related config.

> having multiple procedure files that would get included conditionally, would be a better structure.

Isn't snippet similar to files included conditionaly?

jafiala:
@arvind4501 OK, makes sense!

jafiala (approved):

Lennonka:
Cherry picked:
- 2c9e30b649..ec864e3029  3.18 -> 3.18

PR #4761, opened 2026-04-09
Update information about read-only MCP server

AkshayGadhaveRH:
If the MCP server is allowed to perform REx jobs or CV actions, it ceases to function in read-only mode. Editing the earlier information regarding this.

JIRA:
https://redhat.atlassian.net/browse/SAT-44151

#### Contributor checklists

### DIFF
--- a/guides/common/modules/con_connecting-ai-applications-to-the-mcp-server-for-project.adoc
+++ b/guides/common/modules/con_connecting-ai-applications-to-the-mcp-server-for-project.adoc
@@ -6,7 +6,6 @@
 [role="_abstract"]
 {Project} provides a Model Context Protocol (MCP) server to enable integration with your MCP-compatible AI applications.
 By connecting your MCP client to the MCP server, you can obtain comprehensive AI-generated reports and data analysis of your {Project} inventory.
-You can also start the MCP server with permissions to manage content views on your {Project} and trigger remote execution jobs.
 
 :FeatureName: MCP server for {Project}
 include::snip_technology-preview.adoc[]
--- a/guides/common/modules/con_connecting-ai-applications-to-the-mcp-server-for-project.adoc
+++ b/guides/common/modules/con_connecting-ai-applications-to-the-mcp-server-for-project.adoc
@@ -5,7 +5,8 @@
 
 [role="_abstract"]
 {Project} provides a Model Context Protocol (MCP) server to enable integration with your MCP-compatible AI applications.
-By connecting your MCP client to the MCP server, you can obtain comprehensive AI-generated reports and data analysis of your {Project} inventory.
+You can obtain comprehensive AI-generated reports and data analysis of your {Project} inventory.
+Optionally, you can start the MCP server with permissions to edit the inventory.
 
 :FeatureName: MCP server for {Project}
 include::snip_technology-preview.adoc[]
--- a/guides/common/modules/con_connecting-ai-applications-to-the-mcp-server-for-project.adoc
+++ b/guides/common/modules/con_connecting-ai-applications-to-the-mcp-server-for-project.adoc
@@ -6,6 +6,7 @@
 [role="_abstract"]
 {Project} provides a Model Context Protocol (MCP) server to enable integration with your MCP-compatible AI applications.
 By connecting your MCP client to the MCP server, you can obtain comprehensive AI-generated reports and data analysis of your {Project} inventory.
+You can also start the MCP server with permissions to manage content views on your {Project} and trigger remote execution jobs.
 
 :FeatureName: MCP server for {Project}
 include::snip_technology-preview.adoc[]
--- a/guides/common/modules/con_project-mcp-integration-overview.adoc
+++ b/guides/common/modules/con_project-mcp-integration-overview.adoc
@@ -37,5 +37,9 @@ When the connection is established, you query your AI application for data about
 The MCP client within the AI application obtains the required data from your {Project} inventory and returns it to the AI application.
 The AI application processes the data to answer your queries.
 
+The MCP server runs in read-only mode by default.
+It extracts data from your {Project} inventory but cannot write, edit, or make any updates to the inventory.
+Note that the MCP server can edit the inventory, if you enable the MCP server to trigger remote execution jobs or content view actions. 
+
 .Additional resources
 * link:https://modelcontextprotocol.io[Model Context Protocol]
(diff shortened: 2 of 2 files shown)


guides/common/modules/con_project-mcp-integration-overview.adoc
```diff
@@ -37,8 +37,9 @@ When the connection is established, you query your AI application for data about
 The MCP client within the AI application obtains the required data from your {Project} inventory and returns it to the AI application.
 The AI application processes the data to answer your queries.
 
-The MCP server runs in read-only mode.
+The MCP server runs in read-only mode by default.
 It extracts data from your {Project} inventory but cannot write, edit, or make any updates to the inventory.
+Note that the MCP server can edit the inventory, if you enable the MCP server to trigger remote execution jobs or content view actions.
```
aneta-petrova: As a user, I'd be curious to know how do I spin up an MCP server in "default/read-only" mode and how in the "non-default/write" mode. And if I already have an MCP server up, how can I tell which is which? Which in turn made me wonder whether we now need a full-blown concept module explaining the read-only and write modes.

aneta-petrova:
Are cherry-picks to previous versions needed as well? If I'm not mistaken, there is only one image for the MCP server so we'd need to update the older docs versions too.

aneta-petrova:
On top of changing that reference to read-only mode, should the whole assembly be reviewed to see if other places need updating as well? For example, looking at the assembly intro, it says that the use cases is obtaining reports and data analysis (https://github.com/theforeman/foreman-documentation/blob/master/guides/common/modules/con_connecting-ai-applications-to-the-mcp-server-for-project.adoc?plain=1). With write mode available, that's now only half the picture.

guides/common/modules/con_project-mcp-integration-overview.adoc
AkshayGadhaveRH: @adamruzicka is there any way to identify which options are enabled for the MCP server (allowed-cv-actions and allowed-rex-features). Also. is there any way to disable these actions?

adamruzicka: > is there any way to identify which options are enabled for the MCP server (allowed-cv-actions and allowed-rex-features)

Technically yes, tools will be only available for the enabled things. For example if you don't enable any allowed-rex-features, you will not have the trigger_remote_execution_job tool available.

Still, I would expect the users to know how they run the mcp server rather than them running the mcp server _somehow_ and then trying to figure out what's enabled and what's not?

> is there any way to disable these actions?

By not passing in the options that enable them.

maximiliankolb:
```suggestion
If you enable the MCP server to trigger remote execution jobs or content view actions, the MCP server can edit the inventory.
```

I suggest to go with the if statement first; this also allows us to prevent starting the line with "note that" which in turn would have led us to create an admonition.

maximiliankolb: Does the mode depend on the MCP module? aka. if you only query your AI to see how many hosts are in location X, then read-only is OK and not changed during the start of the MCP container.

If you want to patch hosts, you'll need to make changes to Foreman/Katello -> rw mode.

Maybe:

````diff
-The MCP server runs in read-only mode by default.
-It extracts data from your {Project} inventory but cannot write, edit, or make any updates to the inventory.
-Note that the MCP server can edit the inventory, if you enable the MCP server to trigger remote execution jobs or content view actions. 
+By default, the MCP server runs in read-only mode and cannot make any changes to {Project}.
+It extracts data from your {Project} inventory but does not write or edit your inventory.
+Depending on the **workflow**, start the MCP server with additional permissions, for example, to allow it to start remote execution jobs or patch hosts.
````

Please check if workflow is the correct term here.

Comment (AkshayGadhaveRH) [reply guides/common/modules/con_project-mcp-integration-overview.adoc]: @maximiliankolb I'll just replace `allow it to start remote execution jobs or patch hosts` with `allow it to start remote execution jobs or publish and promote content views` since the patching is done via REx. 
How about 
`If you start the MCP server with additional permissions, for example, to allow it to start remote execution jobs or publish and promote content views, it can make changes to the inventory.`
Since we need to explicitly call out that the MCP server can run in write mode.

**Comment by maximiliankolb** [comment]
Please rebase to HEAD of "master" and resolve the merge conflicts.

File: guides/common/modules/con_connecting-ai-applications-to-the-mcp-server-for-project.adoc [hunk 2]
```diff
@@ -6,6 +6,7 @@
 [role="_abstract"]
 {Project} provides a Model Context Protocol (MCP) server to enable integration with your MCP-compatible AI applications.
 By connecting your MCP client to the MCP server, you can obtain comprehensive AI-generated reports and data analysis of your {Project} inventory.
+You can also start the MCP server with permissions to manage content views on your {Project} and trigger remote execution jobs.
```
aneta-petrova: This is now almost 400 characters. We should aim for 50-300 characters in an abstract.

maximiliankolb: I would probably drop this line here because it's already pretty much part of `guides/common/modules/con_project-mcp-integration-overview.adoc`.

AkshayGadhaveRH: Edited the abstract to fit the character limit.

AkshayGadhaveRH: @maximiliankolb reworded the abstract a bit without dropping this line, what do you think?

guides/common/modules/con_project-mcp-integration-overview.adoc
```diff
@@ -37,5 +37,9 @@ When the connection is established, you query your AI application for data about
 The MCP client within the AI application obtains the required data from your {Project} inventory and returns it to the AI application.
 The AI application processes the data to answer your queries.
 
+The MCP server runs in read-only mode by default.
+It extracts data from your {Project} inventory but cannot write, edit, or make any updates to the inventory.
+If you start the MCP server with additional permissions, for example, to allow it to start remote execution jobs or publish and promote content views, it can make changes to the inventory.
```
adamruzicka: Technically it is correct, but it reads a bit like "if you allow it to make changes it will be able to make changes"

Lennonka: Would this sound better? I think it's a good idea to showcase what kinds of changes it can make.
```suggestion
You can start the MCP server with permissions to make changes to the inventory, for example, to start remote execution jobs or publish and promote content views.
```

adamruzicka (approved):
This is technically correct and it reads well.

Lennonka (approved):
LGTM

Lennonka:
Cherry picked:
- f0b099dc53..b5a14789f2  3.18 -> 3.18

PR #4763, opened 2026-04-09
Remove obsolete BZ2177730 reference from Puppet agent registration

aneta-petrova:

Removing a reference to an obsolete Bugzilla ticket

It looks like the ticket will never get addressed, which makes the paragraph obsolete. The rest of the procedure already documents the required steps.

N/A

#### Contributor checklists

* [x] Foreman 3.12/Katello 4.14 (Satellite 6.16; orcharhino 7.2 on EL9 only; orcharhino 7.3)

### DIFF
--- a/guides/common/modules/proc_installing-and-configuring-puppet-agent-during-host-registration.adoc
+++ b/guides/common/modules/proc_installing-and-configuring-puppet-agent-during-host-registration.adoc
@@ -58,11 +58,6 @@ endif::[]
 * Optional: Add a host parameter named `puppet_ca_server`, select the *string* type, and set the value to the hostname of your Puppet CA server, such as `puppet-ca.example.com`.
 If `puppet_ca_server` is not set, the Puppet agent will use the same server as `puppet_server`.
 * Optional: Add a host parameter named `puppet_environment`, select the *string* type, and set the value to the Puppet environment you want the host to use.
-
-ifndef::orcharhino[]
-+
-Until the https://bugzilla.redhat.com/show_bug.cgi?id=2177730[BZ2177730] is resolved, you must use host parameters to specify the Puppet agent configuration even in integrated setups where the Puppet server is a {SmartProxyServer}.
-endif::[]
 ifdef::katello,orcharhino,satellite[]
 . Navigate to *Hosts* > *Register Host* and register your host using an appropriate activation key.
 endif::[]


maximiliankolb (approved):
LGTM based on status of the BZ ticket.

adamruzicka (approved):
Originally I was thinking about just dropping the conditional part and leaving the rest stated as a fact. After reading this and the BZ again, I'm fine with dropping it completely.

guides/common/modules/proc_installing-and-configuring-puppet-agent-during-host-registration.adoc
ekohl, line 8: I don't want to distract too much, but this reads as contradictory with the previous sentence. Was it intended to state that a configured Puppet *server* is required?

ekohl (approved):

guides/common/modules/proc_installing-and-configuring-puppet-agent-during-host-registration.adoc
aneta-petrova: I read this as an explanation of the use case. You need a configured Puppet agent (= you need to follow this procedure) if you want to integrate Puppet with Foreman.

ekohl: I read it as a prerequisite but my understanding is that global registration will install the Puppet packages if needed. https://github.com/theforeman/foreman/blob/f2cbd0f0a8e5762bc480b0caa370d385268f768c/app/views/unattended/provisioning_templates/host_init_config/host_init_config_default.erb#L51-L54 is supposed to do exactly that.

If that doesn't work, then it could explain why the original BZ was filed.

guides/common/modules/proc_installing-and-configuring-puppet-agent-during-host-registration.adoc
```diff
@@ -58,11 +58,6 @@ endif::[]
 * Optional: Add a host parameter named `puppet_ca_server`, select the *string* type, and set the value to the hostname of your Puppet CA server, such as `puppet-ca.example.com`.
 If `puppet_ca_server` is not set, the Puppet agent will use the same server as `puppet_server`.
 * Optional: Add a host parameter named `puppet_environment`, select the *string* type, and set the value to the Puppet environment you want the host to use.
-
```
ekohl, line 61: Looking at this a second time, I now understand better why the BZ comment was there.

This is not how we prefer to integrate: setting the parameters is intended for unmanaged Puppet servers. In the BZ there is a suggestion to use host properties to set the right Puppet server, Puppet CA server and Puppet environment. That is what we indeed prefer and why the comment below existed.

So the BZ may indicate the global registration form doesn't have fields to set the host properties. In that case you can only rely on it being set via the host group and let it be inherited.

ekohl (changes requested):
Sorry for the premature ACK.

PR #4766, opened 2026-04-10
Update Puppet guide for OpenVox

aneta-petrova:

Updating the Managing configurations by using Puppet guide to prefer OpenVox instead of Puppet.

Foreman is transitioning from Puppet to OpenVox.

https://redhat.atlassian.net/browse/SAT-33215

Assisted-by: Cursor

#### Contributor checklists


aneta-petrova:
@ekohl This PR is still a draft but can you please take a quick look? I updated the first (conceptual) chapter and I'd like to know if I got the gist of the transition to OpenVox correctly. Some early feedback would be helpful in ensuring that I'm on the right path when I continue to update the rest of the guide.

guides/common/modules/con_how-puppet-integrates-with-project.adoc
```diff
@@ -1,38 +1,47 @@
...
+Puppet agent::
+Hosts run the Puppet agent that communicates with OpenVox server.
 
 Puppet modules::
-The _desired state_ of a host is defined in a _catalog_.
-The _catalog_ is compiled from Puppet manifests of one or more Puppet modules assigned to the host.
 A Puppet module is a collection of classes, manifests, resources, files, and templates.
 The Puppet modules work as components of host configuration definitions.
 
-Smart Class parameters::
```
ekohl, line 33: AFAIK this feature remains, also with OpenVox

ekohl (commented):
@bastelfreak and @binford2k this is the change for Foreman (and its downstreams) to OpenVox. I struggled to help @aneta-petrova find exactly which terms we change and which ones we don't. For example, do we already speak about OpenVox modules instead of Puppet modules?

I know this PR is still a draft, but I'd appreciate your opinions on the terminology.

guides/common/modules/con_how-puppet-integrates-with-project.adoc
aneta-petrova: It's mentioned below on line 46. Isn't it a feature of Foreman rather than Puppet itself? That's my interpretation of it, which is why I moved it to the part that describes the architecture of the integration and it's no longer listed as a Puppet component.

guides/common/modules/ref_openvox-versions-supported-for-integration-with-project.adoc
```diff
@@ -0,0 +1,21 @@
...
+
+[id="supported-openvox-and-puppet-versions_{context}"]
+= Supported OpenVox and Puppet versions
+
+[role="_abstract"]
+Before you begin with the Puppet integration, review the supported OpenVox and Puppet versions.
+
+The following versions of the OpenVox server are supported on {ProjectServer} and {SmartProxyServers}:
+
+* OpenVox 8
```
aneta-petrova: I was asked to link to OpenVox 8 system requirements here (just like we now link to [Puppet requirements](https://www.puppet.com/docs/puppet/7/system_requirements.html)) but I didn't find an OpenVox equivalent in their docs.

guides/common/modules/con_how-puppet-integrates-with-project.adoc
```diff
@@ -1,38 +1,47 @@
 :_mod-docs-content-type: CONCEPT
 
-[id="How_Puppet_Integrates_With_Project_{context}"]
+[id="how-puppet-integrates-with-project"]
 = How Puppet integrates with {Project}
```
ekohl, line 4: Should this start off with a clarification on Puppet vs OpenVox? IMHO we should explain that for historic reasons we continue to use Puppet in various places, but actually use OpenVox as an implementation. In the UI and tooling users may still see Puppet as well.

guides/common/modules/ref_openvox-versions-supported-for-integration-with-project.adoc
ekohl: There's https://voxpupuli.org/openvox/ but I don't see requirements either. @bastelfreak or @binford2k have you had any discussions on providing a similar page for OpenVox?

aneta-petrova:
> @bastelfreak and @binford2k this is the change for Foreman (and its downstreams) to OpenVox. I struggled to help @aneta-petrova find exactly which terms we change and which ones we don't. For example, do we already speak about OpenVox modules instead of Puppet modules?

My interpretation of the situation:

* Server: There is no more Puppet server. Use "OpenVox server" everywhere.
* Agent: Both OpenVox agent and Puppet agent are supported. Use both as appropriate.
* Module: This is still called a Puppet module.
* Environment: This is still called a Puppet environment.

Or, to summarize, it's the server and agent technology context that is changing. Puppet modules and Puppet environments remain stable, just like the general term for the technology framework ("Puppet").

guides/common/modules/con_how-puppet-integrates-with-project.adoc
aneta-petrova: I'd prefer to see that in the Release notes. It's the RNs that explain what's changing; the product guides explain what exists at the moment in the given version.

If the UI refers to "Puppet modules" for example, then I think that's actually okay (see my understanding of the situation in https://github.com/theforeman/foreman-documentation/pull/4766#issuecomment-4235738084).

If the UI refers to "Puppet server", then that's a bug and we can track it in the Release notes as a known issue.

Does that make sense? Am I missing anything?

ekohl: Summarizing an offline conversation: I think we should clarify why the guide sometimes mentions Puppet and sometimes OpenVox. We liked what https://voxpupuli.org/openvox/install/ writes
> OpenVox is functionally equivalent to Puppet

Making my proposal concrete, perhaps something based on:
> This guide uses Puppet and OpenVox interchangeably.

aneta-petrova: I updated the description of the components in the overview section ("How Puppet integrates with Foreman"): It now lists only OpenVox server and OpenVox agent as components. Puppet server and Puppet agent get a minor reference (server as a component that's no longer compatible, agent as a component that's a legacy option), clarifying that the OpenVox bits are functionally equivalent to the Puppet ones.

This is a tricky part to balance: I'm trying to avoid getting into lengthy explanations of why this is happening, and rather focus on explaining the current state. I think I can now resolve this thread but let me know if there are further concerns.

guides/common/modules/ref_supported-openvox-and-puppet-versions.adoc
```diff
@@ -0,0 +1,24 @@
...
+The Puppet modules you are using must be compatible with the supported version of OpenVox server.
+====
+
+The following version of OpenVox agent is supported on {Project} hosts:
+
+* OpenVox agent 8
+
+Additionally, Puppet agent 7 remains compatible with OpenVox in {Project} as a legacy alternative.
+If your {Project} hosts run Puppet agent 7, they will integrate with the OpenVox implementation in {Project}.
+However, Puppet agent 7 is end of life and it is recommended to migrate {Project} hosts to OpenVox agent 8. #Okay, but how?#
```
aneta-petrova: This should happen transparently during a `dnf upgrade` once the client repository is updated to the one that contains the openvox packages.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-manually.adoc
```diff
@@ -45,6 +46,7 @@ ifndef::orcharhino[]
 ----
 # {client-package-install-el8} puppet-agent
 ----
+#This command works for both OpenVox agent and Puppet agent.#
```
aneta-petrova: Based on whether the host contains packages for puppet or openvox. The command remains the same for both.

guides/common/modules/con_how-puppet-integrates-with-project.adoc
```diff
@@ -1,38 +1,62 @@
...
+OpenVox server::
+OpenVox server is the central component that stores Puppet configuration definitions in {Project}.
+It runs on {ProjectServer} or {SmartProxyServers}.
 +
-The Puppet server forwards _facts_ to {Project} and {Project} stores them for later use.
-Based on the _facts_ and other definitions, {Project} constructs the ENC answer to the Puppet server.
-The Puppet server compiles a _catalog_ based on the ENC answer and sends the _catalog_ to the Puppet agent.
+[NOTE]
+====
+Puppet server is no longer compatible with {Project} for Puppet integration.
```
ekohl: Technically, it's still compatible but no longer receives any security updates. Some would call this unsupported. At least, if we consider Puppet server as the open source version. Right now we don't mention anything about [Puppet core](https://www.puppet.com/products/puppet-core) (which is entirely untested in combination with Foreman).

aneta-petrova: I changed the wording to "not supported".

binford2k:
Yes. It's still a Puppet module in the Puppet ecosystem. And OpenVox is the oss implementation that many of us choose to use.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-provisioning.adoc
```diff
@@ -1,31 +1,30 @@
...
+If you prefer to install and configure the OpenVox agent on a host during the provisioning process, you can do so by adding the `puppet_setup` snippet to your provisioning template.
+A configured OpenVox agent is required on the host for Puppet integration with your {Project}.
 
 .Prerequisites
-* Puppet must be enabled in your {Project}.
-For more information, see xref:enabling-puppet-integration-with-{project-context}[].
 ifdef::satellite[]
-+
 include::snip_prerequisite-project-client-repository-ak.adoc[]
+* The satellite-client-openvox-for-rhel-8 or satellite-client-openvox-for-rhel-9 repository is synchronized on {ProjectServer} and enabled in the activation key you use.
```
ekohl: The name also includes the arch. https://redhat.atlassian.net/browse/SPMM-22387 lists the exact ones.
```suggestion
* The `satellite-client-openvox-for-rhel-8-<arch>-rpms` or `satellite-client-openvox-for-rhel-9-<arch>-rpms` repository is synchronized on {ProjectServer} and enabled in the activation key you use.
```

We have attributes for the regular client repos and should we also introduce these for the Satellite Client OpenVox repo?
https://github.com/theforeman/foreman-documentation/blob/40c450e1f8f92488ccb389219c3d339a630e5a60/guides/common/attributes-satellite.adoc#L164
https://github.com/theforeman/foreman-documentation/blob/40c450e1f8f92488ccb389219c3d339a630e5a60/guides/common/attributes-satellite.adoc#L173
https://github.com/theforeman/foreman-documentation/blob/40c450e1f8f92488ccb389219c3d339a630e5a60/guides/common/attributes-satellite.adoc#L177

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-provisioning.adoc
```diff
@@ -66,13 +65,13 @@ ifndef::katello,orcharhino,satellite[]
...
 Select the *boolean* type and set the value to `true`.
 endif::[]
-. Set configuration for the Puppet agent.
-* If you use an integrated Puppet server, ensure that you select a Puppet {SmartProxy}, Puppet CA {SmartProxy}, and Puppet environment when you create a host.
-* If you use a non-integrated Puppet server, either set the following host parameters in global parameters, or a host group, or when you create a host:
-** Add a host parameter named `puppet_server`, select the *string* type, and set the value to the hostname of your Puppet server, such as `puppet.example.com`.
-** Optional: Add a host parameter named `puppet_ca_server`, select the *string* type, and set the value to the hostname of your Puppet CA server, such as `puppet-ca.example.com`.
-If `puppet_ca_server` is not set, the Puppet agent will use the same server as `puppet_server`.
+. Set configuration for the OpenVox agent.
+* If you use an integrated OpenVox server, ensure that you select a OpenVox {SmartProxy}, OpenVox CA {SmartProxy}, and Puppet environment when you create a host.
```
ekohl: The actual feature names in Foreman haven't been modified:
```suggestion
* If you use an integrated OpenVox server, ensure that you select a Puppet {SmartProxy}, Puppet CA {SmartProxy}, and Puppet environment when you create a host.
```

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-registration.adoc
```diff
@@ -1,43 +1,37 @@
...
-ifdef::managing-configurations-puppet[]
-For more information, see xref:enabling-puppet-integration-with-{project-context}[].
-endif::[]
-ifndef::managing-configurations-puppet[]
-For more information, see {ManagingConfigurationsPuppetDocURL}enabling-puppet-integration-with-{project-context}[Enabling Puppet integration with {Project}] in _{ManagingConfigurationsPuppetDocTitle}_.
-endif::[]
 ifdef::satellite[]
-+
 include::snip_prerequisite-project-client-repository-ak.adoc[]
+* The satellite-client-openvox-for-rhel-8 or satellite-client-openvox-for-rhel-9 repository is synchronized on {ProjectServer} and enabled in the activation key you use.
```
ekohl: Similar note about the repo names as above: the exact name are incorrect and perhaps use an attribute.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-manually.adoc
```diff
@@ -1,30 +1,25 @@
...
-For more information, see xref:enabling-puppet-integration-with-{project-context}[].
-endif::[]
-ifndef::managing-configurations-puppet[]
-For more information, see {ManagingConfigurationsPuppetDocURL}enabling-puppet-integration-with-{project-context}[Enabling Puppet integration with {Project}] in _{ManagingConfigurationsPuppetDocTitle}_.
-endif::[]
 * The host must have a Puppet environment assigned to it.
 ifdef::satellite[]
 +
 include::snip_prerequisite-project-client-repository-enabled.adoc[]
+* The satellite-client-openvox-for-rhel-8 or satellite-client-openvox-for-rhel-9 repository is enabled on the host, available in the content view and the lifecycle environment of the host, and enabled for the host.
```
ekohl: Same note again about the exact names

ekohl (commented):
Overall it looks good. I hope I didn't miss anything. Biggest issue is the exact repo names that are repeated a few times.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-provisioning.adoc
aneta-petrova: Thanks! I added the correct repository names.

I don't think an attribute is necessary. If the value was different for different target builds, that would be a good use case for an attribute. As far as I know, however, the value is the same and to be documented for Satellite builds only, right? Then hiding it behind an attribute would just introduce complexity.

aneta-petrova:
@ekohl I updated the repository names. Can you please re-review?

I'd also like give space to @maximiliankolb so that he can review the changes from his side.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-provisioning.adoc
```diff
@@ -66,13 +69,13 @@ ifndef::katello,orcharhino,satellite[]
...
 Select the *boolean* type and set the value to `true`.
 endif::[]
-. Set configuration for the Puppet agent.
-* If you use an integrated Puppet server, ensure that you select a Puppet {SmartProxy}, Puppet CA {SmartProxy}, and Puppet environment when you create a host.
-* If you use a non-integrated Puppet server, either set the following host parameters in global parameters, or a host group, or when you create a host:
-** Add a host parameter named `puppet_server`, select the *string* type, and set the value to the hostname of your Puppet server, such as `puppet.example.com`.
+. Set configuration for the OpenVox agent.
+* If you use an integrated OpenVox server, ensure that you select a Puppet {SmartProxy}, Puppet CA {SmartProxy}, and Puppet environment when you create a host.
+* If you use a non-integrated OpenVox server, either set the following host parameters in global parameters, or a host group, or when you create a host:
+** Add a host parameter named `puppet_server`, select the *string* type, and set the value to the hostname of your OpenVox server, such as `openvox.example.com`.
```
aneta-petrova, line 69: Is this a strictly defined parameter that should still be called `puppet_server`? Or should we change it to `openvox_server`?

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-provisioning.adoc
```diff
@@ -1,31 +1,34 @@
...
 
-[id="Installing_and_Configuring_Puppet_Agent_during_Host_Provisioning_{context}"]
-= Installing and configuring Puppet agent during host provisioning
+[id="configuring-openvox-agent-on-a-project-host-during-host-provisioning"]
+= Configuring OpenVox agent on a {Project} host during host provisioning
 
 [role="_abstract"]
-You can install and configure the Puppet agent on a host during the provisioning process.
-A configured Puppet agent is required on the host for Puppet integration with your {Project}.
+If you prefer to install and configure the OpenVox agent on a host during the provisioning process, you can do so by adding the `puppet_setup` snippet to your provisioning template.
```
AkshayGadhaveRH:
```suggestion
If you prefer to install and configure the OpenVox agent on a host during the provisioning process, add the `puppet_setup` snippet to your provisioning template.
```

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
```diff
@@ -53,15 +51,15 @@ ifdef::foreman-el,foreman-deb[]
...
+* Add a host parameter named `puppet_server`, select the *string* type, and set the value to the hostname of your OpenVox server, such as `openvox.example.com`.
 * Optional: Add a host parameter named `puppet_ca_server`, select the *string* type, and set the value to the hostname of your Puppet CA server, such as `puppet-ca.example.com`.
-If `puppet_ca_server` is not set, the Puppet agent will use the same server as `puppet_server`.
+If `puppet_ca_server` is not set, the OpenVox agent will use the same server as `puppet_server`.
 * Optional: Add a host parameter named `puppet_environment`, select the *string* type, and set the value to the Puppet environment you want the host to use.
 
 ifndef::orcharhino[]
 +
-Until the https://bugzilla.redhat.com/show_bug.cgi?id=2177730[BZ2177730] is resolved, you must use host parameters to specify the Puppet agent configuration even in integrated setups where the Puppet server is a {SmartProxyServer}.
+Until the https://bugzilla.redhat.com/show_bug.cgi?id=2177730[BZ2177730] is resolved, you must use host parameters to specify the OpenVox agent configuration even in integrated setups where the OpenVox server is a {SmartProxyServer}.
```
AkshayGadhaveRH, line 49: This BZ is CLOSED WONTFIX, should we drop it from here? Something like

```suggestion
Use host parameters to specify the OpenVox agent configuration even in integrated setups where the OpenVox server is a {SmartProxyServer}.
```

guides/common/modules/proc_configuring-openvox-agent-on-a-host-manually.adoc
```diff
@@ -1,30 +1,26 @@
...
+** satellite-client-openvox-for-rhel-8-x86_64-rpms
+** satellite-client-openvox-for-rhel-9-x86_64-rpms
 endif::[]
-ifndef::satellite,orcharhino[]
-* Ensure a repository containing the Puppet agent is enabled on the host, for example https://apt.puppet.com/[apt.puppet.com] or https://yum.puppet.com/[yum.puppet.com].
-endif::[]
-ifdef::orcharhino[]
-* Ensure a repository containing the Puppet agent is enabled on the host, for example {client-puppet-repo-url}[Puppet agent for {client-os}].
+ifndef::satellite[]
+* Ensure a repository containing the OpenVox agent is enabled on the host.
```
AkshayGadhaveRH, line 23: Since the OpenVox agent package is provided by the `satellite-client-openvox-for-rhel-x` repos, I think we should explicitly mention the repo id/label here.

guides/common/modules/ref_openvox-versions-supported-for-integration-with-project.adoc
```diff
@@ -0,0 +1,20 @@
...
+= OpenVox versions supported for integration with {Project}
+
+[role="_abstract"]
+Before you begin with the Puppet integration, review which versions of the OpenVox server and OpenVox agent are supported for integration with {Project}.
+
+OpenVox Server 8 is supported on {ProjectServer} and {SmartProxyServers}.
+
+[IMPORTANT]
+====
+The Puppet modules you are using must be compatible with the supported version of OpenVox server.
```
AkshayGadhaveRH: Should we mention Ruby 3.2 here? It might make it easier for users to check if their modules are compatible.

AkshayGadhaveRH (changes requested):

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
aneta-petrova: I made an attempt to resolve this line in https://github.com/theforeman/foreman-documentation/pull/4763 but couldn't find a solution in the time frame I had. It's not related to the OpenVox transition so I think we can keep it as it is for now because this PR should be strictly about OpenVox.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-manually.adoc
aneta-petrova: I don't know what it is for non-satellite builds :/ However, I think it's still good to update the wording rather than keep the old Puppet reference and repo here.

guides/common/modules/ref_openvox-versions-supported-for-integration-with-project.adoc
aneta-petrova: I'm not sure, actually. Ruby 3.2 is relevant for OpenVox 8 at the moment. The specific Ruby version might change with a later update to OpenVox server. So perhaps it's best to mention Ruby 3.2 only in the upgrading guide?

aneta-petrova:
Changes to procedures are only to s/Puppet/OpenVox and to add the OpenVox repositories => I'll remove the `Needs testing` label.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-provisioning.adoc
ekohl: The code now uses `puppet_server` (https://github.com/theforeman/foreman/blob/84f6d68fa9d2ad0edf008567d8bd13e7976ed57b/app/services/foreman/renderer/scope/macros/host_template.rb#L55-L61) so renaming this is a non-trivial effort.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-manually.adoc
ekohl: Agreed, IMHO this is a good way to cover it.

guides/common/modules/ref_openvox-versions-supported-for-integration-with-project.adoc
ekohl: It's slightly more complex than that, but agreed that Ruby 3.2 is really a small detail of it. In practice you have many pieces that need to be compatible.

You have a Puppet environment, which is a collection of Puppet modules (and some other bits, but let's ignore that for now). That Puppet environment is executed in some places.

It's safe to say that all Puppet modules in a Puppet environment need to be compatible with the running server (be it Puppet or OpenVox). Also very easy to check because you know the server version. I'd suggest this minor rephrasing:

```suggestion
The Puppet modules you are using must be compatible with the running version of OpenVox server.
```

For simplicity, it's safe to say all your Puppet modules need to be compatible with all versions of the agents using your Puppet environment. In practice that means you need wider support because you usually don't upgrade all your hosts at once.

Practical example of how upgrades typically go:
* Start situation
  * foreman.example.com runs Puppetserver 7
  * host-a.example.com runs puppet-agent 7
  * host-b.example.com runs puppet-agent 7
  * Conclusion: all your modules need to be compatible with Puppet 7
* Upgrade server
  * foreman.example.com runs OpenVox server 8
  * host-a.example.com runs puppet-agent 7
  * host-b.example.com runs puppet-agent 7
  * Conclusion: all your modules need to be compatible with Puppet 7 and OpenVox 8
* Upgrade host-a
  * foreman.example.com runs OpenVox server 8
  * host-a.example.com runs OpenVox 8
  * host-b.example.com runs puppet-agent 7
  * Conclusion: all your modules need to be compatible with Puppet 7 and OpenVox 8
* Upgrade host-b
  * foreman.example.com runs OpenVox server 8
  * host-a.example.com runs OpenVox 8
  * host-b.example.com runs OpenVox 8
  * Conclusion: all your modules need to be compatible with OpenVox 8

Users MUST have a pre-upgrade server step where they need to ensure OpenVox 8 support.
Users MAY have a post-upgrade hosts step where they can drop Puppet 7 support.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-manually.adoc
```diff
@@ -52,6 +48,11 @@ ifdef::satellite[]
 ----
 # {client-package-install-el7} puppet-agent
 ----
++
+[NOTE]
+====
+The `openvox-agent` package, included in the OpenVox repository, provides the `puppet-agent` command, which is used to install the OpenVox agent.
```
ekohl: The "which is used to" part of the sentence is a bit confusing to me. Would this help to explain why we do this?
```suggestion
The `openvox-agent` package, included in the OpenVox repository, provides the `puppet-agent` command.
This is used for maximum compatibility.
```

guides/common/modules/proc_disabling-puppet-integration-with-project.adoc
```diff
@@ -15,17 +15,14 @@ This is a known issue, see the https://bugzilla.redhat.com/show_bug.cgi?id=20870
 endif::[]
 ====
 
-.Prerequisites
-* Puppet is enabled on {Project}.
```
ekohl, line 19: IMHO this is still a prerequisite. It only makes sense if you went have enabled the integration.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-manually.adoc
```diff
@@ -1,30 +1,26 @@
...
+A configured OpenVox agent is required on the host for Puppet integration with your {Project}.
 
 .Prerequisites
-* Puppet must be enabled in your {Project}.
-ifdef::managing-configurations-puppet[]
-For more information, see xref:enabling-puppet-integration-with-{project-context}[].
-endif::[]
-ifndef::managing-configurations-puppet[]
-For more information, see {ManagingConfigurationsPuppetDocURL}enabling-puppet-integration-with-{project-context}[Enabling Puppet integration with {Project}] in _{ManagingConfigurationsPuppetDocTitle}_.
-endif::[]
```
ekohl, line 17: Why is this dropped? Out of the box we don't have this enabled so IMHO we should still guide the user. Even if it's already mentioned in the abstract, the concrete links help.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-manually.adoc
```diff
@@ -1,30 +1,26 @@
...
 * The host must have a Puppet environment assigned to it.
 ifdef::satellite[]
 +
 include::snip_prerequisite-project-client-repository-enabled.adoc[]
+* {project-client-name} OpenVox repository is synchronized on {ProjectServer}, available in the content view and the lifecycle environment of the host, and enabled for the host.
+Choose from the following repositories based on the operating system version and the architecture of the host:
+** satellite-client-openvox-for-rhel-8-aarch64-rpms
+** satellite-client-openvox-for-rhel-8-ppc64le-rpms
+** satellite-client-openvox-for-rhel-8-x86_64-rpms
+** satellite-client-openvox-for-rhel-9-x86_64-rpms
```
ekohl: I don't like the duplication of very concrete information that is subject to change. If you're not going to use attributes, does a snippet at least make sense? Otherwise it can go out of sync with `guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-registration.adoc`.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-manually.adoc
```diff
@@ -1,30 +1,26 @@
...
-ifdef::managing-configurations-puppet[]
-For more information, see xref:enabling-puppet-integration-with-{project-context}[].
-endif::[]
-ifndef::managing-configurations-puppet[]
-For more information, see {ManagingConfigurationsPuppetDocURL}enabling-puppet-integration-with-{project-context}[Enabling Puppet integration with {Project}] in _{ManagingConfigurationsPuppetDocTitle}_.
-endif::[]
 * The host must have a Puppet environment assigned to it.
 ifdef::satellite[]
 +
 include::snip_prerequisite-project-client-repository-enabled.adoc[]
```
ekohl: Previously the client repository provided the package, but if you're going to sync the `{project-client-name} OpenVox` repo then you don't need the regular client repo so I think it can be dropped as a prerequisite.
```suggestion
```

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-provisioning.adoc
```diff
@@ -1,31 +1,34 @@
...
-A configured Puppet agent is required on the host for Puppet integration with your {Project}.
+If you prefer to install and configure the OpenVox agent on a host during the provisioning process, add the `puppet_setup` snippet to your provisioning template.
+A configured OpenVox agent is required on the host for Puppet integration with your {Project}.
 
 .Prerequisites
-* Puppet must be enabled in your {Project}.
-For more information, see xref:enabling-puppet-integration-with-{project-context}[].
 ifdef::satellite[]
-+
 include::snip_prerequisite-project-client-repository-ak.adoc[]
```
ekohl: Like in the other file, you don't need the client repo if you sync the the OpenVox repo:
```suggestion
```

ekohl (commented):
I only reviewed the files themselves, not the actual previews.

guides/common/modules/con_how-puppet-integrates-with-project.adoc
```diff
@@ -1,38 +1,63 @@
 :_mod-docs-content-type: CONCEPT
 
-[id="How_Puppet_Integrates_With_Project_{context}"]
+[id="how-puppet-integrates-with-project"]
```
maximiliankolb:
```suggestion
[id="how-puppet-integrates-with-{project-context}"]
```

guides/common/modules/con_how-puppet-integrates-with-project.adoc
```diff
@@ -1,38 +1,63 @@
...
+To list Puppet facts on a host, run the `puppet facts` command.
+====
+
+{Project} acts as an external node classifier (ENC) for the OpenVox server.
+Based on the facts and other definitions {Project} receives from the OpenVox server, {Project} constructs the ENC answer to the OpenVox server.
+
+Based on the ENC answer, the OpenVox server compiles a catalog for the host and sends it to the agent.
+
+The agent evaluates the system state on the host.
+If the agent finds differences between the desired state defined in the catalog and the actual state, also known as drifts, it enforces correction of the state of the host.
```
maximiliankolb:
```suggestion
If the agent finds differences between the desired state defined in the catalog and the actual state, known as drifts, it enforces correction of the state of the host.
```

suggestion (non-blocking): remove fluff

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-provisioning.adoc
```diff
@@ -1,31 +1,34 @@
 :_mod-docs-content-type: PROCEDURE
 
-[id="Installing_and_Configuring_Puppet_Agent_during_Host_Provisioning_{context}"]
-= Installing and configuring Puppet agent during host provisioning
+[id="configuring-openvox-agent-on-a-project-host-during-host-provisioning"]
+= Configuring OpenVox agent on a {Project} host during host provisioning
```
maximiliankolb:
```suggestion
[id="configuring-openvox-agent-on-a-host-during-host-provisioning"]
= Configuring OpenVox agent on a host during host provisioning
```

IMO "host" is enough; we never say "Foreman host". Please also adjust the file name.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-provisioning.adoc
```diff
@@ -1,31 +1,34 @@
...
-For more information, see xref:enabling-puppet-integration-with-{project-context}[].
 ifdef::satellite[]
-+
 include::snip_prerequisite-project-client-repository-ak.adoc[]
+* {project-client-name} OpenVox repository is synchronized on {ProjectServer} and enabled in the activation key you use.
+Choose from the following repositories, based on the operating system version and the architecture of the host you are provisioning:
+** satellite-client-openvox-for-rhel-8-aarch64-rpms
+** satellite-client-openvox-for-rhel-8-ppc64le-rpms
+** satellite-client-openvox-for-rhel-8-x86_64-rpms
+** satellite-client-openvox-for-rhel-9-x86_64-rpms
```
maximiliankolb: just fyi: EL10 is missing. By design? Also, I suggest to mention the latest first.

EL9 alt archs also miss.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-registration.adoc
```diff
@@ -1,43 +1,41 @@
 :_mod-docs-content-type: PROCEDURE
 
-[id="installing-and-configuring-puppet-agent-during-host-registration_{context}"]
-= Installing and configuring Puppet agent during host registration
+[id="configuring-openvox-agent-on-a-project-host-during-host-registration"]
+= Configuring OpenVox agent on a {Project} host during host registration
```
maximiliankolb:
```suggestion
[id="configuring-openvox-agent-on-a-host-during-host-registration"]
= Configuring OpenVox agent on a host during host registration
```

same as above.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-registration.adoc
```diff
@@ -1,43 +1,41 @@
...
 
 [role="_abstract"]
-You can install and configure the Puppet agent on the host during registration.
-A configured Puppet agent is required on the host for Puppet integration with your {Project}.
+If you prefer to install and configure the OpenVox agent on a host during the registration process, you can enable the OpenVox agent with host parameters in global parameters or a host group.
```
maximiliankolb: after "if you prefer", I suggest to drop "you can":


```suggestion
If you prefer to install and configure the OpenVox agent on a host during the registration process, enable the OpenVox agent with host parameters in global parameters or a host group.
```

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-registration.adoc
```diff
@@ -1,43 +1,41 @@
...
-endif::[]
 ifdef::satellite[]
-+
 include::snip_prerequisite-project-client-repository-ak.adoc[]
+* {project-client-name} OpenVox repository is synchronized on {ProjectServer} and enabled in the activation key you use:
+Choose from the following repositories, based on the operating system version and the architecture of the host you are registering:
+** satellite-client-openvox-for-rhel-8-aarch64-rpms
+** satellite-client-openvox-for-rhel-8-ppc64le-rpms
+** satellite-client-openvox-for-rhel-8-x86_64-rpms
+** satellite-client-openvox-for-rhel-9-x86_64-rpms
```
maximiliankolb: same as above.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-manually.adoc
```diff
@@ -1,30 +1,26 @@
 :_mod-docs-content-type: PROCEDURE
 
-[id="Installing_and_Configuring_Puppet_Agent_Manually_{context}"]
-= Installing and configuring Puppet agent manually
+[id="configuring-openvox-agent-on-a-project-host-manually"]
+= Configuring OpenVox agent on a {Project} host manually
```
maximiliankolb:
```suggestion
[id="configuring-openvox-agent-on-a-host-manually"]
= Configuring OpenVox agent on a host manually
```

same as above.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-manually.adoc
maximiliankolb: Yes, maybe

````diff
-This is used for maximum compatibility.
+This ensures compatibility of Puppet commands with OpenVox agents.
````

File: guides/common/modules/proc_installing-openvox-server-on-projectserver-and-smartproxyservers.adoc [hunk 30]
```diff
@@ -0,0 +1,32 @@
...
+----
+# {foreman-installer} \
+--enable-foreman-cli-puppet \
+--enable-foreman-plugin-puppet \
+--enable-puppet \
+--foreman-proxy-puppet true \
+--foreman-proxy-puppetca true \
+--puppet-server true
+----
+. Optional: If you want manage configuration with Puppet on {SmartProxyServers}, install an OpenVox server on {SmartProxyServers}:
```
maximiliankolb:
```suggestion
. Optional: If you want manage configuration with Puppet on {SmartProxyServers}, install OpenVox server on {SmartProxyServers}:
```

to align it with line 11.

guides/common/modules/ref_openvox-versions-supported-for-integration-with-project.adoc
```diff
@@ -0,0 +1,20 @@
+:_mod-docs-content-type: REFERENCE
```
maximiliankolb, line 1:
```suggestion
:_mod-docs-content-type: REFERENCE

```

guides/common/modules/ref_openvox-versions-supported-for-integration-with-project.adoc
```diff
@@ -0,0 +1,20 @@
...
+[id="openvox-versions-supported-for-integration-with-project"]
```
maximiliankolb:
```suggestion
[id="openvox-versions-supported-for-integration-with-{project-context}"]
```

guides/common/modules/ref_openvox-versions-supported-for-integration-with-project.adoc
```diff
@@ -0,0 +1,20 @@
...
+Before you begin with the Puppet integration, review which versions of the OpenVox server and OpenVox agent are supported for integration with {Project}.
+
+OpenVox Server 8 is supported on {ProjectServer} and {SmartProxyServers}.
+
+[IMPORTANT]
+====
+The Puppet modules you are using must be compatible with the supported version of OpenVox server.
+====
+
+OpenVox Agent 8 is supported on {Project} hosts.
```
maximiliankolb:
```suggestion
OpenVox Agent 8 is supported on hosts.
```

maximiliankolb (commented):
Thanks Anet, I made some suggestions.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-provisioning.adoc
aneta-petrova: @ekohl Can you please take a look at this? These are the repos I was made aware of in a different conversation. Is the list correct like this?

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-provisioning.adoc
aneta-petrova: Okay, the way I interpret this is that we're sticking to `puppet_server` here and elsewhere.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-manually.adoc
aneta-petrova: This prerequisite is pointing to the ID of the assembly where this module is included:

Image

This makes it an invalid prerequisite from the perspective of modular documentation.

guides/common/modules/proc_disabling-puppet-integration-with-project.adoc
aneta-petrova: We have recently discussed the concept of "obvious prerequisites" on the team recently. I'd say that "X is enabled" is an example of such an obvious prerequisite, if you're documenting how to disable something. You can't disable it if it's not enabled.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-manually.adoc
ekohl: Taking a step back: the goal is that you can also use the procedure with the old Puppet 7 agents.

The alternative is to say `dnf install openvox-agent` with a note after to use `dnf install puppet-agent` if you still use the old packages.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-manually.adoc
ekohl: But today we also have it in the Managing hosts guide (https://docs.theforeman.org/3.18/Managing_Hosts/index-katello.html#installing-and-configuring-puppet-agent-during-host-registration_managing-hosts). In your preview I also see https://theforeman-foreman-documentation-preview-pr-4766.surge.sh/nightly/Managing_Hosts/index-foreman-el.html#configuring-openvox-agent-on-a-project-host-during-host-registration that no longer lists the prerequisite.

guides/common/modules/ref_openvox-versions-supported-for-integration-with-project.adoc
aneta-petrova: That's a great write-up of the upgrade process. @AkshayGadhaveRH and @ekohl, do you think it's something that should be captured in https://github.com/theforeman/foreman-documentation/pull/4769?

guides/common/modules/proc_disabling-puppet-integration-with-project.adoc
ekohl: Fair enough. I missed that conversation.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-provisioning.adoc
ekohl: If you do that, should then also reduce it to:
```suggestion
[id="configuring-openvox-agent-on-a-host-during-provisioning"]
= Configuring OpenVox agent on a host during provisioning
```

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-provisioning.adoc
ekohl: Satellite hasn't provided EL10 builds until now. Due to time constraints we haven't done that yet and is planned for later.

@pcreech since you requested the repos, mind taking a look at the alt archs? Or overall verify these are the correct repos?

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
aneta-petrova: I filed https://redhat.atlassian.net/browse/SAT-44553 as a public tracker for the issue.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-manually.adoc
aneta-petrova: For another ticket I'm about to start working on, I have been considering removing those procedures from Managing hosts. However, it's a good catch that in that guide, the prerequisite is required for these modules so I'll re-add it, but only for Managing hosts.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-manually.adoc
aneta-petrova: Oh right, I forgot. Yes, this is a good use case for a snippet. Added.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
```diff
@@ -1,43 +1,43 @@
...
-endif::[]
 ifndef::managing-configurations-puppet[]
-For more information, see {ManagingConfigurationsPuppetDocURL}enabling-puppet-integration-with-{project-context}[Enabling Puppet integration with {Project}] in _{ManagingConfigurationsPuppetDocTitle}_.
+* Puppet must be enabled in your {Project}.
+For more information, see {ManagingConfigurationsPuppetDocURL}[_{ManagingConfigurationsPuppetDocTitle}_].
++
 endif::[]
 ifdef::satellite[]
-+
 include::snip_prerequisite-project-client-repository-ak.adoc[]
```
aneta-petrova: Shall I drop the client repository from the prerequisites in the host registration procedure as well @ekohl?
```suggestion
```

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-manually.adoc
aneta-petrova: @ekohl This procedure is for configuring a new OpenVox agent so I don't think the Puppet 7 agent scenario applies here. So perhaps it's better to just do s/dnf install puppet-agent/dnf install openvox-agent? Can you please re-review this part @ekohl?

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
ekohl: Yes, I suggested that in https://github.com/theforeman/foreman-documentation/pull/4766#discussion_r3109987044. It's now marked as outdated so I guess some lines moved around and it was missed.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-provisioning.adoc
aneta-petrova: Applied. I moved the `{Project} host` keyword to the abstract because https://redhat-documentation.github.io/supplementary-style-guide/#_style_guidelines says that product name must be included in the heading or the abstract.

guides/common/modules/ref_openvox-versions-supported-for-integration-with-project.adoc
aneta-petrova: I'd like to keep the specific reference here. We are documenting a cross-product situation (OpenVox / Foreman) so the extra keyword here and there helps.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
aneta-petrova: Thanks for the clarification. I'll remove it right away too.

aneta-petrova: And just to rant a bit, I feel like I keep losing track of comments in the new GitHub interface for comments, which is what might have happened here.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-registration.adoc
aneta-petrova: To be resolved in https://github.com/theforeman/foreman-documentation/pull/4766#discussion_r3112283039

aneta-petrova:
Thanks, @ekohl and @maximiliankolb, for all your feedback. Can you please re-review? I left a few specific questions for @ekohl, so those threads remain open.

Apart from that, I'm tracking two remaining open questions:

* There are no OpenVox requirements we could link to (https://github.com/theforeman/foreman-documentation/pull/4766#discussion_r3071639543) -- IMO not a blocker for this PR.
* A final ack on the list of repos is needed from @pcreech (https://github.com/theforeman/foreman-documentation/pull/4766#discussion_r3112283039) -- this should be resolved in this PR.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
```diff
@@ -1,43 +1,42 @@
...
 
-[id="installing-and-configuring-puppet-agent-during-host-registration_{context}"]
-= Installing and configuring Puppet agent during host registration
+[id="configuring-openvox-agent-on-a-host-during-registration"]
+= Configuring OpenVox agent on a host during registration
 
 [role="_abstract"]
-You can install and configure the Puppet agent on the host during registration.
-A configured Puppet agent is required on the host for Puppet integration with your {Project}.
+If you prefer to install and configure the OpenVox agent on a {Project} host during the registration process, enable the OpenVox agent with host parameters in global parameters or a host group.
```
maximiliankolb:
```suggestion
If you prefer to install and configure the OpenVox agent on a host during the registration process, enable the OpenVox agent with host parameters in global parameters or a host group.
```

guides/common/modules/proc_configuring-openvox-agent-on-a-host-manually.adoc
```diff
@@ -1,30 +1,26 @@
...
 
-[id="Installing_and_Configuring_Puppet_Agent_Manually_{context}"]
-= Installing and configuring Puppet agent manually
+[id="configuring-openvox-agent-on-a-host-manually"]
+= Configuring OpenVox agent on a host manually
 
 [role="_abstract"]
-You can install and configure the Puppet agent on a host manually.
-A configured Puppet agent is required on the host for Puppet integration with your {Project}.
+On {Project} hosts that are already registered, you can install and configure the OpenVox agent manually.
```
maximiliankolb:
```suggestion
On hosts that are already registered to {Project}, you can install and configure the OpenVox agent manually.
```

maximiliankolb (commented):
two more minor suggestions, overall style-wise LGTM. I cannot comment much on the tech aspects, though as far as I can tell right now, OpenVox (server and agent) is an open source drop in replacement for Puppet, so 👍

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
aneta-petrova: I'd like to keep the explicit Foreman reference here, see also https://redhat-documentation.github.io/supplementary-style-guide/#_style_guidelines: "Make modules findable and reusable. Include the product name in either the title or the short description to make the module reusable."

maximiliankolb: But we never say "Foreman host". It's either Foreman Server or host registered to Foreman.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
```diff
@@ -1,43 +1,42 @@
...
-ifdef::managing-configurations-puppet[]
-For more information, see xref:enabling-puppet-integration-with-{project-context}[].
-endif::[]
 ifndef::managing-configurations-puppet[]
-For more information, see {ManagingConfigurationsPuppetDocURL}enabling-puppet-integration-with-{project-context}[Enabling Puppet integration with {Project}] in _{ManagingConfigurationsPuppetDocTitle}_.
+* Puppet must be enabled in your {Project}.
+For more information, see {ManagingConfigurationsPuppetDocURL}[_{ManagingConfigurationsPuppetDocTitle}_].
++
 endif::[]
 ifdef::satellite[]
```
ekohl, line 16: Perhaps out of scope, but zooming out: the Satellite, Katello, and Orcharhino case are really the same. The difference is that with Satellite we want to consume the Red Hat built version of OpenVox, in Katello we can only tell the user to get it somewhere else (like https://voxpupuli.org/openvox/install/).

Please consider this because we might be able to deduplicate some instructions.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
```diff
@@ -1,43 +1,42 @@
...
+* {project-client-name} OpenVox repository is synchronized on {ProjectServer} and enabled in the activation key you use:
 +
-include::snip_prerequisite-project-client-repository-ak.adoc[]
+include::snip_openvox-client-repos.adoc[]
 * You have an activation key.
 For more information, see {ContentManagementDocURL}Managing_Activation_Keys_content-management[Managing Activation Keys] in _{ContentManagementDocTitle}_.
 endif::[]
 ifdef::katello[]
-* You created a product and repository for the upstream Puppet agent, such as `\https://yum.puppet.com` or `\https://apt.puppet.com`, and synchronized the repository to {Project}.
+* You created a product and repository for the upstream OpenVox agent.
```
ekohl: What if we use the exact orcharhino phrasing below? That looks good to me and then we can merge the katello and orcharhino cases.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
ekohl: Taking a step back: the `enable-puppet7` and `enable-puppet8` parameters are **exactly** the same. It doesn't matter what you use. We didn't clean up our templates, but should have.

Since we're updating:
```suggestion
Add a host parameter named `enable-puppet8`, select the *boolean* type, and set the value to `true`.
```

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
```diff
@@ -1,43 +1,42 @@
...
+* You created an activation key that enables the OpenVox agent repository for hosts.
 For more information, see {ContentManagementDocURL}Managing_Activation_Keys_content-management[Managing activation keys] in _{ContentManagementDocTitle}_.
 endif::[]
 
 .Procedure
 . In the {ProjectWebUI}, navigate to *Configure* > *Global Parameters* to add host parameters globally.
 Alternatively, you can navigate to *Configure* > *Host Groups* and edit or create a host group to add host parameters only to a host group.
-. Enable the Puppet agent using a host parameter in global parameters or a host group.
+. Enable the OpenVox agent using a host parameter in global parameters or a host group.
 // consolidate with line 54 as soon as Satellite client contains Puppet agent 8
```
ekohl: This looks like a TODO we need to implement.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
ekohl: We can drop this `enable-puppet7`

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
```diff
@@ -53,15 +52,15 @@ ifdef::foreman-el,foreman-deb[]
 To use Puppet 8, add a host parameter named `enable-official-puppet8-repo`, select the *boolean* type, and set the value to `true`.
 To use Puppet 7, add a host parameter named `enable-official-puppet7-repo`, select the *boolean* type, and set the value to `true`.
```
ekohl: We can drop this

guides/common/modules/proc_configuring-openvox-agent-on-a-host-manually.adoc
```diff
@@ -38,19 +34,19 @@ ifdef::orcharhino[]
...
+. Install the OpenVox agent package.
 * On hosts running {EL} 8 and above:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-# {client-package-install-el8} puppet-agent
+# {client-package-install-el8} openvox-agent
 ----
 ifdef::satellite[]
 * On hosts running {EL} 7 and below:
```
ekohl: Keeping the conversation open: today we haven't decided if we'll have EL7 packages for `openvox-agent`.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
aneta-petrova: I can't commit to consolidating at this point. Our goal now is to fix what's wrong.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
aneta-petrova: I merged the two conditionals

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
aneta-petrova: Dropped

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
aneta-petrova: Dropped

guides/common/modules/ref_openvox-versions-supported-for-integration-with-project.adoc
AkshayGadhaveRH: @aneta-petrova yeah I think it'll be better that I record it in the upgrading guide.

aneta-petrova:
Hi @ekohl, is it okay to assume that the only things left to be resolved are the list of repos and the question of EL7? Can I consider the rest tech-acked?

guides/common/modules/proc_configuring-openvox-agent-on-a-host-manually.adoc
aneta-petrova: https://github.com/theforeman/foreman-documentation/blob/bf7914027768c359772c5b53c34eaeabc555adcc/guides/common/modules/con_supported-client-architectures-for-configuration-management.adoc?plain=1#L19 is also something to keep in mind and potentially update

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-provisioning.adoc
```diff
@@ -1,31 +1,30 @@
...
 
-[id="Installing_and_Configuring_Puppet_Agent_during_Host_Provisioning_{context}"]
-= Installing and configuring Puppet agent during host provisioning
+[id="configuring-openvox-agent-on-a-host-during-provisioning"]
+= Configuring OpenVox agent on a host during provisioning
 
 [role="_abstract"]
-You can install and configure the Puppet agent on a host during the provisioning process.
-A configured Puppet agent is required on the host for Puppet integration with your {Project}.
+If you prefer to install and configure the OpenVox agent on a {Project} host during the provisioning process, add the `puppet_setup` snippet to your provisioning template.
```
maximiliankolb:
```suggestion
If you prefer to install and configure the OpenVox agent on a host during the provisioning process, add the `puppet_setup` snippet to your provisioning template.
```

maximiliankolb (commented):
I still have concerns about "{Project} hosts". What about "hosts registered to {Project}"? Maybe "hosts in {Project}"?

It's not never but rare:
````
$ rg "\{Project\} host" | wc
      8     130    1525
$ rg "host" | wc
   4351   53474  689985
````

**Comment by aneta-petrova** [comment]
> I still have concerns about "{Project} hosts". What about "hosts registered to {Project}"? Maybe "hosts in {Project}"?
> 
> It's not never but rare:
> 
> ```
> $ rg "\{Project\} host" | wc
>       8     130    1525
> $ rg "host" | wc
>    4351   53474  689985
> ```

Oh right, sorry! This is the PR where I saw that comment... I'll make sure to resolve your question as well before thinking about merging this.

File: guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-provisioning.adoc (hunk 42 above)
Comment (ekohl) [reply guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-provisioning.adoc]: The `puppet_setup` snippet is part of our default provisioning templates and I'm assuming that most users do that. IMHO it's really a small implementation detail and does that belong in the abstract?

I really like the old text more
```suggestion
You can install and configure an OpenVox agent on a host during the provisioning process.
```

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-provisioning.adoc
```diff
@@ -52,27 +51,25 @@ endif::[]
...
+* If you use a non-integrated OpenVox server, either set the following host parameters in global parameters, or a host group, or when you create a host:
+** Add a host parameter named `puppet_server`, select the *string* type, and set the value to the hostname of your OpenVox server, such as `openvox.example.com`.
 ** Optional: Add a host parameter named `puppet_ca_server`, select the *string* type, and set the value to the hostname of your Puppet CA server, such as `puppet-ca.example.com`.
-If `puppet_ca_server` is not set, the Puppet agent will use the same server as `puppet_server`.
+If `puppet_ca_server` is not set, the OpenVox agent will use the same server as `puppet_server`.
 ** Optional: Add a host parameter named `puppet_environment`, select the *string* type, and set the value to the Puppet environment you want the host to use.
 ifdef::katello,orcharhino,satellite[]
-. Ensure your host has access to the Puppet agent packages from {ProjectServer} by using an appropriate activation key.
+. Ensure your host has access to the OpenVox agent packages from {ProjectServer} by using an appropriate activation key.
```
ekohl, line 74: Pre-existing issue: this really duplicates the prerequisite

guides/common/modules/snip_openvox-client-repos.adoc
```diff
@@ -0,0 +1,8 @@
+:_mod-docs-content-type: SNIPPET
+
+Choose from the following repositories, based on the operating system version and the architecture of the host you are registering:
++
+** satellite-client-openvox-for-rhel-9-x86_64-rpms
+** satellite-client-openvox-for-rhel-8-aarch64-rpms
+** satellite-client-openvox-for-rhel-8-ppc64le-rpms
```
ekohl: @pcreech can you check if we really built these? https://yum.voxpupuli.org/openvox8/el/8/ppc64le/ only really has an ancient version

guides/common/modules/snip_openvox-client-repos.adoc
```diff
@@ -0,0 +1,8 @@
+:_mod-docs-content-type: SNIPPET
+
+Choose from the following repositories, based on the operating system version and the architecture of the host you are registering:
++
+** satellite-client-openvox-for-rhel-9-x86_64-rpms
```
ekohl, line 5: Reviewing the list now: I'm wondering why we don't have RHEL 9 aarch64 because https://yum.voxpupuli.org/openvox8/el/9/x86_64/ does exist.

ekohl (commented):
I agree with @maximiliankolb on the `{Project} hosts` part, but also think it'll be better to keep this PR simple and not introduce too many unrelated changes (regardless of when I made many comments). Overall these procedures are really messy and my brain has a hard time separating the changes.

To merge this I think we need to at least sort out the exact supported client support. Downstream I'm working with @pcreech to sort that out.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-provisioning.adoc
aneta-petrova: Unfortunately, the old text doesn't meet the requirements for abstracts. The purpose of extending it with the additional information is to distinguish it from the other -- alternative -- procedures. I'm updating it from `{Project} hosts` to `hosts registered to {Project}`.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-provisioning.adoc
aneta-petrova: While possibly true, I can't commit to fixing non-critical pre-existing issues at this stage.

aneta-petrova:
> I still have concerns about "{Project} hosts". What about "hosts registered to {Project}"? Maybe "hosts in {Project}"?
> 
Hopefully fixed, for the changes in this PR's diff. Feel free to re-review @maximiliankolb

guides/common/modules/ref_openvox-versions-supported-for-integration-with-project.adoc
```diff
@@ -13,9 +13,9 @@ OpenVox Server 8 is supported on {ProjectServer} and {SmartProxyServers}.
 The Puppet modules you use must be compatible with the OpenVox Server version you run.
 ====
 
-OpenVox Agent 8 is supported on {Project} hosts.
+OpenVox Agent 8 is supported on hosts registered to {Project}.
 
 Additionally, Puppet agent 7 remains compatible with OpenVox in {Project} as a legacy alternative.
-If your {Project} hosts run Puppet agent 7, they can still integrate with the OpenVox implementation in {Project}.
+If the hosts registered to {Project} run Puppet agent 7, they can still integrate with the OpenVox implementation in {Project}.
```
ekohl: Feel free to ignore my phrasing suggestion, but is this better?
```suggestion
Hosts registered to {Project} running Puppet agent 7 can still integrate with the OpenVox implementation in {Project}.
```

ekohl (approved):
ACK pending the resolution of the exact repository support.

guides/common/modules/ref_openvox-versions-supported-for-integration-with-project.adoc
```diff
@@ -0,0 +1,21 @@
...
+
+[IMPORTANT]
+====
+The Puppet modules you use must be compatible with the OpenVox Server version you run.
+====
+
+OpenVox Agent 8 is supported on hosts registered to {Project}.
+
+Additionally, Puppet agent 7 remains compatible with OpenVox in {Project} as a legacy alternative.
+If the hosts registered to {Project} run Puppet agent 7, they can still integrate with the OpenVox implementation in {Project}.
```
aneta-petrova: A great suggestion, let's just reduce the ambiguity just a little bit...
```suggestion
Hosts registered to {Project} that run Puppet agent 7 can still integrate with the OpenVox implementation in {Project}.
```

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-provisioning.adoc
```diff
@@ -1,31 +1,30 @@
...
 
-[id="Installing_and_Configuring_Puppet_Agent_during_Host_Provisioning_{context}"]
-= Installing and configuring Puppet agent during host provisioning
+[id="configuring-openvox-agent-on-a-host-during-provisioning"]
+= Configuring OpenVox agent on a host during provisioning
 
 [role="_abstract"]
-You can install and configure the Puppet agent on a host during the provisioning process.
-A configured Puppet agent is required on the host for Puppet integration with your {Project}.
+If you prefer to install and configure the OpenVox agent on a host registered to {Project} during the provisioning process, add the `puppet_setup` snippet to your provisioning template.
```
maximiliankolb: sorry if I suggested exactly that, but now I would want to reword it again 🙈 


```suggestion
If you prefer to install and configure the OpenVox agent on a host during the provisioning process through {Project}, add the `puppet_setup` snippet to your provisioning template.
```

Alternative: drop "through {Project}".

I think that it does not make sense to call a host "registered" if we're talking about provisioning a (new) host.

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-registration.adoc
```diff
@@ -1,67 +1,52 @@
...
 
-[id="installing-and-configuring-puppet-agent-during-host-registration_{context}"]
-= Installing and configuring Puppet agent during host registration
+[id="configuring-openvox-agent-on-a-host-during-registration"]
+= Configuring OpenVox agent on a host during registration
 
 [role="_abstract"]
-You can install and configure the Puppet agent on the host during registration.
-A configured Puppet agent is required on the host for Puppet integration with your {Project}.
+If you prefer to install and configure the OpenVox agent on a host registered to {Project} during the registration process, enable the OpenVox agent with host parameters in global parameters or a host group.
```
maximiliankolb: Same as above 🙈 


```suggestion
If you prefer to install and configure the OpenVox agent on a host during the registration process to {Project}, enable the OpenVox agent with host parameters in global parameters or a host group.
```

alternatives:
* registration process to {Project} -> registration to {Project}
* registration process to {Project} -> registration process

guides/common/modules/proc_configuring-openvox-agent-on-a-host-during-provisioning.adoc
aneta-petrova: To comply with the guidelines I mentioned elsewhere (to mention important keywords in the heading or the abstract), let's use the version with `through {Project}`. Thanks for the suggestion!

aneta-petrova:
@maximiliankolb I applied your comments, and then slightly updated the wording further. Thanks!

guides/common/modules/snip_openvox-client-repos.adoc
```diff
@@ -0,0 +1,8 @@
...
+** satellite-client-openvox-for-rhel-8-aarch64-rpms
+** satellite-client-openvox-for-rhel-8-ppc64le-rpms
+** satellite-client-openvox-for-rhel-8-x86_64-rpms
```
aneta-petrova, line 7:
```suggestion
** satellite-client-openvox-for-rhel-9-x86_64-rpms
** satellite-client-openvox-for-rhel-8-aarch64-rpms
** satellite-client-openvox-for-rhel-8-x86_64-rpms
```

guides/common/modules/snip_openvox-client-repos.adoc
pcreech: the build system only built for x86_64 for el9.  Based on discussions, it looks like we will be adding aarch64 in the near future, but shouldn't block this PR i think.

guides/common/modules/proc_configuring-openvox-agent-on-a-project-host-during-host-provisioning.adoc
pcreech: Ok, the definitive list (for now)

```
satellite-client-openvox-for-rhel-9-x86_64-rpms
satellite-client-openvox-for-rhel-8-x86_64-rpms
satellite-client-openvox-for-rhel-8-aarch64-rpms
```

This will be expanded at a later date.

guides/common/modules/snip_openvox-client-repos.adoc
pcreech: our build system in fact did build a ppc64le version of the latest build we pulled in.  We don't need to ship that build if we don't feel like it though, and can look into disabling this in the future.

guides/common/modules/snip_openvox-client-repos.adoc
pcreech: This suggested change reflects my comments earlier.  +1

pcreech:
Updated with info requested from me.  Once the suggested change is implemented, we should be G2G

aneta-petrova:
Just to be sure, because there have been a lot of comments and back-and-forth in this PR: @maximiliankolb @ekohl, are you okay with the PR in its current state? Can I merge?

maximiliankolb (approved):
Yes. Style-wise LGTM

guides/common/modules/proc_configuring-openvox-agent-on-a-host-manually.adoc
aneta-petrova: I dropped the EL7 step here, and also updated the list of client architectures for config management in https://github.com/theforeman/foreman-documentation/pull/4781/

ekohl (approved):

aneta-petrova:
List formatting wasn't right after dropping EL7 so the latest force push was to fix that.

maximiliankolb (approved):
LGTM. As a follow-up, I will create a PR to drop "and above" in https://theforeman-foreman-documentation-preview-pr-4766.surge.sh/nightly/Managing_Hosts/index-katello.html#configuring-openvox-agent-on-a-host-manually

aneta-petrova:

534ccd1855..26ed6311ec  3.18 -> 3.18

PR #4769, opened 2026-04-13
Update Upgrading guide for OpenVox

AkshayGadhaveRH:

Updating the Upgrading guide to include information about upgrading to OpenVox.

Foreman is transitioning from Puppet to OpenVox.
https://redhat.atlassian.net/browse/SAT-33215

#### Contributor checklists


guides/common/modules/proc_synchronizing-the-new-repositories.adoc
```diff
@@ -17,10 +17,16 @@ You must enable and synchronize the new {ProjectVersion} repositories before you
 ** *{RHEL} 9 (for x86_64 {endash} AppStream) (RPMs)*
 * Repositories required to upgrade {Project} clients:
 ** *{project-client-name}* for each {RHEL} version your clients are using
+** *{project-client-name} OpenVox* for each {RHEL} version your clients are using
```
AkshayGadhaveRH: Hi @ekohl what is the repo label for the OpenVox repo?

guides/common/modules/con_planning-project-upgrade.adoc
```diff
@@ -14,6 +14,7 @@ ifdef::satellite[]
 For information about changes in these tools, see the {ProjectName} {ProjectVersion} {ReleaseNotesDocURL}[{ReleaseNotesDocTitle}].
 endif::[]
 ifdef::satellite[]
+* Ensure your puppet modules are compatible with Ruby 3.2.
```
ekohl: Is it better to say this? Emphasize OpenVox 8 which happens to include Ruby 3.2
```suggestion
* Ensure your Puppet modules are compatible with OpenVox 8, which includes Ruby 3.2.
```

guides/common/modules/proc_synchronizing-the-new-repositories.adoc
ekohl: https://github.com/theforeman/foreman-documentation/pull/4766#discussion_r3081048526

guides/common/modules/con_planning-project-upgrade.adoc
aneta-petrova: Perhaps prefix this with some context so that users who are not using Puppet can be sure they don't need to worry about this item?
```suggestion
* If you use Puppet to manage your host configuration, ensure your Puppet modules are compatible with OpenVox 8, which includes Ruby 3.2.
```

guides/common/modules/proc_synchronizing-the-new-repositories.adoc
```diff
@@ -17,10 +17,16 @@ You must enable and synchronize the new {ProjectVersion} repositories before you
...
 . In the {ProjectWebUI}, navigate to *Content* > *Sync Status*.
 . Synchronize the following repositories:
 ** All repositories for version {ProjectVersion}
 ** *{project-client-name}* repositories for each {RHEL} version your clients are using
++
+[NOTE]
+====
+Add the *{project-client-name} OpenVox* repository for the {RHEL} version your clients are using to the content views and publish and promote a new version of the content views.
+====
```
aneta-petrova: This is a clear instruction -- an action for users to perform -- so I'd like to see it as a regular step somewhere. Perhaps:
```suggestion
. If you use Puppet to manage host configurations, manually update the content views for your hosts.
Add the *{project-client-name} OpenVox* repository for the {RHEL} version your clients are using to the content views and publish and promote a new version of the content views.
```
This also includes my suggestion from https://github.com/theforeman/foreman-documentation/pull/4769/changes#r3085077404 to clarify that the step is only relevant for users who use Puppet in their Foreman environment.

guides/common/modules/proc_synchronizing-the-new-repositories.adoc
```diff
@@ -17,10 +17,13 @@ You must enable and synchronize the new {ProjectVersion} repositories before you
 ** *{RHEL} 9 (for x86_64 {endash} AppStream) (RPMs)*
 * Repositories required to upgrade {Project} clients:
 ** *{project-client-name}* for each {RHEL} version your clients are using
+** *{project-client-name} OpenVox* for each {RHEL} version your clients are using
```
ekohl: This only applies to users who have enabled the Puppet feature. By default it's off. Perhaps something like this:
```suggestion
** When using the Puppet feature, *{project-client-name} OpenVox* for each {RHEL} version your clients are using
```

ekohl (approved):
Minor note inline. It doesn't hurt to sync it and the content view part already has the correct note.

guides/common/modules/proc_synchronizing-the-new-repositories.adoc
aneta-petrova: Can we make this consistent with the other "If..." condition phrases?
```suggestion
** If you use Puppet to manage host configurations: *{project-client-name} OpenVox* for each {RHEL} version your clients are using
```

guides/common/modules/proc_upgrade-puppet-agent-to-openvox-agent.adoc
```diff
@@ -0,0 +1,18 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="upgrade-puppet-agent-to-openvox-agent"]
+= Upgrade puppet-agent to openvox-agent
+
+[role="_abstract"]
+If you use puppet for host management, you have the option to upgrade `puppet-agent`, installed on your hosts, to `openvox-agent`.
```
ekohl:
```suggestion
If you use Puppet for host management, you have the option to upgrade `puppet-agent`, installed on your hosts, to `openvox-agent`.
```

guides/common/modules/proc_upgrade-puppet-agent-to-openvox-agent.adoc
```diff
@@ -0,0 +1,18 @@
...
+
+.Prerequisites
+* *{project-client-name} OpenVox* repository is enabled on the host.
```
ekohl: For upstream this will be wrong. Perhaps we should keep the procedure specific to Satellite by hiding it behind `ifdef` statements in the `master.adoc` files?

guides/common/modules/proc_upgrade-puppet-agent-to-openvox-agent.adoc
```diff
@@ -0,0 +1,18 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="upgrade-puppet-agent-to-openvox-agent"]
+= Upgrade puppet-agent to openvox-agent
+
+[role="_abstract"]
+If you use Puppet for host management, you have the option to upgrade `puppet-agent`, installed on your hosts, to `openvox-agent`.
```
maximiliankolb: Is "upgrade" the correct term? This sounds like we value OpenVox higher than Puppet.

ekohl: It upgrades from version 7 to 8 as well so IMHO it is an upgrade.

maximiliankolb: Fair enough.

guides/common/modules/proc_upgrade-puppet-agent-to-openvox-agent.adoc
```diff
@@ -0,0 +1,18 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="upgrade-puppet-agent-to-openvox-agent"]
+= Upgrade puppet-agent to openvox-agent
```
aneta-petrova: I'm not convinced we need a full blown procedure for a single step. Have you considered adding this as a sub-step to the list of post-upgrade tasks that's part of the server upgrade procedure?

AkshayGadhaveRH: That is what I was considering, but the problem is this has to happen after the repos are enabled in `Synchronizing the new repositories`. Another (probably trivial) problem is that this step has to be performed on the hosts and not on the foreman server itself.

guides/common/modules/proc_upgrade-puppet-agent-to-openvox-agent.adoc
AkshayGadhaveRH: We are doing a package update from `puppet-agent 7` to `openvox-agent 8`, should I mention the package versions explicitly?

AkshayGadhaveRH: Apologies, missed that this was marked as resolved.

guides/common/modules/proc_upgrade-puppet-agent-to-openvox-agent.adoc
aneta-petrova: Oh, the dependency on syncing the repos is something I missed... okay then! Thanks for considering it :)

guides/common/modules/proc_upgrade-puppet-agent-to-openvox-agent.adoc
aneta-petrova: You could simplify the wording a little bit and I also think it would be better to refer to "product names" (with version) rather than package names here. Using more generic terms seems like a better choice for an abstract:

```suggestion
If you use Puppet for host management, you can upgrade from Puppet agent 7 to OpenVox agent 8.
```

guides/common/modules/proc_upgrade-puppet-agent-to-openvox-agent.adoc
```diff
@@ -0,0 +1,18 @@
...
+
+[role="_abstract"]
+If you use Puppet for host management, you have the option to upgrade `puppet-agent`, installed on your hosts, to `openvox-agent`.
+
```
aneta-petrova: The abstract says that the procedure is optional, but it's not clear when users should take these steps. How do they decide whether they want to upgrade to OpenVox agent or not? Can we add something like this to provide the additional context?

```suggestion
Puppet agent 7 remains compatible with OpenVox in {Project} as a legacy alternative, but {Team} recommends to upgrade to OpenVox agent 8.
For more information, see [link to https://docs.theforeman.org/3.18/Managing_Configurations_Puppet/index-satellite.html#openvox-versions-supported-for-integration-with-satellite].
```

guides/common/modules/proc_upgrade-puppet-agent-to-openvox-agent.adoc
aneta-petrova: The heading should also be updated to match.

guides/common/modules/proc_synchronizing-the-new-repositories.adoc
```diff
@@ -17,10 +17,13 @@ You must enable and synchronize the new {ProjectVersion} repositories before you
 ** *{RHEL} 9 (for x86_64 {endash} AppStream) (RPMs)*
 * Repositories required to upgrade {Project} clients:
 ** *{project-client-name}* for each {RHEL} version your clients are using
+** If you use Puppet to manage host configurations, *{project-client-name} OpenVox* for each {RHEL} version your clients are using
```
Lennonka, line 20: I strongly advise to create a new attribute for the new repository.

Lennonka (changes requested):

guides/common/modules/proc_upgrade-puppet-agent-7-to-openvox-agent-8.adoc
```diff
@@ -0,0 +1,20 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="upgrade-puppet-agent-7-to-openvox-agent-8"]
+= Upgrade Puppet agent 7 to OpenVox agent 8
+
+[role="_abstract"]
+If you use Puppet for host management, you have the option to upgrade `puppet agent 7 to OpenVox agent 8.
```
maximiliankolb:
```suggestion
If you use Puppet to configure hosts, you have the option to upgrade `puppet agent 7 to OpenVox agent 8.
```

guides/common/modules/proc_upgrade-puppet-agent-7-to-openvox-agent-8.adoc
```diff
@@ -0,0 +1,20 @@
...
+Puppet agent 7 remains compatible with OpenVox in {Project} as a legacy alternative, but {Team} recommends to upgrade to OpenVox agent 8.
+For more information, see {ManagingConfigurationsPuppetDocURL}openvox-versions-supported-for-integration-with-satellite[OpenVox versions supported for integration with Satellite] in _{ManagingConfigurationsPuppetDocTitle}_.
```
maximiliankolb:
```suggestion
For more information, see {ManagingConfigurationsPuppetDocURL}openvox-versions-supported-for-integration-with-satellite[OpenVox versions supported for integration with {Project}] in _{ManagingConfigurationsPuppetDocTitle}_.
```

guides/common/modules/proc_synchronizing-the-new-repositories.adoc
aneta-petrova: We discussed this in https://github.com/theforeman/foreman-documentation/pull/4766#discussion_r3081048526 and the conclusion was not to create an attribute.

guides/common/attributes-satellite.adoc
```diff
@@ -181,5 +181,6 @@
 :RepoRHEL7ServerSoftwareCollections: rhel-server-rhscl-7-rpms
 
 :project-client-name: Red Hat Satellite Client 6
+:project-client-openvox: {project-client-name} OpenVox
```
aneta-petrova: Adding this attribute now, with https://github.com/theforeman/foreman-documentation/pull/4766/changes having been merged without it after concluding in the comment threads there that it's unnecessary, now creates inconsistency. Can you please revert creating the attribute here so that all OpenVox documentation is consistent in how it refers to the repo across the docs set? @AkshayGadhaveRH

guides/common/modules/proc_upgrade-puppet-agent-7-to-openvox-agent-8.adoc
```diff
@@ -0,0 +1,20 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="upgrade-puppet-agent-7-to-openvox-agent-8"]
+= Upgrade Puppet agent 7 to OpenVox agent 8
+
+[role="_abstract"]
+If you use Puppet to configure hosts, you have the option to upgrade `puppet agent 7 to OpenVox agent 8.
+Puppet agent 7 remains compatible with OpenVox in {Project} as a legacy alternative, but {Team} recommends to upgrade to OpenVox agent 8.
+For more information, see {ManagingConfigurationsPuppetDocURL}openvox-versions-supported-for-integration-with-satellite[OpenVox versions supported for integration with {Project}] in _{ManagingConfigurationsPuppetDocTitle}_.
```
aneta-petrova:
```suggestion
If you use Puppet to configure hosts, you can upgrade Puppet agent 7 to OpenVox agent 8.

Puppet agent 7 remains compatible with OpenVox in {Project} as a legacy alternative, but {Team} recommends to upgrade to OpenVox agent 8.
```

AkshayGadhaveRH: @aneta-petrova I am guessing this was supposed to be for line 7 and 8?

aneta-petrova: Oh yes, it was. Not sure what happened.

aneta-petrova (approved):

guides/common/modules/proc_upgrade-puppet-agent-7-to-openvox-agent-8.adoc
```diff
@@ -0,0 +1,21 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="upgrade-puppet-agent-7-to-openvox-agent-8"]
+= Upgrade Puppet agent 7 to OpenVox agent 8
+
+[role="_abstract"]
+If you use Puppet to configure hosts, you can upgrade Puppet agent 7 to OpenVox agent 8.
+
+Puppet agent 7 remains compatible with OpenVox in {Project} as a legacy alternative, but {Team} recommends to upgrade to OpenVox agent 8.
+For more information, see {ManagingConfigurationsPuppetDocURL}openvox-versions-supported-for-integration-with-satellite[OpenVox versions supported for integration with {Project}] in _{ManagingConfigurationsPuppetDocTitle}_.
```
maximiliankolb:
```suggestion
For more information, see {ManagingConfigurationsPuppetDocURL}openvox-versions-supported-for-integration-with-{project-context}[OpenVox versions supported for integration with {Project}] in _{ManagingConfigurationsPuppetDocTitle}_.
```

guides/common/modules/proc_upgrade-puppet-agent-7-to-openvox-agent-8.adoc
```diff
@@ -0,0 +1,21 @@
...
+
+.Prerequisites
+* *{project-client-name} OpenVox* repository is enabled on the host.
+
+.Procedure
+* On your hosts, upgrade the `puppet-agent` agent package to `openvox-agent`:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+# {project-package-update} puppet-agent
```
ekohl: You can't use this: `{project-package-update}` is only for Satellite and Capsule machines. It needs to be `client-package-update-el8`:
```suggestion
# {client-package-update-el8} puppet-agent
```

Users can also run `project-package-update` without specifying any package.

guides/common/modules/proc_upgrade-puppet-agent-7-to-openvox-agent-8.adoc
```diff
@@ -0,0 +1,21 @@
...
+[role="_abstract"]
+If you use Puppet to configure hosts, you can upgrade Puppet agent 7 to OpenVox agent 8.
+
+Puppet agent 7 remains compatible with OpenVox in {Project} as a legacy alternative, but {Team} recommends to upgrade to OpenVox agent 8.
+For more information, see {ManagingConfigurationsPuppetDocURL}openvox-versions-supported-for-integration-with-satellite[OpenVox versions supported for integration with {Project}] in _{ManagingConfigurationsPuppetDocTitle}_.
+
+.Prerequisites
+* *{project-client-name} OpenVox* repository is enabled on the host.
+
+.Procedure
```
ekohl, line 15: Note a blocker, but this is a very manual way. Shouldn't we call out that you can also use REX to do it in bulk? I'm struggling to find any good existing documentation to do it but manually doing this for 100s or 1000s of servers isn't really a good experience.

Note that you upgrade all packages (without specifying a specific name) then it'll also be upgraded.

ekohl (dismissed):
I'm questioning if we need an entire procedure. I'd expect users to be familiar with packaging updates. Would it be sufficient to tell users something to the tune of:
> After enabling the {project-client-name} OpenVox repository on a host, you need to upgrade the `puppet-agent` package. Upgrading all packages also works.

guides/common/modules/proc_upgrade-puppet-agent-to-openvox-agent.adoc
aneta-petrova: @ekohl Here is where we discussed the question of whether to have a full procedure or not.

guides/common/modules/proc_upgrade-puppet-agent-7-to-openvox-agent-8.adoc
```diff
@@ -0,0 +1,23 @@
...
+* On your hosts, upgrade the `puppet-agent` agent package to `openvox-agent`:
++
+[options="nowrap" subs="+quotes,verbatim,attributes"]
+----
+# {client-package-update-el8} puppet-agent
+----
```
maximiliankolb, line 21:
```suggestion
----
+
```

guides/common/modules/proc_upgrade-puppet-agent-7-to-openvox-agent-8.adoc
aneta-petrova: There is a precedent for how to refer to rex jobs directly in this guide, actually: https://github.com/theforeman/foreman-documentation/blob/3cc18a41069c1166fa7892f35bf6f0311172a4e4/guides/common/modules/proc_upgrading-smartproxy-server.adoc?plain=1#L14-L20  Perhaps if we knew the right rex job to call, we could add a similar admonition?

guides/common/modules/proc_upgrade-puppet-agent-7-to-openvox-agent-8.adoc
```diff
@@ -0,0 +1,23 @@
...
+You can also use remote execution to upgrade the package on multiple hosts.
+For more information, see {ManagingHostsDocURL}Upgrading_a_package_managing-hosts[Upgrading a package] in _{ManagingHostsDocTitle}_.
```
ekohl, line 24: I had a look at that procedure to, but it's not really what I had in mind. That is a procedure to update a package for a single host, but IMHO you really want to do a bulk action. That usually starts with a host group since they often share content views. However, I couldn't find a procedure that actually does that.

maximiliankolb (approved):
LGTM. I will create a follow-up PR to make the procedure work for all builds/Client OS.

lpramuk (approved):

aneta-petrova:

* 36c22e098b..0069aa3691  3.18 -> 3.18

aneta-petrova:
Wait, was this intended for 3.18 only? Should I revert the commit on master?

aneta-petrova:
> Wait, was this intended for 3.18 only? Should I revert the commit on master?

This was a Satellite-only change so it doesn't break anything for upstream. No change needed right now.

PR #4795, opened 2026-04-27
Limit PR labeling

Lennonka:

- Limiting PRs that will be labeled as **Needs tech review** and **Needs style review** to changes under the `guides/` folder
- Limiting PRs that will be labeled as **Needs testing** to changes in procedure modules

- Changes outside the `guides/` folder don't really need formal tech/style reviews, just community agreement
- Testing is typically only needed for procedures

#### Contributor checklists

Please cherry-pick my commits into: **Not sure how far we label**

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,10 +1,10 @@
 Needs tech review:
 - changed-files:
-  - any-glob-to-any-file: ['**']
+  - any-glob-to-any-file: ['guides/**']
 
 Needs style review:
 - changed-files:
-  - any-glob-to-any-file: ['**']
+  - any-glob-to-any-file: ['guides/**']
 
 Needs testing:
 - changed-files:
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -8,4 +8,4 @@ Needs style review:
 
 Needs testing:
 - changed-files:
-  - any-glob-to-any-file: ['guides/**']
+  - any-glob-to-any-file: ['guides/common/assembly_*.adoc','guides/common/modules/proc_*.adoc']


Lennonka:
I don't understand why a guide is affected by this change... the diff should be clean.

maximiliankolb (approved):
LGTM; more input welcome.

.github/labeler.yml
```diff
@@ -1,11 +1,11 @@
...
 
 Needs style review:
 - changed-files:
-  - any-glob-to-any-file: ['**']
+  - any-glob-to-any-file: ['guides/**']
 
 Needs testing:
 - changed-files:
-  - any-glob-to-any-file: ['guides/**']
+  - any-glob-to-any-file: ['guides/common/modules/proc_*.adoc']
```
ekohl: I think this misses some things. An assembly can be changed to include another pre-existing procedure (that isn't modified), resulting in some final change that does need to be tested. I'd be safe and accept some false positives over false negatives.

ekohl:
> I don't understand why a guide is affected by this change... the diff should be clean.

Perhaps a race condition between merges?

jafiala:
@Lennonka This is just a change to which labels are assigned automatically, correct? We can still manually add "needs testing" to whichever PR we need, right?

.github/labeler.yml
Lennonka: Thanks, Ewoud. In that case, I'll add assemblies as well.

Lennonka:
@jafiala Correct.

.github/labeler.yml
Lennonka:
```suggestion
  - any-glob-to-any-file: ['guides/common/assembly_*.adoc','guides/common/modules/proc_*.adoc']
```

maximiliankolb (approved):

.github/labeler.yml
ekohl: I think you missed my point. I was suggesting to not change the line because the `master.adoc` files could also make a difference. Or attributes. If anything, it's probably `guides/**/*.adoc`.

Lennonka: @aneta-petrova What is your opinion on this?

Lennonka: @ekohl I disagree in principle because I don't think we should make functional changes in such a large scope. When we change `master.adoc`, it's most often a restructuring without functional changes. With attributes, I'm not sure, that's less than 50 % where attributes would be changed that affects procedures. And when the attribute change does affect procedures, it goes with changes in steps -> proc_* files. I'd go with adding the label manually for edge cases.

Lennonka: We can always try it out and change it later ;)

aneta-petrova: > @aneta-petrova What is your opinion on this?

I haven't had a chance to take a closer look so I don't have one :) There are four clever people trying to get this right already, I don't think there's a need to throw my brain into the mix too ;)

ekohl: > @ekohl I disagree in principle because I don't think we should make functional changes in such a large scope. When we change `master.adoc`, it's most often a restructuring without functional changes. With attributes, I'm not sure, that's less than 50 % where attributes would be changed that affects procedures. And when the attribute change does affect procedures, it goes with changes in steps -> proc_* files. I'd go with adding the label manually for edge cases.

And I'm still worried about missing those changes and think it's more important to avoid false negatives at the cost of some false positives.

There's no reason it can't be dropped later when there are too many false positives, but I think the changes to `master.adoc` and attributes are rare enough that still deserve a special look.

ekohl:
> This is just a change to which labels are assigned automatically, correct?

It only affects creation:
https://github.com/theforeman/foreman-documentation/blob/9a807614380dd673c64bcb93510aaaac8bb337cd/.github/workflows/labeler.yml#L8-L11
So if you open an empty PR then it won't attach any labels, even if you later add files.

.github/labeler.yml
Lennonka: "Special look" is what the tech review is for, which is labeled for every change under `guides/**`. Testing means going through a procedure step-by-step.

Lennonka:
I'm going to merge this so that we can try it out. If we're not happy with it later, we can always change it back. I'll keep those commits separate.

And we still can add the labels manually if needed.

Lennonka:
Rebased.

Lennonka:
Cherry picked:
- b24d855bf4..cde3e1d8a9  3.19 -> 3.19
- 34ffcce4ad..a8b35d9013  3.18 -> 3.18
- ac2d976a27..dda2b455d1  3.17 -> 3.17
- 1ae949b7aa..233b8f6d20  3.16 -> 3.16
- 3a3d89a30d..10d2bd060e  3.15 -> 3.15
- 3e9b2230af..d4da47f677  3.14 -> 3.14

PR #4804, opened 2026-04-30
Ensure attribute links get picked up by ConceptLink

aneta-petrova:

Moving `https://` out of link attributes and into the modules.

https://github.com/jhradilek/asciidoctor-dita-vale/issues/186

Per asciidoctor-dita-vale rules, links and cross-references are not allowed in concept modules. Right now, we often include links as attributes (such as `{ManagingContentDocURL}`) and these don't get picked up by the ConceptLink rule. Moving the `https://` URI from attributes to the text ensures that the links get picked up by Vale.

#### Contributor checklists

### DIFF
--- a/guides/common/attributes-base.adoc
+++ b/guides/common/attributes-base.adoc
@@ -31,7 +31,7 @@
 :UpgradingDisconnectedDocURL: {BaseURL}Upgrading_Project_Disconnected/{BaseFilenameURL}#
 
 // Red Hat specific URLs
-:RHDocsBaseURL: https://docs.redhat.com/en/documentation/
+:RHDocsBaseURL: docs.redhat.com/en/documentation/
 :RHELDocsBaseURL: {RHDocsBaseURL}red_hat_enterprise_linux/
 
 // Repositories and subscriptions (used both upstream and Satellite guides)
--- a/guides/common/attributes-orcharhino.adoc
+++ b/guides/common/attributes-orcharhino.adoc
@@ -2,7 +2,7 @@
 
 :upstream-ProjectVersion: {ProjectVersion}
 // Overrides for orcharhino build
-:BaseURL: https://docs.orcharhino.com/
+:BaseURL: docs.orcharhino.com/
 :ProjectVersion: 7.9
 :ProjectVersionPrevious: 7.8
 :ProjectVersionPrevious-Previous: 7.7
--- a/guides/common/attributes.adoc
+++ b/guides/common/attributes.adoc
@@ -65,7 +65,7 @@ ifeval::["{build}" == "orcharhino"]
 endif::[]
 
 // Load base attributes
-:SiteURL: https://docs.theforeman.org
+:SiteURL: docs.theforeman.org
 :BaseURL: {SiteURL}/{ProjectVersion}/
 include::attributes-base.adoc[]
 include::attributes-typography.adoc[]
--- a/guides/common/modules/con_user-groups.adoc
+++ b/guides/common/modules/con_user-groups.adoc
@@ -6,6 +6,6 @@
 [role="_abstract"]
 With {Project}, you can assign permissions to groups of users.
 You can also create user groups as collections of other user groups.
-If you use an external authentication source, you can map {Project} user groups to external user groups as described in {ConfiguringUserAuthenticationDocURL}Configuring_External_User_Groups_authentication[Configuring external user groups] in _{ConfiguringUserAuthenticationDocTitle}_.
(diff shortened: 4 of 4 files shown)


aneta-petrova:
@maximiliankolb Can you please take a quick look at this? Would this work for you? See the description for context and the problem I'm trying to solve.

aneta-petrova:
Including `https://` directly in the text of our modules triggers the ConceptLink rule, which is what I'm trying to accomplish:

Screenshot From 2026-04-30 09-50-48

The consequence of this change would be that we'd have to include `https://` in front of every `{GuideNameDocURL}` attribute from now on.

jafiala:
Makes sense to me.

But shouldn't this be cherry-picked into all to-be-migrated versions?

jafiala (approved):

aneta-petrova:
> Makes sense to me.
> 
> But shouldn't this be cherry-picked into all to-be-migrated versions?

Oh yes, and also applied to all occurrences of all guide attributes... :melting_face: I just first want to find out if this solution is acceptable to everyone on the team.

maximiliankolb:
> @maximiliankolb Can you please take a quick look at this? Would this work for you? See the description for context and the problem I'm trying to solve.

@aneta-petrova This would break downtream builds for me because I overwrite the attribute with a relative path, not a URL. I am unsure if this is working 100% for upstream too.

I hope that there's a more elegant solution that this prefix. IMO this would be a huge effort to patch in upstream docs.

However, if this is the only way to make it work, I am confident to be able to revert this at build time in my downstream build:

````
for a list of all modules:
    for a list of DocURL attributes:
        sed "s|https://attribute|attribute|g" module.adoc
````

Your goal is to make a GHA fail if there are any xrefs/DocURLs in concept modules anywhere (but under ".Additional resources")?

aneta-petrova:
> > @maximiliankolb Can you please take a quick look at this? Would this work for you? See the description for context and the problem I'm trying to solve.
> 
> @aneta-petrova This would break downtream builds for me because I overwrite the attribute with a relative path, not a URL. I am unsure if this is working 100% for upstream too.
> 
> IMO this would be a huge effort to patch in upstream docs.

Why do you think so? The way I imagined this was to simply drop `https://` from our URL attributes and prepending it in front of each `DocURL` attribute. Am I missing something?

> 
> However, if this is the only way to make it work, I am confident to be able to revert this at build time in my downstream build:
> 
> ```
> for a list of all modules:
>     for a list of DocURL attributes:
>         sed "s|https://attribute|attribute|g" module.adoc
> ```
> 
> Your goal is to make a GHA fail if there are any xrefs/DocURLs in concept modules anywhere (but under ".Additional resources")?

Yes. I based this solution on https://github.com/jhradilek/asciidoctor-dita-vale/issues/186#issuecomment-4346510539.

maximiliankolb:
"huge" is maybe an exaggeration but I count 112 instances:

````
$ fd con_ | xargs rg "DocURL" | wc
    112    1216   24102
````

> Am I missing something?

No, that should be it. I am just wondering (openly) if there is another solution because to me, it feels very un-elegant.

Just to clarify: Are links in concepts under "Additional resources" OK?

aneta-petrova:
> Just to clarify: Are links in concepts under "Additional resources" OK?

They are.

Unless that concept is an assembly introduction, because assembly introduction cannot include Additional resources.

aneta-petrova:
> I am just wondering (openly) if there is another solution because to me, it feels very un-elegant.

So was I, which is in part why I opened this PR. I welcome suggestions for alternatives. This is the only one I can think of.

aneta-petrova:
> > Just to clarify: Are links in concepts under "Additional resources" OK?
> 
> They are.
> 
> Unless that concept is an assembly introduction, because assembly introduction cannot include Additional resources.

I was wrong! :partying_face: Assembly intros can include additional resources.

https://github.com/jhradilek/asciidoctor-dita-vale/issues/186#issuecomment-4352673657

This will make addressing the requirements of the rule much easier and less disruptive.

maximiliankolb:
> Assembly intros

as in "text within an assembly" or "first concept with leveloffset 0 in an assembly"?

aneta-petrova:
> > Assembly intros
> 
> as in "text within an assembly" or "first concept with leveloffset 0 in an assembly"?

Both :) But for our team, only the second one is relevant. The first concept with leveloffset 0 in assembly can contain an `.Additional resources` section.

aneta-petrova:
I'll record the result of this conversation, along with the proposed solution, in a downstream ticket for further processing. For now, we can close this PR; once we decided on whether and how to implement it, we can reopen it... or raise a new one.

PR #4802, opened 2026-04-29
Flatten nested assemblies

maximiliankolb:
#### What changes are you introducing?

This patch ensures that no assemblies include other assemblies.

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

This is necessary to make docs work for orcharhino downstream.

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)

````
$ fd assembly_ | xargs rg -i "assembly_"
````

#### Contributor checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.18/Katello 4.20 (Satellite 6.19)
* [ ] Foreman 3.17/Katello 4.19
* [ ] Foreman 3.16/Katello 4.18 (Satellite 6.18; orcharhino 7.6 and 7.7)
* [ ] Foreman 3.15/Katello 4.17
* [ ] Foreman 3.14/Katello 4.16 (Satellite 6.17; orcharhino 7.4; orcharhino 7.5)
* [ ] Foreman 3.13/Katello 4.15 (EL9 only)
* [ ] Foreman 3.12/Katello 4.14 (Satellite 6.16; orcharhino 7.2 on EL9 only; orcharhino 7.3)
* We do not accept PRs for Foreman older than 3.12.

PR #4806, opened 2026-04-30
Cherry pick PR to flatten assemblies to 3.18

maximiliankolb:
Cherry-pick https://github.com/theforeman/foreman-documentation/pull/4802 again after reverting it to get GHA green again.

The second and third commit is based on https://github.com/theforeman/foreman-documentation/pull/4805


aneta-petrova:
@maximiliankolb I updated https://github.com/theforeman/foreman-documentation/pull/4805/changes with what I think is a better way to resolve the conflict. Can you please take a look and re-include the reference module rather than adjust the ID?

aneta-petrova (approved):

PR #4793, opened 2026-04-27
Split Registering hosts into multiple assemblies

aneta-petrova:
#### What changes are you introducing?

Reorganizing the existing "Registering hosts and setting up host integration" assembly and its included assemblies:
* Two assemblies, one for registering and the other for setting up host registration, now exist
* "Registering hosts to Foreman" now includes one assembly per registration method
* "Setting up host registration" groups post-registration tasks around actions like configuring tools
* The overview of registration methods is now gone and the information spread among various assembly introductions

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

To separate the user stories more clearly and present them in focused sub-assemblies.

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)

I used the skill file proposed in https://github.com/theforeman/foreman-documentation/pull/4758 to do most of the heavy lifting

#### Contributor checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.18/Katello 4.20 (Satellite 6.19)
* [x] Foreman 3.17/Katello 4.19
* [x] Foreman 3.16/Katello 4.18 (Satellite 6.18; orcharhino 7.6 and 7.7)
* [ ] Foreman 3.15/Katello 4.17
* [ ] Foreman 3.14/Katello 4.16 (Satellite 6.17; orcharhino 7.4; orcharhino 7.5)
* [ ] Foreman 3.13/Katello 4.15 (EL9 only)
* [ ] Foreman 3.12/Katello 4.14 (Satellite 6.16; orcharhino 7.2 on EL9 only; orcharhino 7.3)
* We do not accept PRs for Foreman older than 3.12.

PR #4809, opened 2026-05-04
wip: TOC checking CI

adamruzicka:
TODOs:
- [x] properly fill in the description template
- [ ] rename the individual jobs to something more sensible

Adding a CI check that verifies that all the headings that the satellite branding plugin links to remain in place. It does so by:
- building the satellite flavour of the docs
- extracting TOC and aliases from the built html
- runs the TOC checking rake from foreman_theme_satellite

Every now and then seemingly innocent changes break the documentation link, leading to unnecessary back and forth. This change moves the check closer to the source.

For the time being, this is just a proof of concept. We may eventually find out it is too heavy handed.

#### Contributor checklists

This should go all the way back to 3.16 probably, but there will be some changes needed to get the branches right.

### DIFF
--- a/.github/workflows/linkchecker.yml
+++ b/.github/workflows/linkchecker.yml
@@ -47,3 +47,53 @@ jobs:
       - name: Check HTML links of all builds
         run: |
           make linkchecker-tryer
+
+  theme-link-check:
+    runs-on: ubuntu-24.04
+    defaults:
+      run:
+        shell: bash
+        working-directory: guides
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ env.RUBY_VERSION }}
+          bundler-cache: true
+
+      - name: Build satellite docs
+        run: |
+          bundle exec make -j ${{ env.MAKE_J }} html BUILD=satellite LINKS=local
+
+      - name: Extract TOC
+        run: |
+          mkdir -p build/ci-toc
+          bundle exec ruby scripts/extract_links_toc \
+              --f2l upstream_filename_to_satellite_link.json \
+              --output-toc-file build/ci-toc/toc.json \
+              --output-aliases-file build/ci-toc/aliases.json \
+              build/**/index-satellite.html
+
+      - name: Stage documentation TOC for CI artifact
+        run: |
+          mkdir -p "${GITHUB_WORKSPACE}/toc-artifact"
+          cp build/ci-toc/toc.json build/ci-toc/aliases.json "${GITHUB_WORKSPACE}/toc-artifact/"
+
+      - name: Upload documentation TOC
+        uses: actions/upload-artifact@v7
+        with:
+          name: documentation-toc
+          path: toc-artifact
+          if-no-files-found: error
+
+  foreman-theme-satellite-plugin:
+    needs: theme-link-check
+    # TODO
+    # uses: theforeman/foreman_theme_satellite/.github/workflows/toc-check.yml@develop
+    uses: adamruzicka/foreman_theme_satellite/.github/workflows/toc-check.yml@ci
+    secrets: inherit
+    with:
+      documentation_toc_artifact: documentation-toc
--- a/.github/workflows/linkchecker.yml
+++ b/.github/workflows/linkchecker.yml
@@ -66,6 +66,9 @@ jobs:
 
       - name: Build satellite docs
         run: |
+          # Undo the "satellite is always containerized" overrides
+          sed -e '/:containerized:/d' -e '/include::attributes-containerized/d' -i common/attributes-satellite.adoc
+
           bundle exec make -j ${{ env.MAKE_J }} html BUILD=satellite LINKS=local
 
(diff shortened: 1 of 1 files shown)


Lennonka:
Huh, that's pretty good in principle. Thank you!

How difficult would it be to get an annotation on each module that would break a UI link, saying "You're about to break an UI link by changing this ID. Perhaps provide an alias in an auxiliary ID."?
It would be more user friendly for writers. A nice-to-have :)

Ps. I can help with job naming once we reach a conclusion.

adamruzicka:
> How difficult would it be to get an annotation on each module that would break a UI link, saying "You're about to break an UI link by changing this ID. Perhaps provide an alias in an auxiliary ID."?
> It would be more user friendly for writers. A nice-to-have :)

Because the link checking goes through two intermediate steps (adoc templates -> html -> toc), we lose the information about "this ID was in file F on line XY". However, since we have the list of "broken links" and the original diff, we should be able to piece it together in a best-effort fashion. It would still be a little bit tricky, but doable.

tl;dr: very difficult to get it 100% right every time, slightly involved to get it mostly right most of the times. Either way, I'd prefer to leave it as a followup

Lennonka:
> leave it as a followup

Totally cool with that.

adamruzicka:
Now I'd like to also turn this green. The tests are currently red due to recent rex guides restructuring in https://github.com/theforeman/foreman-documentation/pull/4753/files#issuecomment-4327312094 . Would just changing the leveloffsets do the trick or would that cause things to break somewhere else?

Lennonka:
Do you mean changing leveloffsets in the toc generation? That might work, but I didn't check.
We can't change the leveloffsets in docs.

Lennonka:
Looks like that fixed most problems, that's good!

However, the following issue will have to be fixed in the theme itself:
```
FAILED: Cannot find managing_hosts/index#Registering_Hosts_by_Using_Global_Registration_managing-hosts in TOC for entry: /Managing_Hosts/registering-a-host_managing-hosts
```
Global registration has been [remade recently](https://github.com/theforeman/foreman-documentation/pull/4793/changes#diff-acca9d9d1087be8ef3b1bf1a484d4a62e9d13c6084559dcf5c057833a762cbcdR3).

adamruzicka:
Going one level deeper in the toc seems to help, but I'd be slightly worried about (back)portability of this approach. The rex changes went all the way back to 3.16, in 3.16 the link checking mechanism works slightly differently so there might be some extra hurdles to overcome. This is a technical problem that can be resolved by a technical solution, but there is also a usability aspect to this.

If I recall correctly, the decision to limit the TOC to only upper three levels was to keep the surface area small (and ideally stable) and make sure that the in-project documentation links point to high level sections rather than pointing to overly low-level sections.

Also, [PR#4793](https://github.com/theforeman/foreman-documentation/pull/4793) brought us this this which will require changes in f-theme for develop and most likely auxiliary ids in stable branches of the docs. cc @aneta-petrova 
```
FAILED: Cannot find managing_hosts/index#Registering_Hosts_by_Using_Global_Registration_managing-hosts in TOC for entry: /Managing_Hosts/registering-a-host_managing-hosts
```

Lennonka:
@adamruzicka We could use this check in 3.16 as well because we need to keep cherry picking changes for the docs migration.

What else needs to be done? And how can I help to move this forward?

adamruzicka:
We have a bit of a circular dependency here. RedHatSatellite/foreman_theme_satellite#325 needs the "one more level" from here, this needs the reusable workflow from 325.

I'd suggest:
1. Extracting the one more level commit to its own PR - see https://github.com/theforeman/foreman-documentation/pull/4898
2. Once 1 gets in, waiting for the updated toc to be propagated to f-theme
3. Doing whatever changes are necessary for f-theme to be green
4. Merging this once it turns green

All that assumes we agree that this is the way to go. We could take a shortcut or two if we were fine with red ci for a while.

Lennonka:
> All that assumes we agree that this is the way to go. We could take a shortcut or two if we were fine with red ci for a while.

I trust your judgement. Let's proceed and do it properly.

Lennonka:
Hi @adamruzicka 

> 1. Extracting the one more level commit to its own PR - see [Include one more level in toc #4898](https://github.com/theforeman/foreman-documentation/pull/4898)

This has been merged. Can we continue with the next steps, please?

adamruzicka:
The next steps would be to make the link checking in f-theme-satellite green, see https://github.com/RedHatSatellite/foreman_theme_satellite/pull/309#pullrequestreview-4498555356 . No idea what to do with those things that just disappeared from the toc, even though we now include an additional level.

Lennonka:
Sorry, I'm slammed. I'll try to do it next week.

Lennonka:
Working on fixing links in https://github.com/RedHatSatellite/foreman_theme_satellite/pull/328

Lennonka:
> Working on fixing links in [RedHatSatellite/foreman_theme_satellite#328](https://github.com/RedHatSatellite/foreman_theme_satellite/pull/328)

Links have been fixed. Can we proceed?

Lennonka:
@adamruzicka Hi, Adam. Where are we at? :) Can I help with anything?

PR #4828, opened 2026-05-07
Refactor predefined tuning profiles into a modular assembly

aneta-petrova:

Replace the monolithic procedure with an assembly that includes two concept modules, a reference module for profile sizing, and a procedure for running the installer with --tuning.

This will make it easier to update the assembly for foremanctl.

Assisted-by: Cursor

#### Contributor checklists


guides/common/modules/ref_predefined-tuning-profiles.adoc
```diff
@@ -0,0 +1,45 @@
+:_mod-docs-content-type: REFERENCE
+
+[id="predefined-tuning-profiles_{context}"]
```
maximiliankolb:
```suggestion
[id="predefined-tuning-profiles"]
```

guides/common/modules/proc_applying-a-predefined-tuning-profile.adoc
```diff
@@ -0,0 +1,34 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="applying-a-predefined-tuning-profile_{context}"]
```
maximiliankolb:
```suggestion
[id="applying-a-predefined-tuning-profile"]
```

guides/common/modules/proc_applying-a-predefined-tuning-profile.adoc
```diff
@@ -0,0 +1,34 @@
...
+= Applying a predefined tuning profile
+
+[role="_abstract"]
+Run `{foreman-installer}` with the `--tuning` option to apply a profile.
+Choosing the right tuning profile helps you set up {ProjectServer} with the optimal number of CPU cores and memory allocation for your deployment.
+
+[NOTE]
+====
+You cannot use predefined tuning profiles on {SmartProxies}.
```
maximiliankolb:
```suggestion
You cannot use predefined tuning profiles on {SmartProxyServers}.
```

guides/common/modules/con_how-predefined-tuning-profiles-apply.adoc
```diff
@@ -0,0 +1,18 @@
+:_mod-docs-content-type: CONCEPT
+
+[id="how-predefined-tuning-profiles-apply_{context}"]
```
maximiliankolb:
```suggestion
[id="how-predefined-tuning-profiles-apply"]
```

guides/common/modules/proc_applying-a-predefined-tuning-profile.adoc
```diff
@@ -0,0 +1,34 @@
...
+.Procedure
+. Optional: If you use a `custom-hiera.yaml` file on your {ProjectServer}, back it up:
++
+----
+# cp /etc/foreman-installer/custom-hiera.yaml \
+/etc/foreman-installer/custom-hiera.original
+----
++
+This ensures you can restore the `custom-hiera.yaml` file to its original state if it becomes corrupted.
+. If you use a `custom-hiera.yaml` file on your {ProjectServer}, ensure the settings in `custom-hiera.yaml` do not override the settings in the default tuning profile and your chosen predefined tuning profile:
```
jafiala: The first step being optional and the second one not being optional makes it confusing.

Strictly speaking, conditional steps are not optional, but conditional, so the "optional" should not apply here. (see Indicating optional and conditional steps in IBM Style).

I'm tempted to create a single conditional, but that would need some restructuring of the substeps. Mockup:

```
.  If you use a `custom-hiera.yaml` file on your {ProjectServer}, perform the following steps:
.. Back up your `custom-hiera.yaml` file so you can restore it if the file becomes corrupted.
.. Remove any settings from your `custom-hiera.yaml` that duplicate the settings of the default tuning profile in `/usr/share/foreman-installer/config/foreman.hiera/tuning/common.yaml`.
```

I think this contains everything that needs to be said, no? We don't actually need sub-substeps.

jafiala (approved):
I really like the structure now. Just one note on simplifying the step structure.

guides/common/modules/proc_applying-a-predefined-tuning-profile.adoc
aneta-petrova: The thing is that backing up custom-hiera.yaml is both conditional and optional 🫠

aneta-petrova: And then the second step (to ensure there are no duplicated settings) is conditional and required.

guides/common/modules/ref_predefined-tuning-profiles.adoc
```diff
@@ -0,0 +1,45 @@
...
+[id="predefined-tuning-profiles"]
+= Predefined tuning profiles
+
+[role="_abstract"]
+The following predefined tuning profiles are available in {Project} to right-size your {ProjectServer} based on the number of hosts your {Project} manages and available hardware resources.
+
+The predefined tuning profiles are available in the `/usr/share/foreman-installer/config/foreman.hiera/tuning/sizes` directory.
+Each profile targets a managed-host range and minimum RAM and CPU cores.
+
+default::
```
aneta-petrova: @jafiala A list, or a table? Normally I'd say a table, but we are discouraged from using them now because AI can't consume them as well as lists, right? (I'd still prefer a table... but not sure.)

aneta-petrova: This has been a list for a long time so keeping it this way just maintains a status quo, which is also why I haven't converted it to a table yet. I could, though.

guides/common/modules/proc_applying-a-predefined-tuning-profile.adoc
jafiala: Ah, well that makes sense, but is not very obvious. But I think it works better with the proposed structure:

```
.  If you use a `custom-hiera.yaml` file on your {ProjectServer}, perform the following steps:
.. Optional: Back up your `custom-hiera.yaml` file so you can restore it if the file becomes corrupted.
.. Remove any settings from your `custom-hiera.yaml` that duplicate the settings of the default tuning profile in `/usr/share/foreman-installer/config/foreman.hiera/tuning/common.yaml`.
```

guides/common/modules/ref_predefined-tuning-profiles.adoc
ekohl: IMHO tabular data should be presented in a table. I never liked the long list because as a human it's really hard to understand.

guides/common/modules/con_how-predefined-tuning-profiles-apply.adoc
ekohl, line 1: Reviewer note: this explains how https://github.com/theforeman/foreman-installer/blob/develop/config/foreman-hiera.yaml is read. The Hiera priority file is read from most important to least important. This file currently lists them from least important to most important.

guides/common/modules/con_how-predefined-tuning-profiles-apply.adoc
```diff
@@ -0,0 +1,18 @@
...
+[id="how-predefined-tuning-profiles-apply"]
+= How predefined tuning profiles apply
+
+[role="_abstract"]
+When you run the `{foreman-installer}` command with the `--tuning` option, deployment configuration settings from multiple sources are applied to {Project} in a specific order.
+This order defines the priority of these settings.
+
+The `{foreman-installer}` command merges configuration settings from the following sources in this order:
+
+. The default tuning profile defined in the `/usr/share/foreman-installer/config/foreman.hiera/tuning/common.yaml` file.
```
ekohl: As a user when I select a tuning profile I expect the default to no longer apply. When we implemented tuning we named it `common.yaml` to make it clear it's that it always applies. Perhaps the documentation could follow this idea?

guides/common/modules/con_how-predefined-tuning-profiles-apply.adoc
```diff
@@ -0,0 +1,18 @@
...
+. The predefined tuning profile, as defined in the `/usr/share/foreman-installer/config/foreman.hiera/tuning/sizes/` directory, that you select when running the `{foreman-installer}` command with the `--tuning` option.
```
ekohl: It feels to me that the selected part should be earlier in the sentence. But when I read it then it doesn't look so great either. 
```suggestion
. The selected predefined tuning profile, as defined in the `/usr/share/foreman-installer/config/foreman.hiera/tuning/sizes/` directory, that you select when running the `{foreman-installer}` command with the `--tuning` option.
```

guides/common/modules/proc_applying-a-predefined-tuning-profile.adoc
```diff
@@ -0,0 +1,34 @@
...
+====
+You cannot use predefined tuning profiles on {SmartProxyServers}.
+====
+
+.Procedure
+. Optional: If you use a `custom-hiera.yaml` file on your {ProjectServer}, back it up:
++
+----
+# cp /etc/foreman-installer/custom-hiera.yaml \
+/etc/foreman-installer/custom-hiera.original
```
ekohl: I'd suggest to keep the same file extension so your editor can still recognize it and apply the correct syntax highlighting
```suggestion
/etc/foreman-installer/custom-hiera.original.yaml
```

guides/common/modules/proc_applying-a-predefined-tuning-profile.adoc
```diff
@@ -0,0 +1,34 @@
...
+. If you use a `custom-hiera.yaml` file on your {ProjectServer}, ensure the settings in `custom-hiera.yaml` do not override the settings in the default tuning profile and your chosen predefined tuning profile:
+.. Review the definitions of the default tuning profile in `/usr/share/foreman-installer/config/foreman.hiera/tuning/common.yaml` and the tuning profile that you want to apply in `/usr/share/foreman-installer/config/foreman.hiera/tuning/sizes/`.
+.. Compare the configuration entries against the entries in your `/etc/foreman-installer/custom-hiera.yaml` file.
+.. Remove any duplicated configuration settings that you find in `custom-hiera.yaml`.
+. Enter the `{foreman-installer}` command with the `--tuning` option for the profile that you want to apply.
+For example, you can apply the medium tuning profile settings:
++
+[options="nowrap" subs="attributes"]
+----
+# {foreman-installer} --tuning medium
```
ekohl: Should we make it clear that `medium` is a value provided by the user?
```suggestion
# {foreman-installer} --tuning _medium_
```

guides/common/modules/proc_applying-a-predefined-tuning-profile.adoc
aneta-petrova: Love it, thanks, applied.

guides/common/modules/proc_applying-a-predefined-tuning-profile.adoc
ekohl, line 1: Reviewer notes: we want users to run `foreman-installer --tuning $profile`. Then since `custom-hiera.yaml` can undo some tuning options we provide. Because of that we're steering them away from having values in `custom-hiera.yaml`. This is a manual process we can't really automate so it's a manual procedure.

If the user doesn't use `custom-hiera.yaml` (ideal case) then the whole procedure really is just applying the tuning profile: `foreman-installer --tuning $profile`.

ekohl (commented):
Overall I think this reads well. Inline there are some minor comments. I also left some reviewer notes on what I think the file does / should do. Those are my own notes and feel free to ignore them.

guides/common/modules/con_how-predefined-tuning-profiles-apply.adoc
aneta-petrova: I reorganized the order in which things are explained here.

guides/common/modules/con_how-predefined-tuning-profiles-apply.adoc
aneta-petrova: So this isn't the `default` profile listed in guides/common/modules/ref_predefined-tuning-profiles.adoc?

guides/common/modules/con_how-predefined-tuning-profiles-apply.adoc
aneta-petrova: I changed the sentence structure in a different way.

guides/common/modules/proc_applying-a-predefined-tuning-profile.adoc
aneta-petrova: We could add a note phrased like a warning to guides/common/modules/con_how-predefined-tuning-profiles-apply.adoc, explaining that this is not the preferred way to manage the server because <reason>. Would that help? And if so, what exactly is the reason?

guides/common/modules/proc_applying-a-predefined-tuning-profile.adoc
aneta-petrova: I can mark it as a replaceable, but will also need to change the lead-in sentence a bit.

guides/common/modules/ref_predefined-tuning-profiles.adoc
jafiala: I agree, this seems like tabular data. There are also concerns about LLMs consuming lists, of all things, so let's just go with a table.

guides/common/modules/proc_applying-a-predefined-tuning-profile.adoc
ekohl: Zooming out, what we really wanted was something that automatically tunes the system based on the instance size. That was hard so we settled on the middle ground: we provide a table of sizes that we guess approximates common uses. Then the admin chooses that and a matching tuning. That is the happy path.

A further complexity: before we had tuning profiles there were users who applied tuning using `custom-hiera.yaml`. That can override the tuning profiles. Technically `custom-hiera.yaml` is an unsupported solution, but common so to mitigate that we provided instructions on how to prune some options from `custom-hiera.yaml`.

Zooming back in, that means what I'd expect is:

* A table of sizes so the admin knows which tuning profile to use (`guides/common/modules/ref_predefined-tuning-profiles.adoc`)
* Instructions to apply that tuning profile (`foreman-installer --tuning $profile`)
* If you have applied `custom-hiera.yaml` then you should prune entries.

For the pruning part you need to understand how we apply tuning (`guides/common/modules/con_how-predefined-tuning-profiles-apply.adoc`) and a procedure to drop entries (most of `guides/common/modules/proc_applying-a-predefined-tuning-profile.adoc`).

Is the pruning part something you'd put in an appendix?

guides/common/modules/proc_applying-a-predefined-tuning-profile.adoc
```diff
@@ -0,0 +1,34 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="applying-a-predefined-tuning-profile"]
+= Applying a predefined tuning profile
+
+[role="_abstract"]
+Run `{foreman-installer}` with the `--tuning` option to apply a profile.
+Choosing the right tuning profile helps you set up {ProjectServer} with the optimal number of CPU cores and memory allocation for your deployment.
```
ekohl: Following up from https://github.com/theforeman/foreman-documentation/pull/4828#discussion_r3258624593 (and a few days of thinking) we really wanted it the other way around: the system adjusts to your system size (that you determined during planning). We couldn't come up with a good implementation so came up with some numbers that we think may work well for certain sizes.

This is a mediocre sentence, but I wonder if this is a better description.
```suggestion
Choosing the right tuning profile helps your {ProjectServer} utilize its CPU cores and memory better.
```

guides/common/modules/proc_applying-a-predefined-tuning-profile.adoc
aneta-petrova: This is the trouble with features that are "unsupported but allowed": If we want to provide the necessary instructions, we should provide them in a place where users will be looking for them. In this case, it's documentation on tuning profiles.

Splitting the custom-hiera.yaml part into an appendix is doable, but would require cross-linking and jumping between sections; users who don't use custom Hiera config would still have their reading disrupted by the cross-link, and users who do use Hiera config would not get the hint that it's unsupported just because it's in an appendix.

In order to keep related information together and to avoid disrupting user workflows, I'd still prefer to keep everything together. If it's desirable to discourage users from using custom-hiera.yaml, then we should IMO either say so explicitly (for example through an admonition) or declare it unsupported and remove it from the docs entirely.

ekohl: We've largely avoided this topic because with foremanctl the whole `custom-hiera.yaml` concept isn't really a thing anymore. With that in mind I've not pushed for a complete restructure of this but thought it was useful to at least share the background how we got where we are today. And it worked, because I could better review that 1 sentence.

So with that in mind, I think this is probably good enough but I'll try to give it a full read through before the end of the day.

guides/common/modules/proc_applying-a-predefined-tuning-profile.adoc
```diff
@@ -0,0 +1,34 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="applying-a-predefined-tuning-profile"]
+= Applying a predefined tuning profile
+
+[role="_abstract"]
+Run `{foreman-installer}` with the `--tuning` option to apply a profile.
+Choosing the right tuning profile helps your {ProjectServer} utilize its CPU cores and memory better.
```
aneta-petrova:
```suggestion
Choosing the right tuning profile helps your {ProjectServer} use its CPU cores and memory better.
```

guides/common/modules/con_how-predefined-tuning-profiles-apply.adoc
```diff
@@ -0,0 +1,25 @@
...
+
+Custom Hiera settings::
+If you use custom Hiera settings on your {ProjectServer}, you can find them in the `/etc/foreman-installer/custom-hiera.yaml` file.
+
+Default tuning profile::
+The settings for this profile are defined in the `/usr/share/foreman-installer/config/foreman.hiera/tuning/common.yaml` file.
+
+Predefined tuning profile::
+This is the profile that you select when running the `{foreman-installer}` command with the `--tuning` option.
+The settings for these profiles are defined in the `/usr/share/foreman-installer/config/foreman.hiera/tuning/sizes/` directory.
```
ekohl, line 17: Isn't it easier to follow if the sources are listed in order of priority rather than alphabetically? That way the reader doesn't have to mentally reorganize the items based on what's written below.

guides/common/modules/con_how-predefined-tuning-profiles-apply.adoc
```diff
@@ -0,0 +1,25 @@
...
+[role="_abstract"]
+When you run the `{foreman-installer}` command with the `--tuning` option, deployment configuration settings from multiple sources are applied to {Project} in a specific order.
+This order defines the priority of these settings.
+
+{Project} loads tuning configuration settings from the following sources:
+
+Custom Hiera settings::
+If you use custom Hiera settings on your {ProjectServer}, you can find them in the `/etc/foreman-installer/custom-hiera.yaml` file.
+
+Default tuning profile::
```
ekohl: I'd like to object to calling this the `default` tuning profile. When I read default then I'll assume that choosing another profile means nothing from the default profile is read. Here that's not the case. I'd suggest:
```suggestion
Base tuning profile::
```

Or

```suggestion
Common tuning profile::
```

ekohl (changes requested):
IMHO the default tuning profile vs base profile (or common) should be changed. The other comment is a suggestion.

I'll leave it to tech writers to judge if the application procedure should be split into 2: for those using `custom-hiera.yaml` and those that aren't.

guides/common/modules/con_how-predefined-tuning-profiles-apply.adoc
aneta-petrova: Can you please look at my question in https://github.com/theforeman/foreman-documentation/pull/4828#discussion_r3235812683? Before I change the terminology, I'd like to know exactly what the term is describing.

ekohl: Sorry, I missed that comment.

guides/common/modules/con_how-predefined-tuning-profiles-apply.adoc
ekohl: No, technically there is no default profile.

Under the hood we use layering. Hiera is effectively a lot of keys and values in YAML files and the first key that is found wins.

A good example is `apache::mod::event::startservers` that only exists within `tuning/common.yaml` and not in `tuning/sizes/*.yaml`.

`postgresql::server::config_entries` it's a bit special, but we merge those. Still, same rules apply: first entry found wins. Here `work_mem` is good example. In `tuning/common.yaml` we see `work_mem: 4MB` while `tuning/sizes/extra-large.yaml` and `tuning/sizes/extra-extra-large.yaml` have `work_mem: 8MB`. Since it's not in `tuning/sizes/medium.yaml` and `tuning/sizes/large.yaml`, it falls back to the value in `tuning/common.yaml` (`4MB`).

guides/common/modules/con_how-predefined-tuning-profiles-apply.adoc
aneta-petrova: Reordered.

aneta-petrova:
@ekohl I changed the terminology to use 'base' profile instead of 'default' profile. Is there anything else?

ekohl (approved):
Tech ack.

aneta-petrova:

3a6800274c..b24d855bf4  3.19 -> 3.19
999c3ece50..97eeb9d8ca  3.18 -> 3.18 (a minor conflict)
9df886d98a..ac2d976a27  3.17 -> 3.17
c63c436b39..1ae949b7aa  3.16 -> 3.16

PR #4879, opened 2026-05-26
Add IGNORE content type in pre-processing

aneta-petrova:

* Removing the ASSEMBLY content type
* Extending the Vale workflow with a pre-processing step that injects the IGNORE content type into assembly files

Each assembly and module must include a content type definition to be evaluated by asciidoctor-dita-vale (https://github.com/jhradilek/asciidoctor-dita-vale/blob/af6c00940de1ce99994ddd1631aa18d35a7ff0a3/README.md?plain=1#L130). With this PR, I'm trying to avoid the need to include the IGNORE content type definition in assembly files by injecting the module type before the Vale check runs.

This saves us the need of adding two lines into 165 files that literally just say "ignore this file".

Made with Claude but reviewed by myself to the best of my abilities :upside_down_face: 

#### Contributor checklists

### DIFF
--- a/.github/scripts/preprocess-vale-dita.sh
+++ b/.github/scripts/preprocess-vale-dita.sh
@@ -24,7 +24,7 @@ for file in "$@"; do
   if [[ $(basename "$file") == assembly_*.adoc ]]; then
     # Inject content type definition at the beginning
     {
-      echo ":_mod-docs-content-type: ASSEMBLY"
+      echo ":_mod-docs-content-type: IGNORE"
       echo ""
       cat "$file"
     } > "$TEMP_DIR/$file"
--- a/.github/scripts/preprocess-vale-dita.sh
+++ b/.github/scripts/preprocess-vale-dita.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+# Preprocess AsciiDoc files for Vale DITA linting by injecting content type metadata
+set -euo pipefail
+
+# Create temp directory for preprocessed files
+TEMP_DIR=".vale-temp"
+mkdir -p "$TEMP_DIR"
+
+# Track if we processed any files
+file_count=0
+
+# Process files passed as arguments
+for file in "$@"; do
+  # Skip if file doesn't exist
+  if [[ ! -f "$file" ]]; then
+    echo "Warning: File not found: $file" >&2
+    continue
+  fi
+
+  # Create directory structure in temp
+  mkdir -p "$TEMP_DIR/$(dirname "$file")"
+
+  # Check if it's an assembly file
+  if [[ $(basename "$file") == assembly_*.adoc ]]; then
+    # Inject content type definition at the beginning
+    {
+      echo ":_mod-docs-content-type: ASSEMBLY"
+      echo ""
+      cat "$file"
+    } > "$TEMP_DIR/$file"
+    echo "Preprocessed assembly: $file" >&2
+  else
+    # Copy non-assembly files as-is
+    cp "$file" "$TEMP_DIR/$file"
+  fi
+
+  file_count=$((file_count + 1))
+done
+
+echo "Preprocessed $file_count file(s)" >&2
+
+# Output preprocessed file paths to stdout (one per line)
+for file in "$@"; do
+  if [[ -f "$file" ]]; then
+    echo "$TEMP_DIR/$file"
+  fi
+done
--- a/.github/workflows/vale.yml
+++ b/.github/workflows/vale.yml
@@ -86,9 +86,6 @@ jobs:
         if: steps.changed-files.outputs.any_changed == 'true'
         id: preprocess
         run: |
-          # Make script executable
-          chmod +x .github/scripts/preprocess-vale-dita.sh
-
           # Convert JSON array to space-separated list
           FILES=$(echo '${{ steps.changed-files.outputs.all_changed_files }}' | jq -r '.[]' | tr '\n' ' ')
 
--- a/.github/workflows/vale.yml
+++ b/.github/workflows/vale.yml
@@ -82,11 +82,30 @@ jobs:
         if: steps.changed-files.outputs.any_changed == 'true'
         run: sudo apt-get install -y asciidoctor
 
+      - name: Preprocess assembly files for DITA linting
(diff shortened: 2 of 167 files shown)


jafiala:
While I like this approach, won't it just silence the vale check but still present problems during the actual migration? Assemblies should be converted to DITA maps in the migration, and the migration workflow might be looking for :_mod-docs-content-type: ASSEMBLY for that conversion.

But I suppose that will also apply to the manual :_mod-docs-content-type: IGNORE approach.

So among the two approaches, I think this one is preferable.

No matter which path we choose, we should also add a note to Contributing.

jafiala (approved):

aneta-petrova:
> While I like this approach, won't it just silence the vale check but still present problems during the actual migration? Assemblies should be converted to DITA maps in the migration, and the migration workflow might be looking for :_mod-docs-content-type: ASSEMBLY for that conversion.

I (vaguely) recall a conversation with the author of https://github.com/jhradilek/asciidoctor-dita-vale/ where he said that adding content types during pre-processing was a valid solution.

For the IGNORE type, I think it is safe. (It's not that any other Vale checks depend on IGNORE being present, it just disables all the checks.)

For the other content types, it might be more tricky, which is why I'm not suggesting to add those to pre-processing. However, a solution certainly exists -- we could add those content types during the downstream sync, or the migration team could implement similar pre-processing on their side.

> 
> But I suppose that will also apply to the manual :_mod-docs-content-type: IGNORE approach.
> 
> So among the two approaches, I think this one is preferable.
> 
> No matter which path we choose, we should also add a note to Contributing.

aneta-petrova:
> No matter which path we choose, we should also add a note to Contributing.

To be honest, I don't think that's necessary because CONTRIBUTING.md currently doesn't mention a content type for assemblies at all:

https://github.com/theforeman/foreman-documentation/blob/d95c628624be50ae1738748e62054657ecf64fe6/CONTRIBUTING.md?plain=1#L156-L161

Unlike the other content types, which are mentioned below these lines.

For contributors, this IMHO doesn't represent anything they would need to actively be aware of -- the assembly files get ignored silently under the hood, which the best way to ignore something ;)

.github/workflows/vale.yml
```diff
@@ -82,11 +82,30 @@ jobs:
         if: steps.changed-files.outputs.any_changed == 'true'
         run: sudo apt-get install -y asciidoctor
 
+      - name: Preprocess assembly files for DITA linting
+        if: steps.changed-files.outputs.any_changed == 'true'
+        id: preprocess
+        run: |
+          # Make script executable
+          chmod +x .github/scripts/preprocess-vale-dita.sh
```
maximiliankolb: You could run this once locally and commit the changes, then drop lines 89,90,91.

.github/scripts/preprocess-vale-dita.sh
maximiliankolb, line 1: I think that you can basically shorten this script to

````
find . -type f -name "assembly_*.adoc" | xargs sed --in-place "1iBLUB\n"
````

Replace "blub" with the attribute. Note that this does not work locally as expected because `git` reports a different diff because the assemblies already have the attribute set. So it works, but `git diff` shows a weird diff 🙃

maximiliankolb (commented):
I left two comments & will reproduce the rest of the diff later locally.

maximiliankolb:
> reproduce the rest of the diff later locally.

`fd -t f "assembly_" | xargs sed --in-place "/:_mod-docs-content-type: ASSEMBLY/,+1d"` so anyone can quickly reproduce the changes on this PR once it's ready for final ACK.

.github/scripts/preprocess-vale-dita.sh
aneta-petrova: The way the script is written is designed to avoid the need to modify the original files. That's why it creates copies in a temp directory.

Then there are also some extra bits for potential debugging purposes (warning messages, tracking file count), which could also be useful, especially if/when we add more content types to the mix later.

As I stated before, this PR was drafted by AI and AI is known to write more lines than are sometimes necessary :upside_down_face: When I reviewed it, it seemed to make sense to make it more robust -- especially that part about copying the files to a temp directory rather than modifying them directly :shrug:

aneta-petrova:
> once it's ready for final ACK.

I dropped the (now unnecessary) step to make the preprocessing script executable, and explained my reasoning for keeping the script a bit more robust than a `sed` command.

Then I also fixed the script to inject IGNORE rather than ASSEMBLY, which I totally missed before :see_no_evil: 

To test the PR, I performed the following steps:

1. I ran the script on an assembly: `/scripts/preprocess-vale-dita.sh guides/common/assembly_managing-organizations.adoc`
2. I checked that the temporary assembly created by the script includes the IGNORE content type on its first line: `cat .vale-temp/guides/common/assembly_managing-organizations.adoc`
3. I ran the DITA Vale check on the temporary assembly: `vale --config=.vale-dita.ini .vale-temp/guides/common/assembly_managing-organizations.adoc`
4. Additionally, I checked that the GH workflow `lint DITA` includes running the pre-processing script: https://github.com/theforeman/foreman-documentation/actions/runs/26814218371/job/79051743435?pr=4879

Is this enough to check that it's working as intended? Am I missing anything?

maximiliankolb:
> Is this enough to check that it's working as intended?

Yes.

aneta-petrova:
Now that I'm looking at it as a whole, should we just modify the Vale workflow to process only guides/common/modules rather than guides/common? If the point is to tell Vale to IGNORE the assembly files, should we just exclude them? Isn't that the best kind of ignoring?

maximiliankolb:
> should we just modify the Vale workflow to process only guides/common/modules rather than guides/common?

Yes, good call.

aneta-petrova:
Let's explore that approach in https://github.com/theforeman/foreman-documentation/pull/4893

aneta-petrova:
https://github.com/theforeman/foreman-documentation/pull/4893 seems to be a much simpler and the preferable way to resolve the ASSEMBLY errors. However, I'm keeping this PR open in draft state because it's worth revisiting in the future when/if we decide to move the other content types (the ones in module files, which remain mandatory) to pre-processing.

PR #4865, opened 2026-05-21
Replace ASSEMBLY content type with IGNORE

aneta-petrova:
#### What changes are you introducing?

Replacing the assembly content type with the ignore content type throughout the repository.

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

Assemblies in foreman-documentation do not match the expectations defined by https://github.com/jhradilek/asciidoctor-dita-vale. Replacing the ASSEMBLY content type with the IGNORE content type prevents related Vale warnings.

https://github.com/jhradilek/asciidoctor-dita-vale/pull/179

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)

N/A

#### Contributor checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.19/Katello 4.21
* [x] Foreman 3.18/Katello 4.20 (Satellite 6.19)
* [x] Foreman 3.17/Katello 4.19
* [x] Foreman 3.16/Katello 4.18 (Satellite 6.18; orcharhino 7.6, 7.7, and 7.8)
* [ ] Foreman 3.15/Katello 4.17
* [ ] Foreman 3.14/Katello 4.16 (Satellite 6.17; orcharhino 7.4; orcharhino 7.5)
* [ ] Foreman 3.13/Katello 4.15 (EL9 only)
* [ ] Foreman 3.12/Katello 4.14 (Satellite 6.16; orcharhino 7.2 on EL9 only; orcharhino 7.3)
* We do not accept PRs for Foreman older than 3.12.

PR #4887, opened 2026-05-28
Review assembly Maintaining Server for CQA

aneta-petrova:
#### What changes are you introducing?

* Introducing sub-assemblies, each focused on a single user story
* Reframing the original assembly from generic and vague "maintaining" to "I'm running low on space -- what can I do?"
* Moving a module on anonymizing audit records out of the assembly and into an appendix

This involves some new abstracts and ID changes but no changes to the contents of the modules.

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

Part of the downstream Content Quality Assessment (CQA) effort.

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)

N/A

#### Contributor checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.19/Katello 4.21
* [x] Foreman 3.18/Katello 4.20 (Satellite 6.19)
* [x] Foreman 3.17/Katello 4.19
* [x] Foreman 3.16/Katello 4.18 (Satellite 6.18; orcharhino 7.6, 7.7, and 7.8)
* [ ] Foreman 3.15/Katello 4.17
* [ ] Foreman 3.14/Katello 4.16 (Satellite 6.17; orcharhino 7.4; orcharhino 7.5)
* [ ] Foreman 3.13/Katello 4.15 (EL9 only)
* [ ] Foreman 3.12/Katello 4.14 (Satellite 6.16; orcharhino 7.2 on EL9 only; orcharhino 7.3)
* We do not accept PRs for Foreman older than 3.12.

PR #4893, opened 2026-06-02
Exclude assemblies from lint-DITA

aneta-petrova:

Modifying the DITA part of the Vale check to run only on modules and not on assemblies.

While each assembly and module must include a content type definition to be evaluated by asciidoctor-dita-vale (https://github.com/jhradilek/asciidoctor-dita-vale/blob/af6c00940de1ce99994ddd1631aa18d35a7ff0a3/README.md?plain=1#L130), assemblies in Foreman docs can be ignored because they serve as simple containers of includes. For the purposes of Red Hat's migration to DITA downstream, checking the modules in `guides/common/modules` is enough.

Third time's the charm, after https://github.com/theforeman/foreman-documentation/pull/4865 and https://github.com/theforeman/foreman-documentation/pull/4879 These two PRs explored how to use the IGNORE content type to exclude assemblies from the Vale checks. However, if Vale is not supposed to check assemblies, we can also just drop it from the GH workflow entirely.

#### Contributor checklists

### DIFF
--- a/.cursor/skills/review-assembly-user-story/SKILL.md
+++ b/.cursor/skills/review-assembly-user-story/SKILL.md
@@ -56,11 +56,10 @@ The assembly is fine if:
 For each identified user story:
 
 1. **Create a new assembly file** - Name it `assembly_<descriptive-name>.adoc`
-2. **Add the assembly header** - Include `:_mod-docs-content-type: ASSEMBLY`
-3. **Include relevant modules** - Copy the appropriate module includes for this user story
-4. **Create or update the introductory concept module** - Write a focused concept module with an abstract for this specific user story if needed
-5. **Update parent files** - Replace the original assembly include with includes for all new assemblies in the `master.adoc` file
-6. **Delete the original assembly** - Once all content is split and includes are updated
+2. **Include relevant modules** - Copy the appropriate module includes for this user story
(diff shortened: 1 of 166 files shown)


Lennonka (approved):
The PR looks correct per the intended goal.

Will removing the content type attribute definitions ensure that our "include containers" assemblies do not appear in Jaromir's reports downstream?

Lennonka:
Perhaps we should also update skills that create assemblies so that they don't add the content type attribute?

maximiliankolb (approved):
Thanks Anet, LGTM.

I checked `.github/workflows/vale.yml` and reproduced the remaining changes locally via `fd -t f "assembly_" | xargs sed --in-place "/:_mod-docs-content-type: ASSEMBLY/,+1d"`.

Lennonka (approved):

aneta-petrova:
Rebased to pull in the changes merged with https://github.com/theforeman/foreman-documentation/pull/4887 and resolve the conflict (the other PR deletes an assembly).

aneta-petrova:
Considering that the ASSEMBLY content type is not needed downstream, and it's not relevant for upstream either, I've decided not to cherry-pick. The only consequence should be that PRs on branches 3.16-3.19 will keep reporting Vale errors for assemblies but I don't think it's a big deal because we've been quite okay with ignoring the same errors up until now :)

PR #4903, opened 2026-06-03
Add workflow to report ID and file name mismatch

aneta-petrova:

Adding a pull request check that reports when a module's file name, ID, or heading don't match.

It happens to me all the time that I forget to keep them in sync in my PRs, and seeing a warning saves precious peer reviewers' time that they otherwise have to spend pointing it out to me :upside_down_face: 

My first choice would have been Vale but Vale can't read file names. Hence a script.

#### Contributor checklists

### DIFF
--- a/.github/workflows/check-filename-id-heading-match.yml
+++ b/.github/workflows/check-filename-id-heading-match.yml
@@ -1,5 +1,5 @@
 ---
-name: Check Filename-ID-Heading Match
+name: check
 on:
   pull_request:
     paths:
@@ -9,7 +9,7 @@ on:
 
 jobs:
   check-match:
-    name: Check filename-ID-heading match
+    name: Filename-ID-Heading match
     runs-on: ubuntu-24.04
     permissions:
       contents: read
--- a/.github/workflows/check-filename-id-heading-match.yml
+++ b/.github/workflows/check-filename-id-heading-match.yml
@@ -1,15 +1,15 @@
 ---
-name: Check ID-Filename Match
+name: Check Filename-ID-Heading Match
 on:
   pull_request:
     paths:
       - 'guides/common/modules/**.adoc'
-      - 'scripts/check-id-filename-match.sh'
-      - '.github/workflows/check-id-filename.yml'
+      - 'scripts/check-filename-id-heading-match.sh'
+      - '.github/workflows/check-filename-id-heading-match.yml'
 
 jobs:
-  check-ids:
-    name: Check module IDs match filenames
+  check-match:
+    name: Check filename-ID-heading match
     runs-on: ubuntu-24.04
     permissions:
       contents: read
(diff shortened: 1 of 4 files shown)


aneta-petrova:
I've tested this on many different filename/ID/heading changes. A few examples of the errors the check produced when I was testing it.

Heading includes "Content flows" but the ID and filename are "content-flow":
Screenshot From 2026-06-04 13-58-20

ID includes "-test", and the extra word produces a warning:
Screenshot From 2026-06-04 10-00-23

ID is capitalized but the expectation is a lower-case ID:
Screenshot From 2026-06-04 09-52-32

Nothing happens for snippets because they're not supposed to have IDs; the workflow skips them entirely:
Screenshot From 2026-06-04 09-31-50

ID contains `_{context}`, which gets reported as a mis-match. This is in line with our convention to remove `_{context}` from new IDs:
Screenshot From 2026-06-04 09-31-36

Some of these examples showcase that the check handles attributes in IDs as well, as long as they're defined in the list of attributes to be normalized in the script itself. (`{project-context}` in ID versus `project` in filename, for example).

aneta-petrova:
I tried to account for capitalization requirements too:

A capitalized word in the heading doesn't trigger a warning:
Screenshot From 2026-06-04 14-51-44

But a capitalized element in the ID does because IDs are supposed to be lower case only:
Screenshot From 2026-06-04 14-52-43

aneta-petrova:
Ready for review now. I've done a lot of testing by breaking various IDs/headings/file names and took screenshots of the messages (see comments above).

.github/workflows/check-filename-id-heading-match.yml
```diff
@@ -0,0 +1,88 @@
+---
+name: Check Filename-ID-Heading Match
```
Lennonka:
```suggestion
name: check
```
This of this as the "pipeline type".

.github/workflows/check-filename-id-heading-match.yml
```diff
@@ -0,0 +1,88 @@
...
+on:
+  pull_request:
+    paths:
+      - 'guides/common/modules/**.adoc'
+      - 'scripts/check-filename-id-heading-match.sh'
+      - '.github/workflows/check-filename-id-heading-match.yml'
+
+jobs:
+  check-match:
+    name: Check filename-ID-heading match
```
Lennonka:
```suggestion
    name: Filename-ID-Heading match
```
Think of this as the "check subject".

Lennonka (commented):
Oh, this is very elegant! I appreciate that it's attribute aware.

Can we just tweak the naming a bit so that we get an equally elegant pipeline name, ie. `check / Filename-ID-Heading match` ? Pretty please?

Lennonka (approved):
Thank you! LGTM

aneta-petrova:
> Oh, this is very elegant!

This is one of those situations 'why hasn't anyone thought of this before', right? :laughing: 

> I appreciate that it's attribute aware.

Agreed, that's great. Courtesy of Claude but very neat and seems to work reliably. Also, it should be easy for anyone to update the list in the script with more attributes if needed.

PR #4899, opened 2026-06-03
Make unready containerized docs not ready 

jafiala:
#### What changes are you introducing?
Hide contents of unready containerized guides, and include the "guide not ready" module.

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)
https://redhat.atlassian.net/browse/SAT-45577

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)
Discussed on matrix

#### Contributor checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.19/Katello 4.21
* [ ] Foreman 3.18/Katello 4.20 (Satellite 6.19)
* [ ] Foreman 3.17/Katello 4.19
* [ ] Foreman 3.16/Katello 4.18 (Satellite 6.18; orcharhino 7.6, 7.7, and 7.8)
* [ ] Foreman 3.15/Katello 4.17
* [ ] Foreman 3.14/Katello 4.16 (Satellite 6.17; orcharhino 7.4; orcharhino 7.5)
* [ ] Foreman 3.13/Katello 4.15 (EL9 only)
* [ ] Foreman 3.12/Katello 4.14 (Satellite 6.16; orcharhino 7.2 on EL9 only; orcharhino 7.3)
* We do not accept PRs for Foreman older than 3.12.

PR #4905, opened 2026-06-04
Document foremanctl impact on Pulp import/export paths

aneta-petrova:

* Updating existing information on allowed import paths for foremanctl
* Adding a snippet with information on export paths

https://github.com/theforeman/foremanctl/pull/455
https://github.com/theforeman/foremanctl/blob/744ba31fd0ee4e4a74eef97e2c705acca1b2961d/src/roles/pulp/defaults/main.yaml#L29-L30
https://github.com/theforeman/foremanctl/issues/445

Assisted-by: Cursor

#### Contributor checklists

### DIFF
--- a/guides/common/modules/proc_creating-a-local-source-for-a-custom-file-type-repository.adoc
+++ b/guides/common/modules/proc_creating-a-local-source-for-a-custom-file-type-repository.adoc
@@ -37,12 +37,22 @@ endif::[]
 ----
 # mkdir -p /var/lib/pulp/__local_repos__/__my_file_repo__
 ----
+ifndef::containerized[]
 . Add the parent folder into allowed import paths:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
 # {foreman-installer} --foreman-proxy-content-pulpcore-additional-import-paths /var/lib/pulp/__local_repos__
 ----
+endif::[]
+ifdef::containerized[]
+. Add the parent folder to allowed import paths:
++
+[options="nowrap", subs="+quotes,attributes"]
+----
+# {foremanctl} deploy --content-import-path /var/lib/pulp/__local_repos__
+----
+endif::[]
 . Add files to the directory or create a test file:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
--- a/guides/common/modules/snip_allowed-export-paths-prerequisite.adoc
+++ b/guides/common/modules/snip_allowed-export-paths-prerequisite.adoc
@@ -1,11 +1,5 @@
 :_mod-docs-content-type: SNIPPET
 
 * Exports are written to one of your allowed export paths.
-ifndef::containerized[]
-The default export path is `/var/lib/pulp/exports/`.
-To configure additional paths, use `ALLOWED_EXPORT_PATHS` in `/etc/pulp/settings.py`.
-endif::[]
-ifdef::containerized[]
 The default export path is `/var/lib/pulp/exports/`.
 To configure additional paths, run `{foremanctl} deploy` with one or more `--content-export-path` options.
-endif::[]
--- a/guides/common/modules/snip_allowed-export-paths-prerequisite.adoc
+++ b/guides/common/modules/snip_allowed-export-paths-prerequisite.adoc
@@ -1,0 +1,11 @@
+:_mod-docs-content-type: SNIPPET
+
+* Exports are written to one of your allowed export paths.
+ifndef::containerized[]
+The default export path is `/var/lib/pulp/exports/`.
+To configure additional paths, use `ALLOWED_EXPORT_PATHS` in `/etc/pulp/settings.py`.
+endif::[]
+ifdef::containerized[]
+The default export path is `/var/lib/pulp/exports/`.
+To configure additional paths, run `{foremanctl} deploy` with one or more `--content-export-path` options.
+endif::[]
--- a/guides/scripts/find_unused_modules
+++ b/guides/scripts/find_unused_modules
@@ -43,6 +43,10 @@ unused = ((modules + assemblies) - absolute_included).map do |path|
   Pathname.new(path).relative_path_from(doc_root).to_s
 end
 
+# containerized content build isn't published
+# https://github.com/theforeman/foreman-documentation/issues/4964
+unused.delete('guides/common/modules/snip_allowed-export-paths-prerequisite.adoc')
+
 if unused.any?
   puts unused
   if ENV.key?('GITHUB_ACTIONS')
--- a/guides/common/modules/proc_exporting-a-content-view-version-in-the-syncable-format.adoc
+++ b/guides/common/modules/proc_exporting-a-content-view-version-in-the-syncable-format.adoc
@@ -16,7 +16,9 @@ include::snip_generating-content.adoc[]
 include::snip_content-types-syncable-export.adoc[]
 
 .Prerequisites
+ifdef::containerized[]
 include::snip_allowed-export-paths-prerequisite.adoc[]
+endif::[]
 * Ensure that you set the download policy to *Immediate* for all repositories within the content view you export.
 For more information, see xref:Download_Policies_Overview_{context}[].
 * Ensure that you synchronize products you export to the required date.
--- a/guides/common/modules/proc_exporting-a-content-view-version-in-the-syncable-format.adoc
+++ b/guides/common/modules/proc_exporting-a-content-view-version-in-the-syncable-format.adoc
@@ -16,6 +16,7 @@ include::snip_generating-content.adoc[]
 include::snip_content-types-syncable-export.adoc[]
 
 .Prerequisites
+include::snip_allowed-export-paths-prerequisite.adoc[]
 * Ensure that you set the download policy to *Immediate* for all repositories within the content view you export.
 For more information, see xref:Download_Policies_Overview_{context}[].
 * Ensure that you synchronize products you export to the required date.
--- a/guides/common/modules/proc_exporting-a-repository-in-the-syncable-format.adoc
+++ b/guides/common/modules/proc_exporting-a-repository-in-the-syncable-format.adoc
@@ -11,6 +11,7 @@ include::snip_generating-content.adoc[]
 include::snip_content-types-syncable-export.adoc[]
 
 .Prerequisites
+include::snip_allowed-export-paths-prerequisite.adoc[]
 * Ensure that you set the download policy to *Immediate* for the repository within the Library lifecycle environment you export.
 For more information, see xref:Download_Policies_Overview_{context}[].
 
--- a/guides/common/modules/proc_exporting-a-repository-in-the-syncable-format.adoc
+++ b/guides/common/modules/proc_exporting-a-repository-in-the-syncable-format.adoc
@@ -11,7 +11,9 @@ include::snip_generating-content.adoc[]
 include::snip_content-types-syncable-export.adoc[]
 
 .Prerequisites
+ifdef::containerized[]
 include::snip_allowed-export-paths-prerequisite.adoc[]
+endif::[]
 * Ensure that you set the download policy to *Immediate* for the repository within the Library lifecycle environment you export.
 For more information, see xref:Download_Policies_Overview_{context}[].
 
--- a/guides/common/modules/proc_exporting-a-repository-incrementally-in-the-syncable-format.adoc
+++ b/guides/common/modules/proc_exporting-a-repository-incrementally-in-the-syncable-format.adoc
@@ -14,6 +14,9 @@ include::snip_content-types-syncable-export.adoc[]
 
 The procedure below shows an incremental export of a repository in the Library lifecycle environment.
 
+.Prerequisites
+include::snip_allowed-export-paths-prerequisite.adoc[]
+
 .Procedure
 . Create an incremental export:
 +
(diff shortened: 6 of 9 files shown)


aneta-petrova:
Hi @aidenfine, can you please take a look at this PR and let me know if you think these are the changes needed to cover https://github.com/theforeman/foremanctl/pull/455?

Note that the preview links in https://github.com/theforeman/foreman-documentation/pull/4905#issuecomment-4619777379 include containerized and non-containerized docs preview builds. So for example to review the foremanctl changes, the right HTML builds to look at would be mainly https://theforeman-foreman-documentation-preview-pr-4905.surge.sh/nightly/Managing_Content/index-containerized-katello.html.

guides/common/modules/proc_creating-a-local-source-for-a-custom-file-type-repository.adoc
```diff
@@ -37,12 +37,22 @@ endif::[]
...
 ----
+endif::[]
+ifdef::containerized[]
+. Add the parent folder to allowed import paths:
++
+[options="nowrap", subs="+quotes,attributes"]
+----
+# {foremanctl} deploy --content-import-path /var/lib/pulp/__local_repos__
+----
+endif::[]
```
vsedmik, line 55: This is probably not the only part that needs update in this chapter for the containerized SAT, the preceding foreman-maintain will need an update too.

vsedmik (approved):
Left one note bellow which is going to be addressed by another PR I have been told, so ACK here, looks good to me. 👍

guides/common/modules/proc_creating-a-local-source-for-a-custom-file-type-repository.adoc
aneta-petrova: Looking at the satellite-maintain instructions above makes me wonder if they're even correct right now. Line 30 is telling users to use `satellite-maintain packages` after unlocking package management, and that doesn't seem right (https://github.com/theforeman/foreman-documentation/blob/master/guides/common/modules/ref_managing-packages-on-the-base-operating-system.adoc).

guides/common/modules/snip_allowed-export-paths-prerequisite.adoc
```diff
@@ -0,0 +1,11 @@
+:_mod-docs-content-type: SNIPPET
+
+* Exports are written to one of your allowed export paths.
+ifndef::containerized[]
+The default export path is `/var/lib/pulp/exports/`.
+To configure additional paths, use `ALLOWED_EXPORT_PATHS` in `/etc/pulp/settings.py`.
+endif::[]
+ifdef::containerized[]
+The default export path is `/var/lib/pulp/exports/`.
```
maximiliankolb, line 4: line 5 and 9 are identical, you could move this line below line 3.

maximiliankolb (commented):
Two minor suggestion, overall LGTM style-wise.

I also suggest to swap the order in the snippet file name:

````
$ fd snip_ | rg prereq
guides/common/modules/snip_prerequisite-activation-key-available.adoc
guides/common/modules/snip_prerequisite-configured-smart-proxy-registration-provisioning.adoc
guides/common/modules/snip_prerequisite-networking-for-provisioning.adoc
guides/common/modules/snip_prerequisite-project-client-repository-ak.adoc
guides/common/modules/snip_prerequisite-project-client-repository-enabled.adoc
guides/common/modules/snip_prerequisite-repositories-with-oscap.adoc
guides/common/modules/snip_prerequisites-common-compute-resource.adoc
guides/common/modules/snip_prerequisites-configuring-smartproxyservers-for-load-balancing.adoc
guides/common/modules/snip_prerequisites-deploying-ca-cert-rex.adoc
guides/common/modules/snip_registering-a-host-prerequisites.adoc
````

So maybe "snip_prerequisite-allowed-export-paths.adoc`.

**Comment by maximiliankolb** [comment]
Tech wise, I think that `foreman-installer` can already configure those paths in some cases, so I would appreciated it someone double-checked this:

````
# foreman-installer --full-help | grep -i import
    --foreman-proxy-puppet                                                               Enable Puppet module for environment imports and Puppet runs (current: true)
    --foreman-proxy-content-pulpcore-additional-import-paths                             Additional allowed paths that Pulpcore can use for content imports, or sync from using file:// protocol (current: [])
    --reset-foreman-proxy-content-pulpcore-additional-import-paths                       Reset pulpcore_additional_import_paths to the default value ([])
    --foreman-proxy-content-pulpcore-import-workers-percent                              What percentage of available-workers will pulpcore use for import tasks at a time (current: UNDEF)
    --reset-foreman-proxy-content-pulpcore-import-workers-percent                        Reset pulpcore_import_workers_percent to the default value (UNDEF)
# foreman-installer --full-help | grep -i export
    --foreman-proxy-content-pulpcore-additional-export-paths                             Additional allowed paths that Pulpcore can use for content exports (current: [])
    --reset-foreman-proxy-content-pulpcore-additional-export-paths                       Reset pulpcore_additional_export_paths to the default value ([])
````

Tested on Foreman 3.18/Katello 4.20.

File: guides/common/modules/proc_importing-syncable-exports.adoc [hunk 3]
```diff
@@ -7,8 +7,15 @@
 You can import syncable exports into the Library environment of an organization.
 
 .Prerequisites
-* The syncable exports must be located in one of your `ALLOWED_IMPORT_PATHS` as specified in `/etc/pulp/settings.py`.
-By default, this includes `/var/lib/pulp/imports/`.
+* The syncable exports must be located in one of your allowed import paths.
+ifndef::containerized[]
+The default import path is `/var/lib/pulp/imports/`.
+To configure additional paths, use `ALLOWED_IMPORT_PATHS` in `/etc/pulp/settings.py`.
```
ekohl: Users should **never** modify `settings.py` themselves. Even with `satellite-installer`. Users can use `{foreman-installer} --foreman-proxy-content-pulpcore-additional-import-paths $path` to add additional paths. Looks like this bug was originally introduced in 1afa311ae7b3b115c46407c25f002752dbb31b88.

Note that in https://github.com/theforeman/foremanctl/blob/master/docs/user/parameters.md#mapped we document this parameter mapping.

guides/common/modules/proc_importing-syncable-exports.adoc
```diff
@@ -7,8 +7,15 @@
 You can import syncable exports into the Library environment of an organization.
 
 .Prerequisites
-* The syncable exports must be located in one of your `ALLOWED_IMPORT_PATHS` as specified in `/etc/pulp/settings.py`.
-By default, this includes `/var/lib/pulp/imports/`.
+* The syncable exports must be located in one of your allowed import paths.
+ifndef::containerized[]
+The default import path is `/var/lib/pulp/imports/`.
```
ekohl: Not quite true: https://github.com/theforeman/puppet-foreman_proxy_content/blob/9f441f35d1a403f7576aa2845353c2d6d40246f9/manifests/init.pp#L151-L161 shows the logic. On content proxies (downstream: Capsule) it's `/var/lib/pulp/sync_imports` while on the main server (downstream: Satellite) it's both `/var/lib/pulp/sync_imports` and `/var/lib/pulp/imports`.

guides/common/modules/snip_allowed-export-paths-prerequisite.adoc
```diff
@@ -0,0 +1,11 @@
+:_mod-docs-content-type: SNIPPET
+
+* Exports are written to one of your allowed export paths.
+ifndef::containerized[]
+The default export path is `/var/lib/pulp/exports/`.
+To configure additional paths, use `ALLOWED_EXPORT_PATHS` in `/etc/pulp/settings.py`.
```
ekohl: Similar to above: please use `--foreman-proxy-content-pulpcore-additional-export-paths`. It's also odd that it mentions `--content-export-path` doesn't have an existing entry while `--foreman-proxy-content-pulpcore-additional-export-paths` exists for that.

ekohl (changes requested):
Please keep the installer and foremanctl commands in sync as much as possible. I realize it was already wrong before, but this is a good opportunity to fix it. Note that users who manually modify `settings.py` will lose the value on the next installer run and we also can't migrate them from non-containerized to containerized. For that latter part is exactly why we maintain the parameter mapping.

aidenfine:
> Hi @aidenfine, can you please take a look at this PR and let me know if you think these are the changes needed to cover [theforeman/foremanctl#455](https://github.com/theforeman/foremanctl/pull/455)?
> 
> Note that the preview links in [#4905 (comment)](https://github.com/theforeman/foreman-documentation/pull/4905#issuecomment-4619777379) include containerized and non-containerized docs preview builds. So for example to review the foremanctl changes, the right HTML builds to look at would be mainly https://theforeman-foreman-documentation-preview-pr-4905.surge.sh/nightly/Managing_Content/index-containerized-katello.html.

Sorry for late reply. I would agree with what @ekohl is saying.

aneta-petrova:
The `Unused modules` failure is probably due to https://github.com/theforeman/foreman-documentation/pull/4899/changes having been merged: Because the containerized guide's preview is empty, the new module seems unused. And it _is_ unused right now, but won't be when we start exposing and publishing the containerized contents.

aneta-petrova:
> The `Unused modules` failure is probably due to https://github.com/theforeman/foreman-documentation/pull/4899/changes having been merged: Because the containerized guide's preview is empty, the new module seems unused. And it _is_ unused right now, but won't be when we start exposing and publishing the containerized contents.

By the way, this is also the reason why the diff in https://github.com/theforeman/foreman-documentation/pull/4905#issuecomment-4619777379 is empty, so to review this type of change, there is no preview and looking at the changed lines should be enough.

aneta-petrova:
Hi @aidenfine and/or @ekohl, can you please re-review?

aidenfine (approved):
LGTM matches PR I created. Please have @ekohl also review since he may know in depth details about this :)

maximiliankolb (approved):
style-wise LGTM

ekohl (approved):

guides/common/modules/snip_allowed-export-paths-prerequisite.adoc
ekohl, line 1: The unused modules check complains `guides/common/modules/snip_allowed-export-paths-prerequisite.adoc` this is unused. So it looks like the content guide isn't being published yet.

If this is merged now, it'll be red on all future PRs and that's going to be misleading.

aneta-petrova: @ekohl See also https://github.com/theforeman/foreman-documentation/pull/4905#issuecomment-4692452077. I was going to consider it a false positive. I can't expose the content in the guide because other steps in the affected procedures must be updated for foremanctl first.

maximiliankolb: Ah, I didn't consider that the check runs for all files, not just modified files. Still, I am OK with red GHA for a while until we provide built docs for all build targets.

ekohl: I'd be worried about having to understand a check is red and ignore it.

2 options I'd see:
1. First get the content guide published so the module is published
2. Silence the check for that particular file

I'm guessing option 2 is an easier short term so I'd recommend:
* Open a GH issue to track it
* Patch `find_unused_modules`

For example, with the following:

```patch
diff --git a/guides/scripts/find_unused_modules b/guides/scripts/find_unused_modules
index 13b58e691441..757eda933213 100755
--- a/guides/scripts/find_unused_modules
+++ b/guides/scripts/find_unused_modules
@@ -43,6 +43,10 @@ unused = ((modules + assemblies) - absolute_included).map do |path|
   Pathname.new(path).relative_path_from(doc_root).to_s
 end
 
+# containerized content build isn't published
+# https://github.com/theforeman/foreman-documentation/issues/....
+unused.delete('guides/common/modules/snip_allowed-export-paths-prerequisite.adoc')
+
 if unused.any?
   puts unused
   if ENV.key?('GITHUB_ACTIONS')
```

maximiliankolb: Great! So now that we already have a patch for the script, I lean towards extending the script with this option for blocklist certain modules that we know are used but not published which makes the GHA turn red. -> So green GHA for all; and a tiny bit more work adding a line to the script.

aneta-petrova:
@maximiliankolb I extended the script and raised https://github.com/theforeman/foreman-documentation/issues/4964 to track the exception. Can you please re-review?

maximiliankolb (approved):
Thank you Anet, LGTM!

guides/scripts/find_unused_modules
```diff
@@ -43,6 +43,10 @@ unused = ((modules + assemblies) - absolute_included).map do |path|
   Pathname.new(path).relative_path_from(doc_root).to_s
 end
 
+# containerized content build isn't published
+# https://github.com/theforeman/foreman-documentation/issues/....
```
ekohl: Please fill in the issue number
```suggestion
# https://github.com/theforeman/foreman-documentation/issues/4964
```

guides/scripts/find_unused_modules
```diff
@@ -43,6 +43,10 @@ unused = ((modules + assemblies) - absolute_included).map do |path|
...
+unused.delete('guides/common/modules/snip_allowed-export-paths-prerequisite.adoc')
```
ekohl, line 48: Noting down my thought and don't feel you need to implement this now. I wondered if we could force the cleanup of this somehow. Should be relatively straight forward by checking it actually deleted something.
```suggestion
unless unused.delete('guides/common/modules/snip_allowed-export-paths-prerequisite.adoc')
  puts "Close https://github.com/theforeman/foreman-documentation/issues/4964"
  exit 1
end
```

aneta-petrova:
Thanks, everyone!

PR #4922, opened 2026-06-10
Optimize (CQA) - Mng content - ISS structure

Lennonka:
#### What changes are you introducing?

Optimization of headings and structure in the chapter _Synchronizing content between Foreman servers_

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

- Moving configuration from an appendix to the beginning of the chapter - cleaner for upstream users
- Optimization of structure - grouped exporting and importing procedures for better navigation via TOC
- Improved some headings for better descriptivity, e.g. "a syncable format" -> "the syncable format" because there's only one syncable format
- Headings have to be 3-11 words long

#### Contributor checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.19/Katello 4.21
* [x] Foreman 3.18/Katello 4.20 (Satellite 6.19)
* [x] Foreman 3.17/Katello 4.19
* [x] Foreman 3.16/Katello 4.18 (Satellite 6.18; orcharhino 7.6, 7.7, and 7.8)
* [ ] Foreman 3.15/Katello 4.17
* [ ] Foreman 3.14/Katello 4.16 (Satellite 6.17; orcharhino 7.4; orcharhino 7.5)
* [ ] Foreman 3.13/Katello 4.15 (EL9 only)
* [ ] Foreman 3.12/Katello 4.14 (Satellite 6.16; orcharhino 7.2 on EL9 only; orcharhino 7.3)
* We do not accept PRs for Foreman older than 3.12.

PR #4923, opened 2026-06-10
Add ISS to normalization strings in the pipeline check

Lennonka:

Adding ISS attributes to the Filename-ID-Heading check

image

Found in #4922 

#### Contributor checklists

Please cherry-pick my commits into: **???**

### DIFF
--- a/scripts/check-filename-id-heading-match.sh
+++ b/scripts/check-filename-id-heading-match.sh
@@ -72,6 +72,7 @@ check_file() {
     normalized_id="${normalized_id//\{freeipa-context\}/freeipa}"
     normalized_id="${normalized_id//\{insights-id\}/insights}"
     normalized_id="${normalized_id//\{insights-iop-id\}/insights}"
+    normalized_id="${normalized_id//\{ISS-id\}/iss}"
     normalized_id="${normalized_id//\{foreman-installer\}/foreman-installer}"
     normalized_id="${normalized_id//\{awx-context\}/awx}"
     normalized_id="${normalized_id//\{compute-resource-context\}/computeresource}"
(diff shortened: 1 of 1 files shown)


scripts/check-filename-id-heading-match.sh
```diff
@@ -72,6 +72,7 @@ check_file() {
     normalized_id="${normalized_id//\{freeipa-context\}/freeipa}"
     normalized_id="${normalized_id//\{insights-id\}/insights}"
     normalized_id="${normalized_id//\{insights-iop-id\}/insights}"
+    normalized_id="${normalized_id//\{ISS-id\}/iss}"
```
Lennonka: Does the normalization string have to match the filename substring or not? @aneta-petrova

aneta-petrova: Yes, it does. This should work:

```suggestion
    normalized_id="${normalized_id//\{ISS-id\}/inter-server-synchronization}"
```

scripts/check-filename-id-heading-match.sh
```diff
@@ -95,6 +96,7 @@ check_file() {
         normalized_heading="${normalized_heading//\{freeipa\}/freeipa}"
         normalized_heading="${normalized_heading//\{insights\}/insights}"
         normalized_heading="${normalized_heading//\{insights-iop\}/insights}"
+        normalized_heading="${normalized_heading//\{ISS\}/iss}"
```
aneta-petrova: The headings are converted to lowercase first, so the attributes here should actually be lowercase for the check to pass:

```suggestion
        normalized_heading="${normalized_heading//\{iss\}/inter-server-synchronization}"
```

aneta-petrova (commented):
The normalized_id part of the check is still failing because while that line in the script uses `{ISS-id}`, the actual ID in `con_inter-server-synchronization-configurations.adoc` is `{iss-id}`. AsciiDoc attributes are not case sensitive, but the bash script string replacement is.

My preference is to update the ID in the file to match the attribute's definition in attributes-base.adoc, which is `:ISS-id:`. That means changing `[id="{iss-id}-configurations"]` to `[id="{ISS-id}-configurations"]`.

There's also the alternative of changing the normalization line from `{ISS-id}` to `{iss-id}`. I have not tested that solution but it should also ensure a green check.

aneta-petrova (approved):

PR #4982, opened 2026-06-26
Review headings in Admin

aneta-petrova:
#### What changes are you introducing?

Reviewing headings to comply with the requirements in https://github.com/theforeman/foreman-documentation/blob/master/.claude/skills/heading/SKILL.md
Mostly, this involved making sure each heading has at least 3 words.

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

Part of the downstream Content Quality Assessment (CQA) task

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)

Assisted-by: Cursor

#### Contributor checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.19/Katello 4.21
* [x] Foreman 3.18/Katello 4.20 (Satellite 6.19)
* [x] Foreman 3.17/Katello 4.19
* [x] Foreman 3.16/Katello 4.18 (Satellite 6.18; orcharhino 7.6, 7.7, and 7.8)
* [ ] Foreman 3.15/Katello 4.17
* [ ] Foreman 3.14/Katello 4.16 (Satellite 6.17; orcharhino 7.4; orcharhino 7.5)
* We do not accept PRs for Foreman older than 3.14.

Issue #4964, opened 2026-06-24
Clear exceptions in find_unused_modules after containerization work is finished


aneta-petrova:
https://github.com/theforeman/foreman-documentation/pull/4905#discussion_r3464815469 resulted in adding an exception to the find_unused_modules script. It's possible that other containerization-related PRs will add more exceptions. We should clear all these exceptions after we start publishing all containerized documentation -- then the exceptions won't be needed anymore.

aneta-petrova:
https://github.com/theforeman/foreman-documentation/pull/4995 is another PR that proposes additional exceptions (in progress, not merged yet).

PR #4995, opened 2026-06-30
Switch Satellite builds to containerized

aneta-petrova:

* Switching Satellite preview build to show the containerized version of docs
* Dropping the containerized-satellite build target

We shouldn't need two separate Satellite builds anymore

Assisted-by: Claude

#### Contributor checklists


maximiliankolb:
This is a big step. :tada:

Lennonka (approved):
The changeset looks sane to me.

jafiala (approved):

aneta-petrova:
Similar to https://github.com/theforeman/foreman-documentation/pull/4905, I extended the find_unused_modules script with exceptions. These are needed because quite a few modules really are unused now -- they are satellite-only but not yet containerized, so they are not included anywhere (yet).

guides/common/attributes-containerized.adoc
```diff
@@ -4,7 +4,7 @@
 // This file is included by all containerized flavor attribute files:
 // - attributes-containerized-katello.adoc
 // - attributes-containerized-orcharhino.adoc
-// - attributes-containerized-satellite.adoc
+// - attributes-satellite.adoc (satellite is containerized by default)
```
ekohl: By default implies there's an option, but there isn't
```suggestion
// - attributes-satellite.adoc (always containerized)
```

guides/common/attributes-satellite.adoc
```diff
@@ -197,3 +197,9 @@
 :project-client-name: Red Hat Satellite Client 6
 :project-client-RHEL7-url: {RepoRHEL7ServerSatelliteToolsProjectVersion}
 :project-client-RHEL8-url: {RepoRHEL8ServerSatelliteToolsProjectVersion}
+
+// Containerized deployment overrides
+include::attributes-containerized.adoc[]
+
+// Satellite-specific containerized overrides
+:project-package-install: {package-install}
```
ekohl: Perhaps drop the definition on line 108 instead? While you're at it, lines 107 and 109 too?

guides/common/attributes.adoc
```diff
@@ -50,14 +49,11 @@ ifeval::["{build}" == "containerized-orcharhino"]
...
 endif::[]
-ifeval::["{build}" == "containerized-satellite"]
-:containerized:
-:satellite:
-endif::[]
 ifeval::["{build}" == "katello"]
 :katello:
 endif::[]
 ifeval::["{build}" == "satellite"]
+:containerized:
```
ekohl: If `guides/common/attributes-containerized.adoc` defines this attribute, can this file be simplified?

If we decide not to implement this now, I think that is still a good optimization to do right now.

guides/scripts/find_unused_modules
```diff
@@ -47,6 +46,95 @@ end
 # https://github.com/theforeman/foreman-documentation/issues/4964
 unused.delete('guides/common/modules/snip_allowed-export-paths-prerequisite.adoc')
 
+# Non-containerized Satellite content excluded since satellite build is now containerized-only
```
ekohl, line 49: If we're skipping so many, perhaps implement the safeguard I mentioned last time: https://github.com/theforeman/foreman-documentation/pull/4905#discussion_r3465289151. Otherwise, what's the plan to actually remove these entries again?

ekohl (commented):
Please consider all the inline comments as thoughts when finally pulling the trigger.

I'm not strictly opposed to this, but there is a risk to having unintended consequences when doing cherry picks. Have you considered this risk?

guides/scripts/find_unused_modules
aneta-petrova: I wrote https://github.com/theforeman/foreman-documentation/issues/4964 as a generic issue and now I've posted a link to this PR in there so hopefully that's sufficient to track these removals as well.

ekohl: That also works for me.

guides/common/attributes.adoc
aneta-petrova: Done -- the `:containerized:` attr is now included directly in `attributes-satellite.adoc` rather than here

guides/common/attributes-satellite.adoc
aneta-petrova: Dropped all project-package related attributes

aneta-petrova:
I've implemented the feedback, and along with the list of exceptions in the `find_unused_modules`, that means a lot of changes since the acks were granted so I'd like to ask for re-reviews.

aneta-petrova:
We had a separate conversation about the cherry-picking concerns elsewhere, so let me summarize it here for maximum transparency:

There are some concerns due to our cherry-picking workflows -- removing the non-containerized build in preview might create a blind spot: If I update something on master and only view the containerized preview, I might accidentally break things on the non-containerized preview after the update is cherry-picked to 3.19.

However, the other build targets still include non-containerized previews, which would mitigate the risk.

Option 1: If we want to feel extra safe about preventing that risk, we could mandate separate cherry-picking PRs for branches 3.19 and below (the workflow would be 1. open a PR against master + get it reviewed, 2. open a new PR against 3.19 with the same changes + get it reviewed).

Option 2: If we don't feel like we need this extra safeguard and the non-satellite non-containerized previews are enough, we should be able to continue cherry-picking as usual. Without mandating a separate PR for 3.19.

Please @ekohl @jafiala @Lennonka feel free to review if I captured all arguments and counter-arguments correctly.

aneta-petrova:
As the author of the PR, I'm leaning towards Option 2 :) provided that all maintainers are mindful of the difference between containerized and non-containerized. Which, I think, by now we all are.

Lennonka (approved):
LGTM

Option 2

maximiliankolb:
Option 2.

jafiala:
Option 2 as well.

jafiala (approved):

guides/common/attributes-satellite.adoc
```diff
@@ -197,3 +194,7 @@
 :project-client-name: Red Hat Satellite Client 6
 :project-client-RHEL7-url: {RepoRHEL7ServerSatelliteToolsProjectVersion}
 :project-client-RHEL8-url: {RepoRHEL8ServerSatelliteToolsProjectVersion}
+
+// Containerized deployment overrides
+:containerized:
```
ekohl, line 201: I suggested to set the `containerized` attribute in `attributes-containerized.adoc` so it's set for all containerized builds. Then you can also drop it from the `ifeval::["{build}" == "containerized-katello"]` and `ifeval::["{build}" == "containerized-orcharhino"]` sections in `guides/common/attributes.adoc`.

aneta-petrova: This PR is about making changes to Satellite build targets and I'd rather not touch the other ones. In https://github.com/theforeman/foreman-documentation/pull/4995#discussion_r3499964270, I explained I'm making the change only for Satellite.

guides/scripts/find_unused_modules
ekohl: Minor nit I always like including the actual issue link within the code so there's an easy way to find out where it's tracked. Mind doing the same here?

aneta-petrova: I was hoping referencing the issue only once would be enough but okay, I'll take another look at this

aneta-petrova:
A rebase was needed to pull in a new attribute added in https://github.com/theforeman/foreman-documentation/pull/4996/changes
And I pushed an additional commit to reflect a change in https://github.com/theforeman/foreman-documentation/pull/4982 which renamed one of the modules that I'm adding to the exceptions in find_unused_modules

PR #5008, opened 2026-07-03
Add attribute for errata source based on Client OS

### LINKED ISSUE #4991: Introduce attribute for errata source based on Client OS
note for @maximiliankolb: Introduce attribute for source of errata by Client OS

_Originally posted by @maximiliankolb in https://github.com/theforeman/foreman-documentation/pull/4962#discussion_r3479443917_

maximiliankolb:

This patch ensures that for orcharhino builds, the errata source is based on the Client OS. Examples:

* RHEL Clients: errata come from Red Hat
* Ubuntu Clients: errata come from orcharhino via ATIX Errata Service

Fixes #4991 (Add attribute for Client errata source)
Refs "con_making-errata-available-through-content-views.adoc"
Refs https://orcharhino.com/en/about-orcharhino/features/

#### Contributor checklists

### DIFF
--- a/guides/common/attributes-orcharhino.adoc
+++ b/guides/common/attributes-orcharhino.adoc
@@ -80,6 +80,7 @@
 :client-content-type: Deb
 :client-cv-filter: Deb
 :client-enable-chrony-service: systemctl enable --now chrony
+:client-errata-source: ATIX Errata Service for Ubuntu
 :client-extracted-gpg-key-extracted-path: ~/_My_GPG_Public_Key_
 :client-extracted-gpg-key-target-path: ~/_My_GPG_Public_Key_
 :client-install-yggdrasil-package: apt-get install foreman-ygg-migration
--- a/guides/common/modules/ref_prerequisites-for-installing-errata-on-hosts.adoc
+++ b/guides/common/modules/ref_prerequisites-for-installing-errata-on-hosts.adoc
@@ -6,11 +6,14 @@
 [role="_abstract"]
 Installing errata on hosts requires synchronized repositories, installable errata in the content view environment, registered hosts, and remote execution configuration.
 
+ifdef::katello[]
+* Repositories on {ProjectServer} are synchronized with the latest errata available from upstream.
+endif::[]
+ifdef::orcharhino[]
+* Repositories on {ProjectServer} are synchronized with the latest errata available from {client-errata-source}.
+endif::[]
 ifdef::satellite[]
 * Repositories on {ProjectServer} are synchronized with the latest errata available from Red{nbsp}Hat.
-endif::[]
-ifndef::satellite[]
-* Repositories on {ProjectServer} are synchronized with the latest errata available from upstream.
 endif::[]
 * The host is registered to a content view environment.
 * The host is configured for remote execution.


guides/common/attributes-orcharhino.adoc
```diff
@@ -80,6 +80,7 @@
 :client-content-type: Deb
 :client-cv-filter: Deb
 :client-enable-chrony-service: systemctl enable --now chrony
+:client-errata-source: ATIX Errata Service for Ubuntu
```
maximiliankolb, line 83: @sbernhard Is there a specific name for the errata service for orcharhino customers? This example is for Ubuntu Clients.

sbernhard: I think, this is OK.

guides/common/modules/ref_prerequisites-for-installing-errata-on-hosts.adoc
```diff
@@ -6,12 +6,15 @@
 [role="_abstract"]
 Installing errata on hosts requires synchronized repositories, installable errata in the content view environment, registered hosts, and remote execution configuration.
 
+ifdef::katello[]
+* Repositories on {ProjectServer} are synchronized with the latest errata available from upstream.
+endif::[]
+ifdef::orcharhino[]
+* Repositories on {ProjectServer} are synchronized with the latest errata available from {client-errata-source}.
```
aneta-petrova, line 13: IMO a good use cases for an attribute is when the value varies across build targets. Have you considered simply doing this:

```suggestion
* Repositories on {ProjectServer} are synchronized with the latest errata available from ATIX Errata Service for Ubuntu.
```

rather than introducing the attribute?

(I know this is how we handle other attributes as well, but I'm curious about whether people really prefer it this way or not :) When an attribute is used for a single build target only, in my eyes it only adds complexity without much benefit.)

aneta-petrova (approved):
Ack, although I'm curious to hear some thoughts on my comment :) I'd prefer not to introduce attributes for a single build target but it's also not really blocking this change if this really is the preferred way.

guides/common/modules/ref_prerequisites-for-installing-errata-on-hosts.adoc
maximiliankolb: The reason that I add an attribute is that I can overwrite it in downstream for all other Client OS, e.g. Debian with a similar value, "Red Hat" for RHEL Clients, or "SUSE" for SLES. In downstream, this module is built nine times: https://docs.orcharhino.com/or/docs/sources/guides/ubuntu/managing_hosts/managing_packages.html#applying-errata-to-hosts-by-using-web-ui & https://docs.orcharhino.com/or/docs/sources/guides/rocky_linux/managing_hosts/managing_packages.html#applying-errata-to-hosts-by-using-web-ui etc.

Do you think that I should add the same attribute for Katello and Satellite too? Or keep it as is?

Lennonka: I'm okay with having this attribute only where it's needed, ie. keeping it as is.

We have a ton of satellite-only attributes in the satellite build. I think we should grant orcharhino the same privilege :)

aneta-petrova: > We have a ton of satellite-only attributes in the satellite build.

Yes, and the overall attribute situations adds up to a bit of a mess that is being reported as a significant hurdle in onboarding. Having fewer of them would help, and a single-use attributes are IMO not a good use case for attributes.

Not a blocker for the PR, though, as expressed by my earlier ack, which still stands.

maximiliankolb:
   1c92115742..daf84843fe  3.19 -> 3.19
   538e69ee5e..bcdf3f2792  3.18 -> 3.18

> single-use attributes

It is only used _once_ in the modules, but in downstream orcharhino docs, I will overwrite it differently _nine times_ to render different guides based on the Client OS.

The only alternative for me would be to duplicate the text:

````
ifdef::almalinux[]
* Repositories on {ProjectServer} are synchronized with the latest errata available from the AlmaLinux community.
endif::[]
ifdef::centos[]
* Repositories on {ProjectServer} are synchronized with the latest errata available from the CentOS community.
endif::[]
ifdef::oracle_linux[]
* Repositories on {ProjectServer} are synchronized with the latest errata available from Oracle.
endif::[]
...
````

PR #5009, opened 2026-07-03
Add Vale rule to limit procedure steps to 10

aneta-petrova:

Adding a new Vale rule to the foreman-documentation style package that reports an error when a procedure module contains more than 10 steps. The rule counts only high-level steps (lines starting with `. `) and ignores substeps (lines starting with `..`, `...`, etc.).

Procedures that have too many steps feel overwhelming. This proposes a convention (already applied downstream) to keep procedure steps to a maximum of 10.

N/A

#### Contributor checklists

### DIFF
--- a/.vale/styles/foreman-documentation/ProcedureStepLimit.yml
+++ b/.vale/styles/foreman-documentation/ProcedureStepLimit.yml
@@ -11,9 +11,19 @@ script: |
   count := 0
   step_positions := []
   position := 0
+  in_procedure := false
 
   for line in lines {
-    if text.re_match("^\\. [^\\.]", line) {
+    // Detect .Procedure heading (only one per file)
+    if text.re_match("^\\.Procedure", line) {
+      in_procedure = true
+    }
+    // Reset when we encounter another heading at the same level
+    if text.re_match("^\\.[A-Z]", line) && !text.re_match("^\\.Procedure", line) {
+      in_procedure = false
+    }
+    // Count steps only when inside a .Procedure section
+    if in_procedure && text.re_match("^\\. [^\\.]", line) {
       count += 1
       step_positions = append(step_positions, position)
     }
--- a/.vale/styles/foreman-documentation/ProcedureStepLimit.yml
+++ b/.vale/styles/foreman-documentation/ProcedureStepLimit.yml
@@ -9,11 +9,22 @@ script: |
   matches := []
   lines := text.split(scope, "\n")
   count := 0
+  step_positions := []
+  position := 0
+
   for line in lines {
     if text.re_match("^\\. [^\\.]", line) {
       count += 1
+      step_positions = append(step_positions, position)
     }
+    position += len(line) + 1
   }
+
   if count > 10 {
-    matches = append(matches, {begin: 0, end: 1})
+    // Report error on ALL steps to ensure it appears in GitHub Actions diff context
+    // when ANY step is modified in a procedure that exceeds the 10-step limit
+    for i := 0; i < count; i++ {
+      step_pos := step_positions[i]
+      matches = append(matches, {begin: step_pos, end: step_pos + 1})
+    }
   }
--- a/.vale/styles/foreman-documentation/ProcedureStepLimit.yml
+++ b/.vale/styles/foreman-documentation/ProcedureStepLimit.yml
@@ -1,7 +1,7 @@
 # Report procedures that contain more than 10 steps.
 ---
 extends: script
-message: "Procedure contains more than 10 steps. Split it into multiple procedures or leverage substeps to group related steps."
+message: "Procedure contains more than 10 steps. Split it into multiple procedures or leverage substeps to group related steps. If the procedure contains conditionals, review the preview to ensure all built targets comply with the limit of 10 steps."
 level: error
 scope: raw
 script: |
--- a/.vale/styles/foreman-documentation/ProcedureStepLimit.yml
+++ b/.vale/styles/foreman-documentation/ProcedureStepLimit.yml
@@ -1,7 +1,7 @@
 # Report procedures that contain more than 10 steps.
 ---
 extends: script
-message: "Procedure contains more than 10 steps. Consider splitting into multiple procedures."
+message: "Procedure contains more than 10 steps. Split it into multiple procedures or leverage substeps to group related steps."
 level: error
 scope: raw
 script: |
--- a/.vale/styles/foreman-documentation/ProcedureStepLimit.yml
+++ b/.vale/styles/foreman-documentation/ProcedureStepLimit.yml
@@ -1,0 +1,19 @@
+# Report procedures that contain more than 10 steps.
+---
+extends: script
+message: "Procedure contains more than 10 steps. Consider splitting into multiple procedures."
+level: error
+scope: raw
+script: |
+  text := import("text")
+  matches := []
+  lines := text.split(scope, "\n")
+  count := 0
+  for line in lines {
+    if text.re_match("^\\. [^\\.]", line) {
+      count += 1
+    }
+  }
+  if count > 10 {
+    matches = append(matches, {begin: 0, end: 1})
+  }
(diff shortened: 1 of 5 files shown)


aneta-petrova:
The rule seems to be working => ready for review.

The GHA adds an error message for every step in the procedure that's displayed in the diff. This amounts to a lot of messages but without it, for example, when the the warning was displayed only at the problematic step, if that problematic step wasn't included in the diff, the GHA workflow didn't pick up on it (because the workflow runs on the diff only).

guides/common/modules/proc_monitoring-server.adoc
```diff
@@ -15,3 +15,4 @@ You can use this information for maintenance and troubleshooting.
 ----
 # foreman-rake audits:list_attributes
 ----
+. Add another step.
```
aneta-petrova: This doesn't trigger the error because the procedure now has 3 steps.

guides/common/modules/proc_performing-post-cloning-tasks.adoc
```diff
@@ -29,3 +29,4 @@ For more information, see {AdministeringDocURL}renaming-{project-context}_admin[
 Copy all the `virt-who` configuration files in the `/etc/virt-who.d/` directory from the source server to the same directory on the target server.
 For more information, see {ConfiguringVMSubscriptionsDocURL}[_{ConfiguringVMSubscriptionsDocTitle}_].
 . Decommission the source server.
+. Add another step.
```
aneta-petrova: This doesn't trigger the error because the total number of steps is less than 10. There are some substeps, and if the Vale rule counted steps + substeps, this would be an error. But we're intentionally excluding substeps and counting only steps.

guides/common/modules/proc_synchronizing-ansible-collections.adoc
```diff
@@ -22,6 +22,7 @@ Locate the *SSO URL* section to obtain the SSO URL.
 
 .Procedure
 . In the {ProjectWebUI}, navigate to *Content* > *Products*.
+. Add another step.
```
aneta-petrova: This triggers an error because Vale recognizes that the procedure has more than 10 steps. There is an error for each step to ensure that the error is picked up by the GHA workflow. If there wasn't an error for each step, it's possible that it would not be reported in the part that is included in the diff, and the GHA workflow would miss it. (GHA looks only at the diff.)

.vale/styles/foreman-documentation/ProcedureStepLimit.yml
```diff
@@ -0,0 +1,30 @@
+# Report procedures that contain more than 10 steps.
+---
+extends: script
+message: "Procedure contains more than 10 steps. Consider splitting into multiple procedures."
```
maximiliankolb: suggestion (non-blocking): Reprhase "consider".

IMO: If we make this an error (see line 5), then we should "force" users to restructure the procedure and not just make a suggestion. So maybe "consider" is too nice.

.vale/styles/foreman-documentation/ProcedureStepLimit.yml
```diff
@@ -0,0 +1,30 @@
...
+script: |
+  text := import("text")
+  matches := []
+  lines := text.split(scope, "\n")
+  count := 0
+  step_positions := []
+  position := 0
+
+  for line in lines {
+    if text.re_match("^\\. [^\\.]", line) {
```
maximiliankolb: out of scope: Does this also work for macros? Example:

````
ifdef::kvm-provisioning,openstack-provisioning[. In the *Password* field, enter the SSH password for image access.]
````

in guides/common/modules/proc_adding-compute-resource-images-to-project-by-using-web-ui.adoc

IMO it's better to keep the logic small and have some missed edge cases than overcomplicate it.

Another example: including snippets that contain steps.

**Review by maximiliankolb — COMMENTED** [review]
I like this check!

File: .vale/styles/foreman-documentation/ProcedureStepLimit.yml (hunk 5 above)
Comment (aneta-petrova) [reply .vale/styles/foreman-documentation/ProcedureStepLimit.yml]: Oh, that's a good point. The script wouldn't recognize and process conditionals correctly. Which will result in false positives -- when a procedure seemingly includes more than 10 steps, but in reality the preview doesn't. But I guess a false positive that can be ignored is better than missing an actual error? :upside_down_face:  I'm not sure how to address this in the script, to be honest. Would it work to extend the warning message with information such as "If your procedure includes conditionals, then check your preview because this might be a false positive"?

Snippets are also missed, yes. We have https://github.com/jhradilek/asciidoctor-dita-vale/blob/0d82409212cd9dabccc26cf229c1eb5360c8ecfa/README.md?plain=1#L143 which reports whenever a snippet is included in a procedure so that's at least something. TaskInclude lets you know there's an include that you need to check for compliance yourself.

File: .vale/styles/foreman-documentation/ProcedureStepLimit.yml (hunk 4 above)
Comment (aneta-petrova) [reply .vale/styles/foreman-documentation/ProcedureStepLimit.yml]: You're right. There's another solution to leverage substeps so I'm adding that as well as a hint in the error message.

File: .vale/styles/foreman-documentation/ProcedureStepLimit.yml (hunk 5 above)
Comment (maximiliankolb) [reply .vale/styles/foreman-documentation/ProcedureStepLimit.yml]: I think both limitations are fine for now. I would't mind using this as is for now to get a feeling for how it works.

Comment (maximiliankolb) [reply .vale/styles/foreman-documentation/ProcedureStepLimit.yml]: >  Would it work to extend the warning message with information such as "If your procedure includes conditionals, then check your preview because this might be a false positive"?

Yes.

**Comment by aneta-petrova** [comment]
Some more testing:

1. I ran the new check on all our procedure modules and it returned a very high number of errors:

```
vale --filter='.Name=="foreman-documentation.ProcedureStepLimit"' guides/common/modules/proc_*.adoc
...
✖ 1075 errors, 0 warnings and 0 suggestions in 966 files.
```

This is because each step in a procedure that's too long produces an error, so a procedure with 11 steps amounts to 11 errors. (The logic behind that behavior is explained in https://github.com/theforeman/foreman-documentation/pull/5009#issuecomment-4875040901.)

However, in reality, it's 72 procedures in total that have more than 10 steps.

2. The rule reports also numbered lists in files that are of different content type than procedure. It would be better if it ran only on procedures so I'll try to looking into adjusting it for this behavior.

File: guides/common/modules/con_high-level-steps-for-configuration-management-with-puppet.adoc [hunk 6]
```diff
@@ -16,6 +16,9 @@ Configuration management with Puppet involves the following high-level steps:
 . Configuring intervals for runs of the OpenVox agent on hosts and for configuration enforcement runs of the OpenVox server.
 . Monitoring configuration management using reports in the {ProjectWebUI}.
 . Configuring email notifications.
+. List item 9.
```
aneta-petrova: This doesn't trigger an error because the numbered list is not within a `.Procedure` section.

aneta-petrova:
>     2. The rule reports also numbered lists in files that are of different content type than procedure. It would be better if it ran only on procedures so I'll try to looking into adjusting it for this behavior.

This is now addressed by making sure the Vale rule runs only on numbered lists that follow a `.Procedure` section.

aneta-petrova:
> Some more testing:
> 
>     1. I ran the new check on all our procedure modules and it returned a very high number of errors:
> 
> 
> ```
> vale --filter='.Name=="foreman-documentation.ProcedureStepLimit"' guides/common/modules/proc_*.adoc
> ...
> ✖ 1075 errors, 0 warnings and 0 suggestions in 966 files.
> ```
> 
> This is because each step in a procedure that's too long produces an error, so a procedure with 11 steps amounts to 11 errors. (The logic behind that behavior is explained in [#5009 (comment)](https://github.com/theforeman/foreman-documentation/pull/5009#issuecomment-4875040901).)
> 
> However, in reality, it's 72 procedures in total that have more than 10 steps.

The update to look only at steps within `.Procedure` also addressed a previously undiscovered bug where the script counted steps in the whole file, including steps in `.Verification` and such.

After the latest update that ensures only content in `.Procedure` is checked, the number of errors has dropped from 1075 to 1017.

aneta-petrova:
I have now reverted the commit that I used to show how the rule works and what issues it catches. @maximiliankolb, what do you think?

maximiliankolb (approved):
Great, works for me! I focused on the message; I cannot comment much about the code.

PR #5013, opened 2026-07-03
Add a Vale rule to check abstract length

### LINKED ISSUE #4547: Add Vale rule: Character limit for abstracts
Each abstract should be 50-300 characters long.

Source: Downstream requirement.

aneta-petrova:

Adding a new Vale rule that reports paragraphs that do not conform to the length requirement for abstracts.

Abstracts must be 50-300 characters long.

There are some limitations with this implementation:

* Vale checks the source only, which means that it will count the characters of an attribute (`{RHBK}`) but not the string the attribute translates to in a given build context (`Red Hat build of Keycloak`). This can mean that the real abstract in the preview build could be longer than what Vale checks, potentially exceeding the 300 character limit.
* Some abstracts use conditionals, for example:

```
[role="_abstract"]
ifdef::satellite[]
This is the abstract for Satellite.
endif::[]
ifndef::satellite[]
This is the same abstract, but for other build targets.
endif::[]
```

This, of course, would also not report accurate character count.

IMO this doesn't really pose an issue for the rule as I propose it. Upstream, even an approximate character count limit helps ensure abstracts don't get too long. Downstream, we'll have an additional check (but the upstream one will help us catch some violations before merging, which is still helpful).

#### Contributor checklists

### DIFF
--- a/.vale/styles/foreman-documentation/AbstractLength.yml
+++ b/.vale/styles/foreman-documentation/AbstractLength.yml
@@ -1,7 +1,7 @@
 # Check that abstract paragraphs are between 50 and 300 characters long.
 ---
 extends: script
-message: "Abstract must be between 50 and 300 characters."
+message: "Abstract must be between 50 and 300 characters. Place any information that exceeds 300 characters in a new paragraph or paragraphs. It cannot be part of the abstract."
 level: error
 scope: raw
 script: |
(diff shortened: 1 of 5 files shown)


maximiliankolb:
> IMO this doesn't really pose an issue for the rule as I propose it.

I agree.

Lennonka (approved):
Thank you! Looks good to me.

aneta-petrova:
I reverted the test commit whose purpose was to demonstrate and test the behavior. I'll merge with green GHA.

PR #5030, opened 2026-07-08
Fix proxy structure 

jafiala:
Reorganize the chapter to better follow user flow before containerized content gets added.

Preparation for https://redhat.atlassian.net/browse/SAT-47538

#### Contributor checklists


guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -25,8 +34,64 @@ ifdef::foreman-el,foreman-deb[]
 --no-enable-foreman-cli \
 --puppet-server-ca false
 ----
+. Enable the Registration and Templates features on your {SmartProxyServer} and set the template URL:
```
ekohl: I don't think I'd move this part into the primary installation. IMHO this should still go into the additional conf section.

This also has some considerations on what is default in which installation.

guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -25,8 +34,64 @@ ifdef::foreman-el,foreman-deb[]
...
+----
+# firewall-cmd --permanent --zone=public --add-port=8000/tcp
+----
+. On your {SmartProxyServer}, reload the firewall configuration:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# firewall-cmd --reload
+----
+. On {ProjectServer}, add the {SmartProxy} to the list of trusted proxies.
```
ekohl: Here a new section starts: add the proxy to the list of trusted proxies. I know it's confusing, but it's only needed on Katello (and its derivatives). That's because it's adding it as a trusted HTTP proxy, which is needed on content proxies where Apache is a reverse proxy (for RHSM).

It doesn't apply to a plain Foreman.

guides/common/modules/proc_installing-smart-proxy-server.adoc
ekohl: On content proxies (and downstream Capsules) the registration feature is enabled out of the box. This whole part should be only shown for `foreman-el` and `foreman-deb`. And I'd keep it in the additional configuration as it was because registration is an optional feature.

guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -4,11 +4,20 @@
 = Installing {SmartProxyServer}
 
 [role="_abstract"]
-With {SmartProxy} packages installed, you can proceed to install the {SmartProxyServer}.
+Install {SmartProxyServer} and configure it for host registration and provisioning.
+
+.Prerequisites
+* `{smartproxy-installer-package}` packages are installed.
```
ekohl: This should not be a prerequisite but the procedure should tell the reader how to install this.

ekohl (changes requested):
This PR contains various things that I wouldn't combine in a single PR. The first commit is a bugfix that should be merged and cherry picked. The other parts not because they're restructuring things which I wouldn't do in an existing release.

I'd keep the containerization stuff to the PR that actually implements the containerization. That keeps this PR easier to review IMHO.

guides/common/modules/proc_installing-smart-proxy-server.adoc
jafiala: Don't you think everyone who's managing proxies knows how to install a package?

ekohl: IMHO it needs to be explicit because users also like to copy-paste.

guides/common/modules/proc_installing-smart-proxy-server.adoc
jafiala: Hopefully addressed in 4c289c0da91b3e836f7813b5a97eb6d472b82c7c

guides/common/modules/proc_installing-smart-proxy-server.adoc
jafiala: reverted to the previous state. Please check that's what you intend.

guides/doc-Installing_Proxy/master.adoc
```diff
@@ -29,13 +29,13 @@ include::common/assembly_preparing-your-environment-for-smart-proxy-installation
 include::common/assembly_installing-smart-proxy-server.adoc[leveloffset=+1]
 
 include::common/assembly_performing-additional-configuration-on-smart-proxy-server.adoc[leveloffset=+1]
+endif::[]
```
ekohl, line 32: I don't think this diff is right and it now shows the scalability concern for containerized builds.

guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -7,6 +7,18 @@
...
+# {package-upgrade}
+----
+. Install `{smartproxy-installer-package}`:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {package-install} {smartproxy-installer-package}
+----
 ifdef::foreman-el,foreman-deb[]
 * To install {SmartProxyServer}, enter the following command:
```
ekohl, line 11: I think this now deserves its own item.
```suggestion
. To install {SmartProxyServer}, enter the following command:
```

I'm unsure if the same is true for Katello below. Probably not, but what do you think?

ekohl (commented):
I think this is largely what I had in mind. Some minor details inline.

guides/doc-Installing_Proxy/master.adoc
jafiala: Sorry, which diff?

Got it, working on it.

guides/doc-Installing_Proxy/master.adoc
```diff
@@ -17,7 +17,7 @@ ifdef::katello[]
 include::common/modules/ref_guide-not-ready.adoc[leveloffset=+1]
 endif::[]
 endif::[]
-ifndef::HideDocumentOnStable,foreman-deb,containerized[]
```
ekohl, line 20: I'd prefer to keep containerized changes out of this PR. I don't think you need to touch this file at all here.

jafiala: This is just a temporary change to make the sat preview visible while I troubleshoot my mess.
I fully intend to revert it, as I already did most of my changes.

jafiala:
So this warrants some explanation.

Revert almost all changes because I finally realized they don't make sense.

Separate package install proc is still valuable because we do want to keep the dnf install step, and katello+satellite don't get the install proc, just the ssl procs. That invalidates most of the other changes.

I wasted a lot of time :/

The only actual change is excluding the appendix from satellite.

stejskalleos (approved):

ekohl (changes requested):
> Revert almost all changes because I finally realized they don't make sense.
> 
> Separate package install proc is still valuable because we do want to keep the dnf install step, and katello+satellite don't get the install proc, just the ssl procs. That invalidates most of the other changes.

In that case, should we close this PR? Because in its current form it's adding the scalability considerations to containerized builds which is something we explicitly don't want.

And looking back at this PR, I think it tried to do too many things at the same time and that's why a lot of time was lost. In the future I urge you to start with smaller PRs that have a single dedicated focus. For a reviewer that's easier and you also don't risk losing as much time on a dead end.

PR #5031, opened 2026-07-08
Switch to  postgres 16 as a minimum requirement

ShimShtein:

Update the minimum version of PostgreSQL db to 16

We are switching from PostgreSQL 13 to 16.

#### Contributor checklists

### DIFF
--- a/guides/common/modules/proc_installing-postgresql-as-an-external-database.adoc
+++ b/guides/common/modules/proc_installing-postgresql-as-an-external-database.adoc
@@ -12,7 +12,10 @@ ifdef::foreman-deb[]
 The examples below assume PostgreSQL 14 as used on Ubuntu 22.04.
 If you run {ProjectServer} on Debian 12, you have to adjust the paths of PostgreSQL config and data directories which depend on the PostgreSQL version.
 endif::[]
-ifndef::foreman-deb[]
+ifdef::containerized[]
+{Project} supports PostgreSQL version 16, which is available in {EL} 10 repositories.
+endif::[]
+ifndef::foreman-deb,containerized[]
 {Project} supports PostgreSQL version 13.
 endif::[]
 
--- a/guides/common/modules/ref_postgresql-as-an-external-database-considerations.adoc
+++ b/guides/common/modules/ref_postgresql-as-an-external-database-considerations.adoc
@@ -15,7 +15,10 @@ ifdef::foreman-deb[]
 {Project} requires PostgreSQL version 13 or later.
 {Team} recommends that you use the PostgreSQL version that is part of the default operating system repositories.
 endif::[]
-ifndef::foreman-deb[]
+ifdef::containerized[]
+{Project} supports PostgreSQL version 16, which is available in {EL} 10 repositories.
+endif::[]
+ifndef::foreman-deb,containerized[]
 {Project} supports PostgreSQL version 13.
 endif::[]
 


maximiliankolb:
I am unsure about this change as is. I checked the currently available versions on the Foreman platform OS:

````
$ podman run --rm -it docker.io/library/ubuntu:jammy /bin/bash -c "apt-get update && apt-cache show postgresql" | rg "^Package:|^Version:"
Package: postgresql
Version: 14+238
$ podman run --rm -it docker.io/library/debian:bookworm /bin/bash -c "apt-get update && apt-cache show postgresql" | rg "^Package:|^Version:"
Package: postgresql
Version: 15+248+deb12u1
$ podman run --rm -it quay.io/almalinuxorg/almalinux:9.8 dnf info postgresql | rg "^Name|^Version|^Source"
Name         : postgresql
Version      : 13.23
Source       : postgresql-13.23-3.el9_8.src.rpm
````

This makes me believe that we cannot say both "please use PG 16" and "use the latest PG version shipped by your OS vendor".

Is there a ticket or any PR in foreman/katello/pulp to track this?

**Comment by maximiliankolb** [comment]
Aha:

````
$ podman run --rm -it quay.io/almalinuxorg/almalinux:10.2 dnf info postgresql | rg "^Name|^Version|^Source"
Name         : postgresql
Version      : 16.14
Source       : postgresql16-16.14-1.el10_2.src.rpm
````

Is this related to Foreman/Foreman+Katello on EL10?

**Comment by ShimShtein** [comment]
You are correct, I was just looking into it. 
I wonder if we should switch to recommending a containerized deployment for custom DB support too. This will definitely simplify our guidelines and make it OS-agnostic.

**Review by ekohl — CHANGES_REQUESTED** [review]
I lean to only state we require PostgreSQL 16 for the containerized installation and not touch the existing RPM/deb-based installations. We don't even test PostgreSQL 16 with those.

**Comment by ShimShtein** [comment]
>I lean to only state we require PostgreSQL 16 for the containerized installation and not touch the existing RPM/deb-based installations. We don't even test PostgreSQL 16 with those.

Let me reiterate the possibilities here, so we can all decide how to continue with this:
**Given:**
 RHEL < 10 and debian < 13 have lower versions of PG available
**Suggested solutions:**
1. stop recommending the PG version from the OS and point to postgres repos
2. set a minimum OS version for the postgres server
3. containerized PG (with the caveat that it is not intuitive for the users)
4. Remove specific instructions on how to obtain PG 16 from https://github.com/theforeman/foreman-documentation/blob/master/guides/common/modules/proc_installing-postgresql-as-an-external-database.adoc?plain=1 and only define the requirements and configuration.

**Comment by ekohl** [comment]
Let's look at the various builds we produce and what I'd recommend:
* foreman-deb: keep it as-is
* foreman-el: keep it as-is
* katello: keep it as-is
* orcharhino: keep it as-is (looking at @maximiliankolb for that)
* containerized: set the supported version to 16. Recommend an EL 10 installation for the external DB and use RPM packages.

**Comment by evgeni** [comment]
I completely agree with what @ekohl said above.

**Comment by ShimShtein** [comment]
OK, so I'll put the PG 16 requirement only for the containerized deployment and add an EL10 + RPM recommendation to the external DB procedure.

**Comment by ShimShtein** [comment]
What do you think about this version?
I have added a special case for containerized deployments.

File: guides/common/modules/proc_installing-postgresql-as-an-external-database.adoc [hunk 1]
```diff
@@ -12,10 +12,14 @@ ifdef::foreman-deb[]
 The examples below assume PostgreSQL 14 as used on Ubuntu 22.04.
 If you run {ProjectServer} on Debian 12, you have to adjust the paths of PostgreSQL config and data directories which depend on the PostgreSQL version.
 endif::[]
-ifndef::foreman-deb[]
+ifdef::containerized[]
+{Project} supports PostgreSQL version 16, available in Enterprise Linux 10 repositories.
```
maximiliankolb:
```suggestion
{Project} supports PostgreSQL version 16 which is available in {EL} 10 repositories.
```

We have an attribute for this: https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-base.adoc#L82

guides/common/modules/ref_postgresql-as-an-external-database-considerations.adoc
```diff
@@ -15,7 +15,10 @@ ifdef::foreman-deb[]
 {Project} requires PostgreSQL version 13 or later.
 {Team} recommends that you use the PostgreSQL version that is part of the default operating system repositories.
 endif::[]
-ifndef::foreman-deb[]
+ifdef::containerized[]
+{Project} supports PostgreSQL version 16, available in Enterprise Linux 10 repositories.
```
maximiliankolb:
```suggestion
{Project} supports PostgreSQL version 16 which is available in {EL} 10 repositories.
```

Same as above.

guides/common/modules/proc_installing-postgresql-as-an-external-database.adoc
```diff
@@ -12,10 +12,14 @@ ifdef::foreman-deb[]
...
+endif::[]
+ifndef::foreman-deb,containerized[]
 {Project} supports PostgreSQL version 13.
 endif::[]
 
+
```
maximiliankolb:
```suggestion
```

ekohl:
> No diff compared to the current base

This is because we don't publish the admin guide yet.

guides/common/modules/proc_installing-postgresql-as-an-external-database.adoc
jafiala:
```suggestion
{Project} supports PostgreSQL version 16, which is available in {EL} 10 repositories.
```

guides/common/modules/ref_postgresql-as-an-external-database-considerations.adoc
jafiala: Same as above :)

```suggestion
{Project} supports PostgreSQL version 16, which is available in {EL} 10 repositories.
```

ShimShtein:
Addressed the comments.

maximiliankolb (approved):
style-wise LGTM; I could not check the rendered version (see https://github.com/theforeman/foreman-documentation/pull/5031#issuecomment-4934119054) but I assume it's OK.

jafiala (approved):

ekohl (approved):

PR #5066, opened 2026-07-15
Fix misplaced heading

jsvarova:
Foreman 3.12 document at https://docs.theforeman.org/3.12/Planning_for_Project/index-katello.html#part-Deployment_Planning 
"8. Foreman deployment planning" is empty.

Proposing a fix by adding the "part" declaration and the "book" doctype.
Other options are on the table as well, however, this seems the most elegant. 
Downstream ticket https://redhat.atlassian.net/browse/SAT-47264

### DIFF
--- a/guides/doc-Planning_for_Project/master.adoc
+++ b/guides/doc-Planning_for_Project/master.adoc
@@ -1,6 +1,5 @@
 include::common/attributes.adoc[]
 include::common/header.adoc[]
-:doctype: book
 :context: planning
 
 = {PlanningDocTitle}
@@ -18,7 +17,6 @@ include::common/modules/proc_providing-feedback-on-red-hat-documentation.adoc[le
 endif::[]
 
 [id="Project_Overview_and_Concepts_{context}"]
-[part]
 == {Project} overview and concepts
 
 {ProjectName} is a centralized tool for provisioning, remote management, and monitoring of multiple {EL} deployments.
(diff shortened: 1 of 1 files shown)


guides/doc-Planning_for_Project/master.adoc
```diff
@@ -1,5 +1,6 @@
 include::common/attributes.adoc[]
 include::common/header.adoc[]
+:doctype: book
```
maximiliankolb: I am not sure if this is the way to go. Did you look at the git history of this file? Maybe a cherry-pick of fbc716deb22d3ae37db84aeb64b243661fb4a54b and f63c697e8fe57ba81cb2683f05c7fd620c54dcd7 solves it?

aneta-petrova: It's true that in the past, the Planning guide leverage parts in a similar way that you're suggesting but we dropped them because of some silent build errors, if I remember correctly. So I agree with Maximilian that this is probably not the best way to go.

However, cherry-picking a commit with many other changes might be too much to fix this single leveling issue.

aneta-petrova (commented):
@jsvarova As an alternative solution, how about simply removing the `{Project} deployment planning` heading altogether?

jsvarova:
Hello @aneta-petrova , tested your proposal locally and seems to be most elegant. 
Pushing a new commit.
Thank you!

aneta-petrova (approved):

maximiliankolb (approved):

jsvarova:
Thank you @aneta-petrova and @maximiliankolb for your perspectives and help with this fix!

PR #5072, opened 2026-07-16
document renewal of self-signed and custom server CA for containerized

evgeni:

title

#### Contributor checklists


guides/common/modules/proc_renewing-self-signed-ca-certificates.adoc
```diff
@@ -0,0 +1,37 @@
...
+[role="_abstract"]
+The self-signed Certiticate Authority (CA) on your {ProjectServer} is valid for 20 years by default.
+If you need to renew it because the expiration date is close, follow the below procedure.
+
+.Procedure
+* Renew CA certificates:
++
+[options="nowrap", subs="+quotes,attributes"]
+----
+# {foremanctl} deploy --certificate-renew-ca
```
ShimShtein: This will break the communication with the smart proxies. Do we want to have a more gradual process, like:
1. generate new CA
2. Generate certs for all the proxies using the new CA
3. Add new CA to the current CA bundle
4. Download and install the new CA and certs on proxy servers (using the old certs for download).
5. Once all the proxies are done, remove the old CA from the bundle.

evgeni: It will not. The private key of the CA doesn't change, thus the old and new signatures are still valid.

evgeni: https://github.com/theforeman/foremanctl/pull/671

evgeni: @ekohl I just realized, this probably should be `--certificate-ca-renew` to match other CA related params (like `--certificate-ca-validity-days`) where the "CA" part is "front"

ekohl: That makes sense to me. It will help to group them since IIRC `foremanctl --help` also shows them sorted.

evgeni: https://github.com/theforeman/foremanctl/pull/672

guides/common/modules/proc_renewing-self-signed-ca-certificates.adoc
```diff
@@ -0,0 +1,37 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="renewing-self-signed-ca-certificates"]
+= Renewing self-signed CA certificates
```
evgeni, line 4: did we ever decide on "self-signed" vs "default" terminology?

foremanctl CLI calls them "default" certs

guides/common/modules/proc_renewing-a-custom-server-ca-certificate.adoc
```diff
@@ -0,0 +1,71 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="renewing-custom-server-ca-certificate"]
+= Renewing custom server CA certificate
```
evgeni: Should that be "Renewing *a* custom server CA certificate"?
Given there is only ever one… it sounds weird

guides/common/modules/proc_renewing-self-signed-ca-certificates.adoc
jafiala: Good question, seems the current usage in the docs is mixed, with self-signed certificate, default certificate, and even default self-signed certificate.

My stylistic preference would be to use "default" in headings for brevity, and then use "default self-signed" in the modules.

guides/common/modules/proc_renewing-a-custom-server-ca-certificate.adoc
jafiala: If there's only ever one, it should be "_the_ custom server CA certificate".

But article usage is kind of consistently telegraphic in the current docs, so the current state at least matches the rest.

My preference would be to fix all of them to be grammatically correct and consistent, but that's a lot of work for minimal benefit.

guides/common/modules/con_renewing-certificates.adoc
```diff
@@ -5,8 +5,8 @@
 
 [role="_abstract"]
 ifdef::containerized[]
-You can renew self-signed server and client certificates on {ProjectServer} to prevent them from expiring.
-You can also renew the {customssl} certificate on {ProjectServer} and on {SmartProxyServer}.
+You can renew self-signed server and client certificates as well as the self-signed Certificate Authority on {ProjectServer} to prevent them from expiring.
+You can also renew the custom SSL certificate as well as the custom Certificate Authority on {ProjectServer} and on {SmartProxyServer}.
```
jafiala: I don't think there's reason to capitalize Certificate Authority.

Also, is there a word missing?


```suggestion
You can also renew the custom SSL certificate as well as the custom certificate authority certificate on {ProjectServer} and on {SmartProxyServer}.
```

guides/common/modules/proc_renewing-a-custom-server-ca-certificate.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+[role="_abstract"]
+If you need to update the Certification Authority (CA) certificate that signed your {ProjectServer} and {SmartProxyServers} certificates, add the new CA certificate and use a temporary dual CA certificate file to retain the HTTPS connections to your {ProjectServer} during the renewal.
```
jafiala:
```suggestion
If you need to update the certificate authority (CA) certificate that signed your {ProjectServer} and {SmartProxyServers} certificates, add the new CA certificate and use a temporary dual CA certificate file to retain the HTTPS connections to your {ProjectServer} during the renewal.
```

guides/common/modules/proc_renewing-self-signed-ca-certificates.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+[role="_abstract"]
+The self-signed Certiticate Authority (CA) on your {ProjectServer} is valid for 20 years by default.
```
jafiala:
```suggestion
The self-signed certiticate authority (CA) on your {ProjectServer} is valid for 20 years by default.
```

Maybe missing "certificate" also?

guides/common/modules/proc_renewing-self-signed-ca-certificates.adoc
```diff
@@ -0,0 +1,37 @@
...
+[options="nowrap", subs="+quotes,attributes"]
+----
+# echo | openssl s_client -connect 127.0.0.1:443 2>/dev/null
+----
++
+The output displays the whole SSL connection information.
+Check the `NotBefore` and `NotAfter` dates of CA entry in the "Certificate chain" section, where `NotBefore` is the date when the certificate becomes valid and `NotAfter` is the date when it expires.
+
+.Next steps
+* Refresh the CA certificate to your {SmartProxyServers}.
```
jafiala: Unless this is meant to be different, we should be consistent.

```suggestion
* Refresh the CA certificate on your {SmartProxyServers}.
```

jafiala (commented):
Some minor style comments.

guides/common/modules/proc_renewing-self-signed-ca-certificates.adoc
evgeni: oups, it was not

guides/common/modules/con_renewing-certificates.adoc
evgeni: it was capitalized in the proc I deleted, so I took that over.

jafiala: Yup, I know. We're inconsistent currently, just stopping the spread for now.

guides/common/modules/proc_renewing-a-custom-server-ca-certificate.adoc
jafiala: Again, this wording is pre-existing, but let's not carry the inconsistency forward.

guides/common/modules/con_refreshing-the-self-signed-ca-certificate-on-hosts.adoc
evgeni, line 1: technically, this concept applies to self-signed and custom, but I didn't want to rename that one too

guides/common/modules/proc_renewing-self-signed-ca-certificates.adoc
maximiliankolb: Default is always self-signed, isn't it? And custom might be self signed or signed by a (generally) trusted CA?

evgeni: Yeah, the more correct explanation is:
- by default we manage everything for you (self-signed internally)
- custom(er) gives us files, however they were conceived is not important, as long they match

evgeni: there is also a third (requested, but not implemented, see https://github.com/theforeman/foremanctl/pull/602) option: give us a CA *private* key and we do everything from there for you

guides/common/modules/con_refreshing-the-self-signed-ca-certificate-on-hosts.adoc
jafiala: Hm, we're also inconsistent on refresh/renew. I think using both in text is fine for searchability, but headings and IDs should use just one of those.

renewal sounds better to me.

evgeni: nah, this one actually makes sense:
- renew: deploy a new thing (cert in this case)
- refresh: make some other system use the thing you deployed somewhere else

ekohl: I'm with @evgeni on this.

Renewal is the process of obtaining new SSL certificates.. In the past that typically meant going to your CA and actually pay money, these days it can often be done for free (like Let's Encrypt). If you own the CA itself it is always free. But that's not what's happening here.

Refreshing is not really a very commonly used term with SSL certificates. In my mind it's more updating (if something was done externally) or deploying (when you start using the newly obtained SSL certificates).

guides/common/modules/proc_renewing-a-custom-server-ca-certificate.adoc
```diff
@@ -0,0 +1,71 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="renewing-a-custom-server-ca-certificate"]
+= Renewing a custom server CA certificate
+
+[role="_abstract"]
+If you need to update the certification authority (CA) certificate that signed your {ProjectServer} and {SmartProxyServers} certificates, add the new CA certificate and use a temporary dual CA certificate file to retain the HTTPS connections to your {ProjectServer} during the renewal.
```
evgeni, line 7: This "dual" thing is only really needed if the CA signing the certificate changed in a way that using the new CA certificate does not validate the server certificate signed by the old CA. That happens when you change CAs (or sub-CAs, which commercial providers do every now and then), but usually does not happen when you "just" provide an updated certificate with a different expiration date (which is what would often happen when you have an in-house CA).

The procedure described below works for both cases, as it takes the "safe" path of trusting both certificates for a while, but is actually too much work for the case where only the dates shall change.

Not sure how to present this detail to the reader without writing half-a-page note that will scare everyone :(

ShimShtein: It's a procedure, can we let the user test the new CA + old server cert and create a different procedure for each one of the cases?
Something along the lines of:
1. Check if the old server certs are validated with the new CA: `openssl --cacert /path/to/new/ca.pem`
2. If the check succeeded skip to step 5
3. create dual CA file and upload it to the server
4. update server and proxy certificates
5. upload only the new CA to the server

guides/common/modules/proc_renewing-self-signed-ca-certificates.adoc
```diff
@@ -0,0 +1,37 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="renewing-self-signed-ca-certificates"]
+= Renewing self-signed CA certificates
+
+[role="_abstract"]
+The self-signed certiticate authority (CA) certificate on your {ProjectServer} is valid for 20 years by default.
+If you need to renew it because the expiration date is close, follow the below procedure.
```
ShimShtein, line 8: Is there another use case where the user will want to renew the self-signed CA except the 20 years expiry?

ekohl: Private key leaking is a case I can think of, but our renewal flow keeps the private key so I don't think that would work.

evgeni: Correct. Important: if you change the (internal) CA key, all your clients will need need certs to access content, so "you really do not want to do that"™ in most cases.

guides/common/modules/proc_renewing-a-custom-server-ca-certificate.adoc
evgeni: That could work. @jafiala what do you think?

guides/common/modules/proc_renewing-self-signed-ca-certificates.adoc
ShimShtein: >if you change the (internal) CA key, all your clients will need certs to access content

That would be the full procedure as I described previously, isn't it?

evgeni: No. You need to reregister the hosts. Maybe even reimport the manifest.

ShimShtein: Oh, the fun of Candlepin... :+1:

guides/common/modules/proc_renewing-a-custom-server-ca-certificate.adoc
jafiala: Sorry, missed the ping yesteday.

Yes, I think that's a reasonable approach. But I would use substeps instead.

```
. Check if the old server certs are validated with the new CA: openssl --cacert /path/to/new/ca.pem
. If the check did not succeed, perform the following steps:
.. create dual CA file and upload it to the server
.. update server and proxy certificates
. upload only the new CA to the server
```

jafiala: We can have something more helpful instead of "perform the following steps" if there are meaningful words for the the two following steps combined.

evgeni: Good idea. I can build on this, thanks!

evgeni: Updated, please have a look

jafiala: I like it, this works really well.

guides/common/modules/con_refreshing-the-self-signed-ca-certificate-on-hosts.adoc
jafiala: Got it, thanks for the explanation.

jafiala (approved):

guides/common/modules/proc_renewing-a-custom-server-ca-certificate.adoc
```diff
@@ -0,0 +1,80 @@
...
+# cat _/root/{project-context}_cert/old_ca_cert_bundle.pem_ _/root/{project-context}_cert/new_ca_cert_bundle.pem_ > _/root/{project-context}_cert/ca_cert_bundle.pem_
+----
+.. Renew the certificates on {ProjectServer}.
++
+[options="nowrap" subs="+quotes,attributes"]
+ifdef::containerized[]
+----
+# {foremanctl} deploy \
+--certificate-source custom_server \
+--certificate-server-certificate "_/root/{project-context}_cert/{project-context}_cert.pem_" \
```
amolpati30, line 30: Is `{project-context}` supposed to be `orcharhino`, or does it automatically pick the value based on whether it's the upstream or downstream project?

evgeni: It's defined as an attribute:
```
% git grep :project-context:
guides/common/attributes-base.adoc::project-context: foreman
guides/common/attributes-orcharhino.adoc::project-context: {Project}
guides/common/attributes-satellite.adoc::project-context: satellite
```
So it ends up as the right value depending on the build target.

ShimShtein (approved):

ShimShtein:
Looks good from technical perspective

PR #4835, opened 2026-05-12
Move FreeIPA realm section to Integrating Provisioning guide

AkshayGadhaveRH:
The `Configuring Project to manage the lifecycle of a host registered to a FreeIPA realm` section is better suited for the `Integrating provisioning infrastructure services` guide.

JIRA:
https://redhat.atlassian.net/browse/SAT-36280

#### Contributor checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [ ] Foreman 3.18/Katello 4.20 (Satellite 6.19)
* [ ] Foreman 3.17/Katello 4.19
* [ ] Foreman 3.16/Katello 4.18 (Satellite 6.18; orcharhino 7.6, 7.7, and 7.8)
* [ ] Foreman 3.15/Katello 4.17
* [ ] Foreman 3.14/Katello 4.16 (Satellite 6.17; orcharhino 7.4; orcharhino 7.5)
* [ ] Foreman 3.13/Katello 4.15 (EL9 only)
* [ ] Foreman 3.12/Katello 4.14 (Satellite 6.16; orcharhino 7.2 on EL9 only; orcharhino 7.3)
* We do not accept PRs for Foreman older than 3.12.

PR #5076, opened 2026-07-16
Improve an ID for SEO

Lennonka:

Differentiating ID for upstream/downstream using attributes

SEO

- I noticed this in #4835
- This ID is linked from the web UI and it will need updating in code
- Which is a great opoprtunity to improve it first
- I'm not using the existing `FreeIPA-context` attribute on purpose because it uses old Snake_Case capitalization
- The ID is not referenced anywhere (I checked), therefore, no other updates needed

#### Contributor checklists

### DIFF
--- a/guides/common/attributes-satellite.adoc
+++ b/guides/common/attributes-satellite.adoc
@@ -85,7 +85,7 @@
 :foreman-maintain: satellite-maintain
 :FreeIPA: Identity{nbsp}Management
 :FreeIPA-context: Identity_Management
-:FreeIPA-id: idm
+:FreeIPA-id: identity-management
 :freeipaserver-example-com: idm-server.example.com
 :hammer-smart-proxy: hammer capsule
 :installer-log-file: /var/log/foreman-installer/satellite.log
--- a/guides/common/attributes-satellite.adoc
+++ b/guides/common/attributes-satellite.adoc
@@ -85,6 +85,7 @@
 :foreman-maintain: satellite-maintain
 :FreeIPA: Identity{nbsp}Management
 :FreeIPA-context: Identity_Management
+:FreeIPA-id: idm
 :freeipaserver-example-com: idm-server.example.com
 :hammer-smart-proxy: hammer capsule
 :installer-log-file: /var/log/foreman-installer/satellite.log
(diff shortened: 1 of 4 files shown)


Lennonka:
The Filename check is failing because the API call fails. Maybe it will work tomorrow.

maximiliankolb (approved):
LGTM. 👍
I am OK not using the existing attribute `FreeIPA-context`; I will look into dropping it if possible as a follow-up.

guides/common/attributes-satellite.adoc
```diff
@@ -85,6 +85,7 @@
 :foreman-maintain: satellite-maintain
 :FreeIPA: Identity{nbsp}Management
 :FreeIPA-context: Identity_Management
+:FreeIPA-id: idm
```
aneta-petrova: Why the two attributes? In `attributes-base.adoc` and `check-filename-id-heading-match.sh` as well. Is there a reason not to use the existing "context" one? Or, if the preference is to use an `*-id` attribute (which I'd understand), why not drop the existing "context" one?

I'd much prefer a PR that updates and improves the existing attribute list than one that adds a new one without addressing the existing ones :)

Lennonka: Because it can break other links where the `-context` attribute appears and they might need additional cleanup, which is out of scope for this PR.
\+ Max ^^ says he'll try to get rid of it.

The reason for not using it is in the PR description.

aneta-petrova: Does the `idm` part really improve SEO, though? That's the benefit of the existing ID attribute -- it spells out the name of the product.

Lennonka:
```suggestion
:FreeIPA-context: Identity_Management
:FreeIPA-id: identity-management
```

guides/common/attributes-satellite.adoc
```diff
@@ -85,6 +85,7 @@
 :foreman-maintain: satellite-maintain
 :FreeIPA: Identity{nbsp}Management
 :FreeIPA-context: Identity_Management
+:FreeIPA-id: identity-management
```
Lennonka, line 88: @aneta-petrova Happy now? :)

aneta-petrova: Are you happy as the author? :) Does this make sense? If the goal is to improve SEO, then using the full product name instead of an acronym does make sense to me,

aneta-petrova (approved):

guides/common/attributes-satellite.adoc
Lennonka: I am completely happy. Thanks.

PR #5097, opened 2026-07-23
Add containerized proxy installation

jafiala:
^^

https://redhat.atlassian.net/browse/SAT-47538

#### Contributor checklists


guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -30,3 +31,49 @@ ifdef::katello[]
...
++
+ifdef::satellite[]
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} deploy-proxy \
+--flavor capsule \
+--certificate-bundle /root/_{smartproxy-example-com}_.tar.gz \
+--foreman-fqdn _{foreman-example-com}_
+----
+endif::satellite[]
```
jafiala: This does not exist yet, and we might treat it differently - for example an attribute in place of foreman-proxy-content and capsule.

jafiala:
Please review particularly the pre-installation sections in the preview. They might need some adjustments but are not (yet) visible in the files changed.

guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -6,7 +6,17 @@
 [role="_abstract"]
 With {SmartProxy} packages installed, you can proceed to install the {SmartProxyServer}.
 
+ifdef::containerized[]
+.Prerequisites
+* {ProjectServer} is deployed in a containerized configuration.
+For more information, see xref:{InstallingServerDocURL}deploy-{project-context}.-server-configuration-with-the-default-database[].
```
maximiliankolb:
```suggestion
For more information, see {InstallingServerDocURL}deploy-{project-context}.-server-configuration-with-the-default-database[].
```

guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -30,3 +40,49 @@ ifdef::katello[]
 * To install {SmartProxyServer} with content, refer to xref:configuring-{smart-proxy-context}-server-with-ssl-certificates[].
 Running `{certs-generate}` is a required prerequisite to installing {SmartProxyServer} with content.
 endif::[]
+endif::[]
+ifdef::containerized[]
+. On your {ProjectServer}, generate a certificate bundle for the{SmartProxyServer}:
```
maximiliankolb:
```suggestion
. On your {ProjectServer}, generate a certificate bundle for {SmartProxyServer}:
```

guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -30,3 +40,49 @@ ifdef::katello[]
...
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} certificate-bundle _{smartproxy-example-com}_
+----
++
+The certificate bundle is stored at `/var/lib/foremanctl/certs/bundles/_{smartproxy-example-com}_.tar.gz`.
```
maximiliankolb: "stored at" could mean a directory. Because you reference the path in line 57, maybe dropp line 51 and 52?

guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -30,3 +40,49 @@ ifdef::katello[]
...
+. From your {ProjectServer}, copy the certificate bundle to the{SmartProxyServer}:
```
maximiliankolb:
```suggestion
. On your {ProjectServer}, copy the certificate bundle to {SmartProxyServer}:
```

guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -30,3 +40,49 @@ ifdef::katello[]
...
+# scp /var/lib/foremanctl/certs/bundles/_{smartproxy-example-com}_.tar.gz \
+root@_{smartproxy-example-com}_:/root/_{smartproxy-example-com}_.tar.gz
+----
++
+[IMPORTANT]
+====
+The certificate bundle contains TLS private keys and OAuth credentials.
+Treat the bundle as highly sensitive and delete any copies after deployment.
+====
+. On the {SmartProxyServer}, deploy the configuration:
```
maximiliankolb:
```suggestion
. On your {SmartProxyServer}, deploy the configuration:
```

guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -30,3 +40,49 @@ ifdef::katello[]
...
++
+ifdef::satellite[]
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} deploy-proxy \
+--flavor capsule \
```
maximiliankolb: If you want this for Satellite, then maybe introduce an attribute. I am unsure about https://github.com/theforeman/foreman-documentation/pull/5097/changes#r3637946337: Do you mean the implementation for upstream does not exist yet?

guides/common/modules/proc_installing-smart-proxy-server.adoc
jafiala: I think this needs the definite article, because this is a specific cert bundle for one specific proxy server.

```suggestion
. On your {ProjectServer}, generate a certificate bundle for the {SmartProxyServer}:
```

guides/common/modules/proc_installing-smart-proxy-server.adoc
jafiala: I actually changed this to "From". To me, it makes more sense to say "Copy from A to B" than "Copy on A to B".

guides/common/modules/proc_installing-smart-proxy-server.adoc
jafiala: I mean the implementation for _downstream_ does not exist yet.

I think I'll just drop it for now and revisit in a later PR.

guides/common/modules/proc_installing-smart-proxy-server.adoc
jafiala: Because a user might have more proxies, I think the "the" is more specific - we need the proxy they are installing now, not just any of their proxies.

guides/common/modules/proc_installing-smart-proxy-server.adoc
maximiliankolb: But that's not how your scp example works. If you changed it, then yes:

````
$ scp root@_{foreman-example-com}_:/var/lib/foremanctl/certs/bundles/_{smartproxy-example-com}_.tar.gz root@_{smartproxy-example-com}_:/root/_{smartproxy-example-com}_.tar.gz
````

In your current example, you _must_ run `scp` on Foreman Server.

**Review by maximiliankolb — COMMENTED** [review]
I don't like "the Foreman server" or "the Smart Proxy server" (in general), but I won't block your PR.

File: guides/common/modules/proc_installing-smart-proxy-server.adoc [hunk 8]
```diff
@@ -6,7 +6,17 @@
 [role="_abstract"]
 With {SmartProxy} packages installed, you can proceed to install the {SmartProxyServer}.
 
+ifdef::containerized[]
+.Prerequisites
+* {ProjectServer} is deployed in a containerized configuration.
+For more information, see {InstallingServerDocURL}deploy-{project-context}.-server-configuration-with-the-default-database[].
```
aneta-petrova: A Satellite server will always be containerized in this version.
```suggestion
ifndef::satellite[]
* {ProjectServer} is deployed in a containerized configuration.
For more information, see {InstallingServerDocURL}deploy-{project-context}.-server-configuration-with-the-default-database[].
endif::[]
```
But also, the URL seems to be not quite right -- the `.` in the path.

jafiala: Right, but users don't have all servers always on the latest versions, correct? I think cross-version architectures (not sure the right word, but different foreman versions of server/proxy/host) might be possible?

Even for Satellite, we don't want customers attempting to connect a non-containerized Sat server to containerized proxy.

aneta-petrova: From the Satellite preview: `The version of Capsule must match the version of Satellite installed. For example, the Capsule version 6.20 cannot be registered with the Satellite version 6.19.` https://theforeman-foreman-documentation-preview-pr-5097.surge.sh/nightly/Installing_Proxy/index-satellite.html#system-requirements_capsule

For upgrades, scenarios where the Proxy is one version behind the server are possible (because you have to start upgrading somewhere) but that won't be the case for containerized Foreman because that will always need to be a fresh installation.

guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -6,7 +6,17 @@
...
+* {SmartProxyServer} packages are installed.
+For more information, see xref:installing-{smart-proxy-context}-packages[].
+* You can connect from {ProjectServer} to {SmartProxyServer} by using SSH.
```
evgeni: I don't think this is (or should be) a prerequisite.
Sure, it's the easiest way to get the bundle from Server to Proxy, but it's not a requirement as long as the bundle ends up there somehow.

guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -30,3 +40,36 @@ ifdef::katello[]
...
+====
+The certificate bundle contains TLS private keys and OAuth credentials.
+Treat the bundle as highly sensitive and delete any copies after deployment.
+====
+. On the {SmartProxyServer}, deploy the configuration:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} deploy-proxy \
+--flavor foreman-proxy-content \
```
evgeni:
```suggestion
--flavor {foremanctl-proxy-scenario} \
```

and then define the attribute to `foreman-proxy-content` upstream (and orcharino, afaik) and `capsule` for Satellite

guides/common/modules/proc_installing-smart-proxy-server.adoc
evgeni: Should this have a title?

guides/doc-Installing_Proxy/master.adoc
```diff
@@ -28,6 +23,7 @@ include::common/assembly_preparing-your-environment-for-smart-proxy-installation
 // Installing {SmartProxyServer}
 include::common/assembly_installing-smart-proxy-server.adoc[leveloffset=+1]
```
evgeni, line 24: This file also includes "Configuring Smart Proxy server with SSL certificates" and "Resetting custom SSL certificate to default self-signed certificate on Smart Proxy server"
The former definitely needs updates, as it uses foreman-installer.
The latter looks like it needs re-verification, as I have doubts it works for either installation method right now. (probably not in scope for this PR)

guides/common/modules/proc_installing-smart-proxy-server.adoc
jafiala: Added in https://github.com/theforeman/foreman-documentation/pull/5097/commits/de1642638baea4a736ee6fd6f57dfc2b75f6d03f

guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -30,3 +37,36 @@ ifdef::katello[]
 * To install {SmartProxyServer} with content, refer to xref:configuring-{smart-proxy-context}-server-with-ssl-certificates[].
 Running `{certs-generate}` is a required prerequisite to installing {SmartProxyServer} with content.
 endif::[]
+endif::[]
+ifdef::containerized[]
+. On your {ProjectServer}, generate a certificate bundle for the {SmartProxyServer}:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} certificate-bundle _{smartproxy-example-com}_
```
arvind4501: Should we tell user, that where the generated bundle will be stored on server? something like https://github.com/theforeman/foremanctl/blob/master/docs/developer/deployment.md#proxy as a note

evgeni: The path is part of the next step, isn't that sufficient?

arvind4501: yes seems Ok, i was a bit more verbose, thanks

guides/common/modules/proc_installing-smart-proxy-server.adoc
ekohl, line 1: Given how many content is hidden behind `ifdef` I wonder, at which point is it easier to have a separate module for containerized vs non-containerized?

guides/common/attributes-containerized.adoc
```diff
@@ -7,8 +7,10 @@
 // - attributes-satellite.adoc (always containerized)
 
 // Containerized deployment overrides
+:foremanctl-proxy-scenario: foreman-proxy-content
 :installer-log-file: /var/log/foremanctl/foremanctl.log
 :project-installer-package: foremanctl
+:smartproxy-installer-package: foremanctl
```
ekohl: I don't think we intend to create a separate package for content proxies but I realize this is to reuse an existing module Perhaps still reuse `project-installer-package` to indicate we don't intend to keep it separate?

```suggestion
:smartproxy-installer-package: {project-installer-package}
```

guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -30,3 +37,36 @@ ifdef::katello[]
...
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} deploy-proxy \
+--flavor {foremanctl-proxy-scenario} \
+--certificate-bundle /root/_{smartproxy-example-com}_.tar.gz \
+--foreman-fqdn _{foreman-example-com}_
+----
++
+The script displays its progress and writes logs to `{installer-log-file}`.
```
ekohl: AFAIK this original line dates back to when `foreman-installer` displayed a progress bar without detailed information. That was later changed to display detailed information as it progressed (in addition to logging) and `foremanctl` does the same. While it's true that users can look it up, I question if this line is still needed.

guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -30,3 +37,36 @@ ifdef::katello[]
...
+====
+The certificate bundle contains TLS private keys and OAuth credentials.
+Treat the bundle as highly sensitive and delete any copies after deployment.
+====
+. On the {SmartProxyServer}, deploy the configuration:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} deploy-proxy \
+--flavor {foremanctl-proxy-scenario} \
```
ekohl: In upstream this line isn't needed since it's the default. For downstream I wonder if we could make it default to `capsule` as well.

guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -30,3 +37,36 @@ ifdef::katello[]
...
+--certificate-bundle /root/_{smartproxy-example-com}_.tar.gz \
+--foreman-fqdn _{foreman-example-com}_
```
ekohl: Unrelated to this, but I'll dump my brainfart here: I debated writing this value into the bundle too so the command becomes `{foremanctl} deploy-proxy --certificate-bundle /root/_{smartproxy-example-com}_.tar.gz`.

ekohl (commented):
In the preview I see [2.1. Registering Capsule Server](https://theforeman-foreman-documentation-preview-pr-5097.surge.sh/nightly/Installing_Proxy/index-containerized-katello.html#registering-smart-proxy-server) and upstream that really isn't needed because step [2.2. Configuring repositories](https://theforeman-foreman-documentation-preview-pr-5097.surge.sh/nightly/Installing_Proxy/index-containerized-katello.html#configuring-repositories_smart-proxy) pulls the content directly from the internet anyway.

I'm always a fan of minimal procedures; should we drop this and only add it later if we feel it's needed?

guides/common/modules/proc_installing-smart-proxy-server.adoc
evgeni: This will "break" if people want to use an alias or so.

ekohl: Fair point. Let's leave it for now.

guides/doc-Installing_Proxy/master.adoc
jafiala: Since these are katello/satellite only, I didn't want to include them in this PR but in a follow-up downstreamization PR.

So I will exclude both from the containerized builds here, thanks for bringing this up.

Will keep this open in case I am wrong about that.

guides/common/modules/proc_installing-smart-proxy-server.adoc
arvind4501: thats true as it would deploy-proxy without passing `--flavor`, but then `foremanctl features` and `foremanctl health` etc commands will be in broken state as they need flavor specifc vars https://github.com/theforeman/foremanctl/blob/master/src/playbooks/features/features.yaml#L12

ekohl: In the old installer we stored the scenario as soon as we could. IMHO any `--option` is optional and not mandatory. Should we see it as a bug that it doesn't get stored?

guides/common/attributes-katello.adoc
```diff
@@ -8,4 +8,4 @@
 :project-minimum-memory: 20 GB
 :smart-proxy-minimum-memory: 12 GB
 :smartproxy_port: 9090
-:smartproxy-installer-package: foreman-proxy-content
+:smartproxy-installer-package: {project-installer-package}
```
ekohl: This is not what I meant in https://github.com/theforeman/foreman-documentation/pull/5097/changes#r3657027602. The Katello attributes shouldn't change.

guides/common/modules/proc_installing-smart-proxy-server.adoc
jafiala: @ekohl It would be easier to write and read the code, but we'd need a different filename and ID.

Given that we already have `installing-a-smart-proxy-server` vs `installing-smart-proxy-server` (though one is con and the other proc), I thought using the existing file was the better choice.

Also we don't want to have "containerized" in file names or headings.

ekohl: > Also we don't want to have "containerized" in file names or headings.

Why not? If the procedures are sufficiently different then I think it's actually very useful to have it in filenames. That's what snippets are for IMHO.

Edit: IIRC we also differentiate setting up repositories between upstream EL, upstream Debian, and downstream. In the end they have the same result, but the implementations are wildly different that combining them in a single file hurts maintainability.

guides/common/modules/proc_installing-smart-proxy-server.adoc
arvind4501: I think defaults variables are not stored in `parameters.yaml` is by design. but we should improve commands like features and health to dynamically get flavor value somehow, or maybe store only flavor var be it default

guides/common/attributes-katello.adoc
jafiala: Gotcha, [088ecde](https://github.com/theforeman/foreman-documentation/pull/5097/commits/088ecde8208f0964f50101ea8cbf3fc6289b54d4)

ekohl:
Could you separate out the removal of the registration step in upstream to its own PR? Originally I had hoped that https://github.com/theforeman/foreman-documentation/pull/5030 would already take care of that. Reducing the scope of this PR will help with reviewing.

guides/common/modules/proc_installing-smart-proxy-server.adoc
jafiala: @ekohl What is the outcome here? Shall we keep --flavor here for now, and remove it when the improvement hits? We'd need to track that as part of that improvement. Not sure where that's tracked currently.

guides/common/modules/proc_installing-smart-proxy-server.adoc
jafiala: Hm, I don't remember exactly why, but it's a practice I was taught by the rest of the doc team and follow.

Looking at guides/common/modules/proc_configuring-repositories.adoc, I don't think the structure is more readable than the current ifdef structure. (putting aside the snippets within snippets approach for repos that we don't need) .

If you don't mind, I'd go with the current structure for now so we can move forward.

guides/common/modules/proc_installing-smart-proxy-server.adoc
ekohl: I've created https://github.com/theforeman/foremanctl/issues/708 to track it. For now we can keep this as a known limitation.

jafiala:
@ekohl Good call. Opened https://github.com/theforeman/foreman-documentation/pull/5115 with just the registration removal.

guides/common/modules/proc_installing-smart-proxy-server.adoc
```diff
@@ -30,3 +37,34 @@ ifdef::katello[]
 * To install {SmartProxyServer} with content, refer to xref:configuring-{smart-proxy-context}-server-with-ssl-certificates[].
 Running `{certs-generate}` is a required prerequisite to installing {SmartProxyServer} with content.
 endif::[]
+endif::[]
+ifdef::containerized[]
+. On your {ProjectServer}, generate a certificate bundle for the {SmartProxyServer}:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} certificate-bundle _{smartproxy-example-com}_
```
arvind4501:
```suggestion
# {foremanctl} auth-bundle _{smartproxy-example-com}_
```
since https://github.com/theforeman/foremanctl/pull/698 is merged

guides/doc-Installing_Proxy/master.adoc
evgeni: "Configuring Smart Proxy server with SSL certificates" is something that is expected to work with containers too.

arvind4501:
https://theforeman-foreman-documentation-preview-pr-5097.surge.sh/nightly/Installing_Proxy/index-containerized-katello.html#configuring-repositories_smart-proxy here i see we are pulling foremanctl from our nighty repos but since we have not release any version after (2.3.0) which contains any proxy stuff, should we tempraly also enable 
```
dnf copr enable @theforeman/develop rhel-9-x86_64
```
as this contains latest master builds?

guides/common/modules/proc_installing-smart-proxy-server.adoc
jafiala: Should we also start calling it "authentication bundle" or do we want to keep the terminology consistent with RPM?

arvind4501: "authentication bundle" sounds reasonable to me

guides/doc-Installing_Proxy/master.adoc
jafiala: @evgeni Okay. Custom certs I can see. But could you help me understand how is `proc_installing-smart-proxy-server.adoc` different from what we currently have in `proc_installing-smart-proxy-server.adoc`? I thought that's exactly what we are doing, just in containers and with simpler steps.

EDIT: Is that for post-installation tasks, like when the cert changes?

EDIT EDIT: If that's the case, how is that actually different from the installation proc?

You still need to generate bundle, copy to proxy, then `foremanctl deploy-proxy`, no?

evgeni: Yes, cert changes look like new deployments.

> But could you help me understand how is proc_installing-smart-proxy-server.adoc different from what we currently have in proc_installing-smart-proxy-server.adoc

they_are_the_same_picture.jpg

did you want to link something else?

jafiala: Yes I did. I meant `proc_configuring-smart-proxy-server-with-a-default-ssl-certificate.adoc` (in its not yet existent containerized version).

All I can see is `--certificate-renew`, which probably doesn't warrant an entire new procedure, we could just add something like "to regenerate certificates for an existing proxy, run this with `--certificate-renew`."

evgeni: Oooooh, I think I understand!

Look at the [current structure for non-containers](https://docs.theforeman.org/nightly/Installing_Proxy/index-katello.html#installing-smart-proxy-packages):
image

2.3. install proxy packages: dnf yada yada
2.4. install proxy: this is not a procedure, just do the next step
2.5. configure proxy with certs: do the actual work

2.4 is rather pointless IMHO, as the only instruction is "go, follow instruction in next step"

But now in this PR you added the "do certs things" to 2.4, instead of doing an "if containers/else" in 2.5.

So if you just move things around (and maybe accidentally drop 2.4), we're good :)

evgeni: Oooooooh²: proc_installing-smart-proxy-server.adoc (2.4) has actual content for foreman-deb and foreman-el builds, as those don't use Katello's certs tooling like everyone else. That makes a bit of sense.

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
```diff
@@ -1,13 +1,16 @@
 :_mod-docs-content-type: PROCEDURE
+:containerized:
```
maximiliankolb:
```suggestion
```

Please don't.

jafiala: Noticed it just a little afterwards, sorry!

guides/doc-Installing_Proxy/master.adoc
jafiala: @evgeni Could you re-review please? I think everything should be in place now.

guides/common/modules/proc_configuring-smart-proxy-server-with-a-default-ssl-certificate.adoc
```diff
@@ -23,6 +23,39 @@ For more information, see xref:opening-required-ports[].
...
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} deploy-proxy \
+--flavor {foremanctl-proxy-scenario} \
+--auth-bundle /root/_{smartproxy-example-com}_.tar.gz \
+--foreman-fqdn _{foreman-example-com}_
+----
++
+To regenerate certificates for an existing {SmartProxyServer}, add the `--certificate-renew` option to the `{foremanctl} deploy-proxy` command.
```
evgeni: No, to the `auth-bundle` command.

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
```diff
@@ -24,8 +26,58 @@ endif::[]
...
+For more information, see xref:installing-{smart-proxy-context}-packages[].
+endif::[]
 
 .Procedure
+ifdef::containerized[]
+. On your {ProjectServer}, generate an authentication bundle for the {SmartProxyServer}:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} auth-bundle _{smartproxy-example-com}_
```
evgeni:
```suggestion
# {foremanctl} auth-bundle --certificate-source custom_server --certificate-server-key proxy.key --certificate-server-certificate proxy.crt _{smartproxy-example-com}_
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
```diff
@@ -24,8 +26,58 @@ endif::[]
...
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} deploy-proxy \
+--flavor {foremanctl-proxy-scenario} \
+--auth-bundle /root/_{smartproxy-example-com}_.tar.gz \
+--foreman-fqdn _{foreman-example-com}_
+--certificate-source custom_server \
+--certificate-server-certificate "_/root/{project-context}_cert/{project-context}_cert.pem_" \
+--certificate-server-key "_/root/{project-context}_cert/{project-context}_cert_key.pem_" \
+--certificate-server-ca-certificate "_/root/{project-context}_cert/ca_cert_bundle.pem_"
```
evgeni:
```suggestion
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
```diff
@@ -24,8 +26,58 @@ endif::[]
...
+----
++
+--
+The options used in the command include the following:
+
+`--certificate-source custom_server`:: Sets the certificate source to custom server certificates provided by the user.
+`--certificate-server-certificate`:: Path to {ProjectServer} certificate file that is signed by a Certificate Authority.
+`--certificate-server-key`:: Path to the private key for the {ProjectServer} certificate.
+`--certificate-server-ca-certificate`:: Path to the Certificate Authority bundle.
```
evgeni: we don't accept `ca-certificate`, it has to be the same as for the main server.

the rest you must place at `auth-bundle` time

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
```diff
@@ -24,8 +26,58 @@ endif::[]
...
+
+To regenerate certificates for an existing {SmartProxyServer}, add the `--certificate-renew` option to the `{foremanctl} deploy-proxy` command.
```
evgeni:
```suggestion
To regenerate the bundle for an existing {SmartProxyServer}, add the `--certificate-renew` option to the `{foremanctl} auth-bundle` command.
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
```diff
@@ -4,10 +4,12 @@
 = Deploying a custom SSL certificate to {SmartProxyServer}
 
 [role="_abstract"]
-Use this procedure to configure your {SmartProxyServer} with a custom SSL certificate signed by a Certificate Authority.
+If your {ProjectServer} is configured with a custom certificate, you must configure your {SmartProxyServer} with a custom SSL certificate signed by a Certificate Authority.
```
evgeni:
```suggestion
If your {ProjectServer} is configured with a custom certificate, you must configure your {SmartProxyServer} with a custom SSL certificate signed by the same Certificate Authority.
```

IMHO we can be more explicit here

guides/doc-Installing_Proxy/master.adoc
evgeni: Yepp, posted a few comments on some details, but the flow is now correct! Thanks!

guides/common/modules/proc_configuring-smart-proxy-server-with-a-default-ssl-certificate.adoc
jafiala:
```suggestion
To regenerate certificates for an existing {SmartProxyServer}, add the `--certificate-renew` option to the `{foremanctl} auth-bundle` command.
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
jafiala: Moved the options in [90ad11e](https://github.com/theforeman/foreman-documentation/pull/5097/commits/90ad11e8510f0b3887e13fb270db9aba6680c85b).

Just wanted to double check - the replaceable values can/should stay the same, or did you mean to change them as well?

evgeni: the replaceable values were fine, I am just a lazy bum at typing ;)

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
```diff
@@ -24,8 +26,56 @@ endif::[]
...
+--certificate-server-certificate "_/root/{project-context}_cert/{project-context}_cert.pem_" \
+--certificate-server-key "_/root/{project-context}_cert/{project-context}_cert_key.pem_" \
+----
++
+--
+The options used in the command include the following:
+
+`--certificate-source custom_server`:: Sets the certificate source to custom server certificates provided by the user.
+`--certificate-server-certificate`:: Path to {ProjectServer} certificate file that is signed by a Certificate Authority.
+`--certificate-server-key`:: Path to the private key for the {ProjectServer} certificate.
```
evgeni:
```suggestion
`--certificate-server-certificate`:: Path to {SmartProxyServer} certificate file that is signed by a Certificate Authority.
`--certificate-server-key`:: Path to the private key for the {SmartProxyServer} certificate.
```

Those are the ones you created in "Creating a custom SSL certificate for Smart Proxy server"

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
```diff
@@ -24,8 +26,56 @@ endif::[]
...
 .Procedure
+ifdef::containerized[]
+. On your {ProjectServer}, generate an authentication bundle for the {SmartProxyServer}:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} auth-bundle _{smartproxy-example-com}_
+--certificate-source custom_server \
+--certificate-server-certificate "_/root/{project-context}_cert/{project-context}_cert.pem_" \
+--certificate-server-key "_/root/{project-context}_cert/{project-context}_cert_key.pem_" \
```
evgeni: I would have expected `project-context` to resolve to `smart-proxy` here, but the rendering shows `foreman` -- that's wrong

evgeni: Ooooh, just grepped around, the others are using `{cert-name}` which uses different contextes:
```
% git grep :cert-name:
guides/common/assembly_configuring-satellite-custom-server-certificate.adoc::cert-name: {project-context}
guides/common/assembly_configuring-smart-proxy-servers-with-custom-ssl-certificates-for-load-balancing-with-puppet.adoc::cert-name: {smart-proxy-context}
guides/common/assembly_configuring-smart-proxy-servers-with-custom-ssl-certificates-for-load-balancing-without-puppet.adoc::cert-name: {smart-proxy-context}
guides/common/assembly_installing-smart-proxy-server.adoc::cert-name: {smart-proxy-context}
```

evgeni: Or directly `smart-proxy-context` as the context here is clearly the proxy

guides/common/modules/proc_configuring-smart-proxy-server-with-a-default-ssl-certificate.adoc
```diff
@@ -23,6 +23,39 @@ For more information, see xref:opening-required-ports[].
...
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} deploy-proxy \
+--flavor {foremanctl-proxy-scenario} \
+--auth-bundle /root/_{smartproxy-example-com}_.tar.gz \
+--foreman-fqdn _{foreman-example-com}_
+----
++
+To regenerate certificates for an existing {SmartProxyServer}, add the `--certificate-renew` option to the `{foremanctl} auth-bundle` command.
```
evgeni: I'd move this to the step where you call `auth-bundle` instead of the very end?

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
jafiala: I don't think there's a polite way of saying "that's what I thought", so I'll just leave it at that :}

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
jafiala:
```suggestion
--certificate-server-certificate "_/root/{project-context}_cert/{project-context}_cert.pem_" \
--certificate-server-key "_/root/{project-context}_cert/{smart-proxy-context}_cert_key.pem_" \
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
evgeni: permission to be honest and impolite granted :)

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
```diff
@@ -24,8 +26,56 @@ endif::[]
...
 .Procedure
+ifdef::containerized[]
+. On your {ProjectServer}, generate an authentication bundle for the {SmartProxyServer}:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} auth-bundle _{smartproxy-example-com}_
+--certificate-source custom_server \
+--certificate-server-certificate "_/root/{project-context}_cert/{project-context}_cert.pem_" \
+--certificate-server-key "_/root/{project-context}_cert/{smart-proxy-context}_cert_key.pem_" \
```
evgeni:
```suggestion
--certificate-server-certificate "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem_" \
--certificate-server-key "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert_key.pem_" \
```

much more smart needed

guides/common/assembly_installing-smart-proxy-server.adoc
```diff
@@ -27,8 +27,11 @@ endif::[]
 include::modules/proc_installing-smart-proxy-packages.adoc[leveloffset=+1]
 
 ifdef::foreman-el,foreman-deb,katello[]
+ifndef::containerized[]
 include::modules/proc_installing-smart-proxy-server.adoc[leveloffset=+1]
 endif::[]
+endif::[]
+
```
maximiliankolb, line 34:
```suggestion
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
```diff
@@ -24,8 +26,56 @@ endif::[]
...
+For more information, see xref:installing-{smart-proxy-context}-packages[].
+endif::[]
 
 .Procedure
+ifdef::containerized[]
+. On your {ProjectServer}, generate an authentication bundle for the {SmartProxyServer}:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} auth-bundle _{smartproxy-example-com}_
```
maximiliankolb:
```suggestion
# {foremanctl} auth-bundle _{smartproxy-example-com}_ \
```

guides/common/modules/proc_deploying-a-custom-ssl-certificate-to-smart-proxy-server.adoc
```diff
@@ -24,8 +26,56 @@ endif::[]
...
+--certificate-source custom_server \
+--certificate-server-certificate "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert.pem_" \
+--certificate-server-key "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert_key.pem_" \
```
maximiliankolb:
```suggestion
--certificate-server-key "_/root/{smart-proxy-context}_cert/{smart-proxy-context}_cert_key.pem_"
```

jafiala:
> https://theforeman-foreman-documentation-preview-pr-5097.surge.sh/nightly/Installing_Proxy/index-containerized-katello.html#configuring-repositories_smart-proxy here i see we are pulling foremanctl from our nighty repos but since we have not release any version after (2.3.0) which contains any proxy stuff, should we tempraly also enable
> 
> ```
> dnf copr enable @theforeman/develop rhel-9-x86_64
> ```
> 
> as this contains latest master builds?

I'm not sure, can/should we do that? @evgeni @ekohl ?

guides/common/modules/proc_configuring-smart-proxy-server-with-a-default-ssl-certificate.adoc
```diff
@@ -23,6 +23,39 @@ For more information, see xref:opening-required-ports[].
...
 .Procedure
+ifdef::containerized[]
+. On your {ProjectServer}, generate an authentication bundle for the {SmartProxyServer}:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} auth-bundle _{smartproxy-example-com}_
+----
++
+To regenerate certificates for an existing {SmartProxyServer}, add the `--certificate-renew` option to the `{foremanctl} auth-bundle` command.
```
arvind4501:
```suggestion
To regenerate the bundle for an existing {SmartProxyServer}, add the `--certificate-renew` option to the `{foremanctl} auth-bundle` command.
```
to match https://github.com/theforeman/foreman-documentation/pull/5097/changes#diff-559b348c8d71b62118321274424b32aa88c644e8131ded3b78825eed44dbfcd8R54

evgeni:
> > https://theforeman-foreman-documentation-preview-pr-5097.surge.sh/nightly/Installing_Proxy/index-containerized-katello.html#configuring-repositories_smart-proxy here i see we are pulling foremanctl from our nighty repos but since we have not release any version after (2.3.0) which contains any proxy stuff, should we tempraly also enable
> > ```
> > dnf copr enable @theforeman/develop rhel-9-x86_64
> > ```
> > as this contains latest master builds?
> 
> I'm not sure, can/should we do that? @evgeni @ekohl ?

We dropped the copr in https://github.com/theforeman/foreman-documentation/pull/4832, but I agree, for *nightly* this might be beneficial to include it. But probably out of scope for this particular PR?

arvind4501:
> > > https://theforeman-foreman-documentation-preview-pr-5097.surge.sh/nightly/Installing_Proxy/index-containerized-katello.html#configuring-repositories_smart-proxy here i see we are pulling foremanctl from our nighty repos but since we have not release any version after (2.3.0) which contains any proxy stuff, should we tempraly also enable
> > > ```
> > > dnf copr enable @theforeman/develop rhel-9-x86_64
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > as this contains latest master builds?
> > 
> > 
> > I'm not sure, can/should we do that? @evgeni @ekohl ?
> 
> We dropped the copr in #4832, but I agree, for _nightly_ this might be beneficial to include it. But probably out of scope for this particular PR?

agree, PR should not be blocked on that

guides/common/modules/proc_configuring-smart-proxy-server-with-a-default-ssl-certificate.adoc
evgeni: Funnily, both are right in *this* case. For default/self-signed, we renew the certs AND regenerate the bundle when passing `--certificate-renew`, while for custom server we only regenerate the bundle as we assume the user passed in the new certs. (Not sure this detail is important to the reader, so I'll let Jan decide which wording he likes more)

evgeni (approved):
tech ack

maximiliankolb (approved):
style-wise LGTM

arvind4501 (approved):
Ack, I have followed the guide with `default certs` and was able to deploy proxy following the guide step by step. and it all reads good to me. thanks @jafiala , this is much helpful

PR #5098, opened 2026-07-23
Include content management introduction in containerized build

aneta-petrova:

* Reviewing the existing "Introduction to content" assembly for relevance (and subsequently dropping parts of it)
* Including the remaining content in the containerized build

To extend the containerized guide.

To make Managing Content shorter and avoid duplication.

The module I propose to drop describes 'content management capabilities'. These have recently been documented as 'Foreman use cases' in https://docs.theforeman.org/nightly/Planning_for_Project/index-containerized-katello.html#foreman-use-cases. Additionally, the remaining 'content types' module also partially describes these capabilities. I don't think we need to maintain 'content management capabilities' anymore.

N/A

#### Contributor checklists


maximiliankolb (approved):
Works for me 👍 but I would like an additional ACK.

aneta-petrova:
For the purposes of `tech review`, I'd like to ask reviewers to look at this section:

* (Katello) 1. Content types in Foreman: https://theforeman-foreman-documentation-preview-pr-5098.surge.sh/nightly/Managing_Content/index-containerized-katello.html#Content_Types_in_Foreman_content-management
* (Satellite) 1. Content types in Red Hat Satellite: https://theforeman-foreman-documentation-preview-pr-5098.surge.sh/nightly/Managing_Content/index-satellite.html#Content_Types_in_Satellite_content-management

Are the listed content types supported and available in containerized deployments?

ianballou:
> For the purposes of `tech review`, I'd like to ask reviewers to look at this section:
> 
>     * (Katello) 1. Content types in Foreman: https://theforeman-foreman-documentation-preview-pr-5098.surge.sh/nightly/Managing_Content/index-containerized-katello.html#Content_Types_in_Foreman_content-management
> 
>     * (Satellite) 1. Content types in Red Hat Satellite: https://theforeman-foreman-documentation-preview-pr-5098.surge.sh/nightly/Managing_Content/index-satellite.html#Content_Types_in_Satellite_content-management
> 
> 
> Are the listed content types supported and available in containerized deployments?

There are a number of content types missing from a Katello perspective:

- Errata
- Package groups
- Ansible collections
- Container images
- Python packages (not for Satellite)


---

Kickstart trees are importable in containerized Katello today, but it may be worth mentioning that the mention of provisioning is confusing because containerized Foreman doesn't yet support provisioning as far as I'm aware.

---

Provisioning templates don't really fit our classical definition of content, but perhaps that it outside of the scope of what you're looking for.

aneta-petrova:
> There are a number of content types missing from a Katello perspective:

The text says "**For example**, Foreman supports the following content types:"

So I can either change the section to list all content types, or accept that it's supposed to be an overview rather than a complete list... or just drop the section entirely because a partial overview is confusing.

What if I drop the section and instead add a single sentence along the lines of "Foreman supports a broad range of content types, ranging from RPM packages to ... to custom file types" in [Foreman use cases](https://docs.theforeman.org/nightly/Planning_for_Project/index-katello.html#foreman-use-cases)? I like that idea because it would make Managing Content shorter (it has 22 chapters :scream:) while still providing an overview of content types in a relevant place (a section describing content management as a feature).

> Provisioning templates don't really fit our classical definition of content, but perhaps that it outside of the scope of what you're looking for.

If it doesn't fit, I can remove it. Unless we decide to drop the whole section.

aneta-petrova:
"Foreman use cases" in _Planning_ now has a note that explains what content is and provides examples: https://theforeman-foreman-documentation-preview-pr-5098.surge.sh/nightly/Planning_for_Project/index-containerized-katello.html#foreman-use-cases
And _Managing Content_ no longer includes a section with a list of content types, which I think is okay, because all those chapters on "Managing Content-Type-XYZ" are also a kind of a list... one that will always be complete and up-to-date.

Would this solution work for you @ianballou @maximiliankolb?

ianballou:
> "Foreman use cases" in _Planning_ now has a note that explains what content is and provides examples: https://theforeman-foreman-documentation-preview-pr-5098.surge.sh/nightly/Planning_for_Project/index-containerized-katello.html#foreman-use-cases And _Managing Content_ no longer includes a section with a list of content types, which I think is okay, because all those chapters on "Managing Content-Type-XYZ" are also a kind of a list... one that will always be complete and up-to-date.
> 
> Would this solution work for you @ianballou @maximiliankolb?

I'm okay with this idea. Users looking for an overview will see that content management is a capability, and if they're curious, they can dive into the procedures for different kinds of content. 

You do have to hunt around a little bit to see the different kinds of content types, but a scroll through the sidebar should answer the question "what content types does Katello support managing?".

aneta-petrova:
> You do have to hunt around a little bit to see the different kinds of content types, but a scroll through the sidebar should answer the question "what content types does Katello support managing?".

That is certainly a downside of this change. I wouldn't say it makes things worse compared to the current state of things, though -- which is that there is an overview with a "For example" list that's incomplete, but users might understand it as complete and notice missing content types... Just like it happened here and with @vsedmik :)

An alternative would be to keep this set of changes... but also add a proper, complete reference table with a full list of content types. I'm open to feedback and to hearing everyone's preferences :)

ianballou:
We have a number of places in "1. Introduction to content management" that casually mention some of the content types we support, often mentioning Deb and Yum primarily. This might lead to users wonder why some types are mentioned but not others. So, I think it is nice to have this header page that answers what content type we support, but I agree in its current state that it's unhelpful since it is incomplete.

Since we're breaking into new work scope, I'd say let's remove it since it's essentially incorrect by modern standards, but track that we should include a proper list here soon in a separate chunk of work.

guides/common/modules/con_project-use-cases.adoc
```diff
@@ -44,6 +44,12 @@ Patch management and content distribution::
 Apply updates to hosts in a consistent, scalable, and repeatable manner.
 Import, organize, and distribute content to hosts.
 +
+[NOTE]
+====
+In the context of {Project}, content is defined as the software installed on hosts.
+For example, you can use {Project} to distribute a broad range of content ranging from packages and errata to container images, Ansible content, and custom file types.
```
maximiliankolb, line 50: We have a similar attribute for this: https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-base.adoc#L67 

Although I wonder: Is there non-custom file type content for Satellite? Isn't all file-type content for all builds always "just" file content? Out of scope: but maybe we can simplify this by dropping the attribute and just say "file content" everywhere.

maximiliankolb (commented):
One small suggestion.

My thoughts: Deb and Yum _are_ the main content types, so it's not a bad thing to feature them more prominently. However, a complete & correct list of content types supported by each Katello-flavor is still nice to have. If we add it in a follow-up PR, OK. For now, using the navigation is also OK.

guides/common/modules/con_project-use-cases.adoc
ianballou: There actually are Red Hat file-type repositories. I believe technically any kind of repository could be defined in the subscription manifest.

maximiliankolb: I did not know this. Maybe then just drop "custom" from line 50?

Out of scope: If there is RH file content, then we'd probably have to revisit every instance of `{customfiletype}`.

aneta-petrova:
> We have a number of places in "1. Introduction to content management" that casually mention some of the content types we support, often mentioning Deb and Yum primarily. This might lead to users wonder why some types are mentioned but not others. So, I think it is nice to have this header page that answers what content type we support, but I agree in its current state that it's unhelpful since it is incomplete.
> 
> Since we're breaking into new work scope, I'd say let's remove it since it's essentially incorrect by modern standards, but track that we should include a proper list here soon in a separate chunk of work.

I created https://redhat.atlassian.net/browse/SAT-48256 (public) and tied it to a specific future version to help ensure it gets done in a timely manner. The request is to provide a list of available content types, conditionalized to address the fact that different flavors support different content types. This should eventually provide the needed clarity on what content types are supported.

I think this is a good path to take with these containerization PRs: The goal of a containerization docs PR is to publish containerization-ready documentation, in the timeframe we have. That does not have to mean that all pre-containerization docs must be published for containerized docs: it can also mean that we're publishing whatever we can for the containerization application, and we are tracking gaps through tickets.

guides/common/modules/con_project-use-cases.adoc
aneta-petrova: "custom file type" is used in various places (such as https://github.com/theforeman/foreman-documentation/blob/master/guides/common/assembly_managing-custom-file-type-content-in-project.adoc) and this is the file type that line 50 is referring to. I think it should be okay to keep the term as is -- the point of the line is to hint at the "broad range of content" that can be supported, and naming "custom file type" is one of them is IMO okay.

aneta-petrova:
There have been significant changes since the ack that was originally given so I'd like to ask for re-reviews. @ianballou and @maximiliankolb, are you okay with this PR in its current state? Note the Jira tracker I raised to make sure we provide the complete reference of supported content types eventually.

maximiliankolb (approved):
My ACK still stands.

ianballou (approved):
Works for me, thanks

PR #5105, opened 2026-07-27
Fix parallel build race on shared images and CSS

aneta-petrova:

This is an attempt to prevent race condition failures that seem to be appearing randomly on build jobs.

I've been getting build-html errors quite often, such as this one https://github.com/theforeman/foreman-documentation/actions/runs/30290215039/job/90073083775. Re-running the failed job helps but it would be best to prevent them entirely.

I couldn't figure out what the problem was and a race condition was just a hunch. Claude seems to agree with me and this is the fix it proposes.

#### Contributor checklists

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,12 @@ clean:
 	$(MAKE) -C guides/ clean
 	rm -rf $(DEST) web/output/
 
-html: build-foreman-el build-foreman-deb build-containerized-katello build-containerized-orcharhino build-katello build-orcharhino build-satellite
+BUILDS = foreman-el foreman-deb containerized-katello containerized-orcharhino katello orcharhino satellite
+
+html: FORCE prep
+	@for b in $(BUILDS); do \
+		$(MAKE) -C guides/ html BUILD=$$b || exit 1; \
+	done
 
 build-%: FORCE prep
 	$(MAKE) -C guides/ html BUILD=$*


Makefile
Lennonka, line 1: I assume that you mean to replace the `build-%` build target with the `for` loop? In that case, I think you can delete the whole `build-%` block because it will never run after your change.

I wouldn't mind someone more experienced with make to confirm this.

aneta-petrova: > I assume that you mean to replace the `build-%` build target with the `for` loop? In that case, I think you can delete the whole `build-%` block because it will never run after your change.

I don't think so. The `build-% block` should still be relevant for running a single target builds.

> I wouldn't mind someone more experienced with make to confirm this.

Agreed. Do you have someone particular in mind?

Lennonka: No, I don't have anyone particular in mind.

ekohl: I'm with @aneta-petrova here. The individual targets are still relevant if you want to create an individual build and don't think they would suffer from any concurrency issues. You can still run `make build-foreman-el` to build all `foreman-el` guides. Or `make -j$(nproc) build-foreman-el` if you want concurrency.

Lennonka: @ekohl Can you, please, take a look at the code and review/ack it?

ekohl:
> I couldn't figure out what the problem was and a race condition was just a hunch. Claude seems to agree with me and this is the fix it proposes.

It's a lack of dependency specification. If we look at the error:
```
cp -rf -lL ./images/* ../build/Contributing/images
cp: cannot create directory '../build/Contributing/images': No such file or directory
```

This tells us that it's trying to run `cp` before `../build/Contributing` exists.

In our regular builds we have a prepare step:
https://github.com/theforeman/foreman-documentation/blob/fdf01cbbc1141686abde70f183fc93c5ea45f2f7/guides/common/Makefile#L46-L47

Then `html` depends on the `prepare` step, ensuring `IMAGES_DIR` exists. I suspect it's somehow here:
https://github.com/theforeman/foreman-documentation/blob/fdf01cbbc1141686abde70f183fc93c5ea45f2f7/guides/common/Makefile#L72-L75
That needs the `IMAGES_DIR` to exist, but doesn't depend on it.

I'm proposing https://github.com/theforeman/foreman-documentation/pull/5182 as an alternative fix.

aneta-petrova:
Closing in favor of https://github.com/theforeman/foreman-documentation/pull/5182. Thanks @ekohl and @Lennonka!

PR #5117, opened 2026-07-29
Add troubleshooting reject unknown headers

jafiala:
Add troubleshooting module and upgrade warning

https://redhat.atlassian.net/browse/SAT-47536

#### Contributor checklists

### DIFF
--- a/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
+++ b/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
@@ -33,7 +33,7 @@ HTTP/1.1 403 Forbidden
 .Example output
 [source, none, options="nowrap"]
 ----
-Blocked hosts: _hostname.example.com_
+Jul 15 21:49:28 quadlet.example.com foreman[52746]: 2026-07-15T21:49:28 [E|app|] [ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked hosts: _hostname.example.com_, fo>
 ----
 . Retrieve the FQDN and any server aliases that {ProjectServer} currently accepts:
 +
--- a/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
+++ b/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
@@ -1,0 +1,65 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="troubleshooting-host-header-rejection"]
+= Troubleshooting host header rejection
+
+[role="_abstract"]
+{Project} validates the HTTP `Host` header in each request against the server's FQDN and any names declared with `--server-alias`.
+{Project} rejects requests from undeclared hostnames or IP addresses with a `403 Forbidden` error.
+Over HTTPS, the TLS handshake can also fail if the certificate does not cover the hostname.
+
+To resolve this, declare the hostname as a server alias.
+
+.Procedure
+. Confirm that requests with the FQDN succeed:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# curl -s https://_server.example.com_/api/v2/ping | python3 -m json.tool
+----
++
+A successful response returns JSON with service status information, for example:
++
+[options="nowrap"]
+----
+{
+    "results": {
+        "foreman": {
+            "database": {"active": true, "duration_ms": "1"},
+            "cache": {"servers": [{"status": "ok", "duration_ms": "0"}]}
+        }
+    }
+}
+----
+. Confirm that the alternate hostname is rejected:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# curl -sk https://_alternate.example.com_/api/v2/ping
+----
++
+Over HTTPS, you might see the following client-side TLS rejection:
++
+[options="nowrap"]
+----
+SSL: no alternative certificate subject name matches target hostname 'alternate.example.com'
+----
++
+If you bypass certificate verification with the `--insecure` option, or connect over HTTP, {Project} returns:
++
+[options="nowrap"]
+----
+HTTP/1.1 403 Forbidden
+----
+. If the hostname is legitimate, declare it as a server alias.
+For more information, see xref:configuring-an-alternate-cname_{context}[].
+
+.Verification
+* Repeat the request for the alternate hostname without the `--insecure` option:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# curl -s https://_alternate.example.com_/api/v2/ping | python3 -m json.tool
+----
++
+The request returns JSON with service status information.
--- a/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
+++ b/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
@@ -12,7 +12,7 @@ To allow access from an alternate hostname, declare it as a server alias.
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-# curl -s https://_server.example.com_/api/v2/ping | python3 -m json.tool
+# curl --silent https://_server.example.com_/api/v2/ping | python3 -m json.tool
 ----
 +
 A successful response returns JSON with service status information, for example:
@@ -32,7 +32,7 @@ A successful response returns JSON with service status information, for example:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-# curl -Ik https://_alternate.example.com_/api/v2/ping
+# curl --head --insecure https://_alternate.example.com_/api/v2/ping
 ----
 +
 Over HTTPS, you might see the following client-side TLS rejection:
@@ -56,7 +56,7 @@ For more information, see xref:configuring-an-alternate-cname_{context}[].
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-$ curl -s https://_alternate.example.com_/api/v2/ping | python3 -m json.tool
+$ curl --silent https://_alternate.example.com_/api/v2/ping | python3 -m json.tool
 ----
 +
 The request returns JSON with service status information.
--- a/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
+++ b/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
@@ -44,7 +44,7 @@ Jul 15 21:49:28 quadlet.example.com foreman[52746]: 2026-07-15T21:49:28 [E|app|]
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-# sudo podman secret inspect foreman-settings-yaml --showsecret --format "{{.SecretData}}" | grep ":hosts:" --after-context=3
+# podman secret inspect foreman-settings-yaml --showsecret --format "{{.SecretData}}" | grep ":hosts:" --after-context=3
 ----
 +
 If _hostname.example.com_ does not appear in either output, host header validation is rejecting the request.
--- a/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
+++ b/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
@@ -1,7 +1,7 @@
 :_mod-docs-content-type: PROCEDURE
 
 [id="troubleshooting-host-header-rejection"]
-= Troubleshooting 403 Forbidden errors when connecting to {Project}
+= Troubleshooting host header rejection
 
 [role="_abstract"]
 If you receive a `403 Forbidden` error when connecting to a {ProjectServer} by using a hostname, the request might be rejected by host header validation.
--- a/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
+++ b/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
@@ -44,7 +44,7 @@ Blocked hosts: _hostname.example.com_
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-# grep ServerAlias /etc/httpd/conf.d/05-foreman-ssl.conf
+# sudo podman secret inspect foreman-settings-yaml --showsecret --format "{{.SecretData}}" | grep ":hosts:" --after-context=3
 ----
 +
 If _hostname.example.com_ does not appear in either output, host header validation is rejecting the request.
--- a/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
+++ b/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
@@ -1,62 +1,73 @@
 :_mod-docs-content-type: PROCEDURE
 
 [id="troubleshooting-host-header-rejection"]
-= Troubleshooting host header rejection
+= Troubleshooting 403 Forbidden errors when connecting to {Project}
 
 [role="_abstract"]
-{Project} rejects HTTP requests whose `Host` header does not match the server FQDN or a declared `--server-alias` with a `403 Forbidden` error.
-To allow access from an alternate hostname, declare it as a server alias.
+If you receive a `403 Forbidden` error when connecting to a {ProjectServer} by using a hostname, the request might be rejected by host header validation.
+
+{Project} validates the HTTP `Host` header on every incoming request.
+Only requests whose `Host` header matches the server FQDN or a declared server alias are accepted.
+Requests that use an IP address, a short hostname, or any other undeclared name are rejected with a `403 Forbidden` error.
 
 .Procedure
-. Confirm that requests with the FQDN succeed:
+. Confirm that the error is a `403 Forbidden` response from your {ProjectServer}:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-# curl --silent https://_server.example.com_/api/v2/ping | python3 -m json.tool
+$ curl --head --insecure https://_hostname.example.com_
 ----
-+
-A successful response returns JSON with service status information, for example:
-+
-[source, none, options="nowrap"]
-----
-{
-    "results": {
-        "foreman": {
-            "database": {"active": true, "duration_ms": "1"},
-            "cache": {"servers": [{"status": "ok", "duration_ms": "0"}]}
-        }
-    }
-}
-----
-. Confirm that the alternate hostname is rejected:
-+
-[options="nowrap", subs="+quotes,verbatim,attributes"]
-----
-# curl --head --insecure https://_alternate.example.com_/api/v2/ping
-----
-+
-Over HTTPS, you might see the following client-side TLS rejection:
-+
-[source, none, options="nowrap"]
-----
-SSL: no alternative certificate subject name matches target hostname 'alternate.example.com'
-----
-+
-If you bypass certificate verification with the `--insecure` option, or connect over HTTP, {Project} returns:
 +
 [source, none, options="nowrap"]
 ----
 HTTP/1.1 403 Forbidden
 ----
-. If the hostname is legitimate, declare it as a server alias.
-For more information, see xref:configuring-an-alternate-cname_{context}[].
-
-.Verification
-* Repeat the request for the alternate hostname without the `--insecure` option:
+. Verify that host header validation caused the rejection:
+.. Retrieve the request ID from the response headers:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-$ curl --silent https://_alternate.example.com_/api/v2/ping | python3 -m json.tool
+$ curl --head --insecure https://_hostname.example.com_ 2>&1 | grep -i x-request-id
+----
++
+.Example output
+[source, none, options="nowrap"]
+----
+X-Request-Id: 1a2b3c4d-5e6f-7a8b-9c0d-1e2f3a4b5c6d
+----
+.. Search the {ProjectServer} production log for the request ID from the response header:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# grep _request_ID_ /var/log/foreman/production.log
+----
+. Retrieve the FQDN and any server aliases that {ProjectServer} currently accepts:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# hostname --fqdn
+----
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# grep ServerAlias /etc/httpd/conf.d/05-foreman-ssl.conf
+----
++
+If _hostname.example.com_ does not appear in either output, host header validation is rejecting the request.
+. Resolve the problem:
+* If _hostname.example.com_ is a legitimate alternate name for your {ProjectServer}, declare it as a server alias.
+For more information, see xref:configuring-an-alternate-cname_{context}[].
+* If _hostname.example.com_ is not a name you expect your {Project} to answer, the request might indicate a misconfiguration in your network or a security concern.
+ifdef::satellite[]
+Open a Red{nbsp}Hat Support case for further investigation.
+endif::[]
+
+.Verification
+* After you configure a server alias, verify that the request succeeds:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+$ curl --silent https://_hostname.example.com_/api/v2/ping | python3 -m json.tool
 ----
 +
 The request returns JSON with service status information.
--- a/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
+++ b/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
@@ -4,11 +4,8 @@
 = Troubleshooting host header rejection
 
 [role="_abstract"]
-{Project} validates the HTTP `Host` header in each request against the server's FQDN and any names declared with `--server-alias`.
-{Project} rejects requests from undeclared hostnames or IP addresses with a `403 Forbidden` error.
-Over HTTPS, the TLS handshake can also fail if the certificate does not cover the hostname.
-
-To resolve this, declare the hostname as a server alias.
+{Project} rejects HTTP requests whose `Host` header does not match the server FQDN or a declared `--server-alias` with a `403 Forbidden` error.
+To allow access from an alternate hostname, declare it as a server alias.
 
 .Procedure
 . Confirm that requests with the FQDN succeed:
--- a/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
+++ b/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
@@ -27,7 +27,7 @@ HTTP/1.1 403 Forbidden
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-# journalctl --unit foreman --reverse --lines 50 --grep "Blocked hosts"
+# journalctl --unit foreman --reverse --lines 5 --grep "Blocked hosts"
 ----
 +
 .Example output
--- a/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
+++ b/guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
@@ -31,9 +31,9 @@ HTTP/1.1 403 Forbidden
 ----
 +
 .Example output
-[source, none, options="nowrap"]
+[source, none, options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-Jul 15 21:49:28 quadlet.example.com foreman[52746]: 2026-07-15T21:49:28 [E|app|] [ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked hosts: _hostname.example.com_, _hostname.example.com_
+Jul 15 21:49:28 {foreman-example-com} foreman[52746]: 2026-07-15T21:49:28 [E|app|] [ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked hosts: _hostname.example.com_, _hostname.example.com_
 ----
 . Retrieve the FQDN and any server aliases that {ProjectServer} currently accepts:
 +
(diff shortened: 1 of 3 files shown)


ShimShtein (approved):
Tech ACK. 
LGTM

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,62 @@
...
+----
++
+Over HTTPS, you might see the following client-side TLS rejection:
++
+[options="nowrap"]
+----
+SSL: no alternative certificate subject name matches target hostname 'alternate.example.com'
+----
++
+If you bypass certificate verification with the `--insecure` option, or connect over HTTP, {Project} returns:
```
evgeni: You can't access the API over HTTP. Only HTTPS.

And if the certificate does not include the name, this is a setup issue in itself. For default self-signed certs, `--server-alias` will add the name to the cert, otherwise the user needs to do it.

evgeni: @ekohl this made me realize, should `--server-alias` be verified by `certificates_check`?

guides/doc-Release_Notes/topics/foreman.adoc
```diff
@@ -79,14 +79,24 @@ endif::[]
...
 
 //To add a headline feature, use AsciiDoc's description list: https://docs.asciidoctor.org/asciidoc/latest/lists/description/
 //Summary::
 //Description
 //+
 //Optional second paragraph of description
 
+ifdef::containerized[]
+{Project} rejects HTTP requests with unrecognized host headers::
+{Project} now validates the HTTP `Host` header against the server's FQDN and any declared aliases.
```
maximiliankolb:
```suggestion
{Project} now validates the HTTP `Host` header against the FQDN of the server and any declared aliases.
```

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,62 @@
...
+----
+. If the hostname is legitimate, declare it as a server alias.
+For more information, see xref:configuring-an-alternate-cname_{context}[].
+
+.Verification
+* Repeat the request for the alternate hostname without the `--insecure` option:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# curl -s https://_alternate.example.com_/api/v2/ping | python3 -m json.tool
```
maximiliankolb:
```suggestion
$ curl -s https://_alternate.example.com_/api/v2/ping | python3 -m json.tool
```

Should work as normal user. Same applies to lines 15 and 25.

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
maximiliankolb, line 1: Please use long options for curl for "s" and "k". See `man curl`.

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,62 @@
...
+. Confirm that requests with the FQDN succeed:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# curl -s https://_server.example.com_/api/v2/ping | python3 -m json.tool
+----
++
+A successful response returns JSON with service status information, for example:
++
+[options="nowrap"]
```
maximiliankolb:
```suggestion
[source, none, options="nowrap"]
```

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,62 @@
...
+. Confirm that the alternate hostname is rejected:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# curl -sk https://_alternate.example.com_/api/v2/ping
+----
++
+Over HTTPS, you might see the following client-side TLS rejection:
++
+[options="nowrap"]
```
maximiliankolb:
```suggestion
[source, none, options="nowrap"]
```

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,62 @@
...
+----
+SSL: no alternative certificate subject name matches target hostname 'alternate.example.com'
+----
++
+If you bypass certificate verification with the `--insecure` option, or connect over HTTP, {Project} returns:
++
+[options="nowrap"]
```
maximiliankolb:
```suggestion
[source, none, options="nowrap"]
```

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,62 @@
...
+            "cache": {"servers": [{"status": "ok", "duration_ms": "0"}]}
+        }
+    }
+}
+----
+. Confirm that the alternate hostname is rejected:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# curl -sk https://_alternate.example.com_/api/v2/ping
```
ShimShtein:
```suggestion
# curl -Ik https://_alternate.example.com_/api/v2/ping
```
This will return:
```
$ curl https://foo.example2.com/api/v2/ping -I -k
HTTP/1.1 403 Forbidden
Date: Wed, 15 Jul 2026 15:00:06 GMT
Server: Apache/2.4.62 (CentOS Stream) OpenSSL/3.5.1
Content-Type: text/html; charset=utf-8
strict-transport-security: max-age=631139040; includeSubdomains
x-frame-options: sameorigin
x-content-type-options: nosniff
x-xss-protection: 0
x-download-options: noopen
x-permitted-cross-domain-policies: none
content-security-policy: default-src 'self'; child-src 'self'; connect-src 'self' ws: wss:; img-src 'self' data:; script-src 'unsafe-eval' 'unsafe-inline' 'self'; style-src 'unsafe-inline' 'self'
```
Otherwise `curl` will just return empty output.

Explanation: `-I` to return status and headers, `-k` to bypass mTLS

ShimShtein (changes requested):

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
ekohl: I can't find `certificates_check` in the source, but it may make sense to provide the aliases on the server even if it's not on the certificate because I don't know for sure if any certificate request procedure verifies the target name.

A warning to the user that it may be a misconfiguration (instead of an error) will be helpful though.

evgeni: I of course fucked up the name, it's `certificate_checks`: https://github.com/theforeman/foremanctl/blob/master/src/roles/certificate_checks/tasks/main.yml

Basically the old script from the installer, wrapped into an Ansible role (to be replaced with something more smart in the future)

ekohl (commented):
In troubleshooting I'd expect a flow:
* Start with the problem
* Provide a possible explanation
* Verification the explanation is really the problem
* Provide a solution

In this case, I'd suggest:
* I'm trying to connect to `hostname.example.com` and I'm getting a 403 Forbidden
* A description of why that 403 may be shown (host header validation)
* Provide a mechanism to verify this really is caused by host header validation (I'd assume there is a detailed log in the journal: `curl -I https://hostname.example.com | grep -i x-request-id` and query the production log for that particular request ID)
* Explain to the reader that we take the FQDN and any aliases
* Provide a way to obtain the FQDN and aliases so the user can see if `hostname.example.com` is actually expected to answer. If not, direct the user to configure a CNAME. If it is, there may be a serious problem and they should open a support case.

I think in Red Hat's knowledge base articles this is usually the structure. The user at the beginning doesn't know anything, just the problem they observe.

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
ekohl: That script isn't fed in the aliases right now and I'd prefer to avoid making large changes to it now. We should start planning that "something smart" and include that as an idea. I've created https://github.com/theforeman/foremanctl/issues/714 for that.

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,73 @@
...
+[source, none, options="nowrap"]
+----
+HTTP/1.1 403 Forbidden
+----
+. Verify that host header validation caused the rejection:
+.. Retrieve the request ID from the response headers:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+$ curl --head --insecure https://_hostname.example.com_ 2>&1 | grep -i x-request-id
```
ShimShtein: There is no `x-request-id` header for 403 errors:
```
$ curl --resolve foo.example2.com:443:192.168.122.112 https://foo.example2.com/ --head -k
HTTP/1.1 403 Forbidden
Date: Wed, 15 Jul 2026 21:49:28 GMT
Server: Apache/2.4.62 (CentOS Stream) OpenSSL/3.5.1
Content-Type: text/html; charset=utf-8
strict-transport-security: max-age=631139040; includeSubdomains
x-frame-options: sameorigin
x-content-type-options: nosniff
x-xss-protection: 0
x-download-options: noopen
x-permitted-cross-domain-policies: none
content-security-policy: default-src 'self'; child-src 'self'; connect-src 'self' ws: wss:; img-src 'self' data:; script-src 'unsafe-eval' 'unsafe-inline' 'self'; style-src 'unsafe-inline' 'self'
```
And the log also does not show any request ID. I think the request is rejected before the request-id is generated:
```
Jul 15 21:49:28 quadlet.example.com foreman[52746]: 2026-07-15T21:49:28 [E|app|] [ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked hosts: foo.example2.com, fo>
```
You can see that the last part of `[E|app|]` is empty.
For successful requests you would see:
```
Jul 15 14:52:58 quadlet.example.com foreman[52746]: 2026-07-15T14:52:58 [I|app|4325d823] Completed 200 OK in 143ms (Views: 0.6ms | ActiveRecord: 14.2ms | Allocations: 18481)
```

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,73 @@
...
+.Example output
+[source, none, options="nowrap"]
+----
+X-Request-Id: 1a2b3c4d-5e6f-7a8b-9c0d-1e2f3a4b5c6d
+----
+.. Search the {ProjectServer} production log for the request ID from the response header:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# grep _request_ID_ /var/log/foreman/production.log
```
ShimShtein: We don't have `/var/log/foreman/production.log` anymore. 
We are logging everything using journald. To inspect the logs you have to run:
```
sudo journalctl --unit foreman --reverse --lines 10
```
This will show the last 10 lines of the log in reversed order (latest on top)

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,73 @@
...
+. Retrieve the FQDN and any server aliases that {ProjectServer} currently accepts:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# hostname --fqdn
+----
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# grep ServerAlias /etc/httpd/conf.d/05-foreman-ssl.conf
```
ShimShtein: This won't work for containerized too.
@evgeni do we want to point the user to the `parameters.yaml` file?

ShimShtein (changes requested):

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,73 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="troubleshooting-host-header-rejection"]
+= Troubleshooting host header rejection
+
+[role="_abstract"]
+If you receive a `403 Forbidden` error when connecting to a {ProjectServer} by using a hostname, the request might be rejected by host header validation.
```
aneta-petrova: I'm wondering if the abstract should also hint at the solution. Kind of like this: https://github.com/theforeman/foreman-documentation/blob/master/guides/common/modules/proc_troubleshooting-remote-jobs-timing-out-after-yggdrasil-update.adoc 

An abstract is a summary of a module and the troubleshooting solution is a crucial part of it. I'm also thinking of AI tools using abstracts to collect and process information.

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
ekohl: Ah, it probably doesn't hit the middleware that adds the request ID if the request gets rejected. https://github.com/theforeman/foreman/blob/develop/config/application.rb may need to be modified if that's an option.

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,73 @@
...
+----
++
+[source, none, options="nowrap"]
+----
+HTTP/1.1 403 Forbidden
+----
+. Verify that host header validation caused the rejection:
+.. Retrieve the request ID from the response headers:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
```
jafiala, line 28: The vale error reported on the following lines is a false positive. Opened an issue against `vale at redhat`
https://github.com/redhat-documentation/vale-at-red-hat/issues/1058

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
jafiala: With your suggestion above I think we don't need this step at all since you get the Blocked hosts entry directly.

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
evgeni: Why won't it?

evgeni: Ah, because we never ever set ServerAlias in httpd.conf?

ShimShtein: Because you don't have `/etc/httpd/conf.d/05-foreman-ssl.conf` on your server anymore:
```
[vagrant@quadlet ~]$ ls -l /etc/httpd/conf.d/
total 40
-rw-r--r--. 1 root root 2916 Jun  1 13:26 autoindex.conf
-rw-r--r--. 1 root root  579 Jun  1 13:24 centos-snipolicy.conf
-rw-r--r--. 1 root root 2702 Jul 15 14:43 foreman.conf
-rw-r--r--. 1 root root 5145 Jul 15 14:43 foreman-ssl.conf
-rw-r--r--. 1 root root  400 Jun  1 13:26 README
-rw-r--r--. 1 root root 8720 Jun  1 13:24 ssl.conf
-rw-r--r--. 1 root root 1252 Jun  1 13:24 userdir.conf
[vagrant@quadlet ~]$ grep -R ServerAlias /etc/httpd/conf.d/
[vagrant@quadlet ~]$ 
```

evgeni: Right, the path is different. But even if it was the right path, it wouldn't help.

So, I guess, checking parameters until https://github.com/theforeman/foremanctl/issues/281 is solved is the way to go

ekohl: > Ah, because we never ever set ServerAlias in httpd.conf?

Shouldn't we set it as part of https://github.com/theforeman/foremanctl/pull/471? I suppose we relying on Apache just falling back to the default vhost, which works because we don't have anything else on that machine.

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,67 @@
...
++
+[source, none, options="nowrap"]
+----
+HTTP/1.1 403 Forbidden
+----
+. Verify that host header validation caused the rejection by checking the {ProjectServer} journal for the blocked-host message:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# journalctl --unit foreman --reverse --lines 50 | grep "Blocked hosts"
```
ekohl: `journalctl` has `--grep` so perhaps:
```suggestion
# journalctl --unit foreman --reverse --lines 50 --grep "Blocked hosts"
```

I also found https://github.com/theforeman/foremanctl/issues/467 helpful because it contains a sample:
```
13:34:20 rails.1   | 2026-04-22T13:34:20 [E|app|] [ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked hosts: host.containers.internal, host.containers.internal
```

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
ekohl: I looked at `httpd -t -D DUMP_VHOSTS` to dump the vhosts config, but that doesn't appear to include aliases. Shame.
```console
# httpd -t -D DUMP_VHOSTS
VirtualHost configuration:
*:443                  foreman.example.com (/etc/httpd/conf.d/05-foreman-ssl.conf:6)
*:80                   foreman.example.com (/etc/httpd/conf.d/05-foreman.conf:6)
```

It could still be useful to direct users to the right file.

evgeni: https://github.com/theforeman/foremanctl/pull/718 implements the aliases, but given the rejection comes from rails, it seems odd to look at httpd config?

jafiala: Something like this?

```suggestion
# grep server_aliases /var/lib/obsah/parameters.yaml
```

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,67 @@
...
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# journalctl --unit foreman --reverse --lines 50 --grep "Blocked hosts"
+----
++
+.Example output
+[source, none, options="nowrap"]
+----
+Blocked hosts: _hostname.example.com_
```
jafiala: @ekohl Did you mean use it verbatim just as example output?

```suggestion
13:34:20 rails.1   | 2026-04-22T13:34:20 [E|app|] [ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked hosts: host.containers.internal, host.containers.internal
```

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
evgeni: @jafiala `grep` doesn't really work for yaml, as the values are on the next lines, not the same line.

ShimShtein: @evgeni What's your opinion on:
```
[vagrant@quadlet ~]$ sudo podman secret inspect foreman-settings-yaml --showsecret --format "{{.SecretData}}" | grep ":hosts:" -A 10
:hosts:
  - quadlet.example.com
  - foo.example.com

:ssl_certificate: /etc/foreman/client_cert.pem
:ssl_ca_file: /etc/foreman/katello-default-ca.crt
:ssl_priv_key: /etc/foreman/client_key.pem

:require_ssl: true

:rails_cache_store:
[vagrant@quadlet ~]$
```
This inspects the rails configuration (which is exactly what we are looking for) minus the grep quirks, since we don't have `yq` in base repo.

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
ekohl: I don't have a setup at hand, but I tend to think that a verbatim example from the journal is better than a filtered one. As a user it removes the question whether I really am looking at the right error or not. Still, I'd keep `_hostname.example.com_` instead of `host.containers.internal` as you had previously.

ShimShtein: [This](https://github.com/theforeman/foreman-documentation/pull/5117#discussion_r3684550961) is a verbatim example. I would also recommend using it as is. It is much less confusing to the user.

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
evgeni: `-A10` seems rather arbitrary. In most cases `-A3` is sufficient. but yeah, I guess that works. really would love to have https://github.com/theforeman/foremanctl/pull/718 tho ;)

ShimShtein: It's in already, but you have correctly pointed out that the `ServerAlias` is not entirely related to the bug iteslf, and we should rather examine the rails configuration.

evgeni: that's what you get when commenting between meetings. I meant https://github.com/theforeman/foremanctl/issues/281

ShimShtein: BTW, speaking of `-A 10`: I don't mind changing it to a smaller default, I really don't know how many aliases the users usually have. Also we should use the full option format: `--after-context=NUM`

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
jafiala:
```suggestion
Jul 15 21:49:28 quadlet.example.com foreman[52746]: 2026-07-15T21:49:28 [E|app|] [ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked hosts: _hostname.example.com_, fo>
```

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
jafiala: Added in [c139f25](https://github.com/theforeman/foreman-documentation/pull/5117/commits/c139f25db6de35ba903fb5b982187e14dd71665e)

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
jafiala: I don't think there's anything actionable for this PR here left - the current version uses HTTPS.

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,67 @@
...
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# journalctl --unit foreman --reverse --lines 50 --grep "Blocked hosts"
+----
++
+.Example output
+[source, none, options="nowrap"]
+----
+Jul 15 21:49:28 quadlet.example.com foreman[52746]: 2026-07-15T21:49:28 [E|app|] [ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked hosts: _hostname.example.com_, fo>
```
ShimShtein: Nitpick:

```suggestion
Jul 15 21:49:28 quadlet.example.com foreman[52746]: 2026-07-15T21:49:28 [E|app|] [ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked hosts: _hostname.example.com_, _hostname.example.com_
```

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,67 @@
...
++
+[source, none, options="nowrap"]
+----
+HTTP/1.1 403 Forbidden
+----
+. Verify that host header validation caused the rejection by checking the {ProjectServer} journal for the blocked-host message:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# journalctl --unit foreman --reverse --lines 50 --grep "Blocked hosts"
```
ShimShtein: Since this greps only the "Blocked hosts" lines, we can reduce the amount of lines:

```suggestion
# journalctl --unit foreman --reverse --lines 5 --grep "Blocked hosts"
```
There is no need for more than 5 "Blocked hosts" messages.

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
jafiala: We can't use foo, bar, or other typical metasyntactic variables. I can put in two examples though.

```suggestion
Jul 15 21:49:28 quadlet.example.com foreman[52746]: 2026-07-15T21:49:28 [E|app|] [ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked hosts: _hostname.example.com_, _hostname1.example.com_, _hostname2.example.com_
```

ShimShtein: Yes please! For me it was the same name twice, but I suppose it may vary for the user

ShimShtein:
Testing ACK. I have tried all the commands and the output looked as expected. 
Tech ACK too. We have discussed all the issues.

ShimShtein (approved):

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,67 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="troubleshooting-host-header-rejection"]
+= Troubleshooting host header rejection
+
+[role="_abstract"]
+If you receive a `403 Forbidden` error when connecting to a {ProjectServer} by using a hostname, the request might be rejected by host header validation.
```
maximiliankolb:
```suggestion
If you receive a `403 Forbidden` error when connecting to {ProjectServer} by using a hostname, the request might be rejected by host header validation.
```

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,67 @@
...
+. Retrieve the FQDN and any server aliases that {ProjectServer} currently accepts:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# hostname --fqdn
+----
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# sudo podman secret inspect foreman-settings-yaml --showsecret --format "{{.SecretData}}" | grep ":hosts:" --after-context=3
```
evgeni:
```suggestion
# podman secret inspect foreman-settings-yaml --showsecret --format "{{.SecretData}}" | grep ":hosts:" --after-context=3
```

`#` already implies a root shell, no sudo required

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,67 @@
...
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# journalctl --unit foreman --reverse --lines 5 --grep "Blocked hosts"
+----
++
+.Example output
+[source, none, options="nowrap"]
+----
+Jul 15 21:49:28 quadlet.example.com foreman[52746]: 2026-07-15T21:49:28 [E|app|] [ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked hosts: _hostname.example.com_, _hostname1.example.com_, _hostname2.example.com_
```
jafiala:
```suggestion
Jul 15 21:49:28 quadlet.example.com foreman[52746]: 2026-07-15T21:49:28 [E|app|] [ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked hosts: _hostname.example.com_, _hostname.example.com_
```

@evgeni Based on your demo this is the actual (example) output, right? I had misrepresented @ShimShtein 's previous suggestion as two separate rejected hosts.

evgeni: yeah, it doubles the "bad" hostname

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,67 @@
...
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# journalctl --unit foreman --reverse --lines 5 --grep "Blocked hosts"
+----
++
+.Example output
+[source, none, options="nowrap"]
+----
+Jul 15 21:49:28 quadlet.example.com foreman[52746]: 2026-07-15T21:49:28 [E|app|] [ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked hosts: _hostname.example.com_, _hostname.example.com_
```
evgeni:
```suggestion
Jul 15 21:49:28 {foreman-example-com} foreman[52746]: 2026-07-15T21:49:28 [E|app|] [ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked hosts: _hostname.example.com_, _hostname.example.com_
```

guides/common/modules/proc_troubleshooting-host-header-rejection.adoc
```diff
@@ -0,0 +1,67 @@
...
+----
+. Verify that host header validation caused the rejection by checking the {ProjectServer} journal for the blocked-host message:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# journalctl --unit foreman --reverse --lines 5 --grep "Blocked hosts"
+----
++
+.Example output
+[source, none, options="nowrap"]
```
evgeni:
```suggestion
[source, none, options="nowrap", subs="+quotes,verbatim,attributes"]
```

evgeni (approved):
tiny nitpick, but tech/test ack²

maximiliankolb (approved):

PR #5094, opened 2026-07-22
Allow webhooks in containerized deployments

adamruzicka:
https://redhat.atlassian.net/browse/SAT-47668

#### What changes are you introducing?

Once https://github.com/theforeman/foremanctl/pull/694 goes in, webhooks (without shellhooks) will be available in containerized deployments. This change enables the relevant parts of documentation.

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

See above.

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)

See the TODOs in the changes.

#### Contributor checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [ ] Foreman 3.19/Katello 4.21
* [ ] Foreman 3.18/Katello 4.20 (Satellite 6.19; orcharhino 7.9)
* [ ] Foreman 3.17/Katello 4.19
* [ ] Foreman 3.16/Katello 4.18 (Satellite 6.18; orcharhino 7.6, 7.7, and 7.8)
* [ ] Foreman 3.15/Katello 4.17
* [ ] Foreman 3.14/Katello 4.16 (Satellite 6.17; orcharhino 7.4; orcharhino 7.5)
* We do not accept PRs for Foreman older than 3.14.

PR #5118, opened 2026-07-29
Allow ansible in containerized deployments

adamruzicka:

Allowing ansible in containers, a companion to https://github.com/theforeman/foremanctl/pull/444

See above.

The support for ansible in containers is incomplete, hence the ifdefs applied liberally.

#### Contributor checklists

### DIFF
--- a/guides/common/assembly_configuring-ansible-definitions-to-apply-to-hosts.adoc
+++ b/guides/common/assembly_configuring-ansible-definitions-to-apply-to-hosts.adoc
@@ -1,7 +1,9 @@
 include::modules/con_configuring-ansible-definitions-to-apply-to-hosts.adoc[]
 
 ifndef::foreman-deb[]
+ifndef::containerized[]
 include::modules/proc_adding-rhel-system-roles.adoc[leveloffset=+1]
+endif::[]
 endif::[]
 
 include::modules/proc_importing-ansible-roles-and-variables.adoc[leveloffset=+1]
--- a/guides/common/modules/con_ansible-versions-supported-for-integration-with-project.adoc
+++ b/guides/common/modules/con_ansible-versions-supported-for-integration-with-project.adoc
@@ -20,6 +20,6 @@ ifndef::foreman-deb[]
 // Must match https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/plugins/ansible-core/ansible-core.spec
 The supported version of Ansible is `ansible-core` 2.16.
 ifdef::containerized[]
-This version is shipped with official {SmartProxyServer} container images.
+This version is shipped with {SmartProxyServer} container images.
 endif::[]
 endif::[]
--- a/guides/common/modules/con_ansible-versions-supported-for-integration-with-project.adoc
+++ b/guides/common/modules/con_ansible-versions-supported-for-integration-with-project.adoc
@@ -19,4 +19,7 @@ ifndef::foreman-deb[]
 {Project} repositories include `ansible-core` as a packaged dependency for Ansible integration with {ProjectServer} and {SmartProxyServers}.
 // Must match https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/plugins/ansible-core/ansible-core.spec
 The supported version of Ansible is `ansible-core` 2.16.
+ifdef::containerized[]
+This version is shipped with official {SmartProxyServer} container images.
+endif::[]
 endif::[]
--- a/guides/common/modules/proc_enabling-the-ansible-plugin.adoc
+++ b/guides/common/modules/proc_enabling-the-ansible-plugin.adoc
@@ -19,8 +19,7 @@ ifndef::containerized[]
 endif::[]
 ifdef::containerized[]
 # {foremanctl} \
---add-feature ansible \
---add-feature foreman-proxy
+--add-feature ansible
 endif::[]
 ----
 . On all {SmartProxyServers} that you want to use to run Ansible roles on hosts, enable the Ansible plugin:
@@ -35,6 +34,6 @@ ifndef::containerized[]
 # {foreman-installer} --enable-foreman-proxy-plugin-ansible
 endif::[]
 ifdef::containerized[]
-# {foremanctl} --add-feature ansible
+# {foremanctl} deploy-proxy --add-feature ansible
 endif::[]
 ----
--- a/guides/common/modules/proc_enabling-the-ansible-plugin.adoc
+++ b/guides/common/modules/proc_enabling-the-ansible-plugin.adoc
@@ -9,6 +9,7 @@ Before you can use Ansible for configuration management in {Project}, enable Ans
 .Procedure
 ifndef::satellite[]
 . Enable the Ansible plugin on your {ProjectServer}:
+ifndef::containerized[]
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
@@ -16,6 +17,16 @@ ifndef::satellite[]
 --enable-foreman-plugin-ansible \
 --enable-foreman-proxy-plugin-ansible
 ----
+endif::[]
+ifdef::containerized[]
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} \
+--add-feature ansible \
+--add-feature foreman-proxy
+----
+endif::[]
 . On all {SmartProxyServers} that you want to use to run Ansible roles on hosts, enable the Ansible plugin:
 endif::[]
 ifdef::satellite[]
@@ -23,6 +34,13 @@ ifdef::satellite[]
 endif::[]
 +
 [options="nowrap" subs="+quotes,attributes"]
+ifndef::containerized[]
 ----
 # {foreman-installer} --enable-foreman-proxy-plugin-ansible
 ----
+endif::[]
+ifdef::containerized[]
+----
+# {foremanctl} --add-feature ansible
+----
+endif::[]
--- a/guides/common/modules/proc_enabling-the-ansible-plugin.adoc
+++ b/guides/common/modules/proc_enabling-the-ansible-plugin.adoc
@@ -9,24 +9,20 @@ Before you can use Ansible for configuration management in {Project}, enable Ans
 .Procedure
 ifndef::satellite[]
 . Enable the Ansible plugin on your {ProjectServer}:
-ifndef::containerized[]
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
+ifndef::containerized[]
 # {foreman-installer} \
 --enable-foreman-plugin-ansible \
 --enable-foreman-proxy-plugin-ansible
-----
 endif::[]
 ifdef::containerized[]
-+
-[options="nowrap" subs="+quotes,attributes"]
-----
 # {foremanctl} \
 --add-feature ansible \
 --add-feature foreman-proxy
-----
 endif::[]
+----
 . On all {SmartProxyServers} that you want to use to run Ansible roles on hosts, enable the Ansible plugin:
 endif::[]
 ifdef::satellite[]
@@ -34,13 +30,11 @@ ifdef::satellite[]
 endif::[]
 +
 [options="nowrap" subs="+quotes,attributes"]
+----
 ifndef::containerized[]
-----
 # {foreman-installer} --enable-foreman-proxy-plugin-ansible
-----
 endif::[]
 ifdef::containerized[]
-----
 # {foremanctl} --add-feature ansible
-----
 endif::[]
+----
--- a/guides/common/modules/ref_ansible-role-and-collection-paths-in-project.adoc
+++ b/guides/common/modules/ref_ansible-role-and-collection-paths-in-project.adoc
@@ -10,19 +10,19 @@ System-installed roles and collections are located in `/usr/share/ansible`, whil
 {Project} imports and runs Ansible roles from the following paths:
 
 ifdef::containerized[]
-* `/var/lib/foreman-proxy/ansible/roles`
-* `/var/lib/foreman-proxy/ansible/collections`
+* `/var/lib/foreman-proxy/ansible/roles/`
+* `/var/lib/foreman-proxy/ansible/collections/`
 
-If you want to add custom roles or collections, place them under `/var/lib/foreman-proxy/ansible`.
+If you want to add custom roles or collections, place them under `/var/lib/foreman-proxy/ansible/`.
 endif::[]
 ifndef::containerized[]
-* `/etc/ansible/roles`
-* `/usr/share/ansible/roles`
-* `/etc/ansible/collections`
-* `/usr/share/ansible/collections`
+* `/etc/ansible/roles/`
+* `/usr/share/ansible/roles/`
+* `/etc/ansible/collections/`
+* `/usr/share/ansible/collections/`
 
-Roles and collections from installed packages are placed under `/usr/share/ansible`.
-If you want to add custom roles or collections, place them under `/etc/ansible`.
+Roles and collections from installed packages are placed under `/usr/share/ansible/`.
+If you want to add custom roles or collections, place them under `/etc/ansible/`.
 endif::[]
 
 ifdef::satellite[]
--- a/guides/common/modules/ref_ansible-role-and-collection-paths-in-project.adoc
+++ b/guides/common/modules/ref_ansible-role-and-collection-paths-in-project.adoc
@@ -9,6 +9,13 @@ System-installed roles and collections are located in `/usr/share/ansible`, whil
 
 {Project} imports and runs Ansible roles from the following paths:
 
+ifdef::containerized[]
+* `/var/lib/foreman-proxy/ansible/roles`
+* `/var/lib/foreman-proxy/ansible/collections`
+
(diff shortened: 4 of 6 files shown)


guides/common/modules/proc_enabling-the-ansible-plugin.adoc
```diff
@@ -9,20 +9,38 @@ Before you can use Ansible for configuration management in {Project}, enable Ans
...
 +
 [options="nowrap" subs="+quotes,attributes"]
+ifndef::containerized[]
 ----
 # {foreman-installer} --enable-foreman-proxy-plugin-ansible
 ----
+endif::[]
+ifdef::containerized[]
+----
+# {foremanctl} --add-feature ansible
```
adamruzicka: Can ifdefs go inside the code blocks? The diff could be slightly smaller.

guides/common/modules/proc_enabling-the-ansible-plugin.adoc
```diff
@@ -9,20 +9,38 @@ Before you can use Ansible for configuration management in {Project}, enable Ans
...
 --enable-foreman-proxy-plugin-ansible
 ----
+endif::[]
+ifdef::containerized[]
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# {foremanctl} \
+--add-feature ansible \
+--add-feature foreman-proxy
```
adamruzicka: Do we want to suggest this from the get go or should we aim for cleaner foreman vs proxy separation?

guides/common/assembly_configuring-ansible-definitions-to-apply-to-hosts.adoc
```diff
@@ -1,8 +1,10 @@
 include::modules/con_configuring-ansible-definitions-to-apply-to-hosts.adoc[]
 
 ifndef::foreman-deb[]
+ifndef::containerized[]
 include::modules/proc_adding-rhel-system-roles.adoc[leveloffset=+1]
```
adamruzicka, line 5: note: to be solved, most likely they'll need to be baked into the image?

guides/doc-Managing_Configurations_Ansible/master.adoc
```diff
@@ -25,7 +17,9 @@ include::common/assembly_configuring-ansible-definitions-to-apply-to-hosts.adoc[
 
 include::common/assembly_applying-ansible-definitions-to-hosts.adoc[leveloffset=+1]
 
+ifndef::containerized[]
 include::common/modules/proc_using-ansible-vault-with-project.adoc[leveloffset=+1]
+endif::[]
```
adamruzicka, line 22: note: this relies on `ansible.cfg` being propagated to the container, which still needs to be solved.

adamruzicka (commented):
I'll just park some notes here

guides/common/modules/ref_ansible-role-and-collection-paths-in-project.adoc
```diff
@@ -9,13 +9,21 @@ System-installed roles and collections are located in `/usr/share/ansible`, whil
 
 {Project} imports and runs Ansible roles from the following paths:
 
+ifdef::containerized[]
+* `/var/lib/foreman-proxy/ansible/roles`
```
adamruzicka: shush vale, it is a filesystem path, I'm not at liberty to capitalize it freely

guides/common/modules/proc_enabling-the-ansible-plugin.adoc
aneta-petrova: Yes, that should work.

adamruzicka: done

guides/common/modules/proc_enabling-the-ansible-plugin.adoc
evgeni: I would prefer cleaner separation. But I am not in a position to decide it.

adamruzicka:
https://github.com/theforeman/foremanctl/pull/444 just went in, undrafting.

guides/common/modules/con_ansible-versions-supported-for-integration-with-project.adoc
```diff
@@ -19,4 +19,7 @@ ifndef::foreman-deb[]
 {Project} repositories include `ansible-core` as a packaged dependency for Ansible integration with {ProjectServer} and {SmartProxyServers}.
 // Must match https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/plugins/ansible-core/ansible-core.spec
 The supported version of Ansible is `ansible-core` 2.16.
+ifdef::containerized[]
+This version is shipped with official {SmartProxyServer} container images.
```
maximiliankolb: I would drop "official" here: does are always only for whatever {Team} provides/ships.

guides/common/modules/ref_ansible-role-and-collection-paths-in-project.adoc
maximiliankolb:
```suggestion
* `/var/lib/foreman-proxy/ansible/roles/`
```

guides/common/modules/ref_ansible-role-and-collection-paths-in-project.adoc
```diff
@@ -9,13 +9,21 @@ System-installed roles and collections are located in `/usr/share/ansible`, whil
...
+* `/var/lib/foreman-proxy/ansible/collections`
```
maximiliankolb:
```suggestion
* `/var/lib/foreman-proxy/ansible/collections/`
```

I suggest to add a trailing slash to indicate that it's a path.

guides/common/modules/ref_ansible-role-and-collection-paths-in-project.adoc
```diff
@@ -9,13 +9,21 @@ System-installed roles and collections are located in `/usr/share/ansible`, whil
...
+
+If you want to add custom roles or collections, place them under `/var/lib/foreman-proxy/ansible`.
```
maximiliankolb:
```suggestion
If you want to add custom roles or collections, place them under `/var/lib/foreman-proxy/ansible/`.
```

guides/common/modules/ref_customizing-ansible-configuration.adoc
```diff
@@ -7,6 +7,11 @@
 {Project} manages essential Ansible configuration that is required for Ansible integration with {Project}.
 However, you can customize other Ansible configuration options as usual.
 
+ifdef::containerized[]
+Ansible reads configuration from the environment provided by {SmartProxy}.
```
maximiliankolb: non-blocking: this is a bit unclear to me. Do you mean env vars or files accessible to SmartProxies? What is the environment exactly and how does the user configure it?

adamruzicka: Some time ago, we wanted to allow users to be able to put whatever they want into their /etc/ansible.cfg (tbd in containers), while still leaving us some means of setting the options in the way we need to keep the foreman<->ansible integration working. What we landed on is this two-phased approach:
1) The installer deploys /etc/ansible.env (in "traditional" deployments, in containers it is a podman secret), where we set whatever we need
2) Before SmartProxy starts an ansible job, it sources this file and executes ansible with env vars from that file set
3) Ansible itself reads /etc/ansible.cfg, but the values from environment "win"

In containers, we don't have the modify-able ansible.cfg, so we're only left with the environment that we provide.

> Do you mean env vars or files accessible to SmartProxies?

A little bit of both, a file (/etc/ansible.env) sourced by SmartProxy and passed to ansible as environment variables.

> how does the user configure it?

They don't and that's the point. It is there strictly for us to be able to set the options that are needed for the integration to work fully.

guides/common/modules/ref_ansible-role-and-collection-paths-in-project.adoc
adamruzicka: Also added to the non-containerized paths

guides/common/modules/ref_customizing-ansible-configuration.adoc
maximiliankolb: > They don't and that's the point.

OK, then maybe the following suggestion:

```suggestion
Ansible reads configuration provided by {SmartProxy}.
```

guides/common/modules/ref_ansible-role-and-collection-paths-in-project.adoc
```diff
@@ -9,13 +9,21 @@ System-installed roles and collections are located in `/usr/share/ansible`, whil
...
-* `/usr/share/ansible/roles`
-* `/etc/ansible/collections`
-* `/usr/share/ansible/collections`
+ifdef::containerized[]
+* `/var/lib/foreman-proxy/ansible/roles/`
+* `/var/lib/foreman-proxy/ansible/collections/`
 
-Roles and collections from installed packages are placed under `/usr/share/ansible`.
-If you want to add custom roles or collections, place them under `/etc/ansible`.
+If you want to add custom roles or collections, place them under `/var/lib/foreman-proxy/ansible`.
```
maximiliankolb:
```suggestion
If you want to add custom roles or collections, place them under `/var/lib/foreman-proxy/ansible/`.
```

maximiliankolb (commented):
another tiny suggestion.

maximiliankolb (approved):
style-wise LGTM

guides/common/modules/ref_ansible-role-and-collection-paths-in-project.adoc
```diff
@@ -9,13 +9,21 @@ System-installed roles and collections are located in `/usr/share/ansible`, whil
 
 {Project} imports and runs Ansible roles from the following paths:
 
-* `/etc/ansible/roles`
-* `/usr/share/ansible/roles`
-* `/etc/ansible/collections`
-* `/usr/share/ansible/collections`
+ifdef::containerized[]
+* `/var/lib/foreman-proxy/ansible/roles/`
```
ekohl, line 13: Question for the docs folks. Perhaps the vale rule can differentiate `Ansible` (surrounded by whitespace) with `/ansible` (part of a path) to reduce false positives.

guides/common/assembly_configuring-ansible-definitions-to-apply-to-hosts.adoc
ekohl: Yes, I think so too.

guides/common/modules/ref_ansible-role-and-collection-paths-in-project.adoc
maximiliankolb: Maybe, yes. Other affected examples: `[id="enabling-the-ansible-plugin"]` and `{ansible-namespace}.convert2rhel`.

But maybe also not trivial.

aneta-petrova: Vale should ignore strings in backticks, attributes, etc. When running the check locally, I don't see either of these (backticks, ID, attribute) flagged. Could it be that the CI check uses a different version of Vale than my local check? I'll see if there's anything we can do about it.

However, that said, the comment here also says "unless you are using the term in a context that requires different capitalization". We added that part specifically for these cases.

aneta-petrova: It's not about the versions. This looks like a Vale bug.

It's not just line 13 that's being flagged; it's also line 33 with a link where the URL includes `ansible`. There is a second link on line 34 with `ansible` in its URL, but that is not being flagged -- because Vale gets confused and instead of reporting the URL, it reports line 13. It's weird but when I delete the link on line 34, the error for line 13 suddenly disappears.

ekohl: > Vale should ignore strings in backticks, attributes, etc. When running the check locally, I don't see either of these (backticks, ID, attribute) flagged. Could it be that the CI check uses a different version of Vale than my local check? I'll see if there's anything we can do about it.

If I look at the action output it installed Vale 3.17.0 in this run. Not sure which version you have deployed.

> However, that said, the comment here also says "unless you are using the term in a context that requires different capitalization". We added that part specifically for these cases.

My point is that you still see a "Check failure" in the file overview, requiring a human to make a judgement. If we can avoid that, I'd be happy.

ekohl: Somehow I didn't see your last reply when I created mine.

> It's not just line 13 that's being flagged; it's also line 33 with a link where the URL includes `ansible`. There is a second link on line 34 with `ansible` in its URL, but that is not being flagged -- because Vale gets confused and instead of reporting the URL, it reports line 13. It's weird but when I delete the link on line 34, the error for line 13 suddenly disappears.

That is very confusing.

aneta-petrova: Based on my local testing, this should make the misattributed error disappear: https://github.com/theforeman/foreman-documentation/pull/5156

The PR proposes to update the `foreman-documentation/Capitalization.yml` rule to ignore lowercase `ansible` when preceded or followed by `-`. In other words, to ignore lowercase `ansible` in the URL paths we often use in this project -- `{GuideNameDocURL}blah-blah-ansible-blah`. (When the URL is spelled out, such as `https://docs.theforeman.org/blah-blah-ansible-blah`, Vale correctly ignores it.)

The vale-at-red-hat style package has its own capitalization rule at https://github.com/redhat-documentation/vale-at-red-hat/blob/main/.vale/styles/RedHat/CaseSensitiveTerms.yml  That one still reports line 13 but because the severity of `CaseSensitiveTerms.yml` is set to `warning` only (unlike `Capitalization.yml`, which is set to `error`), this vale-at-red-hat issue does not produce a visible reviewdog message here on GitHub.

aneta-petrova: It took me a looooooooong time to write the message above and I only saw your comments after I sent it.

> My point is that you still see a "Check failure" in the file overview, requiring a human to make a judgement. If we can avoid that, I'd be happy.

Agreed, I'm trying to work my way to the same thing :) I think the PR I linked above should achieve it. Or at least it worked for me locally 🙃

adamruzicka: So, since this pr got sort of hijacked anyway, I'll rebase this on top of yours and we'll see what happens?

aneta-petrova: @adamruzicka That would certainly help test my PR, yes :) Thanks

adamruzicka: vale seems to be happy now

aneta-petrova: Yay! Some might consider it an ugly-ish hack, as I'm explaining in https://github.com/theforeman/foreman-documentation/pull/5156#issuecomment-5193004329, so I guess I'll file an issue to track this change to the rule, while also suggesting something more appropriate as a long-term solution -- for whenever someone has time to look into it.

ekohl: > Based on my local testing, this should make the misattributed error disappear: #5156

That PR looks pretty to what I had in mind, except the whole dashes vs slashes confusion. Good stuff.

aneta-petrova: I merged https://github.com/theforeman/foreman-documentation/pull/5156 because I found another PR that was hitting a similar issue with `ansible` in a URL path. That means a conflict for your here @adamruzicka but I also think it's a cleaner approach to merge the Vale fix elsewhere.

adamruzicka: No problem, rebased

aneta-petrova: And here's the issue to track implementing a more systematic fix: https://github.com/theforeman/foreman-documentation/issues/5160 We might end up closing it if we find the existing fix to be good enough... but it's good to have the context recorded in an issue.

guides/common/modules/proc_enabling-the-ansible-plugin.adoc
adamruzicka: done

adamlazik1 (approved):
tech-wise LGTM

guides/common/modules/proc_enabling-the-ansible-plugin.adoc
```diff
@@ -12,9 +12,15 @@ ifndef::satellite[]
...
 [options="nowrap" subs="+quotes,attributes"]
 ----
+ifndef::containerized[]
 # {foreman-installer} \
 --enable-foreman-plugin-ansible \
 --enable-foreman-proxy-plugin-ansible
+endif::[]
+ifdef::containerized[]
+# {foremanctl} \
+--add-feature ansible
```
ekohl, line 22: Same comment as https://github.com/theforeman/foreman-documentation/pull/5094/changes#r3773405597.

This was later fixed in #5184. @adamlazik1 when providing the tech ack, please verify the commands actually work in the rendered docs. Ideally by copy-pasting and actually running them.

PR #5132, opened 2026-08-03
Make ISS requirement for disconnected

jafiala:
Add ISS as a requirement for the disconnected installation

ISS is required, after other methods were removed
https://redhat.atlassian.net/browse/SAT-48329

Reorganizing the content further is out of scope.

#### Contributor checklists

* [x] Foreman 5.0/Katello 5.0
* [x] Foreman 3.18/Katello 4.20 (Satellite 6.19; orcharhino 7.9)

### DIFF
--- a/guides/common/modules/con_configuring-inter-server-synchronization.adoc
+++ b/guides/common/modules/con_configuring-inter-server-synchronization.adoc
@@ -6,7 +6,7 @@
 [role="_abstract"]
 ifdef::satellite[]
 A disconnected {ProjectServer} does not have direct access to the Red Hat Content Delivery Network.
-You must configure {ISS} to synchronize content from a connected {ProjectServer} to your disconnected {ProjectServer}.
+Configure {ISS} on your disconnected {ProjectServer} to synchronize content from a connected {ProjectServer}.
 endif::[]
 ifndef::satellite[]
 Configure {ISS} on your downstream {ProjectServer} to provide content to its hosts.
--- a/guides/common/modules/con_configuring-inter-server-synchronization.adoc
+++ b/guides/common/modules/con_configuring-inter-server-synchronization.adoc
@@ -4,7 +4,7 @@
 = Configuring {ISS}
 
 [role="_abstract"]
-A disconnected {ProjectServer} does not have direct access to the Red Hat Content Delivery Network.
+A disconnected {ProjectServer} does not have direct access to networked content, such as content delivery networks, container registries, or software repositories.
 Configure {ISS} on your downstream {ProjectServer} to provide content to its hosts.
 
 .Additional resources
--- a/guides/common/modules/con_configuring-inter-server-synchronization.adoc
+++ b/guides/common/modules/con_configuring-inter-server-synchronization.adoc
@@ -5,8 +5,12 @@
 
 [role="_abstract"]
 ifdef::satellite[]
-Configure {ISS} on your disconnected {ProjectServer} to provide content in your disconnected network.
+A disconnected {ProjectServer} does not have direct access to the Red Hat Content Delivery Network.
+You must configure {ISS} to synchronize content from a connected {ProjectServer} to your disconnected {ProjectServer}.
 endif::[]
 ifndef::satellite[]
 Configure {ISS} on your downstream {ProjectServer} to provide content to its hosts.
 endif::[]
+
+.Additional resources
+* {PlanningDocURL}{ISS-id}-scenarios[{ISS} scenarios in _{PlanningDocTitle}_]
--- a/guides/common/modules/con_configuring-inter-server-synchronization.adoc
+++ b/guides/common/modules/con_configuring-inter-server-synchronization.adoc
@@ -4,13 +4,8 @@
 = Configuring {ISS}
 
 [role="_abstract"]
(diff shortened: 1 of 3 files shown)


guides/common/modules/con_configuring-inter-server-synchronization.adoc
```diff
@@ -5,8 +5,12 @@
 
 [role="_abstract"]
 ifdef::satellite[]
```
jafiala, line 7: The abstract is fine, the conditionals make it too long. It will resolve into a correct abstract.

maximiliankolb (approved):
style-wise LGTM

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -46,6 +46,8 @@ endif::[]
 
 ifndef::containerized[]
 
+include::common/assembly_configuring-inter-server-synchronization.adoc[leveloffset=+1]
```
evgeni: I don't think we should pull it that far upfront and raise the level.

ISS is the primary way how people should obtain content, but it's not the only one. We also document "custom CDN". So I'd put it before Custom CDN (L55/57).

Or you create a separate "Obtaining content" chapter, and place it before "additional configuration" (as it's really not additional, you need to do it).

(Also, this all smells a lot like it doesn't belong to "Installing", but into "Content" guides, but oh well…)

guides/common/modules/con_configuring-inter-server-synchronization.adoc
```diff
@@ -5,8 +5,12 @@
...
-Configure {ISS} on your disconnected {ProjectServer} to provide content in your disconnected network.
+A disconnected {ProjectServer} does not have direct access to the Red Hat Content Delivery Network.
+You must configure {ISS} to synchronize content from a connected {ProjectServer} to your disconnected {ProjectServer}.
 endif::[]
 ifndef::satellite[]
 Configure {ISS} on your downstream {ProjectServer} to provide content to its hosts.
```
ekohl, line 12: I'm wondering why we don't use upstream and downstream `{ProjectServer}` in our Satellite documentation. It's IMHO the right naming for it when using ISS. For an abstract, isn't this line perfect?

ekohl (commented):
In a discussion with @ehelms he pointed out that some customers sync content from other sources than ISS so perhaps it shouldn't be a hard requirement, but a strong recommendation as the default. I'm sorry if I misled you there.

> No diff compared to the current base

This makes it really hard to review this. Perhaps we should wait until we have a disconnected guide again so that we can verify if the entire flow works?

guides/common/modules/con_configuring-inter-server-synchronization.adoc
```diff
@@ -5,8 +5,12 @@
...
 endif::[]
+
+.Additional resources
```
archanaserver, line 15: Just by looking at it, i'm  trying to understand when we want to make it as a requirement for disconnected then why it goes  in the additional config, that makes it optional, doesn't it?

jafiala: Yes, that was Evgeni's suggestion, now supported by Ewoud's investigations. I'll have to rephrase the hard requirement, apparently.

guides/common/modules/con_configuring-inter-server-synchronization.adoc
jafiala: Connected/disconnected is what the existing reference (https://docs.redhat.com/en/documentation/red_hat_satellite/6.19/html-single/overview_concepts_and_deployment_considerations/index#iss-scenarios_planning) uses, so that's what I used to be consistent. We can change to upstream/downstream, but to me, "upstream/downstream" might have different meanings. Connected/disconnected is immediately understandable, and also directs users to the respective install guides.

ekohl: Fair enough, but shouldn't we keep non-Satellite aligned? What I'm really driving at: do we need these `ifdef` macros? IMHO where possible we should always aim to align.

guides/common/modules/con_configuring-inter-server-synchronization.adoc
```diff
@@ -4,9 +4,8 @@
 = Configuring {ISS}
 
 [role="_abstract"]
-ifdef::satellite[]
-Configure {ISS} on your disconnected {ProjectServer} to provide content in your disconnected network.
-endif::[]
-ifndef::satellite[]
+A disconnected {ProjectServer} does not have direct access to the Red Hat Content Delivery Network.
```
ekohl: Perhaps it should first be generic? Users can also sync in content from container registries or yum repositories like EPEL and a disconnected install doesn't have direct access to any networked content (such as the RH CDN).

evgeni (commented):
ACK based on diff, given there is no working preview right now

evgeni (approved):

guides/common/modules/con_configuring-inter-server-synchronization.adoc
jafiala: @ekohl Reworded, please ack.

ekohl (approved):

PR #5136, opened 2026-08-03
Remove unused CandlepinVersion attribute

zjhuntin:
## Summary
- Update `:CandlepinVersion:` from 4.8 to 5.0 in `guides/common/attributes.adoc`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/guides/common/attributes.adoc
+++ b/guides/common/attributes.adoc
@@ -13,7 +13,6 @@
 :ProjectVersionPrevious-Previous: 3.18
 :KatelloVersion: nightly
 :PulpcoreVersion: 3.105
-:CandlepinVersion: 4.8
 
 //
 // WHERE ARE MY ATTRIBUTES?


maximiliankolb (commented):
The attribute is not used anywhere. So maybe it would be easier long-term to drop it now?

For Foreman+Katello: https://docs.theforeman.org/nightly/Installing_Server/index-katello.html#configuring-repositories_foreman > `wget https://yum.theforeman.org/katello/nightly/katello/el9/x86_64/katello-repos-latest.rpm`, `rpm2cpio katello-repos-latest.rpm | cpio -idmv`, `cat ./etc/yum.repos.d/katello.repo | grep candlepin` <- this means that the user automatically gets the correct repo for Candlepin.

For containerized Foreman+Katello: there's a container image for Candlepin that should/must provide the correct version of Candlepin.

-> I think that we can safely drop the attribute: we don't use the attribute right now and we don't have to because it's taken care of by the Yum repos or `foremanctl`.

maximiliankolb (approved):
LGTM but I would like an additional ACK in case I missed something.

ekohl (approved):
Looks like it's been unused since d3595a3bfa3203923ef2ab3354fa21ae48c41092.

PR #5169, opened 2026-08-10
Fix external usergroup refresh interval

adamruzicka:
#### What changes are you introducing?
Changing the interval at which `ldap:refresh_usergroups` rake task is expected to run.

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)
`ldap:refresh_usergroups` rake task used to be run by cron every 30 minutes. In https://github.com/theforeman/foreman/pull/10784 and https://github.com/theforeman/foreman-packaging/pull/13001 it was changed to run every hour.

#### Contributor checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing guidelines](https://docs.theforeman.org/nightly/Contributing/index-katello.html#_contributing_to_foreman_documentation).

Please cherry-pick my commits into:

* [x] Foreman 3.19/Katello 4.21
* [ ] Foreman 3.18/Katello 4.20 (Satellite 6.19; orcharhino 7.9)
* [ ] Foreman 3.17/Katello 4.19
* [ ] Foreman 3.16/Katello 4.18 (Satellite 6.18; orcharhino 7.6, 7.7, and 7.8)
* [ ] Foreman 3.15/Katello 4.17
* [ ] Foreman 3.14/Katello 4.16 (Satellite 6.17; orcharhino 7.4; orcharhino 7.5)
* We do not accept PRs for Foreman older than 3.14.

PR #5171, opened 2026-08-10
Display manual procedures for external usergroup refresh in containerized deployments

adamruzicka:
Display manual procedures for external usergroup refresh in containerized deployments.

The procedures are currently hidden, even though both should be available and functional.

Eventually, we'd like the CLI procedure to go away, but as of now, it is there and should work. 

Keeping as a draft until I double check that those do actually work.

#### Contributor checklists

### DIFF
--- a/guides/common/assembly_managing-external-user-groups.adoc
+++ b/guides/common/assembly_managing-external-user-groups.adoc
@@ -4,6 +4,6 @@ include::modules/proc_associating-external-users-with-user-groups.adoc[leveloffs
 
 ifndef::containerized[]
 include::modules/proc_refreshing-external-user-groups-for-ldap-using-cli.adoc[leveloffset=+1]
+endif::[]
 
 include::modules/proc_refreshing-external-user-groups-for-ldap-using-web-ui.adoc[leveloffset=+1]
-endif::[]
--- a/guides/common/assembly_managing-external-user-groups.adoc
+++ b/guides/common/assembly_managing-external-user-groups.adoc
@@ -3,6 +3,9 @@ include::modules/con_managing-external-user-groups.adoc[]
 include::modules/proc_associating-external-users-with-user-groups.adoc[leveloffset=+1]
 
 ifndef::containerized[]
+// While this is available and seems to work, we want to drop this in favor of
+// the other mechanisms (periodic sync through systemd timers, manual refresh using
+// WebUI). Keep things hidden rather than displaying this and then removing it again.
 include::modules/proc_refreshing-external-user-groups-for-ldap-using-cli.adoc[leveloffset=+1]
 endif::[]
 


adamruzicka:
Checked on a containerized nightly instance, both procedures work.

aneta-petrova (approved):

aneta-petrova:
I think simple PRs like this don't even need `tech review` when submitted by an engineer > `style review done` should be enough.

adamruzicka:
> Eventually, we'd like the CLI procedure to go away, but as of now, it is there and should work.

@ekohl since we want to drop it eventually, would you be fine with documenting it while it's there and while it works?

ekohl:
That's https://theforeman-foreman-documentation-preview-pr-5171.surge.sh/nightly/Configuring_User_Authentication/index-containerized-katello.html right? Given we have a UI procedure for it, I'd lean to leaving it out already. At first I thought we had no other way.

I'm also reading this section:
> To set the LDAP source to synchronize user group membership automatically on user login, in the Auth Source page, select the Usergroup Sync option. If this option is not selected, LDAP user groups are refreshed automatically through a scheduled cron job synchronizing the LDAP Authentication source every 30 minutes by default.

We don't have a 30 minute schedule, just hourly: https://github.com/theforeman/foreman/blob/f7e9a6cd0869f76b438a9a9d591eb30eee41489a/lib/tasks/cron.rake#L5. We also migrated both RPM and Debian packaging so no need for conditionals.

adamruzicka:
> That's https://theforeman-foreman-documentation-preview-pr-5171.surge.sh/nightly/Configuring_User_Authentication/index-containerized-katello.html right? Given we have a UI procedure for it, I'd lean to leaving it out already. At first I thought we had no other way.

Ok, will drop.

> We don't have a 30 minute schedule, just hourly:

That's being handled in https://github.com/theforeman/foreman-documentation/pull/5169

guides/common/assembly_managing-external-user-groups.adoc
```diff
@@ -3,7 +3,10 @@ include::modules/con_managing-external-user-groups.adoc[]
 include::modules/proc_associating-external-users-with-user-groups.adoc[leveloffset=+1]
 
 ifndef::containerized[]
+// While this is available and seems to work, we want to drop this in favor of
```
aneta-petrova, line 6: Hm, I'm wondering if this comment should rather be included in the module itself. @maximiliankolb do you have any opinions?

ekohl: I also wouldn't mind if we started to drop it from our entire docs and only support the UI procedure.

maximiliankolb: For the community, I think it would be OK to just show this as admonition. However, I also like the comment which is generally easier to find than git logs. I am OK as is.

IMO this just needs a decision. If Ewoud makes this call, I am very much OK with that. I am unsure who else to ask/I do not know who currently maintains this.

aneta-petrova: > For the community, I think it would be OK to just show this as admonition. However, I also like the comment which is generally easier to find than git logs. I am OK as is.

Rather than an admonition, perhaps we should pair this removal up with a release note, announcing this as removed functionality? Or removed support for a functionality, to be more precise...

@ekohl How does this sound to you as the product owner? We drop the entire procedure, keep only the UI solution documented, and write a release note about it?

adamruzicka: Could we get this in as-is? It shows the supported procedure, leaving the other one to be dealt with later?

aneta-petrova:
> I think simple PRs like this don't even need `tech review` when submitted by an engineer > `style review done` should be enough.

I'll apologize for making this PR more complicated now but because of the discussion in https://github.com/theforeman/foreman-documentation/pull/5171#discussion_r3758606802, it looks like @ekohl's ack or nack should work as `tech review` after all. Just so we're absolutely sure we're doing things by the book.

guides/common/assembly_managing-external-user-groups.adoc
ekohl: > [@ekohl](https://github.com/ekohl) How does this sound to you as the product owner? We drop the entire procedure, keep only the UI solution documented, and write a release note about it?

Do we need a release note to drop the CLI procedure? Users can still execute the required flow (force refresh the LDAP groups) using the UI.

AFAIK we're also planning to write more general "this is what you need to know when working on containerized" and I expect a note there that `foreman-rake` is being phased out.

aneta-petrova: Requests like "this was documented before, now it isn't, please bring the docs back" do come up sometimes. If you think they are unlikely in this case, then no, we don't need a release note.

The release note idea was a reaction to Maximilian's suggestion to turn Adam's comment into a user-facing admonition. That made me think that if we want this change documented, release notes might be a better fit than an admonition that says "this isn't available anymore".

ekohl: On the one hand, it's always good to cover yourself. On the other, I suspect 99% of our users won't care about this difference and it makes the release notes a book that nobody reads. Here I'm slightly leaning to the latter case and not document it as a release note.

Now I just wonder how long it'll take before someone comes to quote me on this with the 1% case.

ekohl:
Short term this is OK, but I'd prefer to drop the CLI procedure for all flavors, not just containerized builds. Long term we're heading to all builds being containerized and it keeps things simpler.

aneta-petrova:

836b3ea3c6..ac6712df7c  5.0 -> 5.0

PR #5182, opened 2026-08-11
Ensure IMAGES_DIR exists when building images

ekohl:

In a high concurrency it may happen that the copy happens before the target is created. This declares the dependency which tells make to order it correctly.

https://github.com/theforeman/foreman-documentation/pull/5105 reported the problem and I think this is a better fix because it allows parallel builds, instead of reverting to sequential builds.

This is hard to test because it's effectively a race condition. I'd suggest to merge this and observe if the problem still occurs.

#### Contributor checklists

### DIFF
--- a/guides/common/Makefile
+++ b/guides/common/Makefile
@@ -39,7 +39,7 @@ endif
 endif
 
 .SHELLFLAGS = -o pipefail -c
-.PHONY = all prepare html clean browser server linkchecker
+.PHONY: all prepare html clean browser server linkchecker
 
 all: html
 
--- a/guides/common/Makefile
+++ b/guides/common/Makefile
@@ -7,7 +7,6 @@ ROOTDIR = $(realpath .)
 NAME = $(notdir $(ROOTDIR))
 DEST_DIR = $(BUILD_DIR)/$(subst doc-,,$(NAME))
 DEST_HTML = $(DEST_DIR)/index-$(BUILD).html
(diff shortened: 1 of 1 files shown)


ekohl:
This is in the base:
```
cp -rf -lL ./images/* ../build/Managing_Configurations_Salt/images
cp -rf -lL ./images/* ../build/Managing_Configurations_Salt/images
cp: cannot create directory '../build/Managing_Configurations_Salt/images/common': File exists
make[2]: *** [../common/Makefile:74: ../build/Managing_Configurations_Salt/.timestamp-images] Error 1
```

Here it tries to copy the images for 2 builds. The issue may be deeper than I thought.

ekohl:
I've restructured the approach a bit:

* Unrelated fix to correctly set `.PHONY`
* Ensure the parent directory exists
* Use `cp` with `--update` and `--target-directory` to make it behave more like `rsync`

I think that should help with any concurrency issues.

aneta-petrova (approved):
Even though I don't have the knowledge to properly review this, I'm willing to give it a try and an ack :) If we merge and the failures mentioned in https://github.com/theforeman/foreman-documentation/pull/5105 still occur, we'll know that additional investigation is needed.

maximiliankolb (approved):
I tested the Makefile locally: There's no regression for my container-based local setup. So 👍 from me.

PR #5195, opened 2026-08-13
Remove the module on content flow

aneta-petrova:

Removing a module on content flow

While reviewing which parts of the Planning guide should be containerized (and what needs updating), this module turned out to be a candidate for removal. The concepts of content management and content flow are sufficiently described in Managing Content and Deployment Scenarios. As long as the Planning guide includes a link to Managing Content in a logical place, we shouldn't need this module.

N/A

#### Contributor checklists

* [ ] Foreman 5.0/Katello 5.0


aneta-petrova:
A quick content gap analysis (with Claude) to find out what we're losing by dropping the module (because it isn't documented anywhere else in the docs set) showed that a list of supported external content sources is now missing. We're already tracking https://redhat.atlassian.net/browse/SAT-48256 to provide a list of supported content types so I'm hoping that would cover this gap.

ekohl (commented):
Since I questioned the usefulness in between meetings, I think I should share my thoughts here too.

I don't think the chapter does a very good job of describing the actual flows and partially overlaps with https://docs.theforeman.org/nightly/Planning_for_Project/index-katello.html#common-deployment-scenarios.

On the other hand, [managing content](https://docs.theforeman.org/nightly/Managing_Content/index-katello.html) just dives right in. No introduction of concepts, no overview of flows. In fact, I'm having a hard time finding out how I sync content to a Smart Proxy at all.

But I'll also emphasize that I'm not an expert on content. These are just thoughts.

aneta-petrova:
> I don't think the chapter does a very good job of describing the actual flows and partially overlaps with https://docs.theforeman.org/nightly/Planning_for_Project/index-katello.html#common-deployment-scenarios.
> 
> On the other hand, [managing content](https://docs.theforeman.org/nightly/Managing_Content/index-katello.html) just dives right in. No introduction of concepts, no overview of flows. In fact, I'm having a hard time finding out how I sync content to a Smart Proxy at all.
> 
> But I'll also emphasize that I'm not an expert on content. These are just thoughts.

They are good thoughts. So we're looking for a better description of content flow after all, rather than giving up on the section entirely? I could see something like that being a good fit for the `Major Foreman components` assembly: After you explain what the server is, proxy, hosts... you show a diagram of how content flows from one through to the other.

To me, that would make more sense than a standalone section like the one we have now.

ekohl:
> They are good thoughts. So we're looking for a better description of content flow after all, rather than giving up on the section entirely? I could see something like that being a good fit for the `Major Foreman components` assembly: After you explain what the server is, proxy, hosts... you show a diagram of how content flows from one through to the other.

I like that idea since it already touches on some of the ideas. Perhaps a relevant observation: right now the guide is top-down: it starts with the features and then lists how it's built up (architecture). This would change it to bottom up: start with the architecture and then see how you can build features with that. Sadly that's a personal preference and there is no right or wrong way.

Not to introduce too much scope creep, but I'm not a huge fan of the "Major components" title. Feels more like an architecture overview.

aneta-petrova:
> I like that idea since it already touches on some of the ideas. Perhaps a relevant observation: right now the guide is top-down: it starts with the features and then lists how it's built up (architecture). This would change it to bottom up: start with the architecture and then see how you can build features with that. Sadly that's a personal preference and there is no right or wrong way.

I updated https://redhat.atlassian.net/browse/SAT-48055. I'll keep this PR opened and linked to it, but I will switch it to draft until I'm ready to pick the ticket up.

> Not to introduce too much scope creep, but I'm not a huge fan of the "Major components" title. Feels more like an architecture overview.

Indeed, this would be needed if we move the content flow description to "Major components". I added a new, better heading for that chapter to the acceptance criteria in https://redhat.atlassian.net/browse/SAT-47862.

PR #5199, opened 2026-08-14
Clean up deprecation and upgrade warnings

ogajduse:
#### What changes are you introducing?
Remove the headline features, upgrade warnings, and deprecations that
were added before the 5.0 branch point (ef598f1f0a) and are now
captured in the 5.0 release notes, so master starts the next cycle
with the clean template. Matches the pattern from #4841 after the
3.19 branch point.

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)
Foreman 5.0 (post) branching

#### Contributor checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing guidelines](https://docs.theforeman.org/nightly/Contributing/index-katello.html#_contributing_to_foreman_documentation).

Please cherry-pick my commits into:

* [ ] Foreman 5.0/Katello 5.0
* [ ] Foreman 3.19/Katello 4.21
* [ ] Foreman 3.18/Katello 4.20 (Satellite 6.19; orcharhino 7.9)
* [ ] Foreman 3.17/Katello 4.19
* [ ] Foreman 3.16/Katello 4.18 (Satellite 6.18; orcharhino 7.6, 7.7, and 7.8)
* [ ] Foreman 3.15/Katello 4.17
* [ ] Foreman 3.14/Katello 4.16 (Satellite 6.17; orcharhino 7.4; orcharhino 7.5)
* We do not accept PRs for Foreman older than 3.14.

PR #5203, opened 2026-08-17
Fix CSS build race condition and redundant per-BUILD compiles

ogajduse:

Fixes a race condition in the guide build's CSS compilation, which is currently causing occasional broken (unstyled) pages in CI-built previews and, presumably, real deploys.

CI builds all 7 product variants (`foreman-el`, `foreman-deb`, `katello`, `orcharhino`, `satellite`, `containerized-katello`, `containerized-orcharhino`) in parallel via `make -j 3 html`. Every variant's build independently recompiles each guide's `style.css`, but the compiled CSS is identical across variants and all of them write to the same output path — so concurrent variant builds can end up truncating and rewriting that file at the same time. If a page gets rendered in the gap, it embeds a corrupted (in the case I hit, completely empty) stylesheet.

Two changes:
- The Sass compile now writes to a temp file and atomically renames it into place, so a reader can never observe a partially-written file, no matter how many builds race on it.
- CSS is now compiled once per guide, up front, and shared across all 7 variant builds, instead of once per variant. Removes the redundant work as well as the race.

This is the CSS half of the parallel-build race #5105 called out; #5182 already fixed the images half of the same underlying problem.

Found while investigating a reviewer report of a broken-looking PR preview (theforeman/foreman-documentation#5199) — traced it to this race via the CI build logs, and confirmed the fix with two full `make -j 3 html` runs (CI's exact invocation).

Related: #5105, #5182

Race conditions are hard to prove fixed just by watching CI logs. I reproduced the actual bug (the empty stylesheet) and confirmed it stopped happening across repeated full builds, but this is still more of a "merge and keep an eye on it" fix than something I can guarantee is 100% gone.

#### Contributor checklists

* [x] Foreman 5.0/Katello 5.0
* [x] Foreman 3.18/Katello 4.20 (Satellite 6.19; orcharhino 7.9)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ DEST := result
 PORT := 5000
 VERSION_LINKS := 5.0 3.19 3.18 3.17 3.16 3.15 3.14 3.13 3.12 3.11 3.10 3.9 3.8 3.7 3.6 3.5 3.4 3.3 3.2 3.1 3.0 2.5 2.4
 
-.PHONY: all clean html web compile serve prep FORCE toc
+.PHONY: all clean html web compile serve prep FORCE toc css
 
 UNAME = $(shell uname)
 ifeq ($(UNAME), Linux)
@@ -27,7 +27,13 @@ clean:
 
 html: build-foreman-el build-foreman-deb build-containerized-katello build-containerized-orcharhino build-katello build-orcharhino build-satellite
 
-build-%: FORCE prep
+# Built once, up front, and shared by every build-% below: the compiled CSS
+# does not vary by BUILD, and compiling it redundantly from N parallel
+# build-% processes races on the same output file.
+css: prep
+	$(MAKE) -C guides/ css
+
+build-%: FORCE prep css
 	$(MAKE) -C guides/ html BUILD=$*
 
 web: prep
--- a/guides/Makefile
+++ b/guides/Makefile
@@ -3,17 +3,24 @@ SHELL := /bin/bash
 SUBDIRS = $(shell ls -d doc-* | grep -v 'doc-Contributing')
 CONTRIBUTING_DIR = doc-Contributing
 
-.PHONY: all clean html linkchecker linkchecker-tryer serve subdirs $(SUBDIRS) contributing
+.PHONY: all clean html linkchecker linkchecker-tryer serve subdirs $(SUBDIRS) contributing contributing-css css
 
 all: html
 
 html: subdirs contributing
+
+css: subdirs contributing-css
 
 subdirs: $(SUBDIRS)
 
 contributing:
 	@if [ -d "$(CONTRIBUTING_DIR)" ]; then \
 		$(MAKE) -C "$(CONTRIBUTING_DIR)" html; \
+	fi
+
+contributing-css:
+	@if [ -d "$(CONTRIBUTING_DIR)" ]; then \
+		$(MAKE) -C "$(CONTRIBUTING_DIR)" css; \
 	fi
 
 $(SUBDIRS):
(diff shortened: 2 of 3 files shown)


aneta-petrova:
Thanks for raising this. Can you think of someone who could review this change? That was our issue with https://github.com/theforeman/foreman-documentation/pull/5105 -- we don't seem to have anyone directly on this team who'd have the right knowledge to be able to review.

ogajduse:
> Thanks for raising this. Can you think of someone who could review this change? That was our issue with #5105 -- we don't seem to have anyone directly on this team who'd have the right knowledge to be able to review.

I can only think of Ewoud and Adam Lazik. They both know how this project works.

aneta-petrova:
Hi @ekohl @adamlazik1, can one of you please take a look at this fix proposal? See the conversation above, which provides some context.

guides/Makefile
```diff
@@ -3,19 +3,26 @@ SHELL := /bin/bash
...
+
 subdirs: $(SUBDIRS)
 
 contributing:
 	@if [ -d "$(CONTRIBUTING_DIR)" ]; then \
 		$(MAKE) -C "$(CONTRIBUTING_DIR)" html; \
 	fi
 
+contributing-css:
+	@if [ -d "$(CONTRIBUTING_DIR)" ]; then \
```
ekohl, line 22: I wondered why this if existed in the `contributing`. Is that just AI being overly cautious?

Makefile
```diff
@@ -27,7 +27,13 @@ clean:
 
 html: build-foreman-el build-foreman-deb build-containerized-katello build-containerized-orcharhino build-katello build-orcharhino build-satellite
 
-build-%: FORCE prep
+# Built once, up front, and shared by every build-% below: the compiled CSS
+# does not vary by BUILD, and compiling it redundantly from N parallel
+# build-% processes races on the same output file.
+css: prep
+	$(MAKE) -C guides/ css
```
ekohl, line 34: This doesn't exist yet, but in relation to what I wrote in the overall review.
```suggestion
guides-prep: prep
	$(MAKE) -C guides/ prep
```

ekohl (commented):
This reminded me of how I want to simplify the images as well that I came up with while working on https://github.com/theforeman/foreman-documentation/pull/5182. I've just submitted it as a PR (https://github.com/theforeman/foreman-documentation/pull/5205) and perhaps we could take a similar approach?

In the `Makefile` we point asciidoctor to a stylesheet per guide:
https://github.com/theforeman/foreman-documentation/blob/31325670df8b379b30a9964addb5ffa6aa28a58a/guides/common/Makefile#L82
Perhaps a better solution is a common `BUILD_DIR` prep that creates `images` and `css`. Then individual builds can layer the HTML files on top.

Effectively create a `prep` goal in `guides/Makefile` to create `guides/build` with images and CSS.

guides/Makefile
ogajduse: This guard mirrors the existing `contributing` and `clean` targets' guard. The Contributing guide was added recently, and it exists only in the `master` branch and in `5.0`, which branched from it last week. If we were to cherry-pick this patch to `3.19` and below, `make` would hit `No such file or directory`.

ogajduse: Basically a backport safety

ogajduse:
This PR doesn't change the build layout; it just adds a `css:` step that compiles CSS once (atomically) before the 7 variant builds run, so it shouldn't get in the way of whatever `guides-prep`/shared `BUILD_DIR` approach #5205 ends up using. I'm happy to either rebase this on #5205 once it's merged, or merge CSS into a shared prep step now if you'd rather handle it in one go.

ekohl (approved):
I think short term this is good enough. My images PR is unlikely to be cherry picked so merging this first is the correct order.

PR #5238, opened 2026-08-21
Update reducing storage use on Foreman server for foremanctl

jberendz:
Updated rake commands and postgresql access instructions for Section 15 of the Administering Satellite guide.

Administering Satellite's Reducing storage use on Satellite Server section needs updated rake commands and PostgreSQL vacuum access instructions. Sections are updated with the approved commands found in https://github.com/theforeman/foremanctl/blob/master/src/roles/foreman/templates/foreman-rake.j2#L8-L29.

#### Contributor checklists

* [x] Foreman 5.0/Katello 5.0


lfu:
### 1.  `foreman_tasks:cleanup` → `foreman_tasks:cleanup:run` — should be reverted

The `foreman-rake` wrapper deployed by foremanctl ([foreman-rake.j2](https://github.com/theforeman/foremanctl/blob/master/src/roles/foreman/templates/foreman-rake.j2#L8-L29)) has an allowlist of supported tasks. That allowlist contains `foreman_tasks:cleanup`, not `foreman_tasks:cleanup:run`.

`foreman_tasks:cleanup` is an alias for `cleanup:run` (defined in [cleanup.rake line 73](https://github.com/theforeman/foreman-tasks/blob/master/lib/foreman_tasks/tasks/cleanup.rake#L73)), so the original command works correctly. Using `cleanup:run` directly would be rejected by the wrapper with "ERROR: Unsupported action detected."

Verified working:
```
# sudo foreman-rake foreman_tasks:cleanup TASK_SEARCH='label = Actions::Katello::Repository::Sync' STATES='stopped'
```

### 2. PostgreSQL vacuum command — needs correction
The PR changes to:
```
# podman exec postgresql vacuumdb -U foreman -d foreman --analyze
```

Issues:
- `-U foreman` should be `-U postgres` — the PostgreSQL admin user in the container is postgres
- `-d foreman` only vacuums one database — the original `--full --all` covered all databases (foreman, candlepin, pulp)
- `--analyze` updates statistics but does not reclaim disk space — the section is about reclaiming space, so `--full` is needed
- Missing `sudo` — podman runs as `root` in containerized deployments

Verified working:
```
# sudo podman exec postgresql vacuumdb -U postgres --full --all
```
This vacuumed: candlepin, foreman, postgres, pulp, template1.

### 3. Missing containerized equivalents
The PR doesn't update these commands which also differ in containerized:


| Original command |	Containerized equivalent |
| :--- | :--- |
| `satellite-maintain service stop` | `sudo systemctl stop foreman.target` |
| `satellite-maintain service start` | `sudo systemctl start foreman.target` |
| `satellite-maintain service stop --exclude postgresql`	 | `sudo systemctl stop foreman.target` && `sudo systemctl start postgresql.service` |
| `satellite-installer --foreman-plugin-tasks-cron-line "..."`	| Systemd timer: `systemctl cat foreman-recurring@daily.timer` (uses `OnCalendar=`) |
| `satellite-installer --foreman-plugin-tasks-automatic-cleanup false` | `sudo systemctl disable foreman-recurring@daily.timer` |
| Edit `/etc/foreman/plugins/foreman-tasks.yaml` `:rules:`  | File is baked into the container image — edits don't persist across restarts.  Use rake task parameters (`TASK_SEARCH`, `AFTER`, `STATES`) or the UI instead. |
| `hammer task info --id ...`	 | Hammer is not installed by default — requires `foremanctl deploy --add-feature hammer` |

Also, all foreman-rake commands in containerized need the `sudo` prefix since the wrapper requires root.

guides/common/modules/ref_configuring-the-cleaning-unused-tasks-feature.adoc
```diff
@@ -15,20 +15,46 @@ By default, {Project} executes a cron job that cleans tasks every day at 19:45.
...
 # {foreman-installer} --foreman-plugin-tasks-cron-line "00 15 * * *"
 ----
+endif::[]
+ifdef::containerized[]
+* To configure the time at which {Project} runs the cron job, set the Systemd timer to the time you want in OnCalendar format.
+For example, to schedule the cron job to run every day at 15:00, enter the following command:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# systemctl cat foreman-recurring@daily.timer "15:00:00"
```
jberendz: Hi @lfu , two things about this containerized command:
* I'm fairly sure I'm not  calling the time correctly. Would the instructions be something like `systemctl edit foreman-recurring@daily.timer` and set `OnCalendar` to `15:00:00`?
* If the calendar event is set to daily, wouldn't that mean strictly at midnight? If so, is the above step even relevant anymore?

guides/common/modules/proc_deleting-a-task-by-id.adoc
```diff
@@ -6,6 +6,16 @@
 [role="_abstract"]
 You can delete tasks by ID, for example if you have submitted confidential data by mistake.
 
+ifdef::containerized[]
+.Prerequisites
```
Lennonka: Prerequisites shouldn't include instructions or commands.
Unfortunately, we currently don't have a link that we could replace it with.

guides/common/modules/proc_deleting-a-task-by-id.adoc
```diff
@@ -20,11 +30,20 @@ You can delete tasks by ID, for example if you have submitted confidential data
...
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
 # foreman-rake foreman_tasks:cleanup TASK_SEARCH="id=_My_Task_ID_"
 ----
+endif::[]
+ifdef::containerized[]
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# sudo foreman-rake foreman_tasks:cleanup TASK_SEARCH='label = Actions::Katello::Repository::Sync' STATES='stopped'
```
Lennonka: There's a convention that commands prefixed with `#` are run under root. (As opposed to commands prefixed with `$` that run under a non-root user.)

```suggestion
# foreman-rake foreman_tasks:cleanup TASK_SEARCH='label = Actions::Katello::Repository::Sync' STATES='stopped'
```
Please, also fix this in the other places below.

Lennonka (changes requested):

lfu:
✅ Task cleanup — back to `foreman-rake foreman_tasks:cleanup` (allowlist-safe; `cleanup:run` would be rejected).
✅ PostgreSQL vacuum — `sudo podman exec postgresql vacuumdb -U postgres --full --all` confirmed working.
✅ Containerized service commands — `systemctl` equivalents for stop/start and disabling the recurring timer are correct.

**One remaining issue — the schedule-change command is wrong.** The current line:

```
# systemctl cat foreman-recurring@daily.timer "15:00:00"
```
isn't valid — `systemctl cat` only prints the unit file and ignores the trailing argument; it changes nothing. 

By default the daily cleanup timer runs at **midnight** (`OnCalendar=daily` expands to `*-*-* 00:00:00`). If you want it to run at a different time — e.g. to move it into a maintenance window or off-peak period — override the timer's schedule. This is the containerized equivalent of the RPM `--foreman-plugin-tasks-cron-line "00 15 * * *"` option, done as a **systemd drop-in override**. Verified procedure:

```
# systemctl edit foreman-recurring@daily.timer
```
Add in the editor — the empty `OnCalendar=` first clears the inherited `daily` value so ours replaces it:

```
[Timer]
OnCalendar=
OnCalendar=15:00:00
```
Then apply and verify:

```
# systemctl daemon-reload
# systemctl restart foreman-recurring@daily.timer
# systemctl list-timers foreman-recurring@daily.timer
```
I tested this with a target ~2 minutes out: the timer fired and the service ran the cleanup container (`foreman-rake cron:daily`, exit `0/SUCCESS`, log finished `foreman_tasks:cleanup`).

**Note worth adding to the doc**: `foreman-recurring@daily.timer` ships with `RandomizedDelaySec=120`, so the actual run can be up to 2 minutes after the configured time — expected, not a misconfiguration.

guides/common/modules/proc_deleting-a-task-by-id.adoc
jberendz: I am moving it to the containerized's version new step 1

lfu:
Re-reviewed the latest commits — the schedule-change section now matches the procedure I verified on a containerized deployment (timer fired and ran `foreman-rake cron:daily` → `0/SUCCESS`).👍

One optional nit worth adding: `foreman-recurring@daily.timer` ships with `RandomizedDelaySec=120`, so after setting `OnCalendar=15:00:00` the job can actually run up to 2 minutes later. A one-line note would save readers from thinking the schedule is misconfigured — e.g.:

The timer uses `RandomizedDelaySec=120`, so the cleanup may start up to 2 minutes after the configured time.

Not blocking — LGTM either way.

@jeremylenz Final check?

guides/common/modules/proc_deleting-task-records.adoc
maximiliankolb, line 1: The commands are identical, so there's no reason to have them both IMO.

guides/common/modules/proc_deleting-a-task-by-id.adoc
```diff
@@ -20,11 +28,20 @@ You can delete tasks by ID, for example if you have submitted confidential data
...
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
 # foreman-rake foreman_tasks:cleanup TASK_SEARCH="id=_My_Task_ID_"
 ----
+endif::[]
+ifdef::containerized[]
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# foreman-rake foreman_tasks:cleanup TASK_SEARCH='label = Actions::Katello::Repository::Sync' STATES='stopped'
```
maximiliankolb: The search differs between containerized and non-containerized. Is this intentional? It looks off to me.

guides/common/modules/proc_recovering-from-a-full-disk.adoc
```diff
@@ -34,17 +34,35 @@ If you use an untypical file system (other than for example ext3, ext4, or xfs),
...
 [options="nowrap" subs="+quotes,attributes"]
 ----
 # {foreman-maintain} service start
 ----
+endif::[]
+ifdef::containerized[]
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# systemctl start foreman.target
```
maximiliankolb, line 53: You could move line 60 and 64 below 57, so that the code block is as follows:

1. if containerized
2. cmd for containerized
3. endif
4. if not containerized
5. cmd for non-containerized
6. endif

This would mean we only need one code block definition & it's clear that the user instruction is the same; only the commands differ based on build target.

maximiliankolb: Same applies to the example above.

guides/common/modules/proc_deleting-a-task-by-id.adoc
```diff
@@ -7,6 +7,14 @@
 You can delete tasks by ID, for example if you have submitted confidential data by mistake.
 
 .Procedure
+ifdef::containerized[]
+. Install Hammer:
++
+[options="nowrap", subs="+quotes,verbatim,attributes"]
+----
+# foremanctl deploy --add-feature hammer
```
jeremylenz: You will not always need to install Hammer. It's included in the `satellite` flavor downstream, and I am not sure this is the place to document installing it.

guides/common/modules/proc_deleting-a-task-by-id.adoc
jeremylenz: The procedure is called "Deleting a task by ID," but this search would find multiple tasks and not by ID. This doesn't seem correct to me. The existing command above `foreman-rake foreman_tasks:cleanup TASK_SEARCH="id=_My_Task_ID_"` should work for both.

guides/common/modules/proc_deleting-task-records.adoc
```diff
@@ -13,7 +13,15 @@ This helps maintain system performance and keeps your task history organized.
...
+ifndef::containerized[]
 +
 ----
 # foreman-rake foreman_tasks:cleanup TASK_SEARCH='label = Actions::Katello::Repository::Sync' STATES='stopped'
 ----
+endif::[]
+ifdef::containerized[]
++
+----
+# foreman-rake foreman_tasks:cleanup TASK_SEARCH='label = Actions::Katello::Repository::Sync' STATES='stopped'
```
jeremylenz: This is the same as the noncontainerized. No need to have separate ifdefs here.

jeremylenz (changes requested):

guides/common/modules/proc_deleting-a-task-by-id.adoc
jberendz: Doing some digging, I'm surprised this was included in the original request because we already describe how to install Hammer in the most relevant doc: https://docs.theforeman.org/nightly/Hammer_CLI/index-foreman-el.html#installing-standalone-hammer

I think we can go back to having "Hammer is installed" on the prereq and then tell users to refer to the reference link for more info.

guides/common/modules/proc_deleting-a-task-by-id.adoc
```diff
@@ -6,6 +6,10 @@
 [role="_abstract"]
 You can delete tasks by ID, for example if you have submitted confidential data by mistake.
 
+.Prerequisites
+* Hammer is installed.
+For more information, see link:https://docs.theforeman.org/nightly/Hammer_CLI/index-foreman-el.html#installing-standalone-hammer[_Installing Standalone Hammer_].
```
jeremylenz: I think this link is only needed in non-Satellite configurations

guides/common/modules/proc_recovering-from-a-full-disk.adoc
```diff
@@ -33,18 +33,27 @@ endif::[]
 If you use an untypical file system (other than for example ext3, ext4, or xfs), you might need to unmount the file system so that it is not in use.
 In that case, complete the following steps:
 
-. Stop {Project} services:
-+
+. Stop {Project} services:+
```
Lennonka:
```suggestion
. Stop {Project} services:
+
```

guides/common/modules/ref_configuring-the-cleaning-unused-tasks-feature.adoc
```diff
@@ -15,23 +15,83 @@ By default, {Project} executes a cron job that cleans tasks every day at 19:45.
...
 [options="nowrap" subs="+quotes,attributes"]
 ----
 # {foreman-installer} --foreman-plugin-tasks-cron-line "00 15 * * *"
 ----
+endif::[]
+ifdef::containerized[]
+* To configure the time at which {Project} runs the cron job, set the Systemd timer to the time you want in OnCalendar format.
+For example, to schedule the cron job to run every day at 15:00, run the following commands:
+
+. Edit the timer's schedule with a drop-in override:
```
Lennonka: Per IBM Style Guide, we must not anthropomorphize nonliving things, which includes the possessive:
```suggestion
. Edit the timer schedule with a drop-in override:
```

guides/common/modules/proc_deleting-a-task-by-id.adoc
jberendz: Good catch. Fix has been pushed

guides/common/modules/ref_configuring-the-cleaning-unused-tasks-feature.adoc
Lennonka, line 1: Please, remove the blank lines between steps.

guides/common/modules/ref_configuring-the-cleaning-unused-tasks-feature.adoc
```diff
@@ -15,23 +15,83 @@ By default, {Project} executes a cron job that cleans tasks every day at 19:45.
...
 * To configure the time at which {Project} runs the cron job, set the `--foreman-plugin-tasks-cron-line` parameter to the time you want in cron format.
 For example, to schedule the cron job to run every day at 15:00, enter the following command:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
 # {foreman-installer} --foreman-plugin-tasks-cron-line "00 15 * * *"
 ----
+endif::[]
+ifdef::containerized[]
+* To configure the time at which {Project} runs the cron job, set the Systemd timer to the time you want in OnCalendar format.
```
Lennonka: I haven't seen it capitalized anywhere else.
```suggestion
* To configure the time at which {Project} runs the cron job, set the systemd timer to the time you want in OnCalendar format.
```

guides/common/modules/ref_configuring-the-cleaning-unused-tasks-feature.adoc
```diff
@@ -15,23 +15,83 @@ By default, {Project} executes a cron job that cleans tasks every day at 19:45.
...
+[options="nowrap" subs="+quotes,attributes"]
+----
+[Timer]
+OnCalendar=
+OnCalendar=15:00:00
+----
++
+[NOTE]
+====
+The timer uses `RandomizedDelaySec=120`, so the cleanup may start up to 2 minutes after the configured time.
```
Lennonka: Is this a default?

guides/common/modules/ref_configuring-the-cleaning-unused-tasks-feature.adoc
```diff
@@ -15,23 +15,83 @@ By default, {Project} executes a cron job that cleans tasks every day at 19:45.
...
+# systemctl restart foreman-recurring@daily.timer
+# systemctl list-timers foreman-recurring@daily.timer
+----
+
+endif::[]
+ifndef::containerized[]
 * To configure the period after which {Project} deletes the tasks, edit the `:rules:` section in the `/etc/foreman/plugins/foreman-tasks.yaml` file.
+endif::[]
+ifdef::containerized[]
+* To configure the period after which {Project} deletes the tasks, use `foreman-rake` task parameters (`TASK_SEARCH`, `AFTER`, `STATES`, etc.) or the web UI.
```
Lennonka:
```suggestion
* To configure the period after which {Project} deletes the tasks, use `foreman-rake` task parameters, such as `TASK_SEARCH`, `AFTER`, `STATES`, or the {ProjectWebUI}.
```

jeremylenz (approved):

guides/common/modules/proc_deleting-a-task-by-id.adoc
```diff
@@ -6,6 +6,10 @@
 [role="_abstract"]
 You can delete tasks by ID, for example if you have submitted confidential data by mistake.
 
+.Prerequisites
```
Lennonka, line 10: For your consideration: I'm wondering if we can hide this prereq for Satellite when Hammer is enabled in Satellite by default:

```suggestion
ifndef::satellite[]
.Prerequisites
```

guides/common/modules/proc_deleting-a-task-by-id.adoc
Lennonka:
```suggestion
For more information, see {HammerDocURL}installing-standalone-hammer[Installing Standalone Hammer] in _{HammerDocTitle}_.
```

Lennonka (changes requested):

guides/common/modules/ref_configuring-the-cleaning-unused-tasks-feature.adoc
lfu: Yes — it's a default from the base `foreman-recurring@.timer` unit, not something you configure. You can see it with `systemctl cat foreman-recurring@daily.timer`. Since our drop-in only overrides `OnCalendar`, `systemd` keeps the base unit's `RandomizedDelaySec=120`, so the note stays relevant. If we'd rather not hard-code "2 minutes" (it's a packaged default that could change), we can word it as "a randomized delay, so cleanup may start shortly after the configured time" and reference `systemctl cat` for the exact value.

Lennonka: Well, but the user might potentially override the packaged value, couldn't they? I was wondering about adding this:
```suggestion
The timer uses `RandomizedDelaySec=120` by default, so the cleanup may start up to 2 minutes after the configured time.
```
But I like your idea about a more generic wording. It would be easier to maintain.

For @jberendz's consideration.

jberendz: Thanks for the considerations @Lennonka and @lfu . While it is easier to maintain referring it to a randomized delay, if the package changes, this line will need to be changed either way.  I think in that context, emphasizing the default makes the most sense because we are also rephrasing the step without quoting back the variables they just read/is the most direct.

guides/common/modules/proc_deleting-a-task-by-id.adoc
lfu: How about the containerized path? @jeremylenz 
```
ifdef::containerized[]
.Prerequisites
* Hammer is enabled. To enable it, run `{foremanctl} deploy --add-feature hammer`.
endif::[]
ifndef::satellite,containerized[]
.Prerequisites
* Hammer is installed.
For more information, see {HammerDocURL}installing-standalone-hammer[Installing standalone Hammer] in _{HammerDocTitle}_.
endif::[]
```

guides/common/modules/proc_deleting-a-task-by-id.adoc
```diff
@@ -6,6 +6,12 @@
 [role="_abstract"]
 You can delete tasks by ID, for example if you have submitted confidential data by mistake.
 
+ifndef::satellite[]
+.Prerequisites
+* Hammer is installed.
+For more information, see {HammerDocURL}installing-standalone-hammer[Installing Standalone Hammer] in _{HammerDocTitle}_.
+endif::[]
```
jeremylenz, line 13: Hammer can be installed on the Foreman base system, or on an external host. I think that link is about installing it "standalone" on an external host.

For containerized non-Satellite, Lucy's suggestion in the resolved comment looks good.
For containerized Satellite, we may not even need a prereq since hammer is included by default.

lfu:
🔧 **proc_deleting-a-task-by-id.adoc (lines 9–13) — Hammer prerequisite**
The prereq points containerized non-Satellite readers to Installing standalone Hammer, which is the external-host install. 
For containerized deployments Hammer is enabled on the server via `{foremanctl} deploy --add-feature hammer`. 
Satellite ships Hammer by default, so it needs no prereq (and since Satellite is a containerized build, the containerized bullet must be gated with `ifndef::satellite`):

```
ifndef::satellite[]
.Prerequisites
ifdef::containerized[]
* Hammer is enabled. To enable it, run `{foremanctl} deploy --add-feature hammer`.
endif::[]
ifndef::containerized[]
* Hammer is installed.
For more information, see {HammerDocURL}installing-standalone-hammer[Installing standalone Hammer] in _{HammerDocTitle}_.
endif::[]
endif::[]
```
Coverage: 

-   Satellite (containerized+satellite) → no prereq; 
-   containerized katello/orcharhino → `foremanctl deploy --add-feature hammer`; 
-   non-containerized non-Satellite → external-host standalone link (unchanged).

✏️ **Nits**

- Same file: link text `"Installing Standalone Hammer"` → `"Installing standalone Hammer"` to match the actual section title (sentence case). (Folded into the block above.)
- `ref_configuring-the-cleaning-unused-tasks-feature.adoc` (line ~13): you changed "thirty days" → "30 days" (good), but the next bullet still reads "older than a year" — consider "older than 1 year" for numeral consistency.

guides/common/modules/proc_deleting-a-task-by-id.adoc
jberendz: Hi @lfu, sorry — I closed this thread before your reply, so I didn't see it.

This makes sense to me, but I have a couple of comments before committing any more changes.

1. Is Hammer included in non-containerized builds? It’s default on Satellite and the non-containerized Foreman docs indicate that it’s included (no steps for installing, only using), but I just want to be sure since we are including Hammer instructions on non-Satellite.
2. We can't put the enable command in this section, because we already moved it out of prereqs (Prerequisite sections should not contain instructions — see https://github.com/theforeman/foreman-documentation/pull/5238#discussion_r3840654093). If the "Using the Hammer CLI tool" section is the wrong document because its scope is installing a standalone instance on an external host (based on https://github.com/theforeman/foreman-documentation/pull/5238#discussion_r3873094484), then we need to put the command somewhere. Lena's comment I linked to originally also confirms that "we do not have a place for this content yet."

The only place it's even mentioned is `proc_configuring-hammer-cli-to-accept-freeipa-credentials.adoc`, and that's close, but not what we're telling the user in this procedure:
> Updating Hammer configuration manually is not required on systems that have been configured with foremanctl.
> Running {foremanctl} deploy --add-feature hammer updates the Hammer configuration as necessary.

  If the goal is to say:
* If you are on Satellite... [no prereqs, already done]
* If you are on a containerized build that isn't Satellite, enable Hammer.
* If you are not on a containerized build [Hammer should already be installed and enabled in the prereqs]

  I would suggest the following:

```
.Procedure
ifdef::containerized[]
ifndef::satellite[]
. Enable Hammer:
+
----
# {foremanctl} deploy --add-feature hammer
----
endif::[]
endif::[]

```
Then, because enabling Hammer is out of scope of this procedure, we open an issue to update whatever relevant document needs the enable instructions. Once it's in the right place, in a later update we remove it from here and replace it with prerequisites, for example:

```
 ifdef::containerized[]
  ifndef::satellite[]
  .Prerequisites
  * Hammer is enabled. 
For more information, see {HammerDocURL}[link]
  endif::[]
  endif::[]
  ifndef::satellite,containerized[]
  .Prerequisites
  * Hammer is installed and enabled.
For more information, see {HammerDocURL}[link].
  endif::[]
```

guides/common/modules/proc_deleting-a-task-by-id.adoc
```diff
@@ -6,6 +6,12 @@
 [role="_abstract"]
 You can delete tasks by ID, for example if you have submitted confidential data by mistake.
 
+ifndef::satellite[]
+.Prerequisites
+* Hammer is installed.
+For more information, see {HammerDocURL}installing-standalone-hammer[Installing Standalone Hammer] in _{HammerDocTitle}_.
```
maximiliankolb:
```suggestion
For more information, see {HammerDocURL}installing-standalone-hammer[Installing standalone Hammer] in _{HammerDocTitle}_.
```

guides/common/modules/proc_deleting-a-task-by-id.adoc
lfu: @jberendz Comment https://github.com/theforeman/foreman-documentation/pull/5238#issuecomment-5441825761 should have addressed this.

jberendz: @lfu  I understand the prerequisites table you listed, but we can't include instructions (run `{foremanctl} deploy --add-feature hammer`) in the prerequisites, because those would go with instructions.

lfu: @jberendz I think you got it right. Hammer is installed by default only in Satellite, not in non-satellite system. @jeremylenz Correct me if I'm wrong.

jeremylenz: This is correct. Hammer is in the Satellite flavor, so it's installed by default on Satellite server and is non-default in other configs.

jberendz:
> 🔧 **proc_deleting-a-task-by-id.adoc (lines 9–13) — Hammer prerequisite** The prereq points containerized non-Satellite readers to Installing standalone Hammer, which is the external-host install. For containerized deployments Hammer is enabled on the server via `{foremanctl} deploy --add-feature hammer`. Satellite ships Hammer by default, so it needs no prereq (and since Satellite is a containerized build, the containerized bullet must be gated with `ifndef::satellite`):
> 
> ```
> ifndef::satellite[]
> .Prerequisites
> ifdef::containerized[]
> * Hammer is enabled. To enable it, run `{foremanctl} deploy --add-feature hammer`.
> endif::[]
> ifndef::containerized[]
> * Hammer is installed.
> For more information, see {HammerDocURL}installing-standalone-hammer[Installing standalone Hammer] in _{HammerDocTitle}_.
> endif::[]
> endif::[]
> ```
> 
> Coverage:
> 
>     * Satellite (containerized+satellite) → no prereq;
> 
>     * containerized katello/orcharhino → `foremanctl deploy --add-feature hammer`;
> 
>     * non-containerized non-Satellite → external-host standalone link (unchanged).
> 
> 
> ✏️ **Nits**
> 
>     * Same file: link text `"Installing Standalone Hammer"` → `"Installing standalone Hammer"` to match the actual section title (sentence case). (Folded into the block above.)
> 
>     * `ref_configuring-the-cleaning-unused-tasks-feature.adoc` (line ~13): you changed "thirty days" → "30 days" (good), but the next bullet still reads "older than a year" — consider "older than 1 year" for numeral consistency.

Everything has been accounted for. The steps to enable Hammer have been moved as the new step 1 for non-Satellite containerized builds because of the style rules surround prereqs. Please let me know if there are any more requested changes before the technical approval.

jberendz:
For the style review: it appears Vale is failing because Hammer is a proper noun but must be lowercase for docIDs,  xrefs, (edit: and a code block). It has been checked by Lena and a rule update may get introduced to fix that.

guides/common/modules/proc_deleting-a-task-by-id.adoc
```diff
@@ -6,7 +6,23 @@
...
+ifndef::satellite,containerized[]
+.Prerequisites
+* Hammer is installed.
+For more information, see {HammerDocURL}installing-standalone-hammer[Installing standalone Hammer] in _{HammerDocTitle}_.
+endif::[]
+
 .Procedure
+ifdef::containerized[]
+ifndef::satellite[]
+. Enable Hammer:
```
Lennonka, line 24: If we're enabling Hammer on ProjectServer, it should be the second step after connecting to ProjectServer :)

guides/common/modules/ref_configuring-the-cleaning-unused-tasks-feature.adoc
```diff
@@ -10,28 +10,87 @@ You can configure the automatic cleaning process according to your needs.
...
 * To configure the time at which {Project} runs the cron job, set the `--foreman-plugin-tasks-cron-line` parameter to the time you want in cron format.
 For example, to schedule the cron job to run every day at 15:00, enter the following command:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
 # {foreman-installer} --foreman-plugin-tasks-cron-line "00 15 * * *"
 ----
+endif::[]
+ifdef::containerized[]
+* To configure the time at which {Project} runs the cron job, you can set the systemd timer to the time you want in OnCalendar format.
```
Lennonka: Article?
```suggestion
* To configure the time at which {Project} runs the cron job, you can set the systemd timer to the time you want in the OnCalendar format.
```

guides/common/modules/ref_configuring-the-cleaning-unused-tasks-feature.adoc
```diff
@@ -10,28 +10,87 @@ You can configure the automatic cleaning process according to your needs.
...
+For example, to schedule the cron job to run every day at 15:00, complete the following steps:
++
+--
+. Edit the timer schedule with a drop-in override:
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# systemctl edit foreman-recurring@daily.timer
+----
+. Replace the inherited `daily` values:
```
Lennonka: Nitpick.
When I run the command above, the drop-in override looks like this:
```
...
# [Timer]
# Unit=foreman-recurring@daily.service
# OnCalendar=daily
# Persistent=true
# RandomizedDelaySec=120
# AccuracySec=1min
...
```
So perhaps singular?
```suggestion
. Replace the inherited `daily` value:
```
Or?

```suggestion
. Replace the inherited values:
```

Lennonka (changes requested):
A few suggestions.

guides/common/modules/ref_configuring-the-cleaning-unused-tasks-feature.adoc
jeremylenz: no article needed here :) It's like saying "in PDF format"

guides/common/modules/ref_configuring-the-cleaning-unused-tasks-feature.adoc
jberendz: I think the first one is more conceptually accurate. Thanks for the suggestions Lena

guides/common/modules/proc_deleting-a-task-by-id.adoc
jberendz: Yeah...

lfu:
Re-reviewed the latest branch — the containerized commands and the schedule-change section (including the `RandomizedDelaySec` note) all look correct and match what I verified on a live containerized deployment. One issue remains.

**The default-schedule sentence is wrong for containerized.** This line in `ref_configuring-the-cleaning-unused-tasks-feature.adoc` is currently unconditional, so containerized readers see it too:
```
By default, {Project} executes a cron job that cleans tasks every day at 19:45.
```
On a containerized deployment the cleanup runs via the `foreman-recurring@daily.timer` systemd unit, whose `OnCalendar=daily` resolves to **midnight (00:00)**, not `19:45`. I confirmed this both on a live system (`systemctl list-timers` showed the daily timer's next run at ~00:01) and in the source: the task is registered by cadence only — `Foreman::Cron.register(:daily, 'foreman_tasks:cleanup')` in foreman-tasks — and `Foreman::Cron` in foreman core carries no time-of-day. The `19:45` value comes from the RPM installer's `--foreman-plugin-tasks-cron-line` default and doesn't apply to the containerized timer.

Suggested fix — conditionalize the sentence:
```
ifndef::containerized[]
By default, {Project} executes a cron job that cleans tasks every day at 19:45.
endif::[]
ifdef::containerized[]
By default, {Project} runs a systemd timer that cleans tasks every day at midnight (`OnCalendar=daily`).
endif::[]
```
Everything else in the PR looks good to me.

guides/common/modules/ref_configuring-the-cleaning-unused-tasks-feature.adoc
```diff
@@ -7,31 +7,94 @@
...
 * To configure the time at which {Project} runs the cron job, set the `--foreman-plugin-tasks-cron-line` parameter to the time you want in cron format.
 For example, to schedule the cron job to run every day at 15:00, enter the following command:
 +
 [options="nowrap" subs="+quotes,attributes"]
 ----
 # {foreman-installer} --foreman-plugin-tasks-cron-line "00 15 * * *"
 ----
+endif::[]
+ifdef::containerized[]
+* To configure the time at which {Project} runs the cron job, you can set the systemd timer to the time you want in the OnCalendar format.
```
maximiliankolb, line 32:
```suggestion
* To configure the time at which {Project} runs the systemd timer, define it in the `OnCalendar` format.
```

Maybe we can shorten it?

guides/common/modules/ref_configuring-the-cleaning-unused-tasks-feature.adoc
```diff
@@ -7,31 +7,94 @@
...
+For example, to schedule the cron job to run every day at 15:00, complete the following steps:
```
maximiliankolb, line 33:
```suggestion
For example, to schedule the systemd timer to run every day at 15:00, complete the following steps:
```

See line 14.

guides/common/modules/ref_configuring-the-cleaning-unused-tasks-feature.adoc
```diff
@@ -7,31 +7,94 @@
...
 ----
-* To reenable regular task cleanup on {Project}, enter the following command:
+endif::[]
+ifdef::containerized[]
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# systemctl disable foreman-recurring@daily.timer
+----
+endif::[]
```
maximiliankolb, line 85:
```suggestion
+
[options="nowrap" subs="+quotes,attributes"]
----
ifdef::containerized[]
# systemctl disable foreman-recurring@daily.timer
endif::[]
ifndef::containerized[]
# {foreman-installer} --foreman-plugin-tasks-automatic-cleanup false
endif::[]
----
```

You could combine those two commands and use conditional inside the code block. IMO this works well if the introduction/step of the command is identical, which it it in line 71.

Same applies to line 86: You could merge the two code blocks.

maximiliankolb (commented):
My suggestions on using "systemd timer" over cron job need a tech ACK; I based my comments on the changes to the top part of the module.

lfu:
"systemd timer" is correct for containerized. There's no crontab on that path: `foreman-recurring@daily.timer` (`OnCalendar=daily`) triggers a Quadlet-generated service running `foreman-rake cron:daily`. 
Verified live — `systemctl list-timers` shows the timer and the run logs finished `foreman_tasks:cleanup` (`0/SUCCESS`). 
'cron' only remains in internal code identifiers (the `cron:daily` rake task and foreman's cadence registry), not as an actual cron job.
Keeping "cron job" in the RPM branch and "systemd timer" for containerized is right. 

@jeremylenz @ekohl @chris1984 Not sure who is the best person to confirm this :smile:

guides/common/modules/ref_configuring-the-cleaning-unused-tasks-feature.adoc
ekohl, line 32: This is IMHO separate functionality: the previous foreman-tasks cleanup cron is a very limited scope functionality where the daily timer is for all jobs that need to run daily. The abstract would need to be different so IMHO this should be its own module.

And zooming out: is this even something we want to document as supported? My suggestion is to just hide the entire module in a containerized build.

lfu: @ekohl Could you point us to the team that owns this piece of information? I think this doc needs the care from the team who can make the best and final decision.

ekohl: Sorry, let me be firmer: please don't include this as a supported procedure.

guides/common/modules/proc_deleting-a-task-by-id.adoc
```diff
@@ -6,13 +6,29 @@
 [role="_abstract"]
 You can delete tasks by ID, for example if you have submitted confidential data by mistake.
 
+ifndef::satellite,containerized[]
+.Prerequisites
+* Hammer is installed.
```
ekohl, line 11: This doesn't align with the procedure explicitly enabling Hammer. Either it's a prerequisite or you enable it explicitly.

guides/common/modules/proc_deleting-a-task-by-id.adoc
```diff
@@ -6,13 +6,29 @@
...
 .Procedure
 . Connect to your {ProjectServer} using SSH:
 +
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
 # ssh root@{foreman-example-com}
 ----
+ifdef::containerized[]
+ifndef::satellite[]
+. Enable Hammer:
```
ekohl, line 24: Technically it's optional since you only use it to optionally view the task and then for the optional verification.

guides/common/modules/proc_deleting-a-task-by-id.adoc
ekohl, line 42: I think this prints how many tasks it has deleted. If that number is more than 0, shouldn't we consider it done? Why do we still need Hammer after?

@adamruzicka could you advise? Would we need to pass `VERBOSE=true` for that to work?

ekohl, line 36: Perhaps suggest noop mode to show what would be done?
```suggestion
$ foreman-rake foreman_tasks:cleanup TASK_SEARCH="id=_My_Task_ID_" NOOP=true
```

ekohl (changes requested):

guides/common/modules/proc_deleting-a-task-by-id.adoc
adamruzicka: > I think this prints how many tasks it has deleted.

It does. It is sort of hidden in the output, but it's there.

> Would we need to pass VERBOSE=true for that to work?

VERBOSE=true just makes it extra chatty by having it print progress messages as it goes, but the "recap" at the end is the same.

```
# foreman-rake foreman_tasks:cleanup TASK_SEARCH=id=c595c08c-811a-4c30-9393-1b92524d2fb4
2026-09-16T08:14:17 [I|app|] Rails cache backend: Redis
2026-09-16T08:14:21 [W|app|] You are trying to replace import_subnets from . Adding allowed actions from plugin permissions to the existing one.
2026-09-16T08:14:22 [W|app|] You are trying to replace view_hosts from foreman_ansible. Adding allowed actions from plugin permissions to the existing one.
2026-09-16T08:14:22 [W|app|] You are trying to replace view_hostgroups from foreman_ansible. Adding allowed actions from plugin permissions to the existing one.
2026-09-16T08:14:22 [W|app|] You are trying to replace edit_hosts from foreman_ansible. Adding allowed actions from plugin permissions to the existing one.
2026-09-16T08:14:22 [W|app|] You are trying to replace edit_hostgroups from foreman_ansible. Adding allowed actions from plugin permissions to the existing one.
2026-09-16T08:14:23 [W|app|] You are trying to replace view_smart_proxies from katello. Adding allowed actions from plugin permissions to the existing one.
2026-09-16T08:14:23 [I|app|] Registered app metadata for plugin katello
2026-09-16T08:14:23 [I|app|] Registered app metadata for plugin foreman_rh_cloud
2026-09-16T08:14:27 [I|for|] Running foreman-tasks cleaner with options {:after=>"0s", :verbose=>false, :batch_size=>1000, :noop=>false, :states=>["stopped"], :backup_dir=>"/var/lib/foreman/tasks-backup/20260916", :filter=>"id=c595c08c-811a-4c30-9393-1b92524d2fb4"}
About to remove 1 tasks matching filter
2026-09-16T08:14:27 [I|for|] About to remove 1 tasks matching filter
W, [2026-09-16T08:14:27.360078 #623]  WARN -- : Scoped order is ignored, it's forced to be batch order.
Deleted 1 tasks matching filter
2026-09-16T08:14:27 [I|for|] Deleted 1 tasks matching filter
No orphaned task locks found, skipping.
2026-09-16T08:14:27 [I|for|] No orphaned task locks found, skipping.
No orphaned task links found, skipping.
2026-09-16T08:14:27 [I|for|] No orphaned task links found, skipping.
No orphaned execution plans found, skipping.
2026-09-16T08:14:27 [I|for|] No orphaned execution plans found, skipping.
No orphaned job invocations found, skipping.
2026-09-16T08:14:27 [I|for|] No orphaned job invocations found, skipping.
2026-09-16T08:14:27 [I|dyn|] start terminating throttle_limiter...
2026-09-16T08:14:27 [I|dyn|] start terminating client dispatcher...
2026-09-16T08:14:27 [I|dyn|] stop listening for new events...
2026-09-16T08:14:27 [I|dyn|] start terminating clock...
```

With verbose
```
# foreman-rake foreman_tasks:cleanup TASK_SEARCH=id=34a79290-1b7b-4e23-b950-597f117bd807 VERBOSE=1
2026-09-16T08:15:11 [I|app|] Rails cache backend: Redis
2026-09-16T08:15:15 [W|app|] You are trying to replace import_subnets from . Adding allowed actions from plugin permissions to the existing one.
2026-09-16T08:15:16 [W|app|] You are trying to replace view_hosts from foreman_ansible. Adding allowed actions from plugin permissions to the existing one.
2026-09-16T08:15:16 [W|app|] You are trying to replace view_hostgroups from foreman_ansible. Adding allowed actions from plugin permissions to the existing one.
2026-09-16T08:15:16 [W|app|] You are trying to replace edit_hosts from foreman_ansible. Adding allowed actions from plugin permissions to the existing one.
2026-09-16T08:15:16 [W|app|] You are trying to replace edit_hostgroups from foreman_ansible. Adding allowed actions from plugin permissions to the existing one.
2026-09-16T08:15:17 [W|app|] You are trying to replace view_smart_proxies from katello. Adding allowed actions from plugin permissions to the existing one.
2026-09-16T08:15:17 [I|app|] Registered app metadata for plugin katello
2026-09-16T08:15:17 [I|app|] Registered app metadata for plugin foreman_rh_cloud
2026-09-16T08:15:21 [I|for|] Running foreman-tasks cleaner with options {:after=>"0s", :verbose=>true, :batch_size=>1000, :noop=>false, :states=>["stopped"], :backup_dir=>"/var/lib/foreman/tasks-backup/20260916", :filter=>"id=34a79290-1b7b-4e23-b950-597f117bd807"}
About to remove 1 tasks matching filter
2026-09-16T08:15:21 [I|for|] About to remove 1 tasks matching filter
0/1
W, [2026-09-16T08:15:21.708372 #665]  WARN -- : Scoped order is ignored, it's forced to be batch order.
1/1
Deleted 1 tasks matching filter
2026-09-16T08:15:21 [I|for|] Deleted 1 tasks matching filter
No orphaned task locks found, skipping.
2026-09-16T08:15:21 [I|for|] No orphaned task locks found, skipping.
About to remove 1 orphaned task links
2026-09-16T08:15:21 [I|for|] About to remove 1 orphaned task links
0/1
1/1
Deleted 1 orphaned task links
2026-09-16T08:15:21 [I|for|] Deleted 1 orphaned task links
No orphaned execution plans found, skipping.
2026-09-16T08:15:21 [I|for|] No orphaned execution plans found, skipping.
No orphaned job invocations found, skipping.
2026-09-16T08:15:21 [I|for|] No orphaned job invocations found, skipping.
2026-09-16T08:15:21 [I|dyn|] start terminating throttle_limiter...
2026-09-16T08:15:21 [I|dyn|] start terminating client dispatcher...
2026-09-16T08:15:21 [I|dyn|] stop listening for new events...
2026-09-16T08:15:21 [I|dyn|] start terminating clock...
```

ekohl: Hmm, that output isn't that great even without verbose. So verification can be to look for `Deleted 1 tasks matching filter` in the output?

adamruzicka: Pretty much. Some light scripting around it would go a long way, but that is not exactly great either.

ekohl: I've heard requests about deleting tasks. Is that something we should eventually expose in the UI/API instead of a rake task?

adamruzicka: I would still be worried about it becoming the default debugging strategy for many, but probably yes

PR #5095, opened 2026-07-22
Update logging for containerization

jsvarova:
What changes are you introducing?
Adding a chapter on containerized logging. 

Why are you introducing these changes? (Explanation, links to references, issues, etc.)

Containerized Satellite 6.20 replaces the traditional file-based logging model with a journald-centric architecture. This is a breaking change: log file paths that administrators have relied on for years (/var/log/foreman, /var/log/candlepin, /var/log/httpd) no longer exist in containerized deployments. Without updated documentation, administrators cannot view service logs, configure log levels for troubleshooting, collect diagnostic data for support cases, or manage log retention to prevent disk exhaustion. The parent epic SAT-43762 (priority: Blocker) explicitly requires that administrators can perform all of these tasks. Every existing script, monitoring integration, and mental model built around traditional log paths must be updated.

- https://github.com/theforeman/foremanctl/issues/21
- https://github.com/theforeman/foremanctl/issues/69
- https://github.com/theforeman/foremanctl/issues/49

#### Contributor checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

PR #5246, opened 2026-08-26
Update containerized variant of Troubleshooting in Managing Hosts

jberendz:
Updated the command to view logs based on the most recent demo: https://youtu.be/010dkpNnQDw?si=VfWd_TpYXV42lp47&t=679

Troubleshooting Red Hat Lightspeed in Satellite has outdated instructions for containerized release.

#### Contributor checklists

* [x] Foreman 5.0/Katello 5.0

### DIFF
--- a/guides/common/modules/ref_troubleshooting-insights-iop.adoc
+++ b/guides/common/modules/ref_troubleshooting-insights-iop.adoc
@@ -6,4 +6,30 @@
 [role="_abstract"]
 You can check the logs to troubleshoot issues with {insights-iop}.
 
+ifndef::containerized[]
 To troubleshoot hosts monitoring, you can find the logs in the `/var/log/foreman/production.log` file.
+endif::[]
+
+ifdef::containerized[]
+To troubleshoot hosts monitoring, you can check the logs by running: 
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# journalctl -u foreman.service
+---- 
+endif::[]
+
+To view all available logs, you can list all services by running: 
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# systemctl list-units foreman*
+---- 
+ifdef::containerized[]
+To manually trigger a VMaas response scan, you can run: 
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+#  systemctl start iop-service-vuln-vmaas-sync.service
+---- 
+endif::[]


guides/common/modules/ref_troubleshooting-insights-iop.adoc
```diff
@@ -6,4 +6,29 @@
...
+[options="nowrap" subs="+quotes,attributes"]
+----
+# journalctl -u foreman.service
+---- 
+
+To view all available logs, you can list all services by running: 
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+# systemctl list-units foreman*
```
jberendz, line 26: @jeremylenz would this command capture the other request of the bug ticket to show the other iop services since this is supposed to "show all services"? If not, I can add each of these in:

`podman logs iop-core-kafka`

`podman logs iop-service-vmaas-reposcan`

`podman logs iop-service-advisor-backend-service`

`podman logs iop-service-vuln-manager`

guides/common/modules/ref_troubleshooting-insights-iop.adoc
```diff
@@ -6,4 +6,30 @@
...
+---- 
+ifdef::containerized[]
+To manually trigger a VMaas response scan, you can run: 
++
+[options="nowrap" subs="+quotes,attributes"]
+----
+#  systemctl start iop-service-vuln-vmaas-sync.service
+---- 
+endif::[]
```
jeremylenz, line 35: I don't think `systemctl list-units foreman*` is what we want here. That doesn't list anything for me. perhaps `systemctl list-dependencies foreman.target` ?

I also think this section would benefit from a notion of "which logs to check for which parts of IoP," e.g. for Advisor you can check `podman logs iop-service-advisor-backend-service`; for Vuln you can check `podman logs iop-service-vuln-manager` and `podman logs iop-service-vmaas-reposcan`; for insights-client registration you can check `podman logs iop-core-kafka` and I think there's another one about inventory. Let's ask @vkrizan what he thinks the best strategy is to document troubleshooting IoP.

guides/common/modules/ref_troubleshooting-insights-iop.adoc
ekohl: In #5095 we already added a procedure for viewing logs of services (https://github.com/theforeman/foreman-documentation/blob/master/guides/common/modules/proc_viewing-service-logs.adoc). I think you can just point to that section because IoP shouldn't be too special.

vkrizan: Be sure to quote the `"foreman*"` otherwise if there is something named foreman in the working directory it would be used as a param for the systemctl command.

Regarding iop, `systemctl list-units "iop*"` can be used. For logs, I'd suggest preferring journalctl with `-u`.

guides/common/modules/ref_troubleshooting-insights-iop.adoc
vkrizan: For logs, I generally would suggest using journalctl. `podman logs` can be used as well, however the journal is mapping it to the unit names rather than container names (those are the same for the most part, however it is easier to capture the oneshot/timed runs with journalctl than with podman).

So for iop logs
```
# all logs
journalctl -u "iop*"

# following logs
journalctl -u "iop*" -f 

# getting only errors
journalctl -u "iop*" -g "err"
```

PR #5229, opened 2026-08-20
Update disconnected update procedure for containerization

aneta-petrova:
#### What changes are you introducing?

Adding a procedure for updating a disconnected and containerized server.

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

https://github.com/theforeman/foremanctl/pull/621

https://redhat.atlassian.net/browse/SAT-48361

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)

N/A

#### Contributor checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing guidelines](https://docs.theforeman.org/nightly/Contributing/index-katello.html#_contributing_to_foreman_documentation).

Please cherry-pick my commits into:

* [x] Foreman 5.0/Katello 5.0
* [ ] Foreman 3.19/Katello 4.21
* [ ] Foreman 3.18/Katello 4.20 (Satellite 6.19; orcharhino 7.9)
* [ ] Foreman 3.17/Katello 4.19
* [ ] Foreman 3.16/Katello 4.18 (Satellite 6.18; orcharhino 7.6, 7.7, and 7.8)
* [ ] Foreman 3.15/Katello 4.17
* [ ] Foreman 3.14/Katello 4.16 (Satellite 6.17; orcharhino 7.4; orcharhino 7.5)
* We do not accept PRs for Foreman older than 3.14.

PR #4929, opened 2026-06-12
Make tuning containerized

jafiala:
#### What changes are you introducing?
Add tuning section to containerized installer docs

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)
https://redhat.atlassian.net/browse/SAT-41407

#### Anything else to add? (Considerations, potential downsides, alternative solutions you have explored, etc.)
Because there is no basic or custom hiera in containerized, con_how-predefined-tuning-profiles-apply is excluded.

#### Contributor checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [x] I am familiar with the [contributing](https://github.com/theforeman/foreman-documentation/blob/master/CONTRIBUTING.md) guidelines.

Please cherry-pick my commits into:

* [x] Foreman 3.19/Katello 4.21
* [x] Foreman 3.18/Katello 4.20 (Satellite 6.19)
* [ ] Foreman 3.17/Katello 4.19
* [ ] Foreman 3.16/Katello 4.18 (Satellite 6.18; orcharhino 7.6, 7.7, and 7.8)
* [ ] Foreman 3.15/Katello 4.17
* [ ] Foreman 3.14/Katello 4.16 (Satellite 6.17; orcharhino 7.4; orcharhino 7.5)
* [ ] Foreman 3.13/Katello 4.15 (EL9 only)
* [ ] Foreman 3.12/Katello 4.14 (Satellite 6.16; orcharhino 7.2 on EL9 only; orcharhino 7.3)
* We do not accept PRs for Foreman older than 3.12.

PR #5265, opened 2026-09-03
Enable disconnected containerized

ShimShtein:
#### Contributor checklists

* [ ] I am familiar with the [contributing guidelines](https://docs.theforeman.org/nightly/Contributing/index-katello.html#_contributing_to_foreman_documentation).

* [x] Foreman 5.0/Katello 5.0


guides/common/assembly_installing-satellite-server-disconnected.adoc
```diff
@@ -2,18 +2,19 @@ ifdef::context[:parent-context: {context}]
...
 
 include::modules/proc_configuring-the-base-operating-system-with-offline-repositories.adoc[leveloffset=+1]
 
 include::modules/proc_installing-from-the-offline-repositories.adoc[leveloffset=+1]
 
-include::modules/proc_resolving-package-dependency-errors.adoc[leveloffset=+1]
-
+ifndef::containerized[]
 include::modules/con_considerations-for-using-the-foreman-installer-utility.adoc[leveloffset=+1]
+endif::[]
```
jafiala: I think we can delete it. No point keeping it in 5.0+, no? Or is there some possibility that the upstream disconnected will be non-containerized?

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -38,24 +36,20 @@ ifdef::containerized[]
 include::common/modules/proc_tuning-performance-with-predefined-profiles.adoc[leveloffset=+1]
 endif::[]
 
-ifndef::containerized[]
-include::common/assembly_tuning-with-predefined-profiles.adoc[leveloffset=+1]
 endif::[]
 
 endif::[]
```
jafiala: Move this endif to L71 to fix the build error.

In your current state, the additional config is displayed for non-satellite builds, and because they include xrefs to hidden content, the xrefs are orphans and the build (correctly) breaks.

```suggestion
```

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -74,10 +68,7 @@ include::common/modules/proc_reverting-server-to-download-content-from-red-hat-c
 
 [appendix]
 include::common/modules/proc_reverting-server-to-download-content-from-red-hat-cdn-by-using-cli.adoc[leveloffset=+1]
-endif::[]
-endif::[]
```
jafiala, line 59:
```suggestion
endif::[]
```

jafiala (changes requested):

guides/common/modules/proc_installing-from-the-offline-repositories.adoc
```diff
@@ -40,13 +40,20 @@ You can install the {Project} packages from the offline repositories by using lo
...
 Install is complete.
-Please run {installer-scenario}.
+Please run {foremanctl} deploy.
+----
+
+. Start the installation process:
++
+[options="nowrap"]
+----
+# {foremanctl} deploy
```
ShimShtein: I have added this step here, since the next thing the doc is talking about is Satellite configuration using hammer (meaning it assumes Sat is already installed at that point). On the other hand the whole certificates handling is way below this line. IoP installation is also mentioned later. 
I think what we need to have is a "flowchart" that will tell the user which flags the user needs to add to the final `foremanctl deploy` line. Otherwise the user will run `foremanctl deploy` multiple time during the installation and while it is not harmful, it still feels a bit weird.

ShimShtein:
Now the linkchecker fails for  `https://docs.tuxcare.com/eportal/` and `https://docs.redhat.com/en/documentation/red_hat_lightspeed/` which does not make too much sense.
Anyway I want to focus more on the doc structure and see if it's good enough, or do we want to make more changes to the disconnected guide

guides/common/assembly_installing-satellite-server-disconnected.adoc
maximiliankolb: Since that hasn't happened until now, I agree with Jan. If there are ever docs for disconnected upstream, it'll probably be containerized only.

guides/common/modules/proc_installing-from-the-offline-repositories.adoc
```diff
@@ -40,13 +40,20 @@ You can install the {Project} packages from the offline repositories by using lo
...
 [options="nowrap" subs="attributes"]
 ----
 Install is complete.
-Please run {installer-scenario}.
+Please run {foremanctl} deploy.
+----
+
+. Start the installation process:
++
+[options="nowrap"]
```
maximiliankolb:
```suggestion
[options="nowrap" subs="attributes"]
```

You have allow the code block to use attributes. I copied this from line 48. Currently, it's not resolved:

Image

guides/common/modules/proc_installing-from-the-offline-repositories.adoc
jafiala: I'm not sure that's necessary if users can run it multiple times and the flags are additive.

I did add "To extend the default deployment, you can select one or more features or add one or more configuration options." in https://docs.theforeman.org/nightly/Installing_Server/index-satellite.html#deploy-satellite-server-configuration-with-the-default-database, which is an attempt to address this at least a little.

ShimShtein: That's a huge refactor that we should probably leave to another PR. Let's keep it as is for the meantime.

guides/common/modules/proc_installing-from-the-offline-repositories.adoc
ShimShtein: fixed

guides/common/modules/proc_installing-from-the-offline-repositories.adoc
ShimShtein: > I'm not sure that's necessary if users can run it multiple times and the flags are additive.

Maybe I misunderstood you: if you are talking about necessity to specify it here - I think we do need it. If the user does not want to perform any additional tasks like certs or IoP, this is the only mention that they have to run this command.

jafiala: > That's a huge refactor that we should probably leave to another PR
Absolutely.

> if you are talking about necessity to specify it here
No, I think making it clear here as well is definitely needed.

Do I understand it correctly that this deploy command gets some parameters from the prepare-system script? I'm thinking if we need to make it more explicit, so that perhaps users don't override some of those. Also, are there any params they need to add to this deploy?

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -31,31 +29,27 @@ include::common/assembly_planning-project-server-installation.adoc[leveloffset=+
...
 
-ifndef::containerized[]
-include::common/assembly_tuning-with-predefined-profiles.adoc[leveloffset=+1]
-endif::[]
-
 endif::[]
 
-ifndef::containerized[]
-
+ifdef::satellite[]
```
ekohl: Can I please ask you to keep the scope smaller and keep additional configuration for a separate PR? I think this PR will be complex enough without it and being able to do a basic installation is something I'd like to have sooner rather than later.

If we have that in place, the IoP team can verify their procedures for disconnected as well.

guides/scripts/find_unused_modules
```diff
@@ -131,7 +131,8 @@ non_containerized_satellite_content = [
   'guides/common/assembly_monitoring-hosts-by-using-insights-in-project.adoc',
   'guides/common/assembly_updating-disconnected-satellite.adoc',
   'guides/common/assembly_upgrading-project-disconnected.adoc',
-  'guides/common/assembly_usage-metrics-collection-in-project.adoc'
+  'guides/common/assembly_usage-metrics-collection-in-project.adoc',
+  'guides/common/modules/proc_configuring-project-server.adoc'
```
ekohl: Trailing commas help to keep the diff smaller
```suggestion
  'guides/common/modules/proc_configuring-project-server.adoc',
```

guides/common/modules/proc_installing-from-the-offline-repositories.adoc
ShimShtein: Nope. the `satellitectl deploy` command may get some additional parameters for optional features: for example custom certifiactes will add some flags, or enabling IoP will add `--add-feature iop`. 
The prepare script by itself does not add any parameters.

jafiala: This should be addressed by including the two "deploy" procedures as we discussed in a call.

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -44,40 +47,43 @@ endif::[]
...
 
-include::common/modules/proc_configuring-server-to-consume-content-from-a-custom-cdn-by-using-web-ui.adoc[leveloffset=+2]
+include::common/assembly_importing-the-project-client-name-repository.adoc[leveloffset=+2]
 
-include::common/modules/proc_configuring-server-to-consume-content-from-a-custom-cdn-by-using-cli.adoc[leveloffset=+2]
+ifndef::containerized[]
+ifdef::foreman-el,foreman-deb[]
+include::common/modules/proc_enabling-remote-execution.adoc[leveloffset=+2]
+endif::[]
+endif::[]
```
jafiala: I think we can just delete this. There should be no non-containerized foreman build of the disconnected guide.


```suggestion
```

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -44,40 +47,43 @@ endif::[]
...
 
+ifdef::katello,orcharhino,satellite[]
 include::common/modules/proc_configuring-pull-based-transport-for-remote-execution.adoc[leveloffset=+2]
+endif::[]
```
jafiala: I think this condition also is unnecessary. No disconnected katello, and orcharhino should just have the initial note with the external link.


```suggestion
include::common/modules/proc_configuring-pull-based-transport-for-remote-execution.adoc[leveloffset=+2]
```

jafiala (changes requested):

jafiala (approved):

guides/common/modules/proc_deploy-project-server-configuration-with-an-external-database.adoc
```diff
@@ -21,6 +21,8 @@ For more information, see {InstallingServerDocURL}planning-{project-context}-ser
 * Your system is prepared for the installation.
 For more information, see {InstallingServerDocURL}preparing-environment-for-{project-context}-server-installation_{project-context}[Preparing environment for {ProjectServer} installation].
 
+ifndef::installing-satellite-server-disconnected[]
```
maximiliankolb, line 25: As is right now, you'd lose the ".Procedure" heading from line 26 that is required for other cases.

jafiala:
The filename-id-heading match script is broken on master (https://github.com/theforeman/foreman-documentation/pull/5282), not by this PR.

guides/common/modules/proc_deploy-project-server-configuration-with-the-default-database.adoc
```diff
@@ -18,6 +18,8 @@ For more information, see xref:{InstallingServerDocURL}planning-{project-context
 * Your system is prepared for the installation.
 For more information, see xref:{InstallingServerDocURL}preparing-environment-for-{project-context}-server-installation_{project-context}[Preparing environment for {ProjectServer} installation].
 
+ifndef::installing-satellite-server-disconnected[]
```
maximiliankolb, line 22: Same as above.

guides/common/modules/proc_installing-insights-iop-with-the-project-context-iso-image.adoc
```diff
@@ -18,25 +18,15 @@ ifndef::satellite[]
...
-For more information, see https://access.redhat.com/solutions/7055981[How to switch to netavark network backend in Podman in the Red{nbsp}Hat Knowledgebase].
 
 .Procedure
 . Download and mount the {Project} ISO image:
-For more information, see xref:installing-from-the-offline-repositories_{context}[].
-. Set up the local repositories for disconnected environments.
-For more information, see xref:installing-from-the-offline-repositories_{context}[].
-. Set up containers on your {ProjectServer}:
+For more information, see xref:installing-the-{foremanctl}-packages-from-the-offline-repositories[].
+. Set up the local repositories and containers for disconnected environments.
```
maximiliankolb:
```suggestion
. Set up the local repositories and containers for disconnected environments:
```

It's a step in a procedure -> Please use a colon instead of a dot.

guides/common/modules/proc_installing-insights-iop-with-the-project-context-iso-image.adoc
```diff
@@ -52,4 +42,10 @@ $ curl -o cvemap.xml https://security.access.redhat.com/data/meta/v1/cvemap.xml
 # cp cvemap.xml /var/lib/foreman/
 ----
 +
-{Project} detects the file and automatically publishes it to `/var/www/html/pub/iop/data/meta/v1/cvemap.xml`.
+{Project} detects the file and automatically publishes it to `/var/www/html/pub/iop/data/meta/v1/cvemap.xml` after the plugin is enabled.
+. Enable the plugin:
```
maximiliankolb:
```suggestion
. Enable the {Insights} plugin:
```

https://github.com/theforeman/foreman-documentation/blob/master/guides/common/attributes-base.adoc#L174 Unless you think it should say "Insights on Satellite", then "{insights-iop}".

guides/common/assembly_installing-and-configuring-insights-iop.adoc
```diff
@@ -9,5 +9,4 @@ endif::[]
 ifdef::installing-satellite-server-disconnected[]
 include::modules/proc_installing-insights-iop-with-the-project-context-iso-image.adoc[leveloffset=+1]
```
maximiliankolb:
```suggestion
```

guides/common/assembly_installing-server-connected.adoc
```diff
@@ -6,14 +6,6 @@ include::modules/proc_deploy-project-server-configuration-with-the-default-datab
 include::modules/proc_deploy-project-server-configuration-with-an-external-database.adoc[leveloffset=+1]
 endif::[]
 
-ifndef::containerized[]
```
maximiliankolb, line 9: Isn't this still relevant for upstream Foreman+Katello based on `foreman-installer`?

Image

maximiliankolb (changes requested):

guides/common/assembly_installing-server-connected.adoc
jafiala: You're right Max, we should not remove this.

guides/common/modules/proc_installing-insights-iop-with-the-project-context-iso-image.adoc
ShimShtein: Do we really need to be that specific? We are talking about enabling IOP (that's the whole procedure), so we have multiple steps - download some files, do some other preparations, and the next step is to enable the plugin in the foreman server.

maximiliankolb: I am not sure what is best, but my gut feeling is "the Insights plugin". non-blocking suggestion; maybe @jafiala  can suggest what's best

jafiala: I think it's better to be specific.

```suggestion
. Enable the {insights-iop} plugin:
```

maximiliankolb:
Please rebase to HEAD of "master".

ShimShtein:
Should be ready for another round.

guides/common/modules/proc_deploy-project-server-configuration-with-an-external-database.adoc
```diff
@@ -22,6 +22,9 @@ For more information, see {InstallingServerDocURL}planning-{project-context}-ser
 For more information, see {InstallingServerDocURL}preparing-environment-for-{project-context}-server-installation_{project-context}[Preparing environment for {ProjectServer} installation].
 
 .Procedure
+
+ifndef::installing-satellite-server-disconnected[]
+
```
maximiliankolb:
```suggestion
ifndef::installing-satellite-server-disconnected[]
```

Same for `guides/common/modules/proc_deploy-project-server-configuration-with-the-default-database.adoc`.

guides/common/modules/proc_installing-insights-iop-with-the-project-context-iso-image.adoc
```diff
@@ -52,4 +42,10 @@ $ curl -o cvemap.xml https://security.access.redhat.com/data/meta/v1/cvemap.xml
 # cp cvemap.xml /var/lib/foreman/
 ----
 +
-{Project} detects the file and automatically publishes it to `/var/www/html/pub/iop/data/meta/v1/cvemap.xml`.
+{Project} detects the file and automatically publishes it to `/var/www/html/pub/iop/data/meta/v1/cvemap.xml` after the plugin is enabled.
```
maximiliankolb, line 49: Stupid question: Is that `/var/www/html/pub/` on Foreman Server aka. FQDN/pub/ aka. the directory that we're trying to get rid of?

guides/common/modules/proc_upgrading-container-images-of-insights-iop.adoc
```diff
@@ -8,91 +8,16 @@ If you have {insights-iop} enabled on your {ProjectServer}, prepare the containe
 
 .Prerequisites
 * {insights-iop} is enabled on your disconnected {ProjectServer}.
-* You know which method you used for the initial installation of {insights-iop}: the ISO image method or the export and import method.
 
 .Procedure
-* If you used the ISO image method for the initial installation of {insights-iop}, update the {insights-iop} container images from the ISO image for {ProjectVersion}:
+* Update the {insights-iop} container images from the ISO image for {ProjectVersion}:
```
maximiliankolb, line 13: AFAIK, this procedure was previously split into two parts; depending on the source of IoP. Now, this could probably be simplified.

Maybe you can move line 13 in some way to the abstract, and then remove one level of the list structure in line 14,16,18 so that those three lines start with a single dot.

guides/common/assembly_installing-server-connected.adoc
```diff
@@ -9,8 +9,6 @@ endif::[]
 ifndef::containerized[]
 include::modules/con_considerations-for-using-the-foreman-installer-utility.adoc[leveloffset=+1]
 
-include::modules/proc_installing-project-server-packages.adoc[leveloffset=+1]
```
maximiliankolb, line 12: This is still needed for upstream for `foreman-installer`-based installations, isn't it? Those are still supported, aren't they? @ekohl

ekohl: Yes, they are still supported in upstream.

guides/common/modules/proc_installing-insights-iop-with-the-project-context-iso-image.adoc
ShimShtein: Yes, that's the folder on the host. I know, but for now we don't have a good solution for exposing data files on the foreman server. I suppose eventually we will collaborate with katello to store those files in a proper repo.

guides/common/modules/proc_upgrading-container-images-of-insights-iop.adoc
ShimShtein: While it's a valid argument, I suggest leaving  the iop refactoring to a separate effort and PR. I am more interested in having a proper disconnected procedure laid out at least on the high level, even if the details will require deeper refactoring.

ShimShtein:
Fixed everything  have talked about. The filename-id should be fixed by https://github.com/theforeman/foreman-documentation/pull/5282

guides/common/modules/proc_installing-insights-iop-with-the-project-context-iso-image.adoc
```diff
@@ -18,25 +18,15 @@ ifndef::satellite[]
...
 [options="nowrap", subs="+quotes,verbatim,attributes"]
 ----
-# /media/sat6/setup_containers
-----
-. Enable the plugin:
-+
-[options="nowrap", subs="+quotes,verbatim,attributes"]
-----
-# {foreman-installer} --enable-iop
+# /media/sat6/prepare_system
```
ekohl, line 33: Is it needed to run `prepare_system` again? I'd think that all the containers have already been imported from the initial installation.

guides/common/modules/proc_installing-the-foremanctl-packages-from-the-offline-repositories.adoc
```diff
@@ -1,10 +1,10 @@
 :_mod-docs-content-type: PROCEDURE
 
-[id="installing-from-the-offline-repositories_{context}"]
-= Installing the {Project} packages from the offline repositories
+[id="installing-the-{foremanctl}-packages-from-the-offline-repositories"]
+= Installing the {foremanctl} packages from the offline repositories
```
ekohl: What was wrong with the old title? I preferred it because it's generic: it didn't point to the installer and just said that it was to install the packages. We shouldn't be so specific in titles.

guides/scripts/find_unused_modules
ekohl: This is still used in upstream. It shouldn't be here.

```suggestion
  'guides/common/assembly_usage-metrics-collection-in-project.adoc'
```

guides/common/assembly_installing-server-connected.adoc
ekohl: Yes they are still supported upstream.

guides/common/modules/proc_upgrading-container-images-of-insights-iop.adoc
ekohl: That's why I've been suggesting to not touch it at all. Right now you're giving the impression that it's correct but I suspect we should drop this entire file: it's now part of the regular upgrade process.

guides/common/modules/proc_installing-insights-iop-with-the-project-context-iso-image.adoc
ShimShtein: We mention downloading the ISO in the procedure. I suppose we need either both or none of them.
For updating IoP - it will make sence, for clean installs - it would not.

guides/common/modules/proc_installing-the-foremanctl-packages-from-the-offline-repositories.adoc
ShimShtein: IMHO it was misleading - I am not installing Satellite from offline repos/rpms - I install only the `satellitectl` utility. That's why I wanted to be a bit more specific.

guides/scripts/find_unused_modules
ShimShtein: fixed.

guides/common/modules/proc_installing-insights-iop-with-the-project-context-iso-image.adoc
ekohl: Why not for clean installs? I'd expect that an admin prepares the system. Then all images are imported into the local system. From then you can list an installed Satellite as a prerequisite.

guides/common/modules/proc_installing-the-foremanctl-packages-from-the-offline-repositories.adoc
ekohl: You don't only install `satellitectl`, you also install all dependencies. That as a whole makes it `{Project} packages`. IMHO titles shouldn't be too specific because the implementation may change. The intent should stay the same.

@jafiala any opinion/guidance?

guides/common/modules/proc_installing-insights-iop-with-the-project-context-iso-image.adoc
ShimShtein: I may be wrong here, but if I read the docs I can see that the user basically has an option to update only the IoP containers without updating the Satellite rpms (at least that was the option in the foreman-installer world).
For such an update, the user will need to mount a new ISO, run the prepare script and then run the `satellitectl` command. 
For fresh installs, the user mounts the ISO and runs the prepare script before installing the Satellite, so this step won't be needed to be repeated for the IoP.

Maybe we should put a conditional statement there, something like "mount the ISO and run the prepare script if you haven't done it earlier".

guides/common/modules/proc_upgrading-container-images-of-insights-iop.adoc
ShimShtein: You still have the specialized data files treatment. I don't think we can get rid of this file yet.

guides/common/modules/proc_installing-insights-iop-with-the-project-context-iso-image.adoc
ekohl: > I may be wrong here, but if I read the docs I can see that the user basically has an option to update only the IoP containers without updating the Satellite rpms (at least that was the option in the foreman-installer world).

Correct, but that was the foreman-installer world. In a containerized world everything is image-based and thus no longer special.

> For such an update, the user will need to mount a new ISO, run the prepare script and then run the `satellitectl` command.

Yes, but that should update the entire installation, IoP included. https://github.com/theforeman/foreman-documentation/pull/5229 should cover that so in the scope of this PR we can ignore it.

> For fresh installs, the user mounts the ISO and runs the prepare script before installing the Satellite, so this step won't be needed to be repeated for the IoP.

Exactly.

> Maybe we should put a conditional statement there, something like "mount the ISO and run the prepare script if you haven't done it earlier".

This is what I'd like to see. You already have a prerequisite pointing to the procedure where `prepare_system` ran. IMHO you can make the prerequisite stronger and say that that module should have been executed entirely, not just mounted the ISO. Possibly even require a full installation.

guides/common/modules/proc_installing-the-foremanctl-packages-from-the-offline-repositories.adoc
ekohl: Today with a fresh mind.

If we look at the previous chapter, the header was `Installing the {Project} packages from the offline repositories` and the last step was `./install_packages`.

Now the last step is `./prepare_system`. How about:

```suggestion
= Preparing the system for {Project} installation from the offline repositories
```

Or since the next step is called deploy:
```suggestion
= Preparing the system for {Project} deployment from the offline repositories
```

guides/common/assembly_installing-server-connected.adoc
ShimShtein: added it back.

guides/common/modules/proc_installing-insights-iop-with-the-project-context-iso-image.adoc
ShimShtein: If you think we should drop the concept of "I want to update only the IoP containers", then we can simplify this whole procedure to contain only the steps for updating the data files (`cvemap` and maybe others). 
Everything else would be covered in the general update Satellite procedure.

guides/common/assembly_installing-server-connected.adoc
```diff
@@ -8,9 +8,7 @@ endif::[]
 
 ifndef::containerized[]
 include::modules/con_considerations-for-using-the-foreman-installer-utility.adoc[leveloffset=+1]
-
```
maximiliankolb, line 11: empty lines between modules are by design.

guides/common/modules/proc_installing-the-foremanctl-packages-from-the-offline-repositories.adoc
jafiala: Either way works, but your point about intent and implementation is strong.

I'd go with "Preparing the system for {Project} deployment" to tie in with the following procs.

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -44,40 +47,35 @@ endif::[]
...
-[appendix]
-include::common/modules/proc_restoring-manual-changes-overwritten-by-a-puppet-run.adoc[leveloffset=+1]
-
 [appendix]
 include::common/modules/proc_reverting-server-to-download-content-from-red-hat-cdn-by-using-web-ui.adoc[leveloffset=+1]
 
 [appendix]
 include::common/modules/proc_reverting-server-to-download-content-from-red-hat-cdn-by-using-cli.adoc[leveloffset=+1]
-endif::[]
+
```
maximiliankolb:
```suggestion
```

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -31,9 +29,14 @@ include::common/assembly_planning-project-server-installation.adoc[leveloffset=+
 include::common/assembly_preparing-your-environment-for-project-server-installation.adoc[leveloffset=+1]
 
 include::common/assembly_installing-satellite-server-disconnected.adoc[leveloffset=+1]
+endif::[]
 
 ifdef::katello,orcharhino,satellite[]
```
maximiliankolb:
```suggestion

```

For upstream, there should be no content as is right now: https://docs.theforeman.org/nightly/Installing_Server_Disconnected/index-katello.html

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -31,9 +29,14 @@ include::common/assembly_planning-project-server-installation.adoc[leveloffset=+
...
 
+include::common/assembly_configuring-satellite-custom-server-certificate.adoc[leveloffset=+1]
+
+include::common/modules/proc_resetting-custom-ssl-certificate-to-default-self-signed-certificate-on-project-server.adoc[leveloffset=+1]
+
 ifdef::containerized[]
```
maximiliankolb:
```suggestion
```

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -31,9 +29,14 @@ include::common/assembly_planning-project-server-installation.adoc[leveloffset=+
...
 include::common/modules/proc_tuning-performance-with-predefined-profiles.adoc[leveloffset=+1]
 endif::[]
```
maximiliankolb:
```suggestion
```

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -44,40 +47,35 @@ endif::[]
 
 endif::[]
 
-ifndef::containerized[]
-
+ifdef::satellite[]
```
maximiliankolb:
```suggestion
```

maximiliankolb (changes requested):
I left some comments for "master.adoc" because there is content visible on https://theforeman-foreman-documentation-preview-pr-5265.surge.sh/nightly/Installing_Server_Disconnected/index-katello.html.

guides/doc-Installing_Server_Disconnected/master.adoc
ShimShtein: Why? we have:
```
ifdef::containerized[]
include::common/modules/proc_tuning-performance-with-predefined-profiles.adoc[leveloffset=+1]
endif::[]

ifndef::containerized[]
include::common/assembly_tuning-with-predefined-profiles.adoc[leveloffset=+1]
endif::[]
```
So it's either one or the other.
We don't want to show them both I think.

maximiliankolb: disconnected, as is right now on "master", is only filled with content for Satellite. orcharhino has its own downstream-only offline docs: https://docs.orcharhino.com/or/docs/sources/installation_and_maintenance/installing_orcharhino_server_offline.html#Installing_orcharhino_Server_using_ISO_images

upstream does not support this. So IMO the include macros can all assume "containerized Satellite", because we have line 16 & 19 for other cases.

ShimShtein:
Should be ready now

guides/doc-Installing_Server_Disconnected/master.adoc
maximiliankolb, line 1: IMO we could drop line 10-14: This scenarios is not supported by/for the community, as indicated by line 17.

maximiliankolb (commented):
* `guides/common/modules/proc_installing-insights-iop-with-the-project-context-iso-image.adoc`: right now, it's a procedure without ".Procedure" and any steps for Satellite builds due to line 10 and the last line.
* `guides/common/modules/proc_upgrading-container-images-of-insights-iop.adoc`: style-wise, we should probably make minor adjustments to the steps, or, as Ewoud suggests, adjust this procedure in a follow-up PR.
* `guides/doc-Installing_Server_Disconnected/master.adoc`: one tiny comment, but style-wise LGTM. Thanks Shim for the changes; it's now much easier to read.

ShimShtein:
I want to leave the IoP to a proper follow-up. Fixed the small comment on the master.adoc.

guides/common/modules/proc_installing-insights-iop-with-the-project-context-iso-image.adoc
evgeni: I am very much :+1: on not updating IoP on its own. This doesn't match with our support strategy, which is all or nothing.

evgeni:
The built result contains the following chapters which I think is wrong for disconnected?
- 2.3. Configuring the HTTP proxy to connect to Red Hat CDN
- 2.4. Registering to Red Hat Subscription Management
- 2.5. Configuring repositories
- 2.6. Logging in to the Red Hat container registry

https://docs.redhat.com/en/documentation/red_hat_satellite/6.19/html/installing_satellite_server_in_a_disconnected_network_environment/index doesn't show these right now


Besides that the non-IOP disconnected installation workflow looks correct to me.

ShimShtein:
Thanks @evgeni! Gated the chapters you mentioned inside `ifndef::installing-satellite-server-disconnected[]` statement and rebased.
Should be more consistent now.

evgeni:
I think this is not a change from this PR, but in "3.2. Configuring the base operating system with offline repositories" the last steps are "Create a directory to serve as the mount point for the ISO file of Satellite Server." and "Mount the ISO image for Satellite Server to the mount point." and I am rather sure Satellite is not part of the base operating system. ;)

ACK otherwise.

jafiala:
@evgeni I added your point as a separate issue so that we don't lose it and don't hold up the PR.

https://github.com/theforeman/foreman-documentation/issues/5315

guides/common/modules/proc_installing-insights-iop-with-the-project-context-iso-image.adoc
```diff
@@ -6,6 +6,10 @@
 [role="_abstract"]
 You can use the {Project} installation ISO image to access the required container content.
 
+//Gated the content until it's refined
```
ekohl: Please hide the entire module in the assembly.

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -32,50 +24,39 @@ include::common/assembly_preparing-your-environment-for-project-server-installat
...
 
 include::common/modules/con_performing-additional-configuration.adoc[leveloffset=+1]
 
 include::common/assembly_installing-and-configuring-insights-iop.adoc[leveloffset=+2]
 
 include::common/assembly_configuring-inter-server-synchronization.adoc[leveloffset=+2]
 
+include::common/assembly_importing-the-project-client-name-repository.adoc[leveloffset=+2]
+
+include::common/modules/proc_configuring-pull-based-transport-for-remote-execution.adoc[leveloffset=+2]
```
ekohl: We haven't delivered pull mode yet in containerized builds.

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -32,50 +24,39 @@ include::common/assembly_preparing-your-environment-for-project-server-installat
...
-include::common/assembly_tuning-with-predefined-profiles.adoc[leveloffset=+1]
-endif::[]
-
-endif::[]
+include::common/modules/proc_resetting-custom-ssl-certificate-to-default-self-signed-certificate-on-project-server.adoc[leveloffset=+1]
 
-ifndef::containerized[]
+include::common/modules/proc_tuning-performance-with-predefined-profiles.adoc[leveloffset=+1]
 
 include::common/modules/con_performing-additional-configuration.adoc[leveloffset=+1]
```
ekohl: In the initial story I explicitly kept additional config out of scope to reduce the amount of things to review. Right now you're asking the reviewers to test the whole epic at once in a single PR.

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -32,50 +24,39 @@ include::common/assembly_preparing-your-environment-for-project-server-installat
...
 :!numbered:
 
-[appendix]
-include::common/modules/proc_restoring-manual-changes-overwritten-by-a-puppet-run.adoc[leveloffset=+1]
-
 [appendix]
 include::common/modules/proc_reverting-server-to-download-content-from-red-hat-cdn-by-using-web-ui.adoc[leveloffset=+1]
 
 [appendix]
 include::common/modules/proc_reverting-server-to-download-content-from-red-hat-cdn-by-using-cli.adoc[leveloffset=+1]
```
ekohl, line 53: Do we actually want to support this? We've been leaning more into making it a hard choice to go disconnected by also extending `prepare_system` to make sure it's configured as such. I'm leaning to leaving those out and perhaps removing them altogether.

guides/doc-Installing_Server_Disconnected/master.adoc
jafiala: I added tuning for disconnected in https://github.com/theforeman/foreman-documentation/pull/4929. But now I think about it, it couldn't have been tested, right?

Do we need a separate task for testing and enabling tuning of disconnected servers?

ShimShtein: Tuning should be the same for connected and disconnected (unless we define some specific bandwidth requirements, which I highly doubt).

guides/doc-Installing_Server_Disconnected/master.adoc
ShimShtein: We talked about it - this guide can be used for semi-disconnected installs too (where there is another Satellite to pull the content from). I think there is some rationale to leave these procedures available here.
And the prepare_system script (at least for today) stops before running satellitectl command, so I don't really see how do you want to use it to configure an already running Satellite.
Of course we can extend the script to wrap around the satellitectl deploy command, but I don't think that was your intention.

guides/common/modules/proc_installing-insights-iop-with-the-project-context-iso-image.adoc
ShimShtein: I think we agreed with @jafiala that htis was the better option.

guides/doc-Installing_Server_Disconnected/master.adoc
ShimShtein: removed

guides/doc-Installing_Server_Disconnected/master.adoc
ShimShtein: The problem is that I need to to validate the coherence of the whole process, and it's hard to do so without seeing the document laid out fully. 
Otherwise we will have steps that are out of order, or steps that repeat themselves. 
I have encountered this when working on the preparing part, that I wanted to add the `satellitectl deploy` execution to it, but other procedures, like the IoP needed to come before that command.

jafiala: Shim, I don't think that applies entirely for the additional config - those by definition should be optional.

ShimShtein: I am not sure I would like to remove it, especially since we want to have the disconnected and connected guides as close as possible. Ideally they should be one guide with optional steps for the disconnected experience.
Anyway the whole additional configuration chapter is the same as in connected. What would be the benefit of removing a reference to an already reviewed content?

ShimShtein: Removed to make the PR more palatable

guides/common/modules/proc_installing-insights-iop-with-the-project-context-iso-image.adoc
jafiala: We did.

However, I double-checked now, and it appears we can remove the include because the file is listed in 
guides/scripts/find_unused_modules. It is a cleaner solution.

ShimShtein: Removed

guides/doc-Installing_Server_Disconnected/master.adoc
ekohl: More concretely: if we move forward with https://github.com/theforeman/foremanctl/pull/785 then does the chapter still make sense?

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -32,50 +24,35 @@ include::common/assembly_preparing-your-environment-for-project-server-installat
...
-include::common/assembly_tuning-with-predefined-profiles.adoc[leveloffset=+1]
-endif::[]
-
-endif::[]
+include::common/assembly_configuring-satellite-custom-server-certificate.adoc[leveloffset=+1]
 
-ifndef::containerized[]
+include::common/modules/proc_resetting-custom-ssl-certificate-to-default-self-signed-certificate-on-project-server.adoc[leveloffset=+1]
 
-include::common/modules/con_performing-additional-configuration.adoc[leveloffset=+1]
```
Lennonka, line 49: Why are we removing this? This module is a chapter heading.

guides/doc-Installing_Server_Disconnected/master.adoc
ShimShtein: Yes, the air gapped PR does not deal with Foreman object manipulation needed for this type of deployments. It does lay out the base for such actions, and we will be able to add those, but for now we still don't do it.

guides/doc-Installing_Server_Disconnected/master.adoc
jafiala: Shim could you please revert 35b3327f0ee3f5a62560c70af982bcd443d32764 ? Not sure why it happened.

guides/common/modules/con_installing-and-configuring-insights-iop.adoc
```diff
@@ -13,11 +13,7 @@ When you install {insights-iop} locally, you can generate {Insights} recommendat
 Enabling {insights-iop} prevents you from using any {RHCloud} services on hosts registered to {Project}.
 * If you install {Project} with external databases, you cannot enable {insights-iop}.
 Enabling {insights-iop} prevents you from using external databases.
-* Podman must be configured to use the netavark network backend.
```
jafiala: Is this intentionally removed from ALL builds and guides? I see it removed also from orcharhino and connected satellite.

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -32,50 +24,29 @@ include::common/assembly_preparing-your-environment-for-project-server-installat
...
-
-ifdef::containerized[]
-include::common/modules/proc_tuning-performance-with-predefined-profiles.adoc[leveloffset=+1]
-endif::[]
-
-ifndef::containerized[]
-include::common/assembly_tuning-with-predefined-profiles.adoc[leveloffset=+1]
-endif::[]
-
-endif::[]
```
jafiala: Can we just have 

ifndef::containerized[]
include::common/modules/proc_tuning-performance-with-predefined-profiles.adoc[leveloffset=+1]
endif::[]

here? So that when it's verified we know where it goes, and can just revert the guard?

guides/doc-Installing_Server_Disconnected/master.adoc
```diff
@@ -32,50 +24,29 @@ include::common/assembly_preparing-your-environment-for-project-server-installat
...
+include::common/assembly_configuring-satellite-custom-server-certificate.adoc[leveloffset=+1]
 
-ifndef::containerized[]
+include::common/modules/proc_resetting-custom-ssl-certificate-to-default-self-signed-certificate-on-project-server.adoc[leveloffset=+1]
 
 include::common/modules/con_performing-additional-configuration.adoc[leveloffset=+1]
 
-include::common/assembly_installing-and-configuring-insights-iop.adoc[leveloffset=+2]
```
jafiala: Same, would prefer conditionals instead of plain removals.

Alternatively, you should make these commits granular so we can then selectively revert them when verified.

guides/common/modules/con_installing-and-configuring-insights-iop.adoc
ShimShtein: Yes, it's the same for all containers now, so IOP is not an exception.

guides/doc-Installing_Server_Disconnected/master.adoc
ShimShtein: This line should be part of the iop work. Once the IoP is done, it would need to add itself to this guide, otherwise it won't be testable. So I think pure removal is OK here. Otherwise what condition would you put there? something like `if false`? It's the same as commenting lines out, and we don't do it.

guides/doc-Installing_Server_Disconnected/master.adoc
ShimShtein: But the whole guide is under `ifdef satellite`, and we don't intend to publish non-containerized Satellite.
And again the same argument here - I could just comment out the line, but it is considered bad practice.

Again, since we're not done with the tuning, once it gets in, it will need to register itself here in order to test it.

guides/doc-Installing_Server_Disconnected/master.adoc
jafiala: ifndef::containerized

I know effectively it is the same as commenting out, but the logic is - we started containerization with all incudes inside ifndef::containerized and removed those guards for verified content.

So your PR should only MOVE those guards around so that the verified content is shown, but the unverified content is hidden. It's more a process thing than state-of-code thing.

guides/common/modules/con_installing-and-configuring-insights-iop.adoc
jafiala: Also non-containerized orcharhino?

guides/doc-Installing_Server_Disconnected/master.adoc
ShimShtein: Originally we had:
```
ifdef::containerized[]
include::common/modules/proc_tuning-performance-with-predefined-profiles.adoc[leveloffset=+1]
endif::[]

ifndef::containerized[]
include::common/assembly_tuning-with-predefined-profiles.adoc[leveloffset=+1]
endif::[]
```
which one do you want me to re-add?

guides/common/modules/con_installing-and-configuring-insights-iop.adoc
ShimShtein: re-added as conditional. Containerized actually has a built-in check.

PR #5267, opened 2026-09-04
Add weekly gh-pages history squash workflow

lzap:
Adds a scheduled workflow that automatically squashes the gh-pages branch history every week, keeping only the last 3 months (90 days, configurable). This prevents the gh-pages branch from growing unboundedly while retaining recent deployment history.

Considered using `force_orphan: true` in the peaceiris/actions-gh-pages action to nuke history entirely on each deploy, but opted for the weekly squash approach because:

- Preserves recent deployment history for debugging and rollback purposes
- Less aggressive than force_orphan (which would discard all history immediately)
- Configurable threshold
- Runs on a schedule independent of actual deployments
- Runs weekly on Sunday at midnight UTC
- Can be manually triggered via workflow_dispatch from the Actions tab

Context: https://community.theforeman.org/t/proposing-to-squash-gh-pages-for-foreman-documentation/47552

### DIFF
--- a/.github/workflows/squash-gh-pages-history.yml
+++ b/.github/workflows/squash-gh-pages-history.yml
@@ -1,0 +1,63 @@
+---
+name: Squash gh-pages History
+
+on:
+  schedule:
+    - cron: '0 0 * * 0'  # Weekly on Sunday at midnight UTC
+  workflow_dispatch:     # Allow manual triggers
+    inputs:
+      retention_days:
+        description: 'Days of commit history to retain'
+        required: false
+        default: '90'
+        type: string
+
+jobs:
+  squash-history:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: gh-pages
+          fetch-depth: 0  # Full history needed to count commits
+
+      - name: Configure Git
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Squash history retaining last N days
+        env:
+          RETENTION_DAYS: ${{ inputs.retention_days || '90' }}
+        run: |
+          CUTOFF_DATE=$(date -d "$RETENTION_DAYS days ago" --utc +%Y-%m-%dT%H:%M:%S)
+          echo "Retention period: $RETENTION_DAYS days (cutoff: $CUTOFF_DATE)"
+
+          # Count commits within retention window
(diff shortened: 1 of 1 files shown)


.github/workflows/squash-gh-pages-history.yml
```diff
@@ -0,0 +1,63 @@
+---
+name: Squash gh-pages History
+
+on:
+  schedule:
+    - cron: '0 0 * * 0'  # Weekly on Sunday at midnight UTC
```
maximiliankolb, line 6: non-blocking: We could probably also do it at any other time on Sunday. I don't know if this would help GH or make our GHA more reliable. It's probably uncool if _all_ projects run stuff at midnight.

.github/workflows/squash-gh-pages-history.yml
```diff
@@ -0,0 +1,63 @@
...
+  workflow_dispatch:     # Allow manual triggers
```
maximiliankolb, line 7: I assume only maintainers can trigger this?

maximiliankolb (approved):
I did not run all lines from the script locally because I do not actually want to bust my gh-pages branch just yet, but I checked the man pages. IMO we can try this.

I left one comment about the cron line & I'm also wondering if the GHA need any write permissions to allow the (force) push to GH: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions

I am OK trying this and then adjusting it later on.

.github/workflows/squash-gh-pages-history.yml
lzap: They probably already solved this problem and have a queue of items.

.github/workflows/squash-gh-pages-history.yml
lzap: Hard to tell, GitHub Actions...

Whoever merges this, please test this and ping me after this is done so I can validate the result.

maximiliankolb (approved):

lzap:
By "please test this" I meant - run the workflow manually and ping me so I can inspect the result. I do not have permissions to do that :-) Thanks.

maximiliankolb:
Merged and tested:

image

-> https://github.com/theforeman/foreman-documentation/actions/runs/35079532495

I fetched `gh-pages` from GitHub and compared it. There was a diff, but it's limited to a [PR that I merged to `3.18` recently](https://github.com/theforeman/foreman-documentation/pull/5303).

From my understanding, this works as expected. :heavy_check_mark: Thanks!

ping @lzap to confirm that this matches your expectation.

lzap:
Thank you it all looks good.

PR #5279, opened 2026-09-07
Make -j default and serve without full rebuild

lzap:
Building the whole site is extremely slow, `-j` should be the default. It is used already on CICD (`-j3`) and there is no reason not to use it. Passing it via `make -j` does not break your shell history it just works.

Also, pass the `-j` down the processes, specifically this helps a bit for CSS.

Finally, the `serve` target always rebuilds everything. This is a problem since I very often want to only build the upstream nightly guides which is much faster. The patch changes this behavior and updates the README with instructions how to now achieve the required behavior:

	make html serve

Or a tad faster:

	make build-foreman-el serve

Added one more commit - ignores `docs.tuxcare.com/eportal` link which always fails on CICD (probably the linter was blocked now).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 SHELL := /bin/bash
+MAKEFLAGS += -j
 DEST := result
 PORT := 5000
 VERSION_LINKS := 5.0 3.19 3.18 3.17 3.16 3.15 3.14 3.13 3.12 3.11 3.10 3.9 3.8 3.7 3.6 3.5 3.4 3.3 3.2 3.1 3.0 2.5 2.4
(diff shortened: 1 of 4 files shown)


Lennonka (approved):
I'm not an expert, but it looks sane and I trust your expertise :)

Will this also speed up the complete CI/CD builds?

lzap:
> Will this also speed up the complete CI/CD builds?

The newly introduced option for CSS can, but just a tad. Don't get your hopes up.

If we have some slow CICD let me know what exactly and I can take a look. When building all guides (`make html`) the biggest offender is the sheer amount of all combinations and versions. What could help is start removing older versions, we currently have 2.4 - 5.0 and maybe we might drop few very, very old versions.

Edit: The oldest version is from 2021 that is not that old. Perhaps we are good.

maximiliankolb (approved):
It is difficult to me to benchmark this locally, but I tested my own wrapper around make and that still works.

PR #5261, opened 2026-09-03
Use EL 10 & remove the maintenance repository

stejskalleos:
#### What changes are you introducing?

* Remove `RepoRHEL9ServerSatelliteMaintenanceProjectVersion` from the codebase.
* Use EL10

#### Why are you introducing these changes? (Explanation, links to references, issues, etc.)

The maintenance repository will no longer exist.
In the upcoming downstream release, repositories will be merged into two:

* satellite-6.Y-for-rhel-10-x86_64-rpms
* satellite-utils-6.Y-for-rhel-10-x86_64-rpms

https://redhat.atlassian.net/browse/SAT-49920

#### Contributor checklists

* [x] I am okay with my commits getting squashed when you merge this PR.
* [ ] I am familiar with the [contributing guidelines](https://docs.theforeman.org/nightly/Contributing/index-katello.html#_contributing_to_foreman_documentation).

Please cherry-pick my commits into:

* [ ] Foreman 5.0/Katello 5.0

PR #5284, opened 2026-09-08
Turn off Vale's AsciiDocDITA.TaskInclude for upstream

ekohl:

Use Vale's [comments](https://docs.vale.sh/formats/asciidoc#comments) to disable a specific rule for upstream content: `AsciiDocDITA.TaskInclude`.

In https://github.com/theforeman/foreman-documentation/pull/5261 I saw this warning and wonder if it's safe to disable the rule for upstream. This is a test to look at that.

I have no idea if my assumption is valid so this is just a draft and I'm abusing CI to see if it even works.

#### Contributor checklists

* [ ] Foreman 5.0/Katello 5.0

### DIFF
--- a/guides/common/modules/proc_configuring-repositories.adoc
+++ b/guides/common/modules/proc_configuring-repositories.adoc
@@ -26,7 +26,9 @@ ifdef::satellite[]
 ----
 endif::[]
 ifdef::foreman-deb[]
+pass:[<!-- vale AsciiDocDITA.TaskInclude = NO -->]
 include::snip_configuring-repositories-foreman-deb.adoc[]
+pass:[<!-- vale AsciiDocDITA.TaskInclude = YES -->]
 endif::[]
 ifdef::foreman-el,katello[]
 . Clear any metadata:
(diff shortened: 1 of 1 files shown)


aneta-petrova:
I wouldn't mind disabling the rule because we've recently done it for another one, but perhaps like this? https://github.com/theforeman/foreman-documentation/blob/de2e27b41dd902ddccc169e0bdd4070b39c8f7c2/.vale-dita.ini#L17

ekohl:
Isn't that just disabling the rule globally, which may not be desirable for downstream? And looking at the test it's also surprising to me that it works for 1, but not the other.

ekohl:
Hmm, doesn't appear to work.
> Make sure to surround the inline passthrough statements with newlines, as shown below.

This isn't 100% clear to me.  Maybe it really needs the empty lines around it as well?

aneta-petrova:
> Isn't that just disabling the rule globally, which may not be desirable for downstream? And looking at the test it's also surprising to me that it works for 1, but not the other.

Oh right, the respective team runs the tests separately downstream, so I don't think we need to worry about that. The parts committed upstream are really just for upstream.

ekohl:
I have no strong opinion on this. For now I'll close this PR because I think I have my answer: control statements don't work as I expected.

PR #5294, opened 2026-09-09
Add vale and vale-all make targets + instructions

lzap:
TIL about vale, I was lazy to install it on my system so I figure out how to run it via podman. This PR adds

```
make vale
```

which runs vale via podman. It matches configuration on our CICD and checks only staged files or files from the last commits if the checkout is clean. Additionally:

```
make vale-all
```

Runs it against the whole project.

---

The second commit is up for discussion, I am introducing AGENTS file which describes how to ensure quality by always building relevant guides locally and also running the newly introduced vale tool. I do not want do go into AI or not-AI discussion in this PR, so this is specifically very brief instruction file with focus only on what agents should run when they are done with some changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,11 @@
+- Follow the checklists and technical/style-review guidance in
+  [CONTRIBUTING.md](CONTRIBUTING.md).
+- Use [guides/README.md](guides/README.md) and the
+  [Contributors' Guide](guides/doc-Contributing/) for documentation structure
+  and AsciiDoc conventions.
+- Treat [.vale.ini](.vale.ini) and the project [Vale rules](.vale/styles/foreman-documentation/)
+  as the style authority.
+- ALWAYS run `make -C guides/XXX` of all changed guides after changes.
+- Avoid `make html` unless necessary, it is very slow.
+- ALWAYS run `make vale`, and
+  [check-filename-id-heading-match.sh](scripts/check-filename-id-heading-match.sh).


.vale.ini
```diff
@@ -10,10 +10,6 @@ Packages = RedHat, AsciiDoc
 
 Vocab = Foreman
 
-# Use local Vocab terms
-Vale.Terms = YES
-Vale.Avoid = YES
-
```
lzap, line 16: I am under impression these are no longer needed, but prove me wrong and I will drop this.

Makefile
```diff
@@ -22,6 +25,15 @@ prep:
 	cd web && bundle config set path 'vendor' && bundle install
 	mkdir -p $(DEST)/nightly
 
+vale-all:
+	podman run --rm --userns=keep-id -v "$(CURDIR):/work:Z" -w /work --entrypoint /bin/sh $(VALE_IMAGE) -c 'vale sync && vale --config=.vale.ini $(VALE_FLAGS) $(VALE_FILES)'
+
+vale:
+	changed_files="$$(git diff --name-only --diff-filter=ACMR HEAD -- 'guides/common/*.adoc' 'guides/common/modules/*.adoc' ':!guides/common/attributes*.adoc' ':!guides/common/header.adoc' ':!guides/common/ribbons.adoc')"; \
+	if [ -z "$$changed_files" ]; then changed_files="$$(git diff-tree --no-commit-id --name-only --diff-filter=ACMR -r HEAD -- 'guides/common/*.adoc' 'guides/common/modules/*.adoc' ':!guides/common/attributes*.adoc' ':!guides/common/header.adoc' ':!guides/common/ribbons.adoc')"; fi; \
+	if [ -z "$$changed_files" ]; then echo "No changed AsciiDoc files to lint."; exit 0; fi; \
```
lzap, line 33: Little bit clunky, but this mirrors what we have on CICD. Ability to run vale only on changed files is extremely useful. Actually, I have an idea how to put this into a single place.

Edit: Added a commit that unifies how changed files are detected.

lzap:
Intereting, during researching on how changed files for vale validation on CICD works, I found out that for every single check, the whole repo must be cloned (currently over 2.5 GiB). This is done by the changed file GHA automatically.

I replaced this behavior with cloning just shallow copy (depth: 0) and then checking out only the required commits from the PR. Then the script is used to find changed files so the functionality is shared both from the Makefile and CICD.

.vale.ini
Lennonka: Why do you think that? I'm not so sure.

lzap:
Dropped the unwanted `vale.ini` change, rebased.

Makefile
```diff
@@ -22,6 +25,14 @@ prep:
 	cd web && bundle config set path 'vendor' && bundle install
 	mkdir -p $(DEST)/nightly
 
+vale-all:
+	podman run --rm --userns=keep-id -v "$(CURDIR):/work:Z" -w /work --entrypoint /bin/sh $(VALE_IMAGE) -c 'vale sync && vale --config=.vale.ini $(VALE_FLAGS) $(VALE_FILES)'
```
ekohl: The `:Z` flag in the volume only works on SELinux. Not sure how many non-SELinux users we have out there.

Makefile
```diff
@@ -2,9 +2,12 @@ SHELL := /bin/bash
 MAKEFLAGS += -j
 DEST := result
 PORT := 5000
+VALE_IMAGE ?= docker.io/jdkato/vale
```
maximiliankolb, line 5: https://docs.vale.sh/topics/installation#docker ✔️

README.md
```diff
@@ -61,6 +61,8 @@ Stable versions are symlinks to the nightly (current) version, which can cause i
 This repository uses the [Vale](https://vale.sh/) linter.
 The `.vale/styles/` subdirectory includes a project-specific `foreman-documentation` style package with rules to check for Foreman documentation project conventions.
 When adding a new rule to the `foreman-documentation` style, make sure it does not duplicate the other styles included in `.vale/styles/`.
+Run `make vale` to run the CI-style changed-file check in Podman without
+installing Vale locally. Use `make vale-all` to scan all guides.
```
maximiliankolb, line 65:
```suggestion
Run `make vale` to run the CI-style changed-file check in Podman without installing Vale locally.
Use `make vale-all` to scan all guides.
```

Makefile
maximiliankolb, line 29: CURDIR: https://www.gnu.org/software/make/manual/html_node/Recursion.html ✔️

Makefile
```diff
@@ -22,6 +25,14 @@ prep:
 	cd web && bundle config set path 'vendor' && bundle install
 	mkdir -p $(DEST)/nightly
 
+vale-all:
+	podman run --rm --userns=keep-id -v "$(CURDIR):/work:Z" -w /work --entrypoint /bin/sh $(VALE_IMAGE) -c 'vale sync && vale --config=.vale.ini $(VALE_FLAGS) $(VALE_FILES)'
+
+vale:
+	@changed_files="$$(scripts/changed-files.py)"; \
+	if [ -z "$$changed_files" ]; then echo "No changed AsciiDoc files to lint."; exit 0; fi; \
```
maximiliankolb, line 33: Is the exit 0 necessary to not run line 34?

If so, there's another way: "if there are changed files, run line 34; else run the echo command". I am totally fine as is. Your call.

Makefile
```diff
@@ -22,6 +25,14 @@ prep:
 	cd web && bundle config set path 'vendor' && bundle install
 	mkdir -p $(DEST)/nightly
 
+vale-all:
+	podman run --rm --userns=keep-id -v "$(CURDIR):/work:Z" -w /work --entrypoint /bin/sh $(VALE_IMAGE) -c 'vale sync && vale --config=.vale.ini $(VALE_FLAGS) $(VALE_FILES)'
+
+vale:
+	@changed_files="$$(scripts/changed-files.py)"; \
```
maximiliankolb, line 32: I think you could also assign the stdout of the Python script to a Make variable that is evaluted when it's first used.

My minimal reproducer:

````Makefile
blub = $(shell sleep 3)

anything:
    @echo "a"

wait:
    @echo "$(blub)"
````

I think that this proofs that the Python script would only run if you run `vale` which actually uses the variable.

File: AGENTS.md
Comment (maximiliankolb) [inline AGENTS.md:1]: We generally do not split lines at n characters.

**Review by maximiliankolb — COMMENTED** [review]
Thanks lzap, it works well. I left some suggestion.

# QA

````
$ make vale
No changed AsciiDoc files to lint.
$ echo $?
0
$ make vale-all
podman run --rm --userns=keep-id -v "/fake/foreman-documentation:/work:Z" -w /work --entrypoint /bin/sh docker.io/jdkato/vale -c 'vale sync && vale --config=.vale.ini --minAlertLevel=error guides/common'
Trying to pull docker.io/jdkato/vale:latest...
Getting image source signatures
Copying blob 186e7016cfda done   | 
Copying blob c1cacf48f26d done   | 
Copying blob 55afa1ecc21d skipped: already exists  
Copying blob 82a9ac3e57e4 done   | 
Copying config 1e59420e10 done   | 
Writing manifest to image destination
W101 'Vale.Terms' isn't a core option; Vale is ignoring it.
W101 'Vale.Avoid' isn't a core option; Vale is ignoring it.
Syncing AsciiDoc [2/2] ███████████████████████████████████████████████ 100% | 1s
 SUCCESS  Synced 2 package(s) to '/work/.vale/styles'.
W101 'Vale.Terms' isn't a core option; Vale is ignoring it.
W101 'Vale.Avoid' isn't a core option; Vale is ignoring it.

 guides/common/assembly_configuring-an-alternate-cname.adoc
 7:1   error  Abstract must be between 50 and 300 characters (currently 395 characters).  foreman-documentation.AbstractLength 
 9:60  error  Use 'before' rather than 'prior to'.                                        RedHat.TermsErrors                   

...
````

Exit 2 due to the errors and only takes about three seconds locally which is very fast for all modules. 👍

File: Makefile (hunk 6 above)
Comment (lzap) [reply Makefile]: Did you mean to use if-else instead? I changed it, it is slightly more readable at the more characters used :-)

File: Makefile (hunk 7 above)
Comment (lzap) [reply Makefile]: Great call, I had no idea this is evaluated this way. GNU Make surprises me every year at least once :-)

File: AGENTS.md
Comment (lzap) [reply AGENTS.md:1]: I am not following this one. Is this about word wrapping?

**Comment by lzap** [comment]
Updated the `Makefile` as you suggested:

```diff
diff --git a/Makefile b/Makefile
index cb38f6736f..628f046c2c 100644
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ PORT := 5000
 VALE_IMAGE ?= docker.io/jdkato/vale
 VALE_FILES ?= guides/common
 VALE_FLAGS ?= --minAlertLevel=error
+CHANGED_FILES = $(shell scripts/changed-files.py)
 VERSION_LINKS := 5.0 3.19 3.18 3.17 3.16 3.15 3.14 3.13 3.12 3.11 3.10 3.9 3.8 3.7 3.6 3.5 3.4 3.3 3.2 3.1 3.0 2.5 2.4

 .PHONY: all clean html web compile serve prep vale vale-all FORCE toc css
@@ -29,9 +30,11 @@ vale-all:
        podman run --rm --userns=keep-id -v "$(CURDIR):/work:Z" -w /work --entrypoint /bin/sh $(VALE_IMAGE) -c 'vale sync && vale --config=.vale.ini $(VALE_FLAGS) $(VALE_FILES)'

 vale:
-       @changed_files="$$(scripts/changed-files.py)"; \
-       if [ -z "$$changed_files" ]; then echo "No changed AsciiDoc files to lint."; exit 0; fi; \
-       podman run --rm --userns=keep-id -v "$(CURDIR):/work:Z" -w /work --entrypoint /bin/sh $(VALE_IMAGE) -c "vale sync && vale --config=.vale.ini $(VALE_FLAGS) $$changed_files"
+       @if [ -z "$(CHANGED_FILES)" ]; then \
+               echo "No changed AsciiDoc files to lint."; \
+       else \
+               podman run --rm --userns=keep-id -v "$(CURDIR):/work:Z" -w /work --entrypoint /bin/sh $(VALE_IMAGE) -c "vale sync && vale --config=.vale.ini $(VALE_FLAGS) $(CHANGED_FILES)"; \
+       fi

 clean:
        $(MAKE) -C guides/ clean
diff --git a/README.md b/README.md
index 715ba2e75e..2d6f9c1f05 100644
--- a/README.md
+++ b/README.md
@@ -61,8 +61,8 @@ Stable versions are symlinks to the nightly (current) version, which can cause i
 This repository uses the [Vale](https://vale.sh/) linter.
 The `.vale/styles/` subdirectory includes a project-specific `foreman-documentation` style package with rules to check for Foreman documentation project conventions.
 When adding a new rule to the `foreman-documentation` style, make sure it does not duplicate the other styles included in `.vale/styles/`.
-Run `make vale` to run the CI-style changed-file check in Podman without
-installing Vale locally. Use `make vale-all` to scan all guides.
+Run `make vale` to run the CI-style changed-file check in Podman without installing Vale locally.
+Use `make vale-all` to scan all guides.
```

AGENTS.md
Lennonka: Yes. We typically have one whole sentence per line.

PR #5295, opened 2026-09-09
Rewrite check-filename-id-heading-match in Python

lzap:
I rewrote (Codex) check-filename-id-heading-match script from Bash, which was spawning many, many subprocesses and it was extremely slow, to Python. I tested the output from both is exactly the same. Then I created a simple shell wrapper so CICD is not broken by the change.

The checker improved from approximately 11.4 seconds to 0.06 seconds on my system, this is now instant so we can run it as much as we like.


jafiala:
The performance gain is impressive, but I don't know enough to review it really. How do you propose that we review?

lzap:
> The performance gain is impressive, but I don't know enough to review it really. How do you propose that we review?

You can just test it if the output is expected or it works as intended. This is a helper script, I do not expect too many code comments anyway. One issue is there are no tests for the code to validate correctness for all use cases, I can add that.

lzap:
Amended an integration test with fixtures, ran it against the original script.

Lennonka (approved):
Since you wrote the tests for it, I think we can consider the update tested.
We don't really have the skills to do code review.
I suggest we merge it as-is and if we run into any issues later, we (you) can fix them as we go.

If there are no blocking comments, we can merge tomorrow.

scripts/test/fixtures/not-a-module.txt
maximiliankolb, line 1: Please duplicate this file as `scripts/test/fixtures/snip_must-not-contain-an-id.adoc`.

Snippets must not contain anchors, so the check should not run for it. In generall, only modules named `con_*.adoc`, `proc_*.adoc`, and `ref_*.adoc` require an anchor.

scripts/test/fixtures/proc_project-server.adoc
maximiliankolb, line 1: Unsure if this is really a different test case:

Could you please duplicate this file as `scripts/test/fixtures/proc_smart-proxy-server.adoc` with the anchor "{smart-proxy-context}-server" and "{SmartProxyServer" as title.

IMO it's different because some attributes differ between headings and anchors. We are not yet very consistent with dedicated attributes with a "-id" or "-context" suffix.

scripts/test/fixtures/ref_by-using-cli.adoc
maximiliankolb, line 1: Could you please add test cases for "by-using-web-ui" -> "{ProjectWebUI}" and "by-using-api" -> "by using {Project} API" and "by-using-ansible" -> "by using Ansible".

maximiliankolb (commented):
I left some comments for more possible test cases.

lzap:
Rebased, added one more commit with what you requested.

PR #5296, opened 2026-09-09
Add Planning guide content to Maps guide

aneta-petrova:

Adding content from the Configuring_virt_who_VM_Subscriptions guide to the Satellite_Documentation_Maps guide.

In some cases, this resulted in minor changes in the Planning guide:

* Minor adjustments to the contents of some of the modules (limited to IDs, headings, abstracts)
* Minor adjustments to the structure of some assemblies

Changes in the Satellite Documentation Maps guide are as follows:

* Extending a few existing top-level jobs with content from the Planning guide
* Introducing a few new top-level jobs to place content from the Planning guide
* Dropping some of the existing top-level jobs (Plan Satellite deployment, Get started with Satellite -- these are replaced with more specific top-level jobs, to avoid duplicating the "Plan" and "Get started" categories)
* New entries in the Reference category

This is part of a downstream documentation reorg. https://redhat.atlassian.net/browse/SAT-50282

Section headings are currently inconsistent in some places: Some sections use imperative headings ("Troubleshoot"), others stick to gerund headings ("Configuring"). That's okay for now, I'll look at making the headings consistent later.

#### Contributor checklists

* [x] Foreman 5.0/Katello 5.0


guides/common/modules/con_smart-proxy-server-features.adoc
ekohl, line 72: I know this was already here, but since I'm reading the docs I'm commenting my notes.

This is not supported in a Katello (and derivatives) installation: the installer will always enable Pulp. I'd love to support that, but the option isn't really usable for most users because it requires hacking some things today. You certainly can't get this by following documentation.

guides/doc-Satellite_Documentation_Maps/maps/understand-administration-tools.adoc
ekohl, line 1: When I look at the content I see 2 categories:
* How to use the application (UI, CLI, API, Ansible collection)
* Features within the application

In the current Satellite docs I only see a single feature rendered, but is it fair to assume the number of features will grow? I'm also wondering if the split I mentioned above should be visible to users. Have you given any thought to where this will grow in the future?

ekohl (commented):
Is there a reference somewhere what the various categories should be?

Discover: I don't have any expectations what this would be, but the use cases is a very strong start. Then the chapter for how to manage it under get started (UI, CLI, API, Ansible) would IMHO also fit well under Discover.

As an open source user: my expectation of Get started will guide me to install the application and get started. For example, https://docs.github.com/en/get-started is basically an index to get you started using the actual product. Same for https://docs.docker.com/get-started/ which is also aimed at actually getting up and running.

Plan makes sense. It's good to connect this back to the Discover parts: if a use case was mentioned as possible then in Plan I'd love to figure out how I can deploy that use case for my needs. The deployment scenarios and workflows fit that well.

Then when I look at the index: the Get started in the middle of Discover and Plan feels odd. Perhaps this was decided elsewhere, but isn't this more logical:

* Get started (more of a landing place) 
* Discover
* Plan
* Install

aneta-petrova:
> Discover: I don't have any expectations what this would be, but the use cases is a very strong start. Then the chapter for how to manage it under get started (UI, CLI, API, Ansible) would IMHO also fit well under Discover.
> 
> As an open source user: my expectation of Get started will guide me to install the application and get started. For example, https://docs.github.com/en/get-started is basically an index to get you started using the actual product. Same for https://docs.docker.com/get-started/ which is also aimed at actually getting up and running.

I like this and think you're right that the tools don't belong in `Get started`. We could try to reserve that category only for tutorials like [Getting started with content management](https://docs.theforeman.org/nightly/Managing_Content/index-satellite.html#getting-started-with-content-management) or perhaps [Getting started with remote execution](https://docs.theforeman.org/nightly/Managing_Hosts/index-katello.html#getting-started-with-remote-execution).

aneta-petrova:
> Then when I look at the index: the Get started in the middle of Discover and Plan feels odd. Perhaps this was decided elsewhere, but isn't this more logical:
> 
>     * Get started (more of a landing place)
> 
>     * Discover
> 
>     * Plan
> 
>     * Install

Just like category names, the way they are ordered is not something we can influence in this PR. Or, more precisely, we can change the order here but that won't be reflected downstream. Because this is a downstream-only build target, changing things arbitrarily here would make the preview misleading because it's not what we'd get downstream.

guides/doc-Satellite_Documentation_Maps/maps/understand-administration-tools.adoc
aneta-petrova: Hm, right. This is taken from the Planning guide where all these sections are grouped together but remote execution is not an administration tool to manage the Satellite server so I guess it doesn't really belong there..?

I might have to place the "Remote execution in Satellite" section somewhere else, probably in the Discover section.

ekohl: Yes, this may be a process where we're forced to deal with a lot of debt in the planning guide. I know you reviewed some sections in the past but I don't think the guide as a whole received the love it deserves.

aneta-petrova: Okay, the module on rex is no longer part of the tools job. I wasn't sure where to place it but the job on architecture seems like a good fit... so that's its new home.

guides/common/modules/con_smart-proxy-server-features.adoc
aneta-petrova: I excluded these lines from katello+orcharhino+satellite. Let me know if this is not what you had in mind.

aneta-petrova:
I think I've implemented all your feedback so far @ekohl. Can you please re-review?

guides/common/modules/con_configuring-provisioning-in-project.adoc
```diff
@@ -1,7 +1,7 @@
 :_mod-docs-content-type: CONCEPT
 
-[id="configuring-provisioning_{context}"]
-= Configuring provisioning
+[id="configuring-provisioning-in-project"]
```
maximiliankolb:
```suggestion
[id="configuring-provisioning-in-{project-context}"]
```

guides/common/modules/con_design-deployment-workflow.adoc
```diff
@@ -0,0 +1,8 @@
+:_mod-docs-content-type: CONCEPT
+
+[id="design-deployment-workflow"]
+= Design deployment workflow
```
maximiliankolb: Your call:

```suggestion
= Design your deployment workflow
```

same for anchor/file name.

guides/common/modules/con_project-ansible-collection.adoc
```diff
@@ -1,7 +1,7 @@
 :_mod-docs-content-type: CONCEPT
 
-[id="automating-{project-context}-management-with-{project-context}-ansible-collection"]
-= Automating {Project} management with {Project} Ansible Collection
+[id="project-ansible-collection"]
```
maximiliankolb:
```suggestion
[id="{project-context}-ansible-collection"]
```

guides/common/modules/con_supported-usage-and-versions-of-project-components.adoc
```diff
@@ -1,7 +1,7 @@
 :_mod-docs-content-type: CONCEPT
 
-[id="Supported-Usage-and-Versions-of-Project-Components_{context}"]
+[id="supported-usage-and-versions-of-project-components"]
```
maximiliankolb:
```suggestion
[id="supported-usage-and-versions-of-{project-context}-components"]
```

guides/common/modules/con_major-project-components.adoc
```diff
@@ -9,3 +9,64 @@ ifdef::katello,orcharhino,satellite[]
 content and
 endif::[]
 configuration from {ProjectServer} and {SmartProxyServers}.
+
+{ProjectServer}::
+{ProjectServer} is the central component of a {Project} deployment.
```
maximiliankolb:
```suggestion
{ProjectServer} is the central component of your {Project} deployment.
```

That would avoid "a orcharhino".

guides/common/modules/con_major-project-components.adoc
```diff
@@ -9,3 +9,64 @@ ifdef::katello,orcharhino,satellite[]
...
+* Configuration of {SmartProxyServers}
+* Configuration of hosts
+* Host provisioning
+ifdef::katello,orcharhino,satellite[]
+* Patch management
+* Subscription management
+endif::[]
++
+ifdef::foreman-deb,foreman-el[]
+{ProjectServer} delegates host provisioning and communication to {SmartProxyServers}.
```
maximiliankolb, line 30: tech wise, I am unsure. I assume that only Foreman Server talks to the compute resource to request new hosts. If you use Kickstart content and a Smart Proxy Server as a content source, then yes, Smart Proxy Servers are involved.

non-blocking/maybe out of scope; I think it's fine but maybe not 100% correct.

guides/common/modules/ref_expense-management-additional-resources.adoc
maximiliankolb, line 1: out of scope: do we really need this as a separate module?

guides/common/modules/ref_key-open-source-components-of-project.adoc
```diff
@@ -1,7 +1,7 @@
 :_mod-docs-content-type: REFERENCE
 
-[id="list-of-key-open-source-components-of-{project-context}"]
-= List of key open source components of {Project}
+[id="key-open-source-components-of-{project-context}"]
+= Key open source components of {Project}
 
 [role="_abstract"]
 {Project} consists of several open source projects integrated with each other, such as the following:
```
maximiliankolb: AFAIK the abstract cannot end with a colon.

guides/common/modules/ref_supported-usage-of-project-components.adoc
```diff
@@ -1,6 +1,6 @@
 :_mod-docs-content-type: REFERENCE
 
-[id="supported-usage-of-project-components_{context}"]
+[id="supported-usage-of-project-components"]
```
maximiliankolb:
```suggestion
[id="supported-usage-of-{project-context}-components"]
```

guides/doc-Satellite_Documentation_Maps/maps/understand-administration-tools.adoc
```diff
@@ -0,0 +1,13 @@
...
+
+include::../../common/modules/con_web-ui-overview.adoc[leveloffset=+1]
+
+include::../../common/modules/con_hammer-cli-overview.adoc[leveloffset=+1]
+
+include::../../common/modules/con_project-api-overview.adoc[leveloffset=+1]
+
+include::../../common/modules/con_project-ansible-collection.adoc[leveloffset=+1]
+
+include::../../common/modules/con_automated-installation-with-kickstart.adoc[leveloffset=+1]
```
maximiliankolb, line 13: out of scope: this module feels a bit odd to me in this map.

aneta-petrova: @ekohl Do you have any advice on what to do with the kickstart installation module? It's been in the [Planning guide](https://docs.theforeman.org/3.19/Planning_for_Project/index-katello.html#Kickstart-Workflow_planning) for years and I agree it feels odd there (and here). Is it still useful? Should we remove it? Should we rewrite it?

guides/common/modules/con_major-project-components.adoc
aneta-petrova: @ekohl Do you have any opinions on this? It's what the module currently says in [Planning](https://docs.theforeman.org/3.19/Planning_for_Project/index-katello.html#Major-Foreman-Components_planning) so this is not a new change.

guides/common/modules/ref_expense-management-additional-resources.adoc
aneta-petrova: It does make sense when looking at the Planning guide's table of contents:

Image

It's not what we typically do in this repo but it serves its purpose and is not wrong... so no change in this PR.

maximiliankolb (approved):
style-wise LGTM

guides/doc-Satellite_Documentation_Maps/maps/understand-administration-tools.adoc
```diff
@@ -0,0 +1,15 @@
...
+
+include::../../common/modules/con_project-api-compared-to-hammer-cli.adoc[leveloffset=+1]
```
aneta-petrova, line 15: FYI @jsvarova, this is one of the modules from the API guide.

guides/common/modules/con_project-api-overview.adoc
```diff
@@ -0,0 +1,32 @@
+:_mod-docs-content-type: CONCEPT
```
aneta-petrova, line 1: FYI @jsvarova This module now contains information from the API overview module in the API guide. (Turns out we had two API overview concepts! This one now includes information from both, to avoid duplication in the Maps guide.)

jafiala (approved):

lzap:
First of all, I was never fan of the multiple books approach, so I welcome the change, I will say the timing is a bit unfortunate. With that, I thought about it from the upstream user and contributor perspective. I created myself a structure of all chapters available including all upstream chapters (which might not be relevant for Satellite):

https://gist.github.com/lzap/ab526cb02470d08fc39b0209b4bd70ae

The proposal (the current PR) currently renders to:

What’s new → Discover → Get started → Plan → Install → Deploy → Upgrade → Migrate → Disconnected environments → Administer → Develop → Configure → Secure → Observe → Integrate → Optimize → Extend → Troubleshoot → Reference

I think the most important goal at this point is to get the top-level list of chapters right. So I took the approach of reviewing what is currently proposed in the PR and it is looking good, I propose the following change:

Learn → Quickstart → Plan → Install → Upgrade → Backup → Restore → Migrate → Integrate → Secure → Provision → Patch → Automate → Observe → Optimize → Reference → Contribute

Not only I added, renamed and dropped some chapters, I also carefully reordered them in the direction how I believe Satellite customers deal with things. We should keep the similar order in sub-chapters if that is not against the new rules and where it makes sense, specifically in the first one. While the order in JTBD might not be significant, at least it helped me to sort out my thoughts and there will be some users reading the docs top to bottom.

Let me comment on each mentioned chapters, old and new:

* Learn - Instead of Discover, which is a term widely used in Satellite (discovery of hosts), I suggest to use a different term. This chapter should explain a LOT of content, Satellite covers a lot of ground and we should cherry pick many bits from various books including subscription, content, provisioning, automation, CLI. I mentioned many times that our documentation is good explaining HOW but weaker in delivering WHY. This is a challenging one.
* Get Started - I do not like this, if I understand the motivation correctly nobody wants to "get started" they want to solve things. This is out.
* Quickstart - this is new, we have this upstream and I believe this is a very important one. We serve a simplified installation instructions for those who just want to take a look and deploy a POC environment without too much thinking about it. We do not explain anything here, everything is just defaults, get started. Probably replaces Get Started from above, the name Quickstart might be different but the idea should stay.
* Plan - very important part because Satellite is not an iPhone that can be set up in an hour. Probably an easier one since we have the planning guide.
* Install - the ultimate chapter that most of readers reaches out for. We already have a book so this could be 1:1 mapping. This describes the installation in full including all the options. This must also include disconnected environments, not sure why there is Disconnected environments in the top-level in the proposal, that is probably an overlook.
* Upgrade - 1:1 mapping. I dropped Deploy, not sure what it was meant for.
* Backup/Restore - I intentionally promoted these to the top level, I think generally speaking we should aim for even more top-level chapters but that can be done later. Backups are extremely important and should be considered very early, ideally after installation.
* Migrate - also important job, not only we should cover Satellite/Satellite migration, but also database migration (to external) or renaming smart proxies as well as renaming certificates possibly.
* Integrate - this is a tough one. Initially I thought it is a good idea, it can cover things like LDAP/MSAD/IPA/SSO (originally security section) and DNS/DHCP/TFTP/PXE (originally in provisioning). I also considered moving Compute Resources here. Generally speaking, this would cover the setup phase while operations would be left in the original chapters. Probably too much work for the initial phase, consider doing this as a followup.
* Secure - initially, this can contain a lot from Administering guide (orgs, locs, users, roles, permissions, audit, certificates). Later we can also duplicate some hardening stuff from installation guide.
* Provision - 1:1 mapping with the guide
* Patch - I think we need to have this job, a customer deployed systems now a security bug appears, how do I patch my systems, quick? Maps a lot of chapters from the Content Guide.
* Automate - similar story, less urgent but deserves to be a top-level job. Covers Configuration Guide.
* Observe - this is a good call, we can collect a lot of chapters from various books in here. Generally speaking: audit, host reports, job reports, facts, metrics, monitoring, PCP
* Optimize - again a good one already selected, this will be also a mix of various chapters: tuning, reducind storage, reconfiguring for better hardware
* Troubleshoot - I am not fan of any kind of FAQ, one clever colleague of mine once said: once you have FAQ section in your documentation, you already failed. Dropped.
* Extend - not sure what the plans here are, dropped.
* Reference - yes, we need that
* Contribute - this might be upstream only section, however, now that I see it, we have many contributions from customers as well so why not.

Note there is no Administer top-level chapter, I think this is a dangerous one, a "catch-all" term that we should generally avoid.

I think that before diving into actual docs changes, we should agree on a mapping on all chapters in some format, could be an Excel or just a Markdown table. Therefore I headed this direction too and with LLM I created a mapping table suggestion where my goal was to validate the top-level chapters. This is NOT how I envision the mapping, this is merely a preview of the fact that each chapter gets an acceptable home. Results are not great, LLM shuffled some chapters aggressively but as an inspiration of how I would like to talk about the mapping going forward:

https://gist.github.com/lzap/7d138999e066ff8008f7239cd617568b

I think it makes no sense to talk about this over individual PRs, that is way too late.

aneta-petrova:
Thanks so much, @lzap, for your time and sharing your expertise. Under any other circumstances, I'd love to get into a conversation about what makes sense and what doesn't, how to best explain things to the user, etc. However, a large part of what you're suggesting to change is determined centrally downstream and therefore we cannot change it here.

For example, if you'd like to see different categories ("Learn" instead of "Discover", for example), that's a conversation we'd have to start downstream first and it would affect all of our other products; it's not something to be determined here in this PR.

I'll reach out to your in a direct message to talk further about what needs to stay the way it is currently set up in the preview on nightly (https://docs.theforeman.org/nightly/Satellite_Documentation_Maps/index-satellite.html) and what we can improve in this PR.

lzap (approved):
I do not think the the top-level list of chapters provides a good experience for users of the product. I believe this is a work in progress and can be changed.

ekohl:
> Integrate - this is a tough one. Initially I thought it is a good idea, it can cover things like LDAP/MSAD/IPA/SSO (originally security section) and DNS/DHCP/TFTP/PXE (originally in provisioning). I also considered moving Compute Resources here. Generally speaking, this would cover the setup phase while operations would be left in the original chapters. Probably too much work for the initial phase, consider doing this as a followup.

This is where our REST API guide and Foreman Ansible Collection documentation would fit in well. Webhooks can also live here.

> Troubleshoot - I am not fan of any kind of FAQ, one clever colleague of mine once said: once you have FAQ section in your documentation, you already failed. Dropped.

I still like a section where I see the error codes and that helps guide me to a solution. Ideally the application itself helps with that, but it's not always realistic.

ekohl:
Also deferring to @lzap's actual review since I don't have the time to properly review it myself. I largely recognize his concerns.

aneta-petrova:

f1145a7a82..4aab7214cd  5.0 -> 5.0

aneta-petrova:
> I do not think the the top-level list of chapters provides a good experience for users of the product.

Because the top-level list of chapters is centrally determined downstream, I'll bring up these concerns downstream.

> I believe this is a work in progress and can be changed.

I can confirm this. We're still only beginning to build the new docs structure and getting parts of the plan in will help give all of us a better picture of the end result should be. Please do let me know if any additional concerns emerge later.

PR #5309, opened 2026-09-11
Map Chapters 1-3 of Mng hosts

Lennonka:

Adding the following chapters of [the _Managing hosts_ guide](https://docs.theforeman.org/nightly/Managing_Hosts/index-satellite.html):

- Reviewing hosts in Satellite web UI
- Managing host placement and associations
- Managing host content settings and system purpose

...to the doc-Satellite_Documentation_Maps

JTBD restructuralization of Satellite docs from guides into a single TOC

- The `:context:` definitions are necessary for now, otherwise the build fails
- Relevant to the Satellite build only

#### Contributor checklists

* [x] Foreman 5.0/Katello 5.0

### DIFF
--- a/guides/doc-Satellite_Documentation_Maps/maps/assign-hosts-to-logical-groups.adoc
+++ b/guides/doc-Satellite_Documentation_Maps/maps/assign-hosts-to-logical-groups.adoc
@@ -1,0 +1,14 @@
+:_mod-docs-content-type: MAP
+:context: assign-hosts-to-logical-groups
+
+include::../../common/modules/con_assign-hosts-to-logical-groups.adoc[]
+
+include::../../common/modules/proc_assigning-a-host-to-a-specific-organization.adoc[leveloffset=+1]
+
+include::../../common/modules/proc_assigning-a-host-to-a-specific-location.adoc[leveloffset=+1]
+
+include::../../common/modules/proc_changing-the-content-view-environments-of-a-host.adoc[leveloffset=+1]
+
+include::../../common/modules/proc_associating-a-virtual-machine-from-a-hypervisor.adoc[leveloffset=+1]
+
+include::../../common/modules/proc_disassociating-a-vm-from-project-without-removing-it-from-a-hypervisor.adoc[leveloffset=+1]
--- a/guides/doc-Satellite_Documentation_Maps/maps/configure-system-purpose-for-subscription-reporting.adoc
+++ b/guides/doc-Satellite_Documentation_Maps/maps/configure-system-purpose-for-subscription-reporting.adoc
@@ -1,0 +1,10 @@
+:_mod-docs-content-type: MAP
+:context: configure-system-purpose-for-subscription-reporting
+
+include::../../common/modules/con_configure-system-purpose-for-subscription-reporting.adoc[]
+
+include::../../common/modules/proc_editing-the-system-purpose-of-a-host-by-using-web-ui.adoc[leveloffset=+1]
+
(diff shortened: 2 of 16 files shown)


guides/doc-Satellite_Documentation_Maps/maps/assign-hosts-to-logical-groups.adoc
```diff
@@ -0,0 +1,14 @@
+:_mod-docs-content-type: MAP
+:context: assign-hosts-to-logical-groups
+
+include::../../common/modules/con_assign-hosts-to-logical-groups.adoc[]
+
+include::../../common/modules/proc_assigning-a-host-to-a-specific-organization.adoc[leveloffset=+1]
+
+include::../../common/modules/proc_assigning-a-host-to-a-specific-location.adoc[leveloffset=+1]
+
+include::../../common/modules/proc_changing-the-content-view-environments-of-a-host.adoc[leveloffset=+1]
```
Lennonka, line 10: I think this should go to "Configure which content hosts consume".

aneta-petrova:
We also have pre-existing top-level jobs Patch and update hosts and Organize Satellite resources for management. (There is also Organize resources by geography but please ignore that, that's a halluciation and is being fixed in https://github.com/theforeman/foreman-documentation/pull/5304). They seem very related to the top-level jobs that you're adding, to the point where your jobs perhaps might actually be sub-jobs of these pre-existing jobs?

* Assign hosts to logical groups seems like something that would be part of Organize Satellite resources for management.
* Configure which content hosts consume seems like something that would be part of Patch and update jobs.

There might be others, these two are just what jumped out at me. Have you considered including the content you're migrating as sub-jobs rather than top-level jobs?

guides/doc-Satellite_Documentation_Maps/maps/assign-hosts-to-logical-groups.adoc
```diff
@@ -0,0 +1,14 @@
+:_mod-docs-content-type: MAP
+:context: assign-hosts-to-logical-groups
```
aneta-petrova, line 2: Is this because the included modules have `_{context}` in their IDs? How about instead removing those `_{context}` elements from those IDs? It's what we've started doing across the repo. It might be cleaner than adding additional attributes like this to maps (the other maps that have been committed to the Maps guide are not using this).

guides/doc-Satellite_Documentation_Maps/maps/configure-system-purpose-for-subscription-reporting.adoc
```diff
@@ -0,0 +1,10 @@
+:_mod-docs-content-type: MAP
```
aneta-petrova, line 1: There is a pre-existing top-level job "Track subscription usage to manage spending". It seems like this would be part of that, doesn't it?

guides/doc-Satellite_Documentation_Maps/maps/view-all-hosts-in-a-central-location.adoc
```diff
@@ -0,0 +1,8 @@
+:_mod-docs-content-type: MAP
```
aneta-petrova, line 1: At first glance, this does not seem to match the description of the "Discover" category: `Orient me to what the product is, what it does, and when I should use it. | Understand whether this product solves my problem, what it can do, and how it conceptually works. | High-level orientation, value proposition, key capabilities, conceptual diagrams (not architectural choices).` Procedures don't seem to be a good fit for a category like that, and perhaps users wouldn't even have anything installed yet at this point.

Lennonka:
@aneta-petrova Thank you for feedback! I'll revisit it next week.

PR #5314, opened 2026-09-14
Include RH subscription management in containerized

aneta-petrova:

Adding assembly on RH subscription management to containerized guides

To extend the containerized docs set

https://redhat.atlassian.net/browse/SAT-47721

N/A

#### Contributor checklists

* [x] Foreman 5.0/Katello 5.0


guides/common/modules/con_managing-red-hat-subscriptions.adoc
```diff
@@ -34,4 +36,3 @@ ifdef::satellite[]
 * {InstallingServerDisconnectedDocURL}configuring-{project-context}-server-to-consume-content-from-a-custom-cdn-by-using-web-ui[Configuring {ProjectServer} to consume content from a custom CDN in _{InstallingServerDisconnectedDocTitle}_]
 * https://access.redhat.com/solutions/7065718[How to register and subscribe a system offline to the Red Hat Customer Portal?]
 endif::[]
-* xref:tracking-subscription-usage-by-using-the-subscriptions-service[]
```
Lennonka, line 37: `ifndef::containerized[]` ?

aneta-petrova: This is intentional. The target section follows immediately after the one that includes the xref. I don't think that's a standard structure for assemblies -- that a module should link to a module that immediately follows in the ToC.

Lennonka: All good. Thank you for clarification!

Lennonka (approved):
LGTM style wise

guides/common/modules/proc_tracking-subscription-usage-by-using-the-subscriptions-service.adoc
```diff
@@ -8,26 +8,28 @@ You can configure your {ProjectServer} to report usage data to the {RHCloud}.
 Track and manage your subscription usage to analyze historical trends for planning and compliance.
 
 ifndef::satellite[]
-.Prerequisites
-* The `foreman_rh_cloud` plugin is enabled on {ProjectServer}.
-You can enable the plugin by running `{foreman-installer}`:
+.Procedure
```
aneta-petrova, line 11: Making sure the steps in the procedure render correctly for all builds was tricky. The result that I'm proposing is:

* For Satellite builds, there is a single step procedure where the step is a number (`1.`) rather than a bullet point (`*`). This is because eventually, the second step in the procedure will also get containerized and the number (`1.`) will be correct. So this divergence from style guidelines is only temporary and the alternative would be to use a lot of conditionals to turn that `1.` into `*` (IMO not worth it)
* For containerized and non-containerized Katello builds, the procedure includes only the step for connected servers (because disconnected is not available in Katello builds).
* The disconnected step is actually not shown in either build right now; but I'd like to keep it in the source because we expect for it to be added eventually/soon (after the containerized solution becomes available).

aneta-petrova:
Hi @jeremylenz, can you please review this update?

jeremylenz (approved):
Checked 3 previews (index-satellite, index-katello, index-containerized-katello) and they all look correct. Tech ACK 👍

aneta-petrova:
Thanks @jeremylenz! We'll also need a testing ack for this PR. Could you perhaps check whether the functionality described in the chapter is covered by automated tests for containerized deployments? Or feel free to use any other way to ensure this is tested based on your team's preference.

guides/common/modules/proc_tracking-subscription-usage-by-using-the-subscriptions-service.adoc
```diff
@@ -37,9 +39,9 @@ You can generate this report by using CLI:
 ----
 # foreman-rake rh_cloud_inventory:report:generate
```
jeremylenz, line 40: We will no longer recommend or allow generating inventory reports via foreman-rake. See https://github.com/theforeman/foremanctl/pull/825

I think this will need to be replaced with API instructions, or ideally, a command added to Hammer that makes it easier.

jeremylenz (changes requested):

guides/common/modules/proc_tracking-subscription-usage-by-using-the-subscriptions-service.adoc
Lennonka: This can be done via web UI, no?

jeremylenz: actually, yes, web UI would be the most user-friendly!

aneta-petrova: This is the part that is currently excluded for containerized because the functionality is not available in containerized deployments. When I last inquired about this particular foreman-rake task, I found out that not even the web UI counterpart would work in containerized. Has that changed? If not, I believe we can leave this excluded for containerized guides.

jeremylenz: @nofaralfasi After your PR merge, web UI download of the reports should work correctly, right?

nofaralfasi: Correct.

aneta-petrova: I think it's past feature freeze, is that right? Will https://github.com/theforeman/foremanctl/pull/825 make it into 5.0?

jeremylenz: We can go past feature freeze for bugs. Depends how you look at it :smile:

jeremylenz: This is blocking basic functionality, so I view it more as a bug

PR #5332, opened 2026-09-17
Update NFS pulp note for foremanctl

lzap:
We currently run pulp containers unconfined because of the issue with Gunicorn:

https://github.com/benoitc/gunicorn/pull/3747

Since the pulp policy is not installed, the label for NFS client cannot be `pulpcore_var_t` it needs to be the OS default.

Once we figure out the fix for Gunicorn/Pulp, we can launch the services in enforcing and then this would become `container_file_t:s0` (must be mounted with `-z` not `-Z`).


evgeni (approved):

Lennonka (approved):
LGTM style wise

Lennonka:
I'm assuming we won't be testing this.

evgeni:
Yeah, I think not explicitly testing this is fine.

Lennonka:
Any cherry picks? I would expect this in 5.0. Please, confirm.

evgeni:
Yes, 5.0 please.

Lennonka:
Cherry picked:
- 12ff347a4c..ad8dc13404  5.0 -> 5.0

PR #5333, opened 2026-09-17
Add content-related content from Managing Content to Maps guide

aneta-petrova:
#### Contributor checklists

* [x] Foreman 5.0/Katello 5.0


guides/common/assembly_curate-and-filter-content-views.adoc
```diff
@@ -0,0 +1,101 @@
+include::modules/con_curate-and-filter-content-views.adoc[]
```
aneta-petrova, line 1: I'll need to verify if this is a good heading for a sub job: it probably shouldn't mention a feature (CV)

guides/common/assembly_distribute-content-to-hosts.adoc
```diff
@@ -0,0 +1,23 @@
+include::modules/con_distribute-content-to-hosts.adoc[]
+
+include::assembly_content-access-control-for-hosts.adoc[leveloffset=+1]
+
+include::assembly_managing-activation-keys.adoc[leveloffset=+1]
+
+include::modules/con_managing-alternate-content-sources.adoc[leveloffset=+1]
```
aneta-petrova, line 7: ACS might fit better under Optimize (per https://www.redhat.com/en/blog/synchronizing-software-between-red-hat-satellite-servers )

aneta-petrova:
A quick list of notes from my meeting with @parthaa and @ianballou about this part of the docs reorg:

* We discussed whether the main chunk of content management docs might be a better fit for the Administer category rather than Configure. (Conclusion: Keep this in mind, but let's not change it right now. After reviewing the descriptions of categories, Configure does seem like a slightly better fit.)
* Configure might be better listed before Administer. This is also related to the point above because the content management actions described in Configure would typically be performed before Administer. (Conclusion: I'll try to propose changing the order of categories downstream because it's determined centrally downstream and changing it here wouldn't have any effect.)
* Add a cross link from the main content job to 13.1.1. Authenticating and securing content with content credentials
* Add cross link also to 16.2. Optimize content synchronization for bandwidth
* "Storing content on an NFS share" is not a good fit for the Optimize performance for scale top-level job (or the Optimize category) but right now we're not 100% clear on what the right category should be. Perhaps Install.

aneta-petrova:
>     * Configure might be better listed before Administer. This is also related to the point above because the content management actions described in Configure would typically be performed before Administer. (Conclusion: I'll try to propose changing the order of categories downstream because it's determined centrally downstream and changing it here wouldn't have any effect.)

I asked downstream and will provide an update when/if I have one.

>     * Add a cross link from the main content job to 13.1.1. Authenticating and securing content with content credentials
> 
>     * Add cross link also to 16.2. Optimize content synchronization for bandwidth

Added both cross links.

>     * "Storing content on an NFS share" is not a good fit for the Optimize performance for scale top-level job (or the Optimize category) but right now we're not 100% clear on what the right category should be. Perhaps Install.

No change yet, this might become clearer as the PR review continues here.

